PR #116, opened 2013-05-14
Update to MongoClientURI javadoc

ramazanpolat:
URI can contain collection name which is not mentioned in javadoc. Fixed.


grossws:
Is this dropped later by accident in current `MongoClientURI` or `collection` will be dropped eventually from it?

PR #141, opened 2013-09-05
Allow access to setGridFS

rozza:
Allows for custom implementations of GridFS.  One usecase is to use JodaDateTime for uploaded date, instead of java date.


rozza:
Closing - in casbah we still can override `setGridFS` and use the super to set the db.

PR #150, opened 2013-10-12
update to GridFSFile.java: putAll method is implemented

ramazanpolat:
putAll method is implemented.
So the user can provide a Map.
Each key-value pair will be saved as file property.


rozza:
The new GridFS implementation allows for arbitrary metadata via a {{Document}} which has this functionality.

The legacy implementation is effectively deprecated as such I'm closing this ticket.

PR #151, opened 2013-10-13
Update to GridFSFile: toMap method is implemented.

ramazanpolat:
toMap method is implemented.
it puts all fields and corresponding values in the files collection to a HashMap object.


rozza:
Apologies - for the lack of recent feedback.  The new GridFS implementation allows for arbitrary metadata via a {{Document}} which is an implementation of a Map and should meet your requirements.

The legacy implementation is effectively deprecated as such I'm closing this ticket.

PR #194, opened 2014-05-08
Improved covariance of DBCollection.insert

minisu:
Changed signature of the insert method from `insert(List<DBOject>)` to `insert(List<? extends DBOject>)`. This is needed to (for example) enable insert to accept a `List<BasicDBObject>`.


rozza:
Apologies for not coming back to this ticket sooner, I'm pleased to say this functionality was added in the 3.x series of the driver and your contribution help make it so!  Many thanks.

PR #224, opened 2014-07-08
ObjectId serialization optimizations

hoffrocket:
- ObjectId(byte[]) reads directly from array rather than wrapping in a
  ByteBuffer
- toString/fromString using fewer intermediate objects
- new ObjectId(ByteBuffer) constructor

### DIFF
--- a/src/main/org/bson/types/ObjectId.java
+++ b/src/main/org/bson/types/ObjectId.java
@@ -18,6 +18,7 @@
 
 package org.bson.types;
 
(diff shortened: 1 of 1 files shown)


jyemin:
The ObjectId class has changed substantially in master since this was submitted.  Any interest in trying to re-apply your changes to the latest on master?

rozza:
Closing, much has changed since this initial patch with the new 3.x series of the driver.  As such this no longer can be applied cleanly.

PR #230, opened 2014-07-21
Re-modularizing 3.0.x

jyemin:
This is still a work in progress, but I'd like to get the first part pushed so that others can contribute on top.  Themes:
1. Removed almost all duplication of exceptions and moved the com.mongodb ones into driver core
2. Simplified CommandResult and removed dependencies on it from exceptions. 
3. Started removing other duplication, including ServerAddress, ReadPreference, WriteConcern, and MongoCredential. Moved com.mongodb ones into driver core.
4. Renamed org.mongodb.<package packages to com.mongodb.<package>.  The org.mongodb package itself is the only one left.


nbkhope:
Could someone clarify why we declare a dependency like `compile 'org.mongodb:mongodb-driver:3.6.0'` in gradle using org.mongodb, but import the packages uses `com.mongodb...`?

What's the difference between org.mongodb and com.mongodb and why did you change it?

jyemin:
It's historical.  Event before this commit,  the driver's packages have always been prefixed with `com.mongodb` but the Maven groupId has always been `org.mongodb`.

PR #234, opened 2014-08-13
Expose the ServerType and ClusterType enumerations for public consumption

luketn:
Expose the ServerType and ClusterType enumerations for public consumption. Re-use the private method that was used by the ServerMonitor class to determine the ServerType for a DB or MongoClient.


rozza:
Apologies for not coming back to this ticket sooner.  [JAVA-805](https://jira.mongodb.org/browse/JAVA-805) is still open and followups should be continued there.

PR #238, opened 2014-08-13
Create Jackson-based implementation of Codec

guoyr:
Create a Codec implementation on top of the Jackson JSON library as proof that the abstraction is correct and so that there is at least one functional POJO codec that people can use with the 3.0 driver's generic collection API.

[JAVA-1314](https://jira.mongodb.org/browse/JAVA-1314?filter=-1)


jyemin:
This pull request doesn't compile, and when I fix the compilation errors many of the tests fail.

PR #260, opened 2014-09-26
Utility functions to view state of BulkWriteOperation objects

mskott:
BulkWriteOperation.execute() will throw an exception if the operations have all ready been executed or no operations have been added to the object, but it doesn't provide any way to test if this is the case. I have added two simple methods which make it possible.

For me the practical use-case is that we often create have thousands of operations to run in bulk, but would like to run execute when n have been added. We can easily do this by adding an external counter, but it makes much more sense for the BulkWriteOperation to make this information available. This also removes the need for code like the one below when no operations have been added to the object.

    try {
       bulkOp.execute();
    } catch (IllegalStateException ise) {
       //Do nothing
    }

### DIFF
--- a/src/main/com/mongodb/BulkWriteOperation.java
+++ b/src/main/com/mongodb/BulkWriteOperation.java
@@ -112,5 +112,12 @@ public class BulkWriteOperation {
         requests.add(request);
     }
 
+    /**
+     * Returns true if the bulk operations have been executed.
+     * @return whether the bulk operations have been executed
+     */
+    public boolean isExecuted() {
+    	return closed;
+    }
 
 }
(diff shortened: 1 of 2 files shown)


rozza:
Hi, apologies for not coming back to this ticket sooner.  As you may be aware much has changed in the codebase with the new and improved 3.x series.

The `DB`/`DBCollection` code path has been deprecated in favour of the new `MongoDatabase`/`MongoCollection` path.  Bulk operations now follow the [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst).  And for those reasons I'm closing this ticket.

PR #269, opened 2014-10-24
parse a list of ServerAddress

twillouer:
Hi,

just an helper to have multiple ServerAddress from a simple String.


rozza:
Hi, apologies for not coming back to this ticket sooner.  As you may be aware much has changed in the codebase with the new and improved 3.x series.  This ticket can no longer be merged, so its being closed.

Currently the standardized way of dealing with multiple hosts in a string is to use a connection string.  Adding an alternative way is not currently planned, as it simple enough to do with a quick loop.

PR #282, opened 2014-11-18
JAVA-1529: JSON Parser now does not allow unmatched trailing quotes on field names

arpitbbhayani:
1. JSONTest.java file to include 2 tests 
           JSON.parse("{ a\": 1 }");
           JSON.parse("{ a: { b\":1 } }");
Both of the above test cases should throw JSONParseException because of unmatched quote.

2. Added a snippet of code in file JSONCallback.java in the objectDone() method to fix the given issue.

### DIFF
--- a/src/main/com/mongodb/util/JSONCallback.java
+++ b/src/main/com/mongodb/util/JSONCallback.java
@@ -35,10 +35,7 @@ import org.bson.types.ObjectId;
 
 import java.text.ParsePosition;
 import java.text.SimpleDateFormat;
-import java.util.Date;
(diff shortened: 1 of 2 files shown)


rozza:
Hi, apologies for not coming back to this ticket sooner. As you may be aware much has changed in the codebase with the new and improved 3.x series. This ticket can no longer be merged, so its being closed.

In 3.x `JSON.parse` is effectively deprecated and `Document.parse` or `BasicDBObject.parse` is preferred.  These new methods will throw an exception with the invalid json.

PR #285, opened 2014-12-01
[JAVA-1390] Phase 2: doing the release. 

trishagee:
Working release process.  It fits together end to end but it's not pretty, and could use some tidy up, however I thought it best to get some feedback on it as it is. The last couple of stages were more complicated than I expected.

You can test the release by running ./gradlew release -DreleaseVersion=3.0.0

Roll it back by removing the 3.0.0 tag, deleting the draft release notes and resetting the current branch back two commits.


jyemin:
There were a few things that didn't work properly in this pull request.  In particular, it didn't properly sign all the artifacts and Sonatype rejected the upload as a result.  I was unable to get signing to work properly with a multi-project Gradle build using the new Gradle publish plugin.  The only way I got it to work was using the old nexus plugin.

PR #286, opened 2015-01-02
[JAVA-1282] basic async GridFS support

st-h:
Actually I was hoping that the 3.x driver would be released before I would need more sophisticated things like GridFS support. As this hasn't been the case, I started to write some GridFS support for the async driver.

Obviously there is quite some work left to do (probably add some tests? :) , however I wanted to reach out early to get some feedback if this will be of any use to you before putting more time into it.


jyemin:
Steve, 

This will definitely be useful to us, but will have to leave it on the back burner for a bit, as we are working on a new cross-driver GridFS spec, and I don't want to add any new GridFS APIs before that spec is finalized.

Also, definitely need some tests ;)


Jeff

driver-async/src/main/com/mongodb/async/client/gridfs/GridFS.java
```diff
@@ -0,0 +1,205 @@
...
+import org.bson.Document;
+import org.bson.types.ObjectId;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Objects;
+
+
+public class GridFS {
```
jyemin: Let's follow the convention of MongoDatabase/Collection and make this an interface.

driver-async/src/main/com/mongodb/async/client/gridfs/GridFS.java
```diff
@@ -0,0 +1,205 @@
...
+
+    private final int chunkSize;
+    private final MongoDatabase db;
+    private final MongoCollection<GridFSFile> fileCollection;
+    private final MongoCollection<Document> chunkCollection;
+    private final String bucketName;
+
+    public GridFS(final MongoDatabase db, final String bucket, final Integer chunkSize) {
+
+        Objects.requireNonNull(db);
```
jyemin: While it would be nice to be able to use the Objects class, we have decided not to rely on any Java 7 APIs except AsynchronousSocketChannel (which can be replaced by Netty).

driver-async/src/main/com/mongodb/async/client/gridfs/GridFS.java
```diff
@@ -0,0 +1,205 @@
...
+        this.fileCollection = this.db.getCollection(bucketName  + ".files", GridFSFile.class);
+        this.chunkCollection = this.db.getCollection(bucketName + ".chunks");
+
+        // from sync driver...
+        // ensure standard indexes as long as collections are small
+        fileCollection.count(new SingleResultCallback<Long>() {
+            @Override
+            public void onResult(final Long result, final Throwable t) {
+                if (t != null) {
+                    //TODO log error
```
jyemin: We have a logger class available: com.mongodb.diagnostics.logging.Loggers

driver-async/src/main/com/mongodb/async/client/MongoClientImpl.java
```diff
@@ -48,7 +49,8 @@
 
     private static final RootCodecRegistry DEFAULT_CODEC_REGISTRY = new RootCodecRegistry(asList(new ValueCodecProvider(),
                                                                                                  new DocumentCodecProvider(),
-                                                                                                 new BsonValueCodecProvider()));
+                                                                                                 new BsonValueCodecProvider(),
+                                                                                                 new GridFSFileCodecProvider()));
```
jyemin: Better to figure out a way to add this provider to the registry in the MongoDatabase used by the GridFS instance.

driver-async/src/main/com/mongodb/async/client/gridfs/GridFS.java
```diff
@@ -0,0 +1,205 @@
...
+    String getBucketName() {
+        return bucketName;
+    }
+
+    /**
+     * Get db.
+     *
+     * @return com.mongodb.async.client.MongoDatabase
+     */
+    MongoDatabase getDb() {
```
jyemin: Rename to getDatabase

driver-async/src/main/com/mongodb/async/client/gridfs/GridFSFileCodecProvider.java
```diff
@@ -0,0 +1,17 @@
+package com.mongodb.async.client.gridfs;
+
+import org.bson.codecs.Codec;
+import org.bson.codecs.configuration.CodecProvider;
+import org.bson.codecs.configuration.CodecRegistry;
+
+public class GridFSFileCodecProvider implements CodecProvider {
```
jyemin: Probably can use org.bson.codecs.configuration.RootCodecRegistry#withCodec instead of needing this class at all.

driver-async/src/main/com/mongodb/async/client/gridfs/GridFSFileCodec.java
```diff
@@ -0,0 +1,108 @@
...
+import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.types.ObjectId;
+
+import java.util.Date;
+
+public class GridFSFileCodec implements CollectibleCodec<GridFSFile> {
+
+    private static final String ID_FIELD_NAME = "_id";
+
+    private final CodecRegistry registry;
```
jyemin: Unused field

driver-async/src/main/com/mongodb/async/client/gridfs/GridFS.java
```diff
@@ -0,0 +1,205 @@
...
+            }
+        });
+    }
+
+    /**
+     * List all GridFSFiles, restricted by the provided query.
+     * @param query a document to limit the results that are returned
+     * @param callback callback that will be completed when the result is available
+     */
+    public void listFiles(final Document query, final SingleResultCallback<List<GridFSFile>> callback) {
```
jyemin: How about returning a MongoIterable<GridFSFile> instead, since the list could be rather large in some use cases?

driver-async/src/main/com/mongodb/async/client/gridfs/GridFS.java
```diff
@@ -0,0 +1,205 @@
...
+        });
+    }
+
+    /**
+     * Delete a GridFSFile by its id.
+     *
+     * @param id the id as ObjectId
+     * @param callback callback that is completed when the file is deleted
+     */
+    public void delete(final ObjectId id, final SingleResultCallback<DeleteResult> callback) {
```
jyemin: Since we know we're deleting a single file, DeleteResult is not useful.  Better to return Void

driver-async/src/main/com/mongodb/async/client/gridfs/GridFS.java
st-h: Is this what you have in mind?

````java
    public MongoIterable<GridFSFile> find(final Object query) {
        return fileCollection.find(query).sort(new Document("filename", 1));
    }
````

This brings up one of the things I wanted to discuss:
The GridFSFile class needs a reference to the MongoCollections it uses. I am currently using a similar workaround as the sync driver ( by calling the injectContext() method after the GridFSFile has been retrieved, which sets the required references).
I think it would be a lot nicer, if we could access the MongoCollection (and MongoDatabase) a Document has been retrieved from through the DecoderContext in the Codec. That way these fields could be initialized when the Object is created in the Codec. However, I am currently not sure if this is something the DecoderContext is meant for or if this has any other impact.

driver-async/src/main/com/mongodb/async/client/gridfs/GridFS.java
st-h: Hm, that currently appears to be consistent with the MongoCollection interface:

````java
void deleteOne(Object filter, SingleResultCallback<DeleteResult> callback);
````

Providing the DeleteResult allows to answer the question "Did I actually delete a file?", which might be useful in some cases. Returning Void could mean to things: 
- one file has been deleted
- no file has been deleted
 
I favour keeping the DeleteResult in this case.

For acknowledged writes we could check if the correct number of chunks has been deleted as well and return an error if not? However if such a operation fails, this usually leaves the database in an inconsistent state. I was wondering if we should provide an option to mark the files document as being deleted before issuing any actual delete operation, delete the chunks first and then finally delete the files document. That at least would make it easy to find any GridFS files where anything went wrong.

driver-async/src/main/com/mongodb/async/client/MongoClientImpl.java
st-h: As of JAVA-1596: Now adding the codec to the files collection/database which is used by the gridfs instance. Reverted any changes to the CodecRegistry

st-h:
Jeff, could you please provide some details about what is going to happen to GridFS support for the async driver?

jyemin:
Apologies for closing this PR on you summarily.  I didn't predict the ramifications of deleting the 3.0.x branch after merging it into master, which is that all the PRs on 3.0.x are automatically closed.  

Can you please re-submit against master?

st-h:
Sure. No problem.
Great that 3.0.x made it into master :)

PR #287, opened 2015-01-07
Updated API listCollectionNames and listIndexes so now match respective ...

rozza:
...commands

JAVA-1589


driver-async/src/main/com/mongodb/async/client/MongoDatabase.java
```diff
@@ -138,9 +136,47 @@
...
+    <C> MongoIterable<C> listCollections(Class<C> clazz);
+
+    /**
+     * Finds the collections in this database.
+     *
+     * @param filter the query filter
+     * @return an iterable containing all the collections in this database
+     * @mongodb.driver.manual reference/command/listCollections listCollections
+     */
+    MongoIterable<Document> listCollections(Object filter);
```
jyemin: Did you consider a ListCollectionsOptions class in case more options are added in the future (like a projection)?

driver-async/src/test/unit/com/mongodb/async/client/MongoDatabaseSpecification.groovy
```diff
@@ -134,7 +134,7 @@ class MongoDatabaseSpecification extends Specification {
         def futureResultCallback = new FutureResultCallback()
 
         when:
-        new MongoDatabaseImpl(name, options, executor).getCollectionNames(futureResultCallback)
+        new MongoDatabaseImpl(name, options, executor).listCollectionNames().into([], futureResultCallback)
```
jyemin: Need tests for the other listCollection* methods

driver-core/src/test/functional/com/mongodb/operation/ListCollectionsOperationSpecification.groovy
```diff
@@ -64,7 +64,7 @@ class ListCollectionsOperationSpecification extends OperationFunctionalSpecifica
 
     def 'should return collection names if a collection exists'() {
         given:
-        def operation = new ListCollectionsOperation(databaseName, new DocumentCodec())
+        def operation = new ListCollectionsOperation(databaseName, new DocumentCodec()).batchSize(10)
```
jyemin: Can we test that batchSize is actually being applied?

driver-core/src/test/functional/com/mongodb/operation/ListIndexesOperationSpecification.groovy
```diff
@@ -90,7 +90,7 @@ class ListIndexesOperationSpecification extends OperationFunctionalSpecification
 
     def 'should return created indexes on Collection'() {
         given:
-        def operation = new ListIndexesOperation(getNamespace(), new DocumentCodec())
+        def operation = new ListIndexesOperation(getNamespace(), new DocumentCodec()).batchSize(10)
```
jyemin: Can we test that batch size is actually being applied?

driver-core/src/main/com/mongodb/operation/ListCollectionsOperation.java
```diff
@@ -186,7 +235,24 @@ private MongoNamespace getNamespace() {
     }
 
     private BsonDocument getCommand() {
-        return new BsonDocument("listCollections", new BsonInt32(1)).append("cursor", new BsonDocument());
+        BsonDocument cursorDocument = batchSize == null ? new BsonDocument() : new BsonDocument("batchSize", new BsonInt32(batchSize));
```
jyemin: Perhaps abstract this line into a static method in CursorHelper?

driver-core/src/main/com/mongodb/operation/ListCollectionsOperation.java
```diff
@@ -186,7 +235,24 @@ private MongoNamespace getNamespace() {
...
+        BsonDocument command = new BsonDocument("listCollections", new BsonInt32(1)).append("cursor", cursorDocument);
+        if (filter != null) {
+            command.append("filter", filter);
+        }
+        return command;
+    }
+
+    private BsonDocument getLegacyFilter() {
```
jyemin: Can you explain what we're trying to accomplish here?  Does the query filter work differently for the command and the legacy query?

jyemin:
Initial review complete.

driver-async/src/main/com/mongodb/async/client/MongoDatabase.java
rozza: Yes but at the moment we only have batchSize as an extra option and it is only applied in 2.8.  I thought this was more akin to the fluentFind which we I could implement for listCollections and listIndexes and then we can expand in the future.

driver-core/src/test/functional/com/mongodb/operation/ListCollectionsOperationSpecification.groovy
rozza: Will add a new test for 2.8 and above to test the batches.

driver-core/src/test/functional/com/mongodb/operation/ListIndexesOperationSpecification.groovy
rozza: As previous

driver-async/src/test/unit/com/mongodb/async/client/MongoDatabaseSpecification.groovy
rozza: Will do although they are implicitly tested here.

driver-core/src/main/com/mongodb/operation/ListCollectionsOperation.java
rozza: Yes, for pre 2.8 we have to ensure that the filter is a string and append the dbname to the string.  This is so we comply with the spec.

rozza:
Todos
-------

  - [ ] Make listCollections fluid - so can set the batchSize from the client api and to help be future compatible.
  - [ ] Make listIndexes fluid - so can set the batchSize from the client api and to help be future compatible.
  - [ ] Test batchSize in listCollections - can we do an initial batchSize (0) and then change it?
  - [ ] Test batchSize in listIndexes - can we do an initial batchSize (0) and then change it?
  - [ ] Add Tests for listCollections* - to ensure its covered in async

rozza:
Have pushed an update - was more in depth than I initially thought but we now support `batchSize` and `maxTimeMS`.

We currently force the listCollection & listIndex methods to read from primary but the spec says:

> Drivers MUST run listCollections on the primary node in "replicaset" mode, unless directly connected to a secondary node in "standalone" mode.

I'm not sure how best to achieve that, the operation however just uses the passed in readPreference - so I'm hoping thats acceptable.

rozza:
Inprogress

rozza:
I did a some what painful rebase onto 3.0.x and actioned all points in aea41f4 

I updated `FindFluentSpecification` as thats where I based the fluent specs from and also added unit tests to check results for the various interfaces to ensure they handle the canned data as expected.  The Operation Specs themselves will handle any logic in `execute` based on server version and the acceptance tests / smoketest also test the functionality.

rozza:
Ready for review again @jyemin :)

jyemin:
lgtm

rozza:
Rebased and merged

PR #291, opened 2015-02-06
[JAVA-1282]  async GridFS support

st-h:
Basic GridFS support, rebased against master.


rozza:
Hi, apologies for not coming back to this ticket sooner. As you may be aware much has changed in the codebase with the new and improved 3.x series. 

@st-h I think your PR is great and would like to thank you for your contribution and the time you spent publishing it.  As a a drivers team we are continually trying to ensure we improve and standardize across all the drivers.  The outcome of which has been a new [GridFS specification](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst) which introduces a new API. The sync driver already has an implementation of this new spec and I'm currently working on the async version ([JAVA-1282](https://jira.mongodb.org/browse/JAVA-1282)).  For that reason I'm having to close this ticket.

st-h:
@rozza thanks for comming back in this. Is there currently any information about the ETA?

rozza:
@st-h its still in progress at the moment but unfortunately its not going to make the 3.2.0 release.  So will be in the following 3.3.0 release.  

I'll update this ticket once its available in a 3.3.0-SNAPSHOT.

PR #294, opened 2015-02-24
Added abstract Filter class

jyemin:
and using it in sync MongoCollection everywhere a filter was previously taken as an object.

Also did a POC of a FilterBuilder which is missing a lot of operators but proves the point that it can be done cleanly. I particularly like that all the methods are static and can therefore be statically imported, so you can make calls like:

    collection.find(and(eq("x", 1), eq("y", 2), gt("z", 3)));

If everyone is happy with this we can proceed to eradicate Object everywhere else as well.

[Review on Reviewable](https://reviewable.io/reviews/mongodb/mongo-java-driver/294)

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/FilterBuilder.java
+++ b/driver-core/src/main/com/mongodb/client/model/FilterBuilder.java
@@ -1,0 +1,297 @@
+package com.mongodb.client.model;
+
+import org.bson.BsonArray;
+import org.bson.BsonBoolean;
+import org.bson.BsonDocument;
+import org.bson.BsonDocumentWriter;
+import org.bson.BsonValue;
+import org.bson.codecs.Encoder;
+import org.bson.codecs.EncoderContext;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import java.util.Map;
+
+import static com.mongodb.assertions.Assertions.notNull;
+import static java.util.Arrays.asList;
+
+/**
+ * A builder for {@link Filter} instances
+ *
+ * @since 3.0
+ */
+public final class FilterBuilder {
+
+    private FilterBuilder() {
+    }
+
+    /**
+     * Creates a filter that ands together the provided list of filters.  Note that this will only generate a "$and" operator if absolutely
+     * necessary, as the query language implicity ands together all the keys.
+     *
+     * @param filters the list of filters to and together
+     * @return the filter
+     * @mongodb.driver.manual reference/operator/query/and $and
+     */
+    public static Filter and(final Iterable<Filter> filters) {
+        return new AndFilter(filters);
+    }
+
+    /**
+     * Creates a filter that ands together the provided list of filters.  Note that this will only generate a "$and" operator if absolutely
+     * necessary, as the query language implicity ands together all the keys.
+     *
+     * @param filters the list of filters to and together
+     * @return the filter
+     * @mongodb.driver.manual reference/operator/query/and $and
+     */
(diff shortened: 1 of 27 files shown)


driver-core/src/main/com/mongodb/client/model/FilterBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+    /**
+     * Creates a filter that ands together the provided list of filters.  Note that this will only generate a "$and" operator if absolutely
+     * necessary, as the query language implicity ands together all the keys.
+     *
+     * @param filters the list of filters to and together
+     * @return the filter
+     * @mongodb.driver.manual reference/operator/query/and $and
+     */
+    public static Filter and(final Filter... filters) {
+        return and(asList(filters));
```
rozza: :+1:

driver/src/main/com/mongodb/MongoCollectionImpl.java
```diff
@@ -254,29 +260,29 @@ public BulkWriteResult bulkWrite(final List<? extends WriteModel<? extends T>> r
         }
 
         return executor.execute(new MixedBulkWriteOperation(namespace, writeRequests, options.isOrdered(),
-                this.writeConcern));
+                                                            this.writeConcern));
```
rozza: Formatting rules have changed again :(

rozza:
I *really* like the builders!

Without implicits this seems verbose but its way better than raw objects!  I think we could improve upon it.  Are we expecting the user to supply a different type to `Sort` or `Projection` than they supply to `asFilter(T document)`?  

I'm not sure we do in the tests, we currently let the user do that, but should we support it? Or do we think it's better to force the user to wrap all their types in `asFilter()` `asSort()`, `asProjection()` because it provides ultimate flexibility, at some cost usability?  In essence `asFilter()` is just a helper just to convert type `T => BsonDocument`.

What if we added a constraint and stopped supporting a mixture of types for all cases? We could simplify all of this by making `MongoCollection` take two types:  `MongoCollection<TCollection, TDocument>`

  *  `TCollection` could be the thing the collection stores `People`, `Workers`, `Documents` etc
  * `TDocument` would be the document like type we use for defining filters, sorts, projections

That way users wouldn't have to wrap everything in `asFilter()`, `asSort()`, `asProjection()` just to create the `BsonDocumentWrappers` we could do it for them.  If they wanted to mix types they could still do this but they would either have to wrap themselves or we we could make some special `TDocument`  type that mirrors what `asFilter()` does has a generic name and can be used everywhere where we use `object` now - basically making creating a `BsonDocumentWrapper` a doddle.

I think we should look to help the common case the most and I think making the user declare a type could be the simplest approach.  Although I'm sure I've missed something..

jyemin:
I think you're on to something.  I added Update last night, and since it required the same render method as Filter, I created a Renderable interface for that method which both Filter and Update implemented.

So what if we just make all the types Renderable instead of Filter, Update, etc, and then make Document, BasicDBObject, BsonDocument implement that type, and have all the builder methods return that type?  That way we don't need any wrapping for the types we support, and third party libraries that want to produce Renderables can either implement the interface directly or provide a wrapper, as is done in the Filter class here.

As is usually the case, the biggest problem is naming.  I was ok with a render method since no one really sees it, but Renderable is just an awful name to sprinkle everywhere.  In an earlier iteration of MongoCollection, we had a ConvertibleToDocument marker interface, which is more descriptive, but ConvertibleToBsonDocument is surely a mouthful.

jyemin:
rebased and merged

PR #295, opened 2015-02-24
Added Renderable interface

jyemin:
OK, this is take two.  In this patch, there's a Renderable interface and most occurrences of Object in MongoCollection and in the model classes have been replaced with Renderable.  In addition, Document and BsonDocument have been retrofitted to implement Renderable, and all the methods on FilterBuilder also return Renderable.

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/Filters.java
+++ b/driver-core/src/main/com/mongodb/client/model/Filters.java
@@ -1,0 +1,300 @@
+package com.mongodb.client.model;
+
+import org.bson.BsonArray;
+import org.bson.BsonBoolean;
+import org.bson.BsonDocument;
+import org.bson.BsonDocumentWriter;
+import org.bson.BsonValue;
+import org.bson.codecs.Encoder;
+import org.bson.codecs.EncoderContext;
+import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.conversions.Bson;
+
+import java.util.Map;
+
+import static com.mongodb.assertions.Assertions.notNull;
+import static java.util.Arrays.asList;
+
+/**
+ * A factory for query filters. A convenient way to use this class is to statically import all of its methods, which allows usage like:
+ * <p><blockquote><pre>
+ *    collection.find(and(eq("x", 1), lt("y", 3)));
+ * </p></pre></blockquote>
+ * @since 3.0
+ */
+public final class Filters {
+
+    private Filters() {
+    }
+
+    /**
+     * Creates a filter that ands together the provided list of filters.  Note that this will only generate a "$and" operator if absolutely
+     * necessary, as the query language implicity ands together all the keys.
+     *
+     * @param filters the list of filters to and together
+     * @return the filter
+     * @mongodb.driver.manual reference/operator/query/and $and
+     */
+    public static Bson and(final Iterable<Bson> filters) {
+        return new AndFilter(filters);
+    }
+
+    /**
+     * Creates a filter that ands together the provided list of filters.  Note that this will only generate a "$and" operator if absolutely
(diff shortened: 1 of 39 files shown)


bson/src/main/org/bson/Renderable.java
```diff
@@ -0,0 +1,20 @@
...
+public interface Renderable {
+    /**
+     * Render the filter into a BsonDocument.
+     *
+     * @param documentClass the document class in scope for the collection
+     * @param codecRegistry the codec registry
+     * @param <TDocument> the type of the document class
+     * @return the filter as a BsonDocument
+     */
+    <TDocument> BsonDocument render(Class<TDocument> documentClass, CodecRegistry codecRegistry);
```
rozza: Is `documentClass` needed? I can't see where its actually used by a `Renderable`.

driver/src/main/com/mongodb/FindIterableImpl.java
```diff
@@ -34,20 +35,23 @@
...
+final class FindIterableImpl<TResult, TDocument> implements FindIterable<TResult> {
     private final MongoNamespace namespace;
-    private final Class<T> clazz;
+    private final Class<TResult> clazz;
     private final ReadPreference readPreference;
     private final CodecRegistry codecRegistry;
     private final OperationExecutor executor;
     private final FindOptions findOptions;
-    private Object filter;
+    private final Class<TDocument> collectionClass;
```
rozza: I'm not sure how `collectionClass` is used which lead me to double check the `Renderable` use of it - is it needed?

rozza:
Couple of questions inline - but overall it looks good.

bson/src/main/org/bson/Renderable.java
jyemin: This may be a problem.  The reason for including it is to allow implementors to re-write filters, projections, sorts, and updates based on knowledge of the class of the document that is in effect for the collection.

Consider a Person class, a MongoCollection<Person>, and a Codec<Person> that transforms the "name" member of Person to an "n" field in the BSON representation.  Then consider a query like this:

     coll.find(eq("name", "Ross"))

when the filter is rendered to a BsonDocument, we'd like to be able to transform the filter to 

     {n : "Ross"}

and in order to do that we need to know about the Person class, so we can look up its Codec in the registry and pull out this information (assuming we have a sub-interface of Codec which provides this information).

The problem is that with a single Renderable interface, we don't know whether we're rendering a filter, a projection, a sort, or an update.

rozza:
Also I saw on #294 your comment about the name - my vote is for: `ConvertibleToDocument` Document to me is strong enough concept that we could just document the interface that we mean `BsonDocument`  either way naming things is hard!

driver-core/src/main/com/mongodb/client/model/Filters.java
```diff
@@ -0,0 +1,300 @@
...
+import java.util.Map;
+
+import static com.mongodb.assertions.Assertions.notNull;
+import static java.util.Arrays.asList;
+
+/**
+ * A factory for query filters. A convenient way to use this class is to statically import all of its methods, which allows usage like:
+ * <p><blockquote><pre>
+ *    collection.find(and(eq("x", 1), lt("y", 3)));
+ * </p></pre></blockquote>
```
rozza: Minor nit - but the html is closed out of order?

rozza:
LGTM

driver-core/src/main/com/mongodb/client/model/Filters.java
jyemin: Done

jyemin:
Rebased and merged, minus the Filters class, which I'll commit separately after it's completed

jyemin:
rebased and merged

PR #296, opened 2015-02-26
Completed work of replacing Object with the Bson type in the high-level API

jyemin:
Looks look I missed a bunch of places in the Iterable implementations where Object has to change to Bson.
Also normalized generic type names and variables names.


jyemin:
@craiggwilson can you review too

rozza:
This the one in codereview?  I reviewed there but also fixed it up in https://github.com/rozza/mongo-java-driver/tree/j1569-2

rozza:
This work has been refactored, rebased and merged into master.

PR #297, opened 2015-03-12
java-1529 JSON parser allows unmatched trailing quote on field names

chithirp:
This change detects unmatched trailing quotes by checking if the current quote char has a corresponding opening quote and throws and exception if not.


rozza:
Hi, apologies for not coming back to this ticket sooner.  This is similar #282 but because 3.x JSON.parse is effectively deprecated, I'm closing this ticket.  Some users may be relying on this behaviour.

The good news is `BasicDBObject.parse`, `Document.parse` and `BsonDocument.parse` will throw an exception with the invalid Json - so those methods should be preferred.

PR #299, opened 2015-03-21
JAVA-1674: Simplified locking in InternalStreamConnection

jyemin:
I took a walk this evening and came up with this design.  The basic insight is that we don't have to hold the reading lock while waiting for IO.  It's enough just to hold it while setting the state.

Ross, maybe this is what you were saying earlier today and I just didn't understand until now.

More comments in line.

The main problem still is that I haven't figured out how to support concurrent async and sync readers together.  With this design, the sync readers are using the reading lock in a way that is not compatible with the async readers.  

One possibility is to not allow sync and async readers to be combined.  In practice, I can't see that they would be.  But not sure how to actually prevent it at compile time without creating a parallel AsyncCluster/AsyncServer/AsyncConnection set of interfaces.

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
+++ b/driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
@@ -275,7 +275,7 @@ class InternalStreamConnection implements InternalConnection {
 
         Response response = null;
         readerLock.lock();
-
+        boolean mustRead = false;
         try {
             response = messages.get(responseTo);
 
@@ -285,8 +285,7 @@ class InternalStreamConnection implements InternalConnection {
 
             if (!readQueue.isEmpty() && !isReading) {
                 isReading = true;
-                fillAndFlipBuffer(REPLY_HEADER_LENGTH,
-                                  errorHandlingCallback(new ResponseHeaderCallback(new ResponseBuffersCallback()), LOGGER));
+                mustRead = true;
             }
         } finally {
             readerLock.unlock();
@@ -294,6 +293,10 @@ class InternalStreamConnection implements InternalConnection {
 
         if (response != null) {
             callback.onResult(response.getResult(), response.getError());
+        }
+
+        if (mustRead) {
+            receiveResponseAsync();
         }
     }
 
@@ -305,8 +308,12 @@ class InternalStreamConnection implements InternalConnection {
         return description.getServerAddress();
     }
 
-    private void fillAndFlipBuffer(final int numBytes, final SingleResultCallback<ByteBuf> callback) {
-        notNull("open", stream, callback);
+    private void receiveResponseAsync() {
+        readAsync(REPLY_HEADER_LENGTH,
+                  errorHandlingCallback(new ResponseHeaderCallback(new ResponseBuffersCallback()), LOGGER));
+    }
+
+    private void readAsync(final int numBytes, final SingleResultCallback<ByteBuf> callback) {
         if (isClosed()) {
             callback.onResult(null, new MongoSocketClosedException("Cannot read from a closed stream", getServerAddress()));
         } else {
@@ -404,8 +411,8 @@ class InternalStreamConnection implements InternalConnection {
                 if (replyHeader.getMessageLength() == REPLY_HEADER_LENGTH) {
                     onSuccess(new ResponseBuffers(replyHeader, null));
                 } else {
-                    fillAndFlipBuffer(replyHeader.getMessageLength() - REPLY_HEADER_LENGTH,
-                                      new ResponseBodyCallback(replyHeader));
+                    readAsync(replyHeader.getMessageLength() - REPLY_HEADER_LENGTH,
+                              new ResponseBodyCallback(replyHeader));
                 }
             }
         }
@@ -509,7 +516,7 @@ class InternalStreamConnection implements InternalConnection {
         }
     }
 
-    private void processFailedRead(final Throwable t) {
+    private void failAllQueuedReads(final Throwable t) {
         close();
         Iterator<Map.Entry<Integer, SingleResultCallback<ResponseBuffers>>> it = readQueue.entrySet().iterator();
         while (it.hasNext()) {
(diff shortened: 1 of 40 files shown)


jyemin:
@craiggwilson fyi

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -67,7 +69,9 @@
     private final AtomicBoolean isClosed = new AtomicBoolean();
     private final AtomicBoolean opened = new AtomicBoolean();
     private final Semaphore writing = new Semaphore(1);
-    private final Semaphore reading = new Semaphore(1);
+    private final Lock readerLock = new ReentrantLock(false);
```
jyemin: We don't need a semaphore here.  It's just a lock. And I had to make it re-entrant because otherwise the tests with mock streams will deadlock.  I'll point out where the deadlock would be below.

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -212,7 +216,7 @@ public ResponseBuffers receiveMessage(final int responseTo) {
         } else {
             while (!messages.containsKey(responseTo)) {
                 try {
-                    reading.acquire();
+                    readerLock.lock();
```
jyemin: The async reader is now using readerLock in a different way than the sync reader, so they won't work together properly.  

If we end up using this design, I'm not sure yet how to make it work with concurrent sync and async readers.

But we don't have any tests of concurrent async and sync readers.

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -259,12 +263,37 @@ public void sendMessageAsync(final List<ByteBuf> byteBuffers, final int lastRequ
...
+            if (!readQueue.isEmpty() && !isReading) {
+                isReading = true;
+                fillAndFlipBuffer(REPLY_HEADER_LENGTH,
+                                  errorHandlingCallback(new ResponseHeaderCallback(new ResponseBuffersCallback()), LOGGER));
+            }
+        } finally {
+            readerLock.unlock();
+        }
+
+        if (response != null) {
```
jyemin: No need to hold the lock just to call the callback

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -259,12 +263,37 @@ public void sendMessageAsync(final List<ByteBuf> byteBuffers, final int lastRequ
...
+        if (isClosed()) {
+            callback.onResult(null, new MongoSocketClosedException("Can not read from a closed socket", getServerAddress()));
+        }
+
+        readerLock.lock();
+
+        Response response = messages.get(responseTo);
+
+        try {
+            if (response == null) {
```
jyemin: There are two possibilities:

# my response has already been read: if so, no need to add myself to the read queue.  Just grab the callback and execute it (after I give up the lock)
# my response has not yet been read: I need to put myself on the read queue.

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -259,12 +263,37 @@ public void sendMessageAsync(final List<ByteBuf> byteBuffers, final int lastRequ
...
+                readQueue.put(responseTo, callback);
+            }
+
+            if (!readQueue.isEmpty() && !isReading) {
```
jyemin: This conditional is the one I'm least sure about.  The second condition, I believe, is necessary to prevent simultaneous readers.    And without the first condition the "InternalStreamConnectionSpecification#should handle out of order messages on the stream asynchronously" test fails on the second async read.

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -259,12 +263,37 @@ public void sendMessageAsync(final List<ByteBuf> byteBuffers, final int lastRequ
...
+                isReading = true;
+                fillAndFlipBuffer(REPLY_HEADER_LENGTH,
+                                  errorHandlingCallback(new ResponseHeaderCallback(new ResponseBuffersCallback()), LOGGER));
+            }
+        } finally {
```
jyemin: I don't need to hold this lock, now that I've updated all my internal state

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -524,36 +508,7 @@ public void failed(final Throwable t) {
...
-                    }
-                }
-                it.remove();
-            }
-        }
-    }
-
-    private void processUnknownFailedRead(final Throwable t) {
-        processPendingResults();
+    private void processFailedRead(final Throwable t) {
```
jyemin: processPendingResults is not necessary because at most one result ever becomes available at a time.

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -624,4 +579,35 @@ public boolean hasError() {
...
+            if (t != null) {
+                processFailedRead(t);
+            } else {
+                SingleResultCallback<ResponseBuffers> callback = null;
+                readerLock.lock();
+                try {
+                    isReading = false;
+                    callback = readQueue.remove(result.getReplyHeader().getResponseTo());
+
+                    if (callback == null) {
```
jyemin: It's possible that the recipient of this message has not put itself on the read queue yet, so we have to store the message for them.

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -624,4 +579,35 @@ public boolean hasError() {
...
+                        messages.put(result.getReplyHeader().getResponseTo(), new Response(result, null));
+                    }
+
+                    if (!readQueue.isEmpty()) {
```
jyemin: If there are more messages to read, read one of them

driver-core/src/main/com/mongodb/connection/InternalStreamConnection.java
```diff
@@ -624,4 +579,35 @@ public boolean hasError() {
...
+                        isReading = true;
+                        fillAndFlipBuffer(REPLY_HEADER_LENGTH, new ResponseHeaderCallback(this));
+                    }
+                } finally {
+                    readerLock.unlock();
+                }
+
+                if (callback != null) {
```
jyemin: Again, no need to hold the lock while executing the callback

jyemin:
Re-opening on a different branch

PR #304, opened 2015-04-08
When checking the presence of slf4j, ClassNotFoundException isn't enough

oallouch:
I'm getting a java.lang.NoClassDefFoundError: org/apache/log4j/Level
because I use org.apache.logging.log4j.Level


oallouch:
Actually, I messed up a bit.
I left some old slf4j in my lib dir, but now I use a new log4j (with the updated packages name).
So the old slf4j is looking for an old log4j, hence the exception.
When I remove the faulty jars, everything works like a charm.
The only thing you could do is add a general catch stating that there is a badly configured slf4j.

Sorry again.

jyemin:
OK, closing this out then.  Thanks.

PR #305, opened 2015-04-10
JAVA-1664 

jyemin:
Added GeoJson object model with enough functionality for use as query filters
Added GeoJson object model codecs (with only encoding support initially)
Added Filters.geoWithin, Filters.geoIntersects, Filters.near, Filters.nearSphere

 JAVA-1664


jyemin:
cc @craiggwilson @rstam

driver-core/src/main/com/mongodb/client/model/geojson/GeometryCollection.java
```diff
@@ -0,0 +1,103 @@
...
+     * Construct an instance with the given list of Geometry objects
+     *
+     * @param coordinateReferenceSystem the coordinate reference system
+     * @param geometries  the list of Geometry objects
+     */
+    public GeometryCollection(final CoordinateReferenceSystem coordinateReferenceSystem,
+                              final List<Geometry> geometries) {
+        super(coordinateReferenceSystem);
+        notNull("geometries", geometries);
+        isTrueArgument("geometrys contain only non-null elements", !geometries.contains(null));
```
rozza: Suggestion: the warning should read "geometrys can only contain non-null elements" or "geometrys cannot contain null elements"

driver-core/src/main/com/mongodb/client/model/geojson/LineString.java
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * @param coordinateReferenceSystem the coordinate reference system
+     * @param coordinates the coordinates
+     */
+    public LineString(final CoordinateReferenceSystem coordinateReferenceSystem,
+                      final List<Position> coordinates) {
+        super(coordinateReferenceSystem);
+        notNull("coordinates", coordinates);
+        isTrueArgument("coordinates must contain at least two positions", coordinates.size() >= 2);
+        isTrueArgument("coordinates contains only non-null positions", !coordinates.contains(null));
```
rozza: Suggestion: "coordinates cannot contain null values"

driver-core/src/main/com/mongodb/client/model/geojson/MultiLineString.java
```diff
@@ -0,0 +1,110 @@
...
+     * @param coordinates the coordinates of each line
+     */
+    public MultiLineString(final CoordinateReferenceSystem coordinateReferenceSystem, final List<List<Position>> coordinates) {
+        super(coordinateReferenceSystem);
+
+        notNull("coordinates", coordinates);
+
+        for (List<Position> line : coordinates) {
+            notNull("line", line);
+            isTrueArgument("line contains only non-null positions", !line.contains(null));
```
rozza: Suggestion "line cannot contain null values"

driver-core/src/main/com/mongodb/client/model/geojson/MultiPoint.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Construct an instance with the given coordinates and coordinate reference system.
+     *
+     * @param coordinateReferenceSystem the coordinate reference system
+     * @param coordinates               the coordinates
+     */
+    public MultiPoint(final CoordinateReferenceSystem coordinateReferenceSystem, final List<Position> coordinates) {
+        super(coordinateReferenceSystem);
+        notNull("coordinates", coordinates);
+        isTrueArgument("coordinates contains only non-null positions", !coordinates.contains(null));
```
rozza: Suggestion: "coordinates cannot contain null values"

driver-core/src/main/com/mongodb/client/model/geojson/MultiPolygon.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Construct an instance.
+     *
+     * @param coordinateReferenceSystem the coordinate reference system
+     * @param coordinates the coordinates
+     */
+    public MultiPolygon(final CoordinateReferenceSystem coordinateReferenceSystem, final List<PolygonCoordinates> coordinates) {
+        super(coordinateReferenceSystem);
+        notNull("coordinates", coordinates);
+        isTrueArgument("coordinates has no null elements", !coordinates.contains(null));
```
rozza: Suggestion: "coordinates cannot contain null values"

driver-core/src/main/com/mongodb/client/model/geojson/PolygonCoordinates.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Construct an instance.
+     *
+     * @param exterior the exterior ring of the polygon
+     * @param holes    optional interior rings of the polygon
+     */
+    public PolygonCoordinates(final List<Position> exterior, final List<Position>... holes) {
+        notNull("exteriorRing", exterior);
+        isTrueArgument("ring contains only non-null positions", !exterior.contains(null));
```
rozza: Suggestion: "ring cannot contain null values"

driver-core/src/main/com/mongodb/client/model/geojson/PolygonCoordinates.java
```diff
@@ -0,0 +1,112 @@
...
+        isTrueArgument("ring must contain at least four positions", exterior.size() >= 4);
+        isTrueArgument("first and last position must be the same", exterior.get(0).equals(exterior.get(exterior.size() - 1)));
+
+        this.exterior = Collections.unmodifiableList(exterior);
+
+        List<List<Position>> holesList = new ArrayList<List<Position>>(holes.length);
+        for (List<Position> hole : holes) {
+            notNull("interiorRing", hole);
+            isTrueArgument("ring contains only non-null positions", !hole.contains(null));
```
rozza: Suggestion: "ring cannot contain null values"

driver-core/src/main/com/mongodb/client/model/geojson/Position.java
```diff
@@ -0,0 +1,103 @@
...
+    private final List<Double> values;
+
+    /**
+     * Construct an instance.
+     *
+     * @param values the non-null values
+     */
+    public Position(final List<Double> values) {
+        notNull("values", values);
+        isTrueArgument("value contains only non-null elements", !values.contains(null));
```
rozza: Suggestion: "values cannot contain null values"

driver-core/src/test/functional/com/mongodb/client/model/GeoFiltersFunctionalSpecification.groovy
```diff
@@ -0,0 +1,77 @@
...
+        expect:
+        find(geoWithinCenter('geo', 2d, 2d, 4d)) == [firstPoint, thirdPoint]
+    }
+
+    @IgnoreIf({ !serverVersionAtLeast([2, 6, 0]) })
+    def '$geoWithin $centerSphere'() {
+        expect:
+        find(geoWithinCenterSphere('geo', 2d, 2d, 4d)) == [firstPoint, secondPoint, thirdPoint]
+    }
+}
```
rozza: nit: no new line

driver-core/src/test/functional/com/mongodb/client/model/GeoJsonFiltersFunctionalSpecification.groovy
```diff
@@ -0,0 +1,82 @@
...
+        find(near('geo', new Point(new Position(1.01d, 1.01d)), 10000d, null)) == [firstPoint]
+    }
+
+    @IgnoreIf({ !serverVersionAtLeast([2, 4, 0]) })
+    def '$nearSphere'() {
+        expect:
+        find(nearSphere('geo', new Point(new Position(1.01d, 1.01d)), 10000d, null)) == [firstPoint]
+    }
+
+}
```
rozza: nit: no new line

driver-core/src/test/unit/com/mongodb/client/model/FiltersSpecification.groovy
```diff
@@ -206,6 +216,275 @@ class FiltersSpecification extends Specification {
...
+                                                                       '[40.0, 18.0]]]}}}}')
+    }
+
+    def 'should render $near'() {
+        given:
+        def point = new Point(new Position(-73.9667, 40.78))
+        def pointDocument = parse(point.toJson())
+
+        expect:
+        toBson(Filters.near('loc', point, 5000d, 1000d)) == parse('{' +
```
rozza: Groovy supports multiline strings which might have been easier to use eg:

```
parse('''{loc : {
     $near: {
        $geometry: {
           type : "Point",
           coordinates : [ -73.9667, 40.78 ]
        },
        $maxDistance: 5000.0,
        $minDistance: 1000.0,
     }
  }
}''')
```

driver-core/src/test/unit/com/mongodb/client/model/geojson/LineStringSpecification.groovy
```diff
@@ -0,0 +1,77 @@
...
+        expect:
+        new LineString(coordinates) == new LineString(coordinates)
+        new LineString(coordinates).hashCode() == new LineString(coordinates).hashCode()
+        new LineString(coordinates).toString() ==
+        'LineString{coordinates=[Position{values=[40.0, 18.0]}, ' +
+        'Position{values=[40.0, 19.0]}, ' +
+        'Position{values=[41.0, 19.0]}, ' +
+        'Position{values=[40.0, 18.0]}]}'
+    }
+}
```
rozza: nit

driver-core/src/test/unit/com/mongodb/client/model/geojson/MultiLineStringSpecification.groovy
```diff
@@ -0,0 +1,72 @@
...
+        expect:
+        new MultiLineString(coordinates) == new MultiLineString(coordinates)
+        new MultiLineString(coordinates).hashCode() == new MultiLineString(coordinates).hashCode()
+        new MultiLineString(coordinates).toString() ==
+        'MultiLineString{coordinates=[' +
+        '[Position{values=[1.0, 1.0]}, Position{values=[2.0, 2.0]}, Position{values=[3.0, 4.0]}], ' +
+        '[Position{values=[2.0, 3.0]}, Position{values=[3.0, 2.0]}, Position{values=[4.0, 4.0]}]]}'
+
+    }
+}
```
rozza: nit

driver-core/src/test/unit/com/mongodb/client/model/geojson/MultiPointSpecification.groovy
```diff
@@ -0,0 +1,71 @@
...
+        expect:
+        new MultiPoint(coordinates) == new MultiPoint(coordinates)
+        new MultiPoint(coordinates).hashCode() == new MultiPoint(coordinates).hashCode()
+        new MultiPoint(coordinates).toString() ==
+        'MultiPoint{coordinates=[Position{values=[40.0, 18.0]}, ' +
+        'Position{values=[40.0, 19.0]}, ' +
+        'Position{values=[41.0, 19.0]}, ' +
+        'Position{values=[40.0, 18.0]}]}'
+    }
+}
```
rozza: nit

rozza:
Looking good.  Comments are:

* I find the CRS parts confusing - why only 3 named systems?  Do we support different CRS systems, Does it impact storage internally and can you use multiple systems and query them?  Very confusing.

* I found some error message to hard to decipher so suggested alternatives.
* Groovy files seem to be missing newlines - you might want to tweak intellij to auto add on save.
* Also, its much easier to use multiline strings rather than concatenate them in the tests ;)  

* Missing documentation!

jyemin:
* I find the CRS parts confusing - why only 3 named systems?  Do we support different CRS systems, Does it impact storage internally and can you use multiple systems and query them?  Very confusing.
** Those are the three named systems that I found supported by the server.  The CRS names are stored by the server, but the only one that has any different effect from the default is STRICT_WINDING, and that one is only allowed on a query filter, not for stored documents.  And it does change the semantics of the query, according to the documentation.  
* I found some error message to hard to decipher so suggested alternatives.
** Will take a look
* Groovy files seem to be missing newlines - you might want to tweak intellij to auto add on save.
** Found the setting and changed it.  Will fix.
* Also, its much easier to use multiline strings rather than concatenate them in the tests ;)  
** I did it this way because those strings were copied from our documentation, and IntelliJ added the newlines automatically
* Missing documentation!
** Aware :)

rozza:
:+1:

driver-core/src/main/com/mongodb/client/model/geojson/GeometryCollection.java
jyemin: The problem with this wording is that, if the condition is false, the IllegalArgumentException's message will be: "state should be: geometrys cannot contain null values".  The way it is now, it will be: "state should be: geometrys contain only non-null elements"

jyemin:
Addressed all comments except for reference documentation.

driver-core/src/main/com/mongodb/client/model/geojson/Geometry.java
```diff
@@ -0,0 +1,114 @@
...
+import org.bson.json.JsonWriterSettings;
+
+import java.io.StringWriter;
+
+/**
+ * An abstract class for representations of GeoJSON geometry objects.
+ *
+ * @since 3.1
+ */
+public abstract class Geometry {
```
rstam: Two comments:

1. Do you not want to prefix your GeoJson classes with GeoJson to avoid ambiguity or name collissions?
2. Why does this not have GeoJsonObject as a base class? I realize that since you chose not to implement GeoJsonFeature (yet) that GeoJsonGeometry would be the only GeoJsonObject subclass so far, but it seems like the right thing to do is to have a GeoJsonObject class.

driver-core/src/main/com/mongodb/client/model/geojson/Geometry.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private static final CodecRegistry REGISTRY = CodecRegistries.fromProviders(new GeoJsonCodecProvider());
+
+    private final CoordinateReferenceSystem coordinateReferenceSystem;
```
rstam: Where's the bounding box?

What about extra members?

driver-core/src/main/com/mongodb/client/model/geojson/Geometry.java
jyemin: 1. I considered it, but decided against it, as there are no other classes named Point, Position, etc that are likely to be used in a server application.
2. Just because it can be added if and when we add features support, without breaking binary compatibility.

driver-core/src/main/com/mongodb/client/model/geojson/MultiPolygon.java
```diff
@@ -0,0 +1,103 @@
...
+import static com.mongodb.assertions.Assertions.notNull;
+
+/**
+ * A representation of a GeoJSON MultiPolygon.
+ *
+ * @since 3.1
+ */
+public final class MultiPolygon extends Geometry {
+
+    private final List<PolygonCoordinates> coordinates;
```
rstam: Why do some coordinates get special classes (e.g. PolygonCoordinates) and others are just something like List<List<Position>>?

driver-core/src/main/com/mongodb/client/model/geojson/Point.java
```diff
@@ -0,0 +1,107 @@
...
+package com.mongodb.client.model.geojson;
+
+import static com.mongodb.assertions.Assertions.notNull;
+
+/**
+ * A representation of a GeoJSON Point.
+ *
+ * @since 3.1
+ */
+public final class Point extends Geometry {
```
rstam: Seems like Point is a class name destined for lots of name collisions...

driver-core/src/main/com/mongodb/client/model/geojson/Position.java
```diff
@@ -0,0 +1,103 @@
...
+import static com.mongodb.assertions.Assertions.isTrueArgument;
+import static com.mongodb.assertions.Assertions.notNull;
+
+/**
+ * A representation of a GeoJSON Position.
+ *
+ * @since 3.1
+ */
+@Immutable
+public final class Position {
```
rstam: Another class name that seems like it's destined for name collisions...

driver-core/src/main/com/mongodb/client/model/geojson/MultiPoint.java
```diff
@@ -0,0 +1,106 @@
...
+import static com.mongodb.assertions.Assertions.notNull;
+
+/**
+ * A representation of a GeoJSON MultiPoint.
+ *
+ * @since 3.1
+ */
+public final class MultiPoint extends Geometry {
+
+    private final List<Position> coordinates;
```
rstam: I understand the simplification you are making here... but this means that every position could have a different number of dimensions, which would be invalid.

Are there any run-time checks to make sure the dimensionality of all the coordinates are in agreement?

driver-core/src/main/com/mongodb/client/model/geojson/codecs/LineStringCodec.java
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param registry the registry
+     */
+    public LineStringCodec(final CodecRegistry registry) {
+        this.registry = notNull("registry", registry);
+    }
+
+    @Override
+    public void encode(final BsonWriter writer, final LineString value, final EncoderContext encoderContext) {
+        encodeGeometry(writer, value, encoderContext, registry, new Runnable() {
```
rstam: Is Runnable because you don't have lambdas in older versions of Java?

driver-core/src/main/com/mongodb/client/model/Filters.java
```diff
@@ -485,10 +490,266 @@ public static Bson size(final String fieldName, final int size) {
...
+     *
+     * @param fieldName the field name
+     * @param points    a list of pairs of x, y coordinates.  Any extra dimensions are ignored
+     * @return the filter
+     * @mongodb.driver.manual reference/operator/query/geoWithin/ $geoWithin
+     * @mongodb.driver.manual reference/operator/query/polygon/#op._S_polygon $polygon
+     * @mongodb.server.release 2.4
+     * @since 3.1
+     */
+    public static Bson geoWithinPolygon(final String fieldName, final List<List<Double>> points) {
```
rstam: Shouldn't the parameter be of type GeoJson Polygon?

driver-core/src/main/com/mongodb/client/model/geojson/Position.java
jyemin: That's what import is for, to disambiguate.  I just don't see people using a different Point or Position class in the same .java file

driver-core/src/main/com/mongodb/client/model/geojson/codecs/LineStringCodec.java
jyemin: Yes, exactly.

driver-core/src/main/com/mongodb/client/model/Filters.java
jyemin: No, this one if for the legacy $polygon: http://docs.mongodb.org/manual/reference/operator/query/polygon/#op._S_polygon.

It's the same in .NET's FilterDefinition, I believe.

driver-core/src/main/com/mongodb/client/model/geojson/MultiPoint.java
jyemin: No, I'm relying on the server to throw in this case.  I'll look into adding run time checks.

driver-core/src/main/com/mongodb/client/model/geojson/MultiPolygon.java
jyemin: This is the only one, I believe.  And it's just because polygon coordinates have two parts: exterior and holes.  So this was the only way to pass a list of them.  Note that the Polygon class has a constructor that takes one of these, and another one that takes the two pieces separately.

driver-core/src/main/com/mongodb/client/model/geojson/Point.java
jyemin: There are two Point classes in the JDK: java.awt.Point and javafx.scene.effect.Point.  Neither are likely to be used in the same Java file along with the Java driver.

jyemin:
Added reference documentation of the Filters builder methods, include geospatial methods.

rozza:
Currently failing checkstyle.

rozza:
Looks great, I like the documentation.  Is it worth mentioning that the helpers don't validate the document that is built but rather build document as requested.

jyemin:
Do you mean that you can do:

   and(new Document("field", new Document("$crazyOperator", 1)...)

and it will just send it to the server as is?  This will only happen if you leave the Filters world, and I don't show any examples of that yet.

rozza:
Yes or an unsupported $not query etc..

jyemin:
In that case why not leave it as is?  Either way, the user will get an error reported, just by the server instead of the driver.

rozza:
Thats fine, but a note in the documentation as well would hurt either.

jyemin:
Rebased and merged.

PR #307, opened 2015-04-17
parse datestrings under $date in JsonReader

fbuecklers:
implement "$date": "2015-04-14T14:55:57.626Z" and "$date": "2015-04-14T14:55:57+02:00" extended json parsing in JsonReader

See Jira ticket https://jira.mongodb.org/browse/JAVA-1768

### DIFF
--- a/bson/src/main/org/bson/json/JsonReader.java
+++ b/bson/src/main/org/bson/json/JsonReader.java
@@ -784,7 +784,7 @@ public class JsonReader extends AbstractBsonReader {
     }
 
     private long visitDateTimeConstructor() {
-        DateFormat df = new SimpleDateFormat("EEE MMM dd yyyy HH:mm:ss z");
+        DateFormat format = new SimpleDateFormat("EEE MMM dd yyyy HH:mm:ss z", Locale.ENGLISH);
 
         verifyToken("(");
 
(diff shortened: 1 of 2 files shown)


jyemin:
Florian,

I posted a comment on the ticket about JsonWriter, as well as $numberLong.  Feel free to tackle those as well.  Generally, JsonReader and JsonWriter should comply with the documentation here; http://docs.mongodb.org/manual/reference/mongodb-extended-json/

Thanks for the report and the pull request.

jyemin:
I found a few more things when I brought it into my editor:

* If you run ./gradlew check, you'll see that it fails due to the *-style imports, which we disallow.
* Please change the exception message for the fist JsonParseException to say something like: "JSON reader expected an ISO-8601 date string but found '%s'."
* Please add a test for the exceptional condition where the date is in an invalid format
* Please change the exception message for the second JsonParseException to say something like: "JSON reader expected an integer or string but found '%s'.".  Extra credit if you add a test for that branch too, which I noticed we missed.

Please also squash the commits at the end, as that's our practice.

Thanks,
Jeff

fbuecklers:
Hopes that it looks good for you now ;-)

jyemin:
It looks great.  Merged to 3.0.x and master.

Thanks!

PR #308, opened 2015-04-29
faster ObjectId.toHexString method

stevenbenjamin:
I found that the ObjectId.toHexString method was taking up a significant amount of time in profiling. This version seems to be about 30x faster.

### DIFF
--- a/docs/landing/data/releases.toml
+++ b/docs/landing/data/releases.toml
@@ -2,7 +2,6 @@ current = "3.0.0"
 
 [[versions]]
   version = "3.1.0-SNAPSHOT"
-  status = "snapshot"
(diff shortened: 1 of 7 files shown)


jyemin:
Hi Steven,

org.bson.types.ObjectIdTest#testHexStringConstructor is failing with this patch

stevenbenjamin:
Sorry about that - This updated version is 4x faster than the previous and passes that test.

rozza:
Hi @stevenbenjamin looks like its still failing on travis, this time it looks like its the check style rules.

Or you ok to fix? It'll probably be worth squashing and rebasing into a single commit.

Thanks Ross.

PR #311, opened 2015-05-20
try.mongodb.org now redirects to university.mongodb.com

gianpaj:


docs/landing/layouts/partials/header/topRight.html
```diff
@@ -1,6 +1,6 @@
 <div>
 <div class="nav-items pull-right">
-  <a href="http://try.mongodb.org/" data-toggle="tooltip" data-placement="bottom" title="Try MongoDB in the browser">Try it Out</a>
+  <a href="https://university.mongodb.com" data-toggle="tooltip" data-placement="bottom" title="Free Online Classes">MongoDB University</a>
```
evanchooly: probably also needs to be done in reference/

gianpaj:
thanks. my bad

evanchooly:
I meant you'll need to do it in both places since we define that header twice.

gianpaj:
Oh sorry (again). didn't know there were two places. I've made a second commit on #312 - hope that's ok. happy to do one commit instead

PR #314, opened 2015-05-24
Update QueryBuilder.java

MorrisLaw:
Renamed the _query and _currentKey variable so that it did not have a symbol in the variable name; makes for easier reading of the code.


rozza:
Thanks for the PR @MorrisLaw - the `_varName` is an existing convention for private variables in some of the legacy parts of the codebase and as its used elsewhere.

Although we may no longer follow this convention in the new parts of the codebase, for legacy parts we are happy to maintain the existing convention.

PR #315, opened 2015-05-24
Update AcknowledgeBulkWriteResult.java

MorrisLaw:
Fixed the exception error in getModifiedCount and removed the word "this" from "the server this must".... also added version 2.6 "or higher" to the end of the statement. Also added spacing to the toString return values.

### DIFF
--- a/driver/src/main/com/mongodb/AcknowledgedBulkWriteResult.java
+++ b/driver/src/main/com/mongodb/AcknowledgedBulkWriteResult.java
@@ -66,8 +66,8 @@ class AcknowledgedBulkWriteResult extends BulkWriteResult {
     public int getModifiedCount() {
         if (modifiedCount == null) {
             throw new UnsupportedOperationException("The modifiedCount is not available because at least one of the servers that was "
-                                                    + "updated was not able to provide this information (the server is must be at least "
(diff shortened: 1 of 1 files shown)


Jeremy1229:
"is" not "this" was the grammatical error I found in the exception error under getModifiedCount.

driver/src/main/com/mongodb/AcknowledgedBulkWriteResult.java
```diff
@@ -120,11 +120,11 @@ public int hashCode() {
     @Override
     public String toString() {
         return "AcknowledgedBulkWriteResult{"
-               + "insertedCount=" + insertedCount
-               + ", matchedCount=" + matchedCount
-               + ", removedCount=" + removedCount
-               + ", modifiedCount=" + modifiedCount
-               + ", upserts=" + upserts
+               + "insertedCount = " + insertedCount
```
jyemin: Please revert this change.  We've been using IntelliJ's default toString generator and this will introduce an inconsistency with all the other generated methods.

rozza:
Thanks @MorrisLaw  - I updated the exception based on your PR, but left the `toString()` method as is.

PR #317, opened 2015-06-09
Argument for validation method changed according to the conventions

OzMolaim:
A minor thing I noticed during reading the code.
All parameters are passed to the validation accompanied with their actual name as a string.


rozza:
Thanks @OzMolaim!

PR #318, opened 2015-06-11
Put a note on DB that it is deprecated

benmccann:
We don't want to use the deprecation tag so as not to cause spurious warnings in our users' codebases, but should clarify that the class does indeed have a successor

### DIFF
--- a/driver/src/main/com/mongodb/DB.java
+++ b/driver/src/main/com/mongodb/DB.java
@@ -46,6 +46,13 @@ import static com.mongodb.ReadPreference.primary;
 import static java.util.Arrays.asList;
 
 /**
+ * This class has been superseded by {@link com.mongodb.client.MongoDatabase}, instances of which
+ * can be created via {@link MongoClient#getDatabase(String)}. While instances of this class can
+ * still be created via {@link MongoClient#getDB(String)}, this method has been deprecated in order
+ * to signal that {@link com.mongodb.client.MongoDatabase} is now preferred. This class, however,
+ * has not been deprecated yet, in order to give users time to migrate their code without
+ * experiencing a huge number of deprecation warnings.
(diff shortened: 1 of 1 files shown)


benmccann:
It was massively confusing to me whether DB was deprecated or just Mongo.getDB. I think this makes it clearer to users what the intended path forward is.

rozza:
@kgignatyev the class has effectively deprecated, in 3.0 `Mongo.getDB` was annotated as `@Deprecated` so all code under is effectively deprecated.  Theres no need to annotate every class as that would be painful for legacy users of the API.

Thinking about it I would prefer the `Mongo.getDB` documentation be clarified rather than `DB` as that seems to have been the source of the confusion.  Anyone using `getDB` will have a deprecation warning and that tells users to use `getDatabase`.  I'm not sure we need the note in the `DB` class because if we need it there then shouldn't we have it also on `DBCollection` and `DBCursor`?   

@benmccann what do you think?

benmccann:
Adding the note to DBCollection and DBCursor wouldn't be a bad idea. Do you mind if we go that way instead? The thing that really threw me was that there's nothing on these classes at all to indicate that they're deprecated.

rozza:
Having discussed this further with the team, we don't have current plans to mark `DB`, `DBCollection`, `DBCursor` and others under that code path as deprecated.  We will continue to fully support those classes during the 3.x series of the driver.  Once a final timeline has decided for 4.0 we will then look to annotating and deprecating classes under `DB`.   

At this present time, we will continue to advise users and 3rd party libraries to continue adopting the new API.  I've reopened JAVA-1858 to help improve both the documentation and API docs for the forthcoming 3.1 release.

Many thanks @benmccann for being patient and persevering with this.

PR #319, opened 2015-06-17
Conventions

evanchooly:
@jyemin @craiggwilson @rozza

Here's the first pass at an implementation.  I wouldn't consider this "done" by any means but it's on its way.  This uses the phases and weights approach we discussed and builds out the basic infrastructure.  There are several more tests to write as well as some validation prototyping (which should essentially be Conventions that throw ClassMappingExceptions).

At any rate, I'm reasonably pleased with the progress and direction.  I've managed to hide away all the classmate references so, with luck, and Convention authors will only ever see ClassModel and FieldModel references with nary a clue there's anything else at play.

### DIFF
--- a/bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
+++ b/bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
@@ -33,7 +33,7 @@ import java.util.TreeMap;
 /**
  * This class represents the various generics and field metadata of a class for use in mapping data to and from the database.
  */
-public class ClassModel extends MappedType {
+public final class ClassModel extends MappedType {
     private final Map<String, FieldModel> fields = new TreeMap<String, FieldModel>();
     private final CodecRegistry registry;
     private final TypeResolver resolver;
--- a/bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
+++ b/bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
@@ -20,6 +20,7 @@ import com.fasterxml.classmate.ResolvedType;
 import com.fasterxml.classmate.ResolvedTypeWithMembers;
 import com.fasterxml.classmate.TypeResolver;
 import com.fasterxml.classmate.members.ResolvedField;
+import com.fasterxml.classmate.members.ResolvedMethod;
 import org.bson.codecs.configuration.CodecRegistry;
 
 import java.util.ArrayList;
@@ -34,6 +35,10 @@ public class ClassModel extends MappedType {
     private final Map<String, FieldModel> fields = new TreeMap<String, FieldModel>();
     private final CodecRegistry registry;
     private final TypeResolver resolver;
+    private final MemberResolver memberResolver;
+    private final WeightedValue<String> collectionName = new WeightedValue<String>();
+    private final Map<String, List<MethodModel>> methods = new TreeMap<String, List<MethodModel>>();
+    private boolean mapped;
 
     /**
      * Construct a ClassModel for the given Classs.
@@ -42,35 +47,20 @@ public class ClassModel extends MappedType {
      * @param resolver the TypeResolver used in discovery of Class metatadata
      * @param aClass   the Class to model
      */
-    public ClassModel(final CodecRegistry registry, final TypeResolver resolver, final Class<?> aClass) {
+    ClassModel(final CodecRegistry registry, final TypeResolver resolver, final Class<?> aClass) {
         super(aClass);
         this.registry = registry;
         this.resolver = resolver;
-        map();
-    }
-
-    protected void map() {
-        final ResolvedType type = resolver.resolve(getType());
-        final List<ResolvedType> resolvedTypes = type.getTypeParameters();
-        for (final ResolvedType resolvedType : resolvedTypes) {
-            addParameter(resolvedType.getErasedType());
-        }
-
-        final ResolvedTypeWithMembers bean = new MemberResolver(resolver)
-                                                 .resolve(type, null, null);
-        final ResolvedField[] fields = bean.getMemberFields();
-        for (final ResolvedField field : fields) {
-            addField(new FieldModel(this, registry, field));
-        }
+        memberResolver = new MemberResolver(resolver);
     }
 
     /**
-     * Adds a field to the model
+     * Returns the collection name to use when de/encoding BSON documents.
      *
-     * @param field the field to add
+     * @return the collection name
      */
-    public void addField(final FieldModel field) {
-        fields.put(field.getName(), field);
+    public String getCollectionName() {
+        return collectionName.get();
     }
 
     /**
@@ -90,5 +80,73 @@ public class ClassModel extends MappedType {
      */
     public List<FieldModel> getFields() {
         return new ArrayList<FieldModel>(fields.values());
+    }
+
+    /**
+     * Returns the name of the class represented by this ClassModel
+     *
+     * @return the name
+     */
+    public String getName() {
+        return getType().getName();
+    }
+
+    /**
+     * Suggests a new value for the collection name.
+     *
+     * @param weight         The weight to give this value relative to other values set
+     * @param collectionName the new collection name to suggest
+     * @see WeightedValue
+     */
+    public void setCollectionName(final Integer weight, final String collectionName) {
+        this.collectionName.set(weight, collectionName);
+    }
+
+    /**
+     * Executes the actual mapping of the class.
+     */
+    public void map() {
+        if (!mapped) {
+            final ResolvedType resolved = resolver.resolve(getType());
+            final ResolvedTypeWithMembers type = memberResolver.resolve(resolved, null, null);
+
+            for (final ResolvedType resolvedType : resolved.getTypeParameters()) {
+                addParameter(resolvedType.getErasedType());
+            }
+
+            for (final ResolvedField field : type.getMemberFields()) {
+                addField(field);
+            }
+
+            for (final ResolvedMethod memberMethod : type.getMemberMethods()) {
+                addMethod(memberMethod);
+            }
+            mapped = true;
+        }
+    }
+
+    private void addField(final ResolvedField field) {
+        final FieldModel model = new FieldModel(this, registry, field);
+        fields.put(model.getName(), model);
+    }
+
+    private void addMethod(final ResolvedMethod method) {
+        final MethodModel model = new MethodModel(this, registry, method);
+        final List<MethodModel> list = getMethods(model.getName());
+        if (list.isEmpty()) {
+            methods.put(model.getName(), list);
+        }
+        list.add(model);
+    }
+
+    /**
+     * Gets the named method if it exists
+     *
+     * @param name the name of the method to fetch
+     * @return the MethodModel named
+     */
+    public List<MethodModel> getMethods(final String name) {
+        final List<MethodModel> list = methods.get(name);
+        return list == null ? new ArrayList<MethodModel>() : list;
     }
 }
--- a/bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
+++ b/bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
@@ -15,6 +15,8 @@
  */
 package org.bson.codecs.configuration.mapper;
 
+import com.fasterxml.classmate.AnnotationConfiguration.StdConfiguration;
+import com.fasterxml.classmate.AnnotationInclusion;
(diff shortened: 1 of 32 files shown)


bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
```diff
@@ -34,6 +35,10 @@
     private final Map<String, FieldModel> fields = new TreeMap<String, FieldModel>();
     private final CodecRegistry registry;
     private final TypeResolver resolver;
+    private final MemberResolver memberResolver;
+    private final WeightedValue<String> collectionName = new WeightedValue<String>();
```
craiggwilson: I'm of two minds about having this in here. Strictly speaking, this is just codec and should know anything about a collection. For instance, what if this class is intended for use in more than one collection?

bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
```diff
@@ -91,4 +81,72 @@ public FieldModel getField(final String name) {
...
+    public void map() {
+        if (!mapped) {
+            final ResolvedType resolved = resolver.resolve(getType());
+            final ResolvedTypeWithMembers type = memberResolver.resolve(resolved, null, null);
+
+            for (final ResolvedType resolvedType : resolved.getTypeParameters()) {
+                addParameter(resolvedType.getErasedType());
+            }
+
+            for (final ResolvedField field : type.getMemberFields()) {
```
craiggwilson: I think adding all the fields and methods is an elegant way to handle the added/removed problem. +1.

bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodecProvider.java
```diff
@@ -52,12 +60,17 @@ public static ProviderBuilder builder() {
     @Override
     @SuppressWarnings({"unchecked", "rawtypes"})
     public <T> Codec<T> get(final Class<T> clazz, final CodecRegistry registry) {
-        Codec<T> codec = null;
+        ClassModelCodec<?> codec = null;
         if (registered.contains(clazz)) {
-            final ClassModel model = new ClassModel(registry, resolver, (Class<Object>) clazz);
-            codec = (Codec<T>) new ClassModelCodec(model);
+            codec = mapped.get(clazz);
+            if (codec == null && conventionPack.isMappable(clazz)) {
```
craiggwilson: Isn't, by definition, the fact that the class was registered with the codec provider enough to justify that it should be mapped?

bson/src/main/org/bson/codecs/configuration/mapper/MethodModel.java
```diff
@@ -0,0 +1,70 @@
...
+
+package org.bson.codecs.configuration.mapper;
+
+import com.fasterxml.classmate.members.ResolvedMethod;
+import org.bson.codecs.configuration.CodecRegistry;
+
+/**
+ * Represents the metamodel of a method on a class mapped by a ClassModel
+ */
+public class MethodModel {
```
craiggwilson: This doesn't seem complete... I assumed it would inherit from MappedType just like FieldModel.

bson/src/main/org/bson/codecs/configuration/mapper/Weights.java
```diff
@@ -0,0 +1,37 @@
...
+ * Represents some standard weight values to help break up the spectrum of possible weight values.
+ */
+public final class Weights {
+    public static final int DEFAULT = 0;
+
+    public static final int BUILT_IN_CONVENTION = 100;
+
+    public static final int USER_CONVENTION = 200;
+
+    public static final int BUILT_IN_ATTRIBUTE = 300;
```
craiggwilson: Probably should change these to annotation.

bson/src/main/org/bson/codecs/configuration/mapper/conventions/Convention.java
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @return the phase
+     */
+    String getPhase();
+
+    /**
+     * @return the weight to be applied to any suggested values coming from this Convention
+     * @see org.bson.codecs.configuration.mapper.WeightedValue
+     */
+    Integer getWeight();
```
craiggwilson: I hadn't thought of having this out there. Is there any reason for this other than debugging purposes?

bson/src/main/org/bson/codecs/configuration/mapper/conventions/BeanPropertiesConvention.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * This convention ensures only fields that are private, nonstatic, nontransient, and nonfinal that have a proper getter and setter as per
+ * the Beans Specification are included in BSON processing.
+ *
+ * @see <a target="_blank" href="http://www.oracle.com/technetwork/articles/javaee/spec-136004.html">JavaBeans Spec</a>
+ */
+public class BeanPropertiesConvention implements Convention {
+    @Override
+    public Integer getWeight() {
+        return 100;
```
craiggwilson: Might as well use your provided constants, Weights.BUILT_IN_CONVENTION.

bson/src/main/org/bson/codecs/configuration/mapper/conventions/BeanPropertiesConvention.java
```diff
@@ -0,0 +1,73 @@
...
+        return ConventionPack.FIELD_MAPPING;
+    }
+
+    @Override
+    public void apply(final ClassModel model) {
+        final List<FieldModel> resolvedFields = model.getFields();
+        for (final FieldModel field : resolvedFields) {
+            if (field.isPrivate() && !field.isTransient() && !field.isFinal()) {
+                final String name = field.getName();
+                final String propertyName = name.substring(0, 1).toUpperCase() + name.substring(1);
```
craiggwilson: Does your upper casing here suffer from any cultural problems?

bson/src/main/org/bson/codecs/configuration/mapper/conventions/LowerCaseClassNamingConvention.java
```diff
@@ -0,0 +1,40 @@
...
+
+package org.bson.codecs.configuration.mapper.conventions;
+
+import org.bson.codecs.configuration.mapper.ClassModel;
+import org.bson.codecs.configuration.mapper.Weights;
+
+/**
+ * Attempts to configure the collection name as the lower case form of the class name.
+ */
+public class LowerCaseClassNamingConvention implements Convention {
```
craiggwilson: If this is the default, I wouldn't have a convention for it and simply set the default when you construct the ClassModel.

bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
evanchooly: I see your point but this isn't not that different than putting a field name on the FieldModel:  it's just telling the codec how/where to map this to.

As for multiple destinations, morphia, for example, allows only a single collection mapping but you can override it when calling the Datastore methods by passing the name of the collection to use.

bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
evanchooly: I realized that by doing this, i can effectively hide all the reflection/introspection away.

bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodecProvider.java
evanchooly: Not necessarily.  Morphia's implementation will allow for mapping whole packages as well.  Probably the driver version will allow that as well.  Can't see why it wouldn't.

bson/src/main/org/bson/codecs/configuration/mapper/MethodModel.java
evanchooly: It's a bit of a smell that i haven't quite mitigated.  It initially did but MappedType has a getType() and the closest a method has to a type is its return type.  I think I changed it when I hit an NPE trying to extract the class of the return type on a void method.  I can *probably* fix it but I just haven't come back around to this one yet.  Ideally, yes, it'd be MappedType as well.

bson/src/main/org/bson/codecs/configuration/mapper/Weights.java
evanchooly: come again?

bson/src/main/org/bson/codecs/configuration/mapper/conventions/Convention.java
evanchooly: Hadn't thought about what now?

bson/src/main/org/bson/codecs/configuration/mapper/conventions/BeanPropertiesConvention.java
evanchooly: *facepalm*  yeah...

bson/src/main/org/bson/codecs/configuration/mapper/conventions/BeanPropertiesConvention.java
evanchooly: Potentially some i18n issues here but since this is all coming from source, we're basically guaranteed English names.  The Beans Spec all but requires it really with its use of "get" and "set" prefixes.  I don't know that there's much wiggle room in this case.

bson/src/main/org/bson/codecs/configuration/mapper/conventions/LowerCaseClassNamingConvention.java
evanchooly: That *could* be said about almost everything that ends up in the DefaultConventionPack, though...  Morphia defaults to just the classname, though.  Before enshrining too many defaults hard coded in to the model classes, I'd want to have some conversations about those default behaviors should be.

bson/src/main/org/bson/codecs/configuration/mapper/ClassModel.java
jyemin: We have to consider how this will be used in the context of the Java driver.  Currently, you have to be explicit about the collection name:

     database.getCollection("people", Person.class);

I suppose the collectionName could be ignored when used this way, and could be taken into account with an alternate API like:

    database.getCollection(Person.class)

For this usage, the Codec interface would have to be optionally extendable to provide the collection name.

evanchooly: That's a great point actually.  With the driver, we'd have an explicit MongoCollection in hand.  This name is really for ODMs (read:  Morphia) to know how to get the appropriate MongoCollection to use.  hrm...

craiggwilson: So, perhaps the solution here isn't to add something explicit into the code, but rather to add a way for ODMs to annotate things at runtime with desirable information. For instance, have a Map<string, string> or something.

bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodecProvider.java
craiggwilson: It just seems completely redundant to me. Even in the MorphiaCodecProvider, presumably it wouldn't recognize things that morphia couldn't map at the codec provider level... hence, the isMappable check is still unnecessary.

bson/src/main/org/bson/codecs/configuration/mapper/Weights.java
craiggwilson: Do you call them attributes in Java? Don't you call them annotations?

bson/src/main/org/bson/codecs/configuration/mapper/conventions/Convention.java
craiggwilson: Why is getWeight() on the interface? It doesn't seem like any consumers would ever use this. ConventionPack, for instance, doesn't use it. And, in practice, it is quite restrictive as to what weights could be applied inside a convention. For instance, a hypothetical CompositeConvention that has a couple of conventions inside it with different weights, which one gets returned out of this method...  So, I think that getWeight() shouldn't be on the interface.

evanchooly: getWeight() is used internally when setting values.  It could probably be removed but I do like that it forces convention writers to explicitly deal with that concept.  I do see your point, though...  I'm on the fence.

PR #320, opened 2015-06-25
Notify listener about wait queue enter before throwing an exception if the queue is full (JAVA-1873)

fzhinkin:
Without this fix DefaultConnectionPool::get throws exception about a full queue first and then notify listener about wait queue exit in finally clause.
As a result, wait queue size reported by the ConnectionPoolStatistics may become negative.

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/connection/DefaultConnectionPool.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2014 MongoDB, Inc.
+ * Copyright (c) 2008-2015 MongoDB, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -83,10 +83,10 @@ class DefaultConnectionPool implements ConnectionPool {
     @Override
     public InternalConnection get(final long timeout, final TimeUnit timeUnit) {
         try {
+            connectionPoolListener.waitQueueEntered(new ConnectionPoolWaitQueueEvent(serverId, currentThread().getId()));
             if (waitQueueSize.incrementAndGet() > settings.getMaxWaitQueueSize()) {
                 throw createWaitQueueFullException();
             }
-            connectionPoolListener.waitQueueEntered(new ConnectionPoolWaitQueueEvent(serverId, currentThread().getId()));
             PooledConnection pooledConnection = getPooledConnection(timeout, timeUnit);
             if (!pooledConnection.opened()) {
                 try {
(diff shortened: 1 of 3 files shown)


fzhinkin:
Hi Jeff,

seems like OOME that caused org.bson.GenericBsonTest::shouldPassAllOutcome failure does not relate to my change, is it?

Thanks,
Filipp.

jyemin:
Yes, unrelated.

driver-core/src/main/com/mongodb/connection/DefaultConnectionPool.java
```diff
@@ -83,10 +83,10 @@ public InternalConnection get() {
     @Override
     public InternalConnection get(final long timeout, final TimeUnit timeUnit) {
         try {
+            connectionPoolListener.waitQueueEntered(new ConnectionPoolWaitQueueEvent(serverId, currentThread().getId()));
```
jyemin: I think this is incorrect.  The thread can't enter the wait queue if the wait queue is full, so the event should not be generated.

How about instead we pull the throw of the wait queue full exception up out of the try block? That way the finally block will not be executed when the exception is thrown.

driver-core/src/test/functional/com/mongodb/connection/QueueEventsConnectionPoolListener.java
```diff
@@ -0,0 +1,40 @@
...
+import com.mongodb.event.ConnectionPoolListenerAdapter;
+import com.mongodb.event.ConnectionPoolWaitQueueEvent;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+class QueueEventsConnectionPoolListener extends ConnectionPoolListenerAdapter {
+    private final AtomicInteger waitQueueSize = new AtomicInteger();
+
+    @Override
+    public void waitQueueExited(ConnectionPoolWaitQueueEvent event) {
```
jyemin: Make param final (to check for style errors, run "gradlew check")

driver-core/src/test/functional/com/mongodb/connection/QueueEventsConnectionPoolListener.java
```diff
@@ -0,0 +1,40 @@
...
+        waitQueueSize.decrementAndGet();
+    }
+
+    @Override
+    public void waitQueueEntered(ConnectionPoolWaitQueueEvent event) {
```
jyemin: As above.

jyemin:
Initial review complete.  I'll merge as soon as the comments are addressed.

Thanks,
Jeff

jyemin:
Since I haven't heard back from you in a while, I went ahead and made the changes that I suggested, and pushed to master.

Thanks!

PR #323, opened 2015-07-08
Conventions

evanchooly:
@jyemin @rozza @craiggwilson 

This should cover most of the conventions stuff we've discussed so far.  The biggest outstanding pieces are probably generics and polymorphism support.

### DIFF
--- a/bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodec.java
+++ b/bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodec.java
@@ -59,12 +59,10 @@ public class ClassModelCodec<T extends Object> implements Codec<T> {
     }
 
     @Override
-    public void encode(final BsonWriter writer, final Object value, final EncoderContext encoderContext) {
+    public void encode(final BsonWriter writer, final Object entity, final EncoderContext encoderContext) {
         writer.writeStartDocument();
         for (final FieldModel fieldModel : classModel.getFields()) {
-            writer.writeName(fieldModel.getName());
-            final Codec<Object> codec = fieldModel.getCodec();
-            codec.encode(writer, fieldModel.get(value), encoderContext);
+            fieldModel.store(writer, entity, encoderContext);
         }
         writer.writeEndDocument();
     }
--- a/bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodec.java
+++ b/bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodec.java
@@ -48,7 +48,7 @@ public class ClassModelCodec<T extends Object> implements Codec<T> {
             while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
                 final String name = reader.readName();
                 final FieldModel fieldModel = classModel.getField(name);
-                fieldModel.set(entity, fieldModel.getCodec().decode(reader, decoderContext));
+                fieldModel.decode(entity, reader, decoderContext);
 
             }
             reader.readEndDocument();
@@ -62,7 +62,7 @@ public class ClassModelCodec<T extends Object> implements Codec<T> {
     public void encode(final BsonWriter writer, final Object entity, final EncoderContext encoderContext) {
         writer.writeStartDocument();
         for (final FieldModel fieldModel : classModel.getFields()) {
-            fieldModel.store(writer, entity, encoderContext);
+            fieldModel.encode(entity, writer, encoderContext);
         }
         writer.writeEndDocument();
     }
--- a/bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodecProvider.java
+++ b/bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodecProvider.java
@@ -105,9 +105,11 @@ public class ClassModelCodecProvider implements CodecProvider {
          * Sets the conventions to use when mapping types.
          *
          * @param conventionPack the ConventionPack to use with the ClassModelCodecProvider
+         * @return this
          */
-        public void setConventionPack(final ConventionPack conventionPack) {
+        public ProviderBuilder setConventionPack(final ConventionPack conventionPack) {
             this.conventionPack = conventionPack;
+            return this;
         }
     }
 }
--- a/bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodecProvider.java
+++ b/bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodecProvider.java
@@ -19,8 +19,12 @@ import com.fasterxml.classmate.TypeResolver;
 import org.bson.codecs.Codec;
 import org.bson.codecs.configuration.CodecProvider;
 import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.codecs.configuration.mapper.conventions.ConventionPack;
+import org.bson.codecs.configuration.mapper.conventions.DefaultConventionPack;
 
+import java.util.HashMap;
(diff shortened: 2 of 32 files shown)


evanchooly:
*bump*

bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodec.java
```diff
@@ -59,12 +59,10 @@ public T decode(final BsonReader reader, final DecoderContext decoderContext) {
     }
 
     @Override
-    public void encode(final BsonWriter writer, final Object value, final EncoderContext encoderContext) {
+    public void encode(final BsonWriter writer, final Object entity, final EncoderContext encoderContext) {
```
craiggwilson: I liked value. entity implies something in DDD, and an entity is not a value type. "value" seems sufficiently generic.

bson/src/main/org/bson/codecs/configuration/mapper/ClassModelCodecProvider.java
```diff
@@ -52,12 +60,17 @@ public static ProviderBuilder builder() {
     @Override
     @SuppressWarnings({"unchecked", "rawtypes"})
     public <T> Codec<T> get(final Class<T> clazz, final CodecRegistry registry) {
-        Codec<T> codec = null;
+        ClassModelCodec<?> codec = null;
         if (registered.contains(clazz)) {
-            final ClassModel model = new ClassModel(registry, resolver, (Class<Object>) clazz);
-            codec = (Codec<T>) new ClassModelCodec(model);
+            codec = mapped.get(clazz);
+            if (codec == null && conventionPack.isMappable(clazz)) {
```
craiggwilson: I'm still questioning the isMappable requirement here... For now, I think it is probably fine, but in the long this will turn into something else. For instance, when the user doesn't use conventions to map a class, but rather uses explicit code to handle it manually.

bson/src/main/org/bson/codecs/configuration/mapper/FieldModel.java
```diff
@@ -17,22 +17,36 @@
...
+    private final WeightedValue<Boolean> included = new WeightedValue<Boolean>(true);
+    private final WeightedValue<Boolean> storeNulls = new WeightedValue<Boolean>(true);
+    private final WeightedValue<Boolean> storeEmpties = new WeightedValue<Boolean>(true);
     private final ClassModel owner;
     private final CodecRegistry registry;
-    private final Field field;
-    private Codec<Object> codec;
+    private final ResolvedField field;
+    private Codec<?> codec;
+    private Converter<?, ?> converter = new IdentityConverter();
```
craiggwilson: I thought you got rid of converters...

bson/src/main/org/bson/codecs/configuration/mapper/conventions/BeanPropertiesConvention.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * This convention ensures only fields that are private, nonstatic, nontransient, and nonfinal that have a proper getter and setter as per
+ * the Beans Specification are included in BSON processing.
+ *
+ * @see <a target="_blank" href="http://www.oracle.com/technetwork/articles/javaee/spec-136004.html">JavaBeans Spec</a>
+ */
+public class BeanPropertiesConvention implements Convention {
+    private Integer getWeight() {
+        return Weights.BUILT_IN_CONVENTION;
```
craiggwilson: Since this is the default convention for mapping, this should probably be DEFAULT.

bson/src/main/org/bson/codecs/configuration/mapper/conventions/LowerCaseClassNamingConvention.java
```diff
@@ -0,0 +1,39 @@
...
+import org.bson.codecs.configuration.mapper.ClassModel;
+import org.bson.codecs.configuration.mapper.Weights;
+
+/**
+ * Attempts to configure the collection name as the lower case form of the class name.
+ */
+public class LowerCaseClassNamingConvention implements Convention {
+    @Override
+    public void apply(final ClassModel model) {
+        model.setCollectionName(getWeight(), model.getType().getSimpleName().toLowerCase());
```
craiggwilson: Do you want all lower case, or just camelCase?

evanchooly: Could be debated.  This is mostly here for POC more than any recommendation.  The final set of default conventions has yet to be decided but will probably mirror the c# driver's where it can.

bson/src/main/org/bson/codecs/configuration/mapper/conventions/Converter.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    Class<T> getType();
+
+    /**
+     * Converts a value of type T back to one of type F
+     *
+     * @param value the value to convert
+     * @return the new F value
+     */
+    F unapply(T value);
```
rozza: I'm not sure how well this will play in Scala as `unapply` is special cased at least in objects and maybe its only a convention - see: http://docs.scala-lang.org/tutorials/tour/extractor-objects.html

evanchooly: `apply()` is as well but I don't think it'll be a problem since it's not on
a case class.  @rozza?

*--------------------------------*

*{ *name     : *"**Justin Lee**", *
  title    : *"*Software Engineer*"*,
  twitter  : *"*@evanchooly <http://twitter.com/evanchooly>*"*,
  web      : *[* "10gen.com", *"*antwerkz.com*"* *]*,
  location : *"*New York, NY*"* *}*

rozza: `apply` is fine as it can return type `T`, its more the `unapply` convention of returning `Option[F]` rather than a flat `F`.

I'm not sure it has any real implications, other than perhaps surprising to some Scala users.  It could mean you have to use a class rather than an `object` when creating custom `Converter`s.

evanchooly: I'm open to changing the names if it'd help.  `apply()` felt the natural
choice which led to `unapply()` which, i'll admit, was a nod to Scala.  We
could change to `convert()`/`unconvert()`  or `deconvert()`.  I'm just not
sure what a decent name would be.

*--------------------------------*

*{ *name     : *"**Justin Lee**", *
  title    : *"*Software Engineer*"*,
  twitter  : *"*@evanchooly <http://twitter.com/evanchooly>*"*,
  web      : *[* "10gen.com", *"*antwerkz.com*"* *]*,
  location : *"*New York, NY*"* *}*

bson/src/main/org/bson/codecs/configuration/mapper/conventions/LowerCaseClassNamingConvention.java
craiggwilson: kk. In this case, we don't store collection names with our codecs, so you can do what you want.

bson/src/main/org/bson/codecs/configuration/mapper/conventions/Converter.java
rozza: It could help, `to()` and `from()` could work as its all ready in the documentation, or expanded out to `convertFrom` etc...  Naming things is hard!  ;)

jyemin:
Closing this, as we're not ready to pull the trigger on this work.

PR #324, opened 2015-07-29
Update installation-guide.md

saarp:
Would save a lot of time if the repository information was included.

### DIFF
--- a/docs/reference/content/driver/getting-started/installation-guide.md
+++ b/docs/reference/content/driver/getting-started/installation-guide.md
@@ -28,6 +28,18 @@ For OSGi-based applications: due to the fact that there are classes from the `co
 jar (described below) instead. 
 {{% /note %}}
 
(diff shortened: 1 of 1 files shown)


evanchooly:
For the released artifacts, the standard maven repository is used and doesn't need configuration.  This repository is used when you want -SNAPSHOT artifacts.

rozza:
Closing as per @evanchooly 's explanation.  

The artifacts are also published on Maven Central making this not required for non -SNAPSHOT releases. The docs do include information for how to setup for snapshots in the same style as you've included.

Sometimes it can take a few minutes for sonatype to sync to maven central which may have been the reason you needed to manually add sonatype coordinates.

saarp:
Thanks for the update. For some reason my local config was not fetching the artifact from Maven Central, possibly due to an incorrectly configured local proxy.

PR #326, opened 2015-08-14
fix typos in javadocs

alb-i986:
As per title.


jyemin:
Much obliged.  Thanks!

PR #327, opened 2015-08-25
Fix header-content alignment

m-vojvodic:
Minor fix to nav-items content alignment on the docs landing page

### DIFF
--- a/docs/landing/static/s/css/frontpage.css
+++ b/docs/landing/static/s/css/frontpage.css
@@ -48,8 +48,8 @@ dt {
 }
 
 #header-db .header-content {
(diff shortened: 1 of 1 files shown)


rozza:
The downside to this is it floats the nav items which then wrap on a smaller screen (when in fact we probably should just hide them).

rozza:
Merged in: 39c8691e6250a107551465c943b8f91c0d7ec136

Added hidden-xs to topRight.html in 0cc1f7018d16621fe13c56fddcab64128282ffe0

rozza:
Thanks @m-vojvodic !

PR #330, opened 2015-10-20
Added WriteConcern support to find and modify operations

rozza:
JAVA-908

### DIFF
--- a/driver-core/src/main/com/mongodb/operation/FindAndDeleteOperation.java
+++ b/driver-core/src/main/com/mongodb/operation/FindAndDeleteOperation.java
@@ -17,9 +17,15 @@
 package com.mongodb.operation;
 
 import com.mongodb.MongoNamespace;
+import com.mongodb.WriteConcern;
 import com.mongodb.async.SingleResultCallback;
 import com.mongodb.binding.AsyncWriteBinding;
 import com.mongodb.binding.WriteBinding;
+import com.mongodb.connection.AsyncConnection;
+import com.mongodb.connection.Connection;
+import com.mongodb.connection.ConnectionDescription;
+import com.mongodb.operation.OperationHelper.AsyncCallableWithConnection;
+import com.mongodb.operation.OperationHelper.CallableWithConnection;
 import org.bson.BsonBoolean;
 import org.bson.BsonDocument;
 import org.bson.BsonString;
@@ -28,10 +34,14 @@ import org.bson.codecs.Decoder;
 import java.util.concurrent.TimeUnit;
 
 import static com.mongodb.assertions.Assertions.notNull;
+import static com.mongodb.internal.async.ErrorHandlingResultCallback.errorHandlingCallback;
 import static com.mongodb.operation.CommandOperationHelper.executeWrappedCommandProtocol;
 import static com.mongodb.operation.CommandOperationHelper.executeWrappedCommandProtocolAsync;
 import static com.mongodb.operation.DocumentHelper.putIfNotNull;
 import static com.mongodb.operation.DocumentHelper.putIfNotZero;
+import static com.mongodb.operation.OperationHelper.releasingCallback;
+import static com.mongodb.operation.OperationHelper.serverIsAtLeastVersionThreeDotTwo;
+import static com.mongodb.operation.OperationHelper.withConnection;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 /**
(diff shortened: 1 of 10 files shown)


driver-core/src/main/com/mongodb/operation/FindAndDeleteOperation.java
```diff
@@ -47,6 +57,7 @@
     private BsonDocument projection;
     private BsonDocument sort;
     private long maxTimeMS;
+    private WriteConcern writeConcern;
```
jyemin: I'd like these classes to be more consistent with BaseWriteOperation. To do so would require the following changes:

* Add a new constructor that takes a non-null WriteConcern parameter
* Deprecate the existing constructor  (so as not to break binary compatibility for any users of core)
* Remove the new writeConcern setter method
* For backwards compatibility, default the writeConcern field to ACKNOWLEDGED
* When creating the command document, instead of checking for null, check for WriteConcern.isServerDefault()  (as in BaseWriteCommandMessage)

driver-core/src/main/com/mongodb/operation/FindAndDeleteOperation.java
```diff
@@ -169,28 +180,69 @@ public BsonDocument getSort() {
...
 
-    private BsonDocument getFindAndRemoveDocument() {
+    private BsonDocument asCommandDocument(final ConnectionDescription description) {
         BsonDocument command = new BsonDocument("findandmodify", new BsonString(namespace.getCollectionName()));
         putIfNotNull(command, "query", getFilter());
         putIfNotNull(command, "fields", getProjection());
         putIfNotNull(command, "sort", getSort());
         putIfNotZero(command, "maxTimeMS", getMaxTime(MILLISECONDS));
         command.put("remove", BsonBoolean.TRUE);
+        if (writeConcern != null & serverIsAtLeastVersionThreeDotTwo(description)) {
```
jyemin: This should be:

     serverIsAtLeastVersionThreeDotTwo(description) && !writeConcern.isServerDefault() && writeConcern.isAcknowledged()

as I don't think the driver should be sending 

    writeConcern : { w : 0}

to the server.

driver/src/test/unit/com/mongodb/MongoCollectionSpecification.groovy
```diff
@@ -606,7 +606,9 @@ class MongoCollectionSpecification extends Specification {
     def 'should use FindOneAndDeleteOperation correctly'() {
         given:
         def collection = new MongoCollectionImpl(namespace, Document, codecRegistry, readPreference, writeConcern, executor)
-        def expectedOperation = new FindAndDeleteOperation(namespace, new DocumentCodec()).filter(new BsonDocument('a', new BsonInt32(1)))
+        def expectedOperation = new FindAndDeleteOperation(namespace, new DocumentCodec())
+                .filter(new BsonDocument('a', new BsonInt32(1)))
+                .writeConcern(writeConcern)
```
jyemin: Given the above comments, should add a test with at least one non-default but still acknowledged write concern.

driver-core/src/test/functional/com/mongodb/operation/FindAndDeleteOperationSpecification.groovy
```diff
@@ -107,4 +160,113 @@ class FindAndDeleteOperationSpecification extends OperationFunctionalSpecificati
...
+        }
+
+        operation.execute(writeBinding)
+
+        then:
+        1 * connection.command(getNamespace().getDatabaseName(), expectedCommand, _, _, _) >> cannedResult
+        2 * connection.release()
+
+        where:
+        serverVersion << [new ServerVersion([3, 0, 0]), new ServerVersion([3, 2, 0])]
```
jyemin: Given above comment, should test with three different write concerns:

* ACKNOWLEDGED
* UNACKNOWLEDGED
* W1

jyemin:
One thing missing: writeConcern should be supported in DBCollection as well for all the findAndModify helpers.

jyemin:
Initial review complete.

driver-core/src/main/com/mongodb/operation/FindAndDeleteOperation.java
rozza: Done

driver-core/src/main/com/mongodb/operation/FindAndDeleteOperation.java
rozza: Done

driver-core/src/test/functional/com/mongodb/operation/FindAndDeleteOperationSpecification.groovy
rozza: Done

rozza:
Updated

jyemin:
lgtm

PR #331, opened 2015-11-01
Fix a couple of typos

eltimn:


jyemin:
Thanks for the sharp eyes.  Keep 'em coming if you spot more.

PR #333, opened 2015-11-07
Update DocumentCodec.java

ralscha:
The assertion is unnecessary because "new ObjectIdGenerator()" is never not null


rozza:
Thanks @ralscha this has been merged!

PR #334, opened 2015-11-07
instanceof can be used here instead of isAssignableFrom

ralscha:
The right side is known at compile time.
And it's a tiny tiny tiny bit faster
http://stackoverflow.com/questions/496928/what-is-the-difference-between-instanceof-and-class-isassignablefrom


rozza:
Thanks @ralscha this has been merged!

PR #336, opened 2015-11-16
Add 2.14 reference documentation

jyemin:
Most of the links had already been updated to 2.13.

Note there is no What's New page yet, as there is with 3.x releases, but probably makes sense to add one.

### DIFF
--- a/docs/reference/content/getting-started/installation-guide.md
+++ b/docs/reference/content/getting-started/installation-guide.md
@@ -17,7 +17,7 @@ The recommended way to get started using one of the drivers in your project is w
 
 This jar that contains everything you need including the BSON library.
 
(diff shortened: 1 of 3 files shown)


rozza:
We should be able to use the `apiref` helper - so we dont have to change urls for the apidocs.

jyemin:
Using apiref, but the internal links like WriteConcern#ACKNOWLEDGED don't work.  I can't remember why...

rozza:
We had to add extra options for the java doc including tracking - that should solve it:
See: https://github.com/mongodb/mongo-java-driver/blob/master/build.gradle#L89-L110

jyemin:
I'm not going to mess with 2.x javadoc generation now, so I think it's best to revert the change to use apiref.

rozza:
No problems - LGTM

PR #337, opened 2015-11-19
Update index.md

steowens:
Disclaimer so mere mortals don't waste their time reading this documentation.


rozza:
I understand you are frustrated @steowens and are venting that frustration, but this really isn't the best way to achieve anything positive.  

Have you tried emailing the [mongodb-user group](https://groups.google.com/forum/#!forum/mongodb-user) asking for help on this issue?  We do actively monitor the user group and help users.

Can I ask you to open a [Jira](http://jira.mongodb.org/browse/JAVA) ticket with more specifics about what the API isn't achieving for you?  This way we can work on it and help other new users to the API.

All the best,

Ross

PR #343, opened 2016-03-10
JAVA-2149: add ability to configure NettyStream channel type in async driver

jamel:


rozza:
Thanks @jamel we used this PR as the basis for the improvement 👍

jamel:
Cool! Thanks a lot!

PR #344, opened 2016-03-23
BasicBSONObjectBuilder : A builder for creating BasicBSONObject model…

mutaherul:
…s from scratch

 A builder for creating BasicBSONObject models from scratch.
    Utility for building complex BasicBSONObject.
    This builder and provides methods to add name/value pairs to the object model and to return the resulting object.
    The methods in this class can be chained to add multiple name/value pairs to the object.
    This class does not allow null to be used as a name(key/field) while building the BasicBSON object
    Example Code:
    # BasicBSONObjectBuilder.getInstance().add(key,value).add(key,mapValue).build()  
    # BasicBSONObject  bObject =  BasicBSONObjectBuilder.createObject(map)

### DIFF
--- a/BasicBSONObjectBuilder.java
+++ b/BasicBSONObjectBuilder.java
@@ -1,5 +1,5 @@
 
-package com.infolytx.utils.bson;
+package com.mongodb;
 import java.security.InvalidParameterException;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


rozza:
Hi @mutaherul,

Firstly, apologies for the lack of response on this pull request. Thanks for sharing your code and creating the pull request. I can see the utility of your builder and hopefully it will be of interest to other users.  However, as its working with the old BsonObject API we have no plans to incorporate into the  driver.  The 3.x version of the driver introduced the new `Document` and `BsonDocument` classes as the new preferred way of creating Bson Documents.

All the best,

Ross

mutaherul:
Thank you for your response. I used this at the project where I have to deal with  com.mongodb.hadoop.io.BSONWritable  ( BSONWritable(BSONObject doc)  ) . Can the new `Document`   be possible to pass at BSONWritable ?  I am using mongo-hadoop-core  (V 2.0.1) in my project .

rozza:
@mutaherul I don't believe it can - the hadoop connector still works with the older style BSONObject and BSONEncoders / Decoders.

PR #346, opened 2016-03-31
JAVA-1280 : add Simple type mode

Dufgui:
Hello,

i try to respond to this jira because i experiment some bugs when i try to parse my json from Document.toJson() with a jackson project (emfjson-jackson and emfjson-mongo).

Hope it can help other people ;o)

### DIFF
--- a/bson/src/main/org/bson/json/JsonMode.java
+++ b/bson/src/main/org/bson/json/JsonMode.java
@@ -34,5 +34,9 @@ public enum JsonMode {
      * While not formally documented, this output mode will attempt to produce output that corresponds to what the MongoDB shell actually
      * produces when showing query results.
      */
-    SHELL
+    SHELL,
+    /**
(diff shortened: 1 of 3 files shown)


rozza:
Hi @Dufgui,

Firstly, apologies for the lack of response on this pull request. I like the idea behind your code however, as a drivers team we need to ensure that we are able to interoperate with the different drivers and server tools (such as MongoImport). Because of this I can't consider merging your code into the Java driver. 

However, if you feel strongly that the MongoDB Drivers really should support this feature, then I urge you to file a ticket to the [DRIVERS](https://jira.mongodb.org/browse/DRIVERS) project.  If you explain the rational for new feature and the conversions to and from Bson types, then as a drivers team we can best look at the feature.

All the best,

Ross

Dufgui:
See https://jira.mongodb.org/browse/DRIVERS-342

PR #347, opened 2016-04-11
JAVA-2169: update the connection release order in pool

Gauravshah:
Release the connection to the pool in a different order to how it is picked. Helps avoid using the same buffer again and again

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
@@ -93,7 +93,7 @@ public class ConcurrentPool<T> implements Pool<T> {
         if (prune) {
             close(t);
         } else {
-            available.addLast(t);
+            available.addFirst(t);
         }
 
         releasePermit();
--- a/driver-core/src/test/unit/com/mongodb/internal/connection/ConcurrentPoolTest.java
+++ b/driver-core/src/test/unit/com/mongodb/internal/connection/ConcurrentPoolTest.java
@@ -23,11 +23,7 @@ import org.junit.Test;
 import java.io.Closeable;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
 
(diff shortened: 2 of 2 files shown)


Gauravshah:
same tests are failing on master since commit d12a26efad19f876e53829c8a8444d5e5340c765

jyemin:
The failures are due to Checkstyle issues in ConcurrentPoolTest:

[ant:checkstyle] /home/travis/build/mongodb/mongo-java-driver/driver-core/src/test/unit/com/mongodb/internal/connection/ConcurrentPoolTest.java:26: Using the '.*' form of import should be avoided - org.junit.Assert.*.
[ant:checkstyle] /home/travis/build/mongodb/mongo-java-driver/driver-core/src/test/unit/com/mongodb/internal/connection/ConcurrentPoolTest.java:70:31: ',' is not followed by whitespace.

Gauravshah:
thanks @jyemin fixed it.

jyemin:
One issue with the change is that it doesn't interact well with the maxConnectionIdleTime setting, which is sometimes used to handle short-term connection spikes.  It lets the connection pool temporarily expand to its maximum allowed size during a spike, and then shrink back down to its minimum allowed size as load decreases and connections are idle. But this relies on the connection pool being treated as as LIFO queue rather than a FIFO queue.  Otherwise, all the connections are cycled through and none are ever idle for long, unless usages effectively drops to zero.

While it would be possible to detect this and switch to FIFO if maxConnectionIdleTime is set, that seems to violate the element of least surprise for users.

Gauravshah:
I understand the concern there. Is there a better way to solve this ? I am
not able to think of a clean solution.

jyemin:
Perhaps the answer lies not in tweaking the connection pool implementation but rather in exposing a new API.  The [MongoDB driver for golang](https://godoc.org/labix.org/v2/mgo) has a Session-based API that allows the user finer-grained control over what socket the application is using.   An approach like that might be applicable for this use case, as then the application itself could reserve as many sockets as it needed for unacknowledged writes, and use a different set for other operations.

Gauravshah:
Probably that sounds better.

I don't think I know java that well to implement new API.

rozza:
Hi @Gauravshah,

I'm going to close this pull request for the reasons discussed above. Feel free to add a ticket to the [JAVA](https://jira.mongodb.org/browse/JAVA) project with the suggested improvement.

All the best,

Ross

Gauravshah:
create issues in https://jira.mongodb.org/browse/JAVA-2169

PR #348, opened 2016-04-29
Log offensive document when exceeding maxDocumentSize

andresoviedo:
Please log the document id when exceeding the max document size when doing bulk updates to mongo with the java driver.  Otherwise it is almost impossible to know which is the offensive document. 

The only workaround to this is to split the bulk update (in my case of thousand documents), to do the update 1 by 1 and print the id when the exception is thrown.

I opened a ticket in JIRA to throw the requested information in the exception. But this pull is only for writing a message to file log. I have tested it with 3.2.0 version and it didn't break anything. This is the related ticket: [https://jira.mongodb.org/browse/JAVA-2180]

Thank you.

### DIFF
--- a/driver/src/main/com/mongodb/DBObjectCodec.java
+++ b/driver/src/main/com/mongodb/DBObjectCodec.java
@@ -23,6 +23,7 @@ import org.bson.BsonDbPointer;
 import org.bson.BsonDocument;
 import org.bson.BsonDocumentWriter;
 import org.bson.BsonReader;
+import org.bson.BsonSerializationException;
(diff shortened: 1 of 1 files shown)


rozza:
Hi @andresoviedo,

Firstly, apologies for the lack of response on this pull request. We've been busy working on the new 3.4 version of the driver.

If we implement this ticket we'll have to do it across the various bson document codecs (DBObject, Document and BsonDocument). Because of that I'm going to close this pull request and any follow up will be done in the scope of JAVA-2180.

All the best,

Ross

PR #349, opened 2016-05-04
JAVA-2186 Add the ability to parse/output a BasicDBList

gsmet:
For now in Hibernate OGM, we only need the ability to decode a BasicDBList.

I also provided the ability to output a BasicDBList which needed some tweaking to AbstractBsonWriter. The other option might be to simply throw an unsupported exception when trying to encode a BasicDBList.

Comments welcome.

### DIFF
--- a/bson/src/main/org/bson/AbstractBsonWriter.java
+++ b/bson/src/main/org/bson/AbstractBsonWriter.java
@@ -307,7 +307,11 @@ public abstract class AbstractBsonWriter implements BsonWriter, Closeable {
 
     @Override
     public void writeStartArray() {
-        checkPreconditions("writeStartArray", State.VALUE);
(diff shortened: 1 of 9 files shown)


rozza:
Hi @gsmet,

Firstly, apologies for the lack of response on this pull request. Thanks for sharing your code and creating the pull request.  I can see the utility of your code and hopefully it will be of interest to other users. However, as its working with the old DBObject API we have no plans to incorporate into the driver. The 3.x version of the driver introduced the new `Document` and `BsonDocument` classes as the new preferred way of creating Bson Documents.  

However, as your code includes the codec (`BasicDBListCodec`) you can still use that to handle the conversion of Json to and from BasicDBLists without the need for an explicit helper method.

All the best,

Ross

PR #350, opened 2016-05-04
JAVA-2185 Add support for NumberInt/$numberInt in the JsonReader

gsmet:
As mentioned in the issue, it's only added in the JsonReader as it causes all sorts of problems if added in the JsonWriter. AFAICS, judging from the tickets open, the server has the same behavior.


rozza:
Hi @gsmet,

Firstly, apologies for the lack of response on this pull request. Thanks for sharing your code and creating the pull request. As you've mentioned there are some tickets open regarding this issue and as a drivers team we need to ensure that we are able to interoperate with the different drivers and server tools (such as MongoImport). Because of this internally we are currently looking at formalising the extended Json specification.  Once this has been completed we'll look to unify in our approach to dealing with Json. 

I'll update JAVA-2185 with progress  in the future.

All the best,

Ross

PR #352, opened 2016-05-25
Fix NaN handling

xardazz:
Currently if BasicDBObject conatins Double.NaN toString() and toJSON() methods produce invalid json - NaN isn't escaped with quotes. This PR fixes it.


xardazz:
@rozza Can you please review?

rozza:
Thanks @xardazz for the PR - I'll make sure we review in time for the 3.3 release.  
Will need to test / support `Document` and `BsonDocument` as well.

rozza:
Hi @xardazz,

Firstly, apologies for the lack of response on this pull request. As a drivers team we need to ensure that we are able to interoperate with the different drivers and server tools (such as MongoImport). Because of this internally we are currently looking at formalising the extended Json specification.  Once this has been completed we'll look to unify our approach to dealing with Json.

All the best,

Ross

PR #353, opened 2016-06-10
JAVA-2216 - add partialFilterExpression to indexRequest in async MongoCollectionImpl

ekandlen:


rozza:
Thanks @ekandlen for the ticket and the patch

PR #355, opened 2016-06-14
JAVA-2224 Support the /pattern/options syntax for regexps in JsonReader

gsmet:
https://jira.mongodb.org/browse/JAVA-2224

It would be nice if it could get backported into 3.2.

### DIFF
--- a/bson/src/main/org/bson/json/JsonReader.java
+++ b/bson/src/main/org/bson/json/JsonReader.java
@@ -936,10 +936,19 @@ public class JsonReader extends AbstractBsonReader {
     private BsonRegularExpression visitRegularExpressionExtendedJson() {
         verifyToken(":");
         JsonToken patternToken = popToken();
-        if (patternToken.getType() != JsonTokenType.STRING) {
-            throw new JsonParseException("JSON reader expected a string but found '%s'.", patternToken.getValue());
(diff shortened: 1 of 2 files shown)


raohenglena:
Is the completing of the API about GridFSFile on your schedu?

` 
for example:
    @Override
    public void putAll(final BSONObject o) {
        throw new UnsupportedOperationException();
    }

    @Override
    public void putAll(@SuppressWarnings("rawtypes") final Map m) {
        throw new UnsupportedOperationException();
    }

    @Override
    public Map<?, ?> toMap() {
        throw new UnsupportedOperationException();
    }

    @Override
    public Object removeField(final String key) {
        throw new UnsupportedOperationException();
    }`

jyemin:
Please see my latest comment on https://jira.mongodb.org/browse/JAVA-2224.

rozza:
Hi @gsmet,

As mentioned in #350 we are currently looking at formalising the extended Json specification. Once this has been completed we'll look to unify in our approach to dealing with Json. I'll update JAVA-2224 with progress in the future.

I should have thanked you in the previous PR's for your work, they have helped ensure we look at how we deal with JSON across the drivers / server tools.

All the best,

Ross

PR #356, opened 2016-06-15
Add toString() for com.mongodb.BulkWriteError

Powerrr:
This adds `toString()` implementation for `com.mongodb.BulkWriteError` (similar to the one in `driver-core/src/main/com/mongodb/bulk/BulkWriteError.java`). I suspect that this class is some sort of "old" API, but it is used by mongo-hadoop. And other similar Errors have string representations, so this one should too.
I believe this closes https://jira.mongodb.org/browse/JAVA-1903 and https://jira.mongodb.org/browse/HADOOP-214.


Powerrr:
Already done in 0fa058cdd1530bc4bc078305e8955b7a90d54598.

PR #357, opened 2016-06-15
Fix typo in BulkWriteError.hashCode()

Powerrr:
This fixes small error in `com.mongodb.bulk.BulkWriteError.hashCode()` (`index` field was effectively not used in hash code), which was introduced during parent class extraction.


rozza:
Thanks @Powerrr!

PR #360, opened 2016-07-08
Javadoc: Fix result of MongoCollection.updateMany

squarejaw:
The `@return` is slightly wrong for `MongoCollection.updateMany`.


rozza:
Many thanks @squarejaw I have merged into master.

PR #361, opened 2016-07-14
correct ipv6 address handling when building MongoClients

evanchooly:
fixes JIRA-2245

### DIFF
--- a/driver-core/src/main/com/mongodb/ServerAddress.java
+++ b/driver-core/src/main/com/mongodb/ServerAddress.java
@@ -109,7 +109,7 @@ public class ServerAddress implements Serializable {
                 }
                 portToUse = Integer.parseInt(host.substring(portIdx + 2));
             }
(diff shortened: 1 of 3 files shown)


driver-core/src/main/com/mongodb/ServerAddress.java
```diff
@@ -129,6 +129,16 @@ public ServerAddress(final String host, final int port) {
         this.port = portToUse;
     }
 
+    private String join(final String delimiter, final String[] array, final int start, final int end) {
```
rozza: Don't think this method is needed.

rozza:
Other than the join method - LGTM

driver-core/src/main/com/mongodb/ServerAddress.java
evanchooly: You're right.  Was using that for a failed approach and missed it in the clean up.

jyemin:
lgtm

PR #362, opened 2016-07-22
JAVA-2257 Handle autoIndexId without capped option enabled

ia3andy:
It's seems that autoIndexId can't be activated without enabling capped option when creating a collection.
Since it is possible using the mongo shell, why blocking this action on the java driver.

Thanks,

### DIFF
--- a/driver-core/src/test/functional/com/mongodb/operation/CreateCollectionOperationSpecification.groovy
+++ b/driver-core/src/test/functional/com/mongodb/operation/CreateCollectionOperationSpecification.groovy
@@ -185,6 +185,29 @@ class CreateCollectionOperationSpecification extends OperationFunctionalSpecific
         }
     }
 
+    def 'should create collection in respect to the autoIndex option'() {
+        given:
+        assert !collectionNameExists(getCollectionName())
+
+        when:
+        new CreateCollectionOperation(getDatabaseName(), getCollectionName())
+                .autoIndex(autoIndex)
+                .execute(getBinding())
+
+        then:
+        def stats = new CommandWriteOperation<Document>(getDatabaseName(),
+                new BsonDocument('collStats', new BsonString(getCollectionName())),
+                new DocumentCodec()).execute(getBinding())
+
+        expect:
+        stats.getInteger('nindexes') == expectedNumberOfIndexes
+
+        where:
+        autoIndex | expectedNumberOfIndexes
+        true | 1
+        false | 0
+    }
+
     @IgnoreIf({ !serverVersionAtLeast(asList(3, 1, 8)) })
     def 'should allow indexOptionDefaults'() {
         given:
--- a/driver-core/src/main/com/mongodb/operation/CreateCollectionOperation.java
+++ b/driver-core/src/main/com/mongodb/operation/CreateCollectionOperation.java
@@ -317,10 +317,10 @@ public class CreateCollectionOperation implements AsyncWriteOperation<Void>, Wri
 
     private BsonDocument asDocument() {
         BsonDocument document = new BsonDocument("create", new BsonString(collectionName));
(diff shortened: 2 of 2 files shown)


jyemin:
Thanks for submitting this pull request.

Please run ./gradlew check on your branch, as it appears from Travis that the patch is violating our Groovy test coding standards.

Also, as per our [contributing guidelines](https://github.com/mongodb/mongo-java-driver/blob/master/CONTRIBUTING.md), please also open an issue in our [JIRA](https://jira.mongodb.org/browse/JAVA) project so that we have an issue to link this pull request to.

Also, please note in our [documentation](https://docs.mongodb.com/manual/reference/method/db.createCollection/) that autoIndexId has been deprecated in the MongoDB server as of the 3.2 release and is scheduled for removal in 3.4.  
 
Thanks,
Jeff

ia3andy:
The issue has been created (since even if it will be deprecated, it's still a bug).

Talking about my use case, I'm looking for a way to "archive/restore" a read-only big data collection on the fly when it hasn't been used for a while (because of indexes size). I was thinking about just removing indexes, but since autoIndex=false will be deprecated (I can't get rid of _id index size), do you suggest using mongo dump/restore instead? Do you have some plan on doing such a feature?

Thanks,
Andy.

driver-core/src/test/functional/com/mongodb/operation/CreateCollectionOperationSpecification.groovy
```diff
@@ -185,6 +185,46 @@ class CreateCollectionOperationSpecification extends OperationFunctionalSpecific
...
+
+        when:
+        new CreateCollectionOperation(getDatabaseName(), getCollectionName())
+                .autoIndex(true)
+                .execute(getBinding())
+
+        then:
+        collectionNameExists(getCollectionName())
+
+        when:
```
jyemin: Couple of suggestions for the tests:

1. Combine the two tests into one using Spock's where syntax: e.g.

    where:
    autoIndex  | expectedNumberOfIndexes
    true            | 1
    false           | 0

2. Combine the two "then" clauses.  The "when" clause is for the code that is under test, so the second one is a mis-use of it, and can be folded into the "then" clause

jyemin:
I suggest that you ask your archiving question in the mongodb-users Google group, as this isn't the right forum for that sort of discussion.

I have a few suggestions about the tests.  After handling those, please
squash the commits and prefix the single commit message with "JAVA-2257: ", and we'll get this merged.

Thanks,
Jeff

driver-core/src/test/functional/com/mongodb/operation/CreateCollectionOperationSpecification.groovy
```diff
@@ -185,6 +185,29 @@ class CreateCollectionOperationSpecification extends OperationFunctionalSpecific
...
+        new CreateCollectionOperation(getDatabaseName(), getCollectionName())
+                .autoIndex(autoIndex)
+                .execute(getBinding())
+
+        then:
+        def stats = new CommandWriteOperation<Document>(getDatabaseName(),
+                new BsonDocument('collStats', new BsonString(getCollectionName())),
+                new DocumentCodec()).execute(getBinding())
+
+        expect:
```
jyemin: Closer... Generally there is no need for to follow "then:" with "expect:".  Just get rid of the "expect:" and even the stats local variable.

ia3andy: Sorry, I'm not familiar with Spock, it's seems to me that all the test file should be rewritten from what you said because it's using the pattern I've been copying from. This change should be in another pull request since it's not part of the bug fixing.

Thanks,



Thanks,

jyemin:
Merged via command line.  Thanks again!

PR #363, opened 2016-07-28
Fix read concern/preference propagation from connection uri for async…

sunjaec:
…-driver

For async driver, the primary method of configuring the client is to specify the options
in the connection string. However, not all options are honored during client creation.

The expectation is that what was set in the connection uri should be the default...


rozza:
Many thanks @sunjaec for reporting this.  

We've added [JAVA-2270](https://jira.mongodb.org/browse/JAVA-2270) to track the progress. We'll take your patch and add some regression test cases as well.

Ross

sunjaec:
Thanks for the quick response. Looking forward to having the fix generally available.

rozza:
[JAVA-2270](https://jira.mongodb.org/browse/JAVA-2270) has now been fixed and is awaiting release.

PR #365, opened 2016-08-22
JAVA-1977 : Updated readers/writers and codecs pages with worker example

vedala:


rozza:
@vedala thanks for the ticket and your work.  We've realised the brittleness in the WorkerCodec implementation around the field order, so we decided not to include it in the documentation. As it could lead others to write brittle codecs.

However, we are working on [JAVA-1812](https://jira.mongodb.org/browse/JAVA-1812) and hope to have a better solution in the future.

All the best,

Ross

vedala:
Thanks, Ross. I look forward to the solution to be provided by JAVA-1812.

PR #366, opened 2016-08-26
update documentation for running tests to explicitly mention required mongod parameter

tdyas:
Update the documentation to explicitly mention the mongod parameter that must be for the mongod used for test runs to be properly configured. (CONTRIBUTING.md did not mention the requirement.)


rozza:
Thanks for the PR @tdyas!

PR #367, opened 2016-08-26
Changed the current year

upa8:


jyemin:
Thanks for noticing that!

upa8:
You are most welcome . 
Found similar change in all the scala drivers document as well . 
Please have a look .

PR #368, opened 2016-08-26
JAVA-2300: add conversions of ObjectId to/from java.nio.ByteBuffer

tdyas:
Add conversions to/from java.nio.ByteBuffer via the new putToByteBuffer method and a new
constructor. This is useful for users who want to avoid the GC overhead of allocating a
byte array when serializing an ObjectId to bytes.

A good example of where this is useful is when serializing or deserializing an `ObjectId` within a larger ByteBuffer where it does not make sense to allocate a byte array for the sole purpose of extracting the `ObjectId` from the larger buffer.

### DIFF
--- a/bson/src/main/org/bson/types/ObjectId.java
+++ b/bson/src/main/org/bson/types/ObjectId.java
@@ -244,17 +244,10 @@ public final class ObjectId implements Comparable<ObjectId>, Serializable {
      * @throws IllegalArgumentException if array is null or not of length 12
      */
     public ObjectId(final byte[] bytes) {
-        if (bytes == null) {
-            throw new IllegalArgumentException();
-        }
-        if (bytes.length != 12) {
-            throw new IllegalArgumentException("need 12 bytes");
-        }
-
-        timestamp = makeInt(bytes[0], bytes[1], bytes[2], bytes[3]);
-        machineIdentifier = makeInt((byte) 0, bytes[4], bytes[5], bytes[6]);
-        processIdentifier = (short) makeInt((byte) 0, (byte) 0, bytes[7], bytes[8]);
-        counter = makeInt((byte) 0, bytes[9], bytes[10], bytes[11]);
+        // This null check allows the delegate constructor to throw the expected IllegalArgumentException.
+        // (ByteBuffer.wrap throws NullPointerException if bytes is null, which violates ObjectId's contract
+        // to callers.)
+        this(bytes != null ? ByteBuffer.wrap(bytes) : null);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tdyas:
Any response to this feature?

bson/src/main/org/bson/types/ObjectId.java
```diff
@@ -308,6 +330,35 @@ public ObjectId(final byte[] bytes) {
     }
 
     /**
+      * Convert to bytes and put those bytes to the provided ByteBuffer.
+      * Note that the numbers are stored in big-endian order.
+      *
+      * @param buffer the ByteBuffer
+      * @throws IllegalArgumentException if the buffer is null or does not have at least 12 bytes remaining
+      */
+    public void putToByteBuffer(final ByteBuffer buffer) {
```
jyemin: DRY this up by implementing toByteArray in terms of it

bson/src/main/org/bson/types/ObjectId.java
```diff
@@ -268,6 +268,28 @@ public ObjectId(final byte[] bytes) {
         this(legacyToBytes(timestamp, machineAndProcessIdentifier, counter));
     }
 
+    /**
+     * Constructs a new instance from the given ByteBuffer
+     *
+     * @param buffer the ByteBuffer
+     * @throws IllegalArgumentException if the buffer is null or does not have at least 12 bytes remaining
+     */
+    public ObjectId(final ByteBuffer buffer) {
```
jyemin: DRY this up by implementing ObjectId(final byte[] bytes) in terms of this

bson/src/test/unit/org/bson/types/ObjectIdTest.java
```diff
@@ -58,6 +59,33 @@ public void testBytes() {
     }
 
     @Test
+    public void testByteBuffer() {
```
jyemin: I think it would be clearer if, instead of adding a new test, you just add another assertion to testToByteArray and to testFromByteArray (perhaps renaming to testToBytes and testFromBytes)

bson/src/main/org/bson/types/ObjectId.java
tdyas: Will do

bson/src/main/org/bson/types/ObjectId.java
tdyas: Will do

tdyas: Writing toByteArray in terms of putToByteBuffer will incur an allocation of a ByteBuffer that will just be thrown away (and same for the ByteBuffer.wrap in the constructor). Is that okay?

tdyas: Done.

bson/src/main/org/bson/types/ObjectId.java
tdyas: Done.

jyemin:
Looks great.  I'm just going to make a few additions to conform to our coding standards (@since javadoc annotations, etc.) and will merge prior to the 3.4.0 release candidate.

Thanks!

jyemin:
Merged after amending the commit to add the Jira issue to the commit message.

PR #372, opened 2016-09-07
JAVA-2301 add read preferences to schema read operations list DB,DBna…

tgrall:
Add read preferences to schema read operations list DB,DBname, Collection, Index

Test:
- All existing code is running with the new methods (using the new code path)
- Added few path with a ReadPreference.primaryPreferred

Functional Testing:
 I have encountered the bug while using MongoDB & Apache Drill in a context where Drill does not have access at all to primary.
- Tested with these patch and it is running as expected (get the schema information from secondary without no access to the primary - different network)

### DIFF
--- a/driver-async/src/main/com/mongodb/async/client/MongoClient.java
+++ b/driver-async/src/main/com/mongodb/async/client/MongoClient.java
@@ -16,6 +16,7 @@
 
 package com.mongodb.async.client;
 
+import com.mongodb.ReadPreference;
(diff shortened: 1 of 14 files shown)


jyemin:
Hi Tug,

Thank you for the pull request.  I commented on https://jira.mongodb.org/browse/JAVA-2301 to explain why it works the way it does.  Let's continue the discussion there.

rozza:
Closing as per: JAVA-2301

PR #375, opened 2016-09-18
Update Gradle

kazuki-ma:
Result of `gradle clean wrapper --gradle-version 3.0 --no-daemon`

If 3.0 is too new, stitching to 2.14.1.


rozza:
Hi @kazuki-ma,

Thanks for the PR - we may look to upgrade gradle in the future, if the current version doesn't meet our need. We just want to minimise any potential risk by upgrading in isolation to other build changes.

All the best,

Ross

rozza:
Hi @kazuki-ma just to let you know - we have updated gradle to 2.14.1 in aa82a6de6c34124e0645505b812467d7904a39bb

PR #376, opened 2016-11-07
Added `collection` back to `MongoClientURI` and `ConnectionString` javadocs

grossws:
Both `MongoClientURI` and `ConnectionString` support `/database.collection` format in uri but it's not in javadoc (previously fixed by #116)


rozza:
Thanks @grossws for the javadoc update!

PR #380, opened 2016-11-28
JAVA-1796 accept hexadecimal string for value of $type in binary data

tomdcsmith:
Fixed jira issue [JAVA-1796](https://jira.mongodb.org/browse/JAVA-1796).

I had the same problem as in this ticket, where the value for $type for $binary data was in a hexadecimal string. This specification can be found [here](https://docs.mongodb.com/v3.2/reference/mongodb-extended-json/#data_binary).

### DIFF
--- a/driver/src/main/com/mongodb/util/JSONCallback.java
+++ b/driver/src/main/com/mongodb/util/JSONCallback.java
@@ -125,7 +125,7 @@ public class JSONCallback extends BasicBSONCallback {
         } else if (b.containsField("$uuid")) {
             o = UUID.fromString((String) b.get("$uuid"));
         } else if (b.containsField("$binary")) {
-            int type = (b.get("$type") instanceof String) ? Integer.decode((String) b.get("$type")) : (Integer) b.get("$type");
+            int type = (b.get("$type") instanceof String) ? Integer.valueOf((String) b.get("$type"), 16) : (Integer) b.get("$type");
             byte[] bytes = DatatypeConverter.parseBase64Binary((String) b.get("$binary"));
             o = new Binary((byte) type, bytes);
         } else if (b.containsField("$undefined") && b.get("$undefined").equals(true)) {
--- a/driver/src/main/com/mongodb/util/JSONCallback.java
+++ b/driver/src/main/com/mongodb/util/JSONCallback.java
@@ -125,12 +125,7 @@ public class JSONCallback extends BasicBSONCallback {
         } else if (b.containsField("$uuid")) {
             o = UUID.fromString((String) b.get("$uuid"));
         } else if (b.containsField("$binary")) {
-            int type;
-            if (b.get("$type") instanceof String) {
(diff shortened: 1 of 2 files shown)


rozza:
Hi @tomdcsmith,

Could you add a regression test to `JSONCallbackTest` and we'll look to squash and merge?

Ross

tomdcsmith:
I've added [two tests](https://github.com/mongodb/mongo-java-driver/pull/380/commits/9b2a5ea706cd601c358d4003c5774b64baedb268) to JSONCallbackTest, one to test the parsing of binary data with the $type value as an integer, and one with the $type value as a hexadecimal string.

driver/src/test/unit/com/mongodb/util/JSONCallbackTest.java
```diff
@@ -139,4 +142,45 @@ public void numberDecimalParsing() {
         Decimal128 number = (Decimal128) JSON.parse(("{\"$numberDecimal\" : \"314E-2\"}"));
         assertEquals(number, Decimal128.parse("314E-2"));
     }
+
+    private Object objectDoneTestHelper(final String jsonString, final String rootName) {
```
rozza: You can simplify the test cases by extending the existing `binaryParsing` test and add a test case similar to the existing one with a hex string eg:

```
Binary parsedBinaryWithHexType = (Binary) JSON.parse(("{ \"$binary\" : \"YWJjZA==\", \"$type\" : \"80\" }"));
assertEquals(128, parsedBinaryWithHexType.getType());
```

driver/src/main/com/mongodb/util/JSONCallback.java
```diff
@@ -125,7 +125,7 @@ public Object objectDone() {
         } else if (b.containsField("$uuid")) {
             o = UUID.fromString((String) b.get("$uuid"));
         } else if (b.containsField("$binary")) {
-            int type = (Integer) b.get("$type");
+            int type = (b.get("$type") instanceof String) ? Integer.decode((String) b.get("$type")) : (Integer) b.get("$type");
```
rozza: This is incorrect because when `$type` is a String it represents a hex string and not an int string.

It needs to be updated to: `Integer.valueOf((String) b.get("$type"), 16)`

driver/src/test/unit/com/mongodb/util/JSONCallbackTest.java
tomdcsmith: I've extended this method now, but the 128 int needed to be cast to a byte:

`assertEquals((byte) 128, parsedBinaryWithHexType.getType());`

driver/src/main/com/mongodb/util/JSONCallback.java
tomdcsmith: I've updated this now.

rozza:
Many thanks @tomdcsmith this has been squashed and merged into master!

PR #382, opened 2016-12-06
Fix BSON decoder inconsistency in java-driver to workaround a bug in MongoDB.

chusse3t:
This is a hotfix to workaround a regression in MongoDB 3.4 (works fine with 3.2), which causes the "rolesInfo" command to return a document with a duplicate "roles" field.

Unfortunately, the first "roles" entry has the data, while the second one is empty. The java driver behaves differently compared to the C based drivers, since Map::put() will overwrite existing entries, thus making the last occurrence matter, while the C based drivers use a linear scan that will always return the first occurrence.

Further, I want to point out that this is NOT a bug in the java driver, but since the broken MongoDB has already been shipped to production, we still need a fix to deal with the broken MongoDBs that are in circulation.

### DIFF
--- a/bson/src/main/org/bson/BsonDocument.java
+++ b/bson/src/main/org/bson/BsonDocument.java
@@ -69,7 +69,10 @@ public class BsonDocument extends BsonValue implements Map<String, BsonValue>, C
      */
     public BsonDocument(final List<BsonElement> bsonElements) {
         for (BsonElement cur : bsonElements) {
-            put(cur.getName(), cur.getValue());
+            final String name = cur.getName();
+            if (!containsKey(name)) {
+                put(name, cur.getValue());
+            }
         }
     }
 
--- a/driver-core/src/test/functional/com/mongodb/operation/RolesInfoOperationsSpecification.groovy
+++ b/driver-core/src/test/functional/com/mongodb/operation/RolesInfoOperationsSpecification.groovy
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2008-2016 MongoDB, Inc.
+ *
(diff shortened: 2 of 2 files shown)


rozza:
Hi @chusse3t,

I can the server ticket for this [SERVER-27271](https://jira.mongodb.org/browse/SERVER-27271) is scheduled for a 3.4.1 release.  

In the meantime I can suggest creating a custom `BsonDocumentCodec` that handles duplicate keys and then use that codec when running your `CommandReadOperation`.

Ross

chusse3t:
Hi Ross,

thanks for the quick response. I just want to underline that we would really appreciate this change being made to the official java driver too, since even though the server is going to be fixed, the broken servers are still out there :). And if it's officially fixed we have one fewer patch to make to the java driver.

Cheers
Chris

jyemin:
Hi Chris,

I do not want to make this change until we have agreement from all supported drivers on how to handle this situation.  This would likely come on the heels of an amendment to the BSON specification that describes the recommended behavior for BSON decoders that encounter duplicate fields.  

Note that a third possibility, which is implemented in the .NET driver, is to treat duplicate fields as an error and throw an exception.

After all, it's just as likely that the "correct" value for the duplicate field is the last occurrence rather than the first.  Or there could be more than two occurrences and the "correct" value is neither the first not the last.  It just so happens that the particular manifestation of a duplicate field in [SERVER-27271](https://jira.mongodb.org/browse/SERVER-27271) places the correct value in the first occurrence of the roles field.

Please also note that a new server ticket has been added to address the general problem on the server side: [SERVER-19091](https://jira.mongodb.org/browse/SERVER-19091)

PR #386, opened 2017-01-21
JAVA-1354: JSON date parsing - support for timezones other than 'Z'

adamwasila:
This is an elaborated version of first fix (which only used 'X' in parsing pattern). Right now it works best on java 1.7+ but has some fallback code to support java 1.6 as well. Quick performance tests shows that on java 1.6 it is around 3 times slower than original version.

### DIFF
--- a/driver/src/main/com/mongodb/util/JSONCallback.java
+++ b/driver/src/main/com/mongodb/util/JSONCallback.java
@@ -41,7 +41,9 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.SimpleTimeZone;
+import java.util.TimeZone;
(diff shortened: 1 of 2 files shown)


jyemin:
As of MongoDB 3.0 mongoexport outputs dates in Zulu time:

     ~$ /mongodb/mongodb-3.0.default/bin/mongoexport -d test -c d
    {"_id":{"$oid":"58c1e8ef28489b68ac9feeb8"},"d":{"$date":"2017-03-09T23:44:47.898Z"}}

Since MongoDB 2.6 (the release prior to 3.0) will be EOLed this fall, I'm not convinced that this issue should be addressed at all.

I'm going to close this PR, but please let us know if you have a use case that really requires this, and we can reconsider.

PR #390, opened 2017-02-08
Correct class level javadoc

jsonking:
Updated class level javadoc for DBCollectionDistinctOptions.java
Previously stated it was for count operations, which applies to a different class: DBCollectionCountOptions.java


rozza:
thanks @jsonking 👍

PR #392, opened 2017-02-16
Quoted strings allowed in BinData constructor

dkublik:


dkublik:
## Issue
Currently **org.bson.json.JsonReader** in unable to read all base64 characters when creating **BinData**
Troublesome characters: **'+', '/', '='**

This makes for example **org.bson.Document.parse(final String json)** fail when reading documents with binary uuids.


## to reproduce:
in **org.bson.json.JsonReaderTest.testBinDataWithNew()**  

replace
`String json = "{ \"a\" : new BinData(3, AQID) }";`

with
`String json = "{ \"a\" : new BinData(3, AQIDBA==) }";`

to get new byte[]{1, 2, 3, 4}


it will fail as **BinDataConstructor** expects only unquotted string as byte value and
**org.bson.json.JsonScanner.scanUnquotedString()**
reads only '$', '_', letters and digits

jyemin:
I changed the commit message to add the issue number, rebased, and pushed.

https://github.com/mongodb/mongo-java-driver/commit/e53fb26fb9733c12f9e0844c2fcf2d0c28185873

PR #393, opened 2017-02-24
Fix javadoc

almostimplemented:
does --> doesn't

I _think_ this is what is actually meant... Apologies if I'm mistaken.


jyemin:
Thanks, you're right.  As I was reviewing I noticed that there's a typo in the same sentence: 'equalss'.  You want to fix that too and then we'll merge?

rozza:
Thanks @almostimplemented !

PR #395, opened 2017-03-01
JAVA-2459: Add support for ConnectionListener and ConnectionPoolListener to MongoClientOptions

aantono:


aantono:
Any update on the status of this PR to be merged?

jyemin:
We're going to have to devote more time to this.  Issues that we need to address before moving forward include:

*Review the events to make sure they are what we want.  For example, the message-related events in ConnectionListener should probably go because they duplicate what's available in the CommandListener interface.
* Coordinate with other drivers to make sure we agree on the event types
* Remove the @Beta flag on the interfaces

jyemin:
Thanks for pushing this along.  We're going to work on this in scope of JAVA-805, making some of the changes that I mentioned above.  Please look for progress on that ticket, which is currently in internal code review.

PR #397, opened 2017-04-07
Added ulimit with command to start mongod when building; removed lega…

jsonking:
When first checking out the project + running the commands to build and test the driver several tests fail with "Too many open files" errors causing the suite to hang and server to crash.

This pull request helps new users get up and running quickly by stating a `ulimit` command that can be used. I also provided the subsequent command to start a local `mongod` using this limit.

There was a broken build status image/link pointing to 10gen which looks old/legacy so I removed it.

Also fixed intellij image showing extra braces due to an extra line break

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,12 +115,17 @@ $ ./gradlew check
 ```
 
 The test suite requires mongod to be running with [`enableTestCommands`](http://docs.mongodb.org/manual/reference/parameters/#param.enableTestCommands), which may be set with the `--setParameter enableTestCommands=1`
-command-line parameter.
+command-line parameter:
+```
+$ mongod --dbpath ./data/db --logpath ./data/mongod.log --port 27017 --logappend --fork --setParameter enableTestCommands=1
+```
 
+If you encounter `"Too many open files"` errors when running the tests then you will need to increase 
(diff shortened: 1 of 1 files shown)


jsonking:
The travis CI build seems to be stuck 'in progress', so I tried closing and re-opening the pull request to re-trigger it. It doesn't seem to have helped :(

rozza:
Hi @jsonking thanks for the PR.  

I don't think we should blanket advise to set ulimits to 1024, however, I like the idea of pointing out that they may cause the test suite to fail.  

So instead could you update to add a paragraph describing the error and how to fix? It would be good if it linked to the manual: https://docs.mongodb.com/manual/reference/ulimit/ 

Thanks again,

Ross

jsonking:
Hi @rozza I have pushed a fresh commit with the change suggested.

The reason I thought it would be ok to set the `ulimit` explicitly is that when running the test suite needs it to be above the default of 256 (on macOS) and when logging in via the shell I get the following: "WARNING: soft rlimits too low. Number of files is 256, should be at least 1000". Agree the single command is cleaner and should work for most people.

Thanks
Jason

jsonking:
Hello,

Is this pull request ok to be merged or would you like further changes?

Thanks
Jason

rozza:
All good - thanks @jsonking 👍

PR #401, opened 2017-05-09
Support Unix domain sockets

joankaradimov:
I found this piece of code, which unfortunately does not work with the official Mongo driver:
http://www.allanbank.com/mongodb-async-driver/userguide/unix_domain_sockets.html

This pull request does a couple of things that should fix the incompatibilities that I observed that prevented `junixsocket` from working.

Unix sockets aside - I think these changes make sense on their own.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,6 +20,7 @@ apply plugin: 'eclipse'
 apply plugin: 'idea'
 
 def configDir = new File(rootDir, 'config')
+ext.jnrUnixsocketVersion = '0.18'
 ext.nettyVersion = '4.1.17.Final'
 ext.snappyVersion = '1.1.4'
 
--- a/driver-core/build.gradle
+++ b/driver-core/build.gradle
@@ -29,6 +29,7 @@ clirr {
 dependencies {
     compile project(':bson')
 
+    compile "com.github.jnr:jnr-unixsocket:$jnrUnixsocketVersion", optional
     compile "io.netty:netty-buffer:$nettyVersion", optional
     compile "io.netty:netty-transport:$nettyVersion", optional
     compile "io.netty:netty-handler:$nettyVersion", optional
--- a/driver-core/src/main/com/mongodb/ServerAddress.java
+++ b/driver-core/src/main/com/mongodb/ServerAddress.java
@@ -17,7 +17,9 @@
 package com.mongodb;
 
 import com.mongodb.annotations.Immutable;
+import jnr.unixsocket.UnixSocketAddress;
 
+import java.io.File;
 import java.io.Serializable;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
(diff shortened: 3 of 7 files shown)


rozza:
Fixed the build upstream and now the tests here pass as expected.

joankaradimov:
I've rebased on the latest master.

I've also added a commit that should allow usage with `Socket`/`SocketAddress` implementations other than `InetSocket`/`InetSocketAddress`. So in theory with some `SocketFactory` magic in the `MongoClientOptions` you should be able to use something like [jnr-unixsocket](https://github.com/jnr/jnr-unixsocket).

joankaradimov:
I rebased and added a new commit. The new commit implements Unix sockets inside the main codebase (as opposed to the possibility of a user-defined `SocketFactory`). The change is less than 30 lines. There's only one change in the public API - the addition of a new constructor to `ServerAddress`.

The previous commits are related to the latest change, but they also make sense on their own (as I've already mentioned).

The implementation uses [jnr-unixsocket](https://github.com/jnr/jnr-unixsocket) which in turn uses [jnr-ffi](https://github.com/jnr/jnr-ffi). The latter is interesting because it is the basis of [JEP 191](http://openjdk.java.net/jeps/191) which is soon to be release in Java 9. [Java Native Runtime](https://github.com/jnr) libraries are cool, because there is no need to manually install native components or to change the environment (which is a drawback of junixsocket).

Here are some notes:
- I've not tested it with clusters. I doubt it's a use case.
- The `jnr-unixsocket` dependency is optional. In case the JAR is missing the connection will ultimately fail with:
    `Caused by: java.lang.ClassNotFoundException: jnr.unixsocket.UnixSocketAddress`
- If it is run on Windows (with `jnr-unixsocket` present in the classpath) it will fail with something like:
    `Timed out after 30000 ms while waiting to connect. Client view of cluster state is {type=UNKNOWN, servers=[{address=\tmp\mongodb-27017.sock:0, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoException: java.lang.UnsatisfiedLinkError: The operation completed successfully.}, caused by {java.lang.UnsatisfiedLinkError: The operation completed successfully.}}]`
- I have some examples I played with, but I wouldn't call them tests. If there is interest I can put together some real tests based on them.

jyemin:
For reference, the associated Jira issue: https://jira.mongodb.org/browse/JAVA-163

jyemin:
Hi @joankaradimov thanks for updating this PR.  Our focus right now is on driver support for MongoDB 3.6, but this is definitely on our radar to review as soon as our time frees up a bit.

joankaradimov:
There were some conflicts with latest master. I've rebased and force pushed the branch.

rozza:
Hi @joankaradimov  thanks for your PR, just to let you know your work formed the base for Unix Domain Socket support. 

I rebased your commits into master d2ff8e12a3352d76a297178f75177d7093295ebe and added MongoConnectionString support in e1680a252ae5122ac77f942eed4625d589d4ec72

Many thanks for contribution.

Ross

PR #403, opened 2017-06-03
Typo : missing quote in toString method's result

astik:
No bug here, jus a typo : a missing quote for the id value in the toString method.


jyemin:
Thanks for catching that.  Say no to typos!

PR #404, opened 2017-06-30
Syntax fix and format

nathanosoares:
Missing parentheses and comma


rozza:
Thanks @nathanosoares for the docs fix.

PR #405, opened 2017-07-20
Add Java Bean support to the PojoCodec

rozza:
JAVA-2555


bson/src/main/org/bson/codecs/pojo/ConventionAnnotationImpl.java
```diff
@@ -31,14 +39,65 @@ public void apply(final ClassModelBuilder<?> classModelBuilder) {
...
+                        creatorExecutable = new CreatorExecutable<T>(clazz, (Constructor<T>) constructor);
+                        break;
+                    }
+                }
+            }
+        }
+
+        if (creatorExecutable == null) {
+            for (Method method : clazz.getDeclaredMethods()) {
+                if (isStatic(method.getModifiers())) {
```
jyemin: It should also be required to return an instance of the declaring class, as per 4.5, paragraph 3.

bson/src/main/org/bson/codecs/pojo/ConventionAnnotationImpl.java
```diff
@@ -31,14 +39,65 @@ public void apply(final ClassModelBuilder<?> classModelBuilder) {
...
+            List<Class<?>> parameterTypes = creatorExecutable.getParameterTypes();
+            if (properties.size() != parameterTypes.size()) {
+                throw creatorExecutable.getError("All parameters must be annotated with a @Property");
+            }
+            for (int i = 0; i < properties.size(); i++) {
+                Property property = properties.get(i);
+                Class<?> parameterType = parameterTypes.get(i);
+                PropertyModelBuilder<?> propertyModelBuilder = classModelBuilder.getProperty(property.value());
+                if (propertyModelBuilder == null) {
+                    throw creatorExecutable.getError(format("Missing Property with the value: '%s'", property.value()));
```
jyemin: In Yasson this isn't actually an error.  See org.eclipse.yasson.customization.JsonbCreatorTest#testCreatorWithoutJavabeanProperty

In other words it's perfectly ok to use a creator with a property defined by one of its arguments that doesn't match up with any property defined in the model.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -118,31 +171,47 @@
         }
     }
 
-    private static <T> FieldModelBuilder<T> getFieldBuilder(final Field field, final Class<T> clazz) {
-        return FieldModel.<T>builder(field);
+    private static TypeData<?> getTypeData(final Method method) {
+        if (isGetter(method.getName())) {
+            return getTypeData(method.getGenericReturnType(), method.getReturnType());
+        } else {
+            if (method.getGenericParameterTypes().length != 1) {
```
rozza: It throws in yasson - but I'm happy to ignore.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -188,6 +255,28 @@ private static TypeParameterMap getTypeParameterMap(final List<String> genericTy
         return value;
     }
 
+    private static final String IS_PREFIX = "is";
+
+    private static final String GET_PREFIX = "get";
+
+    private static final String SET_PREFIX = "set";
+
+    static boolean isSetter(final String methodName) {
```
rozza: Done.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -188,6 +255,28 @@ private static TypeParameterMap getTypeParameterMap(final List<String> genericTy
...
+        return methodName.startsWith(SET_PREFIX);
+    }
+
+    static boolean isGetter(final String methodName) {
```
rozza: Done.

bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
```diff
@@ -0,0 +1,122 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+
+final class CreatorExecutable<T> {
```
jyemin: Acknowledged.  I was reading it as an executable like mongod.exe.

bson/src/main/org/bson/codecs/pojo/PropertyAccessorImpl.java
```diff
@@ -0,0 +1,73 @@
...
+                }
+            }
+        } catch (final Exception e) {
+            throw setError(e);
+        }
+    }
+
+    private CodecConfigurationException getError(final Exception cause) {
+        return new CodecConfigurationException(format("Unable to get value for property '%s' in %s", propertyMetadata.getName(),
+                propertyMetadata.getTypeData().getType()), cause);
```
jyemin: Ack

bson/src/main/org/bson/codecs/pojo/PropertyAccessorImpl.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private CodecConfigurationException setError(final Exception cause) {
+        return new CodecConfigurationException(format("Unable to get value for property '%s' in %s", propertyMetadata.getName(),
```
jyemin: Ack

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -43,51 +47,100 @@
...
         }
-        classModelBuilder.instanceCreatorFactory(new InstanceCreatorFactoryImpl<T>(clazz.getSimpleName(), noArgsConstructor));
+
+        classModelBuilder.instanceCreatorFactory(new InstanceCreatorFactoryImpl<T>(new CreatorExecutable<T>(clazz, noArgsConstructor)));
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <T, S> PropertyMetadata<T> getOrCreateProperty(final String propertyName,
+                                                                  final Map<String, PropertyMetadata<?>> propertyNameMap,
+                                                                  final TypeData<T> typeData,
```
jyemin: I'm not quite seeing that use case.  If the public API is all ObjectId there's no point in storing as a private String field.  I could see this:

    private String id; 
    public String getId() {
        return id;
    }

    public void setId(final String id) {
        this.id = id;
    }

and wanting to serialize id as an ObjectId.  But that's a different case.  Let's discuss.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -16,22 +16,26 @@
 
 package org.bson.codecs.pojo;
 
+import org.bson.codecs.configuration.CodecConfigurationException;
+
+import java.beans.Introspector;
```
jyemin: I know I mentioned this class to you, but on further reflection (no pun intended) let's avoid this dependency on java.beans.Introspector.  In Java 9 this will force applications to pull in the jdk.desktop module, which is going to limit our deployment opportunities.  

http://download.java.net/java/jdk9/docs/api/java/beans/Introspector.html#decapitalize-java.lang.String-

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
jyemin: Similarly, these should not be considered get/set methods
  
     int getfoo();  
     void setfoo(int foo);
     int get();
     void set(int foo);

In other words there has to be at least one character after get/set prefix and it has to be a capital letter (using Character.isUpper(char c)).

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
jyemin: Acknowleded

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
jyemin: Acknowledged

bson/src/main/org/bson/codecs/pojo/annotations/Property.java
```diff
@@ -16,30 +16,32 @@
...
  * <p>Note: Requires the {@link org.bson.codecs.pojo.Conventions#ANNOTATION_CONVENTION}</p>
  *
  * @since 3.5
  * @see org.bson.codecs.pojo.Conventions#ANNOTATION_CONVENTION
  */
 @Documented
-@Target(ElementType.FIELD)
+@Target({ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Property {
```
rozza: No, it was still open / tbc. I like the Bson prefix - so have updated and think it looks reasonable.

jyemin:
Make sure you fix the checkstyle errors before pushing.

rozza:
Done, docs updated, rebased and merged  🎉

PR #406, opened 2017-08-11
Add boolean to automatic - PojoCodecProvider.builder().automatic(true).build()

volyx:


rozza:
Thanks @volyx - merged.  The documentation process isn't automatic so the online documentation will be updated in due course.

PR #408, opened 2017-08-17
detection of slf4j logging: remove check for org.slf4j.impl.StaticLoggerBinder

jpstotz:
When using mongo-java-driver as bundle within an OSGi environment the implemented detection mechanism, if slf4j is available or not, fails and will always return false (because `Class.forName()` does not work on OSGi the way as in non-OSGi Java). 

I don't know a away to fix the detection without having access to certain classes that only exist in an OSGi environment. Therefore I added a possibility to override the detection by setting a system property before using the mongo java driver:

	System.setProperty("org.mongodb.driver.use_slf4", "true");


jyemin:
I'd prefer not to use system properties to control the behavior of the driver.  Most of the ones the driver used to rely on have ended up replaced with an options that's controllable via a public API.

What do you think though about the approach in https://jira.mongodb.org/browse/JAVA-2203 as an alternative? That would make it unnecessary to make this change.

jpstotz:
I checked the propose change of  https://jira.mongodb.org/browse/JAVA-2203 and it indeed solves my problems with the detection of slf4j. I changed my push request accordingly.

driver-core/src/main/com/mongodb/diagnostics/logging/Loggers.java
```diff
@@ -59,11 +59,9 @@ private Loggers() {
...
             Class.forName("org.slf4j.LoggerFactory");
-            // Don't use SLF4J unless a logging implementation has been configured for it
-            Class.forName("org.slf4j.impl.StaticLoggerBinder");
             return true;
         } catch (ClassNotFoundException e) {
             return false;
         }
     }
-}
+}
```
kashike: Missing newline at end of file.

jyemin:
@jpstotz many thanks for the contribution, and in particular the testing in an OSGi environment.

I'd like to see a few other changes included in this PR, including:

- Change `Class.forName` check from `org.slf4j.LoggerFactory` to `org.slf4j.Logger`, as the latter is less likely to be removed in any future SLF4J release.
- Apply the same changes to `bson/src/main/org/bson/diagnostics/Loggers.java`
- Document the change in `docs/reference/content/driver/reference/logging.md` and `docs/reference/content/driver-async/reference/logging.md`

Let me know if you'd like to take this on, or if you'd like one of the maintainers to handle it.

jpstotz:
@jyemin As long as I did not miss anything all of your requested changes are now included in this pull request.

jyemin:
Squashed, rebased, and merged to [master](https://github.com/mongodb/mongo-java-driver/commit/9f0cadd8d7b47e39965d4c19b1aea02abe7f40c3).  Thanks very much @kashike !

PR #410, opened 2017-09-01
Don't lowercase hostname

ChappIO:
With Docker becoming more and more popular and it (and other container services) use case-sensitive hostnames it is a bad idea to lowercase the hostname. It causes it not to work when linking to a docker host in the same network when using the bridge driver.

### DIFF
--- a/driver-core/src/main/com/mongodb/ServerAddress.java
+++ b/driver-core/src/main/com/mongodb/ServerAddress.java
@@ -125,7 +125,7 @@ public class ServerAddress implements Serializable {
                 hostToUse = hostToUse.substring(0, idx).trim();
             }
         }
-        this.host = hostToUse.toLowerCase();
+        this.host = hostToUse;
         this.port = portToUse;
     }
 


ChappIO:
I found an explicit test that checks if the hostname is lowercased. What is the reason for this? Why should the hostname be lowercased?

jyemin:
The driver lower-cases hostnames because names in DNS are case-insensitive.  See https://tools.ietf.org/html/rfc4343.  

Are you having a specific problem on Docker or another container service that this change alleviates?  If so, please provide details as we'd certainly like to be aware of the situation so that it can be addressed across all MongoDB drivers.

rozza:
@ChappIO I'm going to close this ticket, the reasoning for lowercasing has been discussed.

If there are specific interop issues with Docker or other container services we are interested to know more. Please follow up either by a post to [mongodb-user mailinglist](https://groups.google.com/forum/#!forum/mongodb-use) or via a [Java jira ticket](https://jira.mongodb.org/browse/JAVA).

Ross

PR #411, opened 2017-09-06
[docs] Fix Markdown table syntax

fhassak:
Hello,

Table is not displaying properly at : 

http://mongodb.github.io/mongo-java-driver/3.6/upgrading/#compatibility

Farès


jyemin:
Thanks for noticing, and also fixing!

jyemin:
Note that the docs publishing process is currently manual, so the change won't show up immediately.

PR #413, opened 2017-09-14
Support BsonCreator with BsonProperty renaming

visualage:
The existing POJO codec does not work if a property is renamed using @BsonProperty("newName") on the property getter and a @BsonCreator with property @BsonProperty annotations to the parameter.

Also added support of @BsonId annotation to the @BsonCreator's parameter.

To my opinion, the value of @BsonProperty should be mapped to the property name in BSON, regardless where it is annotated. Right now, @BsonProperty value on a @BsonCreator maps to a property getter name. For example, if there is a getter @BsonId getPersonId(), the corresponding constructor parameter annotation should be @BsonId or @BsonProperty("_id) (since '_id' is the property name in BSON), not @BsonProperty("personId") ('personId' does not exist in BSON data). This patch corrects the behavior, but falls back to the old behavior if the new behavior cannot find the right property.

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/ClassModelBuilder.java
+++ b/bson/src/main/org/bson/codecs/pojo/ClassModelBuilder.java
@@ -43,7 +43,7 @@ import static org.bson.codecs.pojo.PojoBuilderHelper.stateNotNull;
  * @see ClassModel
  */
 public class ClassModelBuilder<T> {
-    private static final String ID_PROPERTY_NAME = "_id";
+    static final String ID_PROPERTY_NAME = "_id";
     private final List<PropertyModelBuilder<?>> propertyModelBuilders = new ArrayList<PropertyModelBuilder<?>>();
     private InstanceCreatorFactory<T> instanceCreatorFactory;
     private Class<T> type;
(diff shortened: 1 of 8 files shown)


rozza:
Hi @visualage,

Thanks for the ticket. I've added [JAVA-2599](https://jira.mongodb.org/browse/JAVA-2599) to track this issue.

A property has 3 names:

 * The original name of the property 
    (either the field name or the method name (minus any bean conventions get/set).
  * The write name
  * The read name

This allows for a high level of flexibility over how data is controlled.  Currently, in the context of `@bsonCreator` the `@bsonproperty` value relates to the _actual property name_ and not the write name.

On reflection using the write name would have been a better choice and I like your backwards compatible approach, so thanks again for the patch. We'll look to incorporate into the next release.

In the meantime, you can provide your own `Conventions` to the `PojoCodecProvider` to configure these annotations.

Ross

rozza:
Thanks @visualage after rebasing and some minor changes - your PR was commited to master: b376aa4648c0fa82842713eba56dfa3ab7623e5d

Thank you for the excellent PR and your work on improving the PojoCodec.

PR #416, opened 2017-09-21
Allow discriminators to be used in a collection.

visualage:
Right now, there is no way to serialize/deserialize a collection of abstract class or interface correctly, even though BsonDiscriminator annotation is added to all relevant classes (i.e. both parent class and all sub-classes). See the added unit test.
This patch fixes this problem by using the codec of the actual class of the value during encoding instead of the declared class. The decoding already uses a similar process.

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
+++ b/bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
@@ -67,29 +67,34 @@ final class PojoCodecImpl<T> extends PojoCodec<T> {
         }
     }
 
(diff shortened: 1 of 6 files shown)


rozza:
Thanks @visualage for the PR, it looks good.  

I've added [JAVA-2608](https://jira.mongodb.org/browse/JAVA-2608) to track this.

Ross

rozza:
Thanks @visualage after rebasing and some minor changes - your PR was commited to master: 41f8cf0b333d7882a96a8bfb4b10f6dcace45606

Thank you for the excellent PR and your work on improving the PojoCodec.

PR #419, opened 2017-10-05
[JAVA-2625] use default decimal separator

fhassak:
use locale decimal separator


fhassak:
alternative solution : https://github.com/mongodb/mongo-java-driver/pull/420

PR #424, opened 2017-11-01
Fix for https://jira.mongodb.org/browse/JAVA-2648

bcluap:
Allow a POJO without a setter for a collection to have the collection
populated via the getter and addAll()

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/PropertyAccessorImpl.java
+++ b/bson/src/main/org/bson/codecs/pojo/PropertyAccessorImpl.java
@@ -19,6 +19,7 @@ package org.bson.codecs.pojo;
 import org.bson.codecs.configuration.CodecConfigurationException;
 
 import static java.lang.String.format;
+import java.util.Collection;
 
 final class PropertyAccessorImpl<T> implements PropertyAccessor<T> {
 
@@ -47,6 +48,7 @@ final class PropertyAccessorImpl<T> implements PropertyAccessor<T> {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public <S> void set(final S instance, final T value) {
         try {
             if (propertyMetadata.isDeserializable()) {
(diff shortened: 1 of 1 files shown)


bcluap:
Any chance of this being pulled? Tx

rozza:
Hi @bcluap,

Firstly, thanks for the PR, apologies for the lack of response. We are currently focusing on MongoDB 3.6 features for the upcoming release of the Java driver.  I have some PojoCodec tickets to review in due course and I will review this PR as part of that work.

Ross

bson/src/main/org/bson/codecs/pojo/PropertyAccessorImpl.java
```diff
@@ -55,6 +57,16 @@
                 } else {
                     propertyMetadata.getField().set(instance, value);
                 }
+            } else if (propertyMetadata.getGetter() != null
```
72MiguelGomes: Why not directly call isSerializable(), this way you also have the logic around transient fields...

bcluap: Something like this?:

```java
} else if (propertyMetadata.isSerializable() && value instanceof Collection) {
                // PCB : https://jira.mongodb.org/browse/JAVA-2648
                // If a type is a collection and there is no setter then use the getter and addAll to it from the value
                // Also check that the collection is currently empty. Without this check some test cases fail.
                T collection;
                if (propertyMetadata.getGetter() != null) {
                    collection = (T) propertyMetadata.getGetter().invoke(instance);
                } else {
                    collection = (T) propertyMetadata.getField().get(instance);
                }
                Collection<?> col = ((Collection<?>) collection);
                if (col != null && col.isEmpty()) {
                    col.addAll((Collection) value);
                }
            }
```

rozza:
Closing as support for setting collections / maps via getters has been added via a convention.

PR #428, opened 2017-11-09
Avoid interruption exception in tryAcquire when there is no wait.

visualage:
When using the mongo-java-driver-reactivestream with RxJava2, sometimes the driver throws MongoInterruptedException due to RxJava2 interrupts the thread when disposed. See discussion on the RxJava2 side: https://github.com/ReactiveX/RxJava/issues/5711

RxJava developers recommend the callbacks should not be interrupt sensitive. By changing the tryAcquire(0, TimeUnit) to tryAcquire(), it would not throw InterruptedException.

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
@@ -176,7 +176,9 @@ public class ConcurrentPool<T> implements Pool<T> {
         try {
             if (closed) {
                 return false;
-            } else if (timeout >= 0) {
+            } else if (timeout == 0) {
+                return permits.tryAcquire();
+            } else if (timeout > 0) {
                 return permits.tryAcquire(timeout, timeUnit);
             } else {
                 permits.acquire();


rozza:
Hi @visualage,

Thanks for the ticket. I've added [JAVA-2661]( https://jira.mongodb.org/browse/JAVA-2661) to track this.

Ross

rozza:
@visualage it would be good to understand more about the error and if it is *only* triggered when calling `tryAcquire(0, TimeUnit)` .  Could you provide some more information about triggering this error?

rozza:
Double checking the RxJava2 thread, it looks like ReactiveX/RxJava#5715 fixes ReactiveX/RxJava#5711- can you confirm @visualage ?

72MiguelGomes:
@rozza Not sure what error @visualage is trying to solve but it seems total sense if the timeout is 0 try to lock or return immediately. And it is also part of a function documentation:

> timeout  negative - forever 0        - return immediately no matter what positive ms to wait

In this case I would say that you should add this fix

rozza:
@72MiguelGomes the issue with this fix is `tryAcquire()` does not respect other waiters on the permit - its greedy. Whereas `tryAcquire(0, TimeUnit)` does respect other waiters.

Having not heard back from @visualage and the belief that there is a fix in RxJava, I think this can be closed. However, please follow up if that is not the case.

visualage:
Sorry for the late reply. The rxjava fix would solve my problem, so this is not needed.

PR #429, opened 2017-11-09
Update quick-start.md

DerTomm:
Corrected "Get A Single Document That Matches a Filter" java example (inserted missing Filters class before ".eq(...))


fhassak:
Hello,

This is not necessary because the method `eq` of the Filters class is imported statically by the following import statement as specified in Prerequisites paragraph:

``import static com.mongodb.client.model.Filters.*;``

jyemin:
I agree with @fhassak, so I'm going to close this one.

Thanks though for the contribution and feel free to fix any other docs issues that you come across.

PR #430, opened 2017-11-09
JAVA-2636: Move logic from InternalStreamConnection to CommandMessage for getting the BsonDocument to use for command monitoring.

jyemin:
Take advantage of the observation that a command with a payload is likely to be small (minus the payload) so it's a reasonable approach to simple convert the BSON back into a BsonDocument and then add the payload as a BsonArray initialized with a List<ByteBufBsonDocument> from the OP_MSG payload.


bson/src/main/org/bson/BsonDocument.java
```diff
@@ -771,6 +771,17 @@ public BsonDocument append(final String key, final BsonValue value) {
...
     }
 
+    /**
+     * Gets the first key in the document.
+     *
+     * @return the first key in the document
+     * @throws java.util.NoSuchElementException if the document is empty
+     * @since 3.6
+     */
+    public String getFirstKey() {
```
jyemin: Needed so that InternalStreamConnection can just call it without having to know what subclass of BsonDocument it has.

bson/src/main/org/bson/RawBsonDocument.java
```diff
@@ -244,6 +245,21 @@ public int size() {
     }
 
     @Override
+    public String getFirstKey() {
```
jyemin: And might as well optimize it here rather than force creation of a keySet.

driver-core/src/main/com/mongodb/connection/ByteBufBsonDocument.java
```diff
@@ -58,7 +58,17 @@
     }
 
     static ByteBufBsonDocument createOne(final ByteBufferBsonOutput bsonOutput, final int startPosition) {
-        return create(bsonOutput, startPosition).get(0);
+        List<ByteBuf> duplicateByteBuffers = bsonOutput.getByteBuffers();
```
jyemin: This change is necessary because otherwise calling createOne on a document within an OP_MSG that is followed by a payload will fail, as the create method thinks it's just a bunch of contiguous documents and doesn't know anything about the payload structure.  So this is just a copy of createOne but without the loop.  

I didn't see a straightforward way to refactor for code re-use, but I'll take another look...

driver-core/src/main/com/mongodb/connection/ByteBufBsonDocument.java
```diff
@@ -130,6 +140,10 @@ public BsonDocument clone() {
         return new RawBsonDocument(clonedBytes);
     }
 
+    int getSizeInBytes() {
```
jyemin: Used by CommandMessage to find the start of the payload

driver-core/src/main/com/mongodb/connection/CommandMessage.java
```diff
@@ -68,6 +70,31 @@
...
 
+    BsonDocument getCommandDocument(final ByteBufferBsonOutput bsonOutput) {
+        ByteBufBsonDocument byteBufBsonDocument = ByteBufBsonDocument.createOne(bsonOutput,
+                getEncodingMetadata().getFirstDocumentPosition());
+        BsonDocument commandBsonDocument;
+
+        if (useOpMsg() && containsPayload()) {
+            commandBsonDocument = byteBufBsonDocument.toBsonDocument();
+
+            int payloadStartPosition = getEncodingMetadata().getFirstDocumentPosition()
```
jyemin: Not the prettiest but it gets the job done.

bson/src/main/org/bson/RawBsonDocument.java
rozza: Ack

bson/src/main/org/bson/BsonDocument.java
rozza: Ack

driver-core/src/main/com/mongodb/connection/ByteBufBsonDocument.java
rozza: Seems simple enough.

driver-core/src/main/com/mongodb/connection/CommandMessage.java
```diff
@@ -68,6 +70,31 @@
...
+                    + byteBufBsonDocument.getSizeInBytes()
+                    + 1 // payload type
+                    + 4 // payload size
+                    + payload.getPayloadName().getBytes(Charset.forName("UTF-8")).length + 1;  // null-terminated UTF-8 payload name
+            commandBsonDocument.append(payload.getPayloadName(),
+                    new BsonArray(ByteBufBsonDocument.create(bsonOutput, payloadStartPosition)));
```
rozza: As `ByteBufBsonDocument.create` and `ByteBufBsonDocument.createOne` are poorly named as create actually creates a list. This made me double take at whats added to the BsonArray. 

Can we take this opportunity to rename?

jyemin: createMany or createList?

rozza: createList seems most inkeeping with existing API names

PR #432, opened 2017-12-15
JAVA-2703

jyemin:
Ensure that the ElementExtendingBsonWriter appends the extra elements when a reader is piped into the writer at the top level.

The fix is largely restoring the original code which we mistakenly thought was no longer needed.

The bug wasn't caught because of a poorly written test, which was also hiding a bug in the original implementation.

Patch build: https://evergreen.mongodb.com/version/5a34454ce3c3316221004284

### DIFF
--- a/bson/src/main/org/bson/AbstractBsonWriter.java
+++ b/bson/src/main/org/bson/AbstractBsonWriter.java
@@ -21,6 +21,7 @@ import org.bson.types.ObjectId;
 
 import java.io.Closeable;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.Stack;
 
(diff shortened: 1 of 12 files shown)


bson/src/main/org/bson/AbstractBsonWriter.java
```diff
@@ -752,36 +753,58 @@ public void close() {
...
+    }
+
+    /**
+     * Reads a single document from the given BsonReader and writes it to this, appending the given extra elements to the document.
+     *
+     * @param reader the source of the document
+     * @param extraElements the extra elements to append to the document
+     * @since 3.6
+     */
+    public void pipe(final BsonReader reader, final List<BsonElement> extraElements) {
```
jyemin: Unfortunate introduction of a new public method in a patch release.  I don't see a way around it.  Unlike the original commit I didn't introduce this pipe method into the BsonWriter interface.  That would be a bridge too far...

bson/src/test/unit/org/bson/BsonHelper.java
```diff
@@ -85,11 +90,16 @@
     }
 
     public static List<BsonValue> valuesOfEveryType() {
-        return BSON_VALUES;
+        return getBsonValues();
     }
 
     public static BsonDocument documentWithValuesOfEveryType() {
-        return BSON_DOCUMENT;
+        BsonDocument document = new BsonDocument();
```
jyemin: And here's the root cause of the bad test: this method returns a singleton BsonDocument, which the test then goes ahead and mutates.  It wasn't caught because it happened to be the last test that used this method.

driver-core/src/main/com/mongodb/connection/ElementExtendingBsonWriter.java
```diff
@@ -26,17 +27,25 @@
...
         if (getCurrentLevel() == 0) {
-            writeElements(getBsonWriter(), extraElements);
+            writeElements(getBsonBinaryWriter(), extraElements);
         }
         super.writeEndDocument();
     }
 
+    @Override
+    public void pipe(final BsonReader reader) {
+        if (getCurrentLevel() == -1) {
```
jyemin: There was actually a bug in the original version, which compared the current level to 0 instead of -1.  It wasn't caught then either because of the test bug.

driver-core/src/main/com/mongodb/connection/LevelCountingBsonWriter.java
```diff
@@ -29,256 +30,256 @@
 
 
 abstract class LevelCountingBsonWriter implements BsonWriter {
-    private final BsonWriter bsonWriter;
+    private final BsonBinaryWriter bsonBinaryWriter;
```
jyemin: To avoid introducing the overloaded pipe method to the BsonWriter interface, changed this to depend on BsonBinaryWriter.

driver-core/src/test/unit/com/mongodb/connection/ElementExtendingBsonWriterSpecification.groovy
```diff
@@ -72,18 +75,18 @@ class ElementExtendingBsonWriterSpecification extends Specification {
         for (def cur : extraElements) {
             expectedDocument.put(cur.name, cur.value)
```
jyemin: And here's where the test was mutating what used to be a shared global singleton.

PR #433, opened 2017-12-19
Reduce memory usage of QueryResult

vinssor:
QueryResult uses a list to store items, which results in excessive memory usage and unnecessary CPU load.

Indeed, the MongoIterable API is streaming oriented with the functions forEach and iterator, it does not require the use of a list in the background.

In addition, this prevents the use of reusable objects in a ThreadLocal to limit the creation of these and further improve the memory footprint.


rozza:
@vinssor this doesn't appear to have any commits by you in.  Looks like a PR wanting to merge the 3.6.x branch into master?

Please check and resubmit.

PR #437, opened 2017-12-27
Replacing the List data structure with a Collection where it makes sense

ricardoespsanto:


jyemin:
Hi Ricardo, as I mentioned on JAVA-2718 this change would break binary compatibility.

jyemin:
Closing this, as the change would break binary compatibility.  We can consider for the next major release (4.0).

PR #438, opened 2018-01-04
upgrade to mongoDB 3.6.1 for test suite

fhassak:


jyemin:
Good catch, @fhassak.  We generally use our own CI server and not Travis, but good to have this updated so that pull requests use the latest MongoDB server on Travis.

PR #440, opened 2018-01-09
Update ErrorHandlingResultCallback#onResult log level

thiaguten:
When the log appender to the "org.mongodb" package is configured with ERROR level.

```xml
<logger name="org.mongodb" level="ERROR"/>
```

And some exception is thrown inside the callback lambda expression body.

```java
...
mongoCollection.insertOne(document, (result, error) -> {
        if (null == error) {
          // do something that throws some runtime exception.
          throw new RuntimeException("some exception!!!");
        }
      });
```

Nothing is written to the log, shadowing the error that occurred. Because internally the driver writes as warning and not error. ErrorHandlingResultCallback#onResult

```java
public void onResult(final T result, final Throwable t) {
    try {
        wrapped.onResult(result, t);
    } catch (Throwable e) {
        logger.warn("Callback onResult call produced an error", e);
    }
}
```

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/async/ErrorHandlingResultCallback.java
+++ b/driver-core/src/main/com/mongodb/internal/async/ErrorHandlingResultCallback.java
@@ -48,7 +48,7 @@ public class ErrorHandlingResultCallback<T> implements SingleResultCallback<T> {
         try {
             wrapped.onResult(result, t);
         } catch (Throwable e) {
-            logger.warn("Callback onResult call produced an error", e);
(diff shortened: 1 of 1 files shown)


rozza:
Hi @thiaguten,

Thanks for the ticket, I agree that error is a better logging level for exceptions caused by callbacks. 

I've rebased and linked the commit to the ticket [JAVA-2728](https://jira.mongodb.org/browse/JAVA-2728).  f4f4977 is now in master and will be released as part of the 3.7.0 release.

Thanks again,

Ross

PR #441, opened 2018-01-16
JAVA-2735 : Force locale to Locale.US when formatting decimal number

fhassak:


fhassak:
please wait before merge, i will propose an alternative solution

PR #445, opened 2018-01-24
JAVA-2748 : Create a ServerAddress with default host

fhassak:


rozza:
Hi @fhassak,

Thanks for the pull request, can you add your motivation for the change to the description?  What benefit do you see to adding to the API?

Ross

fhassak:
To avoid mistakes and search for default host 

like here https://github.com/apache/camel/commit/c2d613564b673cdeaee7853d25488171d9a22b03

fhassak:
Finally after careful reflection, this constructor does not make sense. 
It is better to use the constructor with explicit values ​​for the host and port. 
It does not make sense to only have the port.
I suggest closing this PR.

rozza:
Ok thanks @fhassak

PR #446, opened 2018-01-26
Fix for NPE in POJO codec when a POJO has a nested collection property and one of the nested values is null

robrua:
Currently if the top-level collection property is null it will be handled gracefully, but if a collection of collections contains a null element it currently causes an NPE on the inner CollectionCodec.encode call while attempting to loop through the null collection's values.


rozza:
Thanks @robrua,

I reworded your commit (bb07462) to link to the jira issue: [JAVA-2756](https://jira.mongodb.org/browse/JAVA-2756).  When adding a test case I realised we had to handle decoding and I noticed Map values had the same issue (e38c3f7). 

Thanks again,

Ross

PR #447, opened 2018-01-31
Support username containing ',' comma

brendanJsonar:
When a username contains ',' comma character, it raises authentication exception. According to [RFC 5801](https://tools.ietf.org/html/rfc5801) , comma should be replace by =2C.

> The server MUST replace any "," (comma) in the string with "=2C".
The server MUST replace any "=" (equals) in the string with "=3D".

EDIT: change the reference to [RFC 5802](https://tools.ietf.org/html/rfc5802)
>The characters ',' or '=' in usernames are sent as '=2C' and
         '=3D' respectively.  If the server receives a username that
         contains '=' not followed by either '2C' or '3D', then the
         server MUST fail the authentication.

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/ScramSha1Authenticator.java
+++ b/driver-core/src/main/com/mongodb/connection/ScramSha1Authenticator.java
@@ -281,7 +281,7 @@ class ScramSha1Authenticator extends SaslAuthenticator {
         }
 
         private String prepUserName(final String userName) {
-            return userName.replace("=", "=3D").replace(",", "=2D");
+            return userName.replace("=", "=3D").replace(",", "=2C");
         }
 
         private byte[] xor(final byte[] a, final byte[] b) {


jyemin:
I'm not clear what your intention is here.  The RFC that you reference is for the GSSAPI SASL mechanism, but the class you modified is for SCRAM-SHA-1 SASL mechanism.

Also, the part of the GSSAPI RFC you quoted seems to apply to servers, while the driver is used on clients.

jyemin:
Please open a ticket at https://jira.mongodb.org/projects/JAVA so we can gather more details before considering a pull request.  I'd like to see some sample code and the full stack trace.

Thanks,
Jeff

brendanJsonar:
As referring to the mongo implementation, it uses =2C instead of =2D to encode/decode comma.
[mongo/src/mongo/db/auth/sasl_scram_server_conversation.cpp](https://github.com/mongodb/mongo/blob/e1f98f1fc5203a8b52a9c254e00cfdb2b6fdc9be/src/mongo/db/auth/sasl_scram_server_conversation.cpp#L77)
[mongo/src/mongo/client/sasl_scram_client_conversation.cpp](https://github.com/mongodb/mongo/blob/e1f98f1fc5203a8b52a9c254e00cfdb2b6fdc9be/src/mongo/client/sasl_scram_client_conversation.cpp#L72)

jyemin:
OK, I see.  From what I now understand, https://tools.ietf.org/html/rfc5802 is the relevant spec, which says:

> The characters ',' or '=' in usernames are sent as '=2C' and '=3D' respectively.  If the server receives a username that contains '=' not followed by either '2C' or '3D', then the server MUST fail the authentication.

Same idea, though.  I've opened https://jira.mongodb.org/browse/JAVA-2763 to track this bug.

This PR will ultimately need some unit tests to verify it.  Did you want to take that on as well or leave it to the maintainers?

Regards,
Jeff

jyemin:
I've dealt with the unit test, so don't worry about it.  I'll push it along with the commit in this PR.

Thanks again,
Jeff

brendanJsonar:
Thanks a lot for such a quick response.

jyemin:
Pushed to master and 3.6.x branch.  See https://jira.mongodb.org/browse/JAVA-2763 for links to the commits.

Thanks Brendan!

PR #454, opened 2018-02-16
JAVA-2641

jyemin:
https://jira.mongodb.org/browse/JAVA-2641

This isn't the whole API.  If this looks good I'll do gridfs, apply the corresponding annotations to driver-async CRUD API, then move down into driver-core and bson

### DIFF
--- a/driver-legacy/src/main/com/mongodb/DBCursor.java
+++ b/driver-legacy/src/main/com/mongodb/DBCursor.java
@@ -582,8 +582,12 @@ public class DBCursor implements Cursor, Iterable<DBObject> {
 
     @Override
     public long getCursorId() {
-        if (cursor != null && cursor.getServerCursor() != null) {
-            return cursor.getServerCursor().getId();
+        if (cursor != null) {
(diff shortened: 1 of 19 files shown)


driver-legacy/src/main/com/mongodb/DBCursor.java
```diff
@@ -582,8 +582,12 @@ public DBCursor skip(final int numberOfElements) {
 
     @Override
     public long getCursorId() {
-        if (cursor != null && cursor.getServerCursor() != null) {
-            return cursor.getServerCursor().getId();
+        if (cursor != null) {
```
jyemin: SpotBugs founds this based on the annotations

driver-legacy/src/main/com/mongodb/MongoCursorAdapter.java
```diff
@@ -28,7 +28,11 @@
 
     @Override
     public long getCursorId() {
-        return cursor.getServerCursor().getId();
+        ServerCursor serverCursor = cursor.getServerCursor();
```
jyemin: SpotBugs found this one too.

PR #455, opened 2018-02-18
JAVA-2786 : Upgrade Gradle to 4.5.1

fhassak:
I do this command to upgrade : https://gradle.org/install/#with-the-gradle-wrapper

$ ./gradlew wrapper --gradle-version=4.5.1 --distribution-type=bin


rozza:
Thanks @fhassak - we have wrapper task in `build.gradle` that will also update.

PR #456, opened 2018-02-19
JAVA-2523: Remove use of methods deprecated in Java 9

fhassak:
There is 1 warning i haven't fixed when i build with Oracle JDK 9 : 

D:\github\mongo-java-driver\driver-legacy\src\main\com\mongodb\DBCursor.java:980: warning: [deprecation] finalize() in Object has been deprecated
        protected void finalize() {
                       ^
1 warning

Use close() method...


rozza:
Thanks @fhassak  - as DBCursor is already part of a deprecated API we will review in the future.

PR #459, opened 2018-03-02
Simplification

ashu01:
1. Removed redundant initializer
2. Simplified multiple if...else... to simple expression


rozza:
Hi @ashu01 

Thank you for the PR - currently equals methods are automatically generated in a readable manner.

Although, there is a simplification in terms of the lines of code, merging all the steps into a single line does increase the code complexity and make the code less readable. This is especially important when users wish to debug / add breakpoints to code.

In this case readability trumps lines of code and for that reason I'm inclined to close this PR. If you disagree with the above, then by all means feel free to explain your motivations for the PR in a comment.

Ross

PR #460, opened 2018-03-04
add more useful error messages for invalid field names

mjrb:
This fixes https://jira.mongodb.org/browse/JAVA-2114. when you do an update and forget to put something like $set or $inc in the query or have an other wise malformed update query the error message wasn't very helpful even if it was a silly error. this pull request checks which field name validator failed and gives an appropriate error message.


imthereddwarf:
IMHO this is incorrect validation. In the mongo client you can update without needing $set, db.collection.update( { _id: key}, { _id: key, field: value } ) works fine, but fails in the java driver with the "Invalid BSON field name _id"  exception.  The same error occurs with upsert=true and having to use $set with an upsert makes no sense. So the correct fix is to remove this validation altogether.

jyemin:
Closing due to inactivity.

PR #466, opened 2018-04-11
Fix incorrect use of async driver in code sample creating CodecRegistry

mikeligthart:
This commit changes MongoClient to MongoClients to call getgetDefaultCodecRegistry() and adds value 'true' to automatic().


jyemin:
Rebasing and merging.  Thanks for the contribution!

PR #468, opened 2018-04-26
JAVA-2724 : Remove %n in String.format call in DocTaglet

fhassak:
The `<dd>` element is block level element : https://www.w3.org/TR/html401/struct/lists.html#edef-DD
So, we don't need `\n` or `%`, or `<br/>`  : 

With `%n`:
- http://mongodb.github.io/mongo-java-driver/3.7/javadoc/?com/mongodb/client/model/Accumulators.html 

Without  `%n`:
image


jyemin:
Great find.  Thank you!

PR #474, opened 2018-06-02
Reference the MongoDB embedded v1 ABI symbols

edaniels:
@jyemin @rozza


rozza:
Thanks @edaniels merged into master

PR #477, opened 2018-06-25
Fix getDefaultCodecRegistry() example

salomvary:
MongoClient.getDefaultCodecRegistry() no longer seems to exist


rozza:
Thanks @salomvary !

PR #478, opened 2018-06-26
specifies compatibility with java 9

fhassak:
Hello,
I think all versions since 3.0 are compatible with Java 9 classpath but only versions since version 3.7 are compatible with Java 9 module path ?


rozza:
8+ might cover it.

fhassak:
And java 11 ?

jyemin:
We're currently not able to build the driver with Java 11 due to Gradle incompatibilities.  See https://jira.mongodb.org/browse/JAVA-2991

rozza:
Hi all,

I'm going to close this as the java version graph was a cause of confusion and removed in the latest release (61282f4). See: http://mongodb.github.io/mongo-java-driver/3.9/upgrading/

Ross

PR #480, opened 2018-08-07
removing . and $ field name validation

metanomi:
When using Java driver for Mongo 3.6+ dots and dollar signs should be allowed in document field names.

https://jira.mongodb.org/browse/JAVA-2810


rozza:
JAVA-2810 is marked as won't fix at this time until all drivers have been specc'd for support.

If field level validation is causing an issue then using a RawBsonDocument is a known work around.

PR #482, opened 2018-08-29
typo in package name

brijeshprasad89:
Looks like there is minor spelling mistake in package name.


jyemin:
Thanks for catching this!

Regards,
Jeff

PR #483, opened 2018-08-30
In WriteCommandResult, the 'nModified' could be either Long or Integer

ToniA:
I have a GRAILS project based on the GRAILS 2.3.11 framework. This includes mongo-java-driver-2.14.2, and we can't easily upgrade to 3.x (where this issue seems to be fixed).

I'm build a proof-of-concept using Azure CosmosDB with MongoDB 3.2.0 API, and ran into a problem, 'java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer' at getModifiedCount() in WriteResultCommandHelper.java.

This pull request fixes this particular problem. Please release java-mongo-driver-2.14.4 with this PR.

### DIFF
--- a/src/main/com/mongodb/WriteCommandResultHelper.java
+++ b/src/main/com/mongodb/WriteCommandResultHelper.java
@@ -88,7 +88,14 @@ final class WriteCommandResultHelper {
     }
 
     private static Integer getModifiedCount(final WriteRequest.Type type, final CommandResult commandResult) {
-        Integer modifiedCount =  (Integer) commandResult.get("nModified");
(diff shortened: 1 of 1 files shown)


rozza:
Hi @ToniA,

Thanks for the ticket, there are no planned releases for the 2.14 version of the driver, the last release was over two years ago. Please look to migrating over to the 3.x series of the MongoDB Java driver. As this not only contains bugfixes but also contains many improvements to the API as well as access to many of MongoDB's new features.

Because there are no planned 2.14.x releases and as this is an issue outside MongoDB, I'm closing this ticket as won't fix.

All the best,

Ross

PR #484, opened 2018-09-05
JAVA-2961 ObjectId generation with number of bytes > 12 should throw expection

umang-jalan:
As per documentation of getting ObjectId from bytes, IllegalArgumentException should be thrown if bytes is null or bytes.length!= 12 . But presently for bytes.length > 12, it returns ObjectId using first 12 bytes.

https://jira.mongodb.org/browse/JAVA-2961


jyemin:
@umang-jalan Our apologies for losing track of this PR and doing our own implementation of this issue.  See https://github.com/mongodb/mongo-java-driver/commit/468bcb95a3d1904c2033a52453476b56b94d9bff for what got committed to the repository.

PR #485, opened 2018-09-05
BsonReader and BsonWriter are interfaces

arunkjn:
In driver 3.8 these are interfaces


rozza:
Thanks for the PR @arunkjn

PR #486, opened 2018-09-07
add support to interface default method

cycchow:
Add support to interface default method from java 8


rozza:
Hi @cycchow,

Do you have an example or test case that uses a default method?

Ross

rozza:
Thanks @cycchow - support has been added in the forthcoming 4.0 release.

PR #488, opened 2018-09-12
Added Connection string with user and password

wrey75:
For entreprise as well as production reasons, having the user and the password in the connection URI is not convenient. Having the connection string to the database can be very useful but we have to hide the user and password. This PR is quite simple and do not change compatibility. If a user is provided, it is used in replacement of the one provided in the URI string which makes sense and do not modify a lot the code.
You can write tests providing a null userName to check the compatibility with the old version.

### DIFF
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -314,7 +314,7 @@ public class ConnectionString {
         String userInfo;
         String hostIdentifier;
         String finalUserName = userName;
-        char finalPassword[] = password;
+        char[] finalPassword = password;
         idx = userAndHostInformation.lastIndexOf("@");
         if (idx > 0) {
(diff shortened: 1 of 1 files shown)


jyemin:
You can already do something similar with the existing API:

        MongoClientSettings settings = MongoClientSettings.builder()
                .credential(MongoCredential.createCredential(userName, password))
                .applyConnectionString(new ConnectionString("mongodb://localhost,localhost:27018"))
                .build();
        MongoClient client = MongoClients.create(settings);

Is that sufficient for your use case?

wrey75:
This is perfect. You should add this example in the API documentation or in the [documentation](http://mongodb.github.io/mongo-java-driver/3.8/driver-async/tutorials/connect-to-mongodb/). It is not very clear that applying the connection string without credentials keeps the previous ones.

jyemin:
Good idea.  Added https://jira.mongodb.org/browse/JAVA-2975 to track the request.  Feel free to open a pull request to address it, as our documentation is stored in the repository.

PR #493, opened 2018-10-18
misspelled

phongtn:
misspelled in describe Block<T> method


docs/reference/content/driver-async/getting-started/quick-start.md
```diff
@@ -83,7 +83,7 @@ SingleResultCallback<Void> callbackWhenFinished = new SingleResultCallback<Void>
 
 ```
 
-- The following `Block` code which the exexamples will use used to print the results of the find operations:
+- The following `Block` code which the examples will use used to print the results of the find operations:
```
jyemin: Please also change "will use used" to just "will use".  As is it's not grammatically correct.

PR #498, opened 2018-12-04
JAVA-2921: causal consistency example for Java driver

jstewart-mongo:
Add Java example for causal consistency.


jstewart-mongo:
(issued PR by accident)

PR #502, opened 2019-01-15
JAVA-3140: set English Locale as default date format symbols

fhassak:


gradle.properties
```diff
@@ -15,3 +15,4 @@
 #
 
 org.gradle.daemon=true
+org.gradle.jvmargs=-Duser.country=US -Duser.language=en
```
jyemin: Please append a newline.

fhassak: done.

PR #503, opened 2019-02-16
[JAVA-2720] support using Enum in queries through PojoCodec

AarjavP:
https://jira.mongodb.org/browse/JAVA-2720
We need this, I am not entirely sure of the implication for existing code. To me it seems like a simple fix, and essential to claim that enums are fully supported.

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/EnumCodec.java
+++ b/bson/src/main/org/bson/codecs/pojo/EnumCodec.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


rozza:
Hi @AarjavP, thanks for the PR. I think that there are some implications for the existing code here.

The simplest fix would be to add a `EnumCodec` to the codec registry that you want to use. The best fix would be to use the Codec that is specifically assigned to: `myPojo.mySpecialEnumProperty`. It maybe if users want to specialize then they will have to have their own enum codec higher up than the codec registry and we fall through to the base `EnumCodec`, considerations also have to be made to ensure any changes don't break existing users.

jyemin:
As part of the fix for JAVA-4440, there is now a `EnumCodec<T extends Enum<T>> implements Codec<T>` in the default codec registry, so it's no longer restricted to being used as part of POJO encoding/decoding, so I think we can close out this PR.  Sorry for the long delay on it.

PR #504, opened 2019-02-16
Fix typo in enum code example

AarjavP:
Currently, `enum Membership` is declared, and used as a constructor argument, but the field is of type `Member`.
Just changed the field to `Membership` to make it consistent.


rozza:
Thanks @AarjavP

PR #509, opened 2019-04-25
read & write on input/output stream. or the socket timeout will not t…

Nov11:
read & write on input/output stream. or the socket timeout will not take effect, as socket timeout is not working if reading / writing on a socketchannel.

mainly adopt from SocketStream.
bytebuf is not fliped in SocketStream, so not here in socketchannelstream neither.

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/SocketChannelStream.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/SocketChannelStream.java
@@ -28,9 +28,10 @@ import com.mongodb.connection.Stream;
 import org.bson.ByteBuf;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.net.SocketTimeoutException;
-import java.nio.ByteBuffer;
 import java.nio.channels.SocketChannel;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


rozza:
Hi @Nov11, thanks for the PR.  

Couldyou to provide some more information / motivation for the change.  I haven't had reports of this being an issue. so anything information detailing why socket timeout isn't applied when using `SocketChannel#write`/`SocketChannel#read` would be great.

Ross

Nov11:
Motivation : I came across a situation that the timeout doesn't work. It actually blocked for like 50 seconds or so, without one socket timeout exception thrown.  Thread pool filled up due to slow queries and bad things happened...... 

sotimeout effects input/ouput stream only : https://docs.oracle.com/javase/7/docs/api/java/net/Socket.html#setSoTimeout(int)
sotimeout does not work on socketchannel and this behavior can always be reproduced:
https://stackoverflow.com/questions/2866557/timeout-for-socketchannel-doesnt-work

Nov11:
屏幕快照 2019-04-16 下午1 44 04
this calling stack is how I locate that socketchannelstream is relative to my problem

rozza:
@Nov11 many thanks for the quick replies. 

I've opened [JAVA-3275](https://jira.mongodb.org/projects/JAVA/issues/JAVA-3275) to track this.

Ross

jyemin:
Hi @Nov11,

I fixed this in a different way: by removing SocketChannelStream entirely.

https://github.com/mongodb/mongo-java-driver/commit/c71a6b09daf43edb80210d4edc95d36f107e163e

Thanks for bringing this to our attention.

PR #512, opened 2019-05-28
Cloudp 42881

navenoxin:
This exposes the new internal `showMigrationEvents` option that may be passed to change streams. Setting this option to true will make change streams show insert, update, and delete events generated by chunk migrations. This option may only be used for a change stream opened against a mongod.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ def javaCodeCheckedProjects = subprojects.findAll { !['util', 'mongo-java-driver
 configure(coreProjects) {
     evaluationDependsOn(':util')
     group = 'org.mongodb'
-    version = '3.11.0-SNAPSHOT'
+    version = '3.11.0-mongot'
 
     repositories {
         google()
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ def javaCodeCheckedProjects = subprojects.findAll { !['util', 'mongo-java-driver
 configure(coreProjects) {
     evaluationDependsOn(':util')
     group = 'org.mongodb'
-    version = '3.11.0-mongot'
+    version = '3.11.0-mongot-SNAPSHOT'
 
     repositories {
         google()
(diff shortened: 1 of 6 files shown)


driver-sync/src/main/com/mongodb/client/ChangeStreamIterable.java
```diff
@@ -120,4 +120,18 @@
...
+    /**
+     * <p>This option is NOT part of the public API.  It may change at any time without notification.</p>
+     *
+     * <p>Configures this change stream to show chunk migration CRUD events.</p>
+     * <p>This option is only allowed to be true when passed to a change stream opened directly with a shard.</p>
+     * <p>Using this option with a change stream opened against a mongos is an error.</p>
+     *
+     * @param showMigrationEvents the showMigrationEvents boolean
+     * @return this
+     * @since never
```
jyemin: You can just remove the `@since` annotation

driver-sync/src/main/com/mongodb/client/internal/ChangeStreamIterableImpl.java
```diff
@@ -98,6 +99,12 @@
         return this;
     }
 
+    @Override
+    public ChangeStreamIterable<TResult> showMigrationEvents(final boolean showMigrationEvents) {
+        this.showMigrationEvents = notNull("showMigrationEvents", showMigrationEvents);
```
jyemin: Remove this line.  Doesn't make sense for a primitive value.  What's happening here is that the primitive boolean is being auto-boxed into a `Boolean`.

build.gradle
```diff
@@ -54,7 +54,7 @@ def javaCodeCheckedProjects = subprojects.findAll { !['util', 'mongo-java-driver
 configure(coreProjects) {
     evaluationDependsOn(':util')
     group = 'org.mongodb'
-    version = '3.11.0-SNAPSHOT'
+    version = '3.11.0-mongot'
```
jyemin: I'm not sure how many dashes are allowed in Maven versions, but you probably want something more like:

     3.11.0-mongot-SNAPSHOT

navenoxin: Seems reasonable, the second dash seems OK as part of the version qualifier per https://cwiki.apache.org/confluence/display/MAVENOLD/Dependency+Mediation+and+Conflict+Resolution#DependencyMediationandConflictResolution-DependencyVersionRanges and . It did not give our build issues.

PR #515, opened 2019-06-18
JAVA-3328: Excessive memory allocation due to unwrapping of updates in bulk ops

astral303:
Change removes check and call to isEmpty() which causes an unnecessary unwrapping.

IT has been relaxed to allow for server-side failure for this case.


jyemin:
@astral303 we don't want to relax the check so ended up going with a BsonWriter decorator approach that you tried initially.  There is a problem with pipe() as you recognized, but it's rare that pipe() will be used in this context so better to check in most cases and miss the ones where pipe() is used than not to check at all.

PR #517, opened 2019-07-13
FLE spec test update

kevinAlbs:


driver-sync/src/main/com/mongodb/client/internal/CommandMarker.java
```diff
@@ -76,6 +76,7 @@ RawBsonDocument mark(final String databaseName, final RawBsonDocument command) {
 
         try {
             try {
+                System.out.println(command);
```
jyemin: Ditch the debugging  :)

kevinAlbs: Oops, fixed.

kevinAlbs:
I'll wait to merge until https://github.com/mongodb/specifications/pull/606 is approved.

PR #518, opened 2019-07-18
Fix FLE tests

kevinAlbs:
I think I got tripped up before because updating the JSON files in:

```
driver-core/src/test/resources/client-side-encryption
```

Didn't update the ones that the test runner seemed to use:

```
driver-core/out/test/resources/client-side-encryption/
```

I think this is the correct fix, but I'm can't fully run the prose test – creating a data key isn't adding the key alt name, and I think I'm not picking up the fix from [keyAltName fix](https://github.com/mongodb/libmongocrypt/commit/feceb84e0ef4305d3ab6b7d3e3552955491c6f67) though I'm publishing to maven local after that commit. I must be doing something wrong locally.

@jyemin would you be able to validate this fix works? If so, I'll make the corresponding change in the specifications repo.

### DIFF
--- a/driver-core/src/test/resources/client-side-encryption-corpus/corpus-schema.json
+++ b/driver-core/src/test/resources/client-side-encryption-corpus/corpus-schema.json
@@ -98,18 +98,6 @@
         }
       }
     },
-    "aws_string_det_auto_altname": {
-      "bsonType": "object",
-      "properties": {
-        "value": {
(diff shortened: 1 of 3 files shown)


kevinAlbs:
Created the corresponding spec update https://github.com/mongodb/specifications/pull/613. I'll merge this after the spec change is approved.

PR #522, opened 2019-08-26
Update evergreen logging for FLE

vincentkam:
This PR partially updates Evergreen logging for FLE, per offline discussions with @jeff (see https://jira.mongodb.org/browse/JAVA-3403 for more information)
https://evergreen.mongodb.com/patch/5d60615e5623432e241a3890

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -120,7 +120,12 @@ if [ "$SLOW_TESTS_ONLY" == "true" ]; then
     ./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI} \
               ${TRANSACTION_URI} ${GRADLE_EXTRA_VARS} ${ASYNC_TYPE} --stacktrace --info testSlowOnly
 else
+    set +x
+    echo ./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI} \
+              -Dorg.mongodb.test.awsAccessKeyId=AWS_ACCESS_KEY_ID -Dorg.mongodb.test.awsSecretAccessKey=AWS_SECRET_ACCESS_KEY \
+              ${TRANSACTION_URI} ${GRADLE_EXTRA_VARS} ${ASYNC_TYPE} --stacktrace --info test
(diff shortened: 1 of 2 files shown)


.evergreen/run-tests.sh
```diff
@@ -120,7 +120,12 @@ if [ "$SLOW_TESTS_ONLY" == "true" ]; then
     ./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI} \
               ${TRANSACTION_URI} ${GRADLE_EXTRA_VARS} ${ASYNC_TYPE} --stacktrace --info testSlowOnly
 else
+    set +x
```
jyemin: I see a lot of docs for `set -x` but none for `set +x`.  Can you point me to something explaining what it does/tell me why you're doing this?

.evergreen/run-tests.sh
```diff
@@ -120,7 +120,12 @@ if [ "$SLOW_TESTS_ONLY" == "true" ]; then
...
+    echo ./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI} \
```
jyemin: Echoing the command seems to defeat the purpose of this change, no?

.evergreen/run-tests.sh
vincentkam: Sure NP! 😸  `+` turns off an option and `-` turns on an option.  Thus `set -x` would result in a trace of all the commands and their arguments being printed and `set +x` would disable this behavior. 

Source: https://ss64.com/bash/set.html
Relevant sections:
> Using + rather than - will cause the option to be turned off. 


and

>    -x  Print a trace of simple commands and their arguments
       after they are expanded and before they are executed

.evergreen/run-tests.sh
vincentkam: Ah sorry for the confusion:  the command is merely echoing the literal strings of `AWS_SECRET_ACCESS_KEY`  and `AWS_ACCESS_KEY_ID` (as opposed to echoing the contents of the variable via   `${AWS_ACCESS_KEY_ID}`  and `${AWS_SECRET_ACCESS_KEY}`

jyemin:
Since Java driver Evergreen projects are private, there's no pressing need for this.

PR #524, opened 2019-11-22
Change "less" to "greater" in error message

pkhruasu-ui:


rozza:
Thank you for the PR!

PR #530, opened 2020-01-14
link to centralized compatibility chart on ecosystem

ccho-mongodb:
In the "Upgrading" section, the MongoDB server and driver version compatibility chart was out of date. Rather than updating this in multiple places, let's link to the ecosystem driver page which is maintained by the Developer Education Team.


jyemin:
I updated the matrix rather than remove it.  Given that we plan to move these docs onto docs.mongodb.com relatively soon, that will be a good time to consolidate the matrix.

PR #532, opened 2020-01-20
Sync the POJO quickstart doc with PojoQuickTour.java

jstewart-mongo:
JAVA-3213


jstewart-mongo:
Yes, the quick tour was updated in JAVA-2751, but the associated code samples in the documentation were not updated along with the code.

PR #533, opened 2020-01-21
changes needed for morphia 2.0

evanchooly:
This opens up a handful of classes so they are visible from Morphia codecs and related mapping code.

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/PojoSpecializationHelper.java
+++ b/bson/src/main/org/bson/codecs/pojo/PojoSpecializationHelper.java
@@ -20,10 +20,10 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
-final class PojoSpecializationHelper {
+public final class PojoSpecializationHelper {
 
     @SuppressWarnings("unchecked")
-    static <V> TypeData<V> specializeTypeData(final TypeData<V> typeData, final List<TypeData<?>> typeParameters,
+    public static <V> TypeData<V> specializeTypeData(final TypeData<V> typeData, final List<TypeData<?>> typeParameters,
                                               final TypeParameterMap typeParameterMap) {
         if (!typeParameterMap.hasTypeParameters() || typeParameters.isEmpty()) {
             return typeData;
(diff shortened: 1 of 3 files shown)


evanchooly:
incorporated some classes from the driver rather than rely on driver changes.  those classes aren't likely to change (much) and it doesn't make much sense to make them public like i would need.  these two changes feel pretty fundamantal but i'll keep whittling away as much as I can.  alternately, i can just pull those classes in, too, but I'm hoping to centralize the kind of work they do as much as i can in the driver and not have duplicated mapping code.

rozza:
Hi @evanchooly hope things are good.

Thanks for the PR and trimming down the scope of the changes.  I've been working on [JAVA-3408](https://jira.mongodb.org/browse/JAVA-3408) to improve the mapping of nested generic types.  In that work I have changed the `TypeParameterMap` class a bit.  See: rozza/mongo-java-driver#354 for the full details.

I'm hoping that will help your cause - but let me know, feel free to email me if easier.

Ross

evanchooly:
rc0 still breaks morphia, sadly.  The types I need still aren't public types.

evanchooly:
updated PR to reflect current needs

rozza:
So from my understanding Morphia doesn't use the PojoCodec / ClassModel implementation (https://github.com/rozza/mongo-java-driver/pull/354#discussion_r373583375) so rather than commit to extending the public API and having to support these internal classes for the 4.0 lifetime. I think we should keep them private at least for the 4.0 release which is imminent.

I'm happy to re-review this decision for a 4.x release but making the API public puts limitations on the flexibility of the API and fixing / improving issues such as: rozza/mongo-java-driver#354

So to conclude I think for now the best path is to pull the code into Morphia and continue to work with your custom EntityModel logic rather than bind the API's.  This also should allow both projects to move at differing speeds to any issues / new features in the future.

I hope that sounds reasonable path for now @evanchooly?

Ross

evanchooly:
I was afraid that'd be your answer.  I am, in fact, currently embedded my modified version of those classes in the morphia jar.  The catch is that i'm actually using the same namespace so my versions are essentially hiding yours.  I went this route because to move them over in to a morphia specific namespace would mean copying the entire pojo package tree and bits even higher.  That led to a staggering number of files to copy over and keep in sync.  As gross as shadowing the driver classes is, it seems to be the simplest solution.

So for now I'll just run with what I have.  It seems to be working and I'll see what I can do to lessen the need for such gymnastics going forward.  I understand the hesitance to open up classes like that but it essentially makes it near impossible to build much on top of these classes.  I understand that isn't your goal or concern, necessarily, but it does make my job a bit more difficult.  But I have something workable now and I guess I'll just run with that until it doesn't work anymore.

PR #534, opened 2020-01-21
Changes for JAVA-3420

niccottrell:


docs/reference/content/driver/tutorials/ssl.md
```diff
@@ -189,11 +194,17 @@ MongoDB server:
...
 [`keytool`](http://docs.oracle.com/javase/8/docs/technotes/tools/unix/keytool.html)
 or the [`openssl`](https://www.openssl.org/docs/apps/openssl.html)
-command line program.
+command line program. For example:
+
+```sh
+openssl pkcs12 -export -out mongodb.pkcs12 -in client.pem
+```
+
+Once this keyStore file is created, it can be set in the JVM at launch with `-Djavax.net.ssl.keyStore=c:\<PATH>\mongodb.pkcs12`.
```
jyemin: Windows is not our most popular platform for the JVM, so perhaps use:

    -Djavax.net.ssl.keyStore=<path to file>/mongodb.pkcs12

instead?

jyemin:
@niccottrell would you mind cleaning up this PR a bit and re-submitting?  It no longer merges on master and seems to have commits on it from other tickets (JAVA-3530) that have already been committed.

PR #535, opened 2020-01-28
Adding BsonArray constructor that allocates an ArrayList with known capacity

gaspardpetit:
This probably does not need much additional explanations. I could not find a way in the current `BsonArray` implementation to specify the capacity.

From https://dzone.com/articles/performance-evaluation-of-java-arraylist

> It is clear from the from the results (considering the add operation of ArrayList) that if the required maximum capacity of the ArrayList is known, we can get the optimal performance (both average latency and throughput) by specifying the initial capacity to the required capacity.

> In doing so, we can get 24% to 34% improvement in average latency and 30% to 50% improvement in throughput.

JAVA-3613

### DIFF
--- a/bson/src/main/org/bson/BsonArray.java
+++ b/bson/src/main/org/bson/BsonArray.java
@@ -242,7 +242,7 @@ public class BsonArray extends BsonValue implements List<BsonValue>, Cloneable {
 
     @Override
     public BsonArray clone() {
-        BsonArray to = new BsonArray();
+        BsonArray to = new BsonArray(this.size());
         for (BsonValue cur : this) {
             switch (cur.getBsonType()) {
                 case DOCUMENT:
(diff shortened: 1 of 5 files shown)


rozza:
Hi @gaspardpetit, thanks for the PR.  I've added https://jira.mongodb.org/browse/JAVA-3613 to cover this and #536

bson/src/main/org/bson/BsonArray.java
```diff
@@ -52,6 +52,17 @@ public BsonArray() {
         this(new ArrayList<BsonValue>(), false);
     }
 
+    /**
+     * Construct an empty BsonArray with the specified initial capacity.
+     *
+     * @param  initialCapacity  the initial capacity of the BsonArray
+     * @throws IllegalArgumentException if the specified initial capacity
+     *         is negative
+     */
```
jyemin, line 61: Add

    @since 4.1

jyemin:
Sorry for leaving this un-merged for so long.  We lost track of it after the initial review.  I applied the requested changed myself and merged: https://github.com/mongodb/mongo-java-driver/commit/55ca500447d2d0126f36e2fa16ba0061113785b6 and https://github.com/mongodb/mongo-java-driver/commit/82e40ed6929c2b0e7dcd65f3615e10b183275b1e

PR #536, opened 2020-01-28
Adding BsonDocument constructor that allocates a LinkedHashMap with known capacity

gaspardpetit:
Similar to https://github.com/mongodb/mongo-java-driver/pull/535 but this time for BsonDocument

JAVA-3613

### DIFF
--- a/bson/src/main/org/bson/BsonDocument.java
+++ b/bson/src/main/org/bson/BsonDocument.java
@@ -49,7 +49,7 @@ import static java.lang.String.format;
 public class BsonDocument extends BsonValue implements Map<String, BsonValue>, Cloneable, Bson, Serializable {
     private static final long serialVersionUID = 1L;
 
-    private final Map<String, BsonValue> map = new LinkedHashMap<String, BsonValue>();
+    private final Map<String, BsonValue> map;
 
     /**
      * Parses a string in MongoDB Extended JSON format to a {@code BsonDocument}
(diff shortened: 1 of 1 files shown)


rozza:
Hi @gaspardpetit, thanks for the PR.  I've added https://jira.mongodb.org/browse/JAVA-3613 to cover this and #535

bson/src/main/org/bson/BsonDocument.java
```diff
@@ -81,13 +82,25 @@ public BsonDocument(final List<BsonElement> bsonElements) {
      * @param value the value
      */
     public BsonDocument(final String key, final BsonValue value) {
+        this(1);
```
jyemin, line 85: It's quite common usage, especially in the driver itself, to do this:

     var doc = new BsonDocument(key, val).append(key2, val2).append(key3, val3)...
     
instead of      

     var doc = new BsonDocument().append(key, val).append(key2, val2).append(key3, val3)...

Given that, this change could do more harm than good.  Can you revert it?

bson/src/main/org/bson/BsonDocument.java
```diff
@@ -81,13 +82,25 @@ public BsonDocument(final List<BsonElement> bsonElements) {
...
         put(key, value);
     }
 
+    /**
+     * Construct an empty document with the specified initial capacity.
+     *
+     * @param  initialCapacity the initial capacity
+     * @throws IllegalArgumentException if the initial capacity is negative
+     */
```
jyemin, line 94: Add

    @since 4.1

jyemin:
Sorry for leaving this un-merged for so long.  We lost track of it after the initial review.  I applied the requested changed myself and merged: https://github.com/mongodb/mongo-java-driver/commit/661985cc0d315ae63cc348576ddef1fb7a946f68

PR #540, opened 2020-03-03
[JAVA-3644] Аdd to call of callback if error raise on decoding of document

evgzakharov:
If some error raise on decoding of document, watch stream is stumping on error and never close or receive other documents
So add catching an error on decoding and if it raise - call callback with error.

https://jira.mongodb.org/browse/JAVA-3644

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
@@ -166,16 +166,19 @@ final class AsyncChangeStreamBatchCursor<T> implements AsyncAggregateResponseBat
                     callback.onResult(null, t);
                 } else if (rawDocuments != null) {
                     List<T> results = new ArrayList<T>();
-                    try {
-                        for (RawBsonDocument rawDocument : rawDocuments) {
-                            if (!rawDocument.containsKey("_id")) {
-                                callback.onResult(null,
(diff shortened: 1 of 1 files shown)


driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
```diff
@@ -168,15 +168,22 @@ public void onResult(final List<RawBsonDocument> rawDocuments, final Throwable t
...
-                                    new MongoChangeStreamException("Cannot provide resume functionality when the resume token is missing.")
+                            callback.onResult(
+                                null,
+                                new MongoChangeStreamException(
+                                    "Cannot provide resume functionality when the resume token is missing."
+                                )
                             );
                             return;
                         }
                         results.add(rawDocument.decode(changeStreamOperation.getDecoder()));
```
rozza: I think this decode method needs to be in the try / catch.

evgzakharov: ohh.. sorry. In the last commit I have hurry up a bit.. 
All fixed now =)

evgzakharov:
@rozza is there anything else i can do with this MR?

rozza:
Squashed and merged in 719b9d1

Many thanks @evgzakharov for the PR and Jira ticket.

evgzakharov:
Great! Thank you too, @rozza!

PR #541, opened 2020-03-09
JAVA-3649 host returned by server may contain dash

bazile-clyde:


bazile-clyde:
Nope. There's no spec test for this. It's only mentioned in the table just above the section [here](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#mongocredential-properties-5)

PR #542, opened 2020-03-19
Made CodecConfigurationException error message more explicit

Kira-Cesonia:
It just took me several hours to figure out the fact that the problem was not that no public constructor existed (because it did), but that the problem was that the BSON deserializer requires the class to have an *empty* constructor with *no arguments*. Once I figured that out, I was able to solve the problem within less than a minute. I hope that this more explicit error message helps other people who run into the same issue with their first time trying to deserialize BSON objects.

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
+++ b/bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
@@ -137,7 +137,9 @@ final class CreatorExecutable<T> {
 
     private void checkHasAnExecutable() {
         if (constructor == null && method == null) {
-            throw new CodecConfigurationException(format("Cannot find a public constructor for '%s'.", clazz.getSimpleName()));
+            throw new CodecConfigurationException(format("Cannot find a public constructor for '%s'.%n"
+                                                       + "Please ensure the class '%s' has a public, empty constructor with no arguments",
+                                                       clazz.getSimpleName()));
         }
     }
 


bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
```diff
@@ -137,7 +137,9 @@ CodecConfigurationException getError(final Class<?> clazz, final String msg) {
 
     private void checkHasAnExecutable() {
         if (constructor == null && method == null) {
-            throw new CodecConfigurationException(format("Cannot find a public constructor for '%s'.", clazz.getSimpleName()));
+            throw new CodecConfigurationException(format("Cannot find a public constructor for '%s'.%n"
```
jyemin, line 140: This is causing `org.bson.codecs.pojo.PojoCustomTest.testConstructorNotPublicModel` to fail with the exception `java.util.MissingFormatArgumentException: Format specifier '%s'`, because there are not enough arguments.  I don't see a need to repeat the class name (or a newline for that matter), so I think it would be fine to just have:

     format("Cannot find a public constructor for '%s'.  "Please ensure the class has a public, empty constructor with no arguments, clazz.getSimpleName())

jyemin:
Merged on command line: https://github.com/mongodb/mongo-java-driver/commit/219ac6a4f335b85b35b3210d8d3b72608200f040

Apologies for leaving this sitting for so long, and thanks for the contribution.

PR #543, opened 2020-03-20
throw exceptions for invalid host labels

bazile-clyde:


.evergreen/run-mongodb-aws-ecs-test.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Don't trace since the URI contains a password that shouldn't show up in the logs
+set -o xtrace
```
bazile-clyde: This is actually ok to trace since the URI for ECS test don't contain a secret.

driver-core/src/test/unit/com/mongodb/internal/connection/AuthorizationHeaderTest.java
```diff
@@ -66,6 +66,26 @@ public void testGetRegion() throws SaslException {
...
+    @Test(expected = SaslException.class)
+    public void shouldThrowErrorOnEmptyHostLabelAtEnd() throws SaslException {
+        AuthorizationHeader.getRegion("abc.def.");
+    }
+
+    @Test(expected = SaslException.class)
+    public void shouldThrowErrorOnEmptyHost() throws SaslException {
+        AuthorizationHeader.getRegion("");
+    }
+
```
bazile-clyde: These tests just make sure that empty values throw exceptions per the [spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#id11).

bazile-clyde:
Found a few fixes that Java needed while reviewing other POCs

PR #549, opened 2020-05-26
DRIVERS-1279 update matrix spec for mongodb-aws

bazile-clyde:


bazile-clyde:
[Here's](https://evergreen.mongodb.com/version/5ecd8914c9ec4449f8a52c75) the evergreen patch I ran before submitting

.evergreen/.evg.yml
```diff
@@ -1069,8 +1069,8 @@ buildvariants:
     - name: "plain-auth-test"
 
 - matrix_name: "aws-auth-test"
-  matrix_spec: { jdk: "jdk11" }
-  display_name: "MONGODB-AWS Auth test"
+  matrix_spec: { auth: "noauth", ssl: "nossl", jdk: "jdk11", version: ["4.4", "latest"], os: "ubuntu" }
+  display_name: "MONGODB-AWS Auth test ${version} ${os}"
```
jyemin, line 1073: Change `os` to `jdk`.  We typically don't display OS name since we only test on Linux anyway.

.evergreen/.evg.yml
```diff
@@ -1069,8 +1069,8 @@ buildvariants:
     - name: "plain-auth-test"
 
 - matrix_name: "aws-auth-test"
-  matrix_spec: { jdk: "jdk11" }
-  display_name: "MONGODB-AWS Auth test"
+  matrix_spec: { auth: "noauth", ssl: "nossl", jdk: "jdk11", version: ["4.4", "latest"], os: "ubuntu" }
```
jyemin, line 1072: Let's just on jdk8 as well, since that's our minimum supported version.

jyemin, line 1072: It's a bit confusing that it's an auth test but we have to specify `auth = "noauth"`.  Is is necessary to have that bit, or will it default to `noauth` if not specified?

.evergreen/.evg.yml
bazile-clyde: Done

.evergreen/.evg.yml
bazile-clyde: Done. Also, test are failing for jdk8 on both latest and 4.4

bazile-clyde: Removed. This was accidental. We actually specify "AUTH: auth" for the AWS tests so this was a noop.

jyemin: Same error for both?

bazile-clyde:
It now builds with jdk11 all tests. The latest evergreen results are [here](https://evergreen.mongodb.com/version/5ed00b107742ae725a55a1f0)

PR #551, opened 2020-06-08
Improve the support for mapped find iterables

fleonce:
Currently, when using FindIterable#map, the resulting MongoIterable will not allow you to use find operation methods like FindIterable#sort, FindIterable#limit and so on. I updated FindIterable#map to return a FindIterable, backed by a new internal class for mapping FindIterables called MappingFindIterable, which provides all the comfort for queries just as the FindIterable interface itself does.


fleonce:
Closed the PR for now as the compilation check failed. Will reopen when the error is found and resolved.

PR #552, opened 2020-06-10
[JAVA-3759] Fix for BsonBinary subtype v4 conversion to UUID

arushi-dhunna:
https://jira.mongodb.org/browse/JAVA-3759
Bug: BsonBinary subtype v4 (UUID Standard) gets converted back to Binary instead of UUID

On reading UUID subtype V4 from MongoDB using mongo-java-driver 3.12.5, BsonBinary subtype v4 (STANDARD Uuid) gets converted back to Binary and not UUID.

Line of code where this bug occurs:
https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/BSONCallbackAdapter.java#L86

Which throws mapping exception only in case of UUID v4 (and not in V3) : Causing: dev.morphia.mapping.MappingException: Could not map Entity with ID: org.bson.types.Binary@e1123829

As a result, calling client is forced to add a decoding hook transformer for Binary to UUID

Options

### DIFF
--- a/bson/src/main/org/bson/BSONCallbackAdapter.java
+++ b/bson/src/main/org/bson/BSONCallbackAdapter.java
@@ -18,6 +18,7 @@ package org.bson;
 
 import org.bson.types.Decimal128;
 import org.bson.types.ObjectId;
+import org.bson.internal.UuidHelper;
 
 import static org.bson.Bits.readLong;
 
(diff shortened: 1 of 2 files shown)


jyemin:
As discussed on the Jira issue, this would be a backwards breaking change.

PR #554, opened 2020-06-19
[JAVA-3402] Override provider URL when initializing DNS context in case it is def…

danieljohngomez:
When a jndi.properties is added on the classpath and `java.naming.provider.url` is set to a value other than `dns:`, connecting via mongodb+srv scheme fails with error `javax.naming.ConfigurationException: Invalid URI`.

This PR ensures that provider url is always set to `dns:`.

The error can be reproduced by adding jndi.properties file with content `java.naming.provider.url=SomeProvider` to test resources then running SRV connection tests.


danieljohngomez:
I also found this change on [Netflix' eureka client](https://github.com/Netflix/eureka/blob/master/eureka-client/src/main/java/com/netflix/discovery/endpoint/DnsResolver.java#L45)

PR #555, opened 2020-07-09
[JAVA-3782] Use the platform-specific line separator in Authorization header

fhassak:


fhassak:
Hello, please do not merge. I think my solution is wrong. I need to investigate further this issue.

fhassak:
The second commit is the correct solution. 
Aws use unix-style line endings (`\n`)

config/findbugs-exclude.xml
```diff
@@ -169,4 +169,10 @@
         <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE"/>
     </Match>
 
+    <!-- AWS use UNIX-style line endings -->
+    <Match>
+        <Class name="com.mongodb.internal.connection.AuthorizationHeader"/>
+        <Bug pattern="VA_FORMAT_STRING_USES_NEWLINE"/>
+    </Match>
```
bazile-clyde, line 176: Is this necessary for the tests to pass on Windows? I suspect the change made in `AuthorizationHeader.java:118` fixes the issue since it no longer uses platform-specific newline characters.

jyemin:
@fhassak thanks for the fix.  We are much obliged!

PR #556, opened 2020-07-25
Fix visit UUID constructor

rmanibus:
The condition in visitUUIDConstructor is always true and make it return a UUID_LEGACY regardless of the uuidConstructorName
```
        BsonBinarySubType subType = BsonBinarySubType.UUID_STANDARD;
        if (!"UUID".equals(uuidConstructorName) || !"GUID".equals(uuidConstructorName)) {
            subType = BsonBinarySubType.UUID_LEGACY;
        }
```


jyemin:
Closed by agreement on https://jira.mongodb.org/browse/JAVA-3799

PR #560, opened 2020-08-06
JAVA-3201: Add workload executor

jyemin:
https://jira.mongodb.org/browse/JAVA-3201

Patch build in the drivers-atlas-testing project: https://spruce.mongodb.com/version/5f2b60bc850e611d15e2c51e/tasks

See also https://github.com/jyemin/drivers-atlas-testing/commit/fc8c05c39920583922396da031191a306228aafd for how it's invoked from Astrolabe

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ def coreProjects = subprojects.findAll { !['util'].contains(it.name) }
 def javaProjects = subprojects.findAll { !it.name.contains('scala') }
 def scalaProjects = subprojects.findAll { it.name.contains('scala') }
 def javaMainProjects = javaProjects.findAll { !['util'].contains(it.name) }
-def javaCodeCheckedProjects = javaMainProjects.findAll { !['driver-benchmarks'].contains(it.name) }
+def javaCodeCheckedProjects = javaMainProjects.findAll { !['driver-benchmarks', 'driver-workload-executor'].contains(it.name) }
 def javaAndScalaTestedProjects = javaCodeCheckedProjects + scalaProjects
 
 configure(coreProjects) {
--- a/driver-sync/src/test/functional/com/mongodb/client/JsonPoweredCrudTestHelper.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/JsonPoweredCrudTestHelper.java
@@ -71,7 +71,6 @@ import org.bson.BsonString;
 import org.bson.BsonValue;
 import org.bson.Document;
 import org.bson.types.ObjectId;
-import org.junit.AssumptionViolatedException;
 import util.Hex;
 
 import java.io.ByteArrayInputStream;
@@ -126,9 +125,6 @@ public class JsonPoweredCrudTestHelper {
         } catch (NoSuchMethodException e) {
             throw new UnsupportedOperationException("No handler for operation " + methodName);
         } catch (InvocationTargetException e) {
-            if (e.getTargetException() instanceof AssumptionViolatedException) {
-                throw (AssumptionViolatedException) e.getTargetException();
-            }
             if (e.getTargetException() instanceof MongoException) {
                 throw (MongoException) e.getTargetException();
             }
--- a/driver-workload-executor/build.gradle
+++ b/driver-workload-executor/build.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright 2008-present MongoDB, Inc.
+ * Copyright 2016-present MongoDB, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,13 +14,33 @@
  * limitations under the License.
  */
 
-include ':bson'
-include ':driver-benchmarks'
-include ':driver-workload-executor'
-include ':driver-core'
-include ':driver-legacy'
-include ':driver-sync'
-include ':driver-reactive-streams'
-include ':bson-scala'
-include ':driver-scala'
-include ':util'
+apply plugin: 'application'
+
+mainClassName = "com.mongodb.workload.WorkloadExecutor"
+
+run {
+    systemProperties System.getProperties()
+}
+
+sourceSets {
+    main {
+        java {
+            srcDir 'src/main'
+        }
+        resources {
+            srcDir 'src/resources'
+        }
+    }
+}
+
+dependencies {
+    api project(':driver-sync')
+
+    implementation project(':driver-sync').sourceSets.test.output
+
+    implementation 'ch.qos.logback:logback-classic:1.1.1'
+}
+
+javadoc {
+    enabled = false
+}
--- a/driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
+++ b/driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
@@ -58,8 +58,8 @@ public class WorkloadExecutor {
      * @param args no arguments are expected
      */
     public static void main(String[] args) {
-        String outputDirectory = System.getProperty("org.mongodb.workload.output.directory") == null
-                ? System.getProperty("user.dir") : System.getProperty("org.mongodb.workload.output.directory");
+        String outputDirectory =
+                System.getProperty("org.mongodb.workload.output.directory", System.getProperty("user.dir"));
         
         Runtime.getRuntime().addShutdownHook(new Thread(() -> {
             try {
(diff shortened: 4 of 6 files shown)


driver-sync/src/test/functional/com/mongodb/client/JsonPoweredCrudTestHelper.java
```diff
@@ -126,9 +125,6 @@ BsonDocument getOperationResults(final BsonDocument operation, @Nullable final C
         } catch (NoSuchMethodException e) {
             throw new UnsupportedOperationException("No handler for operation " + methodName);
         } catch (InvocationTargetException e) {
-            if (e.getTargetException() instanceof AssumptionViolatedException) {
```
jyemin, line 129: This is no longer necessary, and removing it means that the executor doesn't need to take a Junit dependency.

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @param args no arguments are expected
+     */
+    public static void main(String[] args) {
+        String outputDirectory = System.getProperty("org.mongodb.workload.output.directory") == null
+                ? System.getProperty("user.dir") : System.getProperty("org.mongodb.workload.output.directory");
+        
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            try {
+                done = true;
```
jyemin, line 66: This guarantees the executor does at most one more operation before stopping

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -0,0 +1,134 @@
...
+                writeResults(outputDirectory);
```
jyemin, line 67: Write the results from the shutdown hook in order to avoid blocking process exit on hung I/O in the executor.

build.gradle
```diff
@@ -54,7 +54,7 @@ def coreProjects = subprojects.findAll { !['util'].contains(it.name) }
 def javaProjects = subprojects.findAll { !it.name.contains('scala') }
 def scalaProjects = subprojects.findAll { it.name.contains('scala') }
 def javaMainProjects = javaProjects.findAll { !['util'].contains(it.name) }
-def javaCodeCheckedProjects = javaMainProjects.findAll { !['driver-benchmarks'].contains(it.name) }
+def javaCodeCheckedProjects = javaMainProjects.findAll { !['driver-benchmarks', 'driver-workload-executor'].contains(it.name) }
```
jyemin, line 57: Perhaps there's a cleaner way to do this?

rozza: I think this is fine for now, as it allows for centralised  configuration of projects in the main build.

A cleaner alternative would potentially to separate out the logic for code checking into a separate file and only import it in the projects that actually need them.  Then configuration would essentially be contained on a per project basis.  I think when we move to kotlin for the build that would be the ideal time to do that work.

driver-sync/src/test/functional/com/mongodb/client/JsonPoweredCrudTestHelper.java
rozza: Ack

driver-workload-executor/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'application'
+
+mainClassName = "com.mongodb.workload.WorkloadExecutor"
+
+run {
+    systemProperties System.getProperties()
```
rozza, line 22: nice work around.

jyemin:
Ah, that's confusing about the validate task.  It actually failed for a different reason: https://github.com/mongodb-labs/drivers-atlas-testing/issues/82

The errors about `doesNotExist` are by design, it appears: https://github.com/mongodb-labs/drivers-atlas-testing/blob/master/astrolabe/validator.py#L108-L117.  For debugging purposes, the workload executor prints stack trace for any errors (since normally errors are unexpected).

jyemin:
OK, I tried again to use gradle instead of java in the executor script, and now that it's running in the foreground, it works.  So that's great.  But now since it found SLF4J on the classpath it is using logback for logging, and there is no config file, so it's logging at DEBUG.  So I added a logback.xml to bump it up to WARN.

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -0,0 +1,138 @@
...
+
+            collection.withWriteConcern(WriteConcern.MAJORITY).drop();
+            collection.withWriteConcern(WriteConcern.MAJORITY).insertMany(documents);
+        }
+
+        JsonPoweredCrudTestHelper helper = new JsonPoweredCrudTestHelper("Workload executor", database, collection);
+
+        BsonArray operations = workload.getArray("operations");
+
+        outer:
```
rozza, line 101: TIL named loops.

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -0,0 +1,138 @@
...
+    private static final AtomicInteger numFailures = new AtomicInteger();
+    private static final AtomicInteger numErrors = new AtomicInteger();
+
+    /**
+     * The main entry point
+     *
+     * @param args no arguments are expected
+     */
+    public static void main(String[] args) {
+        String outputDirectory = System.getProperty("org.mongodb.workload.output.directory") == null
```
rozza, line 61: Potentially simplify to use a default property.

`String outputDirectory = System.getProperty("org.mongodb.workload.output.directory", System.getProperty("user.dir"))`

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -0,0 +1,138 @@
...
+            for (BsonValue cur : operations.getValues()) {
+                try {
+                    if (done) {
+                        break outer;
+                    }
+                    BsonDocument operation = cur.asDocument().clone();
+                    BsonValue expectedResult = operation.get("result");
+                    BsonDocument resultDocument = helper.getOperationResults(operation);
+
+                    if (expectedResult != null && !resultDocument.get("result").equals(expectedResult)) {
```
rozza, line 112: Can expectedResult be null and the operation return a result?

jyemin: Yes, I think that's allowed. It means that the runner shouldn't check the result and just treat any normal return as a success.

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
jyemin: Done

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -0,0 +1,138 @@
...
+        MongoCollection<BsonDocument> collection = database.getCollection(workload.getString("collection").getValue(),
+                BsonDocument.class);
+
+        if (workload.containsKey("testData")) {
+            List<BsonDocument> documents = new ArrayList<>();
+            for (BsonValue document : workload.getArray("testData")) {
+                documents.add(document.asDocument());
+            }
+
+            collection.withWriteConcern(WriteConcern.MAJORITY).drop();
```
prashantmital, line 93: You can remove this as well - `astrolabe` will drop the collection and insert data ensuring a consistent start state.

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+        BsonDocument workload = BsonDocument.parse(workloadSpec);
+
+        MongoDatabase database = mongoClient.getDatabase(workload.getString("database").getValue());
+        MongoCollection<BsonDocument> collection = database.getCollection(workload.getString("collection").getValue(),
+                BsonDocument.class);
+
+        if (workload.containsKey("testData")) {
+            List<BsonDocument> documents = new ArrayList<>();
+            for (BsonValue document : workload.getArray("testData")) {
```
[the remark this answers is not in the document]
jyemin: Why is it in the workload definition?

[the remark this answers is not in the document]
prashantmital: Not sure why it was designed in this manner. I agree that it can be moved to a top-level key instead of being in the workload definition at some point in the future.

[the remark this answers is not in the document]
jyemin: Done

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
jyemin: Done

PR #561, opened 2020-08-10
Fix scala documentation for updateMany

DrPhil:
The old documentation seems to have been copied from `updateOne`. This makes the docs match the docs from `driver-reactive-streams`.


rozza:
Thanks @DrPhil

PR #564, opened 2020-08-18
Docs: Added a link to the shared core

rozza:
JAVA-3803

http://rozza.github.io/mongo-java-driver/4.1/apidocs/

I opted not to merge the sources for each areas java docs, as that makes the index page very hard to read - losing the initial benefits of splitting out the docs. It also significantly increases the size of the docs - from 45mb to 110mb

### DIFF
--- a/docs/reference/content/apidocs.md
+++ b/docs/reference/content/apidocs.md
@@ -17,3 +17,4 @@ The following API documentation is available:
   - [Java Reactive Streams Driver]({{< apiref "mongodb-driver-reactivestreams" "index.html" >}})
   - [Scala Driver]({{< apiref "mongo-scala-driver" "index.html" >}})
   - [Bson]({{< apiref "bson" "index.html" >}}) (The bson layer)
+  - [Core]({{< apiref ""mongodb-driver-core" "index.html" >}}) (The shared core classes)


rozza:
This fixes the ticketed issue with the core types being linkable to from the core docs site and is the quickest fix.

Search works but only by finding items that use that type - eg: Finding UpdateOptions is two clicks, one for a method that uses it and then one to go to the actual API Docs. The back button works to go between core and sync. So browser navigation is ok.

Option 2: combining the docs makes the index pages harder to read and blows up the size of the docs.
Option 3. That is still possible outside of this fix but would have to evaluated as to what details to add.  Also, the rest of the docs should really introduce the API docs - so is this option overkill?

jyemin:
Agreed that this is a quick fix.

Agreed that search works in the way you describe.   Additionally, if you use the Search bar on the top right, sometimes you can find that top level method when you search for the type of one of its parameters.  e.g. search for UpdateOptions and you'll be presented with all the methods that take one as a param.

> combining the docs makes the index pages harder to read

Not sure what is harder.  Is that it doesn't guide you to the top level of the API effectively when looking at http://rosslawley.co.uk/mongo-java-driver/4.1/apidocs/mongodb-driver-sync/index.html?  If so, I think that's the case with Java API docs in general.

> blows up the size of the docs

Are you referring to the size of the gh-pages branch?  Does that matter?  It's already huge because it has copies for every release.

> so is this option overkill?

I'm not sure, it would need user research to know for sure.  Ultimately, can users find the info they are looking for before they get frustrated and give up. 

If you want to start here and re-evaluate later, I can LGTM this.  Also, I realize that this page will be re-written by the docs team soon as part of the overhaul.

PR #565, opened 2020-08-20
[JAVA-3821] Set specific cipher suites on ssl connection

baekya884:
i add ssloption that set specific ciphersuite when connection. when using ssl connection maybe will help.


baekya884:
i just change heartbeat test wait time from 500ms to 1500ms at commit 4f3ed2e.
why test dose failed?

jyemin:
The discussion on JAVA-3821 about this petered out, so closing this.  We can re-open if necessary.

PR #566, opened 2020-08-29
JAVA-3485: Suppress firing of duplicate cluster events

jyemin:
The design problem presented by this requirement is that currently we combine the ServerListener registered by Cluster implementations to be informed of server changes with any ServerListeners registered by an application.  This turns out to be incorrect, because different rules apply.  The Cluster needs to be informed of all changes to server descriptions, e.g. including changes to RTT.  But the spec requires that server events are only delivered to applications under certain rules of equality.  So to implement this, in the first commit I refactored Server and Cluster implementations to separate the two types of listeners (adding a new private interface for the former).  In the second commit I implemented the event suppression rules required by the specification.

* https://jira.mongodb.org/browse/JAVA-3485
* https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#server-description-equality
* https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#server-description-equality

https://spruce.mongodb.com/version/5f4ab0a92a60ed0c9dcaac25/tasks

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultServer.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultServer.java
@@ -57,6 +57,7 @@ class DefaultServer implements ClusterableServer {
     private final ConnectionFactory connectionFactory;
     private final ServerMonitor serverMonitor;
     private final ChangeListener<ServerDescription> serverStateListener;
+    private final ServerDescriptionChangedListener serverDescriptionChangedListener;
     private final ServerListener serverListener;
     private final CommandListener commandListener;
     private final ClusterClock clusterClock;
@@ -65,7 +66,9 @@ class DefaultServer implements ClusterableServer {
 
     DefaultServer(final ServerId serverId, final ClusterConnectionMode clusterConnectionMode, final ConnectionPool connectionPool,
                   final ConnectionFactory connectionFactory, final ServerMonitorFactory serverMonitorFactory,
-                  final ServerListener serverListener, final CommandListener commandListener, final ClusterClock clusterClock) {
+                  final ServerDescriptionChangedListener serverDescriptionChangedListener, final ServerListener serverListener,
+                  final CommandListener commandListener, final ClusterClock clusterClock) {
+        this.serverDescriptionChangedListener = notNull("internalServerListener", serverDescriptionChangedListener);
         this.serverListener = notNull("serverListener", serverListener);
         this.commandListener = commandListener;
         this.clusterClock = notNull("clusterClock", clusterClock);
@@ -201,7 +204,8 @@ class DefaultServer implements ClusterableServer {
             connectionPool.close();
             serverMonitor.close();
             isClosed = true;
-            serverListener.serverClosed(new ServerClosedEvent(serverId));
+            ServerClosedEvent event = new ServerClosedEvent(serverId);
+            serverListener.serverClosed(event);
         }
     }
 
@@ -298,7 +302,10 @@ class DefaultServer implements ClusterableServer {
             ServerDescription oldDescription = description;
             if (shouldReplace(oldDescription, event.getNewValue())) {
                 description = event.getNewValue();
-                serverListener.serverDescriptionChanged(new ServerDescriptionChangedEvent(serverId, description, oldDescription));
+                ServerDescriptionChangedEvent serverDescriptionChangedEvent =
+                        new ServerDescriptionChangedEvent(serverId, description, oldDescription);
+                serverDescriptionChangedListener.serverDescriptionChanged(serverDescriptionChangedEvent);
+                serverListener.serverDescriptionChanged(serverDescriptionChangedEvent);
             }
         }
 
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultServer.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultServer.java
@@ -205,8 +205,7 @@ class DefaultServer implements ClusterableServer {
             connectionPool.close();
             serverMonitor.close();
             isClosed = true;
-            ServerClosedEvent event = new ServerClosedEvent(serverId);
-            serverListener.serverClosed(event);
+            serverListener.serverClosed(new ServerClosedEvent(serverId));
         }
     }
 
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultServer.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultServer.java
@@ -45,6 +45,7 @@ import static com.mongodb.connection.ServerConnectionState.CONNECTING;
 import static com.mongodb.internal.async.ErrorHandlingResultCallback.errorHandlingCallback;
 import static com.mongodb.internal.connection.ClusterableServer.ConnectionState.AFTER_HANDSHAKE;
 import static com.mongodb.internal.connection.ClusterableServer.ConnectionState.BEFORE_HANDSHAKE;
+import static com.mongodb.internal.connection.EventHelper.equivalentEvent;
 import static com.mongodb.internal.operation.ServerVersionHelper.FOUR_DOT_TWO_WIRE_VERSION;
 import static java.util.Arrays.asList;
 
@@ -305,7 +306,9 @@ class DefaultServer implements ClusterableServer {
                 ServerDescriptionChangedEvent serverDescriptionChangedEvent =
                         new ServerDescriptionChangedEvent(serverId, description, oldDescription);
                 serverDescriptionChangedListener.serverDescriptionChanged(serverDescriptionChangedEvent);
-                serverListener.serverDescriptionChanged(serverDescriptionChangedEvent);
+                if (!equivalentEvent(description, oldDescription)) {
+                    serverListener.serverDescriptionChanged(serverDescriptionChangedEvent);
+                }
             }
         }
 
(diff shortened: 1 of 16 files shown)


driver-core/src/main/com/mongodb/internal/connection/EventHelper.java
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+            if (!equivalentEvent(curNew, matchingPrev)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    static boolean equivalentEvent(final ServerDescription current, final ServerDescription previous) {
```
durran, line 48: The logic here looks almost identical to `ServerDescription#equals`, just in a different order of the checks. Is there a reason this is duplicated? Or am I potentially missing something?

jyemin: It is similar, but it checks fewer fields (according to the rule of equality from the spec, which is not what is implemented in `ServerDescription#equals`.  It didn't seem worth pulling out the common bits and sharing them, since it's pretty much boilerplate code.  I considered putting this code in a public ServerDescription method, but didn't think it was necessary to commit to a public API for this logic.

durran:
LGTM

driver-core/src/main/com/mongodb/internal/connection/EventHelper.java
```diff
@@ -0,0 +1,128 @@
...
+    @SuppressWarnings("BooleanMethodIsAlwaysInverted")
+    private static boolean exceptionsEquals(final Throwable current, final Throwable previous) {
+        // Compare class equality and message as exceptions rarely override equals
+        Class<?> thisExceptionClass = current != null ? current.getClass() : null;
+        Class<?> thatExceptionClass = previous != null ? previous.getClass() : null;
+        if (!Objects.equals(thisExceptionClass, thatExceptionClass)) {
+            return false;
+        }
+
+        String thisExceptionMessage = current != null ? current.getMessage() : null;
```
bdeleonardis1, line 118: Why did you decide to do your null checks this way instead of something like:
```
if (current == null || previous == null) {
      return current == previous
}
```
And then comparing the fields directly (which would save one null check on current and previous). Is it just best practice to always use null checks before accessing a value and then using Object.equals?

driver-core/src/test/unit/com/mongodb/internal/connection/EventHelperTest.java
```diff
@@ -0,0 +1,165 @@
...
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class EventHelperTest {
+
+    @Test
+    public void testServerDescriptionEventEquivalence() {
+        ServerDescription serverDescription = createBuilder().build();
```
bdeleonardis1, line 53: I would've thought having a really big test with a lot of asserts would be frowned upon stylistically. Why is it okay in this case? Because the equivalentEvent method itself has to be pretty long but is also straight forward?

driver-core/src/main/com/mongodb/internal/connection/DefaultServer.java
```diff
@@ -201,7 +205,8 @@ public void close() {
             connectionPool.close();
             serverMonitor.close();
             isClosed = true;
-            serverListener.serverClosed(new ServerClosedEvent(serverId));
```
bdeleonardis1, line 204: Just curious why you decided to break this up when it doesn't seem related to the rest of your pr?

driver-core/src/main/com/mongodb/internal/event/EventListenerHelper.java
```diff
@@ -74,20 +73,14 @@ public static ServerMonitorListener getServerMonitorListener(final ServerSetting
         }
     }
 
-    public static ServerListener createServerListener(final ServerSettings serverSettings, final ServerListener additionalServerListener) {
```
bdeleonardis1, line 77: I'm a little confused what's going on here... First of all why did this method need to change to not need an additionalServerListener anymore? I couldn't really follow the changes you made to where this gets called.

I think I understand what this does - takes in a bunch of server listeners and creates a new one that broadcasts the messages to all them, but I don't understand why this would be necessary? I thought servers would have two listeners after this ticket (one that communicates with the application, and one that communicates with the cluster (for the description change events). Why are there more than that?

bdeleonardis1: Ah I think I get the second part now... The cluster uses this because it has multiple servers and wants to send the same message to all of them, right? I'm still confused about the first question though.

driver-core/src/main/com/mongodb/internal/connection/DefaultServer.java
jyemin: Good catch.  It's an artifact of a previous design that I later rejected.  I will revert.

driver-core/src/main/com/mongodb/internal/connection/EventHelper.java
jyemin: I like that!  The reason I did it this way is because I had copied this code from ServerDescription.equals, which, because it was written at a time when we couldn't use Java 7 classes, didn't use Objects.equals.  Then when I realized I could use Objects.equals here I didn't think to simplify it further.

driver-core/src/main/com/mongodb/internal/event/EventListenerHelper.java
jyemin: The `additionalServerListener` is the one that is created by the `Cluster` implementations.  In this patch `Cluster` implementations no longer create a `ServerListener`.  Instead they create an instance of the new `ServerDescriptionChangedListener` interface, which is no longer combined with the ones from the application.  Instead, DefaultServer takes an instance of `ServerDescriptionChangedListener` separately in its constructor.

> The cluster uses this because it has multiple servers and wants to send the same message to all of them, right? 

Cluster doesn't send messages to Server.  It's the other way around.  Each Server instance sends a message to its containing Cluster.

driver-core/src/main/com/mongodb/internal/connection/EventHelper.java
jyemin: Done. Please resolve if you like the change

driver-core/src/main/com/mongodb/internal/connection/DefaultServer.java
jyemin: Done. Please resolve if you like the change

driver-core/src/test/unit/com/mongodb/internal/connection/EventHelperTest.java
jyemin: I generally do this for long boilerplate methods.  It seemed overkill to have 20 different tests.

rozza (changes requested):
Looks good - one method name is wrong and a nit.

driver-core/src/main/com/mongodb/internal/connection/ServerDescriptionChangedListener.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.internal.connection;
+
+import com.mongodb.event.ServerDescriptionChangedEvent;
+
+// internal interface that Cluster implementations register with Server implementations in order be notified of changes in server state.
```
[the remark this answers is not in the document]
jyemin: Ah, I was using this style to make it less confusing, so readers wouldn't think the interface was public.

Done

driver-core/src/main/com/mongodb/internal/connection/EventHelper.java
```diff
@@ -0,0 +1,127 @@
...
+final class EventHelper {
+
+    static boolean equivalentEvent(final ClusterDescription current, final ClusterDescription previous) {
+        if (!exceptionsEquals(current.getSrvResolutionException(), previous.getSrvResolutionException())) {
+            return false;
+        }
+        if (current.getServerDescriptions().size() != previous.getServerDescriptions().size()) {
+            return false;
+        }
+        for (ServerDescription curNew: current.getServerDescriptions()) {
```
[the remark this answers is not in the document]
jyemin: I find it more natural with the loop, which might just be my lack of familiarity with streams.  OK to keep it as is?

driver-core/src/main/com/mongodb/internal/connection/EventHelper.java
```diff
@@ -0,0 +1,127 @@
...
+                }
+            }
+            if (!equivalentEvent(curNew, matchingPrev)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    static boolean equivalentEvent(final ServerDescription current, final ServerDescription previous) {
```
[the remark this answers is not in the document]
jyemin: Yeah, the naming choice was deliberate, but I see why it's confusing. I didn't want to call it `equivalentDescriptions`, since that would beg the question of why we're not just using `ServerDescription#equals`.  It's really more like 'wouldDescriptionsGenerateEquivalentEvents'.  Would that be better?  Or perhaps just a comment explaining it? Both?

driver-core/src/main/com/mongodb/internal/connection/EventHelper.java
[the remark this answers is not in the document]
rozza: Yeap - that was just for my learning

driver-core/src/main/com/mongodb/internal/connection/EventHelper.java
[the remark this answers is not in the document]
rozza: I think both would be good - it looks like a typo now and future Ross would probably wonder why we're not using equals!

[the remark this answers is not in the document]
jyemin: Done

rozza (approved):
LGTM

jyemin:
Merged on command line

PR #569, opened 2020-09-01
Java 3778: Add matchEverything 

bdeleonardis1:
I'm not crazy about the name but I agreed with Ross' comment on the ticket that empty might be confusing. I also decided against matchAll because I thought people may be confused with $all.

Other questions:
- when making updates like this, should I be updating the scala driver and the docs all in the same ticket?
- are the tests I selected on my evergreen patch appropriate?

Evergreen patch: https://spruce.mongodb.com/version/5f4e6e1c3627e0664a66406c/tasks


jyemin:
- Yes, you should update Scala and reference docs in the same ticket
- The tests are appropriate.  Better to err on the side of too much testing than too little, but in this case I would have just run a single variant against a standalone, just to make sure the integration tests pass.

driver-core/src/main/com/mongodb/client/model/Filters.java
```diff
@@ -853,6 +854,17 @@ public static Bson jsonSchema(final Bson schema) {
...
     }
 
+    /**
+     * Creates a filter that will match all documents.
+     *
+     * @return the filter
+     * @since 4.2
+     */
+    public static Bson matchEverything() {
+        return new Document();
```
jyemin, line 864: For consistency, use BsonDocument instead of Document

docs/reference/content/builders/filters.md
```diff
@@ -32,7 +32,8 @@ The comparison operator methods include:
 - `lte`: Matches values that are less than or equal to a specified value.
 - `ne`: Matches all values that are not equal to a specified value.
 - `in`: Matches any of the values specified in an array.  
-- `nin`: Matches none of the values specified in an array. 
+- `nin`: Matches none of the values specified in an array.
+- `matchEverything`: Matches all the documents. 
 
 #### Examples
```
jyemin, line 38: It seems appropriate to add an example as well

jyemin:
LGTM.  Let's see what Ross thinks about the name.

driver-scala/src/main/scala/org/mongodb/scala/model/Filters.scala
```diff
@@ -690,6 +690,14 @@ object Filters {
     JFilters.nearSphere(fieldName, x, y, maxDistance.asJava, minDistance.asJava)
   }
 
+  /**
+   * Creates a filter that will match all documents.
```
rozza, line 694: Nit: Keep this the same as the Javadoc: ` Creates an empty filter that will match all documents.`

rozza (commented):
Yeah the name is slightly counter intuitive but the javadoc explains the reasoning well. 

One minor nit for the scala doc - after that LGTM

driver-scala/src/main/scala/org/mongodb/scala/model/Filters.scala
bdeleonardis1: done

bdeleonardis1:
Merged on command line.

PR #571, opened 2020-09-02
Update server discovery monitoring tests with directConnection URI option

jyemin:
https://jira.mongodb.org/browse/JAVA-3824

No patch build since it's just a change to unit tests, which I ran locally.

### DIFF
--- a/driver-core/src/test/unit/com/mongodb/internal/connection/ServerDiscoveryAndMonitoringMonitoringTest.java
+++ b/driver-core/src/test/unit/com/mongodb/internal/connection/ServerDiscoveryAndMonitoringMonitoringTest.java
@@ -102,7 +102,7 @@ public class ServerDiscoveryAndMonitoringMonitoringTest extends AbstractServerDi
                 BsonDocument newDescription = topologyDescriptionChangedEventDocument.getDocument("newDescription");
                 assertEqualClusterDescriptions(createClusterDescriptionFromClusterDescriptionDocument(newDescription),
                         event.getNewDescription());
-                if (newDescription.getString("topologyType").getValue().equals("Single")) {
+                if (newDescription.getString("topologyType").getValue().equals("Single") && isSingleServerClusterExpected()) {
                     assertEquals(SingleServerCluster.class, getCluster().getClass());
                 } else {
                     assertEquals(MultiServerCluster.class, getCluster().getClass());
--- a/driver-core/src/test/resources/server-discovery-and-monitoring-monitoring/discovered_standalone.json
+++ b/driver-core/src/test/resources/server-discovery-and-monitoring-monitoring/discovered_standalone.json
@@ -1,6 +1,6 @@
 {
-  "description": "Monitoring a direct connection",
-  "uri": "mongodb://a:27017/?directConnection=true",
+  "description": "Monitoring a discovered standalone connection",
+  "uri": "mongodb://a:27017/?directConnection=false",
   "phases": [
     {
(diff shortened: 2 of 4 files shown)


driver-core/src/test/unit/com/mongodb/internal/connection/ServerDiscoveryAndMonitoringMonitoringTest.java
```diff
@@ -102,7 +102,7 @@ private void assertEvents(final BsonArray events) {
                 BsonDocument newDescription = topologyDescriptionChangedEventDocument.getDocument("newDescription");
                 assertEqualClusterDescriptions(createClusterDescriptionFromClusterDescriptionDocument(newDescription),
                         event.getNewDescription());
-                if (newDescription.getString("topologyType").getValue().equals("Single")) {
+                if (newDescription.getString("topologyType").getValue().equals("Single") && isSingleServerClusterExpected()) {
```
jyemin, line 105: This change is needed because the implementation type of the `Cluster` doesn't exactly match the topology type in the SDAM specification.  In the spec, `mongodb://a:27017/?directConnection=false` results in a topology type of `Single` if the server identifies itself as a Standalone, but in the Java driver it's still a `MultiServerCluster`

jyemin: Without this change, the `"Monitoring a discovered standalone connection"` test in `discovered_standalone.json` fails this assertion.

bdeleonardis1: I'm sorry, I don't understand how this fixes the problem you described...

`isSingleServerClusterExpected` looks to me like it's checking if the connection is to a standalone in the second condition (only one host and no replica set). So your condition here is basically:
if (server identifies as standalone && (directConnection == true || server is standalone))
                assertEquals(SingleServerCluster.class, getCluster().getClass());

So when `directConnection` is `false`, but the connection is still a standalone you're expecting a `SingleServerCluster`, but you said in your comment that the Java driver would create a `MultiServerCluster`. I'm assuming I have a misunderstanding about the definition of standalone...

jyemin: A standalone is a type of MongoDB deployment (see https://docs.mongodb.com/manual/reference/glossary/#term-standalone).

The Java driver creates the `Cluster` instance before it knows what the deployment type is (because it hasn't connected yet.  That's the responsibility of the `Cluster` instance).  The type of `Cluster` that is created depends on the configuration (i.e the connection string.  

So the check in `isSingleServerClusterExpected` is not about the deployment type, but only the configuration.    The second condition is not the same as saying `server is standalone`.    A connection string like `mongodb://host` could be connecting to a deployment where `host` is a standalone or where `host` is a member of a replica set.  Either way, the driver will create a `SingleServerCluster`.  But a connection string like `mongodb:/host/?replicaSet=test` is definitely referring to a replica set deployment, so the driver will create a `MultiServerCluster`.

bdeleonardis1: Ah okay that makes sense now, thanks.

rozza (approved):
LGTM

PR #572, opened 2020-09-02
JAVA-3292: Add a Codec for converting directly to and from a JSON string

bdeleonardis1:


bson/src/main/org/bson/json/JsonString.java
```diff
@@ -0,0 +1,40 @@
...
+
+package org.bson.json;
+
+/**
+ * A wrapper class that holds a JSON string. This class makes decoding straight
+ * to JSON easy.
+ *
+ * @since 4.2
+ */
+public class JsonString {
```
bdeleonardis1, line 25: 1) I wasn't sure where to put this file. This folder contains json stuff, but it's not that related to the purpose of this file. I was debating between this folder and its parent folder.

2) I decided against providing equals and hashcode methods because I thought users shouldn't really keep JsonStrings, they should just use them to access the string. Do you agree with this decision?

3) I don't have any unit tests for this class. Is this okay for simple wrapper classes like this?

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -58,7 +61,10 @@
 @Immutable
 public final class MongoClientSettings {
     private static final CodecRegistry DEFAULT_CODEC_REGISTRY =
-            fromProviders(asList(new ValueCodecProvider(),
```
bdeleonardis1, line 61: I thought this was kind of ugly, but the alternative is to create a Provider for StringCodecs, just so they can be added to the default registry

bdeleonardis1:
Forgot to run a patch when I first put it up, here it is: https://spruce.mongodb.com/version/5f5108c4e3c3312467bc837b/tasks

bson/src/main/org/bson/json/JsonString.java
```diff
@@ -0,0 +1,50 @@
...
+    private final String json;
+
+    /**
+     * Constructs a new instance with the given JSON string
+     *
+     * @param json the JSON string
+     */
+    public JsonString(final String json) {
```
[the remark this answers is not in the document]
rozza: What happens if a user tries to do:

```
MongoCollection<JsonString> collection.insert(new JsonString("['a', 'b', 'c']"));
```

Do you think there should be any validation on the String?

driver-core/src/main/com/mongodb/MongoClientSettings.java
rozza: Looking at the BsonCodecProvider and GridFSFileCodecProvider they are both very simple providers - so it may be worth adding one. Ultimately,  it doesn't really matter but having a single registry rather than two as default may be marginally beneficial.

rozza (changes requested):
Some comments and nits.

docs/reference/content/bson/extended-json.md
```diff
@@ -116,5 +116,29 @@ try {
 The `Document.parse()` static factory method constructs an instance of a `JsonReader` with the given string and returns an instance of an
 equivalent Document. `JsonReader` automatically detects the JSON flavor in the string, so you do not need to specify it.
 
+### Reading Directly to JSON
```
jyemin, line 119: The example shows both writing JSON directly to MongoDB and reading JSON directly from MongoDB.  Let's clarify in the text above the example that both are supported.

docs/reference/content/bson/extended-json.md
```diff
@@ -116,5 +116,29 @@ try {
...
+If you do not need a document and only want to deal with JSON, you can use the `JsonStringCodec` to read JSON directly from the database.
```
jyemin, line 120: Let's emphasize `JsonString` rather than `JsonStringCodec`

docs/reference/content/bson/extended-json.md
```diff
@@ -116,5 +116,29 @@ try {
 The `Document.parse()` static factory method constructs an instance of a `JsonReader` with the given string and returns an instance of an
 equivalent Document. `JsonReader` automatically detects the JSON flavor in the string, so you do not need to specify it.
 
+### Reading Directly to JSON
+If you do not need a document and only want to deal with JSON, you can use the `JsonStringCodec` to read JSON directly from the database. 
+This is more efficient than constructing a `Document` first and then calling `toJson()`. `JsonStringCodec` is part of the default registry,
```
jyemin, line 121: Also more efficient than calling Document#parse, so point that out as well.

docs/reference/content/bson/extended-json.md
```diff
@@ -116,5 +116,29 @@ try {
...
 
+```java
+MongoDatabase database = mongoClient.getDatabase("mydb");
+MongoCollection<JsonString> collection = database.getCollection("json", JsonString.class);
+collection.insertOne(new JsonString("{hello: 1}"));
+JsonString jsonString = collection.find().first();
+```
+`JsonString` is simply a wrapper class that takes in a `String` in the constructor and returns the `String` in the get `getJson()` method.
+
+You can also provide custom `JsonWriterSettings` to the `JsonStringCodec`, by constructing the codec yourself and then creating your own registy:
```
jyemin, line 132: Let's put the customization in a separate example, since it's optional and may confuse people that just need the simple thing

jyemin (changes requested):

bson/src/main/org/bson/json/JsonString.java
bdeleonardis1: Done

driver-core/src/main/com/mongodb/MongoClientSettings.java
bdeleonardis1: Done

bson/src/main/org/bson/codecs/JsonStringCodec.java
```diff
@@ -0,0 +1,70 @@
...
+ * @since 4.2
+ */
+public class JsonStringCodec implements Codec<JsonString> {
+    private final JsonWriterSettings writerSettings;
+
+    /**
+     * Construct a JsonStringCodec with default JsonWriterSettings
+     */
+    public JsonStringCodec() {
+        this(JsonWriterSettings.builder().outputMode(JsonMode.RELAXED).build());
```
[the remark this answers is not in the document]
bdeleonardis1: I know you said no action required, but I agreed that this seemed weird so I got rid of setting the outputMode explicitly.

bson/src/main/org/bson/json/JsonString.java
```diff
@@ -0,0 +1,50 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bson.json;
+
+/**
+ * A wrapper class that holds a JSON string. This class makes decoding straight
```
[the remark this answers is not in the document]
bdeleonardis1: Done

docs/reference/content/bson/extended-json.md
bdeleonardis1: Done

docs/reference/content/bson/extended-json.md
bdeleonardis1: Done

docs/reference/content/bson/extended-json.md
bdeleonardis1: Done

docs/reference/content/bson/extended-json.md
bdeleonardis1: Done

bson/src/main/org/bson/json/JsonString.java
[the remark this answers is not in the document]
jyemin: Now I'm wondering if we should call this class something different.  Since `[1, 2, 3]` (or even `1`?) is valid JSON, should this class be called `JsonObjectString` instead, if indeed it will forever be restricted to JSON objects (see https://www.json.org/json-en.html)?

Even if we don't change the name, if we are committing to only supporting JSON objects, then this class can and should implement `org.bson.conversions.Bson`.  That way, it can be used for query filters, etc.

Alternatively, we could allow any JSON string, not just JSON objects, though currently BsonReader.pipe currently only supports documents

bson/src/main/org/bson/json/JsonString.java
```diff
@@ -17,20 +17,31 @@
 package org.bson.json;
 
 /**
- * A wrapper class that holds a JSON string. This class makes decoding straight
- * to JSON easy.
+ * A wrapper class that holds a JSON document string. This class makes decoding straight
+ * to JSON easy. Note that this class only holds valid JSON documents, not arrays.
```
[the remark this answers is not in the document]
bdeleonardis1: Done

bson/src/main/org/bson/json/JsonString.java
```diff
@@ -17,20 +17,31 @@
...
  */
 public class JsonString {
     private final String json;
 
     /**
-     * Constructs a new instance with the given JSON string
+     * Constructs a new instance with the given JSON string. Clients must ensure
+     * they only pass in valid JSON documents to this constructor. The constructor does not
+     * perform full validation on construction, but an invalid JsonString will cause
+     * errors later on when it is attempted to be inserted.
```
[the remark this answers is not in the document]
bdeleonardis1: Done

bson/src/main/org/bson/json/JsonString.java
```diff
@@ -17,20 +17,31 @@
...
      *
      * @param json the JSON string
      */
     public JsonString(final String json) {
+        if (json == null) {
+            throw new IllegalArgumentException("Json cannot be null");
+        }
+
+        if (json.charAt(0) != '{' || json.charAt(json.length() - 1) != '}') {
```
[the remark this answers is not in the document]
bdeleonardis1: Addressed

bson/src/main/org/bson/json/JsonString.java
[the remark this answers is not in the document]
bdeleonardis1: Done (renamed to `JsonObject` and implements `Bson`

bson/src/main/org/bson/codecs/BsonCodecProvider.java
```diff
@@ -30,7 +31,9 @@
     @Override
     @SuppressWarnings("unchecked")
     public <T> Codec<T> get(final Class<T> clazz, final CodecRegistry registry) {
-        if (Bson.class.isAssignableFrom(clazz)) {
```
bdeleonardis1, line 33: This change became necessary after making JsonObject implement Bson (because now the BsonCodecProvider thinks its BsonCodec can encode/decode JsonObjects, but that's not what we want. 

This change also allowed me to delete the JsonObjectCodecProvider because BsonCodecProvider is already part of the default registry. I also added the test in MongoCollectionTest.java to ensure the default registry was correctly encoding/decoding JsonObjects.

bson/src/main/org/bson/json/JsonObject.java
```diff
@@ -0,0 +1,96 @@
...
+
+package org.bson.json;
+
+import org.bson.BsonDocument;
+import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.conversions.Bson;
+
+/**
+ * A wrapper class that holds a JSON object string. This class makes decoding JSON efficient.
+ * Note that this class only holds valid JSON objects, not arrays.
```
jyemin, line 25: "not arrays or other values"

bson/src/main/org/bson/json/JsonObject.java
```diff
@@ -0,0 +1,96 @@
...
+     * when it is used later on.
+     *
+     * @param json the JSON object string
+     */
+    public JsonObject(final String json) {
+        if (json == null) {
+            throw new IllegalArgumentException("Json cannot be null");
+        }
+
+        String trimmed = json.trim();
```
jyemin, line 45: While use of `trim()` is convenient, it results in a full copy of the string, which is unnecessarily wasteful.  I suggest instead to write a simple loop to skip over any leading whitespace characters defined in the JSON spec (20, A, D, 9) and ensure the first non-whitespace character is `{`.  I don't think it's worth ensuring that the string ends with `}`.  Since we're not checking the validity of the JSON here, it's not really necessary (if the string doesn't end with `}` it's not valid JSON anyway and will throw a parse exception eventually)

jyemin: Also, technically, the `trim()` method defines whitespace differently from the JSON spec.

bson/src/main/org/bson/codecs/BsonCodecProvider.java
jyemin: You can fix that problem by putting the JsonObjectCodecProvider before the BsonCodecProvider in the list of providers.  That's the expected pattern when defining a registry, and there are plenty of cases already where more than one provider will return a codec for a given class (it's why PojoCodecProvider has to go last).

So while this is convenient, I think it's unexpected to include JsonObject here just because it implements Bson interface.

jyemin (changes requested):

bson/src/main/org/bson/json/JsonObject.java
bdeleonardis1: Done

bson/src/main/org/bson/json/JsonObject.java
bdeleonardis1: Done

bson/src/main/org/bson/codecs/BsonCodecProvider.java
bdeleonardis1: Okay that makes sense. Done

jyemin (changes requested):
Looking good.  I just have one more request for additional code coverage of the new whitespace checks.

bson/src/test/unit/org/bson/json/JsonObjectTest.java
```diff
@@ -37,10 +37,15 @@ public void testArray() {
         new JsonObject("['A', 'B', 'C']");
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    public void testWhitespaceArray() {
+        new JsonObject(" ['A']");
+    }
+
```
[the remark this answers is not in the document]
bdeleonardis1: Done

bson/src/main/org/bson/json/JsonObject.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Constructs a new instance with the given JSON object string. Clients must ensure
+     * they only pass in valid JSON objects to this constructor. The constructor does not
+     * perform full validation on construction, but an invalid JsonObject can cause errors
+     * when it is used later on.
+     *
+     * @param json the JSON object string
+     */
+    public JsonObject(final String json) {
+        if (json == null) {
```
rozza, line 41: We have a `notNull` helper for this - see `org.bson.assertions.Assertions.notNull`

bson/src/main/org/bson/json/JsonObject.java
```diff
@@ -0,0 +1,102 @@
...
+            throw new IllegalArgumentException("Json cannot be null");
+        }
+
+        for (int i = 0; i < json.length(); i++) {
```
rozza, line 45: would `new JsonObject("")` pass this test?

Could you use Assertions.isTrue for this check?

bson/src/main/org/bson/json/JsonObject.java
```diff
@@ -0,0 +1,102 @@
...
+            char c = json.charAt(i);
+            if (c == '{') {
+                break;
+            }
+
+            // Valid whitespace characters according to the json spec: https://www.json.org/json-en.html
+            if (c != 0x0020 && c != 0x000A && c != 0x000D && c != 0x0009) {
```
rozza, line 52: I would use `Character.isWhitespace` as its what is actually used when parsing json strings and is probably more readable.

bson/src/main/org/bson/json/JsonObject.java
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @return the JSON object string
+     */
+    public String getJson() {
+        return json;
+    }
+
+    @Override
+    public <TDocument> BsonDocument toBsonDocument(final Class<TDocument> documentClass, final CodecRegistry registry) {
+        return BsonDocument.parse(json);
```
rozza, line 71: Could you use the registry here? Would using BsonDocumentWrapper be beneficial ? Which approach is better?

bson/src/main/org/bson/json/JsonObject.java
jyemin: I recommended this approach because it matches the JSON spec.  Do you think that's not the behavior of most JSON parsers?

rozza: I suppose, my thoughts were the only parser that actually matters is our own Json reader. Note we don't follow the strictness of the Json Spec, eg we support missing quotes / single quoted values, so thought `Character.isWhitespace` was more readable.

bson/src/main/org/bson/json/JsonObject.java
jyemin: Good catch, Ross. 

Brian, this method will be called any place where an application passes an instance of this class to a method expecting an instance of `Bson`.  So it's important that it's efficient.  Have a look at some other implementations of `Bson` to see how they are implemented.

docs/reference/content/bson/extended-json.md
```diff
@@ -117,31 +117,31 @@ The `Document.parse()` static factory method constructs an instance of a `JsonRe
 equivalent Document. `JsonReader` automatically detects the JSON flavor in the string, so you do not need to specify it.
 
 ### Reading and Writing JSON Directly
-If you do not need a document and only want to deal with JSON, you can use `JsonString` to read and write JSON directly. `JsonString` 
-is simply a wrapper class that takes in a `String` in the constructor and returns the `String` in the get `getJson()` method. 
-Reading and writing directly is more efficient than constructing a `Document` first and then calling `toJson()`, and it is also more efficient than calling `Document#parse`. 
-The codec responsible for reading/writing JSON (`JsonStringCodec`) is part of the default registry, so doing this is very simple 
+If you do not need a document and only want to deal with JSON, you can use `JsonObject` to read and write JSON directly. `JsonObject` 
+is simply a wrapper class that takes in a `String` in the constructor and returns the trimmed `String` in the get `getJson()` method.
```
jyemin: Nit: Change `get getJson()` to `getJson()` here and below.

jyemin (changes requested):

bson/src/main/org/bson/json/JsonObject.java
bdeleonardis1: Done

bson/src/main/org/bson/json/JsonObject.java
bdeleonardis1: Nice catch, fixed and added a test case. I used `Assertions.isTrueArgument` instead because we're just sanity checking the argument so I figured we would want an `IllegalArgumentException` instead.

bson/src/main/org/bson/json/JsonObject.java
bdeleonardis1: Seeing that `Character.isWhitespace` matches our own JsonReader's behavior, I agree with Ross and I changed it. Happy to change it back if you strongly disagree @jyemin .

bson/src/main/org/bson/json/JsonObject.java
bdeleonardis1: Yup thanks. I used `BsonDocumentWrapper` as it is how `Document` and `BasicDBObject` implement this method and my understanding is that it will be most performant given that it doesn't actually parse the JSON until necessary.

docs/reference/content/bson/extended-json.md
bdeleonardis1: Done

bson/src/main/org/bson/json/JsonObject.java
rozza: :+1:

bson/src/main/org/bson/json/JsonObject.java
jyemin: I'm ok with this change.  You can resolve this.

rozza (approved):
LGTM!

bson/src/main/org/bson/json/JsonObject.java
jyemin: Looks good!

jyemin (approved):

jyemin:
LGTM!

bdeleonardis1:
Merged on command line.

PR #573, opened 2020-09-04
Docs: fix descending and 2dsphere index Javadoc

StevenWillyS:
Hello

just some minor doc fix here, since i get a little confused about the descending index javadoc


rozza:
Thanks @StevenWillyS

PR #575, opened 2020-09-10
JAVA-3677: Add BsonRepresentation to allow ObjectId to string convers

bdeleonardis1:
Evergreen patch: https://spruce.mongodb.com/version/5f5a7c4557e85a3bd2b99b2a/tasks


bson/src/main/org/bson/codecs/FlexibleCodec.java
```diff
@@ -0,0 +1,37 @@
...
+import org.bson.BsonType;
+
+/**
+ * Implementations of this interface can decode additional types
+ * and translate them to the desired value type depending on the BsonRepresentation.
+ *
+ * @param <T> the value type
+ * @since 4.2
+ */
+public interface FlexibleCodec<T> extends Codec<T> {
```
bdeleonardis1, line 28: The motivation for this design is that we need a way to tell Codecs what other type they can read, aside from what they would do normally (we don't want String codecs reading ObjectIds normally)... There's no way to do this currently, so I decided to define a new interface that has a method that allows you to specify what other type it can read.

bson/src/main/org/bson/codecs/FlexibleCodecProvider.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A {@code CodecProvider} for all codecs that can read in different values depending on
+ * the BsonRepresentation.
+ *
+ * @since 4.2
+ */
+public class FlexibleCodecProvider implements CodecProvider {
```
bdeleonardis1, line 31: This class is a little more complicated than it needs to be, but it copies the style of `BsonValueCodecProvider` because I'm assuming we're going to want to add support for additional BsonRepresentations in the future which will necessitate more FlexibleCodecs

bson/src/main/org/bson/codecs/FlexibleStringCodec.java
```diff
@@ -0,0 +1,74 @@
...
+import org.bson.codecs.configuration.CodecConfigurationException;
+import org.bson.types.ObjectId;
+
+/**
+ * A codec capable of decoding different types to Strings and encoding
+ * Strings to different types depending on the provided BsonRepresentation.
+ *
+ * @since 4.2
+ */
+public class FlexibleStringCodec extends StringCodec implements FlexibleCodec<String> {
```
jyemin, line 32: I'd like to consider a different design for handling `BsonRepresentation` that works within the existing codec registry system.  Instead of creating a new codec, and a new provider, what if we made `StringCodec` implement `FlexibleCodec` itself?  That way, the existing codec for string is responsible for determining whether it can handle alternate representations.  This will allow application-defined codecs to participate in this system, and make it easier to add support to more of the built-in codecs.

To make this work though, we have to preserve immutability of codec instances.  So I imagine the interface looking something like this:

    public interface RepresentationConfigurable<T> {
         // get the current representation
         BsonType getRepresentation();

         // Return a codec with the given representation
         // Be careful to preserve immutability!
         // Throw an exception if the requested representation is not supported
         Codec<T> withRepresentation(BsonType representation)
    }

The POJO codec will do an `instanceof RepresentationConfigurable` on the codec for the field with a BsonRepresentation annotation to determine whether the field can participate in this scheme.

Let me know what you think of this alternative.

jyemin (commented):
It's looking good, Brian. 

I wanted to focus on the design first before proceeding further with the review, so I posted an initial comment to discuss it.

bson/src/main/org/bson/codecs/FlexibleStringCodec.java
bdeleonardis1: Yup this design sounds great to me. I had considered doing something like this but I wasn't sure if doing the `instanceof` check was considered bad. I hadn't thought of the fact that it will allow application-defined codecs to participate, that's a good point. It'll also just make the code simpler and more concise, so I'll get started on changing it.

rozza:
I'll hold off reviewing until the design gets updated but I just wanted to say I really like the direction this is going - good work @bdeleonardis1

bson/src/main/org/bson/codecs/StringCodec.java
```diff
@@ -16,27 +16,74 @@
...
-        writer.writeString(value);
+        switch (representation) {
+            case STRING:
+                writer.writeString(value);
+                break;
+            case OBJECT_ID:
+                writer.writeObjectId(new ObjectId(value));
+                break;
+            default:
+                throw new BsonInvalidOperationException("Cannot encode a String to a " + representation);
```
bdeleonardis1, line 70: This line is pretty much unreachable now (because the only representations `StringCodec` allows to be created are  `STRING` and `OBJECT_ID`. Is this kind of defensive programming good or unnecessary?

bson/src/main/org/bson/codecs/pojo/ConventionAnnotationImpl.java
```diff
@@ -100,6 +111,14 @@ private void processPropertyAnnotations(final ClassModelBuilder<?> classModelBui
         }
     }
 
+    private <T> boolean  bsonRepIsValid(final BsonType bsonRep, final Class<T> clazz) {
```
bdeleonardis1, line 114: Even though we're throwing exceptions in the `Codec` constructor now, I decided to leave this check in there by the principle of catching errors as soon as we can. Is that good or unnecessary?

bson/src/main/org/bson/codecs/RepresentationConfigurable.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Returns an immutable codec with the given representation. If the provided representation
+     * is not supported an exception will be thrown.
+     *
+     * @param representation the BsonRepresentation.
+     * @return a new Codec with the correct representation.
+     * @throws IllegalArgumentException
+     */
+    Codec<T> withRepresentation(BsonType representation);
```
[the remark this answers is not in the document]
jyemin: I support this decision.

bson/src/main/org/bson/codecs/StringCodec.java
jyemin: I think this is good defensive programming, since a change in a different method could break the invariants here, so better to be defensive.  And it would be bad to silently do nothing in this method if somehow the type was unsupported.

bson/src/main/org/bson/codecs/pojo/annotations/BsonRepresentation.java
```diff
@@ -0,0 +1,45 @@
...
+ * An annotation that specifies what type the property is stored as in the database.
+ *
+ * <p>Note: Requires the {@link org.bson.codecs.pojo.Conventions#ANNOTATION_CONVENTION}</p>
+ *
+ * @since 4.2
+ * @see org.bson.codecs.pojo.Conventions#ANNOTATION_CONVENTION
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
```
jyemin, line 37: Should we be targeting anything besides FIELD here?  I notice that some of the other annotations target METHOD, for example.

jyemin (commented):
Just some initial comments on the new design.

bson/src/main/org/bson/codecs/pojo/annotations/BsonRepresentation.java
bdeleonardis1: We could definitely target METHOD. Imo that is the only other `ElementType` that would make any sense to target. I chose not to because of this in our docs:

"Annotations can be applied to read and / or write contexts by configuring the getter / setter methods. Any annotations applied to a field are applied to both read and write contexts."

`BsonRepresentation` applies to both reading and writing, so I didn't want to let users only annotate a getter or setter and have the impression that it only affected reading/writing. Also to me it just seems cleaner to annotate the field as opposed to the getter/setter if you're not trying to distinguish between reading and writing. Do you agree with that rationale, or do you still think we should allow it to target methods as well?

jyemin: @rozza, interested in your opinion here.  I'm not sure why BsonProperty, for example, applies to field, method, and parameter, and what happens if there is disagreement between them.  Maybe it's an error to add it to more than one place.

rozza: If field naming doesn't match the convention then setting on getters and setters would be the only way for a user to apply the annotation equally.  So I would add to METHOD as well.

BsonProperty's are special and can also be applied to parameters - when used in constructors, it allows the mapping of arguments to properties.  There can be conflicting read and write properties as models don't necessarily have to be symmetrical (See the `AsymmetricalModel` examples).

bson/src/main/org/bson/codecs/pojo/ConventionAnnotationImpl.java
rozza: Can users write their own RepresentationConfigurable codecs? If so can we validate them here?

bson/src/main/org/bson/codecs/pojo/annotations/BsonRepresentation.java
bdeleonardis1: Okay sounds good I'll add it.

bson/src/main/org/bson/codecs/pojo/ConventionAnnotationImpl.java
```diff
@@ -85,6 +87,15 @@ private void processPropertyAnnotations(final ClassModelBuilder<?> classModelBui
                 classModelBuilder.idPropertyName(propertyModelBuilder.getName());
             } else if (annotation instanceof BsonIgnore) {
                 propertyModelBuilder.readName(null);
+            } else if (annotation instanceof BsonRepresentation) {
```
bdeleonardis1, line 90: I decided to only check for `BsonRepresentation` in the read annotations, because I felt that there was no reason users would want to only put the tag on the setter method. This annotation feels kind of similar to `BsonId` to me which is also not checked for in the write annotations. The properties that are checked for in both do different things depending on the context (i.e. change the `readName` or the `writeName`). Because `BsonRepresentation` should do the same thing in both contexts, I thought it made sense to only allow it (and check for it) in one place.

bson/src/main/org/bson/codecs/pojo/ConventionAnnotationImpl.java
rozza: @bdeleonardis1 Have you thought about what to do here?

bdeleonardis1: Sorry @rozza I missed this question the first time. If users can write their own `RepresentationConfigurable` codecs I don't think we can check for validity here because there is no way of predicting what types of conversions these codecs will allow and we don't have even have access to the Codec yet at this point in the execution. I think the best option is to just let codecs do error checking as they'red constructed (i.e. in the call to withRepresentation). If they're provided with an invalid BsonRep in that method call, they can throw an exception before ever being created.

docs/reference/content/bson/pojos.md
```diff
@@ -389,7 +393,7 @@ public final class Person {
 The `Person` POJO Will produce BSON similar to:
 
 ```json
-{ "_id": "1234567890", "_t": "Person", "firstName": "Alan", "surname": "Turing",
+{ "_id": "ObjectId(111111111111111111111111)", "_t": "Person", "firstName": "Alan", "surname": "Turing",
```
rozza, line 396: Wouldn't it just store the hex?

bson/src/main/org/bson/codecs/pojo/ConventionAnnotationImpl.java
rozza: Users often may want to do their own thing and with representations you've done all the groundwork to make it extensible. 

I'd remove this check here and add it later when fetching the codec.  You may also want to document how to create a custom BsonRepresentation (@jyemin do you think its worth it?)

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -266,9 +267,15 @@ private void decodeProperties(final BsonReader reader, final DecoderContext deco
         return codec;
     }
 
+    @SuppressWarnings("unchecked")
     private <S> Codec<S> getCodecFromPropertyRegistry(final PropertyModel<S> propertyModel) {
         try {
-            return propertyCodecRegistry.get(propertyModel.getTypeData());
+            Codec<S> codec = propertyCodecRegistry.get(propertyModel.getTypeData());
+            BsonType representation = propertyModel.getBsonRepresentation();
+            if (codec instanceof RepresentationConfigurable && representation != null) {
```
rozza, line 275: This could be a good place to check for user bsonRepresentation errors?

rozza (changes requested):

docs/reference/content/bson/pojos.md
jyemin: I think it amounts to the same time, this is just using shell syntax.  But better to use standard extended JSON syntax in the docs, IMO.

bson/src/main/org/bson/codecs/pojo/ConventionAnnotationImpl.java
bdeleonardis1: @rozza I think I'm a little confused. I thought we were just trying to allow users to to create Codecs that implement `RepresentationConfigurable` so that they can create any conversions that they want.

Or are you saying we want to allow them to create a custom BsonRepresentation that just works with our built in Codecs (for example `@BsonRepresentation(BsonType.INT32)` on a String and then our `StringCodec` handles it)? 

If it's the first one, I don't see why we can't check for validity in `withRepresentation`. Our built in codecs would know all of the conversions that they support, so they can just check them on construction. That would prevent us from having to do type checking/casting where you highlighted in the code below.

If it's the second one, that seems a lot harder. We would need to add support to each of our codecs for all possible conversions, right?

rozza: Sorry, I wasn't very clear.  I can think of two main error scenarios:

1) If a user added `@BsonRepresentation(BsonType.INT32)` and the codec is `RepresentationConfigurable` then the codec should validate the BsonType.
2) If a user adds `@BsonRepresentation(BsonType.STRING)` and the resulting codec isn't `RepresentationConfigurable` then the PojoCodec should error.

bdeleonardis1: Ah I see what you're saying now. So I'll have checks in `withRepresentation` that will handle your case 1, and an additional case in the place that you highlighted that will handle your case 2. Thanks for clarifying for me.

bdeleonardis1: Done. I already had test cases for each of these cases in `PojoCustomTest.java` so I didn't add any cases. I did change the exception being thrown (and expected in the tests) to `IllegalArgumentException` though because `CodecConfigurationException` gets caught in `getCodecFromPropertyRegistry` and we want in thrown.

docs/reference/content/bson/pojos.md
bdeleonardis1: I changed to extended JSON with "$oid" as per this: https://docs.mongodb.com/manual/reference/mongodb-extended-json/#bson.ObjectId

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
bdeleonardis1: Yep added a check here to throw errors if a BsonRepresentation is provided but the codec isn't RepresentationConfigurable. The rest of the checks are done in the `withRepresentation` methods of the codecs that are RepresentationConfigurable.

bson/src/main/org/bson/codecs/RepresentationConfigurable.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    BsonType getRepresentation();
+
+    /**
+     * Returns an immutable codec with the given representation. If the provided representation
+     * is not supported an exception will be thrown.
+     *
+     * @param representation the BsonRepresentation.
+     * @return a new Codec with the correct representation.
+     * @throws IllegalArgumentException
```
jyemin, line 43: I can't recall an interface specification where IllegalArgumentException could be thrown by different implementors for different parameter values.  I think it makes more sense to throw CodecConfigurationException instead of IllegalArgumentException for unsupported representations.

bson/src/main/org/bson/codecs/pojo/ConventionObjectIdGeneratorsImpl.java
```diff
@@ -30,6 +30,8 @@ public void apply(final ClassModelBuilder<?> classModelBuilder) {
                    classModelBuilder.idGenerator(IdGenerators.OBJECT_ID_GENERATOR);
                } else if (classModelBuilder.getIdGenerator() == null && idType.equals(BsonObjectId.class)) {
                    classModelBuilder.idGenerator(IdGenerators.BSON_OBJECT_ID_GENERATOR);
+               } else if (classModelBuilder.getIdGenerator() == null && idType.equals(String.class)) {
```
jyemin, line 33: Wouldn't this only be appropriate if the bson representation of this string property is ObjectId?

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -266,9 +267,18 @@ private void decodeProperties(final BsonReader reader, final DecoderContext deco
...
 
+    @SuppressWarnings("unchecked")
     private <S> Codec<S> getCodecFromPropertyRegistry(final PropertyModel<S> propertyModel) {
         try {
-            return propertyCodecRegistry.get(propertyModel.getTypeData());
+            Codec<S> codec = propertyCodecRegistry.get(propertyModel.getTypeData());
+            BsonType representation = propertyModel.getBsonRepresentation();
+            if (representation != null) {
+                if (codec instanceof RepresentationConfigurable) {
+                    return ((RepresentationConfigurable<S>) codec).withRepresentation(representation);
```
jyemin, line 277: This cast doesn't seem necessary, and based on your comment above, not necessarily safe.

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -266,9 +267,18 @@ private void decodeProperties(final BsonReader reader, final DecoderContext deco
...
+                }
+                throw new IllegalArgumentException("Codec must implement RepresentationConfigurable to support BsonRepresentation");
```
jyemin, line 279: Why not CodecConfigurationException?

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
jyemin, line 277: I'm not clear why this call to withRepresentation couldn't be done earlier, like in PropertyModelBuilder#build, instead of delayed until this point.

jyemin (changes requested):
I've done a more thorough review of the implementation now that the design has been agreed to.

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
bdeleonardis1: This method is wrapped in a try-catch that catches CodecConfigurationExceptions. I can easily refactor the method, but I originally didn't think which error type we threw made a big difference, but your comment above made sense.

The catch is only really intended for that `propertyCodecRegistry.get` call anyway, so it's probably better to refactor anyway.

bson/src/main/org/bson/codecs/RepresentationConfigurable.java
bdeleonardis1: Okay that makes sense

bson/src/main/org/bson/codecs/pojo/ConventionObjectIdGeneratorsImpl.java
bdeleonardis1: Yep, I think you're right.

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
bdeleonardis1: Unless I'm missing something, it's both necessary and safe. The type of `codec` is `Codec` which doesn't necessarily have a `withRepresentation` method, so we need to cast it as `RepresentationConfigurable` before we can call `withRepresentation`. It should be safe because we're checking if `codec` is a `RepresentatoinConfigurable` in the if statement wrapping this line. 

Note that I'm casting `codec`, the caller, not the return value of the call.

bdeleonardis1: This is actually the first time we get the `codec` we need as far as I can tell (line 273, the call to propertyCodecRegistry#get is where we get it). The codec is always null in PropertyModelBuilder#build as far as I can tell. I believe this is the intended behavior, as we only get the codec when we need it the first time, and then it is cached.

jyemin: Whoops, my bad!  I guess this is why they are adding the new pattern matching syntax in Java 15, which will allow you to do:

    if (codec instanceof RepresentationConfigurable rc) {
        rc.withRepresentation(representation);
    }

jyemin: OK.

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
bdeleonardis1: Done

bson/src/main/org/bson/codecs/pojo/ConventionObjectIdGeneratorsImpl.java
bdeleonardis1: Done. I also noticed nothing was really testing this code, so I added a test in `ConvenvtionsTest.java` that tests both `else if` conditions. The first `if` condition is tested in the first test in `ConventionsTest`.

bson/src/main/org/bson/codecs/RepresentationConfigurable.java
bdeleonardis1: Done

bson/src/main/org/bson/codecs/pojo/ConventionObjectIdGeneratorsImpl.java
```diff
@@ -30,6 +31,9 @@ public void apply(final ClassModelBuilder<?> classModelBuilder) {
                    classModelBuilder.idGenerator(IdGenerators.OBJECT_ID_GENERATOR);
                } else if (classModelBuilder.getIdGenerator() == null && idType.equals(BsonObjectId.class)) {
                    classModelBuilder.idGenerator(IdGenerators.BSON_OBJECT_ID_GENERATOR);
+               } else if (classModelBuilder.getIdGenerator() == null && idType.equals(String.class)
```
jyemin, line 34: This seems like a design flaw.  Elsewhere (i.e. PojoCodecImpl) we do not make assumptions about which Codecs support changing BSON representation, but now here we are assuming that whatever Codec happens to be registered for String.class does support it (it's true out default one does, but it could be changed by the application), and no others are (a user could register their own for some other type).

But given that we defer Codec resolution until first use, I don't see a way out of this.  

Any ideas?

bdeleonardis1: I don't really see why this is a problem... This condition will only be true if the user has explicitly marked their String id field with `@BsonRepresentation(BsonType.ObjectId)`. By doing this, they are implying that whatever codec they're using for `Strings` supports changing the BSON rep. 

If they use the annotation and their codec does not support changing the BSON rep, an exception will be thrown anyway when the String codec is resolved the first time.

jyemin: > we are assuming that whatever Codec happens to be registered for String.class does support it

I see your point.  This isn't really a problem

> and no others are

I was thinking of an application-defined class, e.g. 

     public class Guid {
           private final String guid;
     }

but I don't see how we could create an Id generator that populates this anyway.  The application would have to register it as a convention explicitly.

So I think this is ok.  False alarm.

jyemin (approved):

bson/src/main/org/bson/codecs/pojo/PropertyModel.java
```diff
@@ -113,6 +116,13 @@ public boolean isReadable() {
         return codec;
     }
 
+    /**
+     * @return the BsonRepresentation of the field
+     */
```
rozza, line 121: Add `@since 4.2` to the javadoc

bson/src/main/org/bson/codecs/pojo/PropertyModelBuilder.java
```diff
@@ -229,6 +231,26 @@ public Boolean isDiscriminatorEnabled() {
         return this;
     }
 
+    /**
+     * Returns the BsonRepresentation
+     *
+     * @return the BsonRepresentation
+     */
+    public BsonType getBsonRepresentation() {
```
rozza, line 239: Add `@since 4.2` to the javadoc

bson/src/main/org/bson/codecs/pojo/PropertyModelBuilder.java
```diff
@@ -229,6 +231,26 @@ public Boolean isDiscriminatorEnabled() {
...
+        return bsonRepresentation;
+    }
+
+    /**
+     * Sets the BsonRepresentation
+     *
+     * @param bsonRepresentation the BsonRepresentation
+     * @return this
+     */
+    public PropertyModelBuilder<T> bsonRepresentation(final BsonType bsonRepresentation) {
```
rozza, line 249: Add `@since 4.2` to the javadoc

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
```diff
@@ -595,6 +597,22 @@ public void testInvalidGetterAndSetterModelDecoding() {
...
+    @Test(expected = CodecConfigurationException.class)
+    public void testInvalidBsonRepresentationStringEncoding() {
+        encodesTo(getPojoCodecProviderBuilder(BsonRepresentationUnsupportedString.class),
+                new BsonRepresentationUnsupportedString("1"), "");
+    }
+
+    @Test(expected = CodecConfigurationException.class)
+    public void testInvalidBsonRepresentationIntDecoding() {
+        decodingShouldFail(getCodec(BsonRepresentationUnsupportedInt.class), "{'id': 'hello', s: '3'}");
+    }
```
rozza, line 614: I think you can add a failure test for an invalid BsonRepresentationModel instance as well - where the string value cannot be converted to an ObjectId.

bson/src/test/unit/org/bson/codecs/pojo/PojoRoundTripTest.java
```diff
@@ -507,6 +508,10 @@ public void test() {
                 getPojoCodecProviderBuilder(MapListGenericExtendedModel.class),
                 "{values: {a: [1, 2, 3], b: [4, 5, 6]}}"));
 
+        data.add(new TestData("BsonRepresentation is encoded and decoded correctly", new BsonRepresentationModel(1),
```
rozza, line 511: Also add a roundtrip test without the default conventions applied.

rozza (changes requested):
Looks really good, a couple of javadoc nits and I think theres two extra tests we can add that will help catch any future regressions.

bson/src/main/org/bson/codecs/pojo/PropertyModel.java
bdeleonardis1: Done

bson/src/main/org/bson/codecs/pojo/PropertyModelBuilder.java
bdeleonardis1: Done

bson/src/main/org/bson/codecs/pojo/PropertyModelBuilder.java
bdeleonardis1: Done

bson/src/test/unit/org/bson/codecs/pojo/PojoRoundTripTest.java
bdeleonardis1: Done, but I had to add it to `PojoCustomTest`, because `PojoRoundTripTest` does tests with the "AUTOMATIC_BUILDER" and "PACKAGE_BUILDER" which both use default conventions.

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
bdeleonardis1: Done. It's throwing an IllegalArgumentException in this case now. I can easily change that, but I couldn't think of a more appropriate error message. CodecConfigurationException doesn't sound right because the codec could be configured fine, it's just that the data doesn't match it.

rozza (approved):
LGTM!

bdeleonardis1:
Rebased and merged on command line.

PR #576, opened 2020-09-11
JAVA-3839

jyemin:
@rozza this is preliminary work as the Scala build fails when running against Java 14.  See https://jira.mongodb.org/browse/JAVA-3839 for the exception I get.  I found https://github.com/sbt/sbt/issues/5093 but I'm not sure what, if anything, can be updated in our build script to take advantage of the fix.  Can you take a look?


build.gradle
```diff
@@ -311,22 +311,24 @@ configure(javaCodeCheckedProjects) {
...
     }
 
     codenarc {
-        toolVersion = '1.1'
+        toolVersion = '1.6.1'
         reportFormat = project.buildingWith('xmlReports.enabled') ? 'xml' : 'html'
     }
 
-    tasks.withType(com.github.spotbugs.SpotBugsTask) {
+    spotbugsMain {
```
jyemin, line 321: It complained without this change

build.gradle
```diff
@@ -311,22 +311,24 @@ configure(javaCodeCheckedProjects) {
...
         reports {
             xml.enabled = project.buildingWith('xmlReports.enabled')
             html.enabled = !project.buildingWith('xmlReports.enabled')
         }
     }
 
+    spotbugsTest {
```
jyemin, line 328: It complained about the use of sourceSets up above so using this as an alternative.

config/codenarc/codenarc.xml
```diff
@@ -55,6 +55,13 @@
         <exclude name='MethodParameterTypeRequired'/>
         <exclude name='FieldTypeRequired '/>
         <exclude name="VariableTypeRequired"/>
+        <exclude name="CompileStatic"/>
```
jyemin, line 58: Excluding any new rules added since 1.1

build.gradle
```diff
@@ -26,7 +26,7 @@ buildscript {
     }
     dependencies {
         classpath 'com.netflix.nebula:gradle-extra-configurations-plugin:3.0.3'
-        classpath "com.github.spotbugs:spotbugs-gradle-plugin:3.0.0"
+        classpath "gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.5.0"
```
jyemin, line 29: Not sure why the name change, but that's what it is now.

rozza: Ack

build.gradle
rozza: Ack

build.gradle
rozza: Ack

config/codenarc/codenarc.xml
rozza: Ack

rozza (changes requested):
Update `gradle.properties`:

```
scalaVersions=2.11.12,2.12.12,2.13.3
defaultScalaVersions=2.13.3
```

See: https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#version-compatibility-table

You'll have to edit the publishing the scala snapshots / archives and set the jdk to 11 for the 2.11 & 2.12 versions of scala.

rozza (commented):
Typo: Should be `2.13.3`

rozza (approved):
LGTM!

jyemin:
Added one more update to fix an issue when using Java 15.

rozza (approved):

PR #580, opened 2020-09-23
Fix scala publishing

rozza:
Use the latest scala multiversion plugin
Split out the publishing steps

JAVA-3843

Patch: https://spruce.mongodb.com/version/5f6b310a1e2d17450be4b41e/tasks
Snapshots: https://oss.sonatype.org/content/repositories/snapshots/org/mongodb/scala/

### DIFF
--- a/.evergreen/publish.sh
+++ b/.evergreen/publish.sh
@@ -34,4 +34,5 @@ SYSTEM_PROPERTIES="-Dorg.gradle.internal.publish.checksums.insecure=true -Dorg.g
 
 ./gradlew -version
 ./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info  ${TASK}
-./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info :bson-scala:${TASK} :driver-scala:${TASK} -PdefaultScalaVersions=2.11.12,2.12.10
+./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info :bson-scala:${TASK} :driver-scala:${TASK} -PdefaultScalaVersions=2.12.12
+./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info :bson-scala:${TASK} :driver-scala:${TASK} -PdefaultScalaVersions=2.11.12
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ buildscript {
         classpath 'biz.aQute.bnd:biz.aQute.bnd.gradle:5.1.2'
         
         // Scala plugins
-        classpath "com.adtran:scala-multiversion-plugin:1.+"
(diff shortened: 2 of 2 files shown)


.evergreen/publish.sh
```diff
@@ -34,4 +34,5 @@ SYSTEM_PROPERTIES="-Dorg.gradle.internal.publish.checksums.insecure=true -Dorg.g
 
 ./gradlew -version
 ./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info  ${TASK}
-./gradlew ${SYSTEM_PROPERTIES} --stacktrace --info :bson-scala:${TASK} :driver-scala:${TASK} -PdefaultScalaVersions=2.11.12,2.12.10
```
rozza, line 37: I had to split out builds / maybe a change in gradle when we last upgraded, that the multiversion plugin didnt handle.

build.gradle
```diff
@@ -30,7 +30,7 @@ buildscript {
         classpath 'biz.aQute.bnd:biz.aQute.bnd.gradle:5.1.2'
         
         // Scala plugins
-        classpath "com.adtran:scala-multiversion-plugin:1.0.35"
+        classpath "com.adtran:scala-multiversion-plugin:1.+"
```
rozza, line 33: Updated the classpath as per: https://github.com/ADTRAN/gradle-scala-multiversion-plugin#scala-multi-version-plugin

jyemin: I'm not sure why to do this.  Wouldn't it preferable to have repeatable builds by continuing to pin to a specific version?

.evergreen/publish.sh
jyemin: Acknowledged, but kind of weird since I thought the point of this plugin was to allow you to do it the way it is now.

build.gradle
rozza: I think its to catch any bug fixes in the plugin, especially with regards with newer versions of gradle.

rozza: Fixed to latest version.

.evergreen/publish.sh
rozza: Yeah, seems like a bug. Ideally I'd like to know enough gradle to write a stable multi version plugin that allows all versions to be published in one shot, as well as allowing different source routes.

build.gradle
jyemin: Weird that at https://github.com/ADTRAN/gradle-scala-multiversion-plugin/releases still shows 1.0.35 as the latest release.

rozza: https://plugins.gradle.org/plugin/com.adtran.scala-multiversion-plugin

jyemin: Yeah, I wonder why the discrepancy. Anyway, no worries.

jyemin (approved):
LGTM

PR #581, opened 2020-09-24
JAVA-3567: Add support for records

bdeleonardis1:
Leaving as a draft for the following reasons:
1. Testing is not finished. I was unable to get Junit running successfully in preview mode so I wrote my own little test runner. The tests I wrote make me confident that my implementation is correct, but we're going to want to add more coverage before merging this. All of the tests I have written so far are integration tests. I did not write a full suite of tests because I didn't want to waste time writing them in one format, when they'll just need to be converted to Junit tests before this gets merged anyway.
2. Annotations are not supported. I was having a problem getting the ElementType.RecordComponent to work because of preview mode. If we are able to use ElementType.RecordComponent it will be very easy to support annotations, but if this causes a Java 8 compatibility problem we'll have to think harder about how to do that.
3. There is a bunch of junk in this diff that I needed to do to get preview mode to work that needs to be deleted before it is merged.

I think it makes the most sense to hold off on fixing these problems until records are generally available instead of preview mode because it will save a lot of time and configuration headaches.

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/ClassModelBuilder.java
+++ b/bson/src/main/org/bson/codecs/pojo/ClassModelBuilder.java
@@ -55,6 +55,7 @@ public class ClassModelBuilder<T> {
     private String discriminator;
     private String discriminatorKey;
     private String idPropertyName;
+    private boolean isRecord;
 
     ClassModelBuilder(final Class<T> type) {
         configureClassModelBuilder(this, notNull("type", type));
@@ -234,6 +235,23 @@ public class ClassModelBuilder<T> {
      */
     public boolean removeProperty(final String propertyName) {
         return propertyModelBuilders.remove(getProperty(notNull("propertyName", propertyName)));
+    }
+
+    /**
+     * @return returns true if the ClassModelBuilder represents a record
+     */
+    public boolean getIsRecord() {
+        return isRecord;
+    }
+
+    /**
+     * Sets whether or not the ClassModelBuilder represents a record
+     * @param isRecord whether or not the ClassModelBuilder represents a record
+     * @return this
+     */
+    public ClassModelBuilder<T> isRecord(boolean isRecord) {
+        this.isRecord = isRecord;
+        return this;
     }
 
     /**
--- a/bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
+++ b/bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
@@ -21,129 +21,43 @@ import org.bson.codecs.pojo.annotations.BsonId;
 import org.bson.codecs.pojo.annotations.BsonProperty;
 
 import java.lang.annotation.Annotation;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Method;
-import java.lang.reflect.Type;
-import java.util.ArrayList;
 import java.util.List;
+import java.lang.reflect.Type;
 
-import static java.lang.String.format;
-import static java.util.Arrays.asList;
+interface CreatorExecutable<T> {
+    Class<T> getType();
 
-final class CreatorExecutable<T> {
-    private final Class<T> clazz;
-    private final Constructor<T> constructor;
-    private final Method method;
-    private final List<BsonProperty> properties = new ArrayList<BsonProperty>();
-    private final Integer idPropertyIndex;
-    private final List<Class<?>> parameterTypes = new ArrayList<Class<?>>();
-    private final List<Type> parameterGenericTypes = new ArrayList<Type>();
+    List<String> getProperties();
 
-    CreatorExecutable(final Class<T> clazz, final Constructor<T> constructor) {
-        this(clazz, constructor, null);
-    }
+    Integer getIdPropertyIndex();
 
-    CreatorExecutable(final Class<T> clazz, final Method method) {
-        this(clazz, null, method);
-    }
+    List<Class<?>> getParameterTypes();
 
-    private CreatorExecutable(final Class<T> clazz, final Constructor<T> constructor, final Method method) {
-        this.clazz = clazz;
-        this.constructor = constructor;
-        this.method = method;
+    List<Type> getParameterGenericTypes();
+
+    T getInstance(Object... params);
+
+    CodecConfigurationException getError(Class<?> clazz, String msg);
+
+    static Integer initializeDataFromAnnotations(Annotation[][] parameterAnnotations, List<String> properties) {
         Integer idPropertyIndex = null;
+        for (int i = 0; i < parameterAnnotations.length; ++i) {
+            Annotation[] parameterAnnotation = parameterAnnotations[i];
 
-        if (constructor != null || method != null) {
-            Class<?>[] paramTypes = constructor != null ? constructor.getParameterTypes() : method.getParameterTypes();
-            Type[] genericParamTypes = constructor != null ? constructor.getGenericParameterTypes() : method.getGenericParameterTypes();
-            parameterTypes.addAll(asList(paramTypes));
-            parameterGenericTypes.addAll(asList(genericParamTypes));
-            Annotation[][] parameterAnnotations = constructor != null ? constructor.getParameterAnnotations()
-                    : method.getParameterAnnotations();
+            for (Annotation annotation : parameterAnnotation) {
+                if (annotation.annotationType().equals(BsonProperty.class)) {
+                    properties.add(((BsonProperty) annotation).value());
+                    break;
+                }
 
-            for (int i = 0; i < parameterAnnotations.length; ++i) {
-                Annotation[] parameterAnnotation = parameterAnnotations[i];
-
-                for (Annotation annotation : parameterAnnotation) {
-                    if (annotation.annotationType().equals(BsonProperty.class)) {
-                        properties.add((BsonProperty) annotation);
-                        break;
-                    }
-
-                    if (annotation.annotationType().equals(BsonId.class)) {
-                        properties.add(null);
-                        idPropertyIndex = i;
-                        break;
-                    }
+                if (annotation.annotationType().equals(BsonId.class)) {
+                    properties.add(null);
+                    idPropertyIndex = i;
+                    break;
                 }
             }
         }
-
-        this.idPropertyIndex = idPropertyIndex;
-    }
-
-    Class<T> getType() {
-        return clazz;
-    }
-
-    List<BsonProperty> getProperties() {
-        return properties;
-    }
-
-    Integer getIdPropertyIndex() {
         return idPropertyIndex;
-    }
-
-    List<Class<?>> getParameterTypes() {
-        return parameterTypes;
-    }
-
-    List<Type> getParameterGenericTypes() {
-        return parameterGenericTypes;
-    }
-
-    @SuppressWarnings("unchecked")
-    T getInstance() {
-        checkHasAnExecutable();
-        try {
-            if (constructor != null) {
-                return constructor.newInstance();
-            } else {
-                return (T) method.invoke(clazz);
-            }
-        } catch (Exception e) {
-            throw new CodecConfigurationException(e.getMessage(), e);
-        }
-    }
-
-    @SuppressWarnings("unchecked")
-    T getInstance(final Object[] params) {
-        checkHasAnExecutable();
-        try {
-            if (constructor != null) {
-                return constructor.newInstance(params);
-            } else {
-                return (T) method.invoke(clazz, params);
-            }
-        } catch (Exception e) {
-            throw new CodecConfigurationException(e.getMessage(), e);
-        }
-    }
-
-
-    CodecConfigurationException getError(final Class<?> clazz, final String msg) {
-        return getError(clazz, constructor != null, msg);
-    }
-
-    private void checkHasAnExecutable() {
-        if (constructor == null && method == null) {
-            throw new CodecConfigurationException(format("Cannot find a public constructor for '%s'.", clazz.getSimpleName()));
-        }
-    }
-
-    private static CodecConfigurationException getError(final Class<?> clazz, final boolean isConstructor, final String msg) {
-        return new CodecConfigurationException(format("Invalid @BsonCreator %s in %s. %s", isConstructor ? "constructor" : "method",
-                clazz.getSimpleName(), msg));
     }
 
 }
--- a/bson/src/main/org/bson/codecs/pojo/MethodCreatorExecutable.java
+++ b/bson/src/main/org/bson/codecs/pojo/MethodCreatorExecutable.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bson.codecs.pojo;
+
+import org.bson.codecs.configuration.CodecConfigurationException;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+
+final class MethodCreatorExecutable<T> implements CreatorExecutable<T> {
+    private final Class<T> clazz;
+    private final Method method;
+    private final List<String> properties = new ArrayList<>();
+    private final Integer idPropertyIndex;
+    private final List<Class<?>> parameterTypes = new ArrayList<Class<?>>();
+    private final List<Type> parameterGenericTypes = new ArrayList<Type>();
+
+    MethodCreatorExecutable(final Class<T> clazz, final Method method) {
+        this.clazz = clazz;
+        this.method = method;
+        Integer idPropertyIndex = null;
+
+        if (method != null) {
+            Class<?>[] paramTypes = method.getParameterTypes();
+            Type[] genericParamTypes = method.getGenericParameterTypes();
+            parameterTypes.addAll(asList(paramTypes));
+            parameterGenericTypes.addAll(asList(genericParamTypes));
+
+            Annotation[][] parameterAnnotations = method.getParameterAnnotations();
+            CreatorExecutable.initializeDataFromAnnotations(parameterAnnotations, properties);
+        }
+
+        this.idPropertyIndex = idPropertyIndex;
+    }
+
+    @Override
+    public Class<T> getType() {
+        return clazz;
+    }
+
+    @Override
+    public List<String> getProperties() {
+        return properties;
+    }
+
+    @Override
+    public Integer getIdPropertyIndex() {
+        return idPropertyIndex;
+    }
+
+    @Override
+    public List<Class<?>> getParameterTypes() {
+        return parameterTypes;
+    }
+
+    @Override
+    public List<Type> getParameterGenericTypes() {
+        return parameterGenericTypes;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public T getInstance(final Object... params) {
+        checkHasAnExecutable();
+        try {
+            return (T) method.invoke(clazz, params);
+        } catch (Exception e) {
+            throw new CodecConfigurationException(e.getMessage(), e);
+        }
+    }
+
+    @Override
+    public CodecConfigurationException getError(final Class<?> clazz, final String msg) {
+        return new CodecConfigurationException(format("Invalid @BsonCreator method in %s. %s",
+                clazz.getSimpleName(), msg));
+    }
+
+    private void checkHasAnExecutable() {
+        if (method == null) {
+            throw new CodecConfigurationException(format("Cannot find a public method to construct"
+                   + " the object with for '%s'.", clazz.getSimpleName()));
+        }
+    }
+}
--- a/bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
+++ b/bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
@@ -19,8 +19,10 @@ package org.bson.codecs.pojo;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.RecordComponent;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.util.ArrayList;
@@ -42,10 +44,19 @@ import static org.bson.codecs.pojo.PropertyReflectionUtils.isGetter;
 import static org.bson.codecs.pojo.PropertyReflectionUtils.toPropertyName;
 
 final class PojoBuilderHelper {
-
-    @SuppressWarnings("unchecked")
     static <T> void configureClassModelBuilder(final ClassModelBuilder<T> classModelBuilder, final Class<T> clazz) {
         classModelBuilder.type(notNull("clazz", clazz));
+
+        if (isRecord(clazz)) {
+            configureRecordClassModelBuilder(classModelBuilder, clazz);
+        } else {
+            configurePojoClassModelBuilder(classModelBuilder, clazz);
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <T> void configurePojoClassModelBuilder(final ClassModelBuilder<T> classModelBuilder, final Class<T> clazz) {
+        classModelBuilder.isRecord(false);
 
         ArrayList<Annotation> annotations = new ArrayList<Annotation>();
         Set<String> propertyNames = new TreeSet<String>();
@@ -57,10 +68,7 @@ final class PojoBuilderHelper {
         Map<String, PropertyMetadata<?>> propertyNameMap = new HashMap<String, PropertyMetadata<?>>();
         while (!currentClass.isEnum() && currentClass.getSuperclass() != null) {
             annotations.addAll(asList(currentClass.getDeclaredAnnotations()));
-            List<String> genericTypeNames = new ArrayList<String>();
-            for (TypeVariable<? extends Class<? super T>> classTypeVariable : currentClass.getTypeParameters()) {
-                genericTypeNames.add(classTypeVariable.getName());
-            }
+            List<String> genericTypeNames = getGenericTypeNames(currentClass);
 
             PropertyReflectionUtils.PropertyMethods propertyMethods = getPropertyMethods(currentClass);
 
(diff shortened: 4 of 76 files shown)


bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -145,7 +140,48 @@
...
+        for (int i = 0; i < recordComponents.length; i++) {
+            RecordComponent recordComponent = recordComponents[i];
+            String name = recordComponent.getName();
+            types[i] = recordComponent.getType();
+
+            Method getter;
+            try {
+                getter = clazz.getDeclaredMethod(name);
+            } catch (NoSuchMethodException e) {
+                throw new AssertionError("Record must have getter method with same name as field");
```
bdeleonardis1, line 165: I wasn't sure what to put in the catch block. These calls should never fail because the methods are guaranteed to to exist with records.

bson/src/main/org/bson/codecs/pojo/RecordCodecProvider.java
```diff
@@ -0,0 +1,202 @@
...
+                return null;
+            }
+        }
+        return null;
+    }
+
+    /**
+     * A Builder for the RecordCodecProvider
+     */
+    public static final class Builder {
```
bdeleonardis1, line 102: This class unfortunately contains a lot of duplicate code with the Builder in PojoCodecProvider. I considered making an abstract builder, but that would require me to either make the instance variables protected (which the style checker complained about) or add getter methods for all of the instance variables. I figured the extra code that was required to make that work wasn't worth it, but let me know if that was a mistake.

build.gradle
```diff
@@ -79,8 +79,8 @@ configure(coreProjects) {
 configure(javaProjects) {
     apply plugin: 'java-library'
 
-    sourceCompatibility = JavaVersion.VERSION_1_8
```
bdeleonardis1, line 82: You can ignore all this, we'll have to delete it before merging anyway.

record-support/src/main/com/mongodb/record/RecordTest.java
```diff
@@ -0,0 +1,209 @@
...
+import org.bson.types.ObjectId;
+
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
+import static org.bson.codecs.configuration.CodecRegistries.fromRegistries;
+
+public class RecordTest {
```
bdeleonardis1, line 34: This file is very ugly because it won't be merged. You can ignore the test logic, it just implements round trip testing and go straight to the tests if you want to look at them (but like I said those tests aren't comprehensive).

bdeleonardis1: Also I had to put this stuff in a new module because I couldn't get the bson module to run with Java 14 in preview mode. However, the bson module can still access the 14 preview features, if we run the code from a different module.

bson/src/main/org/bson/codecs/pojo/RecordCodecProvider.java
bdeleonardis1: We also may want to consider making an abstract UserDefinedCodecProvider, because RecordCodecProvider and PojoCodecProvider are so similar.

jyemin:
> If we are able to use ElementType.RecordComponent it will be very easy to support annotations, but if this causes a Java 8 compatibility problem we'll have to think harder about how to do that.

I think this will be ok. We will have to compile all the annotations with Java 16, but still target Java 8.  This isn't so different with what we do now: compile with Java 11 but target Java 8.

bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
```diff
@@ -34,20 +35,20 @@
...
     }
 
-    CreatorExecutable(final Class<T> clazz, final Method method) {
-        this(clazz, null, method);
+    CreatorExecutable(final Class<T> clazz, final Method method, final boolean isRecord) {
+        this(clazz, null, method, isRecord);
     }
 
-    private CreatorExecutable(final Class<T> clazz, final Constructor<T> constructor, final Method method) {
+    private CreatorExecutable(final Class<T> clazz, final Constructor<T> constructor, final Method method, final boolean isRecord) {
```
rozza, line 51: This is starting to looks a little complex - do you think this could be simplified more?

I'm not sure if there is a code smell at the InstanceCreatorFactory level as I can't recall why CreatorExecutable instances were required. Perhaps the abstraction for the implementation of InstanceCreators is not the best? Or alternatively perhaps an interface would be better for CreatorExecutables and then you can separate out and specialize the logic for pojos and records into discrete implementations.

Could you have a look and see if there is an opportunity to simplify here?

rozza (changes requested):
Looks really good, I had a couple of comments

bson/src/main/org/bson/codecs/pojo/RecordCodecProvider.java
```diff
@@ -0,0 +1,202 @@
...
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+import static org.bson.assertions.Assertions.notNull;
+
+/**
+ * Provides Codecs for registered Records via the ClassModel abstractions.
+ *
+ * @since TODO
+ */
+public final class RecordCodecProvider implements CodecProvider {
```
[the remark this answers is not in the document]
rozza: I see its to do with the PojoBuilderHelper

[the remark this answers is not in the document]
bdeleonardis1: It does differ a little (because of PojoBuilderHelper like you said), but you're right that we don't need a separate provider. My motivation for doing it this way was so that users could decide whether they wanted to support just records, just Pojos, or both. Now I'm thinking that we can choose a reasonable default for that (probably support both), and add a configuration in PjoCodecProvider where a user can specify if they want something different. Is that's even a useful option or should we just force users to support both?

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -45,8 +45,6 @@
 
     @SuppressWarnings("unchecked")
     static <T> void configureClassModelBuilder(final ClassModelBuilder<T> classModelBuilder, final Class<T> clazz) {
-        classModelBuilder.type(notNull("clazz", clazz));
```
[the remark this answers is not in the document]
bdeleonardis1: Yes we can do that. I can get the isRecord method via reflection and then invoke it. We can refactor to get rid of RecordCodecProvider if that's desirable, see my reply above.

bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
bdeleonardis1: Yup I'll think about it.

bson/src/main/org/bson/codecs/pojo/RecordCodecProvider.java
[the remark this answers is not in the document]
jyemin: Unlike POJO codec, I think record codec can be enabled safely by default.  That was my motivation for having a separate provider for records.

bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
bdeleonardis1: @rozza I ended up changing CreatorExecutable to an interface and then creating ConstructorCreatorExecutable, MethodCreatorExecutable, and RecordCreatorExecutable. I think this makes the code pretty clean, but there is a good amount of duplication between Constructor/MethodCreatorExecutable. I can also make a ClassCreatorExecutable which would represent both of them (basically the original CreatorExecutable).

bson/src/main/org/bson/codecs/pojo/RecordCodecProvider.java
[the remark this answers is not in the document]
bdeleonardis1: @rozza what do you think of what Jeff said? One option is we could just use PojoCodecProvider and by default have it support records, and require it to be manually configured if you want to support pojos as well.

bson/src/main/org/bson/codecs/pojo/CreatorExecutable.java
```diff
@@ -21,129 +21,43 @@
...
-            for (int i = 0; i < parameterAnnotations.length; ++i) {
-                Annotation[] parameterAnnotation = parameterAnnotations[i];
+    CodecConfigurationException getError(Class<?> clazz, String msg);
 
-                for (Annotation annotation : parameterAnnotation) {
-                    if (annotation.annotationType().equals(BsonProperty.class)) {
-                        properties.add((BsonProperty) annotation);
-                        break;
-                    }
+    static Integer initializeDataFromAnnotations(Annotation[][] parameterAnnotations, List<String> properties) {
```
bdeleonardis1, line 42: A couple things strike me as a little sloppy in this method...
- Passing an empty list in, just so things can be added to it
- Doing two things at once (initializing properties list and returning the id index)

If I didn't have this method, the code would need to be duplicated in both Constructor and MethodCreatorExecutable. If I broke it into two separate functions it would require two passes over the annotations list.

bson/src/main/org/bson/codecs/pojo/MethodCreatorExecutable.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    public CodecConfigurationException getError(final Class<?> clazz, final String msg) {
+        return new CodecConfigurationException(format("Invalid @BsonCreator method in %s. %s",
+                clazz.getSimpleName(), msg));
+    }
+
+    private void checkHasAnExecutable() {
+        if (method == null) {
+            throw new CodecConfigurationException(format("Cannot find a public constructor for '%s'.", clazz.getSimpleName()));
```
rozza, line 100: Nit: Change exception message

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -35,18 +36,24 @@
...
 import static org.bson.codecs.pojo.PropertyReflectionUtils.getPropertyMethods;
 import static org.bson.codecs.pojo.PropertyReflectionUtils.isGetter;
 import static org.bson.codecs.pojo.PropertyReflectionUtils.toPropertyName;
 
 final class PojoBuilderHelper {
-
-    @SuppressWarnings("unchecked")
     static <T> void configureClassModelBuilder(final ClassModelBuilder<T> classModelBuilder, final Class<T> clazz) {
-        classModelBuilder.type(notNull("clazz", clazz));
+        classModelBuilder.type(clazz);
```
rozza: Keep the notNull check?

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
rozza: Generally we seem to use a CodecConfiguration exception for pretty much all exceptions in Codecs. I think an IllegalStateException would be fine here and below. Also pass the  NoSuchMethodException exception on as well.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -42,11 +43,18 @@
 import static org.bson.codecs.pojo.PropertyReflectionUtils.toPropertyName;
 
 final class PojoBuilderHelper {
-
-    @SuppressWarnings("unchecked")
     static <T> void configureClassModelBuilder(final ClassModelBuilder<T> classModelBuilder, final Class<T> clazz) {
         classModelBuilder.type(notNull("clazz", clazz));
 
+        if (isRecord(clazz)) {
```
bdeleonardis1, line 49: @jyemin in the design doc you suggested "inverting the dependency" so that whoever constructed the ClassModelBuilder was responsible for configuring it correctly (i.e. RecordCodecProvider calls configureRecordClassModelBuilder). I implemented this initially, but I realized this would introduce a breaking change. 

ClassModel::builder is part of the public API and it constructs a ClassModelBuilder which originally called the configureClassModelBuilder method. To invert the dependency we would get rid of this call. Previously users get a fully configured classModelBuilder just by calling ClassModel#builder, now it isn't configured it all. This is bad because we accept ClassModels in our PojoCodecProvider in one of the register functions. I believe this means we can't invert the dependency so I did it this way instead.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -254,13 +303,33 @@ private static TypeParameterMap getTypeParameterMap(final List<String> genericTy
...
     static <V> V stateNotNull(final String property, final V value) {
         if (value == null) {
             throw new IllegalStateException(format("%s cannot be null", property));
         }
         return value;
     }
 
+    static boolean isRecord(final Class<?> clazz) {
+        try {
+            clazz.getClass().getMethod("isRecord");
```
rozza, line 323: Do you mean to get the Class of the Class<?> here ?

You could just: `return (Boolean) clazz.getMethod("isRecord").invoke()`

bson/src/main/org/bson/codecs/pojo/RecordCodecProvider.java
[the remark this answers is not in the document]
rozza: I'm happy to defer this decision for now.  I think we'll just have to consider if users will use Records in isolation to Pojo's or if they will combine them.  As its such a new feature we just don't have that data yet.  

@jyemin 's idea of having on by default Record support would be really good.

rozza (commented):
This is looking good - couple of minor nits.

bson/src/main/org/bson/codecs/pojo/MethodCreatorExecutable.java
bdeleonardis1: Done

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
bdeleonardis1: Oops, done.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
bdeleonardis1: No this is actually intentional. isRecord is a method on the class, not the object, so to see if it exists I needed to get the class of the class. I added a comment in the code attempting to explain this. I changed the format to how you did it with invoke though because it's clearer.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
bdeleonardis1: Got it, done.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
rozza: Cool, thanks for the clarification!

bson/src/main/org/bson/codecs/pojo/ClassModelBuilder.java
```diff
@@ -55,6 +55,7 @@
     private String discriminator;
     private String discriminatorKey;
     private String idPropertyName;
+    private boolean isRecord;
```
bdeleonardis1, line 58: I added this to avoid a call to `PojoBuilderHelper::isRecord` in `ConventionAnnotationImpl`. It's not added to the `ClassModel` that is built.

bson/src/main/org/bson/codecs/pojo/PojoBuilderHelper.java
```diff
@@ -144,8 +152,61 @@
...
+                throw new IllegalStateException("Record must have getter method with same name as field", e);
+            }
+            PropertyMetadata<?> metadata = getOrCreateMethodPropertyMetadata(name, declaringClassName, propertyNameMap,
+                    TypeData.newInstance(getter, true), propertyTypeParameterMap, null, genericTypeNames,
+                    getGenericType(getter, true));
+
+            metadata.setGetter(getter);
 
-        classModelBuilder.instanceCreatorFactory(new InstanceCreatorFactoryImpl<T>(new CreatorExecutable<T>(clazz, noArgsConstructor)));
+            for (Annotation a : recordComponent.getAnnotations()) {
```
bdeleonardis1, line 190: Because records are immutable, they won't have setters, so we can't really have the read/write breakdown we use for annotations in Pojos. Therefore I just add all the annotations to both the read and write annotations.

Also, by implementing it this way, I am requiring users to put the annotation on the RecordComponent (as opposed to a custom getter if they implement one). This seems reasonable to me, as overriding the getter (or writing a new one) should be pretty rare and I can't think of a reason a user would want to annotate the getter instead of the record component.

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
```diff
@@ -633,7 +633,7 @@ public void testRoundTripWithoutBsonAnnotation() {
         return conventions;
     }
 
-    class ObjectCodec implements Codec<Object> {
+    static class ObjectCodec implements Codec<Object> {
```
bdeleonardis1, line 636: Necessary in order to reuse ObjectCodec in my RecordCustomTest.

bson/src/test/unit/org/bson/codecs/pojo/RecordCustomTest.java
```diff
@@ -0,0 +1,279 @@
...
+    public void testInvalidBsonRepresentationIntDecoding() {
+        decodingShouldFail(getRecordCodec(BsonRepresentationUnsupportedIntRecord.class), "{'id': 'hello', s: '3'}");
+    }
+
+    @Test(expected = CodecConfigurationException.class)
+    public void testStringIdIsNotObjectId() {
+        encodesTo(getRecordCodecRegistry(BsonRepresentationUnsupportedIntRecord.class), new BsonRepresentationRecord("notanobjectid", 1), null);
+    }
+
+    private static <T> RecordCodecProvider getRecordProviderFromClassModel(ClassModelBuilder<T> classModelBuilder) {
```
bdeleonardis1, line 262: These helpers are in this class as opposed to PojoTestCase because I wanted to keep all record related code out of PojoTestCase in case the record test code needs to go in a separate module. The way I organized it is helpers specific to RecordCustomTest go in RecordCustomTest, helpers specific to RecordRoundTripTest go in RecordRoundTripTest, and tests shared by both of them go in RecordRoundTrip and are package private.

bson/src/main/org/bson/codecs/pojo/annotations/BsonId.java
```diff
@@ -32,6 +32,6 @@
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
-@Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER})
+@Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.RECORD_COMPONENT})
```
jyemin, line 35: Just confirming that after review of the Java 15 changes to the record specification, that this change is still necessary.  My reading suggests that we can do without it.

bson/src/main/org/bson/codecs/pojo/RecordCreatorExecutable.java
```diff
@@ -0,0 +1,113 @@
...
+            RecordComponent recordComponent = recordComponents[i];
+            String name = recordComponent.getName();
+
+            boolean isId = name.equals("id") || name.equals("_id");
+            for (Annotation a : recordComponent.getAnnotations()) {
+                if (a.annotationType().equals(BsonId.class)) {
+                    isId = true;
+                    break;
+                }
+                if (a.annotationType().equals(BsonProperty.class)) {
```
jyemin, line 59: I'm not clear why a component annotated with BsonProperty can't also be the id.  Can you explain?  I may be misunderstanding given that AnnotationBsonPropertyIdRecord does exactly that.

bson/src/test/unit/org/bson/codecs/pojo/Student.java
```diff
@@ -0,0 +1,6 @@
+package org.bson.codecs.pojo;
```
jyemin, line 1: Should this be moved to the same package as all the other records?

jyemin (changes requested):
It's looking great, and the tests are very thorough.  Just a few comments.  

Also: so that we don't forget, can you also add the RecordCodecProvider to the default registriy in MongoClientSettings?

bson/src/main/org/bson/codecs/pojo/RecordCodecProvider.java
```diff
@@ -0,0 +1,195 @@
...
+         * Registers codec providers that receive the type parameters of properties for instances encoded and decoded
+         * by a {@link PojoCodec} for records, handled by this provider.
+         *
+         * <p>Note that you should prefer working with the {@link CodecRegistry}/{@link CodecProvider} hierarchy. Providers
+         * should only be registered here if a codec needs to be created for custom container types like optionals and
+         * collections. Support for types {@link Map} and {@link java.util.Collection} are built-in so explicitly handling
+         * them is not necessary.
+         * @param providers property codec providers to register
+         * @return this
+         * @since 3.6
```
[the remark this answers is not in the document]
bdeleonardis1: Done

bson/src/main/org/bson/codecs/pojo/annotations/BsonId.java
bdeleonardis1: This change is actually still necessary. The cleanest way to get the annotations for me is to get them off the RecordCompnent, because that's what I iterate over in PojoBuilderHelper::configureRecordClassModelBuilder. I've manually tested and confirmed that I cannot get them off the RecordComponent in Java 15 without targeting the RECORD_COMPONENT.

I believe what Java 15 allows is for you to get the annotation off the field (or method, or parameter, etc.) if it is applicable. For example, if you have a record like the following:
`
public record Holder(@BsonIgnore String field) {}
`
You could get the annotation by doing: `Holder.class.getDeclaredField("field").getAnnotations()` instead of having to go through the RecordComponent. 

It is possible for us to avoid targeting RECORD_COMPONENT if that is desirable (I would just have to lookup the field by its name and get the annotation from there), but the code will be a little less clean. Let me know what you think is best...

bson/src/test/unit/org/bson/codecs/pojo/Student.java
bdeleonardis1: This actually isn't used by any of the tests. I was using it to manually test something else and forgot to delete it. Deleted now.

bson/src/main/org/bson/codecs/pojo/RecordCreatorExecutable.java
bdeleonardis1: Thanks for asking this, there was a bug here... I had done this originally to get the test involving AnnotationBsonPropertyIdRecord to work. By specifying @BsonProperty("id") the client is actually saying they do not want that field to be stored as the _id (because the write name is now id, so it can't be). There's a corresponding POJO test that does the same thing which is where I got the test case from.

But you're right that a RecordComponent with a BsonProperty can still be the _id if the name is actually "_id." I fixed the code to check for that now and added two test cases to RecordRoundTripTest.

bson/src/main/org/bson/codecs/pojo/annotations/BsonId.java
jyemin: I see now.  Thanks for the explanation.  I think it's fine to leave it as is.

bson/src/main/org/bson/codecs/pojo/RecordCreatorExecutable.java
```diff
@@ -0,0 +1,118 @@
...
+                    String writeName = ((BsonProperty) a).value();
```
jyemin, line 60: Nit: for simple conditionals like this it's cleaner to use a straight assignment:

    isId = ((BsonProperty) a).value().equals("_id");

bson/src/main/org/bson/codecs/pojo/RecordCreatorExecutable.java
```diff
@@ -0,0 +1,118 @@
...
+        RecordComponent[] recordComponents = clazz.getRecordComponents();
+        for (int i = 0; i < recordComponents.length; i++) {
+            RecordComponent recordComponent = recordComponents[i];
+            String name = recordComponent.getName();
+
+            boolean isId = name.equals("id") || name.equals("_id");
+            for (Annotation a : recordComponent.getAnnotations()) {
+                if (a.annotationType().equals(BsonId.class)) {
+                    isId = true;
+                    break;
```
jyemin, line 57: Nit: for simple loops like this where the conditions within are mutually exclusive, better to remove the break statements and use `else if`.  It reveals the intent more clearly.

bson/src/main/org/bson/codecs/pojo/RecordCreatorExecutable.java
```diff
@@ -0,0 +1,118 @@
...
+        parameterGenericTypes.addAll(asList(constructor.getGenericParameterTypes()));
+
+        RecordComponent[] recordComponents = clazz.getRecordComponents();
+        for (int i = 0; i < recordComponents.length; i++) {
+            RecordComponent recordComponent = recordComponents[i];
+            String name = recordComponent.getName();
+
+            boolean isId = name.equals("id") || name.equals("_id");
+            for (Annotation a : recordComponent.getAnnotations()) {
+                if (a.annotationType().equals(BsonId.class)) {
```
jyemin, line 55: Nit: you can simplify this to the less verbose and more efficient form of

     if (a instanceof BsonId))

and same for the next conditional

jyemin (commented):
While reviewing the last commit I found a few more nits for you to consider.

bson/src/main/org/bson/codecs/pojo/RecordCreatorExecutable.java
bdeleonardis1: Done

bson/src/main/org/bson/codecs/pojo/RecordCreatorExecutable.java
bdeleonardis1: Done

bson/src/main/org/bson/codecs/pojo/RecordCreatorExecutable.java
bdeleonardis1: Done, thanks for these tips

rozza:
I think this looks great - I'm looking forward to it landing in master!

Great work on migrating all the pojo test models over to records - I forgot how many there are! And for testing the record / pojo interop.

jyemin:
Also LGTM!  Great work!

jyemin (approved):

bdeleonardis1:
This branch has been pushed to the main repo and will be merged once Java 16 is released.

cardouken:
> This branch has been pushed to the main repo and will be merged once Java 16 is released.

Any news now that it's been released for a while?

jyemin:
@cardouken thanks for your interest.  This is definitely on our radar to merge, but it turns out there is more work to be done on the build side before it's ready to go.  IIRC we have to upgrade Gradle in order to build with Java 16, and that upgrade is not straightforward.

cardouken:
@jyemin thanks for the quick update, looking forward to it coming out.

jyemin:
This is unblocked.  The driver has been upgraded to Gradle 7.3 and builds with Java 17.   So we're going to pick this back up soon.

fmendezp:
@jyemin thanks for the several updates. Do you have any further information about the possibility of this moving forward?

jyemin:
We are still planning to move forward, but encountered a few more build complications.  Apologies for the further delays.

hfoxy:
Any further news on this?

jyemin:
See [JAVA-3567](https://jira.mongodb.org/browse/JAVA-3567).  It was released in 4.6.0.

hfoxy:
I see, thanks!

PR #582, opened 2020-09-25
Fix PojoCodec cache concurrency issue.

rozza:
The LazyPojoCodec was not thread safe when creating the PojoCodecImpl

JAVA-3775

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/LazyPojoCodec.java
+++ b/bson/src/main/org/bson/codecs/pojo/LazyPojoCodec.java
@@ -31,7 +31,7 @@ class LazyPojoCodec<T> extends PojoCodec<T> {
     private final PropertyCodecRegistry propertyCodecRegistry;
     private final DiscriminatorLookup discriminatorLookup;
     private final ConcurrentMap<ClassModel<?>, Codec<?>> codecCache;
-    private volatile PojoCodecImpl<T> pojoCodec;
+    private PojoCodecImpl<T> pojoCodec;
 
     LazyPojoCodec(final ClassModel<T> classModel, final CodecRegistry registry, final PropertyCodecRegistry propertyCodecRegistry,
                   final DiscriminatorLookup discriminatorLookup, final ConcurrentMap<ClassModel<?>, Codec<?>> codecCache) {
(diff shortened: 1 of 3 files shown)


bson/src/main/org/bson/codecs/pojo/LazyPojoCodec.java
```diff
@@ -57,7 +57,7 @@ public T decode(final BsonReader reader, final DecoderContext decoderContext) {
         return getPojoCodec().decode(reader, decoderContext);
     }
 
-    private Codec<T> getPojoCodec() {
+    private synchronized Codec<T> getPojoCodec() {
```
rozza, line 60: The race was here. volatile wasn't enough of a guarantee, different threads called getPojoCodec at different points causing the variable to be rewritten and the state to be non fixed.

> As soon as a thread needs to first read the value of a volatile variable, and based on that value generate a new value for the shared volatile variable, a volatile variable is no longer enough to guarantee correct visibility.
>
> The short time gap in between the reading of the volatile variable and the writing of its new value, creates an race condition where multiple threads might read the same value of the volatile variable, generate a new value for the variable, and when writing the value back to main memory - overwrite each other's values.

bdeleonardis1: Sorry @rozza I understand why this was a problem that needs to be fixed, but I still don't quite understand why this fixes the original bug. The original bug was that a PropertyModel's cached codecs was null. The volatile/synchronized bug might construct PojoCodecImpls and reassign them to pojoCodec unnecessarily, but the value returned by getPojoCodec is always guaranteed to be a fully constructed PojoCodecImpl, right? And from what we discussed in our meeting, that means the property models must be setup correctly because they are setup in the constructor... Do you mind giving a little more explanation?

rozza: So as far as I can tell race is multiple threads reading and overwriting the value at the same time.  The looping nature of registry lookups and caching that is addressed in #585 seems to also be partially the cause of why sometimes the propertyModel's codec is cached with a null value.

Locking the access to the variable until it is written removes the race condition and enables the pojoCodec value to be fully initialized and configured as expected.

I was only able to observe the effects of the propertyCodecRegistry recursively calling the codec registry but wasn't able to pin point the exact conditions where propertyModel.cacheCodec was called with a null value.

bdeleonardis1: Ok got it I think that makes sense.

jyemin:
Should this PR be closed, as #585 removes the LazyPojoCodec class entirely?

rozza:
Closing see: #585

PR #583, opened 2020-09-25
Sync retryable write tests to add error label

bdeleonardis1:
Couple of questions...
1. To sync the tests we just copy the files over from the spec repo right? There's no fancy command or git setup?
2. I didn't see us using any yaml files for the tests, so I'm assuming I only have to copy over the json ones? Drivers get to pick which ones they want to use?
3. I understand why this didn't cause any failures (it's just adding an error label that we need to have at the end), but I'm confused how we were passing it initially. I set breakpoints everywhere I thought we could add a RetryableWriteError and couldn't find where it actually happened. CommandOperationHelper::addRetryableErrorLabel wasn't doing it because the maxWireVersion = 9 and it wasn't a socket exception.

Evergreen patch: https://spruce.mongodb.com/version/5f6e2ad5d1fe07104a068d9a/tasks

### DIFF
--- a/driver-core/src/test/resources/retryable-writes/insertOne-serverErrors.json
+++ b/driver-core/src/test/resources/retryable-writes/insertOne-serverErrors.json
@@ -966,7 +966,10 @@
           ],
           "writeConcernError": {
             "code": 91,
-            "errmsg": "Replication is being shut down"
+            "errmsg": "Replication is being shut down",
+            "errorLabels": [
+              "RetryableWriteError"
+            ]
           }
         }
       },


rozza:
> 1. To sync the tests we just copy the files over from the spec repo right? There's no fancy command or git setup?

Yes, I just do a local copy.  Updates to tests / specs can happen sporadically and the process of using Drivers tickets to manage / schedule the work seems to work well enough.

> 2. I didn't see us using any yaml files for the tests, so I'm assuming I only have to copy over the json ones? Drivers get to pick which ones they want to use? 

Yes, the json files are auto created from the yaml files, this is required as not all drivers / languages have native yaml support, but all drivers have to be able to consume Json to be able to handle extended json.  

I think the reason why yaml is preferred is it supports comments etc - so human readable metadata can be added easily when writing the tests.

> 3. I understand why this didn't cause any failures (it's just adding an error label that we need to have at the end), but I'm confused how we were passing it initially. I set breakpoints everywhere I thought we could add a RetryableWriteError and couldn't find where it actually happened. CommandOperationHelper::addRetryableErrorLabel wasn't doing it because the maxWireVersion = 9 and it wasn't a socket exception.

Its a confusing path, but essentially all CRUD operations are handled as potentially bulk operations, so an `insertOne({})` is actually the same as an `insertMany([{}])`. The problem with bulk operations is a user doesn't know or care about the size of the operation and if it fits on the wire.  So we have to batch up bulk operations and deliver them as efficiently as possible. This creates a complex code path. The results of each batch of the bulk operation then gets merged together into a final result. Its not the cleanest area of code in the driver.

Part of the handling of batch errors is to check the exceptions and add the error labels as per the spec.  So older versions of the server will issue the 91 WriteConcernError with no error label and newer ones will automatically add the error label, even if its not in the fail command.  So that is why it always passed across versions.

See: https://github.com/mongodb/mongo-java-driver/blob/master/driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java#L254-L255 for the handling / labelling of the error if its missing.

bdeleonardis1:
@rozza thanks a lot for that explanation it was really helpful.

jyemin (approved):
LGTM.  Good questions! And good answers, too.

bdeleonardis1:
Merged on command line.

PR #585, opened 2020-09-28
Fix PojoCodec caching 

rozza:
This sits on top of #582 

Reduce the number of required CodecRegistry lookups for properties
Reduce the number of Lazy Codecs for properties

JAVA-3847


bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
```diff
@@ -0,0 +1,145 @@
...
+import org.bson.codecs.RepresentationConfigurable;
+import org.bson.codecs.configuration.CodecConfigurationException;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.bson.codecs.pojo.PojoSpecializationHelper.specializeTypeData;
+
+class LazyPropertyModelCodec<T> implements Codec<T> {
```
rozza, line 33: This is the main change - instead of the LazyPojoCodec - this makes the PropertyModel cached codec itself lazy and reduces the calls back to the registry that previously existed thanks to the PropertyCodecRegistry containing using the FallbackPropertyCodecProvider

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -48,41 +43,28 @@
     private final CodecRegistry registry;
     private final PropertyCodecRegistry propertyCodecRegistry;
     private final DiscriminatorLookup discriminatorLookup;
-    private final ConcurrentMap<ClassModel<?>, Codec<?>> codecCache;
     private final boolean specialized;
 
     PojoCodecImpl(final ClassModel<T> classModel, final CodecRegistry codecRegistry,
                   final List<PropertyCodecProvider> propertyCodecProviders, final DiscriminatorLookup discriminatorLookup) {
         this.classModel = classModel;
-        this.registry = fromRegistries(fromCodecs(this), codecRegistry);
```
rozza, line 57: Removed this nesting of registries

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -48,41 +43,28 @@
...
+        this.registry = codecRegistry;
         this.discriminatorLookup = discriminatorLookup;
-        this.codecCache = new ConcurrentHashMap<ClassModel<?>, Codec<?>>();
```
rozza, line 59: The codecCache wasn't useful and was essentially replaced with a cache in the PropertyCodecRegistryImpl

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -234,9 +216,18 @@ private void decodeProperties(final BsonReader reader, final DecoderContext deco
...
-    private <S> void addToCache(final PropertyModel<S> propertyModel) {
-        Codec<S> codec = propertyModel.getCodec() != null ? propertyModel.getCodec() : specializePojoCodec(propertyModel);
-        propertyModel.cachedCodec(codec);
+    private void specialize() {
+        if (specialized) {
+            classModel.getPropertyModels().forEach(this::cachePropertyModelCodec);
+        }
+    }
+
+    private <S> void cachePropertyModelCodec(final PropertyModel<S> propertyModel) {
```
rozza, line 225: This has been simplified and made lazy.

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -250,89 +241,6 @@ private void decodeProperties(final BsonReader reader, final DecoderContext deco
         return false;
     }
 
-
-
-    @SuppressWarnings("unchecked")
-    private <S> Codec<S> specializePojoCodec(final PropertyModel<S> propertyModel) {
-        Codec<S> codec = getCodecFromPropertyRegistry(propertyModel);
```
rozza, line 257: This call could loop back to the registry creating new PojoCodec instances. These then could then become specialized instances and weren't cached adequately.

So the specialization and caching were split out to the LazyPropertyModelCodec - which handles the specialization of itself and caches locally on demand.

The PropertyCodecRegistry handles the caching of types per PojoCodec, which can then later be specialized.  This reduces the number of roundtrips to the codecRegistry for types.

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -347,7 +255,10 @@ private void decodeProperties(final BsonReader reader, final DecoderContext deco
                 if (discriminatorKey.equals(name)) {
                     discriminatorKeyFound = true;
                     try {
-                        codec = (Codec<T>) registry.get(discriminatorLookup.lookup(reader.readString()));
+                        Class<?> discriminatorClass = discriminatorLookup.lookup(reader.readString());
+                        if (!codec.getEncoderClass().equals(discriminatorClass)) {
+                            codec = (Codec<T>) registry.get(discriminatorClass);
```
rozza, line 260: Only go to the codec registry if needed. Previously this worked because the codec registry was changed to be; `fromRegistries(fromCodecs(this), codecRegistry);` in the constructor.

bson/src/main/org/bson/codecs/pojo/PropertyCodecRegistryImpl.java
```diff
@@ -36,16 +38,25 @@
...
+        }
+
         for (PropertyCodecProvider propertyCodecProvider : propertyCodecProviders) {
-            Codec<S> codec = propertyCodecProvider.get(type, this);
-            if (codec != null) {
-                return codec;
-            }
+                Codec<S> codec = propertyCodecProvider.get(typeWithTypeParameters, this);
+                if (codec != null) {
+                    propertyCodecCache.put(typeWithTypeParameters, codec);
```
rozza, line 54: Cache if we can.

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
```diff
@@ -613,7 +613,7 @@ public void testInvalidBsonRepresentationStringEncoding() {
 
     @Test(expected = CodecConfigurationException.class)
     public void testInvalidBsonRepresentationIntDecoding() {
-        decodingShouldFail(getCodec(BsonRepresentationUnsupportedInt.class), "{'id': 'hello', s: '3'}");
```
rozza, line 616: This failed not in decoding but in creation of the Codec. However, properties are lazy now, so the codec is only created on demand.

bdeleonardis1 (commented):
This was cool, I added two questions to the code. I also have a more meta question...
I don't understand why we have to worry about codec caching in general. Nothing seems particularly costly about looking up the codec from the PropertyCodecRegistry. I would assume executing db commands would be a substantially larger bottleneck. Have we ever done benchmarking to see how much the caching helps or how slow the codec stuff is?

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
rozza: I noticed we did this when I wanted to test the number of registry lookups for classes and wanted to remove the impediment to be able to determine how many times the registry is actually called for lookups.

It has the benefit of simplifying the code / logic.  Codec lookups are cached by the ProvidersCodecRegistry, and there's no need to create a new ProvidersCodecRegistry and cache for each PojoCodec.

Why was it necessary in the first place? 

This was a workaround to handle fully configured PojoCodecs with their type information and not to use the registry to get the raw PojoCodec which is not specialized with type information.  The `NestedGenericHolderFieldWithMultipleTypeParamsModel` test class covers this scenario, the `nested` field contains the fully reified data for the `PropertyWithMultipleTypeParamsModel`. If the PojoCodec went to the codec registry to obtain a `PropertyWithMultipleTypeParamsModel` instance it wouldn't be specialized and would have no concrete types.  So we still need a mechanism to ensure that if the configured PojoCodec is the discriminated class then we should use that.

Check out the `NestedGenericHolderFieldWithMultipleTypeParamsModel` for an example.

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -234,9 +216,18 @@ private void decodeProperties(final BsonReader reader, final DecoderContext deco
...
-        propertyModel.cachedCodec(codec);
+    private void specialize() {
+        if (specialized) {
+            classModel.getPropertyModels().forEach(this::cachePropertyModelCodec);
+        }
+    }
+
+    private <S> void cachePropertyModelCodec(final PropertyModel<S> propertyModel) {
+        if (propertyModel.getCachedCodec() == null) {
+            Codec<S> codec = propertyModel.getCodec() != null ? propertyModel.getCodec()
```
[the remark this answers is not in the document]
rozza: Good question, users can explicitly provide a codec for their properties and this allows those codecs to be used if configured.

rozza:
Thanks @bdeleonardis1 for the review and great questions.

> I don't understand why we have to worry about codec caching in general. Nothing seems particularly costly about looking up the codec from the PropertyCodecRegistry. I would assume executing db commands would be a substantially larger bottleneck. 

So looking up from the `PropertyCodecRegistry` is cheap and as it falls back to the general codec registry - the higher level `ProvidersCodecRegistry` also provides caching. However, Pojo's and generic types are the issue, there is some cost to creating / specializing a codec for a property - setting the generic types etc.. Is it worth spending this each time?

The belief is users will be using their Pojos for CRUD operations, in those scenarios a PojoCodec instance would be repeatedly used to decode / encode values. So having property codecs available and cached, stops repeat processing.  There is bound to be some trade off, the cache is indiscriminate for basic codecs and specialized codecs alike.

> Have we ever done benchmarking to see how much the caching helps or how slow the codec stuff is?
I recall some benchmarking in the early days of the PojoCodec but not in relation to caching. It could be an interesting exercise, I'd expect the data model used would sway the benchmark. A simple model would probably just cost memory. A complex model without caching would repeatedly create new specialized codecs, so would cost memory and cpu.  But I haven't tested that hypothesis.

bdeleonardis1:
@rozza thanks for those very detailed answers, they were really helpful.

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
```diff
@@ -48,41 +43,28 @@
...
         specialize();
     }
 
     PojoCodecImpl(final ClassModel<T> classModel, final CodecRegistry registry, final PropertyCodecRegistry propertyCodecRegistry,
-                  final DiscriminatorLookup discriminatorLookup, final ConcurrentMap<ClassModel<?>, Codec<?>> codecCache,
-                  final boolean specialized) {
+                  final DiscriminatorLookup discriminatorLookup, final boolean specialized) {
         this.classModel = classModel;
-        this.registry = fromRegistries(fromCodecs(this), registry);
+        this.registry = fromRegistries(registry, fromCodecs(this));
```
jyemin, line 61: Why is the order swapped here?

jyemin (changes requested):

bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
```diff
@@ -0,0 +1,145 @@
...
+    private synchronized Codec<T> getPropertyModelCodec() {
+        if (codec == null) {
+            Codec<T> localCodec = getCodecFromPropertyRegistry(propertyModel);
+            if (localCodec instanceof PojoCodec) {
+                PojoCodec<T> pojoCodec = (PojoCodec<T>) localCodec;
+                ClassModel<T> specialized = getSpecializedClassModel(pojoCodec.getClassModel(), propertyModel);
+                localCodec = new PojoCodecImpl<>(specialized, registry, propertyCodecRegistry, discriminatorLookup, true);
+            }
+            if (localCodec == null) {
+                localCodec = new LazyMissingCodec<>(propertyModel.getTypeData().getType(),
```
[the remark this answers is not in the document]
rozza: I've moved it into the `getCodecFromPropertyRegistry` its not dead code just very defensive. If a user creates their own codec registry implementation that returns null from `registry.get(clazz)` then it would be required.

bson/src/main/org/bson/codecs/pojo/PojoCodecImpl.java
rozza: Updated - no need to nest the registry - as per the other constructor.

bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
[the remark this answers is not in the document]
jyemin: That would violate the specification of the CodecRegistry interface but you never know what users will do.

jyemin (approved):

PR #586, opened 2020-09-29
Removed unused fsync field from WriteConcern

rozza:
JAVA-3846


jyemin (approved):
LGTM.  I feel cleaner already.

PR #589, opened 2020-10-06
Potential fix for JAVA-3769

bdeleonardis1:
I spent the day figuring out what was causing the failure, so I figured I might as well put a fix up for this in case that's the route we want to take. I don't believe, however, this should be fixed for the following reasons:
- By fixing we're basically endorsing using multiple PojoCodecProviders while that's not intended. 
- It adds some complexity (albeit not a lot) to the codebase for little reward

In my opinion there are three things we can do instead to make this better for users:
1. Add a line to our documentation saying that only one PojoCodecProvider can be used.

2. Add to the error message in DiscriminatorLookup::lookup to say something like: "Make sure only one PojoCodecProvider is in your registry." My only concern with this is it may confuse users who are getting that error for a different reason. Unfortunately it is not possible to detect when multiple PojoCodecProviders are the cause of the "class could not be found for the discriminator" error without some kind of global state because there is no way of knowing if the other providers have the discriminator that we can't find. Because of this, we cannot give a more specific error message.

3. Throw an exception in ProvidersCodecRegistry's constructor if multiple codecProviders are PojoCodecProviders. This allows us to throw a very descriptive error, but it's potentially a breaking change. This bug only causes failures when the Pojos are nested and separate providers are used (i.e. Person and Address have different providers and Person contains Address). If the providers are for two separate classes that do not contain each other, clients may be running code successfully using two PojoCodecProviders at once and we're breaking that now.

In my opinion the best thing to do is a combo of Option 1 and Option 2, but I wanted to lay out all the options and get some feedback.

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/AutomaticPojoCodec.java
+++ b/bson/src/main/org/bson/codecs/pojo/AutomaticPojoCodec.java
@@ -66,4 +66,9 @@ final class AutomaticPojoCodec<T> extends PojoCodec<T> {
     ClassModel<T> getClassModel() {
         return pojoCodec.getClassModel();
     }
+
+    @Override
+    DiscriminatorLookup getDiscriminatorLookup() {
+        return pojoCodec.getDiscriminatorLookup();
+    }
 }
--- a/bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
+++ b/bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
@@ -67,7 +67,7 @@ class LazyPropertyModelCodec<T> implements Codec<T> {
             if (localCodec instanceof PojoCodec) {
                 PojoCodec<T> pojoCodec = (PojoCodec<T>) localCodec;
                 ClassModel<T> specialized = getSpecializedClassModel(pojoCodec.getClassModel(), propertyModel);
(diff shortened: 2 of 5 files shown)


jyemin:
> By fixing we're basically endorsing using multiple PojoCodecProviders while that's not intended.

I don't have a problem endorsing this.  In fact, I think we had assumed it would work.  Certainly it's the case that we expect multiple providers in general to cooperate with one another

> It adds some complexity (albeit not a lot) to the codebase for little reward

If this fix is indeed correct, it seems like a small bit of extra complexity to pay


What do you think @rozza ?

bdeleonardis1:
Ok that makes sense to me. Note that once @rozza merges https://github.com/mongodb/mongo-java-driver/pull/585 this fix will have to change so we might want to let him merge his first.

rozza:
I think this is great, it removes an impediment to users and makes using PojoCodecs more robust. Also, when we add record support it will naturally fix any discriminator lookups between the RecordCodecProvider and the PojoCodecProvider.

I've merge #585 so this will need a slight update but LGTM

bdeleonardis1:
@jyemin @rozza would you mind taking another look before I merge? The diff between the two commits will be a little weird because of rebasing Ross's code, but the only real change was moving the line where I get the correct DiscriminatorLookup.

rozza (approved):
LGTM

bdeleonardis1:
Rebased and merged on command line.

PR #590, opened 2020-10-07
bson: simplify StringUtils

amarkevich:
JAVA-4013


bson/src/main/org/bson/StringUtils.java
```diff
@@ -16,21 +16,16 @@
...
-        Iterator<?> iter = s.iterator();
-        while (iter.hasNext()) {
-            builder.append(iter.next());
-            if (!iter.hasNext()) {
-                break;
-            }
-            builder.append(delimiter);
+    @SafeVarargs
+    public static <T> String join(final String delimiter, final T... values) {
+        StringJoiner joiner = new StringJoiner(delimiter);
```
jyemin, line 24: This is nice.  Could we do one better by using streams, as shown in the Javadoc for StringJoiner?

amarkevich: @jyemin done

jyemin:
@amarkevich, sorry for losing track of this.  This has now been merged.  Thanks for the contribution!

PR #596, opened 2020-10-07
JAVA-3845: Remove travis config files and references

bdeleonardis1:
Evergreen automatic testing on PRs is now enabled so I'm removing the travis related stuff. Also let me know if you would prefer a different set of default tests to run on evergreen. I currently just run the static-checks and 4.4 tests on repl sets, sharding, and standalone.


jyemin (approved):
LGTM!

bdeleonardis1:
Merged on command line.

PR #597, opened 2020-10-08
DOCSP-12704: fix the link to MongoClientOptions api doc

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-12704

This change may apply to other branches depending on when MongoClientOptions was moved to the legacy library


jyemin (commented):
Were you able to test the links?

ccho-mongodb:
> Were you able to test the links?

I was able to test that the relative URLs resolved to
"apidocs/mongodb-driver-legacy/com/mongodb/MongoClientOptions.html" when running a hugo server. I wasn't immediately sure how to configure it to use 4.1 as the base URL.

Also, this PR will need to be cherry-picked to version branches 3.7 and later right? I'm not familiar with your release process but am happy to learn.

jyemin:
We're probably not going to fix older versions of the docs.  We never have before.

PR #600, opened 2020-10-19
JAVA-3868: Update docs to reflect and filter behaviour

durran:


driver-core/src/main/com/mongodb/client/model/Filters.java
```diff
@@ -208,17 +208,15 @@ private Filters() {
     }
 
     /**
-     * Creates a filter that performs a logical AND of the provided list of filters.  Note that this will only generate a "$and"
-     * operator if absolutely necessary, as the query language implicity ands together all the keys.  In other words, a query expression
-     * like:
+     * Creates a filter that performs a logical AND of the provided list of filters.
```
jyemin, line 211: Also apply this to the overloaded method just below.

jyemin (changes requested):

driver-core/src/main/com/mongodb/client/model/Filters.java
durran: Done

jyemin (approved):

PR #601, opened 2020-10-20
Update serialization of ObjectId to a stable interchange format

jyemin:
The existing serialization format for ObjectId is the default format provided by the
ObjectOutputStream specification, which just serializes each field of the class.  But
in the 3.8.0 release we changed the fields, but neglected to even change the
serialVersionUID of the class, which makes any instances of ObjectId serialized prior
to 3.8.0 dangerously incompatible with subsequent driver releases.

This change updates the serialVersionUID for ObjectId, and also defines the serialized
form using a proxy via the writeReplace method.  This means that serialized instances
of ObjectId prior to this change will be incompatible, but at least an exception
will be thrown indicating the incompatibility. And now we have the freedom to change
the private representation of ObjectId without affecting serialization format in the
future.

JAVA-3854

### DIFF
--- a/bson/src/main/org/bson/types/ObjectId.java
+++ b/bson/src/main/org/bson/types/ObjectId.java
@@ -16,6 +16,8 @@
 
 package org.bson.types;
 
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
 import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


rozza (approved):
LGTM

PR #604, opened 2020-10-28
Replace SimpleDateFormat with DateDateFormatter for ISODate string parsing

jyemin:
The JsonReader class already uses java.time.format.DateTimeFormatter for parsing
date strings from relaxed extended JSON. But for historical reasons it still
used the older java.text.SimpleDateFormat for parsing ISODate strings from shell
mode JSON.

This patch replaces the SimpleDateFormat usage with DateTimeFormatter, which will
preserve the existing behavior, but also remove a dependency on Locale.ENGLISH
time zones, which has caused problems for some users of GraalVM (which in some
configurations does not include any of the resources associated with Locale.ENGLISH.

Note that SimpleDateFormat is still used when parsing date strings for the shell-mode
Date constructor, but that is less commonly used and in addition the current shell
support is not localizable and is tied to English-language usage.

JAVA-3851

### DIFF
--- a/bson/src/main/org/bson/json/DateTimeFormatter.java
+++ b/bson/src/main/org/bson/json/DateTimeFormatter.java
@@ -17,25 +17,30 @@
 package org.bson.json;
 
 import java.time.Instant;
+import java.time.LocalDate;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
-import java.time.format.DateTimeParseException;
 import java.time.temporal.TemporalAccessor;
 import java.time.temporal.TemporalQuery;
 
+import static java.time.format.DateTimeFormatter.ISO_LOCAL_DATE;
 import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
 
 final class DateTimeFormatter {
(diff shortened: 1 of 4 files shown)


bson/src/main/org/bson/json/DateTimeFormatter.java
```diff
@@ -17,25 +17,30 @@
...
+        if (dateTimeString.length() == DATE_STRING_LENGTH) {
+            return LocalDate.parse(dateTimeString, ISO_LOCAL_DATE).atStartOfDay().toInstant(ZoneOffset.UTC).toEpochMilli();
+        } else {
             return ISO_OFFSET_DATE_TIME.parse(dateTimeString, new TemporalQuery<Instant>() {
                 @Override
                 public Instant queryFrom(final TemporalAccessor temporal) {
                     return Instant.from(temporal);
                 }
             }).toEpochMilli();
-        } catch (DateTimeParseException e) {
```
jyemin, line 37: This wrapping in IllegalArgumentException is unnecessary, and also hides the stack trace from the underlying exception,

rozza (approved):
LGTM

durran:
I'm all for simplification. LGTM

durran (approved):

PR #608, opened 2020-11-11
Improve assertion error context in unified test runner

jyemin:
Sorry for the noise on this.  But I found a number of places where the assertion context was insufficient to debug test failures in the versioned API and thought it best to fix it in a small, separate PR.

JAVA-3495

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
@@ -21,16 +21,22 @@ import com.mongodb.event.CommandEvent;
 import com.mongodb.event.CommandFailedEvent;
 import com.mongodb.event.CommandStartedEvent;
 import com.mongodb.event.CommandSucceededEvent;
+import org.bson.BsonArray;
 import org.bson.BsonDocument;
 import org.bson.BsonString;
 import org.bson.BsonValue;
 import org.bson.json.JsonWriterSettings;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 abstract class ContextElement {
-    static ContextElement ofOperation(final BsonDocument operation) {
-        return new OperationContextElement(operation);
+    public static ContextElement ofTest(final BsonDocument definition) {
+        return new TestContextContextElement(definition);
+    }
+
+    static ContextElement ofOperation(final BsonDocument operation, final OperationResult result) {
+        return new OperationContextElement(operation, result);
     }
 
     static ContextElement ofValueMatcher(final BsonValue expected, final BsonValue actual, final String key, final int arrayPosition) {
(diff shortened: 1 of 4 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
```diff
@@ -21,16 +21,22 @@
...
 import org.bson.BsonValue;
 import org.bson.json.JsonWriterSettings;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 abstract class ContextElement {
-    static ContextElement ofOperation(final BsonDocument operation) {
-        return new OperationContextElement(operation);
+    public static ContextElement ofTest(final BsonDocument definition) {
```
jyemin, line 34: Now we have a context for both the test itself, and each operation

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
```diff
@@ -42,28 +48,47 @@ static ContextElement ofError(final BsonDocument expectedError, final Exception
...
 
     static ContextElement ofOutcome(final MongoNamespace namespace, final List<BsonDocument> expectedOutcome,
-                                           final List<BsonDocument> actualOutcome) {
+                                    final List<BsonDocument> actualOutcome) {
         return new OutcomeMatchingContextElement(namespace, expectedOutcome, actualOutcome);
     }
 
-    static ContextElement ofEvents(final String client) {
-        return new EventsMatchingContextElement(client);
+    static ContextElement ofEvents(final String client, BsonArray expectedEvents, List<CommandEvent> actualEvents) {
```
jyemin, line 55: Now the event context includes the expected and actual events, so that if there is a mismatch in the number of event, you can see what the actual ones were.  This helped debug one of the versioned API tests, where I thought it was an extra getMore but it was actually an extra killCursors

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -152,11 +151,15 @@ private void assertOutcome() {
 
     private void assertOperation(final BsonDocument operation) {
         OperationResult result = executeOperation(operation);
+        context.push(ContextElement.ofOperation(operation, result));
         if (operation.containsKey("expectResult")) {
-            valueMatcher.assertValuesMatch(operation.get("expectResult"), requireNonNull(result.getResult()));
+            assertNotNull(context.getMessage("The operation expects a result but an exception occurred"), result.getResult());
```
jyemin, line 156: This change turns an NPE into an assertion, which is correct, as there are cases where the test fails because it expect a result and the driver threw instead

durran (approved):
LGTM

rozza (approved):

PR #610, opened 2020-11-13
Support Versioned API

jyemin:
Supported in driver-sync and driver-legacy only

JAVA-3878


driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -629,6 +648,16 @@ public UuidRepresentation getUuidRepresentation() {
         return uuidRepresentation;
     }
 
+    /**
+     * Gets the server API to use when sending commands to the server.
+     *
+     * @return the server API
+     * @since 4.3
+     */
+    public Optional<ServerApi> getServerApi() {
```
jyemin, line 657: Not sure whether we should start introduction Optional in Settings, as there are many other settings that can be null and a new paradigm might be confusing.

driver-core/src/main/com/mongodb/internal/async/client/AsyncMongoClients.java
```diff
@@ -147,7 +147,7 @@ private static Cluster createCluster(final MongoClientSettings settings,
         return new DefaultClusterFactory().createCluster(settings.getClusterSettings(), settings.getServerSettings(),
                 settings.getConnectionPoolSettings(), streamFactory, heartbeatStreamFactory, settings.getCredential(),
                 getCommandListener(settings.getCommandListeners()), settings.getApplicationName(), mongoDriverInformation,
-                settings.getCompressorList());
+                settings.getCompressorList(), settings.getServerApi().orElse(null));
```
jyemin, line 150: This is the extent of async support at this point.

driver-core/src/main/com/mongodb/internal/binding/ClusterBinding.java
```diff
@@ -81,6 +88,11 @@ public SessionContext getSessionContext() {
         return new ReadConcernAwareNoOpSessionContext(readConcern);
     }
 
+    @Override
+    public ServerApi getServerApi() {
```
jyemin, line 92: Since this has to make it's way all the way down to CommandMessage, it seems best to add it to bindings rather than having to pass it down through every operation.

driver-core/src/main/com/mongodb/internal/binding/SingleConnectionReadBinding.java
```diff
@@ -1,116 +0,0 @@
...
-import com.mongodb.internal.session.SessionContext;
-
-import static com.mongodb.assertions.Assertions.notNull;
-
-/**
- * A read binding that is bound to a single connection.
- *
- * @since 3.2
- */
-public class SingleConnectionReadBinding extends AbstractReferenceCounted implements ReadBinding {
```
jyemin, line 32: Rather than add ServerApi to an unused class, I just removed it.

driver-core/src/main/com/mongodb/internal/connection/ClusterFactory.java
```diff
@@ -1,60 +0,0 @@
...
-import com.mongodb.connection.StreamFactory;
-import com.mongodb.event.ClusterListener;
-import com.mongodb.event.ConnectionPoolListener;
-
-/**
- * Factory for {@code Cluster} implementations.
- *
- * @since 3.0
- */
-public interface ClusterFactory {
```
jyemin, line 32: It was getting confusing to have this interface so I just killed it.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -177,8 +182,14 @@ protected EncodingMetadata encodeMessageBodyWithMetadata(final BsonOutput bsonOu
...
             if (payload == null) {
-                addDocument(getCommandToEncode(), bsonOutput, commandFieldNameValidator, null);
+                List<BsonElement> elements = null;
+                if (serverApi != null) {
+                    elements = new ArrayList<>(3);
+                    addServerApiElements(elements);
+                }
+                addDocument(getCommandToEncode(), bsonOutput, commandFieldNameValidator, elements);
             } else {
+                // TODO: should never get to this point with a serverApi specified?
```
jyemin, line 192: I think I can remove this TODO.  Any server that supports serverApi will also support OP_MSG so should never get here.

driver-core/src/main/com/mongodb/internal/connection/DefaultClusterFactory.java
```diff
@@ -19,39 +19,24 @@
...
 
 /**
  * The default factory for cluster implementations.
  *
  * @since 3.0
  */
-public final class DefaultClusterFactory implements ClusterFactory {
-
-    @Override
-    public Cluster create(final ClusterSettings settings, final ServerSettings serverSettings,
```
jyemin, line 43: Unused method from unused ClusterFactory interface

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -324,51 +326,59 @@ static BsonDocument executeCommand(final WriteBinding binding, final String data
...
     private static <D, T> T executeCommand(final String database, final BsonDocument command,
                                            final FieldNameValidator fieldNameValidator, final Decoder<D> decoder,
                                            final ConnectionSource source, final Connection connection, final ReadPreference readPreference,
-                                           final CommandReadTransformer<D, T> transformer, final SessionContext sessionContext) {
+                                           final CommandReadTransformer<D, T> transformer, final SessionContext sessionContext,
+                                           final ServerApi serverApi) {
 
-        return transformer.apply(connection.command(database, command, fieldNameValidator, readPreference, decoder, sessionContext),
-                source, connection);
+        // TODO: Get ServerApi somehow
```
jyemin, line 349: I can remove the TODO

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -324,51 +326,59 @@ static BsonDocument executeCommand(final WriteBinding binding, final String data
...
 
     private static <D, T> T executeWriteCommand(final String database, final BsonDocument command,
                                                 final FieldNameValidator fieldNameValidator, final Decoder<D> decoder,
                                                 final Connection connection, final ReadPreference readPreference,
-                                                final CommandWriteTransformer<D, T> transformer, final SessionContext sessionContext) {
+                                                final CommandWriteTransformer<D, T> transformer, final SessionContext sessionContext,
+                                                final ServerApi serverApi) {
 
-        return transformer.apply(connection.command(database, command, fieldNameValidator, readPreference, decoder, sessionContext),
+        // TODO: inject ServerApi somehow
```
jyemin, line 379: This one too

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -269,7 +269,7 @@ private void getMore() {
                                                              NO_OP_FIELD_NAME_VALIDATOR,
                                                              ReadPreference.primary(),
                                                              CommandResultDocumentCodec.create(decoder, "nextBatch"),
-                                                             connectionSource.getSessionContext()));
+                                                             connectionSource.getSessionContext(), null));
```
jyemin, line 272: According to server specification, the one command that doesn't accept server version is getMore.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -155,9 +155,10 @@ private void endClosedSessions(final List<BsonDocument> identifiers) {
                 }
             }).getConnection();
 
+            // TODO: how to inject serverApi here, or should we just not?
```
jyemin, line 158: I think this has to be done.  Otherwise a server started with requireApiVersion=true will reject this command

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
```diff
@@ -450,6 +460,16 @@ public static MongoCredentialWithCache getCredentialWithCache() {
         return getConnectionString().getCredential() == null ? null : new MongoCredentialWithCache(getConnectionString().getCredential());
     }
 
+    public static BsonDocument getServerParameters() {
```
jyemin, line 463: Required for new unified test runner requirements for runOn

driver-sync/src/test/functional/com/mongodb/client/unified/ErrorMatcher.java
```diff
@@ -56,6 +56,14 @@ void assertErrorsMatch(final BsonDocument expectedError, final Exception e) {
                     expectedError.getBoolean("isClientError").getValue(),
                     e instanceof MongoClientException || e instanceof IllegalArgumentException);
         }
+        if (expectedError.containsKey("errorContains")) {
```
jyemin, line 59: First use of this, so had to add support

driver-sync/src/test/functional/com/mongodb/client/unified/RunOnRequirementsMatcher.java
```diff
@@ -21,37 +21,62 @@
...
+                        }
+                        break;
+                    case "topologies":
+                        BsonArray topologyTypes = curRequirement.getValue().asArray();
+                        if (!topologyMatches(topologyTypes)) {
+                            requirementMet = false;
+                            break requirementLoop;
+                        }
+                        break;
+                    case "serverParameters":
```
jyemin, line 59: With the addition of serverparameters supports, it's now necessary to break out of the loop as soon as a requirement is not met.  Otherwise we have to handle checking for a server parameter that's not supported by the server.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -140,6 +145,30 @@ OperationResult executeFind(final BsonDocument operation) {
                 new BsonArray(iterable.into(new ArrayList<>())));
     }
 
+    OperationResult executeDistinct(final BsonDocument operation) {
```
jyemin, line 148: Added support for a bunch more operations that are used in the versioned API tests.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -80,6 +87,31 @@ public UnifiedTest(final String schemaVersion, @Nullable final BsonArray runOnRe
         this.context.push(ContextElement.ofTest(definition));
     }
 
+    @NotNull
+    protected static Collection<Object[]> getTestData(final String directory) throws URISyntaxException, IOException {
```
jyemin, line 91: Refactored test data loading now that there are several different unified test runners

driver-sync/src/test/functional/com/mongodb/client/unified/VersionedApiTest.java
```diff
@@ -0,0 +1,55 @@
...
+import com.mongodb.lang.Nullable;
+import org.bson.BsonArray;
+import org.bson.BsonDocument;
+import org.junit.runners.Parameterized;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.Collection;
+
+public class VersionedApiTest extends UnifiedTest {
```
jyemin, line 31: We could just have a single unified test runner, but I think it's clearer to break them out by spec like this

driver-core/src/main/com/mongodb/MongoClientSettings.java
durran: I'm of the mindset that `Optional` should be used in these cases, and I'd think most developers should be comfortable with it, so I agree with the decision here.

driver-core/src/main/com/mongodb/ServerApiVersion.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Gets the {@code ServerApiVersion} that corresponds to the given value.
+     *
+     * @param value the String value of the desired server API version
+     * @return the corresponding {@code ServerApiVersion}
+     * @throws MongoClientException if no matching enumeration exists
+     */
+    public static ServerApiVersion findByValue(final String value) {
+        //noinspection SwitchStatementWithTooFewBranches
```
durran, line 53: Is this just future-proofing for when more API versions exist?

jyemin: Yeah, kind of silly, but I think harmless

durran (approved):
LGTM

driver-core/src/main/com/mongodb/MongoClientSettings.java
rozza: I'm unsure on this, and we probably are the only consumers of the value. To me if we only use it like so: `getServerApi().orElse(null)` then its not worth wrapping in an optional.

Also we dont consistently wrap `getServerApi` methods in optionals (ClusterBinding).

I'd argue the inconsistency also adds extra cognitive load to the code base.

driver-core/src/main/com/mongodb/ServerApi.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /**
+     * Gets whether the application requires strict server API version enforcement.
+     *
+     * <p>
+     * The default is false.
+     * </p>
+     *
+     * @return whether the application requires strict server API version enforcement
```
rozza, line 60: Is it ok not to mention its an optional?

driver-core/src/main/com/mongodb/ServerApi.java
```diff
@@ -0,0 +1,193 @@
...
+import java.util.Optional;
+
+import static com.mongodb.assertions.Assertions.notNull;
+
+/**
+ * A specification of the server API on which the application relies.
+ *
+ * @since 4.3
+ */
+public final class ServerApi {
```
rozza, line 32: Is this really a settings class? Should it be named as such?

driver-core/src/main/com/mongodb/internal/async/client/AsyncMongoClientImpl.java
```diff
@@ -67,7 +67,7 @@ private AsyncMongoClientImpl(final MongoClientSettings settings, final Cluster c
                                  @Nullable final Closeable externalResourceCloser) {
         this.settings = notNull("settings", settings);
         this.cluster = notNull("cluster", cluster);
-        this.serverSessionPool = new ServerSessionPool(cluster);
+        this.serverSessionPool = new ServerSessionPool(cluster, settings.getServerApi().orElse(null));
```
rozza, line 70: Is an optional better?

driver-core/src/main/com/mongodb/internal/binding/SingleConnectionReadBinding.java
rozza: Ack :+1:

driver-core/src/main/com/mongodb/internal/binding/SingleServerBinding.java
```diff
@@ -79,6 +73,11 @@ public SessionContext getSessionContext() {
         return NoOpSessionContext.INSTANCE;
     }
 
+    @Override
+    public ServerApi getServerApi() {
```
rozza, line 77: Returns a nullable rather than an option.

driver-core/src/main/com/mongodb/internal/binding/WriteBinding.java
```diff
@@ -40,6 +41,8 @@
      */
     SessionContext getSessionContext();
 
+    ServerApi getServerApi();
```
rozza, line 44: Returns a nullable rather than an option.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
rozza: Perhaps add a comment inline to explain its not needed.

driver-core/src/test/functional/com/mongodb/internal/connection/AsyncStreamTimeoutsSpecification.groovy
```diff
@@ -50,7 +50,7 @@ class AsyncStreamTimeoutsSpecification extends OperationFunctionalSpecification
         given:
         def connection = new InternalStreamConnectionFactory(
                 new AsynchronousSocketChannelStreamFactory(openSocketSettings, getSslSettings()), getCredentialWithCache(), null, null,
-                [], null).create(new ServerId(new ClusterId(), new ServerAddress(new InetSocketAddress('192.168.255.255', 27017))))
+                [], null,).create(new ServerId(new ClusterId(), new ServerAddress(new InetSocketAddress('192.168.255.255', 27017))))
```
rozza, line 53: Surprised codenarc allows this?

driver-sync/src/main/com/mongodb/client/internal/CryptConnection.java
```diff
@@ -94,7 +95,7 @@ public ConnectionDescription getDescription() {
     @Override
     public <T> T command(final String database, final BsonDocument command, final FieldNameValidator commandFieldNameValidator,
                          final ReadPreference readPreference, final Decoder<T> commandResultDecoder, final SessionContext sessionContext,
-                         final boolean responseExpected, @Nullable final SplittablePayload payload,
+                         final ServerApi serverApi, final boolean responseExpected, @Nullable final SplittablePayload payload,
```
rozza, line 98: Mark as nullable?

driver-sync/src/main/com/mongodb/client/internal/CryptConnection.java
```diff
@@ -127,8 +128,10 @@ public ConnectionDescription getDescription() {
 
     @Override
     public <T> T command(final String database, final BsonDocument command, final FieldNameValidator fieldNameValidator,
-                         final ReadPreference readPreference, final Decoder<T> commandResultDecoder, final SessionContext sessionContext) {
-        return command(database, command, fieldNameValidator, readPreference, commandResultDecoder, sessionContext, true, null, null);
+                         final ReadPreference readPreference, final Decoder<T> commandResultDecoder, final SessionContext sessionContext,
+                         final ServerApi serverApi) {
```
rozza, line 132: Mark as nullable?

driver-sync/src/test/functional/com/mongodb/client/unified/ErrorMatcher.java
rozza: Ack

driver-sync/src/test/functional/com/mongodb/client/unified/RunOnRequirementsMatcher.java
```diff
@@ -21,37 +21,62 @@
...
-            if (requirement.containsKey("maxServerVersion")
-                    && serverVersion.compareTo(getMaxServerVersionForField("maxServerVersion", requirement)) > 0) {
-                requirementMet = false;
-            }
 
-            if (requirement.containsKey("topologies")) {
-                BsonArray topologyTypes = requirement.getArray("topologies");
-                if (!topologyMatches(topologyTypes)) {
-                    requirementMet = false;
+            requirementLoop:
```
rozza, line 37: TIL about java labels!

driver-sync/src/test/functional/com/mongodb/client/unified/VersionedApiTest.java
rozza: Ack

rozza (changes requested):
Looking really good. Found a couple of nits.

driver-core/src/main/com/mongodb/internal/async/client/AsyncMongoClientImpl.java
jyemin: IntelliJ actually complains when you specify an Optional as a parameter:

> Optional was designed to provide a limited mechanism for library method return types where there needed to be a clear way to represent "no result".

driver-core/src/main/com/mongodb/internal/binding/SingleServerBinding.java
jyemin: You want this to return `Optional<ServerApi>`?  

I think I didn't bother because it's not a public API

driver-core/src/test/functional/com/mongodb/internal/connection/AsyncStreamTimeoutsSpecification.groovy
jyemin: I'll check it out

driver-core/src/main/com/mongodb/ServerApi.java
jyemin: I think of it more like ReadPreference or WriteConcern, which are also aggregates of related settings.  The name is also proscribed in the spec.

driver-core/src/main/com/mongodb/ServerApi.java
jyemin: I think it's ok.  The type of the return tell you that, which is convenient.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -177,7 +182,12 @@ protected EncodingMetadata encodeMessageBodyWithMetadata(final BsonOutput bsonOu
             commandStartPosition = bsonOutput.getPosition();
 
             if (payload == null) {
```
rozza, line 184: Could the payload be part of a transaction? If so should it be added here?

driver-core/src/main/com/mongodb/internal/async/client/AsyncMongoClientImpl.java
rozza: Sorry I was really asking if using an optional in the settings is actually better / beneficial

driver-core/src/main/com/mongodb/ServerApi.java
rozza: Ack

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: Yeah, this is confusing.  Here's the issue: for the most part passing serverApi to OP_QUERY (i.e. pre 3.6 servers) is irrelevant, because those servers don't support server api (or transactions) anyway.  But we do need to pass server api via OP_QUERY in one circumstance: when sending the initial isMaster command as part of the handshake.  Because we don't know what server version version it is until we get the isMaster response, we always use OP_QUERY for that isMaster command.  For simplicity, and because we don't actually know the command name here until we serialize it, this code just always adds server api for regular OP_QUERY commands.  But doing the same for ones with payload would require more work, and wouldn't give a benefit, because the server we'd be sending it to doesn't support server api anyway.

Maybe an inline comment explaining this would help?

rozza: Ack - that makes sense. Its worth a comment explaining its added incase its an isMaster call.

driver-core/src/main/com/mongodb/internal/async/client/AsyncMongoClientImpl.java
jyemin: Right.  I don't think it is, and am removing it as per your suggestion

driver-core/src/main/com/mongodb/internal/binding/SingleServerBinding.java
jyemin: Add @Nullable to all of these.

Done

driver-core/src/main/com/mongodb/internal/binding/WriteBinding.java
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: Done

driver-core/src/main/com/mongodb/MongoClientSettings.java
jyemin: Ross persuaded me.  I've remove use of Optional here and in MongoClientOptions

driver-core/src/main/com/mongodb/internal/async/client/AsyncMongoClientImpl.java
jyemin: Done

driver-core/src/test/functional/com/mongodb/internal/connection/AsyncStreamTimeoutsSpecification.groovy
jyemin: Fixed.

driver-sync/src/main/com/mongodb/client/internal/CryptConnection.java
jyemin: Done

driver-sync/src/main/com/mongodb/client/internal/CryptConnection.java
jyemin: Done

rozza (approved):

alcaeus (approved):

jyemin:
Parked in https://github.com/mongodb/mongo-java-driver/tree/j3878 until ready to push to master.  Will have to add reactive support before that happens.

PR #611, opened 2020-11-19
Add Bson#toBsonDocument() overload with a default implementation

jyemin:
The default implementation simply calls Bson#toBsonDocument(Class<T>, CodecRegistry)

JAVA-3885

### DIFF
--- a/bson-scala/src/main/scala/org/mongodb/scala/bson/collection/BaseDocument.scala
+++ b/bson-scala/src/main/scala/org/mongodb/scala/bson/collection/BaseDocument.scala
@@ -242,7 +242,7 @@ private[bson] trait BaseDocument[T] extends Traversable[(String, BsonValue)] wit
   /**
    * Returns a copy of the underlying BsonDocument
    */
-  def toBsonDocument: BsonDocument = copyBsonDocument()
+  override def toBsonDocument: BsonDocument = copyBsonDocument()
 
   override def toBsonDocument[TDocument](documentClass: Class[TDocument], codecRegistry: CodecRegistry): BsonDocument =
     underlying
--- a/bson-scala/src/main/scala/org/mongodb/scala/bson/collection/BaseDocument.scala
+++ b/bson-scala/src/main/scala/org/mongodb/scala/bson/collection/BaseDocument.scala
@@ -239,10 +239,7 @@ private[bson] trait BaseDocument[T] extends Traversable[(String, BsonValue)] wit
    */
   def toJson(settings: JsonWriterSettings): String = underlying.toJson(settings)
 
-  /**
-   * Returns a copy of the underlying BsonDocument
-   */
-  override def toBsonDocument: BsonDocument = copyBsonDocument()
+  override def toBsonDocument: BsonDocument = underlying
 
   override def toBsonDocument[TDocument](documentClass: Class[TDocument], codecRegistry: CodecRegistry): BsonDocument =
     underlying
--- a/bson/src/main/org/bson/RawBsonDocument.java
+++ b/bson/src/main/org/bson/RawBsonDocument.java
@@ -226,17 +226,17 @@ public final class RawBsonDocument extends BsonDocument {
 
     @Override
     public Set<Entry<String, BsonValue>> entrySet() {
-        return toBsonDocument().entrySet();
+        return toBaseBsonDocument().entrySet();
     }
 
     @Override
     public Collection<BsonValue> values() {
-        return toBsonDocument().values();
+        return toBaseBsonDocument().values();
     }
 
     @Override
     public Set<String> keySet() {
-        return toBsonDocument().keySet();
+        return toBaseBsonDocument().keySet();
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


bson/src/main/org/bson/RawBsonDocument.java
```diff
@@ -349,7 +349,8 @@ private BsonBinaryReader createReader() {
         return new BsonBinaryReader(new ByteBufferBsonInput(getByteBuffer()));
     }
 
-    private BsonDocument toBsonDocument() {
```
jyemin, line 352: Had to change the name of this private method to avoid name conflict with the new method.

bson/src/test/unit/org/bson/DocumentTest.java
```diff
@@ -131,6 +132,18 @@ public void toJsonShouldTakeACustomDocumentCodec() {
         assertEquals("{\"database\": {\"name\": \"MongoDB\"}}", customDocument.toJson(customDocumentCodec));
     }
 
+    @Test
+    public void toBsonDocumentShouldCreateBsonDocument() {
```
jyemin, line 136: This was the most convenient place to test.

rozza (approved):

jyemin:
@rozza I jumped the gun.  There is a compilation error in ByteBufBsonDocument, which I fixed.  But bson-scala also fails with:

```
> Task :bson-scala:compileScala
[Error] /Users/jeff.yemin/git/mongo-java-driver/bson-scala/src/main/scala-2.13+/org/mongodb/scala/bson/collection/immutable/Document.scala:102: class Document inherits conflicting members:
  <defaultmethod> def toBsonDocument(): org.bson.BsonDocument (defined in trait Bson) and
  def toBsonDocument(): org.mongodb.scala.bson.BsonDocument (defined in trait BaseDocument)
  (note: this can be resolved by declaring an `override` in class Document.)
[Error] /Users/jeff.yemin/git/mongo-java-driver/bson-scala/src/main/scala-2.13+/org/mongodb/scala/bson/collection/mutable/Document.scala:103: class Document inherits conflicting members:
  <defaultmethod> def toBsonDocument(): org.bson.BsonDocument (defined in trait Bson) and
  def toBsonDocument(): org.mongodb.scala.bson.BsonDocument (defined in trait BaseDocument)
  (note: this can be resolved by declaring an `override` in class Document.)
[Error] /Users/jeff.yemin/git/mongo-java-driver/bson-scala/src/main/scala/org/mongodb/scala/bson/collection/BaseDocument.scala:245: `override` modifier required to override concrete member:
<defaultmethod> def toBsonDocument(): org.bson.BsonDocument (defined in trait Bson)
three errors found
```

Not quite sure yet if that can be fixed....

rozza:
Try:` override def toBsonDocument()`

bson-scala/src/main/scala/org/mongodb/scala/bson/collection/BaseDocument.scala
```diff
@@ -242,7 +242,7 @@ private[bson] trait BaseDocument[T] extends Traversable[(String, BsonValue)] wit
   /**
    * Returns a copy of the underlying BsonDocument
    */
-  def toBsonDocument: BsonDocument = copyBsonDocument()
+  override def toBsonDocument: BsonDocument = copyBsonDocument()
```
jyemin, line 245: I'm not sure about this override.  

`toBsonDocument[TDocument](documentClass: Class[TDocument], codecRegistry: CodecRegistry): BsonDocument`  returns `this`, essentially

But `BaseDocument#toBsonDocument `returns a copy of itself

But the intention in `Bson` is to make those two methods essentially identical in behavior

rozza: I think thats fine, it should return `underlying`.  Note that `ImmutableDocumentCodec#generateIdIfAbsentFromDocument` will have to be updated to clone the document so it doesn't mutate the original:

```
  override def generateIdIfAbsentFromDocument(document: Document): Document = {
    if (!underlying.documentHasId(document.underlying)) {
      Document(underlying.generateIdIfAbsentFromDocument(document.toBsonDocument.clone()))
    } else {
      document
    }
  }
```

jyemin: The doc comment says that it returns a copy.  Could this change break clients that are assuming that it returns a copy?

rozza: I doubt users use the method, its really there for the codecs. Alternatively, you'll have to leave as is.

jyemin: Thanks.  I looked for other usages and found:

```
trait LowPrio {

  /**
   * Transforms `immutable.Document` to `BsonDocument`
   */
  implicit object TransformImmutableDocument extends BsonTransformer[IDocument] {
    def apply(value: IDocument): BsonDocument = value.toBsonDocument
  }
   ...
```

Does that have to change as well?

rozza: No, thats fine

jyemin: Done

rozza (approved):

PR #614, opened 2020-11-25
Builders for aggregate expression

jyemin:
Support a representative sample of expressions, suitable for a draft PR

 * literal expressions, parsed and unparsed
 * object and array expressions
 * field path expressions
 * variable reference expressions
 * custom expressions
 * add expressions
 * gte expressions
 * switch expressions

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/Aggregates.java
+++ b/driver-core/src/main/com/mongodb/client/model/Aggregates.java
@@ -28,6 +28,8 @@ import org.bson.codecs.configuration.CodecRegistry;
 import org.bson.conversions.Bson;
 
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import static java.util.Arrays.asList;
 import static org.bson.assertions.Assertions.notNull;
@@ -65,6 +67,23 @@ public final class Aggregates {
      */
     public static Bson addFields(final List<Field<?>> fields) {
         return new AddFieldsStage(fields);
+    }
+
+    /**
+     * Creates an $addFields pipeline stage
+     *
+     * @param fields        the fields to add
+     * @param <TExpression> the expression type
+     * @return the $addFields pipeline stage
+     * @see com.mongodb.client.model.expressions.Expressions
+     * @mongodb.driver.manual reference/operator/aggregation/addFields/ $addFields
+     * @mongodb.server.release 3.4
+     * @since 4.?
+     */
+    public static <TExpression> Bson addFields(final Map<String, TExpression> fields) {
+        return new AddFieldsStage(fields.entrySet().stream()
+                .map(entry -> new Field<>(entry.getKey(), entry.getValue()))
+                .collect(Collectors.toList()));
     }
 
     /**
--- a/driver-core/src/main/com/mongodb/client/model/Projections.java
+++ b/driver-core/src/main/com/mongodb/client/model/Projections.java
@@ -25,6 +25,7 @@ import org.bson.codecs.configuration.CodecRegistry;
 import org.bson.conversions.Bson;
 
 import java.util.List;
+import java.util.Map;
 
 import static com.mongodb.assertions.Assertions.notNull;
 import static java.util.Arrays.asList;
@@ -55,6 +56,21 @@ public final class Projections {
      */
     public static <TExpression> Bson computed(final String fieldName, final TExpression expression) {
         return new SimpleExpression<TExpression>(fieldName, expression);
+    }
+
+    /**
+     * Creates a projection of fields computed from the given expressions. Projection with an expression is only supported
+     * using the $project aggregation pipeline stage.
+     *
+     * @param fields     the map of fields to expressions
+     * @param <TExpression> the expression type
+     * @return the projection
+     * @see Aggregates#project(Bson)
+     * @see com.mongodb.client.model.expressions.Expressions
+     * @since 4.?
+     */
+    public static <TExpression> Bson computed(final Map<String, TExpression> fields) {
+        return new CompoundExpression<>(fields);
     }
 
     /**
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Branch.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Branch.java
@@ -16,7 +16,7 @@
 
 package com.mongodb.client.model.expressions;
 
-import jdk.nashorn.internal.ir.annotations.Immutable;
+import com.mongodb.annotations.Immutable;
 
 import java.util.Objects;
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Branch.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Branch.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+import jdk.nashorn.internal.ir.annotations.Immutable;
+
+import java.util.Objects;
+
+import static com.mongodb.assertions.Assertions.notNull;
+
+@Immutable
+public final class Branch {
+    private final Expression caseExpr;
+    private final Expression thenExpr;
+
+    Branch(final Expression caseExpr, final Expression thenExpr) {
+        this.caseExpr = notNull("case", caseExpr);
+        this.thenExpr = notNull("then", thenExpr);
+    }
+
+    public Expression getCase() {
+        return caseExpr;
+    }
+
+    public Expression getThen() {
+        return thenExpr;
+    }
+
+    @Override
+    public String toString() {
+        return "Branch{" +
+                "case=" + caseExpr +
+                ", then=" + thenExpr +
+                '}';
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Branch branch = (Branch) o;
+        return Objects.equals(caseExpr, branch.caseExpr) &&
+                Objects.equals(thenExpr, branch.thenExpr);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(caseExpr, thenExpr);
+    }
+}
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+import org.bson.BsonValue;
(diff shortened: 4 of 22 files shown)


driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -67,6 +69,23 @@ public static Bson addFields(final List<Field<?>> fields) {
...
+     *
+     * @param fields        the fields to add
+     * @param <TExpression> the expression type
+     * @return the $addFields pipeline stage
+     * @see com.mongodb.client.model.expressions.Expressions
+     * @mongodb.driver.manual reference/operator/aggregation/addFields/ $addFields
+     * @mongodb.server.release 3.4
+     * @since 4.?
+     */
+    public static <TExpression> Bson addFields(final Map<String, TExpression> fields) {
```
jyemin, line 83: This overload lets you use expressions more easily, where each value in the map is an expression, e.g.

    addFields(Map.of(
                        "foo", add(...),
                        "bar", switch(...)
                      ))

driver-core/src/main/com/mongodb/client/model/Projections.java
```diff
@@ -57,6 +58,21 @@ private Projections() {
...
+     * using the $project aggregation pipeline stage.
+     *
+     * @param fields     the map of fields to expressions
+     * @param <TExpression> the expression type
+     * @return the projection
+     * @see Aggregates#project(Bson)
+     * @see com.mongodb.client.model.expressions.Expressions
+     * @since 4.?
+     */
+    public static <TExpression> Bson computed(final Map<String, TExpression> fields) {
```
jyemin, line 72: Similar to the addFields overload, this makes it easier to project multiple computed fields.  Perhaps this method should be called `of`.

driver-core/src/main/com/mongodb/client/model/expressions/Branch.java
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+import java.util.Objects;
+
+import static com.mongodb.assertions.Assertions.notNull;
+
+public final class Branch {
```
jyemin, line 23: Helper class to support creating switch expressions

driver-core/src/main/com/mongodb/client/model/expressions/ExpressionHelper.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.mongodb.client.model.expressions;
+
+import org.bson.BsonArray;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+interface ExpressionHelper {
```
jyemin, line 25: The new way of writing helper classes ;)

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -0,0 +1,252 @@
...
+
+    static LiteralExpression literal(final int value) {
+        return new LiteralExpression(new BsonInt32(value), true);
+    }
+
+    static LiteralExpression literal(final long value) {
+        return new LiteralExpression(new BsonInt64(value), true);
+    }
+
+    static LiteralExpression literal(final Date value) {
```
jyemin, line 70: Maybe it's stupid to have these overloads, which can create ambiguity with the one below that takes Object.  It means that if you want a null literal you need to write `literal((Object) null)`.  Alternatively could create a special method for null literals, e.g. `nullLiteral()`.

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
```diff
@@ -0,0 +1,117 @@
...
+        this.defaultExpr = defaultExpr;
+    }
+
+    /**
+     * The path to take if no branch case expression evaluates to true.
+     *
+     * @param defaultExpr an expression to evaluate if no branch evaluates to true
+     * @return a new SwitchExpression
+     */
+    public SwitchExpression defaultExpr(final Expression defaultExpr) {
```
jyemin, line 56: This is what drove me initially to create concrete classes for each expression.  With this, you can write:

      switchExpr(branch(...), branch(...)).defaultExpr(...)

without it you need to start adding overloads to the builder, e.g.

      Expression switchExpr(Branch ... branch)
      Expression switchExpr(List<Branch> branches)
      Expression switchExpr(List<Branch> branches, Expression defaultExpr)

which seemed like it would create a lot of noise in the builder class.  This pattern is similar to how we define iterable interfaces and options classes, where the required arguments go in the initiating method, and optional arguments are chained.

driver-core/src/main/com/mongodb/client/model/expressions/codecs/ExpressionCodec.java
```diff
@@ -0,0 +1,70 @@
...
+import org.bson.BsonWriter;
+import org.bson.codecs.BsonValueCodecProvider;
+import org.bson.codecs.Codec;
+import org.bson.codecs.DecoderContext;
+import org.bson.codecs.EncoderContext;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
+
+public final class ExpressionCodec implements Codec<Expression> {
```
jyemin, line 31: Similar to `BsonCodec`

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.stream.Collectors;
+
+import static com.mongodb.assertions.Assertions.notNull;
+import static java.util.Collections.unmodifiableList;
+
+/**
+ * Evaluates a series of case expressions. When it finds an expression which evaluates to true, $switch executes a specified expression
+ * and breaks out of the control flow.
+ *
+ * @since 4.?
```
jyemin, line 37: Will update once we know...

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
```diff
@@ -0,0 +1,117 @@
...
+ */
+@Immutable
```
jyemin, line 39: All expressions should be immutable

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
```diff
@@ -0,0 +1,117 @@
...
+public final class SwitchExpression implements Expression {
+    private final List<Branch> branches;
+    @Nullable
+    private final Expression defaultExpr;
+
+    SwitchExpression(final List<Branch> branches, @Nullable final Expression defaultExpr) {
```
jyemin, line 45: Constructors package-protected to force uniform use of builders

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
```diff
@@ -0,0 +1,117 @@
...
+        this.branches = notNull("branches", branches);
+        this.defaultExpr = defaultExpr;
+    }
+
+    /**
+     * The path to take if no branch case expression evaluates to true.
+     *
+     * @param defaultExpr an expression to evaluate if no branch evaluates to true
+     * @return a new SwitchExpression
```
jyemin, line 54: Preserves immutability

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
```diff
@@ -0,0 +1,117 @@
...
+        return new SwitchExpression(branches, defaultExpr);
+    }
+
+    /**
+     * An array of control branch documents.
+     *
+     * @return a list of branches, which may not be null.
+     */
+    public List<Branch> getBranches() {
+        return unmodifiableList(branches);
```
jyemin, line 66: Preserves immutability

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
```diff
@@ -0,0 +1,117 @@
...
+                        .collect(Collectors.toCollection(BsonArray::new)));
+        if (defaultExpr != null) {
+            switchDocument.append("default", defaultExpr.toBsonValue(codecRegistry));
+        }
+        return new BsonDocument("$switch", switchDocument);
+    }
+
+    @Override
+    public String toString() {
+        return "SwitchExpression{" +
```
jyemin, line 94: Side note: how does everyone feel about turning off the checkstyle rules that make the generated toString and equals methods fail the check?  

 * `+` and `&&` on new line
 * braces required on `if` (this one I feel kind of tied to)

It's really annoying to have to modify all the generated method (I didn't do it yet, so checkstyle actually fails now)

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -0,0 +1,51 @@
...
+ * @since 4.?
+ */
+public interface Expression {
+    /**
+     * Convert an expression to a BSON value
+     *
+     * @param codecRegistry the registry to use to convert literal expressions
+     * @return the BSON value represented by this expression
+     */
+    BsonValue toBsonValue(CodecRegistry codecRegistry);
```
jyemin, line 37: Unlike aggregation stages, expressions can be of any BSON type, even including null

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
durran: I'm fine with turning of the `+` and `&&` on a new line - I agree braces required on `if` could probably stay.

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
durran: I'd lean towards classes for each expression, for the same reasons you state and the unit tests would end up being more focused on a single expression per test class, which I find cleaner and easier to read.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
rozza: Is order important?

driver-core/src/main/com/mongodb/client/model/expressions/Branch.java
```diff
@@ -0,0 +1,68 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+import jdk.nashorn.internal.ir.annotations.Immutable;
```
rozza, line 19: Is this an import we want?

driver-core/src/main/com/mongodb/client/model/expressions/ExpressionHelper.java
rozza: For my edification, why prefer using an interface over a class that can't be constructed?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
rozza: I think nullLiteral works, I'd rather be type friendly and explicit.

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
rozza: Personally, I would be happy to just use the intellij defaults for these auto generated overrides.

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
rozza: Agreed.

rozza (commented):
I'm liking the approach :+1:

driver-core/src/main/com/mongodb/client/model/expressions/Branch.java
jyemin: Good catch.  Replaced with the correct one in all the files that had it.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
jyemin: No, there's no reason I can think of that order should matter here, and in general it shouldn't be relied on as the server doesn't guarantee that order is preserved.

driver-core/src/main/com/mongodb/client/model/expressions/ExpressionHelper.java
jyemin: It's just a bit cleaner since you don't have to take extra steps to ensure non-instantiability (with final) or non-constructability (with a private constructor)

driver-core/src/main/com/mongodb/client/model/expressions/SwitchExpression.java
jyemin: Ok, I'll see if there's a way to disable the checks just for these methods.

jyemin:
Closing this for now.  Thanks everyone for the feedback.,

PR #617, opened 2020-12-08
Support the versioned API in reactive streams driver

jyemin:
This completes the work for versioned API.

This PR is based off of the previous versioned API PR, which I've already rebased on master, so this includes the Project Reactor changes.

JAVA-3878


driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -139,45 +137,31 @@ public T apply(final T t, final AsyncConnectionSource source, final AsyncConnect
     }
 
     static CommandWriteTransformer<BsonDocument, Void> writeConcernErrorTransformer() {
```
jyemin, line 139: This file had so many issues I just decided to modernize it completely

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -236,33 +215,30 @@ public T call(final ConnectionSource source) {
...
-                    if (!canRetryRead(source.getServerDescription(), connection.getDescription(), binding.getSessionContext())) {
-                        throw originalException;
-                    }
-                    BsonDocument retryCommand = commandCreator.create(source.getServerDescription(), connection.getDescription());
-                    logRetryExecute(retryCommand.getFirstKey(), originalException);
-                    return executeCommand(database, retryCommand, decoder, source, connection, binding.getReadPreference(), transformer,
-                            binding.getSessionContext(), source.getServerApi());
-                } finally {
-                    connection.release();
+        return withReleasableConnection(binding, originalException, (source1, connection1) -> {
```
jyemin, line 218: `connection1` is a bit awkward after lambdafy-ing.  Maybe extract a method?

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -562,78 +524,12 @@ public void onResult(final T result, final Throwable t) {
 
     /* Async Write Binding Helpers */
 
-    static void executeCommandAsync(final AsyncWriteBinding binding,
```
jyemin, line 565: No longer used after removing other unused stuff

driver-core/src/main/com/mongodb/internal/operation/CommandWriteOperation.java
```diff
@@ -1,62 +0,0 @@
...
-import static com.mongodb.internal.operation.CommandOperationHelper.executeCommand;
-import static com.mongodb.internal.operation.CommandOperationHelper.executeCommandAsync;
-
-/**
- * An operation that executes an arbitrary command that writes to the server.
- *
- * @param <T> the operations result type.
- * @since 3.0
- */
-public class CommandWriteOperation<T> implements AsyncWriteOperation<T>, WriteOperation<T> {
```
jyemin, line 35: This class was unused, and once I ditched it a lot of the CommandOperationHelper methods were no longer used.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
jyemin: I went ahead and extracted a `retryCommand` method.  I like it better now.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -227,42 +206,44 @@ public T call(final ConnectionSource source) {
 
             if (!shouldAttemptToRetryRead(retryReads, e)) {
                 if (retryReads) {
-                    logUnableToRetry(command.getFirstKey(), e);
+                    logUnableToRetry(Objects.requireNonNull(command).getFirstKey(), e);
```
rozza, line 209: If command is null then both `command.getFirstKey()` and `Objects.requireNonNull(command).getFirstKey()` act the same.  Which is probably isn't desired, I'd imagine if command is null then its better to rethrow the original exception rather than a NPE.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -227,42 +206,44 @@ public T call(final ConnectionSource source) {
...
+    private static <D, T> T retryCommand(final ReadBinding binding, final String database, final CommandCreator commandCreator,
+                                         final Decoder<D> decoder, final CommandReadTransformer<D, T> transformer,
+                                         final MongoException originalException) {
+        return withReleasableConnection(binding, originalException, (source, connection) -> {
+            try {
+                if (!canRetryRead(source.getServerDescription(), connection.getDescription(), binding.getSessionContext())) {
+                    throw originalException;
                 }
+                BsonDocument retryCommand = commandCreator.create(source.getServerDescription(), connection.getDescription());
+                logRetryExecute(retryCommand.getFirstKey(), originalException);
```
rozza, line 229: Note: No null check here.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -227,42 +206,44 @@ public T call(final ConnectionSource source) {
...
-                            binding.getSessionContext(), source.getServerApi());
-                } finally {
-                    connection.release();
+        return retryCommand(binding, database, commandCreator, decoder, transformer, exception);
+    }
+
+    private static <D, T> T retryCommand(final ReadBinding binding, final String database, final CommandCreator commandCreator,
+                                         final Decoder<D> decoder, final CommandReadTransformer<D, T> transformer,
+                                         final MongoException originalException) {
+        return withReleasableConnection(binding, originalException, (source, connection) -> {
```
rozza, line 223: Suggestion: I think in the future we should look to clean up these loan methods `withConnection` and `withReleasableConnection`.  So that they are inline with `withReadConnectionSource` and handle the try/finally releasing.  Reduces the boilerplate and risk of not releasing the connection.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
rozza: Nice

rozza (commented):
Looks good one comment and one suggestion for future work.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
jyemin: Good catch.  I blindly followed IntelliJ's advice here instead of seeing what the real problem was.  There is no way that command should ever be null, and if it is we should certainly not retry. Pulling the initialization of command outside the try block gets rids of the warning.

jyemin: Done

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
jyemin: See above.  No need for it here either.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
jyemin: Ack.

rozza (approved):

PR #619, opened 2020-12-15
JAVA-3331.  InsertOneOptions and InsertManyOptions are final

HbnKing:
this is my fix for the JIRA ticket for the issue [JAVA-3331](https://jira.mongodb.org/browse/JAVA-3331). 
it seems like a simple fix, and essential to  extend them and add things like WriteConcern and ReadPreference options .


jyemin:
Hi @HbnKing,

Please see my comment on JAVA-3331.  I don't think we should proceed with this change.

Thanks,
Jeff

PR #627, opened 2021-01-11
Scala: Ensure implicit conversion keeps type fidelity

rozza:
When converting from a BsonDocument to org.bson.Document
the scala driver previously used json. The implementation
should not rely on the json output format, but rather use
the Codecs to do the conversion.

JAVA-3916

### DIFF
--- a/driver-scala/src/main/scala/org/mongodb/scala/package.scala
+++ b/driver-scala/src/main/scala/org/mongodb/scala/package.scala
@@ -16,17 +16,13 @@
 
 package org.mongodb
 
-import org.bson.codecs.{ BsonDocumentCodec, DecoderContext, DocumentCodec, EncoderContext }
-import org.bson.{ BsonBinaryReader, BsonBinaryWriter, ByteBufNIO }
-import org.bson.io.{ BasicOutputBuffer, ByteBufferBsonInput }
-import org.bson.json.{ JsonMode, JsonWriterSettings }
-
-import _root_.scala.language.implicitConversions
-import _root_.scala.reflect.ClassTag
+import org.bson.BsonDocumentReader
+import org.bson.codecs.{ DecoderContext, DocumentCodec }
 import org.mongodb.scala.bson.BsonDocument
 import org.mongodb.scala.internal.WriteConcernImplicits
 
-import java.nio.ByteBuffer
+import _root_.scala.language.implicitConversions
+import _root_.scala.reflect.ClassTag
 
 /**
  * The MongoDB Scala Driver package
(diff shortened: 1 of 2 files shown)


driver-scala/src/test/scala/org/mongodb/scala/ScalaPackageSpec.scala
```diff
@@ -141,4 +140,33 @@ class ScalaPackageSpec extends BaseSpec {
     val javaCredential5 = JMongoCredential.createGSSAPICredential("userName")
     scalaCredential5 should equal(javaCredential5)
   }
+
+  it should "implicitly convert to org.Bson.document with type fidelity" in {
```
jyemin, line 144: Lowercase the `B`

driver-scala/src/main/scala/org/mongodb/scala/package.scala
```diff
@@ -392,9 +399,18 @@ package object scala extends ClientSessionImplicits with ObservableImplicits wit
 
   implicit def bsonDocumentToDocument(doc: BsonDocument): Document = new Document(doc)
 
-  implicit def bsonDocumentToUntypedDocument(doc: BsonDocument): org.bson.Document =
-    org.bson.Document.parse(doc.toJson())
+  implicit def documentToUntypedDocument(doc: Document): org.bson.Document =
+    bsonDocumentToUntypedDocument(doc.underlying)
+
+  implicit def bsonDocumentToUntypedDocument(doc: BsonDocument): org.bson.Document = {
+    val writer: BsonBinaryWriter = new BsonBinaryWriter(new BasicOutputBuffer())
```
jyemin, line 406: You can use BsonDocumentReader and simplify to

    new DocumentCodec().decode(new BsonDocumentReader(doc), DecoderContext.builder().build())

jyemin (changes requested):

driver-scala/src/main/scala/org/mongodb/scala/package.scala
rozza: Much nicer thanks

driver-scala/src/main/scala/org/mongodb/scala/package.scala
```diff
@@ -392,9 +399,18 @@ package object scala extends ClientSessionImplicits with ObservableImplicits wit
...
+    new BsonDocumentCodec().encode(writer, doc, EncoderContext.builder().build())
```
stIncMale, line 407: I am not sure, but at first glance`BsonDocumentCodec` looks like an immutable class. If that is the case, do we actually want to create a new object per the `bsonDocumentToUntypedDocument` method invocation? Unless we are fortunate and [scalar replacement](https://shipilev.net/jvm/anatomy-quarks/18-scalar-replacement/) kicks in, this creates unnecessary garbage.

rozza: Good call - done.

jyemin (approved):

driver-scala/src/test/scala/org/mongodb/scala/ScalaPackageSpec.scala
```diff
@@ -141,4 +140,33 @@ class ScalaPackageSpec extends BaseSpec {
     val javaCredential5 = JMongoCredential.createGSSAPICredential("userName")
     scalaCredential5 should equal(javaCredential5)
   }
+
+  it should "implicitly convert to org.bson.document with type fidelity" in {
+
+    val bsonDocument = Document(
+      "null" -> BsonNull(),
+      "int32" -> BsonInt32(32),
+      "int64" -> BsonInt64(64L),
```
aaabramov, line 149: I'd change `64L` -> `Long.MaxValue` for reproducing issue in JIRA ticket.

FYI, I am OP of the bug

rozza: Sounds good

driver-scala/src/test/scala/org/mongodb/scala/ScalaPackageSpec.scala
```diff
@@ -141,4 +140,33 @@ class ScalaPackageSpec extends BaseSpec {
...
     scalaCredential5 should equal(javaCredential5)
   }
+
+  it should "implicitly convert to org.bson.document with type fidelity" in {
+
+    val bsonDocument = Document(
+      "null" -> BsonNull(),
+      "int32" -> BsonInt32(32),
+      "int64" -> BsonInt64(Long.MaxValue),
+      "decimal128" -> BsonDecimal128(128),
```
aaabramov, line 150: Also, I would explicitly pass 128.1 as with BsonDouble.

PR #630, opened 2021-01-12
Create internal MongoClient for auto-encryption

jyemin:
JAVA-3932

During the testing of the reactive streams implementation, I discovered a bug in ListCollectionsOperation in which two connections were checked out concurrently (and only one used).  I've fixed that here as well.

### DIFF
--- a/driver-core/src/main/com/mongodb/MongoClientSettings.java
+++ b/driver-core/src/main/com/mongodb/MongoClientSettings.java
@@ -477,12 +477,23 @@ public final class MongoClientSettings {
         /**
          * Sets the auto-encryption settings
          *
+         * A separate, internal {@code MongoClient} is created if any of the following are true:
+         *
+         * <ul>
+         *    <li>{@code AutoEncryptionSettings.keyVaultClient} is not passed</li>
+         *    <li>{@code AutoEncryptionSettings.bypassAutomaticEncryption} is {@code false}</li>
+         * </ul>
+         *
+         * If an internal {@code MongoClient} is created, it is configured with the same
+         * options as the parent {@code MongoClient} except {@code minPoolSize} is set to {@code 0}
+         * and {@code AutoEncryptionSettings} is omitted.
+         *
          * @param autoEncryptionSettings the auto-encryption settings
          * @return this
          * @since 3.11
          * @see #getAutoEncryptionSettings()
          */
-        public Builder autoEncryptionSettings(final AutoEncryptionSettings autoEncryptionSettings) {
+        public Builder autoEncryptionSettings(@Nullable final AutoEncryptionSettings autoEncryptionSettings) {
             this.autoEncryptionSettings = autoEncryptionSettings;
             return this;
         }
--- a/driver-core/src/main/com/mongodb/internal/operation/ListCollectionsOperation.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/ListCollectionsOperation.java
@@ -21,10 +21,10 @@ import com.mongodb.MongoNamespace;
 import com.mongodb.ReadPreference;
 import com.mongodb.ServerAddress;
 import com.mongodb.ServerCursor;
-import com.mongodb.internal.async.AsyncBatchCursor;
-import com.mongodb.internal.async.SingleResultCallback;
 import com.mongodb.connection.ConnectionDescription;
 import com.mongodb.connection.ServerDescription;
+import com.mongodb.internal.async.AsyncBatchCursor;
+import com.mongodb.internal.async.SingleResultCallback;
 import com.mongodb.internal.binding.AsyncConnectionSource;
 import com.mongodb.internal.binding.AsyncReadBinding;
 import com.mongodb.internal.binding.ConnectionSource;
@@ -56,7 +56,7 @@ import static com.mongodb.assertions.Assertions.notNull;
 import static com.mongodb.connection.ServerType.SHARD_ROUTER;
 import static com.mongodb.internal.async.ErrorHandlingResultCallback.errorHandlingCallback;
 import static com.mongodb.internal.operation.CommandOperationHelper.CommandCreator;
-import static com.mongodb.internal.operation.CommandOperationHelper.executeCommandAsync;
+import static com.mongodb.internal.operation.CommandOperationHelper.executeCommandAsyncWithConnection;
 import static com.mongodb.internal.operation.CommandOperationHelper.executeCommandWithConnection;
 import static com.mongodb.internal.operation.CommandOperationHelper.isNamespaceError;
 import static com.mongodb.internal.operation.CommandOperationHelper.rethrowIfNotNamespaceError;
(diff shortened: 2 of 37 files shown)


driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -477,12 +477,23 @@ public Builder uuidRepresentation(final UuidRepresentation uuidRepresentation) {
         /**
          * Sets the auto-encryption settings
          *
+         * A separate, internal {@code MongoClient} is created if any of the following are true:
```
jyemin, line 480: Doc changes as required by the spec

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -477,12 +477,23 @@ public Builder uuidRepresentation(final UuidRepresentation uuidRepresentation) {
...
+         * options as the parent {@code MongoClient} except {@code minPoolSize} is set to {@code 0}
+         * and {@code AutoEncryptionSettings} is omitted.
+         *
          * @param autoEncryptionSettings the auto-encryption settings
          * @return this
          * @since 3.11
          * @see #getAutoEncryptionSettings()
          */
-        public Builder autoEncryptionSettings(final AutoEncryptionSettings autoEncryptionSettings) {
+        public Builder autoEncryptionSettings(@Nullable final AutoEncryptionSettings autoEncryptionSettings) {
```
jyemin, line 496: Made this nullable so that auto-encryption could easily be disabled in the cloned settings.

driver-core/src/main/com/mongodb/internal/operation/ListCollectionsOperation.java
```diff
@@ -262,22 +262,22 @@ public void call(final AsyncConnectionSource source, final AsyncConnection conne
                 if (t != null) {
                     errHandlingCallback.onResult(null, t);
                 } else {
-                    final SingleResultCallback<AsyncBatchCursor<T>> wrappedCallback = releasingCallback(errHandlingCallback,
-                                                                                                        source, connection);
                     if (serverIsAtLeastVersionThreeDotZero(connection.getDescription())) {
-                        executeCommandAsync(binding, databaseName, getCommandCreator(), createCommandDecoder(),
-                                asyncTransformer(), retryReads,
+                        executeCommandAsyncWithConnection(binding, source, databaseName, getCommandCreator(), createCommandDecoder(),
+                                asyncTransformer(), retryReads, connection,
```
jyemin, line 267: Using the same connection to execute the command, as is done in other operations (and in the execute method)

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/MongoClientImpl.java
```diff
@@ -115,7 +115,7 @@ Crypt getCrypt() {
         return crypt;
     }
 
-    MongoClientSettings getSettings() {
+    public MongoClientSettings getSettings() {
```
jyemin, line 118: The class is internal, so this isn't actually an API change

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/KeyRetriever.java
```diff
@@ -23,19 +23,16 @@
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-import java.io.Closeable;
 import java.util.List;
 
 import static com.mongodb.assertions.Assertions.notNull;
 
-class KeyRetriever implements Closeable {
+class KeyRetriever {
```
jyemin, line 30: Now `Crypt` class owns the internal `MongoClient`

driver-core/src/main/com/mongodb/MongoClientSettings.java
stIncMale: I'd be useful to also add the `@Nullable` annotation to the corresponding `MongoClientSettings.autoEncryptionSettings` field. Builder fields are expected to be nullable (builder objects are expected to exist in a partially constructed state) but for fields of other objects nullability is not so obvious.

It seems like `autoEncryptionSettings` was allowed to be null even before this PR, and a few other fields of the `MongoClientSettings` class are effectively nullable, but most of them are not.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/Crypt.java
```diff
@@ -173,7 +178,9 @@ public void close() {
         if (commandMarker != null) {
             commandMarker.close();
         }
-        keyRetriever.close();
+        if (internalClient != null) {
+            internalClient.close();
```
stIncMale, line 182: Do we want to to consider utilizing the try-with-resources block when closing multiple resources?

1. This way nulls are handled for us ([_"A resource is closed only if it initialized to a non-null value"_](https://docs.oracle.com/javase/specs/jls/se15/html/jls-14.html#jls-14.20.3)).
2. If the `close` method of any resource throws a `Throwable`, other resources are still closed, and none of the throwables are lost: [_"An exception from the closing of one resource does not prevent the closing of other resources. Such an exception is suppressed if an exception was thrown previously by an initializer, the try block, or the closing of a resource."_](https://docs.oracle.com/javase/specs/jls/se15/html/jls-14.html#jls-14.20.3) ("suppressed" means [`Throwable.addSuppressed`](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#addSuppressed-java.lang.Throwable-), such exceptions are shown in stack trances).

Here is an example:
```java
@Override
public void close() {
    try (final AutoCloseable _4 = keyManagementService;
        final AutoCloseable _3 = internalClient;
        final AutoCloseable _2 = commandMarker;
        final AutoCloseable _1 = mongoCrypt) {//closed in the order _1, ..., _4
    } catch (final Exception e) {//this is needed because AutoCloseable.close() throws Exception
    }
}
```
However, this does require marking the relevant classes as `implements AutoCloseable`, which is not a backward compatible change exactly because its `close` method is declared as `throws Exception`.

The problem with the current approach is that failing `mongoCrypt.close()` results in not closing any of the rest of the resources.

driver-core/src/test/resources/client-side-encryption/aggregate.json
```diff
@@ -150,18 +150,6 @@
...
-            "command": {
-              "listCollections": 1,
-              "filter": {
-                "name": "datakeys"
-              },
-              "$db": "keyvault"
-            },
-            "command_name": "listCollections"
-          }
-        },
```
stIncMale, line 164: Just confirming my understanding. Is this correct that this expectations were removed because they are gathered from a user-facing `MongoClient`, which does not use the `keyvault` DB anymore (the DB is now used by an internal `MongoClient` owned by a `Crypt` object)?

jyemin: Exactly.

driver-core/src/main/com/mongodb/MongoClientSettings.java
jyemin: I'm not sure what you mean.  The getter for that field is already Nullable:

    @Nullable
    public AutoEncryptionSettings getAutoEncryptionSettings() {
        return autoEncryptionSettings;
    }

jyemin: I agree there is inconsistency in MongoClientSettings about this.  For example, if autoEncryptionSettings can accept null, why not streamFactoryFactory or credential?  I think we missed a use case when considering whether to allow null on these builder setters.  Since the default value of those fields are null, it seemed unnecessary to also accept null in the setter.  But what we're doing in the PR shows another use case: when you construct a builder from an existing settings, and need to null out some of the values from the settings.    So I think it would be best to allow null in the setters for all Nullable properties.  But let's do that in a separate PR than this.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/Crypt.java
stIncMale: Ah, `KeyManagementService` is internal, so can be easily made `Autocloseable` (https://github.com/mongodb/mongo-java-driver/pull/630/commits/37b25ff7f674cd0209ed9993bfcf11bb042c7e99#diff-81695b3bfe7a24372a999b1d55f9af8799b05e515d88011e14e6332f6b6ed58dR43). 👍

rozza: Good catch @stIncMale

rozza (approved):
LGTM

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/Crypt.java
jyemin: I like this approach, though Intellij warns about unused local variables.   We don't have to worry about catching exceptions because all of our close() method overrides are define to not throw checked exceptions.

Done

driver-core/src/main/com/mongodb/MongoClientSettings.java
stIncMale: > The getter for that field is already Nullable

Yes, the getter is marked, I was thinking about additionally marking the field the getter returns, despite having the annotation on the getter is definitely helpful.

> I think it would be best to allow null in the setters for all Nullable properties. But let's do that in a separate PR than this.

Absolutely

stIncMale (approved):

jyemin:
Merged on command line:

* https://github.com/mongodb/mongo-java-driver/commit/c329baa3645384e90fe9a55433803995ad39be43
* https://github.com/mongodb/mongo-java-driver/commit/9b1670af30f197db20b4f70ce94ad35d4743b0df

PR #632, opened 2021-01-13
Remove mentions of the com.mongodb.Block class

stIncMale:
This class is no longer a part of the public API.

JAVA-3934


stIncMale:
These PR is inspired by https://github.com/mongodb/mongo-java-driver/pull/629, but the original one only scratched the surface.

jyemin (approved):
LGTM!

PR #633, opened 2021-01-13
Add .editorconfig

stIncMale:
IntlliJ IDEA support: https://www.jetbrains.com/help/idea/configuring-code-style.html#editorconfig

JAVA-3922


stIncMale:
@rozza This is the config that @jyemin uses. You may be using a different one and may want to discuss some of the options. I am fine with any format :)

jyemin (commented):
Should we prune this file to contain config for only the file formats that we actually use?

stIncMale:
> Should we prune this file to contain config for only the file formats that we actually use?

Done. All other sections contain at least one file extension that we have in the repo.

.editorconfig
```diff
@@ -0,0 +1,1008 @@
...
+tab_width = 4
+ij_continuation_indent_size = 8
+ij_formatter_off_tag = @formatter:off
+ij_formatter_on_tag = @formatter:on
+ij_formatter_tags_enabled = false
+ij_smart_tabs = false
+ij_visual_guides = none
+ij_wrap_on_typing = true
+
+[*.css]
```
jyemin, line 17: Let's remove this one too as we never edit the css files.

.editorconfig
```diff
@@ -0,0 +1,1008 @@
...
+ij_java_use_single_class_imports = true
+ij_java_variable_annotation_wrap = off
+ij_java_visibility = public
+ij_java_while_brace_force = never
+ij_java_while_on_new_line = false
+ij_java_wrap_comments = false
+ij_java_wrap_first_method_in_call_chain = false
+ij_java_wrap_long_lines = false
+
+[*.less]
```
jyemin, line 285: What are .less files?

.editorconfig
```diff
@@ -0,0 +1,1008 @@
...
+ij_less_space_after_colon = true
+ij_less_space_before_opening_brace = true
+ij_less_use_double_quotes = true
+ij_less_value_alignment = 0
+
+[*.sass]
```
jyemin, line 305: What are .sass files?

.editorconfig
```diff
@@ -0,0 +1,1008 @@
...
+ij_scala_use_alternate_continuation_indent_for_params = false
+ij_scala_use_scaladoc2_formatting = false
+ij_scala_variable_annotation_wrap = off
+ij_scala_while_brace_force = never
+ij_scala_while_on_new_line = false
+ij_scala_wrap_before_with_keyword = false
+ij_scala_wrap_first_method_in_call_chain = false
+ij_scala_wrap_long_lines = false
+
+[*.scss]
```
jyemin, line 517: What are .scss files?

.editorconfig
```diff
@@ -0,0 +1,1008 @@
...
+ij_scss_space_after_colon = true
+ij_scss_space_before_opening_brace = true
+ij_scss_use_double_quotes = true
+ij_scss_value_alignment = 0
+
+[*.styl]
```
jyemin, line 537: What are .styl files?

jyemin (commented):
Still looks like there are some unnecessary file formats.

.editorconfig
stIncMale: I don't know what are they, but here is an example: https://github.com/mongodb/mongo-java-driver/blob/master/docs/landing/static/s/lib/font-awesome/less/list.less

.editorconfig
stIncMale: An example: https://github.com/mongodb/mongo-java-driver/blob/master/docs/landing/static/s/lib/font-awesome/scss/_stacked.scss

.editorconfig
stIncMale: We also have `*.js` files, and we probably don't edit them. I will remove front-end related file extensions, even if we have such files in the repo.

.editorconfig
stIncMale: This is my fault 👎 , I missed the section. The same is true for the `*.styl` section.

jyemin:
For the HTML-related files, we almost never edit them, and will soon be removing the reference docs from the repository entirely.  So not sure if it's worth including edit configs for anything in the reference/landing docs.  

What do you think, @rozza

stIncMale:
I removed front-end related sections. Jeff, thanks for spotting the redundant `styl`, `sass` sections!

jyemin (approved):
LGTM

rozza (approved):
LGTM

I've changed now use the default intellij styles and a diff doesn't show much difference - so happy with it.

.editorconfig
```diff
@@ -0,0 +1,716 @@
+[*]
```
rozza, line 1: One suggestion add:

`root = true` to the start of the file.

stIncMale: Thanks 👍 , done.

PR #634, opened 2021-01-14
Ensure that resumeToken is included on resume attempts

jyemin:
JAVA-3871

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
@@ -64,7 +64,12 @@ final class ChangeStreamBatchCursor<T> implements AggregateResponseBatchCursor<T
         return resumeableOperation(new Function<AggregateResponseBatchCursor<RawBsonDocument>, Boolean>() {
             @Override
             public Boolean apply(final AggregateResponseBatchCursor<RawBsonDocument> queryBatchCursor) {
-                return queryBatchCursor.hasNext();
+                try {
+                    return queryBatchCursor.hasNext();
+                } catch (RuntimeException e) {
+                    cachePostBatchResumeToken(queryBatchCursor);
+                    throw e;
+                }
             }
         });
     }
@@ -74,9 +79,11 @@ final class ChangeStreamBatchCursor<T> implements AggregateResponseBatchCursor<T
         return resumeableOperation(new Function<AggregateResponseBatchCursor<RawBsonDocument>, List<T>>() {
             @Override
             public List<T> apply(final AggregateResponseBatchCursor<RawBsonDocument> queryBatchCursor) {
-                List<T> results = convertResults(queryBatchCursor.next());
-                cachePostBatchResumeToken(queryBatchCursor);
-                return results;
+                try {
+                    return convertResults(queryBatchCursor.next());
+                } finally {
+                    cachePostBatchResumeToken(queryBatchCursor);
+                }
             }
         });
     }
@@ -86,9 +93,11 @@ final class ChangeStreamBatchCursor<T> implements AggregateResponseBatchCursor<T
         return resumeableOperation(new Function<AggregateResponseBatchCursor<RawBsonDocument>, List<T>>() {
             @Override
             public List<T> apply(final AggregateResponseBatchCursor<RawBsonDocument> queryBatchCursor) {
-                List<T> results = convertResults(queryBatchCursor.tryNext());
-                cachePostBatchResumeToken(queryBatchCursor);
-                return results;
+                try {
+                    return convertResults(queryBatchCursor.tryNext());
+                } finally {
+                    cachePostBatchResumeToken(queryBatchCursor);
+                }
             }
         });
     }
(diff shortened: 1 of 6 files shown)


driver-core/src/test/functional/com/mongodb/internal/operation/ChangeStreamOperationProseTestSpecification.groovy
```diff
@@ -107,7 +107,10 @@ class ChangeStreamOperationProseTestSpecification extends OperationFunctionalSpe
         waitForLastRelease(async ? getAsyncCluster() : getCluster())
 
         where:
-        async << [true, false]
+        async  | callHasNext
```
jyemin, line 110: Added this hack so we can test the path where the getMore fails due to a call to hasNext rather than next

driver-sync/src/test/functional/com/mongodb/client/ChangeStreamProseTest.java
```diff
@@ -370,8 +370,8 @@ public void testGetResumeTokenReturnsPostBatchResumeTokenAfterGetMore()
             // use reflection to access the postBatchResumeToken
             AggregateResponseBatchCursor<?> batchCursor = getBatchCursor(cursor);
 
-            // check equality in the case where the batch has not been iterated at all
-            assertEquals(cursor.getResumeToken(), batchCursor.getPostBatchResumeToken());
+            // resume token should be null before iteration
+            assertNull(cursor.getResumeToken());
```
jyemin, line 374: This test was wrong.  Before the bug fix, both the values were null.  Afterwards, the postBatchResumeToken is set on the batch cursor but not the high-level cursor.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -100,6 +100,7 @@
         this.decoder = notNull("decoder", decoder);
         if (result != null) {
             this.operationTime = result.getTimestamp(OPERATION_TIME, null);
+            this.postBatchResumeToken = getPostBatchResumeTokenFromResponse(result);
```
jyemin, line 103: Now it matches AsyncQueryBatchCursor constructor

driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
```diff
@@ -64,7 +64,12 @@ public boolean hasNext() {
         return resumeableOperation(new Function<AggregateResponseBatchCursor<RawBsonDocument>, Boolean>() {
             @Override
             public Boolean apply(final AggregateResponseBatchCursor<RawBsonDocument> queryBatchCursor) {
-                return queryBatchCursor.hasNext();
+                try {
+                    return queryBatchCursor.hasNext();
+                } catch (RuntimeException e) {
```
jyemin, line 69: It might also be ok for this to be a finally block, but I wasn't able to convince myself of it.

rozza: I'm interested as to why do it this way? Personally, I'd convert to a finally block as its cleaner and in my head equates to the same thing.

driver-core/src/test/functional/com/mongodb/internal/operation/ChangeStreamOperationProseTestSpecification.groovy
rozza: Is this missing true, true? If so this would work:

`[async, callHasNext] << [[true, false], [true, false]].combinations()`

driver-sync/src/test/functional/com/mongodb/client/ChangeStreamProseTest.java
rozza: Should you add a `assertNotNull(batchCursor.getPostBatchResumeToken())` check as well?

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestValidator.java
```diff
@@ -51,8 +50,6 @@ public UnifiedTestValidator(final String fileDescription, final String testDescr
 
     @Before
     public void setUp() {
-        // TODO: remove after https://jira.mongodb.org/browse/JAVA-3871 is fixed
-        assumeTrue(!(fileDescription.equals("poc-change-streams") && testDescription.equals("Test consecutive resume")));
```
rozza, line 55: Hurrah!

rozza (changes requested):
Some minor nits

driver-sync/src/test/functional/com/mongodb/client/ChangeStreamProseTest.java
jyemin: Done

driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
jyemin: I convinced myself it's ok for this to be a finally block.

driver-core/src/test/functional/com/mongodb/internal/operation/ChangeStreamOperationProseTestSpecification.groovy
rozza: Ack

rozza (approved):
LGTM

PR #636, opened 2021-01-16
Fix the read timeout implementation in NettyStream

stIncMale:
Both the new approach and the original one achieve the guarantee that
there are no concurrent read timeouts scheduled for a channel.
This is an essential property needed for a timeout implementation,
let us call it "at-most-one".

The original approach of achieving the at-most-one property:
- Schedule timeouts only by an event loop thread.
- If another thread needs to schedule a timeout, it submits a new scheduleTimeout task
to the channel's event loop (asynchronous timeout scheduling).
- The scheduleTimeout task schedules a new timeout if none is scheduled.
The original approach allowed executions in which a scheduleTimeout task runs after
completion of the read operation that submitted the task,
which resulted in unexpected timeout exceptions.

The new approach achieves the at-most-one property by using a lock.
As a result, timeouts can be scheduled by any thread and there is no
asynchronous timeout scheduling. This means we cannot miss removing a timeout
because it was submitted for scheduling, but has not been scheduled yet.

Other notable changes:
- [related  bug fix] NettyStream now must always respect the requested additional timeout.
The original implementation had a chance to ignore the requested additional timeout
and schedule a timeout with the default delay. This is again due to the asynchronous
timeout scheduling in the original approach.
- [performance optimization] Public read methods do not schedule timeouts anymore
if the requested number of bytes is available right away.
- [performance optimization] Netty channel handlers do not try to schedule timeouts anymore,
timeouts may be scheduled only by the public read methods.
Trying to schedule timeouts from the method handleReadResponse was unnecessary
even in the original approach.
- [performance optimization] NettyStream does not produce excessive garbage by re-creating PendingReader
objects each time a new piece of data arrives for a pending reader.
- [code improvement] The fields NettyStream.pendingReader, pendingException are always written/read
inside synchronized blocks that use the same NettyStream object, so they can be plain.
Marking them volatile is unnecessary and potentially misleading.
- [code improvement] ReadTimeoutHandler was removed because it wasn't acting as a handler and was not needed.

JAVA-3920


stIncMale:
Almost all improvements described in the "Other notable changes" are done because of the suggestions expressed by [@konrad-kaminski](https://github.com/konrad-kaminski) either verbally or via the proposed code. Thank you, Konrad!

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -58,29 +60,72 @@
...
+ * int1 -> inv2 -> ret2
+ *      \--------> ret1
+ * }</pre>
+ * As shown on the diagram, the method {@link #readAsync(int, AsyncCompletionHandler)} runs concurrently with
+ * itself in the example above. However, there are no concurrent pending readers because the second operation
+ * is invoked after the first operation has completed reading despite the method has not returned yet.
  */
 final class NettyStream implements Stream {
-    private static final String READ_HANDLER_NAME = "ReadTimeoutHandler";
+    private static final int NO_SCHEDULE_TIMEOUT = -1;
```
rozza, line 105: 0 defaults to no timeout can this also be zero? 

Also can we name this to be explicit that it represents no timeout for reads? 
`NO_READ_TIMEOUT` = 0

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -143,6 +188,7 @@ private void initializeChannel(final AsyncCompletionHandler<Void> handler, final
             bootstrap.handler(new ChannelInitializer<SocketChannel>() {
                 @Override
                 public void initChannel(final SocketChannel ch) {
+                    final ChannelPipeline pipeline = ch.pipeline();
```
rozza, line 191: Nit: No need for local final variables.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -152,13 +198,24 @@ public void initChannel(final SocketChannel ch) {
                             enableHostNameVerification(sslParameters);
                         }
                         engine.setSSLParameters(sslParameters);
-                        ch.pipeline().addFirst("ssl", new SslHandler(engine, false));
+                        pipeline.addFirst("ssl", new SslHandler(engine, false));
                     }
+
                     int readTimeout = settings.getReadTimeout(MILLISECONDS);
+                    final long readTimeoutMillis;
```
rozza, line 205: Default to NO_SCHEDULE_TIME and then you can remove the else block.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -152,13 +198,24 @@ public void initChannel(final SocketChannel ch) {
...
                     if (readTimeout > 0) {
```
rozza, line 206: if `NO_SCHEDULE_TIME` was 0 then this could read `if (readTimeout > NO_SCHEDULE_TIME)`

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -211,18 +269,32 @@ public void operationComplete(final ChannelFuture future) throws Exception {
...
+     *                          Timeouts may be scheduled only by the public read methods. Taking into account that concurrent pending
+     *                          readers are not allowed, there must not be a situation when threads attempt to schedule a timeout
+     *                          before the previous one is either cancelled or completed.
+     */
+    private void readAsync(final int numBytes, final AsyncCompletionHandler<ByteBuf> handler, final long readTimeoutMillis) {
         ByteBuf buffer = null;
         Throwable exceptionResult = null;
         synchronized (this) {
             exceptionResult = pendingException;
+            PendingReader pendingReader = this.pendingReader;
```
rozza, line 288: Nit: Does this need to be made a local variable? If so seems we've used `localPendingReader` elsewhere to make it clearer.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -211,18 +269,32 @@ public void operationComplete(final ChannelFuture future) throws Exception {
...
             if (exceptionResult == null) {
                 if (!hasBytesAvailable(numBytes)) {
-                    pendingReader = new PendingReader(numBytes, handler);
+                    if (pendingReader == null) {//called by a public read method
+                        pendingReader = new PendingReader(numBytes, handler, scheduleReadTimeout(readTimeoutTask, readTimeoutMillis));
+                        this.pendingReader = pendingReader;
+                    } else {//called by a Netty channel handler
```
rozza, line 294: This else statement can be removed

rozza (changes requested):
Its looking good.

Some style nits, also you can remove any commented out assert lines of code.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: Thanks! This, combined with https://github.com/mongodb/mongo-java-driver/pull/636#discussion_r559500203 and https://github.com/mongodb/mongo-java-driver/pull/636#discussion_r559502153 makes the changes look so much cooler 😎

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: 👍

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: I agree that it can be, but does removing it make the code clearer? This code is not so trivial, and having comments that express what's going on, or even state some invariants and possible interpretations in case they are violated, appears to be useful. What are the downsides of having code comments like this?

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: The reason I am using local variables for shared state is to reduce the number of read usages we can see in the code (including the "Find usages" search in an IDE). The fewer code pieces include reads from or writes to a shared variable, the easier it is (at least to me) to think about the correctness of concurrent code.

The reason I did not call it `localPendingReader` is because `local` adds verbosity without adding any information: we can immediately see that this is a local variable on the line it is declared, and on the other lines where it is used IntelliJ IDEA IDE differentiates local and object variables (by default by using different boldness of the font).

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: We need to use a reference to the channel's pipeline in multiple places in this method. While calling `ch.pipeline()` each time we need it is correct, it is an unnecessary call, because we know that the method returns the same reference. We may make some assumptions by looking at the source code, but we should treat it as subject for change, and something that we may believe is "free" for now, may not be free in the future.

Additionally, not calling a method multiple times reduces the number of usages that IDE finds when we want to find usages of the method. This simplifies code navigation.

As for the `final` modifier on local variables, it's a simple way to tell the reader that the variable is assigned once and only once.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
rozza: There is a cognitive load and cost to a code base by reusing the same named variable. It also adds future risk to the codebase as  the localization of the variable may be changed.

Adding `local` does add verbosity but I disagree because it does add extra information, its explicit. Explicit should be preferred over implicit. Especially as you can't guarantee users have the same IDE or even settings in their IDE as you. 

So please make it explicit, it costs nothing and improves the readability of the code for future development.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
rozza: This codebase has been created by many authors, it looks homogenous because all the developers have tread lightly on the codebase and kept the code style consistent.  Using `final` for local variables isn't a style that has been adopted in the past.

However, with the base line now being java 8 and the need for local final variable in closures perhaps this is a style we should adopt. It is more explicit as you say.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
rozza: We generally don't inline comment in the code because its considered a code smell.  The code should be readable in intent, however there are scenarios where you may want to add comments. In general methods / classes have comments describing the logic. 

I'll leave this one to you thinks best, It may be useful in x years when we revisit this code.

rozza: Regarding this else statement I don't think it adds to the code as contains no logic.

A single comment checking is `if(pendingReader == null)` should suffice.

rozza (commented):
Looks good, continued the discussion for localPendingReader & pendingReader and the empty else block. Happy to discuss over zoom if its easier.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
jyemin: I'm happy to consider final local variables as a standard going forward, but currently it's our standard not to, so let's stick with that for now.

jyemin: Also, note that IntellJ now differentiates between mutated and effectively final local variables by underlining the former.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
jyemin: Our practice generally in this code base is to minimize use of local variables.  I can see the argument here in favor, since this is state that is protected by a lock, but at the same time I do find the style odd, particularly likes 288-290 below.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
jyemin: If you really want the commented asserts to remain as documentation, how about also commenting out the `else` that contains them?

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -445,47 +519,52 @@ public void operationComplete(final ChannelFuture future) {
...
             }
+        }
+
+        private ScheduledFuture<?> schedule(final long timeoutMillis) {
+            //assert timeoutMillis > 0 : timeoutMillis;
+            return ctx.executor().schedule(this, timeoutMillis, MILLISECONDS);
+        }
     }
-}
+}
```
jyemin, line 570: Newline required at end, I believe

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -152,13 +199,20 @@ public void initChannel(final SocketChannel ch) {
...
-                        ch.pipeline().addFirst("ssl", new SslHandler(engine, false));
+                        pipeline.addFirst("ssl", new SslHandler(engine, false));
                     }
+
                     int readTimeout = settings.getReadTimeout(MILLISECONDS);
-                    if (readTimeout > 0) {
-                        ch.pipeline().addLast(READ_HANDLER_NAME, new ReadTimeoutHandler(readTimeout));
+                    if (readTimeout > NO_SCHEDULE_TIME) {
+                        readTimeoutMillis = readTimeout;
+                        /* We need at least one handler before (in the inbound evaluation order) the InboundBufferHandler,
```
jyemin, line 208: I don't really understand this comment.  Where are we firing exception events?  What would be the observed behavior if we didn't have this handler?

jyemin (changes requested):
The logic seems sound.  Nice job with that.  Most of my comments are about style.

Can you think of any way to test this code?  Certainly our testing of the previous code was insufficient, and we haven't improved matters at all in that respect.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: I removed the `final` modifier. Is there a document with the style specified (not all style aspects are checked automatically)? I apologize for making not compliant changes, but I have not seen the document that specifies it. If there is none, I'll be picking up the style as I go.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: Removed the `pendingReader` local variable.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: Removed the `else` block and the comments within.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -58,29 +60,72 @@
...
+import java.util.concurrent.ScheduledFuture;
 
+import static com.mongodb.assertions.Assertions.isTrueArgument;
 import static com.mongodb.internal.connection.SslHelper.enableHostNameVerification;
 import static com.mongodb.internal.connection.SslHelper.enableSni;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 /**
  * A Stream implementation based on Netty 4.0.
+ * Just like it is for the {@link java.nio.channels.AsynchronousSocketChannel},
```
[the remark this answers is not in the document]
stIncMale: Thanks! Yes, I checked this in https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-core/index.html. However, writing a doc makes thing better anyway. For example, I had to learn that `NettyStream` does not allow concurrent pending readers by asking. The next engineer looking at the code will have a chance to learn this from the doc.

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: Our `.editorconfig` contains `insert_final_newline = false`, so Idea does not allow me to end a file with a newline character. Should I also change the `.editorconfig`?

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
stIncMale: > Where are we firing exception events?
[`ctx.fireExceptionCaught(ReadTimeoutException.INSTANCE)`](https://github.com/mongodb/mongo-java-driver/pull/636/files#diff-38fbf22cab5e77e01fe9fc24165ab0979e93cc5681d7b139798686145f1744b9R552) fires it.

The only way for us to fire an exception event in Netty not as a result of handing an inbound/outbound event in [`ChannelInboundHandler`](https://netty.io/4.1/api/io/netty/channel/ChannelInboundHandler.html)/[`ChannelOutboundHandler`](https://netty.io/4.1/api/io/netty/channel/ChannelOutboundHandler.html), but on demand is by calling [`ChannelHandlerContext.fireExceptionCaught`](https://netty.io/4.1/api/io/netty/channel/ChannelHandlerContext.html#fireExceptionCaught-java.lang.Throwable-). This method results [_"in having the ChannelInboundHandler.exceptionCaught(ChannelHandlerContext, Throwable) method called of the next ChannelInboundHandler contained in the ChannelPipeline of the Channel"_](https://netty.io/4.1/api/io/netty/channel/ChannelHandlerContext.html#fireExceptionCaught-java.lang.Throwable-). So we need to store a reference to a context of a handler (we store it in `ReadTimeoutTask.ctx`), which is a legit approach: [_"You can keep the ChannelHandlerContext for later use, such as triggering an event outside the handler methods, even from a different thread."_](https://netty.io/4.1/api/io/netty/channel/ChannelHandlerContext.html).

Now we need to decide which handler's context we will use. Our `InboundBufferHandler` knows how to process exceptions, so we need timeout exception events to be forwarded to this handler, which means we cannot use its context for firing such events (this follows from the docs quoted above). 👉 At this point you may read the [comment](https://github.com/mongodb/mongo-java-driver/pull/636/files#diff-38fbf22cab5e77e01fe9fc24165ab0979e93cc5681d7b139798686145f1744b9R208) again, and it will probably be clear this time. Hope this helps, but if not, let's have a call.

stIncMale:
> Can you think of any way to test this code?

1. We need a test that calls reads with variable added timeout on a `NettyStream` object and **checks that reads do not fail because of a premature timeout**. Since we cannot have concurrent pending readers on a single `NettyStream`, such a test needs to use many `NettyStream` objects to increase the number of concurrent attempts. This increases our chances to spot a problem without having to run the test for too long. Apparently, we also need a server that sends data with variable delays. It is important to note that while a test like this is probabilistic (may not spot the problem), it does not fail spuriously, i.e., when code is correct, because of the lack of hard real-time constraints (provided that we use `System.nanoTime` to measure time intervals, as it is guaranteed to be monotonic).
2. It would also be nice to have a way to **check that reads do not fail due to timeouts if data is made available in a timely manner**. Unlike 1., such tests are allowed to fail spuriously, i.e., fail even if the code is correct, because hard real-time computing is not what we do. However, if we set the max (soft) server delay as 2 seconds, while the min read timeout value is 30 seconds, then timing out almost certainly means we hit an actual code problem, e.g., a deadlock, rather than just experienced the _expected_ behaviour of a non-real-time computing system.

As an alternative to writing and starting/stopping a test server bound to a loopback interface, we may try using [`EmbeddedChannel`](https://netty.io/4.1/api/io/netty/channel/embedded/EmbeddedChannel.html). Netty has it specifically to facilitate testing. From what I can see in the "Netty in action" book, this channel is supposed to be used for testing one or multiple channel handlers (it accepts an array of handlers as one of its constructor arguments and builds an internal pipeline), and I am not sure it can be used as a transport implementation. In other words, I am not sure we can specify it via `Bootstrap.channel(Class<? extends C> channelClass)`. However, we can try, this will require changing `NettyStream(...Class<? extends SocketChannel> socketChannelClass...)` to `NettyStream(...Class<? extends Channel> channelClass...)` and changing the return type of the `NettyStream.getSocketChannelClass` method.

Writing such a test is not exactly a trivial task, so if we decide to write it, I am proposing to have a separate task in Jira. Should I create one?

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
jyemin: That's strange, since most (all?) existing code ends in a newline.  Since this file did end in a newline let's add it back, and we can figure out the .editorconfig issue separately

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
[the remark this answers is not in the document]
jyemin: We've preferred, I think, to use `//` comments in other places where we want extensive comments on non-public classes and methods, to make it clear that the comments are for maintainers and not users.  But I don't feel that strongly about it.

jyemin (changes requested):
Agreed about the testing.  Let's push it off for now.

rozza (approved):
LGTM @stIncMale - great work.

Unfortunately, we don't have a fixed style guide so somethings will be pick up as you go along. However, its a testament to the pr that its mostly style nits are being discussed.

jyemin (approved):

stIncMale:
The ticket to add tests: JAVA-3946.

PR #637, opened 2021-01-18
Improve TLS docs, add javaseref shortcode

stIncMale:
Formerly https://github.com/mongodb/mongo-java-driver/pull/622.

Note that this MR removes the TLS docs duplication and points reactive and Scala driver docs to the Java driver page. TSL docs for the reactive and Scala drivers were missing some relevant sections from the Java driver. By using only a single copy of this documentation, we can be sure that there are no outdated copies.

I installed `hugo` 0.25 locally from sources and manually checked that the changes, especially those related to the new `javaseref`, work properly.

JAVA-3420


rozza (changes requested):
My worry is that users will get lost jumping from their driver to the sync driver and that would cause confusion to the users. So I would prefer it being duplicated. 

Most of the documentation is duplicated and then updated to the relevant paths / code examples.

In the future once this is migrated over to the docs team, they will be able to include the content.  

Is that something we can do here? Include the content, so theres only one copy but it appears inline with all the languages?

stIncMale:
> My worry is that users will get lost jumping from their driver to the sync driver and that would cause confusion to the users.
> ...
> Include the content, so theres only one copy but it appears inline with all the languages?

The answer is complicated: in theory "yes", in practice "no". Despite I found a way to do so, the approach only works when the Hugo re-generates a page that embeds the content of a different page (this must be related to some sort of internal evaluation order). I created 1fe5cbfa887917b5ff60d344836e7d69ed08545a to show how it could look like, but we can't really do this.

TL;DR The TLS docs are now duplicated again.

---
For the sake of sharing knowledge, below are the possible approaches.

1. The [**official approach**](https://gohugo.io/templates/files/#readfile-example-add-a-project-file-to-content) is to use the [`readFile`](https://gohugo.io/functions/readfile/) function and write our own shortcode. The official page suggests writing a shortcode, but does not provide an example. The *problems* with this approach are
1.1. [Does not include headers from the embedded content into the TOC](https://github.com/gohugoio/hugo/issues/247#issuecomment-569379164).
1.2. [Does not process shortcodes from the embedded content](https://discourse.gohugo.io/t/readfile-does-not-parse-shortcodes/16667/5).
2. A **community approach** of embedding generated page content, described [here](https://discourse.gohugo.io/t/readfile-does-not-parse-shortcodes/16667/8) and [here](https://github.com/gohugoio/hugo/issues/247#issuecomment-342258564). Shortcodes are processed with this approach, which is nice. The *problems* with this approach are:
2.1. Still does not include headers from the embedded content into the TOC. I did not find people complaining, but this is the behaviour I observed.
2.2. Works only then when Huge re-generates a page that embeds another one because I updated either the page or I updated the `embed` shortcode (updating the page that is embedded does not help and results in removing the embedded content). I did not find similar complaints in the Internet.

Judging by 2.1, 2.2, there is a tiny hope that newer versions of Hugo do not have the problems (at least not 2.2.), but we are [locked at Hugo 0.25](https://github.com/mongodb/mongo-java-driver/blob/master/docs/reference/README.md#requirements). Interestingly, Hugo maintainers do not want to provide a built-in way to embed (duplicates the generated content) or include (no content duplication) content: https://github.com/gohugoio/hugo/issues/247.

rozza (approved):
LGTM - thanks for looking into the embedding but I agree duplicating for now seems way easier.

PR #638, opened 2021-01-20
Update reference documentation for the 4.2 release

jyemin:


docs/landing/data/releases.toml
```diff
@@ -1,12 +1,12 @@
-current = "4.1.1"
+current = "4.2.0"
```
stIncMale, line 1: Is there a way to reuse this value so that we only need to update it in a single place? If it cannot be reused directly from `releases.toml` it probably can be specified in `config.toml` as a [site parameter](https://gohugo.io/variables/site/#the-siteparams-variable) and then use the parameter value in pages.

jyemin: Perhaps, but can we investigate that separately?

rozza (approved):

stIncMale (approved):

jyemin:
Added one more commit for Project Reactor changes

rozza (approved):

docs/reference/content/upgrading.md
```diff
@@ -12,6 +12,9 @@ title = "Upgrade Considerations"
 The 4.2 release is a minor release as per the definition of [semantic versioning](https://semver.org). As such, there are no breaking
 changes in this release for users upgrading from the 4.1 release.
 
+Note that in this release the Reactive Streams driver has taken an implementation dependency on 
+[Project Reactor](https://projectreactor. io/).  The minimum supported Project Reactor version is *Californium-SR21*.
```
stIncMale: The space in `https://projectreactor. io/` does not seem intentional.

jyemin: Whoops.  Fixed on master

PR #644, opened 2021-01-26
Fail unified test if an operation throws an exception

jyemin:
A unified test should fail in the following circumstance:

* The operation specification defines neither expectResult nor expectError
* The operation throws an exception

JAVA-3960

A test for this is forthcoming in the spec test validator, but wanted to get the fix in now while it was fresh in my mind.


stIncMale:
Seems like another situation when a test may fail is when an operation specification expects a result, but a error happens. And this case is covered by the change in this PR.

stIncMale (approved):

PR #646, opened 2021-01-26
Code improvements split from JAVA-3907

stIncMale:
* ReferenceCounted.getCount documentation was changed.
It is documented to start at 1, which is with accordance to
what the current driver code expects. The common usage pattern is as follows:
ReferenceCounted resource = new ...;
//note that resource.retain() is not called because we expect the resource is created with counter 1
try {
  //use resource witho;
} finally {
  resource.release();
}
Changing implementations of ReferenceCounted.getCount to start with 0 would result in
resource.release() throwing IllegalStateException("Attempted to decrement the reference count below 0").
As long as the method ReferenceCounted.getCount is exposed, its behavior
is a part of the ReferenceCounted contract.

* Some fields in AsyncQueryBatchCursor were made final.

* Reduced the amount of garbage produced by QueryResult.getCursor.

* Documented Publisher properties in com.mongodb.reactivestreams.client.

### DIFF
--- a/driver-core/src/main/com/mongodb/ServerCursor.java
+++ b/driver-core/src/main/com/mongodb/ServerCursor.java
@@ -16,6 +16,8 @@
 
 package com.mongodb;
 
+import com.mongodb.annotations.Immutable;
+
 import java.io.Serializable;
 
 /**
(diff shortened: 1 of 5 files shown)


stIncMale:
I will split this PR into separate ones.

PR #647, opened 2021-01-26
Document publisher properties in API docs

stIncMale:
JAVA-3963

### DIFF
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/package-info.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/package-info.java
@@ -15,6 +15,17 @@
  */
 
 /**
- * This packages contains classes for the reactive stream client implementation
+ * This packages contains classes for the reactive stream client implementation.
+ * <p>
+ * All API {@link org.reactivestreams.Publisher}s are
+ * <a href="https://projectreactor.io/docs/core/release/reference/#reactor.hotCold">cold</a>,
+ * meaning that nothing happens until they are subscribed to.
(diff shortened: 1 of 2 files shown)


driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/package-info.java
```diff
@@ -15,6 +15,17 @@
  */
 
 /**
- * This packages contains classes for the reactive stream client implementation
+ * This packages contains classes for the reactive stream client implementation.
+ * <p>
+ * All API {@link org.reactivestreams.Publisher}s are
```
jyemin, line 20: Do these links end up resolving correctly to some stable location for the Reactive Streams documentation?

stIncMale: No, they do not. However, we have already had this situation before this PR. For example, see https://mongodb.github.io/mongo-java-driver/4.2/apidocs/mongodb-driver-reactivestreams/com/mongodb/reactivestreams/client/FindPublisher.html - there is no javadoc link to `org.reactivestreams.Publisher`.

I noticed the lack of links previously and saw that the only external API link we have configured is https://docs.oracle.com/en/java/javase/11/docs/api/. If there is no preventive rule, I am glad to add the link to the [`org.reactivestreams` API](https://www.reactive-streams.org/reactive-streams-1.0.2-javadoc/org/reactivestreams/package-summary.html).

I tried building javadocs with my change locally, and the change works as expected.

rozza (approved):
LGTM

jyemin (approved):
LGTM

PR #652, opened 2021-01-29
Fix reactive streams batch cursor regression.

rozza:
The BatchCursorPublisher now waits for demand to be signalled before
creating the cursor.

The batch cursor publisher also internally respects the batch size.
Its either the configured batch size if set or the amount requested.

JAVA-3973

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/client/model/FindOptions.java
+++ b/driver-core/src/main/com/mongodb/internal/client/model/FindOptions.java
@@ -60,6 +60,7 @@ public final class FindOptions {
     public FindOptions() {
     }
 
+    //CHECKSTYLE:OFF
     FindOptions(
             final int batchSize, final int limit, final Bson projection, final long maxTimeMS, final long maxAwaitTimeMS, final int skip,
             final Bson sort, final CursorType cursorType, final boolean noCursorTimeout, final boolean oplogReplay, final boolean partial,
@@ -86,6 +87,7 @@ public final class FindOptions {
         this.showRecordId = showRecordId;
         this.allowDiskUse = allowDiskUse;
     }
+    //CHECKSTYLE:ON
 
     public FindOptions withBatchSize(final int batchSize) {
         return new FindOptions(batchSize, limit, projection, maxTimeMS, maxAwaitTimeMS, skip, sort, cursorType, noCursorTimeout,
--- a/driver-core/src/main/com/mongodb/internal/client/model/FindOptions.java
+++ b/driver-core/src/main/com/mongodb/internal/client/model/FindOptions.java
@@ -60,6 +60,38 @@ public final class FindOptions {
     public FindOptions() {
     }
 
+    FindOptions(
+            final int batchSize, final int limit, final Bson projection, final long maxTimeMS, final long maxAwaitTimeMS, final int skip,
+            final Bson sort, final CursorType cursorType, final boolean noCursorTimeout, final boolean oplogReplay, final boolean partial,
+            final Collation collation, final String comment, final Bson hint, final String hintString, final Bson max, final Bson min,
+            final boolean returnKey, final boolean showRecordId, final Boolean allowDiskUse) {
+        this.batchSize = batchSize;
+        this.limit = limit;
+        this.projection = projection;
+        this.maxTimeMS = maxTimeMS;
+        this.maxAwaitTimeMS = maxAwaitTimeMS;
+        this.skip = skip;
+        this.sort = sort;
+        this.cursorType = cursorType;
+        this.noCursorTimeout = noCursorTimeout;
+        this.oplogReplay = oplogReplay;
+        this.partial = partial;
+        this.collation = collation;
+        this.comment = comment;
+        this.hint = hint;
+        this.hintString = hintString;
+        this.max = max;
+        this.min = min;
+        this.returnKey = returnKey;
+        this.showRecordId = showRecordId;
+        this.allowDiskUse = allowDiskUse;
+    }
+
+    public FindOptions withBatchSize(final int batchSize) {
+        return new FindOptions(batchSize, limit, projection, maxTimeMS, maxAwaitTimeMS, skip, sort, cursorType, noCursorTimeout,
+                oplogReplay, partial, collation, comment, hint, hintString, max, min, returnKey, showRecordId, allowDiskUse);
+    }
+
     /**
      * Gets the limit to apply.  The default is null.
      *
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/AggregatePublisherImpl.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/AggregatePublisherImpl.java
@@ -142,36 +142,33 @@ final class AggregatePublisherImpl<T> extends BatchCursorPublisher<T> implements
     private <E> Publisher<E> publishExplain(final Class<E> explainResultClass, @Nullable final ExplainVerbosity verbosity) {
         notNull("explainDocumentClass", explainResultClass);
         return getMongoOperationPublisher().createReadOperationMono(() ->
-                        asAggregateOperation().asAsyncExplainableOperation(verbosity,
+                        asAggregateOperation(1).asAsyncExplainableOperation(verbosity,
                                 getCodecRegistry().get(explainResultClass)),
                 getClientSession());
     }
 
     @Override
-    AsyncReadOperation<AsyncBatchCursor<T>> asAsyncReadOperation() {
+    AsyncReadOperation<AsyncBatchCursor<T>> asAsyncReadOperation(final int initialBatchSize) {
         MongoNamespace outNamespace = getOutNamespace();
 
         if (outNamespace != null) {
             AsyncWriteOperation<Void> aggregateToCollectionOperation = getAggregateToCollectionOperation();
 
-            FindOptions findOptions = new FindOptions().collation(collation);
-            Integer batchSize = getBatchSize();
-            if (batchSize != null) {
-                findOptions.batchSize(batchSize);
-            }
+            FindOptions findOptions = new FindOptions().collation(collation).batchSize(initialBatchSize);
+
             AsyncReadOperation<AsyncBatchCursor<T>> findOperation =
                     getOperations().find(outNamespace, new BsonDocument(), getDocumentClass(), findOptions);
 
             return new WriteOperationThenCursorReadOperation<>(aggregateToCollectionOperation, findOperation);
         } else {
-            return asAggregateOperation();
+            return asAggregateOperation(initialBatchSize);
         }
     }
 
-    private AggregateOperation<T> asAggregateOperation() {
+    private AggregateOperation<T> asAggregateOperation(final int initialBatchSize) {
         return getOperations()
                 .aggregate(pipeline, getDocumentClass(), maxTimeMS, maxAwaitTimeMS,
-                        getBatchSize(), collation, hint, comment, allowDiskUse, aggregationLevel);
+                           initialBatchSize, collation, hint, comment, allowDiskUse, aggregationLevel);
     }
 
     private AsyncWriteOperation<Void> getAggregateToCollectionOperation() {
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -29,6 +29,7 @@ import java.util.concurrent.atomic.AtomicLong;
 class BatchCursorFlux<T> implements Publisher<T> {
 
     private final BatchCursorPublisher<T> batchCursorPublisher;
+    @Nullable
     private final Integer originalBatchSize;
     private final Runnable resetBatchSize;
     private final AtomicBoolean inProgress = new AtomicBoolean(false);
@@ -48,10 +49,10 @@ class BatchCursorFlux<T> implements Publisher<T> {
     public void subscribe(final Subscriber<? super T> subscriber) {
         Flux.<T>create(sink -> {
             this.sink = sink;
-            sink.onRequest(l -> {
-                if (demandDelta.addAndGet(l) > 0 && inProgress.compareAndSet(false, true)) {
+            sink.onRequest(demand -> {
+                if (calculateDemand(demand) > 0 && inProgress.compareAndSet(false, true)) {
                     if (batchCursor == null) {
-                        batchCursorPublisher.batchSize(calculateBatchSize());
+                        batchCursorPublisher.batchSize(calculateBatchSize(sink.requestedFromDownstream()));
                         batchCursorPublisher.batchCursor().subscribe(bc -> {
                             batchCursor = bc;
                             inProgress.set(false);
@@ -70,7 +71,7 @@ class BatchCursorFlux<T> implements Publisher<T> {
         .subscribe(subscriber);
     }
 
-    void closeCursor() {
+    private void closeCursor() {
         if (batchCursor != null && !batchCursor.isClosed()) {
             batchCursor.close();
         }
@@ -81,7 +82,7 @@ class BatchCursorFlux<T> implements Publisher<T> {
             if (batchCursor.isClosed()) {
                 sink.complete();
             } else {
-                batchCursor.setBatchSize(calculateBatchSize());
+                batchCursor.setBatchSize(calculateBatchSize(sink.requestedFromDownstream()));
                 Mono.from(batchCursor.next())
                         .doOnCancel(this::closeCursor)
                         .doOnError((e) -> {
@@ -91,7 +92,7 @@ class BatchCursorFlux<T> implements Publisher<T> {
                         .doOnSuccess(results -> {
                             if (results != null) {
                                 results.forEach(sink::next);
-                                demandDelta.addAndGet(-results.size());
+                                calculateDemand(-results.size());
                             }
                             if (batchCursor.isClosed()) {
                                 sink.complete();
@@ -105,14 +106,23 @@ class BatchCursorFlux<T> implements Publisher<T> {
         }
     }
 
+    synchronized long calculateDemand(final long demand) {
+        long originalDelta = demandDelta.get();
+        long newDelta = demandDelta.addAndGet(demand);
+        if (demand > 0 && newDelta < originalDelta) {
+            demandDelta.set(Long.MAX_VALUE);
+            return Long.MAX_VALUE;
+        }
+        return newDelta;
+    }
 
-    int calculateBatchSize() {
+    int calculateBatchSize(final long demand) {
         if (originalBatchSize != null) {
             return originalBatchSize;
-        } else if (sink.requestedFromDownstream() > Integer.MAX_VALUE) {
+        } else if (demand > Integer.MAX_VALUE) {
             return Integer.MAX_VALUE;
         }
-        return Math.max(2, (int) sink.requestedFromDownstream());
+        return Math.max(2, (int) demand);
     }
 
 }
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -77,8 +77,11 @@ class BatchCursorFlux<T> implements Publisher<T> {
                 Mono.from(batchCursor.next())
                         .doOnCancel(this::closeCursor)
                         .doOnError((e) -> {
-                            closeCursor();
-                            sink.error(e);
+                            try {
+                                closeCursor();
+                            } finally {
+                                sink.error(e);
+                            }
                         })
                         .doOnSuccess(results -> {
                             if (results != null) {
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -48,7 +48,13 @@ class BatchCursorFlux<T> implements Publisher<T> {
                         batchCursorPublisher.batchCursor(batchSize).subscribe(bc -> {
                             batchCursor = bc;
                             inProgress.set(false);
-                            recurseCursor();
+
+                            // Handle any cancelled subscriptions that happen during the time it takes to get the batchCursor
+                            if (sink.isCancelled()) {
+                                closeCursor();
+                            } else {
+                                recurseCursor();
+                            }
                         }, sink::error);
                     } else {
                         inProgress.set(false);
@@ -97,8 +103,6 @@ class BatchCursorFlux<T> implements Publisher<T> {
                         })
                         .subscribe();
                 }
-        } else if (sink.isCancelled()) {
-            closeCursor();
         }
     }
 
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -30,7 +30,7 @@ class BatchCursorFlux<T> implements Publisher<T> {
     private final BatchCursorPublisher<T> batchCursorPublisher;
     private final AtomicBoolean inProgress = new AtomicBoolean(false);
     private final AtomicLong demandDelta = new AtomicLong(0);
-    private BatchCursor<T> batchCursor;
+    private volatile BatchCursor<T> batchCursor;
     private FluxSink<T> sink;
 
     BatchCursorFlux(final BatchCursorPublisher<T> batchCursorPublisher) {
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -16,7 +16,6 @@
 
 package com.mongodb.reactivestreams.client.internal;
 
-import com.mongodb.lang.Nullable;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import reactor.core.publisher.Flux;
@@ -29,20 +28,13 @@ import java.util.concurrent.atomic.AtomicLong;
 class BatchCursorFlux<T> implements Publisher<T> {
 
     private final BatchCursorPublisher<T> batchCursorPublisher;
-    @Nullable
-    private final Integer originalBatchSize;
-    private final Runnable resetBatchSize;
     private final AtomicBoolean inProgress = new AtomicBoolean(false);
     private final AtomicLong demandDelta = new AtomicLong(0);
     private BatchCursor<T> batchCursor;
     private FluxSink<T> sink;
 
-    BatchCursorFlux(final BatchCursorPublisher<T> batchCursorPublisher,
-                    @Nullable final Integer originalBatchSize,
-                    final Runnable resetBatchSize) {
+    BatchCursorFlux(final BatchCursorPublisher<T> batchCursorPublisher) {
         this.batchCursorPublisher = batchCursorPublisher;
-        this.originalBatchSize = originalBatchSize;
-        this.resetBatchSize = resetBatchSize;
     }
 
     @Override
@@ -52,13 +44,12 @@ class BatchCursorFlux<T> implements Publisher<T> {
             sink.onRequest(demand -> {
                 if (calculateDemand(demand) > 0 && inProgress.compareAndSet(false, true)) {
                     if (batchCursor == null) {
-                        batchCursorPublisher.batchSize(calculateBatchSize(sink.requestedFromDownstream()));
-                        batchCursorPublisher.batchCursor().subscribe(bc -> {
+                        int batchSize = calculateBatchSize(sink.requestedFromDownstream());
+                        batchCursorPublisher.batchCursor(batchSize).subscribe(bc -> {
                             batchCursor = bc;
                             inProgress.set(false);
                             recurseCursor();
                         }, sink::error);
-                        resetBatchSize.run();
                     } else {
                         inProgress.set(false);
                         recurseCursor();
@@ -117,8 +108,9 @@ class BatchCursorFlux<T> implements Publisher<T> {
     }
 
     int calculateBatchSize(final long demand) {
-        if (originalBatchSize != null) {
-            return originalBatchSize;
+        Integer setBatchSize = batchCursorPublisher.getBatchSize();
+        if (setBatchSize != null) {
+            return setBatchSize;
         } else if (demand > Integer.MAX_VALUE) {
             return Integer.MAX_VALUE;
         }
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -63,7 +63,7 @@ class BatchCursorFlux<T> implements Publisher<T> {
     }
 
     private void closeCursor() {
-        if (batchCursor != null && !batchCursor.isClosed()) {
+        if (batchCursor != null) {
             batchCursor.close();
         }
     }
@@ -97,14 +97,11 @@ class BatchCursorFlux<T> implements Publisher<T> {
         }
     }
 
-    synchronized long calculateDemand(final long demand) {
-        long originalDelta = demandDelta.get();
-        long newDelta = demandDelta.addAndGet(demand);
-        if (demand > 0 && newDelta < originalDelta) {
-            demandDelta.set(Long.MAX_VALUE);
-            return Long.MAX_VALUE;
-        }
-        return newDelta;
+    long calculateDemand(final long demand) {
+        return demandDelta.accumulateAndGet(demand, (originalValue, update) -> {
+            long newValue = originalValue + update;
+            return update > 0 && newValue < originalValue ? Long.MAX_VALUE : newValue;
+        });
     }
 
     int calculateBatchSize(final long demand) {
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -68,7 +68,7 @@ class BatchCursorFlux<T> implements Publisher<T> {
         }
     }
 
-    void recurseCursor(){
+    private void recurseCursor(){
         if (!sink.isCancelled() && sink.requestedFromDownstream() > 0 && inProgress.compareAndSet(false, true)) {
             if (batchCursor.isClosed()) {
                 sink.complete();
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -94,6 +94,8 @@ class BatchCursorFlux<T> implements Publisher<T> {
                         })
                         .subscribe();
                 }
+        } else if (sink.isCancelled()) {
+            closeCursor();
         }
     }
 
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.reactivestreams.client.internal;
+
+import com.mongodb.lang.Nullable;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.FluxSink;
+import reactor.core.publisher.Mono;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+class BatchCursorFlux<T> implements Publisher<T> {
+
+    private final BatchCursorPublisher<T> batchCursorPublisher;
+    private final Integer originalBatchSize;
+    private final Runnable resetBatchSize;
+    private final AtomicBoolean inProgress = new AtomicBoolean(false);
+    private final AtomicLong demandDelta = new AtomicLong(0);
+    private BatchCursor<T> batchCursor;
+    private FluxSink<T> sink;
+
+    BatchCursorFlux(final BatchCursorPublisher<T> batchCursorPublisher,
+                    @Nullable final Integer originalBatchSize,
+                    final Runnable resetBatchSize) {
+        this.batchCursorPublisher = batchCursorPublisher;
+        this.originalBatchSize = originalBatchSize;
+        this.resetBatchSize = resetBatchSize;
+    }
+
+    @Override
+    public void subscribe(final Subscriber<? super T> subscriber) {
+        Flux.<T>create(sink -> {
+            this.sink = sink;
+            sink.onRequest(l -> {
+                if (demandDelta.addAndGet(l) > 0 && inProgress.compareAndSet(false, true)) {
+                    if (batchCursor == null) {
+                        batchCursorPublisher.batchSize(calculateBatchSize());
+                        batchCursorPublisher.batchCursor().subscribe(bc -> {
+                            batchCursor = bc;
+                            inProgress.set(false);
+                            recurseCursor();
+                        }, sink::error);
+                        resetBatchSize.run();
+                    } else {
+                        inProgress.set(false);
+                        recurseCursor();
+                    }
+                }
+            });
+            sink.onCancel(this::closeCursor);
+            sink.onDispose(this::closeCursor);
+        }, FluxSink.OverflowStrategy.BUFFER)
+        .subscribe(subscriber);
+    }
+
+    void closeCursor() {
+        if (batchCursor != null && !batchCursor.isClosed()) {
+            batchCursor.close();
+        }
+    }
+
+    void recurseCursor(){
+        if (!sink.isCancelled() && sink.requestedFromDownstream() > 0 && inProgress.compareAndSet(false, true)) {
+            if (batchCursor.isClosed()) {
+                sink.complete();
+            } else {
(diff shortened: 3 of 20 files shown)


driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,118 @@
...
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.FluxSink;
+import reactor.core.publisher.Mono;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+class BatchCursorFlux<T> implements Publisher<T> {
```
rozza, line 29: I separated out the actual flux that gets subscribed to.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,118 @@
...
+        this.originalBatchSize = originalBatchSize;
+        this.resetBatchSize = resetBatchSize;
+    }
+
+    @Override
+    public void subscribe(final Subscriber<? super T> subscriber) {
+        Flux.<T>create(sink -> {
+            this.sink = sink;
+            sink.onRequest(l -> {
+                if (demandDelta.addAndGet(l) > 0 && inProgress.compareAndSet(false, true)) {
```
rozza, line 52: As this is backed by a the OverflowStrategy.BUFFER, only call getMore if needed:  `(demandDelta.addAndGet(l) > 0`

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,118 @@
...
+                            }
+                        })
+                        .subscribe();
+                }
+        }
+    }
+
+
+    int calculateBatchSize() {
+        if (originalBatchSize != null) {
```
rozza, line 110: If the user set the batchSize use that.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,118 @@
...
+            return originalBatchSize;
+        } else if (sink.requestedFromDownstream() > Integer.MAX_VALUE) {
+            return Integer.MAX_VALUE;
+        }
+        return Math.max(2, (int) sink.requestedFromDownstream());
```
rozza, line 115: Ensure that batchSize(1) isn't used as that can close the cursor.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorPublisher.java
```diff
@@ -116,44 +112,8 @@ public Integer getBatchSize() {
 
     @Override
     public void subscribe(final Subscriber<? super T> subscriber) {
-        batchCursor()
-                .flatMapMany(batchCursor -> {
```
rozza, line 120: `flatMapMany` requests data immediately - causing the original issue.

Also there is no hook for setting the initial batchSize based on demand.

driver-reactive-streams/src/test/unit/com/mongodb/reactivestreams/client/internal/AggregatePublisherImplTest.java
```diff
@@ -58,6 +58,7 @@ void shouldBuildTheExpectedOperation() {
 
         AggregateOperation<Document> expectedOperation = new AggregateOperation<>(NAMESPACE, pipeline,
                                                                                   getDefaultCodecRegistry().get(Document.class))
+                .batchSize(Integer.MAX_VALUE)
```
rozza, line 61: Demand is now signalled and the batchSize set via the requested data amount.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
+class BatchCursorFlux<T> implements Publisher<T> {
+
+    private final BatchCursorPublisher<T> batchCursorPublisher;
+    private final Integer originalBatchSize;
+    private final Runnable resetBatchSize;
+    private final AtomicBoolean inProgress = new AtomicBoolean(false);
+    private final AtomicLong demandDelta = new AtomicLong(0);
+    private BatchCursor<T> batchCursor;
```
stIncMale, line 36: This plain field is modified from
```java
sink.onRequest(l -> {
    ...
    batchCursorPublisher.batchCursor().subscribe(bc -> {
        ...
        batchCursor = bc;
        ...
    })
    ...
})
```
and is read from `BatchCursorFlux.closeCursor` and other places.

I have no knowledge of project Reactor, which means I can't tell whether it guarantees all the necessary happens-before relations for us, so I am pointing this out just in case. @rozza, if you thought about this and it should be fine, then I trust it is.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,118 @@
...
+                    }
+                }
+            });
+            sink.onCancel(this::closeCursor);
+            sink.onDispose(this::closeCursor);
+        }, FluxSink.OverflowStrategy.BUFFER)
+        .subscribe(subscriber);
+    }
+
+    void closeCursor() {
```
stIncMale, line 73: Would it be better to make the methods `closeCursor`, `recurseCursor`, `calculateBatchSize` private?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public void subscribe(final Subscriber<? super T> subscriber) {
+        Flux.<T>create(sink -> {
+            this.sink = sink;
+            sink.onRequest(l -> {
+                if (demandDelta.addAndGet(l) > 0 && inProgress.compareAndSet(false, true)) {
+                    if (batchCursor == null) {
+                        batchCursorPublisher.batchSize(calculateBatchSize());
```
jyemin, line 54: It seems dangerous for BatchCursorFlux to be mutating the state of the BatchCursorPublisher, since that may be both observed and changed by the application concurrrently with this call.  

Perhaps another way of accomplishing the goal is for this class to instead tell the publisher what batch size to use.  This could be done by adding a batchSize parameter to BatchCursorPublish.batchCursor().

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
```diff
@@ -207,6 +210,14 @@ private boolean tryMultipleTimes(final int noTimes, final Supplier<Boolean> test
      * @throws AssertionError if not exactly one terminal event notification was received
      */
     public void assertTerminalEvent() {
+
+        try {
+            //noinspection ResultOfMethodCallIgnored
+            latch.await(TIMEOUT_DURATION.toMillis(), TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            // ignore
```
stIncMale, line 218: Are we sure we want to swallow the interrupt signal instead of re-setting `Thread.currentThread().interrupt()`? In the absolute majority of cases, swallowing interrupt signals is not what the writer wants, and should usually be done by the same class (e.g., an executor that manages threads) that sends the signal in the first place.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    @After
+    public void tearDown() {
+        if (collection != null) {
+            drop(collection.getNamespace());
+        }
+        if (client != null) {
+            client.close();
+        }
```
stIncMale, line 67: Do we want to do `try{drop} finally {close}`, which has more chances to close the client if things go wrong?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,118 @@
...
+        this.batchCursorPublisher = batchCursorPublisher;
+        this.originalBatchSize = originalBatchSize;
+        this.resetBatchSize = resetBatchSize;
+    }
+
+    @Override
+    public void subscribe(final Subscriber<? super T> subscriber) {
+        Flux.<T>create(sink -> {
+            this.sink = sink;
+            sink.onRequest(l -> {
```
jyemin, line 51: Consider renaming `l` to something more descriptive like `demand` or `numRequestedDocuments`

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
jyemin, line 52: Do we need to check for overflow of demandDelta?

jyemin, line 52: I think there is a concurrency issue here.  Consider the case where an application calls Subscription.request multiple times.  The first call we will get demand and set inProgress to true.  Now imagine that the second call occurs just after line 57.  We've set the batchCursor but haven't recursed yet.  So the second call will observe demand and inProgress of false and will enter the conditional, then set inProgress to false and call recurseCursor.  Meanwhile the first call will also, concurrently, call recurseCursor.  

Also, batchCursor is not protected by a mutex and so I think it's not safe to assume that writes to it will be observed.

jyemin (changes requested):
I'll think more about how the concurrency issues might be addressed.  @stIncMale might also have some ideas about it.

driver-reactive-streams/src/test/unit/com/mongodb/reactivestreams/client/internal/AggregatePublisherImplTest.java
stIncMale: We expect batch size to be `Integer.MAX_VALUE`, but I do not see us requesting `Integer.MAX_VALUE` (or more), which is what we need to do in order to get `Integer.MAX_VALUE`, according to the `BatchCursorFlux.calculateBatchSize` method. Could you please help me understand where `Integer.MAX_VALUE` comes from?

stIncMale (commented):
It is worth saying that due not not having any knowledge of Project Reactor, I cannot assess the correctness of the core that uses Reactor. I expressed mostly thoughts about other aspects of this PR.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
```diff
@@ -166,7 +169,7 @@ public Subscription getSubscription() {
      * @throws AssertionError if the sequence of items observed does not exactly match {@code items}
      */
     public void assertReceivedOnNext(final List<T> items) {
-        if (tryMultipleTimes(items.size(), () -> getOnNextEvents().size() != items.size())) {
+        if (!tryMultipleTimes(10, () -> getOnNextEvents().size() == items.size())) {
```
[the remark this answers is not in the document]
rozza: This was changed because it didnt work and if items.size() is just 1 then its pretty useless as an eventually consistent check.
This class wasn't used by anything but hadn't been deleted so I just fixed it up.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
```diff
@@ -207,6 +210,14 @@ private boolean tryMultipleTimes(final int noTimes, final Supplier<Boolean> test
      * @throws AssertionError if not exactly one terminal event notification was received
      */
     public void assertTerminalEvent() {
+
+        try {
+            //noinspection ResultOfMethodCallIgnored
```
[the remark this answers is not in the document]
rozza: It wasn't used and was left over

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorPublisher.java
```diff
@@ -116,44 +112,8 @@ public Integer getBatchSize() {
...
-                            } else {
-                                inProgress.compareAndSet(true, false);
-                                recurseCursor(sink, batchCursor, inProgress);
-                            }
-                        })
-                        .subscribe();
-            }
-        }
+        Integer originalBatchSize = getBatchSize();
+        new BatchCursorFlux<>(this, originalBatchSize, () -> batchSize = originalBatchSize).subscribe(subscriber);
```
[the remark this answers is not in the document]
rozza: This was how it was done prior to reactor.  The issue is the batchSize is used as part of the operation creation, so if a user doesn't set it on the Publisher we shouldn't set it on the publisher.

I'll look to refactor around that and make it stable.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+        subscriber.requestMore(Long.MAX_VALUE);
+        subscriber.assertNoErrors();
+        subscriber.assertReceivedOnNext(docs);
+        assertCommandNames(singletonList("find"));
+        subscriber.assertTerminalEvent();
+    }
+
+    private void assertCommandNames(final List<String> commandNames) {
+        assertArrayEquals(commandNames.toArray(),
```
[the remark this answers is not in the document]
rozza: Junit4 lacks the assertIterable helper. So this is as easy and doesn't require hamcrest.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Ack

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
rozza: Here the test would most likely still fail due to the interruption, if it doesn't fail then it doesn't matter anyway.

driver-reactive-streams/src/test/unit/com/mongodb/reactivestreams/client/internal/AggregatePublisherImplTest.java
rozza: Sure, by using the blocking functions we are requesting data.

So ` Flux.from(publisher).blockFirst();` requests `Long.MAX_VALUE`, which as batchSize is an int32 gets converted to `Integer.MAX_VALUE`

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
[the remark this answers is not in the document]
rozza: Note: `assertReceivedOnNext` previously wasn't used.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
rozza: Ignore that we should error - will add.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Done

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Done

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
```diff
@@ -0,0 +1,231 @@
...
+        TestSubscriber<Document> subscriber = new TestSubscriber<>();
+        collection.find().batchSize(0).subscribe(subscriber);
+        assertCommandNames(emptyList());
+
+        subscriber.requestMore(100);
+        subscriber.assertReceivedOnNext(docs.subList(0, 100));
+        assertCommandNames(singletonList("find"));
+
+        subscriber.requestMore(101);
+        subscriber.assertNoErrors();
```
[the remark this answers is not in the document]
rozza: Done

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
```diff
@@ -0,0 +1,231 @@
...
+        subscriber.assertReceivedOnNext(docs.subList(0, 50));
+        assertCommandNames(singletonList("find"));
+
+        subscriber.requestMore(25);
+        subscriber.assertReceivedOnNext(docs.subList(0, 75));
+        subscriber.assertNoTerminalEvent();
+        assertCommandNames(asList("find", "getMore"));
+
+        subscriber.requestMore(25);
+        subscriber.assertNoErrors();
```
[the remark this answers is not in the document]
rozza: Done

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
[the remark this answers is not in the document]
rozza: Updated to Junit5

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
rozza: Done

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,118 @@
...
+import reactor.core.publisher.FluxSink;
+import reactor.core.publisher.Mono;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+class BatchCursorFlux<T> implements Publisher<T> {
+
+    private final BatchCursorPublisher<T> batchCursorPublisher;
+    private final Integer originalBatchSize;
```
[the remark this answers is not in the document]
rozza: Done

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Multiple calls to recurseCursor is fine as it contains a locking mechanism around inProgress - so if there is a race calling it via multiple calls to `request(n)` only one thread can action it.

> batchCursor is not protected by a mutex and so I think it's not safe to assume that writes to it will be observed.

Not sure I follow here, batchCursor is also protected by the inprogress check..

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Done

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
jyemin: @stIncMale can you check my reasoning here?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Correct, we set the field when we lazily fetch the cursor.

We explicitly close the batchCursor if the sink is cancelled / disposed of. If calling `batchCursorPublisher.batchCursor().subscribe` produces an error we flag that up to the sink as an error. In that case batchCursor won't have been set.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: The whole class is essentially private (being internal) I made this private and recurseCursor private and added tests for calculateBatchSize and the new calculateDemand methods.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,120 @@
...
+                                inProgress.set(false);
+                                recurseCursor();
+                            }
+                        })
+                        .subscribe();
+                }
+        }
+    }
+
+    synchronized long calculateDemand(final long demand) {
```
stIncMale, line 100: I propose using `AtomicLong.accumulateAndGet` method, which is meant to solve tasks like this one.
```java
long calculateDemand(final long demand) {
    return demandDelta.accumulateAndGet(demand, (originalDelta, update) -> {
        long newDelta = originalDelta + update;
        return demand > 0 && newDelta < originalDelta ? Long.MAX_VALUE : newDelta;
    });
}
```

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,120 @@
...
+                }
+            });
+            sink.onCancel(this::closeCursor);
+            sink.onDispose(this::closeCursor);
+        }, FluxSink.OverflowStrategy.BUFFER)
+        .subscribe(subscriber);
+    }
+
+    private void closeCursor() {
+        if (batchCursor != null && !batchCursor.isClosed()) {
```
stIncMale, line 66: Do we want to document `BatchCursor.close` as idempotent and spare ourselves from checks like this one? Its implementation is both not idempotent (not required, but could be) and incorrect at this moment, but the bug will be fixed once https://github.com/mongodb/mongo-java-driver/pull/654 is merged, and the method will also become idempotent.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/client/model/FindOptions.java
```diff
@@ -60,6 +60,40 @@
...
+        this.hintString = hintString;
+        this.max = max;
+        this.min = min;
+        this.returnKey = returnKey;
+        this.showRecordId = showRecordId;
+        this.allowDiskUse = allowDiskUse;
+    }
+    //CHECKSTYLE:ON
+
+    public FindOptions withBatchSize(final int batchSize) {
```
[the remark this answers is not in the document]
jyemin: We use this naming all over the current API, e.g. MongoCollection.withCodecRegistry() so I think it's something readers will have to get used to.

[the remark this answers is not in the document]
stIncMale: I see, thank you. Naming should be consistent.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/AggregatePublisherImpl.java
```diff
@@ -142,36 +142,33 @@
...
         return getMongoOperationPublisher().createReadOperationMono(() ->
-                        asAggregateOperation().asAsyncExplainableOperation(verbosity,
+                        asAggregateOperation(1).asAsyncExplainableOperation(verbosity,
                                 getCodecRegistry().get(explainResultClass)),
                 getClientSession());
     }
 
     @Override
-    AsyncReadOperation<AsyncBatchCursor<T>> asAsyncReadOperation() {
+    AsyncReadOperation<AsyncBatchCursor<T>> asAsyncReadOperation(final int initialBatchSize) {
```
jyemin, line 151: Do we lose anything by not letting this be nullable, like it is in `getBatchSize()`?  Or does `0 == null`?

stIncMale: A side but related question: is my understanding correct that we do not care about using boxed types and passing parameters by reference?

jyemin: We generally only box integral types when we need null to signal something, e.g. use the server's default value for this property.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
stIncMale: ### Scenario expressed by Jeff

> Consider the case where an application calls Subscription.request multiple times. The first call we will get demand and set inProgress to true. Now imagine that the second call occurs just after line 57. We've set the batchCursor but haven't recursed yet. So the second call will observe demand and inProgress of false and will enter the conditional, then set inProgress to false and call recurseCursor. Meanwhile the first call will also, concurrently, call recurseCursor.

@jyemin Yes, the method `recurseCursor` may be called concurrently in the scenario you describe. However, this method protects most of its body from being executed concurrently by using the same approach:
```java
if (inProgress.compareAndSet(false, true)) {//start executing mutually exclusive code in thread X
    ...
    doOnSuccess(results -> {
       ...
       inProgress.set(false);//end executing mutually exclusive code in thread Y
                             //(may not be the same as X, depending on how Reactor/BatchCursor work)
    })
}
```

A similar code in the `subscribe` method is meant to guarantee that:
1. `batchCursor` is initialized not more than once
2. `w_in_subscribe(batchCursor, bc)` happens-before (HB) any `r_in_recurseCursor(batchCursor)`.

#### HB(w_in_subscribe(batchCursor, <initial value>), r_in_closeCursor(batchCursor))?
`inProgress` does not provide HB between `w_in_subscribe(batchCursor, bc)` and `r_in_closeCursor(batchCursor)`. This HB may only be provided by the Reactor. @rozza Could you please confirm this HB relation is guaranteed?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
stIncMale: This thread continues here: https://github.com/mongodb/mongo-java-driver/pull/652#discussion_r568185193

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/FindPublisherImpl.java
```diff
@@ -199,14 +200,15 @@
     private <E> Publisher<E> publishExplain(final Class<E> explainResultClass, @Nullable final ExplainVerbosity verbosity) {
         notNull("explainDocumentClass", explainResultClass);
         return getMongoOperationPublisher().createReadOperationMono(() ->
-                        asAsyncReadOperation().asAsyncExplainableOperation(verbosity,
+                        asAsyncReadOperation(0).asAsyncExplainableOperation(verbosity,
```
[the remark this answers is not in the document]
rozza: The explain command doesn't use the batchSize but because of the change to push batchSize down we have to add a value and the server default seems best - as it may be honoured in the future.

`BatchCursorPublisher.asAsyncFirstReadOperation` only will return the first result so BatchSize can be set to 1.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
```diff
@@ -22,51 +22,27 @@
...
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
+import static com.mongodb.ClusterFixture.TIMEOUT_DURATION;
+
 public class TestSubscriber<T> implements Subscriber<T> {
 
-    private final Subscriber<T> delegate;
+    private static final int DEFAULT_MAX_RETRIES = 10;
```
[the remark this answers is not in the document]
rozza: Removed

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/AggregatePublisherImpl.java
rozza: We will always set the batchSize - either the users configured batchSize or the initial demand

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Nice - Done

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Updated

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
```diff
@@ -242,20 +197,26 @@ public void assertNoTerminalEvent() {
...
+            retry++;
+            if (check.get()) {
+                return true;
+            }
+            long sleepTimeMS = 100 + (100 * (long) Math.pow(retry, 2));
+            totalSleepTimeMS += sleepTimeMS;
+            try {
+                Thread.sleep(sleepTimeMS);
+            } catch (InterruptedException e) {
+                return false;
```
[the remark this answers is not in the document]
rozza: Done

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
```diff
@@ -242,20 +197,26 @@ public void assertNoTerminalEvent() {
...
-    /**
-     * Assert that this {@link Subscriber} has received an {@code onError} notification.
-     *
-     * @throws AssertionError if this {@link Subscriber} did not received an {@link #onError} notifications
-     */
-    public void assertErrored() {
-        if (onErrorEvents.size() == 0) {
-            // can't use AssertionError because (message, cause) doesn't exist until Java 7
-            throw new RuntimeException("No onError events");
+    private boolean retry(final Supplier<Boolean> check) {
```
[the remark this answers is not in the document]
rozza: Updated to `waitFor`

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/TestSubscriber.java
rozza: Done

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: `closeCursor` can be called at any time via `sink::onDispose` & `sink::onCancel` - so the batch cursor may not be set.

Looking at the code there is a chance of the sink being cancelled before the batchCursor is set which currently wouldn't close the cursor:

```
batchCursorPublisher.batchCursor(batchSize).subscribe(bc -> {
                            batchCursor = bc;  // sink.isCancelled could be true at this point in time.
                            inProgress.set(false);
                            recurseCursor();
                        }, sink::error);
```

So I've added a check in `recurseCursor` - as `closeCursor` is idempotent it doesn't matter if it gets called multiple times.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/AggregatePublisherImpl.java
jyemin: Ah, I see.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/FindPublisherImpl.java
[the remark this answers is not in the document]
stIncMale: @rozza, thank you for the explanation, it makes all the sense! :)

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
stIncMale: > `closeCursor` can be called at any time via `sink::onDispose` & `sink::onCancel`

And `closeCursor` may be called concurrently with the code we pass to the `onRequest` method, right? If that is the case, then making sure executions do not have data race due to writing to and reading from `batchCursor` is on us. The simplest way of achieving this is by marking `batchCursor` as `volatile`.

rozza: Done.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -94,6 +94,8 @@ private void recurseCursor(){
                         })
                         .subscribe();
                 }
+        } else if (sink.isCancelled()) {
+            closeCursor();
```
stIncMale, line 98: I am confused. My understanding previously was that we register `sink.onDispose(this::closeCursor)` specifically to call `closeCursor` when `sink` is either terminated with `onError`, or with `onComplete`, or a a result of being cancelled. This means:
1. We do not need to sprinkle explicit `closeCursor` in our code. Doing so confuses reader and makes it harder to reason about the code, because it creates expectations that `sink.onDispose(this::closeCursor)` is not enough.
2. We do not need `sink.onCancel(this::closeCursor)`.

We would have a good reason to write something beyond `sink.onDispose(this::closeCursor)` (like registering `sink.onCancel`, or calling `closeCursor` explicitly) if we knew about a bug in Project Reactor and was trying to work our way around it. Is this the case?

stIncMale (changes requested):

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -0,0 +1,119 @@
...
+        if (!sink.isCancelled() && sink.requestedFromDownstream() > 0 && inProgress.compareAndSet(false, true)) {
+            if (batchCursor.isClosed()) {
+                sink.complete();
+            } else {
+                batchCursor.setBatchSize(calculateBatchSize(sink.requestedFromDownstream()));
+                Mono.from(batchCursor.next())
+                        .doOnCancel(this::closeCursor)
+                        .doOnError((e) -> {
+                            closeCursor();
+                            sink.error(e);
```
stIncMale, line 81: If for some reason `closeCursor` throws, we are not going to call `sink.error`. Is there a good reason to not protect ourselves from such a scenario by using `try { close } finally { sink.error }`?

stIncMale (changes requested):

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: I think this fixes a possible race consider:

```
Subscription.request(1)
Subscription.cancel()
```

```
------- batchCursorPublisher.subscribe called
--------------- Close cursor called (batchCursor == null)
----------------------------- batchCursorPublisher.subscribe completes - recurseCursor closes the cursor.
```

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Its probably overly defensive as if it errors then something else is wrong.  It should be idempotent.

rozza: Done

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
stIncMale: I see, you are right! 👏  There is a race condition (but not a data race), that allows to have `batchCursor` left open despite `sink` being cancelled. I spent some significant time trying to prove that the solution you propose is correct, and I failed. Instead, I proved that there are allowed executions in which we will leave `batchCursor` open.

----
### Labels:
- `so` - synchronization order.
- `po` - program order.
- `hb` - happens-before order.
- `w(batchCursor)` - the write action that initializes `batchCursor` as a result of calling `subscribe`. 
- `r_in_closeCursor(batchCursor)` - the read from `batchCursor` action in `closeCursor` that is used to check for null.
- `w(sink.cancelled)` - a synchronization action (equivalent to volatile write) somewhere in the `sink` code that results in `sink.isClosed` returning true. This actions is done as a result of cancelling `sink`.
- `r(sink.cancelled)` - a synchronization action (equivalent to volatile read) somewhere in the `sink.isClosed` code that determines the return value of this method. This action is done as a result of the `sink.isCancelled()` call we do right after (in `po`) the action `w(batchCursor)`.

### Facts:
1. `hb`: `w(sink.cancelled)`, `r_in_closeCursor(batchCursor)` because `sink` calls `closeCursor` as a result of being cancelled.
2. `po`: `w(batchCursor)`, `r(sink.cancelled)` trivially from how the code is written.
3. `hb`: `w(batchCursor)`, `r(sink.cancelled)` because of fact 2.
5. `w(batchCursor)`, `r_in_closeCursor(batchCursor)`, `w(sink.cancelled)`, `r(sink.cancelled)` are synchronization actions, which means they are totally ordered in `so`. Executions with any `so` that does not result in violating JMM constraints is allowed.
6. `so`: `w(batchCursor)`, `r(sink.cancelled)` because of facts 5 and 2 (`so` is consistent with `po`).

### Proof:
Because of the fact 4, all allowed executions can be classified into the following two groups:
1. `so`: `w(batchCursor)`, `r_in_closeCursor(batchCursor)`
2. `so`: `r_in_closeCursor(batchCursor)`, `w(batchCursor)`

#### Executions in group 1
`hb`: `w(batchCursor)`, `r_in_closeCursor(batchCursor)` because of the `so` in this group of executions.
The code in `closeCursor` observes non-null `batchCursor` and calls `batchCursor.close` in all executions in this group.

#### Executions in group 2
```
hb: w(sink.cancelled), r_in_closeCursor(batchCursor)                                    fact 1
so:                    r_in_closeCursor(batchCursor), w(batchCursor)                    for this group of executions
hb:                                                   w(batchCursor), r(sink.cancelled) fact 3
so:                                                   w(batchCursor), r(sink.cancelled) fact 6
```
Or, with arrows:
```
w(sink.cancelled) -hb-> r_in_closeCursor(batchCursor) -\
                                                       so
                                                         \-> w(batchCursor) -po, hb, so-> r(sink.cancelled)
```
Unfortunately, there is nothing in the orders specified above that would either give us `hb`: `w(sink.cancelled)`, `r(sink.cancelled)`, or prohibit `so`: `r(sink.cancelled)`, `w(sink.cancelled)`.

As a result, executions in which `r_in_closeCursor(batchCursor)` does not observe `w(batchCursor)`, and `r(sink.cancelled)` does not observe `w(sink.cancelled)` are allowed (they constitute a subset of executions in group 2). Such executions leave `batchCursor` not closed despite `sink` being cancelled.

stIncMale: We can prevent faulty executions described above by enclosing `r_in_closeCursor(batchCursor)` and `w(batchCursor) & r(sink.cancelled)` in `synchronized` blocks (or equivalent constructs). Once we do that, the following will be true for all the previously faulty executions: from `so: r_in_closeCursor(batchCursor), w(batchCursor)` follows `hb: r_in_closeCursor(batchCursor), w(batchCursor)` (this is not the case with the current code). As a result, we will start having `hb`: `w(sink.cancelled)`, `r(sink.cancelled)` for all the previously faulty executions, and they will stop being faulty.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
stIncMale: Thank you! I would agree that it's overly defensive if we were calling `System.exit` or `Runtime.halt` methods as a result of failed `closeCursor`. As long as it's not the case, the code needs to do its best to try and follow the contracts.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
stIncMale: I gave this situation more thought, and it seems like the happens-before order is provable consistent with the synchronization order: [is HB consistent with SO for not otherwise related synchronization actions?](http://cs.oswego.edu/pipermail/concurrency-interest/2021-February/017286.html), [proof](http://cs.oswego.edu/pipermail/concurrency-interest/2021-February/017287.html) (hopefully, a correct one).

Given this new proved fact, we can come back to the reasoning about executions in group 2.

#### Executions in group 2
```
w(sink.cancelled) -hb-> r_in_closeCursor(batchCursor) -\
                                                       so
                                                         \-> w(batchCursor) -po, hb, so-> r(sink.cancelled)
```
gives us
```
w(sink.cancelled) -so, hb-> r_in_closeCursor(batchCursor) -\
                                                       so
                                                         \-> w(batchCursor) -po, hb, so-> r(sink.cancelled)
```
The above gives `so`: `w(sink.cancelled)`, `r(sink.cancelled)` and, therefore `hb`: `w(sink.cancelled)`, `r(sink.cancelled)`. The code initializing `batchCursor` observes that `sink` is cancelled and calls `batchCursor.close` in all executions in this group.

stIncMale: @rozza Sorry for lots of text, but now we have a formal proof that the current approach of eliminating that race condition you mentioned is correct.

The only thing I'd like to propose, is to remove the `if (sink.isCancelled()) closeCursor()` code from `recurseCursor` and put it right after `batchCursor = bc`. This way we do not introduce unnecessary code to the `recurseCursor`. It may also be useful to explain why `if (sink.isCancelled()) closeCursor()` exists in the code, because this race condition is not so easy to notice.

rozza: Done.

stIncMale: Thank you!

stIncMale (approved):
👍 
I marked all conversations started by me as resolved.

jyemin (approved):

PR #653, opened 2021-01-29
3.8.x continuous matrix testing

jyemin:
This is a patch to the 3.8.x branch to fulfill the goals for continuous matrix testing for the 4.0-era driver, which was 3.8.0.  Similar PRs will be done for the 4.2 and 4.4 era release branches.   What I've done here is two things:

1. Modify or skip tests that will not pass against later server versions.  Mostly because the features under test have been removed
2. In cases where the server made breaking changes to commands that prevents the driver code from working, I changed the driver code to make it work with the later server versions.  This seemed easier than disabling all the tests of that feature, though that is another option we could consider.  Look at the changes to MapReduceToCollectionOperation for an example of this. 

There's nothing really new here.  All these changes have been applied already in one form or another to the master branch of the driver.  It was just a matter of finding them and copying them to this branch.

JAVA-3942

Full patch build: https://spruce.mongodb.com/version/60133b4d9ccd4e019373ad0d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -194,7 +194,7 @@ functions:
   "upload test results":
     - command: attach.xunit_results
       params:
-        file: ./src/*/build/test-results/TEST-*.xml
+        file: ./src/*/build/test-results/test/TEST-*.xml
 
   "bootstrap mongo-orchestration":
     - command: shell.exec
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -316,15 +316,6 @@ functions:
           ${PREPARE_SHELL}
           echo '{"results": [{ "status": "FAIL", "test_file": "Build", "log_raw": "No test-results.json found was created"  } ]}' > ${PROJECT_DIRECTORY}/test-results.json
 
-  "install dependencies":
-    type: test
-    params:
-      working_dir: "src"
-      script: |
-        ${PREPARE_SHELL}
-        file="${PROJECT_DIRECTORY}/.evergreen/install-dependencies.sh"
-        [ -f ${file} ] && sh ${file} || echo "${file} not available, skipping"
-
 # Anchors
 
 hosts: &hosts
@@ -337,7 +328,6 @@ pre:
   - func: "fix absolute paths"
   - func: "init test-results"
   - func: "make files executable"
-  - func: "install dependencies"
 
 post:
   # Removed, causing timeouts
@@ -357,9 +347,6 @@ tasks:
         - func: "upload build"
 
     - name: "test"
-      depends_on:
-        - variant: "static-checks"
-          name: "static-analysis"
       commands:
         - func: "bootstrap mongo-orchestration"
         - func: "run tests"
@@ -408,6 +395,10 @@ axes:
   - id: version
     display_name: MongoDB Version
     values:
+      - id: "4.2"
+        display_name: "4.2"
+        variables:
+          VERSION: "4.2"
       - id: "4.0"
         display_name: "4.0"
         variables:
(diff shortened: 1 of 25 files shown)


.evergreen/.evg.yml
```diff
@@ -194,7 +194,7 @@ functions:
   "upload test results":
     - command: attach.xunit_results
       params:
-        file: ./src/*/build/test-results/TEST-*.xml
+        file: ./src/*/build/test-results/test/TEST-*.xml
```
jyemin, line 197: Test uploading was actually broken in the 3.x branch, so I just fixed it.

.evergreen/.evg.yml
```diff
@@ -316,15 +316,6 @@ functions:
           ${PREPARE_SHELL}
           echo '{"results": [{ "status": "FAIL", "test_file": "Build", "log_raw": "No test-results.json found was created"  } ]}' > ${PROJECT_DIRECTORY}/test-results.json
 
-  "install dependencies":
```
jyemin, line 319: The presence of this function caused evergreen to refuse to load the file, so had to remove it.

driver-core/src/main/com/mongodb/operation/MapReduceHelper.java
```diff
@@ -26,19 +27,19 @@ static MapReduceStatistics createStatistics(final BsonDocument result) {
     }
 
     private static int getInputCount(final BsonDocument result) {
-        return result.getDocument("counts").getNumber("input").intValue();
```
jyemin, line 29: These changes are to prevent the driver from failing when no statistics are present in the reply.  Without this change, we can't run mapReduce operations at all.

driver-core/src/main/com/mongodb/operation/MapReduceToCollectionOperation.java
```diff
@@ -581,21 +583,23 @@ public MapReduceStatistics apply(final BsonDocument result, final ServerAddress
 
     private BsonDocument getCommand(final ConnectionDescription description) {
         BsonDocument outputDocument = new BsonDocument(getAction(), new BsonString(getCollectionName()));
-        outputDocument.append("sharded", BsonBoolean.valueOf(isSharded()));
-        outputDocument.append("nonAtomic", BsonBoolean.valueOf(isNonAtomic()));
+        if (description != null && !serverIsAtLeastVersionFourDotFour(description)) {
```
jyemin, line 586: Without this change, the server will fail to execute the operation

.evergreen/.evg.yml
rozza: Ack

driver-core/src/main/com/mongodb/operation/MapReduceHelper.java
rozza: Ack

driver-core/src/main/com/mongodb/operation/MapReduceToCollectionOperation.java
```diff
@@ -581,21 +583,23 @@ public MapReduceStatistics apply(final BsonDocument result, final ServerAddress
...
-        putIfNotZero(commandDocument, "limit", getLimit());
+                .append("map", getMapFunction())
+                .append("reduce", getReduceFunction())
+                .append("out", outputDocument);
+
+        putIfNotNull(commandDocument, "query", getFilter());
+        putIfNotNull(commandDocument, "sort", getSort());
+        putIfNotNull(commandDocument, "finalize", getFinalizeFunction());
+        putIfNotNull(commandDocument, "scope", getScope());
+        putIfTrue(commandDocument, "verbose", isVerbose());      putIfNotZero(commandDocument, "limit", getLimit());
```
rozza, line 602: nit: missing a new line

rozza (changes requested):
Minor nit but LGTM after that

driver-core/src/main/com/mongodb/operation/MapReduceToCollectionOperation.java
jyemin: Ack.  I'll fix it locally

jyemin:
Pushed to https://github.com/mongodb/mongo-java-driver/tree/3.8.x-continuous-matrix-testing

PR #654, opened 2021-01-31
Make implementations of AsyncAggregateResponseBatchCursor more responsive to close signals

stIncMale:
JAVA-3907

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
@@ -69,9 +69,9 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
     private volatile int batchSize;
     private final AtomicInteger count = new AtomicInteger();
     private volatile BsonDocument postBatchResumeToken;
-    private volatile BsonTimestamp operationTime;
-    private volatile boolean firstBatchEmpty;
-    private volatile int maxWireVersion = 0;
+    private final BsonTimestamp operationTime;
+    private final boolean firstBatchEmpty;
+    private final int maxWireVersion;
 
     /* protected by `this` */
     private boolean isOperationInProgress = false;
@@ -100,6 +100,8 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
         if (result != null) {
             this.operationTime = result.getTimestamp(OPERATION_TIME, null);
             this.postBatchResumeToken = getPostBatchResumeTokenFromResponse(result);
+        } else {
+            this.operationTime = null;
         }
 
         firstBatchEmpty = firstBatch.getResults().isEmpty();
@@ -109,9 +111,7 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
                 killCursor(connection);
             }
         }
-        if (connection != null) {
-            this.maxWireVersion = connection.getDescription().getMaxWireVersion();
-        }
+        this.maxWireVersion = connection == null ? 0 : connection.getDescription().getMaxWireVersion();
     }
 
     @Override
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
@@ -70,9 +70,6 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
     private final AtomicInteger count = new AtomicInteger();
     private volatile BsonDocument postBatchResumeToken;
     private final BsonTimestamp operationTime;
-    /**
-     * May be true for a tailable cursor.
-     */
     private final boolean firstBatchEmpty;
     private final int maxWireVersion;
 
@@ -193,10 +190,12 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
         if (isClosed()) {
             callback.onResult(null, new MongoException(format("%s called after the cursor was closed.",
                     tryNext ? "tryNext()" : "next()")));
-        } else if (firstBatch != null && (tryNext || !firstBatchEmpty)) {
-            List<T> results = tryNext && firstBatchEmpty
-                    ? null
-                    : firstBatch.getResults();
+        } else if (firstBatch != null && (tryNext || !firstBatch.getResults().isEmpty())) {
+            // May be empty for a tailable cursor
+            List<T> results = firstBatch.getResults();
+            if (tryNext && results.isEmpty()) {
+                results = null;
+            }
             firstBatch = null;
             ServerCursor localCursor = getServerCursor();
             if (localCursor == null) {
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
@@ -79,8 +79,8 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
     /* protected by `this` */
     private boolean isOperationInProgress = false;
     private boolean isClosed = false;
-    private boolean isClosePending = false;
     /* protected by `this` */
+    private volatile boolean isClosePending = false;
 
     AsyncQueryBatchCursor(final QueryResult<T> firstBatch, final int limit, final int batchSize, final long maxTimeMS,
                           final Decoder<T> decoder, final AsyncConnectionSource connectionSource, final AsyncConnection connection) {
@@ -338,7 +338,13 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
     private void handleGetMoreQueryResult(final AsyncConnection connection, final SingleResultCallback<List<T>> callback,
                                           final QueryResult<T> result, final boolean tryNext) {
         cursor.set(result.getCursor());
-        if (!tryNext && result.getResults().isEmpty() && result.getCursor() != null) {
+        if (isClosePending) {
+            try {
+                endOperationInProgress();
+            } finally {
+                callback.onResult(null, null);
+            }
+        } else if (!tryNext && result.getResults().isEmpty() && result.getCursor() != null) {
             getMore(connection, result.getCursor(), callback, false);
         } else {
             count.addAndGet(result.getResults().size());
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
@@ -70,6 +70,9 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
     private final AtomicInteger count = new AtomicInteger();
     private volatile BsonDocument postBatchResumeToken;
     private final BsonTimestamp operationTime;
+    /**
+     * May be true for a tailable cursor.
+     */
     private final boolean firstBatchEmpty;
     private final int maxWireVersion;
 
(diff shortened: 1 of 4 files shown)


driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
```diff
@@ -69,15 +69,18 @@
     private volatile int batchSize;
     private final AtomicInteger count = new AtomicInteger();
     private volatile BsonDocument postBatchResumeToken;
-    private volatile BsonTimestamp operationTime;
```
rozza, line 72: :+1:

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
```diff
@@ -337,7 +338,17 @@ private void endOperationInProgress() {
...
                                           final QueryResult<T> result, final boolean tryNext) {
         cursor.set(result.getCursor());
-        if (!tryNext && result.getResults().isEmpty() && result.getCursor() != null) {
+        if (isClosePending) {
+            try {
+                connection.release();
+                if (result.getCursor() == null) {
+                    connectionSource.release();
+                }
+                endOperationInProgress();
```
rozza, line 347: Can we add a test to ensure that this happens and cursors are closed / killed?

rozza (changes requested):
I think adding a functional test cancelling the subscription to a change stream and the cursor is killed / bindings are released would be good.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
```diff
@@ -69,15 +69,18 @@
     private volatile int batchSize;
     private final AtomicInteger count = new AtomicInteger();
     private volatile BsonDocument postBatchResumeToken;
-    private volatile BsonTimestamp operationTime;
-    private volatile boolean firstBatchEmpty;
-    private volatile int maxWireVersion = 0;
+    private final BsonTimestamp operationTime;
+    /**
+     * May be true for a tailable cursor.
+     */
```
jyemin, line 75: It may be true for any type of cursor.  It may be confusing to call it out for tailable cursors only.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
```diff
@@ -190,12 +193,10 @@ private void next(final SingleResultCallback<List<T>> callback, final boolean tr
         if (isClosed()) {
             callback.onResult(null, new MongoException(format("%s called after the cursor was closed.",
                     tryNext ? "tryNext()" : "next()")));
-        } else if (firstBatch != null && (tryNext || !firstBatch.getResults().isEmpty())) {
-            // May be empty for a tailable cursor
-            List<T> results = firstBatch.getResults();
-            if (tryNext && results.isEmpty()) {
-                results = null;
-            }
+        } else if (firstBatch != null && (tryNext || !firstBatchEmpty)) {
```
jyemin, line 196: Is this part of the fix to the bug, or just a refactoring?   If just a refactoring, I suggest reverting it so as not to mix non-obvious refactoring with a bug fix.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
jyemin: Seems like this is the spot where we would kill the cursor, if there is still one active.  I don't see that it's being done though.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
```diff
@@ -337,7 +338,17 @@ private void endOperationInProgress() {
...
-        if (!tryNext && result.getResults().isEmpty() && result.getCursor() != null) {
+        if (isClosePending) {
+            try {
+                connection.release();
+                if (result.getCursor() == null) {
+                    connectionSource.release();
+                }
+                endOperationInProgress();
+            } finally {
+                callback.onResult(null, null);
```
jyemin, line 349: Consider whether we should report an error here instead of indicating that the cursor has been exhausted normally.

jyemin (changes requested):

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
stIncMale: It is the place, and the cursor is killed as a result of the following sequence:
```java
isClosePending == true
    endOperationInProgress()
        close()
            killCursorOnClose()
```
This is exactly the same logic we had (and still have) in the `else` branch of the `handleGetMoreQueryResult` method:
```java
else {
    ...
    else {
        connection.release();
        if (result.getCursor() == null) {
            connectionSource.release();
        }
    }
    endOperationInProgress();
    ...
}
```

stIncMale: > Can we add a test to ensure that this happens and cursors are closed / killed?

I thought about adding a test. While I can see how to create a test that checks that the subscription reacts to the cancellation as expected from the standpoint of the enduser, creating a similar test only for the `AsyncQueryBatchCursor` is not as straightforward. I would be glad to discuss how to proceed with such a test. My first thoughts are expressed in [this comment](https://jira.mongodb.org/browse/JAVA-3907?focusedCommentId=3592460&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3592460) (@rozza, it was written before your request, so we may continue discussing it there).

jyemin: Acknowledged.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
stIncMale: This comment was originally in the code: https://github.com/mongodb/mongo-java-driver/pull/654/commits/0893c181da2e774cc8236760869b6b3ac140f775#diff-a781a219ff4737fc3cbf85dcbd57ee59970d024d0994cdf390b352afd1c38d20L194

Once I roll back 0893c181da2e774cc8236760869b6b3ac140f775, this comment will return to its original place.

jyemin: Acknowledged

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
jyemin: I think my confusion was due to the fact that below the same connection is used to kill the cursor, while here we release the connection, and check out a new one for the killCursors.  Does that difference matter?

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
stIncMale: Yes, we may either return null, thus signalling `onComplete` or throw an exception, thus signalling `onError`. According to https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.2/README.md#user-content-1.6 and https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.2/README.md#user-content-2.4 both options are valid, but `onComplete` seems more natural to me: a user requested to cancel the subscription, and we cancelled it - there is nothing unexpected or exceptional about this situation.

The problem is that Reactor calls neither `onError` nor `onComplete` despite `AsyncChangeStreamBatchCursor.convertResultsCallback` passing the result/exception to the callback provided by the `MongoOperationPublisher.sinkToCallback` method. This is not related to the changes in this MR and may be a manifestation of us not hooking up to Reactor properly.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
stIncMale: We use the provided connection to close the cursor only if we reached the limit, otherwise the code below kills the cursor in a different connection.

The only good reason I can think of for killing the server cursor in a different connection is if the connection we already have is closed or have to be closed due to a failed IO operation in it. I do not understand why the code in `handleGetMoreQueryResult` kills the cursor one way or another, and with regard to killing the cursor, I simply followed the logic that was already in the method.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
stIncMale: Reverted 0893c181da2e774cc8236760869b6b3ac140f775 in 94fc5e71e3e42ad1ea984399c852cdd5693d7bb1.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
rozza: For tests also think of the change stream ticket that is linked. It would be good to confirm that is fixed.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
rozza: I dont think you need to signal anything back to the subscriber - so reactor not calling onError / onComplete post cancellation is correct.

> If a Subscription is cancelled its Subscriber MUST eventually stop being signaled.

Reactor may stop signals reaching the subscriber straight way.

rozza: So for me `callback.onResult(null, null)` seems fine.

driver-core/src/test/functional/com/mongodb/internal/operation/AsyncQueryBatchCursorFunctionalSpecification.groovy
```diff
@@ -306,6 +307,29 @@ class AsyncQueryBatchCursorFunctionalSpecification extends OperationFunctionalSp
         false     | 0
     }
 
+    @Slow
+    def 'should unblock if closed while waiting for more data from tailable cursor'() {
+        given:
+        collectionHelper.create(collectionName, new CreateCollectionOptions().capped(true).sizeInBytes(1000))
+        collectionHelper.insertDocuments(new DocumentCodec(), Document.parse('{}'))
+        def firstBatch = executeQuery(new BsonDocument('_id', BsonNull.VALUE), 0, 1, true, true);
```
stIncMale, line 315: `executeQuery` issues the following command:
```
{"find": "cappedCollectionName", "filter": {"_id": null}, "tailable": true, "awaitData": true, "limit": 0, "batchSize": 1, "$db": "JavaDriverTest", "lsid": {"id": {"$binary": {"base64": "Bfg1KfxTRE6SZBNt0yKANQ==", "subType": "04"}}}}
```
Surprisingly, if before it is issued, the collection contains 0 documents (the line 314 is commented out), then the server does not create a cursor as a result of executing the command. I looked at https://docs.mongodb.com/manual/reference/command/find/, https://docs.mongodb.com/manual/core/capped-collections/, https://docs.mongodb.com/manual/core/tailable-cursors/, and the only possibly related piece of documentation is from the last link: _"Tailable cursors may become dead, or invalid, if either: the query returns no match."_ However, the filter `{"_id": null}` does not match any documents regardless of whether the collection contains any or not.

@jyemin, @rozza do you happen to know the explanation for this behavior?

rozza: I think it will be covered by the documented caveat. There looks to be an old server ticket: [SERVER-13955](https://jira.mongodb.org/browse/SERVER-13955)

rozza (approved):
LGTM

jyemin (approved):

driver-core/src/test/functional/com/mongodb/internal/operation/AsyncQueryBatchCursorFunctionalSpecification.groovy
stIncMale: Ah, so it's a known bug. Thank you!

PR #655, opened 2021-01-31
Introduce self-managing reference counting to AsyncConnection and AsyncConnectionSource

stIncMale:
Please refer to JAVA-3978 for a detailed description of the implemented idea.

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/async/ResourcePreservingCallback.java
+++ b/driver-core/src/main/com/mongodb/internal/async/ResourcePreservingCallback.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


jyemin:
@stIncMale I think we can close this as the solution will look quite different in context of Project Reactor.

stIncMale:
> @stIncMale I think we can close this as the solution will look quite different in context of Project Reactor.

I imagine that it will still boil down to the `Stream`, and since we will still use connection pooling, we will most likely still have a `ReferenceCounted` abstraction over `Stream` with those same `doAsync(Callback)` methods that `Stream` has. Jeff, could you please confirm that what I imagine is incorrect?

P.S. I'd like GitHub to support discussions, instead of having a set of comments which represent multiple concurrent discussions with no way good way to link one comment to another.

jyemin:
I'm not sure what it will look like.  But I imagine that instead of 

`void getConnection(final SingleResultCallback<AsyncConnection> callback) `

we'll have something like

`Mono<AsynConnection> getConnection()`

P.S.  Yeah, that would be nice to have discussions in PRs like that.

stIncMale:
The thing with `Mono<AsynConnection> getConnection()` (or, more generally, `Publisher<Result> doAsync()) API is that it will still have to be adapted from `AsynchronousSocketChannel.read(callback)`/`write(callback)`. So all we can do is to push the adapting code further from the client-facing code and closer to the `AsynchronousSocketChannel`, but the code will still be there. However, if we make this adaptation "before" we need to count references (e.g., before we start pooling connections), then indeed, the idea expressed in this MR would be irrelevant.

stIncMale:
Should I close the corresponding Jira ticket?

jyemin:
I think we can leave the ticket open.

PR #656, opened 2021-02-03
Return `this` from CryptBinding.retain

jyemin:
JAVA-3974


stIncMale (approved):
It's hard to speculatively tell whether or not this breaks some existing code. I will rely on the test results.

PR #658, opened 2021-02-03
Tighten up UUID parsing in JSON Reader

jyemin:
The new spec test uncovered that UUID.fromString is more permissive in its parsing than is required by the test.  

So I added more restrictive parsing, but another option is to ignore the test.  The downside of the change I made is that it slows down the parsing code, but I doubt it will be detectable.

JAVA-3982
JAVA-3959

### DIFF
--- a/bson/src/main/org/bson/json/JsonReader.java
+++ b/bson/src/main/org/bson/json/JsonReader.java
@@ -1208,45 +1208,10 @@ public class JsonReader extends AbstractBsonReader {
         String uuidString = readStringFromExtendedJson();
         verifyToken(JsonTokenType.END_OBJECT);
         try {
-            validateHyphens(uuidString);
+            UuidStringValidator.validate(uuidString);
             return new BsonBinary(UUID.fromString(uuidString));
         } catch (IllegalArgumentException e) {
             throw new JsonParseException(e);
-        }
-    }
-
-    // UUID strings must be in the form 73ffd264-44b3-4c69-90e8-e7d1dfc035d4, but UUID.fromString fails to detect hyphens in the wrong
-    // positions.  For example, it will parse 73ff-d26444b-34c6-990e8e-7d1dfc035d4 (same as previous value but with hyphens in the wrong
-    // positions), but return a UUID that is not equal to the one it returns for the string with the hyphens in the correct positions.
-    // Given that, in order to comply with the Extended JSON specification, we add our own validation before calling UUID.fromString.
-    private void validateHyphens(final String uuidString) {
-        if (uuidString.length() != 36) {
-            throw new IllegalArgumentException(String.format("UUID string \"%s\" must be 36 characters", uuidString));
-        }
-
-        int dash1 = uuidString.indexOf('-', 0);
-        if (dash1 != 8) {
-            throw new IllegalArgumentException(String.format("First dash in UUID string \"%s\" must be at position 8", uuidString));
-        }
-
-        int dash2 = uuidString.indexOf('-', dash1 + 1);
-        if (dash2 != 13) {
-            throw new IllegalArgumentException(String.format("Second dash in UUID string \"%s\" must be at position 13", uuidString));
-        }
-
-        int dash3 = uuidString.indexOf('-', dash2 + 1);
-        if (dash3 != 18) {
-            throw new IllegalArgumentException(String.format("Third dash in UUID string \"%s\" must be at position 19", uuidString));
-        }
-
-        int dash4 = uuidString.indexOf('-', dash3 + 1);
-        if (dash4 != 23) {
-            throw new IllegalArgumentException(String.format("Fourth dash in UUID string \"%s\" must be at position 23", uuidString));
-        }
-
-        int dash5 = uuidString.indexOf('-', dash4 + 1);
-        if (dash5 != -1) {
-            throw new IllegalArgumentException(String.format("Too many dashes in UUID string \"%s\"", uuidString));
         }
     }
 
--- a/bson/src/main/org/bson/json/JsonReader.java
+++ b/bson/src/main/org/bson/json/JsonReader.java
@@ -1208,9 +1208,45 @@ public class JsonReader extends AbstractBsonReader {
         String uuidString = readStringFromExtendedJson();
         verifyToken(JsonTokenType.END_OBJECT);
         try {
+            validateHyphens(uuidString);
             return new BsonBinary(UUID.fromString(uuidString));
         } catch (IllegalArgumentException e) {
             throw new JsonParseException(e);
+        }
+    }
+
+    // UUID strings must be in the form 73ffd264-44b3-4c69-90e8-e7d1dfc035d4, but UUID.fromString fails to detect hyphens in the wrong
+    // positions.  For example, it will parse 73ff-d26444b-34c6-990e8e-7d1dfc035d4 (same as previous value but with hyphens in the wrong
+    // positions), but return a UUID that is not equal to the one it returns for the string with the hyphens in the correct positions.
+    // Given that, in order to comply with the Extended JSON specification, we add our own validation before calling UUID.fromString.
+    private void validateHyphens(final String uuidString) {
+        if (uuidString.length() != 36) {
+            throw new IllegalArgumentException(String.format("UUID string \"%s\" must be 36 characters", uuidString));
+        }
+
+        int dash1 = uuidString.indexOf('-', 0);
+        if (dash1 != 8) {
+            throw new IllegalArgumentException(String.format("First dash in UUID string \"%s\" must be at position 8", uuidString));
+        }
+
+        int dash2 = uuidString.indexOf('-', dash1 + 1);
+        if (dash2 != 13) {
+            throw new IllegalArgumentException(String.format("Second dash in UUID string \"%s\" must be at position 13", uuidString));
+        }
+
+        int dash3 = uuidString.indexOf('-', dash2 + 1);
+        if (dash3 != 18) {
+            throw new IllegalArgumentException(String.format("Third dash in UUID string \"%s\" must be at position 19", uuidString));
+        }
+
+        int dash4 = uuidString.indexOf('-', dash3 + 1);
+        if (dash4 != 23) {
+            throw new IllegalArgumentException(String.format("Fourth dash in UUID string \"%s\" must be at position 23", uuidString));
+        }
+
+        int dash5 = uuidString.indexOf('-', dash4 + 1);
+        if (dash5 != -1) {
+            throw new IllegalArgumentException(String.format("Too many dashes in UUID string \"%s\"", uuidString));
         }
     }
 
--- a/bson/src/main/org/bson/json/UuidStringValidator.java
+++ b/bson/src/main/org/bson/json/UuidStringValidator.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bson.json;
+
+import java.util.BitSet;
+
+final class UuidStringValidator {
+    private static final BitSet HEX_CHARS;
+
+    static {
+        BitSet hexChars = new BitSet(256);
+
+        hexChars.set('0', '9' + 1);
+        hexChars.set('A', 'F' + 1);
+        hexChars.set('a', 'f' + 1);
+
+        HEX_CHARS = hexChars;
+    }
+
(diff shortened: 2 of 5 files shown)


bson/src/test/resources/bson/binary.json
```diff
@@ -1,101 +1,113 @@
...
+  "parseErrors": [
+    {
+      "description": "$uuid wrong type",
+      "string": "{\"x\" : { \"$uuid\" : { \"data\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4\"}}}"
+    },
+    {
+      "description": "$uuid invalid value--too short",
+      "string": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-90e8-e7d1dfc035d4\"}}"
+    },
+    {
```
jyemin, line 100: These are the only new tests.  The rest is just reformatted white space

bson/src/test/unit/org/bson/json/JsonReaderTest.java
```diff
@@ -967,7 +967,80 @@ public void testUuid() {
             assertEquals(AbstractBsonReader.State.DONE, bsonReader.getState());
             return null;
         });
+
```
jyemin, line 970: Added these tests to get full coverage of the changes

jyemin:
I considered reporting a JDK bug but [this thread](http://mail.openjdk.java.net/pipermail/core-libs-dev/2018-January/051136.html) implies that a patch would not be accepted.

bson/src/main/org/bson/json/JsonReader.java
```diff
@@ -1208,12 +1208,48 @@ private BsonBinary visitUuidExtendedJson() {
...
+        if (dash3 != 18) {
+            throw new IllegalArgumentException(String.format("Third dash in UUID string \"%s\" must be at position 19", uuidString));
+        }
+
+        int dash4 = uuidString.indexOf('-', dash3 + 1);
+        if (dash4 != 23) {
+            throw new IllegalArgumentException(String.format("Fourth dash in UUID string \"%s\" must be at position 23", uuidString));
+        }
+
+        int dash5 = uuidString.indexOf('-', dash4 + 1);
```
rozza, line 1247: Dash 5 is checked in UUID.fromString

bson/src/main/org/bson/json/JsonReader.java
```diff
@@ -1208,12 +1208,48 @@ private BsonBinary visitUuidExtendedJson() {
...
             throw new JsonParseException(e);
         }
     }
 
+    // UUID strings must be in the form 73ffd264-44b3-4c69-90e8-e7d1dfc035d4, but UUID.fromString fails to detect hyphens in the wrong
+    // positions.  For example, it will parse 73ff-d26444b-34c6-990e8e-7d1dfc035d4 (same as previous value but with hyphens in the wrong
+    // positions), but return a UUID that is not equal to the one it returns for the string with the hyphens in the correct positions.
+    // Given that, in order to comply with the Extended JSON specification, we add our own validation before calling UUID.fromString.
+    private void validateHyphens(final String uuidString) {
+        if (uuidString.length() != 36) {
```
rozza, line 1223: This is checked in UUID.fromString

rozza (commented):
Looks to be lots of parsing bug reports that are related: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8159339

Only thing of note is I noticed two checks that are also carried out in UUID.fromString and you could just bring in the whole UUID.fromString method.

Also this fix doesn't check that the characters are hex characters so there can still be errors.  See: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8257225

bson/src/main/org/bson/json/JsonReader.java
rozza: Ah only in JDK9 and above.

bson/src/main/org/bson/json/JsonReader.java
rozza: Ah only in JDK9 and above.

bson/src/test/unit/org/bson/json/JsonReaderTest.java
```diff
@@ -967,7 +967,80 @@ public void testUuid() {
...
+        // wrong length
+        json = "{ \"$uuid\" : \"b5f21e0c2-a0d-42d6-ad03-d827008d8ab\"}}";
+        testStringAndStream(json, bsonReader -> {
+            try {
+                bsonReader.readBinaryData();
+                fail("Should fail to parse NumberDecimal constructor with a string");
+            } catch (JsonParseException e) {
+                // all good
```
stIncMale, line 978: Do we want to consider writing a separate test for each scenario with `expected = JsonParseException.class`? For example
```java
@Test(expected = JsonParseException.class)
public void testUuid_wrongLength() {
    testStringAndStream("{ \"$uuid\" : \"b5f21e0c2-a0d-42d6-ad03-d827008d8ab\"}}", bsonReader -> {
        bsonReader.readBinaryData();
        return null;
    });
}
```

stIncMale (changes requested):

bson/src/test/unit/org/bson/json/JsonReaderTest.java
jyemin: I considered it, but decided against it.  But I don't feel too strongly one way or the other.

jyemin: I ended up splitting out the validation to a separate helper class, and wrote table-driven tests of that class.

jyemin:
Given Ross's findings, I decided to just validate everything.  PTAL.

bson/src/main/org/bson/json/JsonReader.java
```diff
@@ -1208,6 +1208,7 @@ private BsonBinary visitUuidExtendedJson() {
         String uuidString = readStringFromExtendedJson();
         verifyToken(JsonTokenType.END_OBJECT);
         try {
+            UuidStringValidator.validate(uuidString);
```
rozza, line 1211: Q: Once we've done all that validation shouldn't we just create the UUID, rather than push it through the validation in UUID.fromString again?

rozza (approved):
LGTM - if we are worried about performance we could just create the UUID now that we've parsed / validated it all.

bson/src/test/unit/org/bson/json/UuidStringValidatorTest.java
```diff
@@ -0,0 +1,107 @@
...
+            "aaaaaaaa-aaaa-aaaa-aaaa-aaaaa:aaaaaa",
+            "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaa:aaaaa",
+            "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaa:aaaa",
+            "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaa:aaa",
+            "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaa:aa",
+            "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaa:a",
+            "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaa:"})
+    public void testInvalidUuidStrings(final String uuidString) {
+        try {
+            validate(uuidString);
```
rozza, line 101: Use assertThrows ?

jyemin: Cool, forgot about that.  I also decided to use `assertDoesNotThrow` in the first test since otherwise there are no assertions.  That method makes the intent more obvious.

bson/src/main/org/bson/json/JsonReader.java
jyemin: I thought about it, but didn't want to take over that responsibility.  We could always do it later if there are complaints about performance.

bson/src/main/org/bson/json/UuidStringValidator.java
```diff
@@ -0,0 +1,75 @@
...
+    private static final BitSet HEX_CHARS;
+
+    static {
+        BitSet hexChars = new BitSet(256);
+
+        hexChars.set('0', '9' + 1);
+        hexChars.set('A', 'F' + 1);
+        hexChars.set('a', 'f' + 1);
+
+        HEX_CHARS = hexChars;
```
jyemin, line 31: @stIncMale the Javadoc for BitSet states that it's not safe for multithreaded use without external synchronization.  But given that the use is read-only after assignment to a static final field, I think that is sufficient synchronization.  Is that your understanding as well?

jyemin: Actually, I don't see anything in https://docs.oracle.com/javase/specs/jls/se15/html/jls-17.html#jls-17.4.5 that indicates that this is safe.

stIncMale: Initializing static state in a static initializer and only reading it outside of the initializer is thread-safe: roughly speaking, all write actions produced by a static initializer happen-before reads (and writes, but that is irrelevant for this specific situation) produced by the code outside of it.

Unfortunately, there is no note in the JMM stating this, and we are left figuring out on our own from different places in the JLS. The above statement follows from [JLS 12.4.1. When Initialization Occurs](https://docs.oracle.com/javase/specs/jls/se15/html/jls-12.html#jls-12.4.1), [12.4.2. Detailed Initialization Procedure](https://docs.oracle.com/javase/specs/jls/se15/html/jls-12.html#jls-12.4.2),

stIncMale: Just in case: having a local variable `hexChars` for the purpose of modifying the `BitSet` state and then writing `HEX_CHARS = hexChars` does not change anything (the same is true for `final` fields because a [_"freeze action on `final` field f of `o` takes place when `c` (constructor) exits, either normally or abruptly"_](https://docs.oracle.com/javase/specs/jls/se15/html/jls-17.html#jls-17.5.1)). So we can just have the following code:
```java
static {
    HEX_CHARS = new BitSet(256);
    HEX_CHARS.set('0', '9' + 1);
    HEX_CHARS.set('A', 'F' + 1);
    HEX_CHARS.set('a', 'f' + 1);
}
```

jyemin: I implemented your suggested change

bson/src/test/unit/org/bson/json/JsonReaderTest.java
```diff
@@ -967,73 +967,17 @@ public void testUuid() {
...
-            return null;
-        });
-
-        // too many hyphens
-        json = "{ \"$uuid\" : \"b5f21e0c-2a0d-42d6-ad03-d827008d8ab-\"}}";
-        testStringAndStream(json, bsonReader -> {
-            try {
-                bsonReader.readBinaryData();
-                fail("Should fail to parse NumberDecimal constructor with a string");
+                fail("Should fail to parse invalid UIUD");
```
stIncMale, line 980: It appears that we could be mo expressive here by using either `assertThrows` just like it is used in  `UuidStringValidatorTest.java`, or `@Test(expected = JsonParseException.class)`.

bson/src/main/org/bson/json/UuidStringValidator.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    private static void validateDash(final String str, final int pos) {
+        if (str.charAt(pos) != '-') {
+            throw new IllegalArgumentException(String.format("Expected dash in UUID string \"%s\" at position %d", str, pos));
+        }
+    }
+
+    // UUID strings must be in the form 73ffd264-44b3-4c69-90e8-e7d1dfc035d4, but UUID.fromString fails to detect hyphens in the wrong
```
stIncMale: This comment is simply a question, not an objection to the changes in the current PR.

Q:
While I understand that we do not publish API docs for this method, what are the reasons to write docs for internal code not in the javadoc format? The downsides of such a practice are:
- it forces reviewers to check whether or not the documented code is a part of public API;
- it introduces inconsistency in how docs are written;
- it prevents reading the doc without jumping to the source code.

bson/src/main/org/bson/json/UuidStringValidator.java
```diff
@@ -22,7 +22,7 @@
     private static final BitSet HEX_CHARS;
 
     static {
-        HEX_CHARS = new BitSet(256);
+        HEX_CHARS = new BitSet('f' + 1);
```
stIncMale, line 25: [Ignore me]
We could make it even smaller by shifting all char values by `(int)'0'` when converting them to `BitSet` indices.

stIncMale (changes requested):

bson/src/main/org/bson/json/UuidStringValidator.java
jyemin: I will ignore you.  I little too clever just to save 64 bits.  ;)

bson/src/main/org/bson/json/UuidStringValidator.java
jyemin: Let's discuss offline

stIncMale: Saved for a retro meeting.

bson/src/test/unit/org/bson/json/JsonReaderTest.java
jyemin: Done

stIncMale: 👍

stIncMale (approved):

jyemin:
Rebased and merged on command line since I wanted to preserve two commits

PR #659, opened 2021-02-04
JAVA-3684: deprecate geohaystack

terakilobyte:
This removes mention of geoHaystack from the docs and marks methods as deprecated.

### DIFF
--- a/docs/reference/content/builders/indexes.md
+++ b/docs/reference/content/builders/indexes.md
@@ -123,19 +123,3 @@ This example specifies a 2d index on the `points` field:
 geo2d("points")
 ```
 
-
-#### geoHaystack
-
-To specify a [geoHaystack]({{< docsref "core/geohaystack/" >}}) index key, use the `geoHaystack` method.
-
-{{% note class="important"%}}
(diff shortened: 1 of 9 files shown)


rozza (changes requested):
Looks good, only one thing, where methods are deprecated you should also add the `@deprecated` tag and message in the java docs

terakilobyte:
I thought I had added the @deprecated tags. Do you mean to add a quick note within the method docs it is deprecated and what to use instead?

rozza:
@terakilobyte - here's an example: https://github.com/mongodb/mongo-java-driver/blob/8574c6daff4b8a1ecf65c67a301287bc919b5b7c/driver-core/src/main/com/mongodb/event/ConnectionPoolListener.java#L133-L138

terakilobyte:
I've added to the method docs for the public facing methods. I was unsure if it needed to be done for the methods in internal based on driver policy.

rozza:
@terakilobyte unfortunately you have to double annotate with Java once for thing being deprecated and then again in the javadoc.  Scala is nicer as the annotation is combined.

I've updated the java docs and will squash and merge.

Many thanks

PR #661, opened 2021-02-05
Regression test for change stream cancellation

rozza:
Ensures that all sessions are returned to the pool - and therefore the connections are released.

JAVA-3938 JAVA-3907

### DIFF
--- a/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ChangeStreamsCancellationTest.java
+++ b/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ChangeStreamsCancellationTest.java
@@ -23,7 +23,8 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Mono;
 
-import static com.mongodb.reactivestreams.client.Fixture.dropDatabase;
+import static com.mongodb.ClusterFixture.TIMEOUT_DURATION;
+import static com.mongodb.reactivestreams.client.Fixture.drop;
 import static com.mongodb.reactivestreams.client.Fixture.getDefaultDatabase;
 import static com.mongodb.reactivestreams.client.Fixture.isReplicaSet;
 import static com.mongodb.reactivestreams.client.Fixture.serverVersionAtLeast;
@@ -32,25 +33,24 @@ import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 public class ChangeStreamsCancellationTest {
 
-    private MongoDatabase database;
+    private MongoCollection<Document> collection;
 
     @BeforeEach
     public void setup() {
         assumeTrue(isReplicaSet() && serverVersionAtLeast(3, 6));
-        database = getDefaultDatabase();
+        collection = getDefaultDatabase().getCollection("changeStreamsCancellationTest");
     }
 
     @AfterEach
     public void tearDown() {
-        if (database != null) {
-            dropDatabase(database.getName());
+        if (collection != null) {
+            drop(collection.getNamespace());
         }
     }
 
     @Test
     public void testCancelReleasesSessions() {
-        MongoCollection<Document> collection = database.getCollection("changeStreamsCancellationTest");
-        Mono.from(collection.insertOne(new Document()));
+        Mono.from(collection.insertOne(new Document())).block(TIMEOUT_DURATION);
 
         TestSubscriber<ChangeStreamDocument<Document>> subscriber = new TestSubscriber<>();
         subscriber.doOnSubscribe(sub -> {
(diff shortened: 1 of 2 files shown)


driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ChangeStreamsCancellationTest.java
```diff
@@ -0,0 +1,73 @@
...
+    @BeforeEach
+    public void setup() {
+        assumeTrue(isReplicaSet() && serverVersionAtLeast(3, 6));
+        database = getDefaultDatabase();
+    }
+
+    @AfterEach
+    public void tearDown() {
+        if (database != null) {
+            dropDatabase(database.getName());
```
jyemin, line 46: We generally avoid dropping the default database?  Is it necessary?

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ChangeStreamsCancellationTest.java
```diff
@@ -0,0 +1,73 @@
...
+        Mono.from(collection.insertOne(new Document()));
+
+        TestSubscriber<ChangeStreamDocument<Document>> subscriber = new TestSubscriber<>();
+        subscriber.doOnSubscribe(sub -> {
+            sub.request(Integer.MAX_VALUE);
+            new Thread(() -> {
+                try {
+                    Thread.sleep(1000);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
```
jyemin, line 62: This is a place where InterruptedException can be safely ignored, since there are no references to this thread anywhere that could interrupt it.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ChangeStreamsCancellationTest.java
```diff
@@ -0,0 +1,73 @@
...
+    public void tearDown() {
+        if (database != null) {
+            dropDatabase(database.getName());
+        }
+    }
+
+    @Test
+    public void testCancelReleasesSessions() {
+        MongoCollection<Document> collection = database.getCollection("changeStreamsCancellationTest");
+        Mono.from(collection.insertOne(new Document()));
```
jyemin, line 53: What's the purpose of the insert?

jyemin (changes requested):

stIncMale:
@rozza Just in case, the test added [here](https://github.com/mongodb/mongo-java-driver/pull/654/commits/ab5162cdc8587a819ceaa26aecfb4e7ad87471f6) checks JAVA-3907 and JAVA-3938. I did not write a full-blown test that checks the behavior from the user perspective because based on the tests I saw, I assumed this is not what we do (also expressed my thoughts about the test and proposed discussing them in this [comment](https://github.com/mongodb/mongo-java-driver/pull/654#discussion_r568202971)).

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ChangeStreamsCancellationTest.java
stIncMale: I would like to express my thoughts about this in two parts: general and specific.

###### General
I fail to see benefits from taking a shortcut when there is an opportunity for one if it does not drastically simplify the code / improve its readability. Not only this impairs developing good coding habits, but also imposes a burden on both writers and readers of assessing whether or not taking a shortcut is safe or not.

###### Specific
We may either write code without taking a shortcut
```java
try {
    Thread.sleep(1000);
}
```
or take the shortcut
```java
try {
    Thread.sleep(1000);
    //cannot happen because no code has a reference to the current thread
}
```

The first approach results in an expected code pattern and does not require additional thinking about the way it handles `InterruptedException`. The second approach produces the same number of lines, but requires a reader to read, understand and verify the comment.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ChangeStreamsCancellationTest.java
stIncMale: Please forgive my interfering. FYI: the test still works without this line and fails (expectedly) if I roll back the changes made in JAVA-3907.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ChangeStreamsCancellationTest.java
jyemin: What's expected is in the eye of the beholder.  To me it's more natural to ignore the exception.
 
But I'm not attached enough to argue the point.

rozza (commented):
So this test confirms that cancellation is propagated and that the cursor is closed.  The tests added as part of JAVA-3907 test the async cursor at a lower level and is desirable. I see value in having both, especially having the higher level regression test, because if a future change didn't propagate the cancellation of the subscription then we wouldn't catch it.

stIncMale (approved):

jyemin (approved):

PR #662, opened 2021-02-10
Use $collStats for estimatedDocumentCount on 5.0+ servers

jyemin:
There are three commits, which I'll summarize here:

1. Refactor CountOperation into three operations: CountOperation was already complex in that it was supporting three different use cases.  One for estimatedDocumentCount, one for countDocuments, and one for the legacy DBCollection.count.  Trying to add the $collStats support into this class was going to make it even messier, as now there would be two different aggregation pipelines.  It was just too much, so I split it out, which allowed me to get rid of all the conditional logic.  
2. Add support for errorCode to unified test runner.  This was just a small missing piece that was needed by the new tests
3. The last commit is the functional change required by the spec, plus all the new tests.

JAVA-3989

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/ServerVersionHelper.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/ServerVersionHelper.java
@@ -31,6 +31,7 @@ public final class ServerVersionHelper {
     public static final int FOUR_DOT_ZERO_WIRE_VERSION = 7;
     public static final int FOUR_DOT_TWO_WIRE_VERSION = 8;
     public static final int FOUR_DOT_FOUR_WIRE_VERSION = 9;
+    public static final int FIVE_DOT_ZERO_WIRE_VERSION = 12;
 
     public static boolean serverIsAtLeastVersionThreeDotZero(final ConnectionDescription description) {
         return description.getMaxWireVersion() >= THREE_DOT_ZERO_WIRE_VERSION;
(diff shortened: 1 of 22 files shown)


driver-core/src/main/com/mongodb/internal/operation/ServerVersionHelper.java
```diff
@@ -31,6 +31,7 @@
     public static final int FOUR_DOT_ZERO_WIRE_VERSION = 7;
     public static final int FOUR_DOT_TWO_WIRE_VERSION = 8;
     public static final int FOUR_DOT_FOUR_WIRE_VERSION = 9;
+    public static final int FIVE_DOT_ZERO_WIRE_VERSION = 12;
```
rozza, line 34: Oww big jump!

driver-sync/src/test/functional/com/mongodb/client/ReadConcernTest.java
```diff
@@ -68,18 +69,16 @@ public void cleanUp() {
     @Test
     public void shouldIncludeReadConcernInCommand() {
         mongoClient.getDatabase(getDefaultDatabaseName()).getCollection("test")
-                .withReadConcern(ReadConcern.LOCAL).estimatedDocumentCount();
+                .withReadConcern(ReadConcern.LOCAL).find().into(new ArrayList<>());
```
rozza, line 72: No Async equivalent?

rozza (approved):

driver-sync/src/test/functional/com/mongodb/client/ReadConcernTest.java
jyemin: There is a ReadConcernTest in reactive streams module that was also modified.

PR #665, opened 2021-02-11
Prepare the change stream code for tests in the unified format

stIncMale:
This PR creates two unified runners for change stream tests and allows putting unified change stream tests in `driver-core/src/test/resources/unified-test-format/change-streams`.

The PR also implements `SyncChangeStreamIterable.withDocumentClass` as this method is used by the `UnifiedTest` runner (see `UnifiedCrudHelper.executeChangeStream`).

All changes are in the test code.

JAVA-4011

### DIFF
--- a/bson/src/test/unit/util/JsonPoweredTestHelper.java
+++ b/bson/src/test/unit/util/JsonPoweredTestHelper.java
@@ -66,7 +66,7 @@ public final class JsonPoweredTestHelper {
     /**
      * @param resourcePath             The {@linkplain Class#getResource(String) name of a resource} representing
      *                                 a directory with test files.
-     * @param excludedRelativeSubpaths Sub-paths relative to the {@code resourcePath} directory to excluded from the results.
+     * @param excludedRelativeSubpaths Sub-paths relative to the {@code resourcePath} directory to exclude from the results.
      * @return A collection with unspecified mutability and element order.
      */
     public static Collection<Path> getTestFiles(final String resourcePath, final Path... excludedRelativeSubpaths)
--- a/bson/src/test/unit/util/JsonPoweredTestHelper.java
+++ b/bson/src/test/unit/util/JsonPoweredTestHelper.java
@@ -29,63 +29,28 @@ import java.io.InputStreamReader;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.Charset;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.util.Collection;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public final class JsonPoweredTestHelper {
-    private static final String JSON_FILE_NAME_SUFFIX = ".json";
 
     public static BsonDocument getTestDocument(final File file) throws IOException {
         return new BsonDocumentCodec().decode(new JsonReader(getFileAsString(file)), DecoderContext.builder().build());
-    }
-
-    public static BsonDocument getTestDocument(final Path file) throws IOException {
-        return getTestDocument(file.toFile());
     }
 
     public static BsonDocument getTestDocument(final String resourcePath) throws IOException, URISyntaxException {
         return getTestDocument(new File(JsonPoweredTestHelper.class.getResource(resourcePath).toURI()));
     }
 
-    /**
-     * @see #getTestFiles(String, Path...)
-     */
-    public static Collection<File> getTestFiles(final String resourcePath) throws URISyntaxException, IOException {
-        return getTestFiles(resourcePath, (Path[]) null)
-                .stream()
-                .map(Path::toFile)
-                .collect(Collectors.toList());
-    }
-
-    /**
-     * @param resourcePath             The {@linkplain Class#getResource(String) name of a resource} representing
-     *                                 a directory with test files.
-     * @param excludedRelativeSubpaths Sub-paths relative to the {@code resourcePath} directory to exclude from the results.
-     * @return A collection with unspecified mutability and element order.
-     */
-    public static Collection<Path> getTestFiles(final String resourcePath, final Path... excludedRelativeSubpaths)
-            throws URISyntaxException, IOException {
+    public static List<File> getTestFiles(final String resourcePath) throws URISyntaxException {
         URL resource = JsonPoweredTestHelper.class.getResource(resourcePath);
         if (resource == null) {
             return Collections.emptyList();
         } else {
-            Path start = Paths.get(resource.toURI());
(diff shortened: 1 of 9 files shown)


driver-sync/src/test/functional/com/mongodb/client/AbstractChangeStreamsTest.java
```diff
@@ -252,8 +253,8 @@ private JsonPoweredCrudTestHelper createJsonPoweredCrudTestHelper(final MongoCli
 
     @Parameterized.Parameters(name = "{0}: {1}")
     public static Collection<Object[]> data() throws URISyntaxException, IOException {
-        List<Object[]> data = new ArrayList<Object[]>();
-        for (File file : JsonPoweredTestHelper.getTestFiles("/change-streams")) {
+        List<Object[]> data = new ArrayList<>();
+        for (Path file : JsonPoweredTestHelper.getTestFiles("/change-streams", Paths.get("unified"))) {
```
jyemin, line 257: I think you can avoid this change if you instead put the unified tests in 

    /unified-test-format/change-streams

instead of

    /change-streams/unified

That's what I plan to do for the new unified crud tests I'm adding in scope of the ticket I'm working on now.

bson/src/test/unit/util/JsonPoweredTestHelper.java
```diff
@@ -27,24 +27,66 @@
...
+        } else {
+            Path start = Paths.get(resource.toURI());
+            List<Path> excluded = excludedRelativeSubpaths == null ? Collections.emptyList() : Stream.of(excludedRelativeSubpaths)
+                    .map(start::resolve)
+                    .collect(Collectors.toList());
+            try (Stream<Path> files = Files.walk(start)) {
+                return files
+                        .filter(path -> !Files.isDirectory(path))
+                        .filter(path -> path.getNameCount() > 0 && path.getFileName().toString().endsWith(JSON_FILE_NAME_SUFFIX))
+                        .filter(path -> excluded.stream().noneMatch(path::startsWith))
```
stIncMale, line 86: The efficient way to implement exclusion of sub-paths is to use the `Files.walkFileTree` method and return `SKIP_SUBTREE` from `FileVisitor.preVisitDirectory`. However, for our purposes here making the code shorter seems to be more important, than making it efficient. However, if someone thinks this is a problem, I can implement exclusion via `Files.walkFileTree`.

driver-sync/src/test/functional/com/mongodb/client/AbstractChangeStreamsTest.java
stIncMale: I thought about doing this, but `VersionedApiTest` has scenarios in `driver-core/src/test/resources/versioned-api`, not in `driver-core/src/test/resources/unified-test-format/change-streams`. So to be more consistent, I had to use `driver-core/src/test/resources/change-streams/unified`.

Given the fact about `VersionedApiTest`, do you still think I should change the unified `ChangeStreamsTest` runners to take tests from `driver-core/src/test/resources/unified-test-format/change-streams`?

jyemin: Yes, I think we should move all the unified tests under that directory.

stIncMale: Great!, this simplifies the changes.

stIncMale: Done.

bson/src/test/unit/util/JsonPoweredTestHelper.java
stIncMale: This comment is not relevant anymore.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncMongoCursor.java
```diff
@@ -27,8 +28,11 @@
 class SyncMongoCursor<T> implements MongoCursor<T> {
     private final Iterator<T> iterator;
 
-    SyncMongoCursor(final Publisher<T> publisher) {
-        iterator = Flux.from(publisher).toIterable().iterator();
+    SyncMongoCursor(final Publisher<T> publisher, final @Nullable Integer batchSize) {
+        iterator = (batchSize == null
```
rozza, line 32: Does this make any difference?

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestValidator.java
```diff
@@ -53,6 +53,6 @@ protected MongoClient createMongoClient(final MongoClientSettings settings) {
 
     @Parameterized.Parameters(name = "{0}: {1}")
     public static Collection<Object[]> data() throws URISyntaxException, IOException {
-        return getTestData("unified-test-format");
+        return getTestData("unified-test-format/valid-pass");
```
jyemin, line 56: Good catch

bson/src/test/unit/util/JsonPoweredTestHelper.java
```diff
@@ -42,9 +44,14 @@ public static BsonDocument getTestDocument(final String resourcePath) throws IOE
     }
 
     public static List<File> getTestFiles(final String resourcePath) throws URISyntaxException {
-        List<File> files = new ArrayList<File>();
-        addFilesFromDirectory(new File(JsonPoweredTestHelper.class.getResource(resourcePath).toURI()), files);
-        return files;
+        URL resource = JsonPoweredTestHelper.class.getResource(resourcePath);
```
jyemin, line 47: I guess this is needed because currently there are no unified change stream tests?  

But I worry that by not throwing an exception in this case we might inadvertently hide a mistake in the resourcePath passed by the caller.

stIncMale: Yes, that may be a problem. I can add a dummy file with no tests and remove this check. Would such solution be acceptable?

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncMongoCursor.java
stIncMale: Without this code we lose the explicitly set batch size and the Reactor uses (requests) 256 instead as its default.

rozza: Ack

bson/src/test/unit/util/JsonPoweredTestHelper.java
stIncMale: Added `unified-test-format/change-streams/change-streams.json` without tests.

jyemin (approved):
LGTM

PR #667, opened 2021-02-13
Add the UpdateDescription.truncatedArrays field

stIncMale:
JAVA-3990 (implements SPEC-1779)


driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
```diff
@@ -56,15 +77,40 @@ public UpdateDescription(@Nullable @BsonProperty("removedFields") final List<Str
...
 
+    /**
+     * Returns information about the updated fields of the {@linkplain org.bson.BsonType#ARRAY array} type
+     * when the changes are reported as truncations.
+     *
+     * @return {@code truncatedArrays}.
+     * @see #getUpdatedFields()
+     */
+    @Nullable
+    public List<TruncatedArray> getTruncatedArrays() {
```
stIncMale, line 98: Returning `null` from a method of a container type (`List` in this case) is an API design that not only violates the principle of least astonishment, but also forces users to write code that checks for both `null` and `isEmpty` instead of checking only `isEmpty`. Unfortunately, the test `ChangeStreamDocumentCodecSpecification.'should round trip ChangeStreamDocument successfully'` is written in such a way that prevents `UpdateDescription` from always returning an empty `List` instead of `null`. This in turn may be related to the behavior of BSON codec functionality.

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
```diff
@@ -31,18 +33,37 @@
...
 
     /**
      * Creates a new instance
      *
      * @param removedFields the list of fields that have been removed.
      * @param updatedFields the updated fields
+     * @deprecated Not part of the public API since 4.3, may be removed in the future.
+     */
+    @Deprecated
+    public UpdateDescription(@Nullable final List<String> removedFields,
```
stIncMale, line 47: We cannot remove this constructor because from the user standpoint it was part of the public API. The reason it had to be `public` is that this is required by the BSON codec functionality, but it could have been documented as not being part of the public API.

driver-core/src/test/functional/com/mongodb/internal/operation/ChangeStreamOperationSpecification.groovy
```diff
@@ -227,7 +227,7 @@ class ChangeStreamOperationSpecification extends OperationFunctionalSpecificatio
         nextDoc.getFullDocument() == BsonDocument.parse('{ _id : 2, x : 3 }')
         nextDoc.getNamespace() == helper.getNamespace()
         nextDoc.getOperationType() == OperationType.UPDATE
-        nextDoc.getUpdateDescription() == new UpdateDescription(['y'], BsonDocument.parse('{x : 3}'))
+        nextDoc.getUpdateDescription() == new UpdateDescription(['y'], BsonDocument.parse('{x : 3}'), null)
```
stIncMale, line 230: It is not specified whether the MongoDB server returns `null` or `[]` when no changes are reported as truncation. The current behavior appears to be returning `[]`. The method `UpdateDescription.equals` treats `null` and empty `List` as equal values, this way we are prepared to face any of `null`/`[]`, and can safely specify `null` in this test as our expectation.

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
rozza: Perhaps change the test? Or always return a list from this method?

driver-core/src/test/functional/com/mongodb/internal/operation/ChangeStreamOperationSpecification.groovy
rozza: Well from released servers its null but from now on it maybe an empty list

stIncMale: Yes, it may be an empty list since MongoDB 4.7, or a null. The server design does not specify that one of these values is not possible, and we cannot depends on the behavior we currently observe: this same design is an example of a situation when the server [changed its behaviour because it was not specified](https://docs.google.com/document/d/1sxGrck8X0r0W37UVLFwV8vaUQv3OM8xlyWWvHB5M2ew/edit#heading=h.22yp33tzmzpp).

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
stIncMale: > Perhaps change the test?

In order to change the test I would also need to implement the possibility to either not serialize empty lists in BSON or not deserialize empty lists from BSON. This comment simply explains a problem with the current `ChangeStreamDocument` and BSON serialization functionality, and states that the newly added method suffers from that same problem.

I think that if we want to improve the `ChangeStreamDocument` API (always return empty `List` values instead of returning `null values), it requires a separate task.

driver-core/src/test/functional/com/mongodb/internal/operation/ChangeStreamOperationSpecification.groovy
rozza: Ack

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
rozza: I agree and think that always returning a list here is what is desirable.

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
rozza: If users can consume these types in their applications, then I think being able to easily create them for their testing scenarios is important. 

For that reason I'm not against the constructors being public and possibly wouldn't even deprecate this method, just have it call the newer constructor.

driver-core/src/main/com/mongodb/client/model/changestream/TruncatedArray.java
```diff
@@ -0,0 +1,92 @@
...
+ * when the change is reported as truncation.
+ *
+ * @since 4.3
+ */
+public final class TruncatedArray {
+    private final String field;
+    private final int newSize;
+
+    /**
+     * Not part of the public API.
```
rozza, line 38: I'm not sure this is the case.

rozza (commented):
I've added some comments for discussion.

driver-core/src/main/com/mongodb/client/model/changestream/TruncatedArray.java
jyemin: I think this would be the first time we've tried to make a single method of an otherwise public class not part of the public API.  I'm not sure I like the precedent.

stIncMale: This is why I asked if users are supposed to create instances of `UpdateDescription` (a part of `ChangeStreamDocument`) on their own. Since the answer was "no", I concluded that constructors of these classes are not part of the public API. If that is not the case, then then the answer to the mentioned question is "yes", i.e., users are supposed to create instances of these classes, including the new `TruncatedArray` class (@rozza pointed out a [good example](https://github.com/mongodb/mongo-java-driver/pull/667#discussion_r576681615) when this may be needed).

Thank you, guys, I will make the changes necessary to express that the constructors are a part of public API.

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/changestream/TruncatedArray.java
stIncMale: Done.

driver-core/src/test/resources/unified-test-format/change-streams/change-streams.yml
```diff
@@ -0,0 +1,74 @@
+description: "change-streams"
```
jyemin, line 1: In java driver we've chosen not to copy the YAML files, but rather only the JSON files.  Can you remove it?

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -339,8 +335,13 @@ OperationResult executeUpdateOne(final BsonDocument operation) {
                     throw new UnsupportedOperationException("Unsupported argument: " + cur.getKey());
             }
         }
-        return resultOf(() ->
-                toExpected(collection.updateOne(filter, update, options)));
+        BsonDocument filter = arguments.getDocument("filter");
```
jyemin, line 338: Nit: In all the other methods the required arguments are pulled out before the options.  Consider whether we should be consistent about that.

jyemin (changes requested):

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
stIncMale: Done. Surprisingly, this did not require changing anything in the serialization mechanism.

bson/src/main/org/bson/codecs/EncoderContext.java
```diff
@@ -25,8 +25,10 @@
  * @since 3.0
  */
 public final class EncoderContext {
-
-    private static final EncoderContext DEFAULT_CONTEXT = EncoderContext.builder().build();
+    /**
+     * The default immutable {@link EncoderContext}. Its method {@link #isEncodingCollectibleDocument()} returns {@code false}.
+     */
+    public static final EncoderContext DEFAULT_CONTEXT = EncoderContext.builder().build();
```
stIncMale, line 31: Please pay attention to this change. Maybe there are reasons why we may not want to expose this field as part of the public API.

I was unable to understand the usage of `EncoderContext` and would appreciate if you guys could help we with answering the following questions:
1. What is the purpose of `EncoderContext.getChildContext`? It returns a final reference `DEFAULT_CONTEXT` to an immutable object, which means that currently this method is futile.
2. The method `EncoderContext.encodeWithChildContext` appears to be futile for the same reason as specified in 1: it always uses `DEFAULT_CONTEXT`.
3. As far as I can tell, currently the class `EncoderContext` could be removed with replacing invocations of the setter method `EncoderContext.Builder.isEncodingCollectibleDocument` with the values currently passed as its arguments (some method signatures and potentially field types may need to be updated).

rozza: EncoderContext / DecodeContext were invisaged so codecs could understand more about how to encode/decode the data they have and specialize if needed.

> 1.What is the purpose of EncoderContext.getChildContext? It returns a final reference DEFAULT_CONTEXT to an immutable object, which means that currently this method is futile.

Currently, it signals you are encoding a nested field / document so don't need to add a _id field.

> 2. The method EncoderContext.encodeWithChildContext appears to be futile for the same reason as specified in 1: it always uses DEFAULT_CONTEXT.

Its the method used to signal you are encoding a nested field.

> 3. As far as I can tell, currently the class EncoderContext could be removed with replacing invocations of the setter method EncoderContext.Builder.isEncodingCollectibleDocument with the values currently passed as its arguments (some method signatures and potentially field types may need to be updated).

It would be a major breaking API change to remove it, requiring all Codec implementations to be changed. So all internal Codecs and more importantly all userland Codecs would have to change. That in itself is prohibitive and doesn't provide added value to do so.

I think what you are really want to know is "_Why does EncoderContext exist and why was it designed this way?_"

The context pattern is extendable for future unknown requirements when processing data to / from bson. Extending a context doesn't require any major version changes. One example where a change has happened in the past has been for Pojo discriminator support in the DecoderContext.  

The success of the Codec and BsonWriter/Reader design has meant that contexts haven't required to be extended. However, in the past flexibility and user customizability for the contexts has been considered / requested for easier integration into other libraries but so far nothing has been added for external users.

rozza (approved):
LGTM

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -692,43 +704,49 @@ public OperationResult executeCreateIndex(final BsonDocument operation) {
...
+        MongoCursor<ChangeStreamDocument<BsonDocument>> cursor = entities.getChangeStream(operation.getString("object").getValue());
 
         if (operation.containsKey("arguments")) {
             throw new UnsupportedOperationException("Unexpected arguments");
         }
 
-        return resultOf(cursor::next);
+        return resultOf(() -> {
+            BsonDocumentWriter bsonDocumentWriter = new BsonDocumentWriter(new BsonDocument());
+            changeStreamDocumentCodec.encode(bsonDocumentWriter, cursor.next(), EncoderContext.DEFAULT_CONTEXT);
```
rozza, line 747: Comment: If this is the only public use of DEFAULT_CONTEXT I probably wouldn't have made it public but I'm not against it being so either.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -68,11 +75,15 @@
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
+import static java.util.Arrays.asList;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
 
 final class UnifiedCrudHelper {
     private final Entities entities;
+    private final Codec<ChangeStreamDocument<BsonDocument>> changeStreamDocumentCodec = ChangeStreamDocument.createCodec(
```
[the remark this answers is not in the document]
jyemin: Ack

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
jyemin: I am opposed to adding it in the context of this commit, as I don't think we should mix changes to the public API with changes to a test runner.  If you feel strongly that DEFAULT_CONTEXT should be public, then let's open a separate ticket for it.

jyemin (changes requested):

jyemin (approved):

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
```diff
@@ -31,18 +35,35 @@
...
+        this(removedFields, updatedFields, null);
+    }
+
+    /**
+     * @param removedFields   Names of the fields that were removed.
+     * @param updatedFields   Information about the updated fields.
+     * @param truncatedArrays Information about the updated fields of the {@linkplain org.bson.BsonType#ARRAY array} type
+     *                        when the changes are reported as truncations.
+     *                        If {@code null}, then {@link #getTruncatedArrays()} returns an {@linkplain List#isEmpty() empty} {@link List}.
      */
```
jyemin, line 58: Add `@since 4.3`

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
```diff
@@ -56,15 +77,60 @@ public UpdateDescription(@Nullable @BsonProperty("removedFields") final List<Str
...
     }
 
+    /**
+     * Returns information about the updated fields of the {@linkplain org.bson.BsonType#ARRAY array} type
+     * when the changes are reported as truncations.
+     *
+     * @return {@code truncatedArrays}.
+     * There are no guarantees on the mutability of the {@code List} returned.
+     * @see #getUpdatedFields()
+     */
```
jyemin, line 116: Add `@since 4.3`

jyemin (changes requested):
Sorry, missed a couple things.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
stIncMale: Done. Thanks for spotting this! I should have been more mindful.

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/changestream/UpdateDescription.java
jyemin: We miss these All The Time.  I wish we could come up with a linter for it.  The problem is that many of the older methods don't have this javadoc annotation, so we can't just apply it to the whole code base.

jyemin (approved):

jyemin:
LGTM!

stIncMale:
All static checks pass for me locally with the latest change.

bson/src/main/org/bson/codecs/EncoderContext.java
stIncMale: @rozza Thank you so much for the detailed explanation! 👍

PR #674, opened 2021-02-19
Remove assumption that jdk.net package is available

jyemin:
This commit removes any assumptions that jdk.net.ExtendedSocketOptions
is available at all at runtime, since the jdk.net package is not actually
part of the standard library (though in practice it is usually available).
If it's not available, no TCP keep-alive options will be set, but socket
creation will proceed instead of failing with a ClassNotFoundException.

JAVA-4005

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/SocketStreamHelper.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/SocketStreamHelper.java
@@ -19,21 +19,21 @@ package com.mongodb.internal.connection;
 import com.mongodb.MongoInternalException;
 import com.mongodb.connection.SocketSettings;
 import com.mongodb.connection.SslSettings;
-import jdk.net.ExtendedSocketOptions;
 
 import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocket;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.net.SocketOption;
-import java.util.Arrays;
 
 import static com.mongodb.internal.connection.SslHelper.enableHostNameVerification;
 import static com.mongodb.internal.connection.SslHelper.enableSni;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
+@SuppressWarnings({"unchecked", "rawtypes"})
 final class SocketStreamHelper {
     // Keep alive options and their values for Java 11+
     private static final String TCP_KEEPIDLE = "TCP_KEEPIDLE";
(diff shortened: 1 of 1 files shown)


driver-core/src/main/com/mongodb/internal/connection/SocketStreamHelper.java
```diff
@@ -78,19 +107,21 @@ static void initialize(final Socket socket, final InetSocketAddress inetSocketAd
...
+                    SET_OPTION_METHOD.invoke(socket, KEEP_COUNT_OPTION, TCP_KEEPCOUNT_LIMIT);
+                }
+                if (KEEP_IDLE_OPTION != null) {
+                    SET_OPTION_METHOD.invoke(socket, KEEP_IDLE_OPTION, TCP_KEEPIDLE_DURATION);
+                }
+                if (KEEP_INTERVAL_OPTION != null) {
+                    SET_OPTION_METHOD.invoke(socket, KEEP_INTERVAL_OPTION, TCP_KEEPINTERVAL_DURATION);
+                }
             }
+        } catch (IllegalAccessException | InvocationTargetException e) {
```
stIncMale, line 123: The method `setExtendedSocketOptions` is "best effort" only with regard to working around access issues and unavailability of a method/class/fields. If everything is present and the code does have access to invoke the `setOption` method, then this method throwing an exception is a problem that must not be ignored, but must be treated like any other exception happening when we set other socket options in a more traditional way.

driver-core/src/main/com/mongodb/internal/connection/SocketStreamHelper.java
```diff
@@ -78,19 +107,21 @@ static void initialize(final Socket socket, final InetSocketAddress inetSocketAd
...
-                setOptionMethod.invoke(socket, ExtendedSocketOptions.class.getDeclaredField(TCP_KEEPIDLE).get(null),
-                        TCP_KEEPIDLE_DURATION);
-                setOptionMethod.invoke(socket, ExtendedSocketOptions.class.getDeclaredField(TCP_KEEPINTERVAL).get(null),
-                        TCP_KEEPINTERVAL_DURATION);
-            } catch (Throwable t) {
+    static void setExtendedSocketOptions(final Socket socket) {
+        try {
+            if (SET_OPTION_METHOD != null) {
+                if (KEEP_COUNT_OPTION != null) {
+                    SET_OPTION_METHOD.invoke(socket, KEEP_COUNT_OPTION, TCP_KEEPCOUNT_LIMIT);
```
stIncMale, line 114: This PR made me think, why does the driver enable the TCP keep alive mechanism when connecting to a MongoDB server? [RFC 1122. 4.2.3.6 TCP Keep-Alives](https://www.rfc-editor.org/rfc/rfc1122#page-101):

> A TCP keep-alive mechanism should only be invoked in
> server applications that might otherwise hang
> indefinitely and consume resources unnecessarily if a
> client crashes or aborts a connection during a network
> failure.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/connection/SocketStreamHelper.java
jyemin: There is discussion of the benefits in https://jira.mongodb.org/browse/DRIVERS-383

driver-core/src/main/com/mongodb/internal/connection/SocketStreamHelper.java
jyemin: Yeah, I noticed that too.  I was trying not to change the existing behavior, which also ignores this exception, just in case there is a problem calling this method that we're not aware.  I don't want to regress by throwing where we didn't throw before. But maybe that's unlikely.  One option is to log the error but still ignore it. What do you think?

stIncMale (approved):

PR #675, opened 2021-03-01
Ensure retryable writes tests fail when unexpected exception is thrown

jyemin:
Also, ensure that the stack trace of the unexpected exception is
included in build output by re-throwing the exception.

JAVA-4025

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableWritesTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableWritesTest.java
@@ -138,14 +138,14 @@ public abstract class AbstractRetryableWritesTest {
 
         BsonDocument operation = definition.getDocument("operation");
         BsonDocument outcome = definition.getDocument("outcome");
-
+        boolean isErrorExpected = outcome.getBoolean("error", BsonBoolean.FALSE).getValue();
+        
         try {
             BsonDocument result = helper.getOperationResults(operation);
-            assertFalse("Expected error but instead got result: " + result.toJson(),
-                    outcome.getBoolean("error", BsonBoolean.FALSE).getValue());
+            assertFalse("Expected error but instead got result: " + result.toJson(), isErrorExpected);
             assertEquals(outcome.getDocument("result", new BsonDocument()), result.getDocument("result", new BsonDocument()));
         } catch (RuntimeException e) {
-            if (!outcome.getBoolean("error", BsonBoolean.FALSE).getValue()) {
+            if (!isErrorExpected) {
                 throw e;
             }
             assertExceptionState(e, outcome.get("result", new BsonNull()), operation.getString("name").getValue());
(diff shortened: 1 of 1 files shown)


driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableWritesTest.java
```diff
@@ -139,22 +139,18 @@ public void shouldPassAllOutcomes() {
...
         try {
-            result = helper.getOperationResults(operation);
+            BsonDocument result = helper.getOperationResults(operation);
+            assertFalse("Expected error but instead got result: " + result.toJson(),
+                    outcome.getBoolean("error", BsonBoolean.FALSE).getValue());
+            assertEquals(outcome.getDocument("result", new BsonDocument()), result.getDocument("result", new BsonDocument()));
         } catch (RuntimeException e) {
-            wasException = true;
+            if (!outcome.getBoolean("error", BsonBoolean.FALSE).getValue()) {
+                throw e;
```
jyemin, line 149: Ensures that we see the stack trace from the unexpected exception.

driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableWritesTest.java
```diff
@@ -139,22 +139,18 @@ public void shouldPassAllOutcomes() {
...
-        boolean wasException = false;
         try {
-            result = helper.getOperationResults(operation);
+            BsonDocument result = helper.getOperationResults(operation);
+            assertFalse("Expected error but instead got result: " + result.toJson(),
+                    outcome.getBoolean("error", BsonBoolean.FALSE).getValue());
+            assertEquals(outcome.getDocument("result", new BsonDocument()), result.getDocument("result", new BsonDocument()));
         } catch (RuntimeException e) {
-            wasException = true;
+            if (!outcome.getBoolean("error", BsonBoolean.FALSE).getValue()) {
```
stIncMale, line 148: [just an expression of my opinion]

While I know the reason behind duplicating the code `outcome.getBoolean("error", BsonBoolean.FALSE).getValue()` instead of introducing a local (`final`) variable, I strongly disagree that it is beneficial to the code in this particular case (arguing  position about this approach in general is harder, so I am not doing it now). I had to read and understand what this line of code does before I realized that it is the same code as on the line above. In other words, I as a reader did the same work twice.

stIncMale (approved):

driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableWritesTest.java
jyemin: I think part of the problem here is that the expression is not so easy to parse.  You have the `BsonBoolean.FALSE` default value, and then also the `getValue()` call tacked on at the end.  So maybe the issue is more with the API that we're calling.  But regardless I introduced a local variable as you suggest.

stIncMale: Thanks!

> problem here is that the expression is not so easy to parse

Absolutely. If it were `isErrorExpected(outcome)` instead of `outcome.getBoolean("error", BsonBoolean.FALSE).getValue()`, the code would be easily readable.

stIncMale (approved):

stIncMale (approved):

rozza (approved):

PR #676, opened 2021-03-01
Ensure that AsyncReadWriteBinding is released prior to subscriber notification

jyemin:
Previously the driver assumed that Mono#doOnTerminate is executed
prior to the subscriber being notified of completion.  But it turns
out that behavior is not guaranteed in the Californium release of
Project Reactor (though it is in later releases).  So instead,
now the SingleResultCallback that is used to notify the Mongo of
completion is wrapped by one that first releases the binding, and
Mono#doOnTerminate is no longer used.

JAVA-4027

### DIFF
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
@@ -66,12 +66,13 @@ public class OperationExecutorImpl implements OperationExecutor {
                         binding.release();
                         return Mono.error(new MongoClientException("Read preference in a transaction must be primary"));
                     } else {
-                        return Mono.<T>create(sink -> operation.executeAsync(binding, sinkToCallback(sink)))
-                                .doOnTerminate(binding::release)
-                                .doOnError((t) -> {
-                                    labelException(session, t);
-                                    unpinServerAddressOnTransientTransactionError(session, t);
-                                });
+                        return Mono.<T>create(sink -> operation.executeAsync(binding, (result, t) -> {
+                            binding.release();
+                            sinkToCallback(sink).onResult(result, t);
+                        })).doOnError((t) -> {
+                            labelException(session, t);
+                            unpinServerAddressOnTransientTransactionError(session, t);
+                        });
                     }
                 });
     }
@@ -86,13 +87,13 @@ public class OperationExecutorImpl implements OperationExecutor {
                                                               session == null && clientSession != null))
                 .switchIfEmpty(Mono.fromCallable(() -> getReadWriteBinding(ReadPreference.primary(), readConcern, session, false)))
                 .flatMap(binding ->
-                    Mono.<T>create(sink -> operation.executeAsync(binding, sinkToCallback(sink)))
-                                .doOnTerminate(binding::release)
-                                .doOnError((t) -> {
-                                    labelException(session, t);
-                                    unpinServerAddressOnTransientTransactionError(session, t);
-                                })
-
+                        Mono.<T>create(sink -> operation.executeAsync(binding, (result, t) -> {
+                            binding.release();
+                            sinkToCallback(sink).onResult(result, t);
+                        })).doOnError((t) -> {
+                            labelException(session, t);
+                            unpinServerAddressOnTransientTransactionError(session, t);
+                        })
                 );
     }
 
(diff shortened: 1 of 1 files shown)


driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
```diff
@@ -66,12 +66,13 @@
...
                     } else {
-                        return Mono.<T>create(sink -> operation.executeAsync(binding, sinkToCallback(sink)))
-                                .doOnTerminate(binding::release)
-                                .doOnError((t) -> {
-                                    labelException(session, t);
-                                    unpinServerAddressOnTransientTransactionError(session, t);
-                                });
+                        return Mono.<T>create(sink -> operation.executeAsync(binding, (result, t) -> {
+                            binding.release();
+                            sinkToCallback(sink).onResult(result, t);
```
jyemin, line 71: Does anyone else find `syncToCallback` confusing?  It seems more like the reverse, `callbackToSink`, makes more sense.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
```diff
@@ -86,13 +87,13 @@
...
                 .switchIfEmpty(Mono.fromCallable(() -> getReadWriteBinding(ReadPreference.primary(), readConcern, session, false)))
                 .flatMap(binding ->
-                    Mono.<T>create(sink -> operation.executeAsync(binding, sinkToCallback(sink)))
-                                .doOnTerminate(binding::release)
-                                .doOnError((t) -> {
-                                    labelException(session, t);
-                                    unpinServerAddressOnTransientTransactionError(session, t);
-                                })
-
+                        Mono.<T>create(sink -> operation.executeAsync(binding, (result, t) -> {
```
jyemin, line 90: I don't know what's going on with the indentation settings here.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
rozza: It converts the sink to a SingleResultCallback so makes sense to me.

rozza (approved):
LGTM - did this cause any tests to fail?

jyemin:
Yes, see the analysis in JAVA-3969.  It's the root cause of the failure of one of the sessions spec tests.

rozza:
Good detective work :)

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
stIncMale, line 71: Should we use try-finally in both places for releasing and calling `sinkToCallback`?

jyemin: like:
 
    try {
         sinkToCallback(sink).onResult(result, t);
    } finally {
        binding.release();
    }

If so, then no, because that would defeat the purpose of the change, which is to release the binding prior to the callback.  Maybe instead you mean:

    try {
         binding.release();
    } finally {
         sinkToCallback(sink).onResult(result, t);
    }

or 

    try {
         binding.release();
    } catch (RuntimeException e) }
         sinkToCallback(sink).onResult(null, e);    
    } finally {
         sinkToCallback(sink).onResult(result, t);
    }

?

stIncMale: I mean `try {release} finally { sinkToCallback.onResult }`. I am sorry if my first comment appears to propose a different order, I did not think it is.

I am not opposed to the simplified approach because passing an exception from `release` to `onResult` is really verbose:
```
(result, t) -> {
    Throwable primaryProblem = t;//if t is not null, then this is the exception a user wants to see as the primary one, not the one from the release method
    try {
        binding.release();
    } catch (RuntimeException e) {
        if (primaryProblem == null) {
            primaryProblem = e;
        } else {
            primaryProblem.addSuppressed(e);
        }
    } finally {
        sinkToCallback(sink).onResult(result, primaryProblem);
    }
}
```

Whichever you chose is fine by me. We also have an alternative:
1.  specify that `release` must not throw anything but `IllegalStateException if the reference count is already 0`;
2. ensure that all the current implementations comply with 1;
3. not use try-finally with `release`.

stIncMale: Actually, it's even worse:

1. On one hand, we actually should do
```java
finally {
    sinkToCallback(sink).onResult(primaryProblem == null ? result : null, primaryProblem);
}
```
2. On the other hand, doing 1 loses the result that was produced, which does not seem like a generally correct approach.

Given this, I am even more inclined to use the simplified version
```
try {
     binding.release();
} finally {
     sinkToCallback(sink).onResult(result, t);
}
```

stIncMale (approved):

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
jyemin: Done

stIncMale (approved):

PR #677, opened 2021-03-02
Re-enable AWS auth tests

jyemin:
JAVA-3906

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -353,6 +353,7 @@ functions:
         script: |
           ${PREPARE_SHELL}
           cd ${DRIVERS_TOOLS}/.evergreen/auth_aws
+          . ./activate_venv.sh
           mongo aws_e2e_regular_aws.js
     - command: shell.exec
       type: test
@@ -376,6 +377,7 @@ functions:
         script: |
           ${PREPARE_SHELL}
           cd ${DRIVERS_TOOLS}/.evergreen/auth_aws
+          . ./activate_venv.sh
           mongo aws_e2e_assume_role.js
     - command: shell.exec
       type: test
@@ -404,6 +406,7 @@ functions:
         script: |
           ${PREPARE_SHELL}
           cd ${DRIVERS_TOOLS}/.evergreen/auth_aws
+          . ./activate_venv.sh
           mongo aws_e2e_ec2.js
     - command: shell.exec
       type: test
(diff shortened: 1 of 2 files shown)


driver-core/src/test/functional/com/mongodb/internal/connection/AwsAuthenticationSpecification.groovy
```diff
@@ -31,7 +31,7 @@ class AwsAuthenticationSpecification extends Specification {
 
         when:
         openConnection(connection, async)
-        executeCommand(getConnectionString().getDatabase(), new BsonDocument('count', new BsonString('test')), connection)
+        executeCommand(getConnectionString().getDatabase(), new BsonDocument('count', new BsonString('test')), null, connection)
```
jyemin, line 34: The API of executeCommand changed with versioned API support, but it was missed here because the test was being skipped at the time and also because Groovy.  Fixed by adding a null ServerAPI instance to the calls.

bazile-clyde: Ack.

bazile-clyde (approved):
LGTM

.evergreen/.evg.yml
```diff
@@ -353,6 +353,7 @@ functions:
         script: |
           ${PREPARE_SHELL}
           cd ${DRIVERS_TOOLS}/.evergreen/auth_aws
+          . ./activate_venv.sh
```
rozza, line 356: Why is this now required?

jyemin: I don't know, actually.  I copied it from pymongo as per @ShaneHarvey 's suggestion.  @bazile-clyde , do you know?

bazile-clyde: It's because of [DRIVERS-1462](https://jira.mongodb.org/browse/DRIVERS-1462). The auth testing scripts (written in python) depend on the boto3 library. The linked issue was caused by the Ubuntu 18.04 evergreen host using an old version of boto3. Rather than having every driver upgrade boto3 themselves (or potentially any other dependencies), `activate_venv.sh` abstracted this complexity away. The script sets up a virtual environment and upgrades boto3.

bazile-clyde: Hopefully, if we run into this same type of issue we can upgrade `activate_venv.sh` without causing driver-wide changes.

rozza: Sounds good thanks @bazile-clyde

rozza (approved):
LGTM

PR #680, opened 2021-03-09
Add getReason method to ConnectionCheckOutFailedEvent

jyemin:
JAVA-4038

Somehow this was missed in the initial commit.  I noticed it when doing the workload executor for planned maintenance.


rozza (approved):
LGTM

PR #682, opened 2021-03-11
Update error handling logic

jyemin:
Update logic for throwing MongoNotPrimaryException or MongoNodeIsRecoveringException:

* Add 10058 to the list of error codes that cause MongoNotPrimaryException
  to be thrown.
* Only examine error messages when no error code is included in the reply.

JAVA-3986


rozza (approved):
LGTM

PR #683, opened 2021-03-15
Document DuplicateKeyException and WriteConcernException are effectively deprecated

jyemin:
JAVA-3919

### DIFF
--- a/driver-core/src/main/com/mongodb/DuplicateKeyException.java
+++ b/driver-core/src/main/com/mongodb/DuplicateKeyException.java
@@ -19,15 +19,21 @@ package com.mongodb;
 import org.bson.BsonDocument;
 
 /**
- * A subclass of {@link WriteConcernException} representing a duplicate key exception
+ * The legacy {@link WriteConcernException}, thrown when trying to insert or update a document containing a duplicate key.
  *
- * <p>
- * This class is effectively deprecated. It is only thrown with use of the long-deprecated MongoClient.getDB method and associated
- * methods, so when using non-deprecated execution paths it should not be caught.  Instead, there are two other exceptions that can be
- * caught. For single-document inserts, a {@link MongoWriteException} is thrown with a {@link WriteError} in the category
- * {@link ErrorCategory#DUPLICATE_KEY}. For bulk inserts, a {@link MongoBulkWriteException} is thrown where one or more of the
- * {@link WriteError}'s in the list of errors is in the category {@link ErrorCategory#DUPLICATE_KEY}.
- * </p>
+ * <p>Only thrown when using the legacy deprecated API, which is accessed via {@code com.mongodb.MongoClient.getDB}.</p>
+ *
+ * <p>For application using the {@code MongoCollection}-based API, duplicate key exceptions can be determined via:</p>
+ * <ul>
+ *   <li>
+ *       Single document inserts or updates: a {@link MongoWriteException} is thrown with a {@link WriteError} in the category
+ *       {@link ErrorCategory#DUPLICATE_KEY}.
+ *   </li>
+ *   <li>
+ *       Bulk document inserts or updates: A {@link MongoBulkWriteException} is thrown where one or more of the {@link WriteError}'s in the
+ *       list of errors is in the category {@link ErrorCategory#DUPLICATE_KEY}.
+ *   </li>
+ * </ul>
  *
  * @since 2.12
  * @see MongoWriteException
--- a/driver-core/src/main/com/mongodb/DuplicateKeyException.java
+++ b/driver-core/src/main/com/mongodb/DuplicateKeyException.java
@@ -19,9 +19,21 @@ package com.mongodb;
 import org.bson.BsonDocument;
 
 /**
- * Subclass of {@link WriteConcernException} representing a duplicate key exception
+ * A subclass of {@link WriteConcernException} representing a duplicate key exception
+ *
+ * <p>
+ * This class is effectively deprecated. It is only thrown with use of the long-deprecated MongoClient.getDB method and associated
+ * methods, so when using non-deprecated execution paths it should not be caught.  Instead, there are two other exceptions that can be
+ * caught. For single-document inserts, a {@link MongoWriteException} is thrown with a {@link WriteError} in the category
+ * {@link ErrorCategory#DUPLICATE_KEY}. For bulk inserts, a {@link MongoBulkWriteException} is thrown where one or more of the
+ * {@link WriteError}'s in the list of errors is in the category {@link ErrorCategory#DUPLICATE_KEY}.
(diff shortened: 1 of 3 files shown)


driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java
```diff
@@ -63,13 +59,6 @@
     private static final Logger PROTOCOL_EVENT_LOGGER = Loggers.getLogger("protocol.event");
     private static final CodecRegistry REGISTRY = fromProviders(new BsonValueCodecProvider());
 
-    private static WriteConcernResult createWriteResult(final BsonDocument result) {
```
jyemin, line 66: I spotted this unused code when looking to see how these exceptions are used by the driver, and thought it should be removed so that it's clear that the exceptions are not actually used except where indicated in the new javadoc.

driver-core/src/main/com/mongodb/DuplicateKeyException.java
```diff
@@ -19,9 +19,21 @@
 import org.bson.BsonDocument;
 
 /**
- * Subclass of {@link WriteConcernException} representing a duplicate key exception
+ * A subclass of {@link WriteConcernException} representing a duplicate key exception
+ *
+ * <p>
+ * This class is effectively deprecated. It is only thrown with use of the long-deprecated MongoClient.getDB method and associated
```
stIncMale, line 25: I propose writing `{@code MongoClient.getDB}` (the same proposal is for `WriteConcernException.java`) since we cannot link `MongoClient.getDB` here.

driver-core/src/main/com/mongodb/DuplicateKeyException.java
```diff
@@ -19,9 +19,21 @@
 import org.bson.BsonDocument;
 
 /**
- * Subclass of {@link WriteConcernException} representing a duplicate key exception
+ * A subclass of {@link WriteConcernException} representing a duplicate key exception
```
stIncMale, line 22: Nit: end the sentence with a period?

driver-core/src/main/com/mongodb/DuplicateKeyException.java
stIncMale, line 25: What is the reason we don't want to use `@Deprecated` and `@deprecated` here and in `WriteConcernException.java`?

stIncMale (changes requested):

driver-core/src/main/com/mongodb/DuplicateKeyException.java
jyemin: Sure, though it's technically not a complete English sentence. :)

driver-core/src/main/com/mongodb/DuplicateKeyException.java
jyemin: Yeah, I debated that.  A little weird because `MongoClient.getDB` not actually valid code.

stIncMale: > not actually valid code

What do you mean by this and why is this relevant to using of the `@code` inline tag?


One more idea, since we cannot have a link here, it is useful to include the package name: `com.mongodb.MongoClient.getDB`.

jyemin: I thought `@code` was used for actual code snippets, and not function signatures.  Is that not the case?

jyemin: The Javadoc for `MongoClient#getDB` states:

>    The {@link DB} class has been superseded by {@link com.mongodb.client.MongoDatabase}.  The deprecation of this method effectively deprecates the {@link DB}, {@link DBCollection}, and {@link DBCursor} classes, among others; but in order to give users time to migrate to the new API without experiencing a huge number of compiler warnings, those classes have not yet been formally deprecated.

These two exception classes would be classified as part of the `among others` phrase in that documentation.

stIncMale: The [`@code`](https://docs.oracle.com/en/java/javase/15/docs/specs/javadoc/doc-comment-spec.html#code) inline tag _"Displays text in code font without interpreting the text as HTML markup or nested Javadoc tags"_. HTML [`code`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-code-element) element also gives a lot of room for using it _"a fragment of computer code. This could be an XML element name, a filename, a computer program, or any other string that a computer would recognize"_.

> actual code snippets

I wonder, how would one define a code snippet in a robust and practically useful way that is different from the one specified in the HTML spec?

stIncMale: I see, thanks you. For the record, I would like to express that I disagree with the approach of deprecating code elements in such a way that tools cannot spot the deprecation.

rozza (changes requested):
I don't think this is clear enough to remove the confusion the user reported in JAVA-3919.

How about something clearer like:

```
/**
 * The legacy {@link WriteConcernException} representing trying to insert a duplicate key.
 *
 * <p>Only thrown when using the legacy deprecated api, which is accessed via {@code mongoClient.getDB}.</p>
 *
 * <p>For users of the modern api, duplicate key exceptions can be determined via:</p>
 * <ul>
 *   <li>
 *       Single document inserts.
 *       A {@link MongoWriteException} is thrown with a {@link WriteError} in the category {@link ErrorCategory#DUPLICATE_KEY}.
 *   </li>
 *   <li>
 *       Bulk document inserts. 
 *       A {@link MongoBulkWriteException} is thrown where one or more of the {@link WriteError}'s in the list of errors is in the 
 *       category {@link ErrorCategory#DUPLICATE_KEY}.
 *   </li>
 * </ul>
 *
 * @since 2.12
 * @see MongoWriteException
 * @see MongoBulkWriteException
 * @see WriteError
 * @see ErrorCategory#DUPLICATE_KEY
 */
```

driver-core/src/main/com/mongodb/DuplicateKeyException.java
jyemin: Done

driver-core/src/main/com/mongodb/DuplicateKeyException.java
jyemin: Done

jyemin:
@rozza I update mostly as per your suggestion, though I modified it a bit because I realized that you can get a duplicate key error on an update as well as an insert, if you have a unique index on a field besides `_id`.  

PTAL

stIncMale (approved):

rozza (approved):
LGTM

jyemin:
Merging on command line

PR #684, opened 2021-03-15
Update workload executor to use unified test runner

jyemin:
JAVA-3952

Not really worth looking at the diff.  It's essentially a new implementation.

[This patch build](https://evergreen.mongodb.com/version/604902a23e8e865aca63dd8c) is actually what proves it works, since the workload executor doesn't run in the driver's own Evergreen configuration.

### DIFF
--- a/driver-workload-executor/build.gradle
+++ b/driver-workload-executor/build.gradle
@@ -44,7 +44,7 @@ dependencies {
     implementation project(':driver-core').sourceSets.test.output
     implementation project(':driver-sync').sourceSets.test.output
     implementation 'ch.qos.logback:logback-classic:1.1.1'
-    implementation(platform("org.junit:junit-bom:5.6.2")) // TODO: grab BOM version from parent
+    implementation(platform("org.junit:junit-bom:$junitBomVersion"))
     implementation('org.junit.jupiter:junit-jupiter')
     implementation('org.junit.vintage:junit-vintage-engine')
 }
--- a/driver-workload-executor/build.gradle
+++ b/driver-workload-executor/build.gradle
@@ -1,11 +1,11 @@
 /*
- * Copyright 2016-present MongoDB, Inc.
+ * Copyright 2008-present MongoDB, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *   http://www.apache.org/licenses/LICENSE-2.0
+ *    http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -14,13 +14,19 @@
  * limitations under the License.
  */
 
+buildscript {
+    repositories {
+        maven { url "https://plugins.gradle.org/m2/" }
+    }
+    dependencies {
+        classpath 'com.github.jengelman.gradle.plugins:shadow:6.1.0'
+    }
+}
+
 apply plugin: 'application'
+apply plugin: 'com.github.johnrengelman.shadow'
 
 mainClassName = "com.mongodb.workload.WorkloadExecutor"
-
-run {
-    systemProperties System.getProperties()
-}
 
 sourceSets {
     main {
@@ -34,13 +40,21 @@ sourceSets {
 }
 
 dependencies {
-    api project(':driver-sync')
-
+    implementation project(':driver-sync')
+    implementation project(':driver-core').sourceSets.test.output
     implementation project(':driver-sync').sourceSets.test.output
-
     implementation 'ch.qos.logback:logback-classic:1.1.1'
+    implementation(platform("org.junit:junit-bom:5.6.2")) // TODO: grab BOM version from parent
+    implementation('org.junit.jupiter:junit-jupiter')
+    implementation('org.junit.vintage:junit-vintage-engine')
 }
 
 javadoc {
     enabled = false
+}
+
+jar {
+    manifest {
+        attributes "Main-Class": "com.mongodb.workload.WorkloadExecutor"
+    }
 }
--- a/driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
+++ b/driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
@@ -16,110 +16,149 @@
 
 package com.mongodb.workload;
 
-import com.mongodb.client.JsonPoweredCrudTestHelper;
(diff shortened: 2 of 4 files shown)


jyemin:
[This patch build](https://evergreen.mongodb.com/version/604902a23e8e865aca63dd8c) is actually what proves it works, since the workload executor doesn't run in the driver's own Evergreen configuration.

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -16,110 +16,149 @@
...
-                        numFailures.incrementAndGet();
-                    } else {
-                        numSuccesses.incrementAndGet();
-                    }
-                } catch (Exception e) {
-                    numErrors.incrementAndGet();
-                    e.printStackTrace();
-                }
+
+        System.out.println(fileDocument.toJson(JsonWriterSettings.builder().indent(true).build()));
```
rozza, line 79: Can we remove these printlns?

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -16,110 +16,149 @@
...
-                .append("numErrors", new BsonInt32(numErrors.intValue()));
-
-        try (BufferedWriter writer = new BufferedWriter(
-                new OutputStreamWriter(new FileOutputStream(new File(outputDirectory, "results.json"))))) {
-            writer.write(resultsDocument.toJson(JsonWriterSettings.builder().indent(true).build()));
-            writer.newLine();
-            writer.flush();
-        }
+    private static void writeFile(final BsonDocument document, final Path path) throws IOException {
+        System.out.println("Writing file: '" + path.toFile().getAbsolutePath());
```
rozza, line 159: Another println.

driver-workload-executor/build.gradle
```diff
@@ -34,13 +40,21 @@ sourceSets {
...
 
 dependencies {
-    api project(':driver-sync')
-
+    implementation project(':driver-sync')
+    implementation project(':driver-core').sourceSets.test.output
     implementation project(':driver-sync').sourceSets.test.output
-
     implementation 'ch.qos.logback:logback-classic:1.1.1'
+    implementation(platform("org.junit:junit-bom:5.6.2")) // TODO: grab BOM version from parent
```
rozza, line 47: Is this todo still needed?

rozza (changes requested):
Looks good.

Two nits: 
* Remove the println statements and / or use logging.  
* Todo in the gradle file - either fix or remove the todo

driver-workload-executor/build.gradle
jyemin: Ah, I meant to comment to you about this.  Do you know how we can share the BOM version between here and the main build.gradle?

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
jyemin: The println's are useful to understand what Astrolabe is asking the workload executor to do.  Since it's not production code, I don't think it matters whether we use a logger or println, as we only care that these show up when running the executor, and we don't mind that they always show up (we don't need runtime control as enabled by a logger).  So I think it's ok to leave as is, but let me know if you think there are other reasons to prefer a logger.

rozza: I still think we should follow the same coding conventions for all our projects.

driver-workload-executor/build.gradle
rozza: Gotcha, check out how projectReactorVersion is set in the top level project and used in the reactive streams project.  That pattern should work here also.

rozza: `junitBomVersion` needs to be added to the `ext` block and then it should be usable across the sub projects.

jyemin: Thanks, done.

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
jyemin: Done

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
jyemin: Done

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
```diff
@@ -16,110 +16,152 @@
...
-        String outputDirectory =
-                System.getProperty("org.mongodb.workload.output.directory", System.getProperty("user.dir"));
-        
+    private static final Logger LOGGER = Loggers.getLogger("workload-executor");
+    private static final CountDownLatch terminationLatch = new CountDownLatch(1);
+    private static volatile boolean terminateLoop;
+
+    public static void main(String[] args) throws IOException {
+        if (args.length != 2) {
+            System.out.println("Usage: AstrolabeTestRunner <path to workload spec JSON file> <path to results directory>");
```
jyemin, line 50: @rozza I left this on as a `println` as it seemed weird to use a logger to report incorrect usage.

jyemin:
@rozza ready for another look

driver-workload-executor/build.gradle
```diff
@@ -14,14 +14,20 @@
...
+buildscript {
+    repositories {
+        maven { url "https://plugins.gradle.org/m2/" }
+    }
+    dependencies {
+        classpath 'com.github.jengelman.gradle.plugins:shadow:6.1.0'
+    }
+}
+
 apply plugin: 'application'
```
rozza, line 26: Nit: doesnt look like you use this plugin.

rozza (approved):
One nit - but LGTM.

driver-workload-executor/src/main/com/mongodb/workload/WorkloadExecutor.java
rozza: Ack

driver-workload-executor/build.gradle
jyemin: Yeah I left it here so that there is still a way to run it directly from gradle without first creating a shadow jar, just for debugging purposes.  Let me know if you think it should go.

rozza: Ah nice I was expecting a configuration block for it

eg: 
```
application {
    mainClass.set("org.gradle.sample.Main")
}
```

But sounds like its not needed - LGTM

PR #686, opened 2021-03-17
Join the thread in DocumentationSamples.testWatch

stIncMale:
JAVA-4075


stIncMale:
I described the problem in detail in JAVA-4075, and I also pointed out there that I fail to explain the test failure. I do not want to fallback to "maybe there is a problem on the server-side" explanation, so most likely my understanding of what is happening is not correct. I would be glad if I could discuss the problem with someone and understand the situation better.

jyemin (approved):

PR #689, opened 2021-03-23
Guarantee that ChangeStreamPublisher for a collection completes after dropping the collection

stIncMale:
JAVA-4044

### DIFF
--- a/driver-core/src/main/com/mongodb/assertions/Assertions.java
+++ b/driver-core/src/main/com/mongodb/assertions/Assertions.java
@@ -18,11 +18,18 @@
 package com.mongodb.assertions;
 
 import com.mongodb.internal.async.SingleResultCallback;
+import com.mongodb.lang.Nullable;
 
 import java.util.Collection;
 
 /**
  * <p>Design by contract assertions.</p> <p>This class is not part of the public API and may be removed or changed at any time.</p>
+ * All {@code assert...} methods throw {@link AssertionError} and should be used to check conditions which may be violated if and only if
+ * the driver code is incorrect. The intended usage of this methods is the same as of the
+ * <a href="https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html">Java {@code assert} statement</a>. The reason
+ * for not using the {@code assert} statements is that they are not always enabled. We prefer having internal checks always done at the
+ * cost of our code doing a relatively small amount of additional work in production.
+ * The {@code assert...} methods return values to open possibilities of being used fluently.
  */
 public final class Assertions {
     /**
@@ -117,6 +124,57 @@ public final class Assertions {
                 throw new IllegalArgumentException(name + " can not contain a null value");
             }
         }
+    }
+
+    /**
+     * @param value A value to check.
+     * @param <T> The type of {@code value}.
+     * @return {@code null}.
+     * @throws AssertionError If {@code value} is not {@code null}.
+     */
+    @Nullable
+    public static <T> T assertNull(@Nullable final T value) throws AssertionError {
+        if (value != null) {
+            throw new AssertionError(value.toString());
+        }
+        return null;
+    }
+
+    /**
+     * @param value A value to check.
+     * @param <T> The type of {@code value}.
+     * @return {@code value}
+     * @throws AssertionError If {@code value} is {@code null}.
+     */
+    public static <T> T assertNotNull(@Nullable final T value) throws AssertionError {
+        if (value == null) {
+            throw new AssertionError();
+        }
+        return value;
+    }
+
+    /**
+     * @param value A value to check.
+     * @return {@code true}.
+     * @throws AssertionError If {@code value} is {@code false}.
+     */
+    public static boolean assertTrue(final boolean value) throws AssertionError {
+        if (!value) {
+            throw new AssertionError();
+        }
+        return true;
+    }
+
+    /**
+     * @param value A value to check.
+     * @return {@code false}.
+     * @throws AssertionError If {@code value} is {@code true}.
+     */
+    public static boolean assertFalse(final boolean value) throws AssertionError {
+        if (value) {
+            throw new AssertionError();
+        }
+        return false;
     }
 
     private Assertions() {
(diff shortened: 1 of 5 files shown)


driver-core/src/main/com/mongodb/assertions/Assertions.java
```diff
@@ -119,6 +126,57 @@ public static void doesNotContainNull(final String name, final Collection<?> col
...
     }
 
+    /**
+     * @param value A value to check.
+     * @param <T> The type of {@code value}.
+     * @return {@code null}.
+     * @throws AssertionError If {@code value} is not {@code null}.
+     */
+    @Nullable
+    public static <T> T assertNull(@Nullable final T value) throws AssertionError {
```
rozza, line 136: None of the other methods throw checked exceptions. So these follow the existing convention?

rozza, line 136: None of the other methods throw checked exceptions. Should these follow the existing convention?

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
```diff
@@ -44,7 +48,12 @@
     private final int maxWireVersion;
 
     private volatile BsonDocument resumeToken;
-    private volatile AsyncAggregateResponseBatchCursor<RawBsonDocument> wrapped;
+    /**
+     * {@code null} {@linkplain AtomicReference#get() value} combined with {@link #isClosed} being {@code false}
+     * represents a situation in which the wrapped object was closed by {@code this} and {@code this} stayed open.
+     * This happens as a result of {@linkplain ChangeStreamBatchCursorHelper#isRetryableError(Throwable, int) retryable errors}.
```
rozza, line 54: TIL about labels for linkplain

rozza: I found the text hard to understand. Perhaps simplify to something like:

```
  /**
   * Note: {@linkplain ChangeStreamBatchCursorHelper#isRetryableError(Throwable, int) Retryable errors} can 
   * result in {@code wrapped} containing a {@code null} value and {@link #isClosed} being {@code false}. 
   * This represents a situation in which the wrapped object was closed by {@code this} but {@code this} remained open.
```

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
```diff
@@ -102,22 +112,63 @@ public void close() {
...
+            observedWrapped.close();
+        }
+    }
+
+    /**
+     * This method guarantees that the {@code wrapped} argument is closed even if
+     * {@link #setWrappedOrCloseIt(AsyncAggregateResponseBatchCursor)} is called concurrently with or after (in the happens-before order)
+     * the method {@link #close()}.
+     */
+    private void setWrappedOrCloseIt(final AsyncAggregateResponseBatchCursor<RawBsonDocument> wrapped) {
```
rozza, line 157: Nit: Use alternative variable name for the replacement wrapped. The differences between `wrapped` and `this.wrapped` can be made more explicit.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
```diff
@@ -282,4 +312,31 @@ private void assertCommandNames(final List<String> commandNames) {
...
+     * This method ensures that the server considers the specified {@code collection} existing for the purposes of, e.g.,
+     * {@link MongoCollection#drop()}, instead of replying with
+     * {@code "errmsg": "ns not found", "code": 26, "codeName": "NamespaceNotFound"}.
+     *
+     * @return {@code operationTime} starting at which the {@code collection} is guaranteed to exist.
+     */
+    private static BsonTimestamp ensureExists(final MongoClient client, final MongoCollection<Document> collection) {
+        BsonObjectId insertedId = Mono.from(collection.insertOne(Document.parse("{}")))
+                .map(InsertOneResult::getInsertedId)
+                .map(BsonValue::asObjectId)
```
rozza, line 325: Observation: No need to cast to object id - an `_id` field can be of any BsonValue type.

rozza (approved):
Only a  few nits / comments / suggestions for you to decide upon.  

Otherwise LGTM.

driver-core/src/main/com/mongodb/assertions/Assertions.java
jyemin: Do you mean unchecked?  I would agree that there is no need for the throws clause but it also doesn't bother me given that throwing this exception is the whole purpose of the method.

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
jyemin: I had the same momenentary confusion when reading.

jyemin (approved):
LGTM

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
stIncMale: Thanks, removed the casting.

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
stIncMale: Renamed the parameter to `newValue`.

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
stIncMale: Ross, thank you for providing the proposed text. Used it as suggested.

driver-core/src/main/com/mongodb/assertions/Assertions.java
stIncMale: I use the `throws` clause to document thrown exceptions that are part of a method contract (even if they are unchecked) in situations when I don't write javadocs. But then when I write javadocs, I still specify the `throws` clause (to have the same approach everywhere) even though it is duplicated with the `@throws` tag. Since there does not seem to be a strong push for removing the clause, I left it.

PR #692, opened 2021-04-09
Add alternate constructor to Polygon and PolygonCoordinates

markaalvaro:
Adds alternate constructors to Polygon and PolygonCoordinates that make it possible to construct a Mongo Polygon given a Polygon from some other library with a similar API.

Posting this PR to demonstrate what it would be nice to have, but this was a breaking change if consuming the driver using Groovy as can be seen in a few tests I changed. I would assume this exact solution won't work, so open to any ideas.

Basically just want to be able to convert a Polygon with this type of API to the Java Driver's Polygon:

```
class Coordinate {
    double[] getCoords();
}

class Polygon {
    List<Coordinate> getExterior();
    List<List<Coordinate>> getHoles();
}
```

which currently is not possible as converting the above Polygon holes to the Java Driver's would require instantiating and populating a `List<Position>[]`, which using conventional Java isn't allowed because of the compile error: "Cannot create a generic array of List<Position>". And I know using `newInstance()` I can make an array that will work, but just looking for some way to avoid using reflection if possible.

JAVA-4098

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/geojson/Polygon.java
+++ b/driver-core/src/main/com/mongodb/client/model/geojson/Polygon.java
@@ -45,6 +45,16 @@ public final class Polygon extends Geometry {
     /**
      * Construct an instance with the given coordinates.
      *
+     * @param exterior the exterior ring of the polygon
+     * @param holes    optional interior rings of the polygon
+     */
+    public Polygon(final List<Position> exterior, final List<List<Position>> holes) {
+        this(new PolygonCoordinates(exterior, holes));
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


markaalvaro:
[JAVA-4098](https://jira.mongodb.org/browse/JAVA-4098)

jyemin:
Added a small commit to address the compiler warning, as well as `@since` Javadoc annotations for the new constructors.

Despite the Groovy breakage, I think this is ok.  It preserves binary compatibility, even if it does not source compatibility.

jyemin (approved):

stIncMale (approved):

PR #694, opened 2021-04-15
Support TLS in AbstractConnectionPoolTest and its subclasses

stIncMale:
Also log exceptions that happen while tests run

JAVA-4104


jyemin (approved):
LGTM

PR #695, opened 2021-04-17
Unpin server from session on abort or first non-transactional operation 

jyemin:
This change is a required precursor for load balancing support.  It fixes a couple of bugs in the pinning behavior of the driver for transactions that were uncovered by the new tests.

1. Server should be unpinned on successful abort of txn
2. Server should be unpinned on the first operation in a session after transaction has been committed 
 
JAVA-4096

### DIFF
--- a/driver-core/src/test/resources/unified-test-format/transactions/java-only-mongos-unpin.json
+++ b/driver-core/src/test/resources/unified-test-format/transactions/java-only-mongos-unpin.json
@@ -1,0 +1,96 @@
+{
+  "description": "java-only-mongos-unpin",
+  "schemaVersion": "1.1",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "4.2",
+      "topologies": [
+        "sharded-replicaset"
+      ]
+    }
+  ],
+  "createEntities": [
+    {
+      "client": {
+        "id": "client0",
+        "useMultipleMongoses": true
+      }
+    },
+    {
+      "database": {
+        "id": "database0",
+        "client": "client0",
+        "databaseName": "mongos-unpin-db"
+      }
+    },
+    {
+      "collection": {
+        "id": "collection0",
+        "database": "database0",
+        "collectionName": "test"
+      }
+    },
+    {
+      "session": {
+        "id": "session0",
+        "client": "client0"
+      }
+    }
+  ],
+  "initialData": [
+    {
+      "collectionName": "test",
+      "databaseName": "mongos-unpin-db",
+      "documents": []
+    }
+  ],
+  "tests": [
+    {
+      "description": "unpin when a non-transaction read operation uses a session",
+      "operations": [
+        {
+          "name": "startTransaction",
+          "object": "session0"
+        },
+        {
+          "name": "insertOne",
+          "object": "collection0",
+          "arguments": {
+            "document": {
+              "x": 1
+            },
+            "session": "session0"
+          }
+        },
+        {
+          "name": "commitTransaction",
+          "object": "session0"
+        },
+        {
+          "name": "assertSessionPinned",
+          "object": "testRunner",
+          "arguments": {
+            "session": "session0"
+          }
+        },
+        {
+          "name": "find",
+          "object": "collection0",
+          "arguments": {
+            "filter": {
+              "x": 1
+            },
+            "session": "session0"
+          }
+        },
+        {
+          "name": "assertSessionUnpinned",
+          "object": "testRunner",
+          "arguments": {
+            "session": "session0"
+          }
+        }
+      ]
+    }
+  ]
+}
--- a/driver-core/src/test/resources/unified-test-format/transactions/mongos-unpin.json
+++ b/driver-core/src/test/resources/unified-test-format/transactions/mongos-unpin.json
@@ -1,0 +1,277 @@
+{
+  "description": "mongos-unpin",
+  "schemaVersion": "1.1",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "4.2",
+      "topologies": [
+        "sharded-replicaset"
+      ]
+    }
+  ],
+  "createEntities": [
+    {
+      "client": {
+        "id": "client0",
+        "useMultipleMongoses": true
+      }
+    },
+    {
+      "database": {
+        "id": "database0",
+        "client": "client0",
+        "databaseName": "mongos-unpin-db"
+      }
+    },
+    {
+      "collection": {
+        "id": "collection0",
+        "database": "database0",
+        "collectionName": "test"
+      }
+    },
+    {
+      "session": {
+        "id": "session0",
+        "client": "client0"
+      }
+    }
+  ],
+  "initialData": [
+    {
+      "collectionName": "test",
+      "databaseName": "mongos-unpin-db",
+      "documents": []
+    }
+  ],
+  "_yamlAnchors": {
+    "anchors": 24
+  },
+  "tests": [
+    {
+      "description": "unpin after TransientTransctionError error on commit",
+      "operations": [
+        {
+          "name": "startTransaction",
+          "object": "session0"
+        },
+        {
+          "name": "insertOne",
+          "object": "collection0",
+          "arguments": {
+            "document": {
+              "x": 1
+            },
+            "session": "session0"
+          }
+        },
+        {
+          "name": "targetedFailPoint",
+          "object": "testRunner",
+          "arguments": {
+            "session": "session0",
+            "failPoint": {
+              "configureFailPoint": "failCommand",
+              "mode": {
+                "times": 1
+              },
+              "data": {
+                "failCommands": [
+                  "commitTransaction"
+                ],
+                "errorCode": 24
+              }
+            }
+          }
+        },
+        {
+          "name": "commitTransaction",
+          "object": "session0",
+          "expectError": {
+            "errorCode": 24,
+            "errorLabelsContain": [
+              "TransientTransactionError"
+            ],
+            "errorLabelsOmit": [
+              "UnknownTransactionCommitResult"
+            ]
+          }
+        },
+        {
+          "name": "assertSessionUnpinned",
+          "object": "testRunner",
+          "arguments": {
+            "session": "session0"
+          }
+        }
+      ]
+    },
+    {
+      "description": "unpin on successful abort",
+      "operations": [
+        {
+          "name": "startTransaction",
+          "object": "session0"
+        },
+        {
+          "name": "insertOne",
+          "object": "collection0",
+          "arguments": {
+            "document": {
+              "x": 1
+            },
+            "session": "session0"
+          }
+        },
+        {
+          "name": "abortTransaction",
+          "object": "session0"
+        },
+        {
+          "name": "assertSessionUnpinned",
+          "object": "testRunner",
+          "arguments": {
+            "session": "session0"
+          }
+        }
+      ]
+    },
+    {
+      "description": "unpin after TransientTransctionError error on abort",
+      "operations": [
+        {
+          "name": "startTransaction",
+          "object": "session0"
+        },
+        {
+          "name": "insertOne",
+          "object": "collection0",
+          "arguments": {
+            "document": {
+              "x": 1
+            },
+            "session": "session0"
+          }
+        },
+        {
+          "name": "targetedFailPoint",
+          "object": "testRunner",
+          "arguments": {
+            "session": "session0",
+            "failPoint": {
+              "configureFailPoint": "failCommand",
+              "mode": {
+                "times": 1
+              },
+              "data": {
+                "failCommands": [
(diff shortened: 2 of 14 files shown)


driver-core/src/test/resources/unified-test-format/transactions/java-only-mongos-unpin.json
```diff
@@ -0,0 +1,96 @@
+{
```
jyemin, line 1: This may be controversial: The spec test doesn't cover all the paths in the Java driver to test this unpinning logic.  That's because there are two paths for operation execution, one for reads and one for writes.  The spec test exercises the write path.  I added this test to execute the read path as well.

driver-core/src/test/resources/unified-test-format/transactions/mongos-unpin.json
```diff
@@ -0,0 +1,277 @@
+{
```
jyemin, line 1: This is the test from the specs repo.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ClientSessionPublisherImpl.java
```diff
@@ -186,9 +193,11 @@ public AsyncClientSession getWrapped() {
                     new AbortTransactionOperation(transactionOptions.getWriteConcern())
                             .recoveryToken(getRecoveryToken()),
                     readConcern, this)
-                    .doOnError(MongoException.class, this::unpinServerAddressOnError)
                     .onErrorResume(Throwable.class, (e) -> Mono.empty())
-                    .doOnTerminate(() ->  cleanupTransaction(TransactionState.ABORTED));
+                    .doOnTerminate(() -> {
+                        setPinnedServerAddress(null);
```
jyemin, line 198: The spec requires that abort unpin on success as well as error, but there were no tests for it.  Now there are, and Java driver fails the test.  This fixes it for reactive

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
```diff
@@ -85,6 +91,11 @@
             @Nullable final ClientSession session) {
         notNull("operation", operation);
         notNull("readConcern", readConcern);
+
+        if (session != null && !(operation instanceof CommitTransactionOperation)) {
```
jyemin, line 95: It's ugly, but I couldn't find a better way to implement the required behavior.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/UnifiedTransactionsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient;
+import org.bson.BsonArray;
+import org.bson.BsonDocument;
+import org.junit.runners.Parameterized;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.Collection;
+
+public class UnifiedTransactionsTest extends UnifiedTest {
```
jyemin, line 32: These are the first transaction tests in the unified format, so had to add a new test class

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -135,10 +135,14 @@ OperationResult executeFind(final BsonDocument operation) {
         MongoCollection<BsonDocument> collection = entities.getCollection(operation.getString("object").getValue());
         BsonDocument arguments = operation.getDocument("arguments");
 
+        ClientSession session = getSession(arguments);
         BsonDocument filter = arguments.getDocument("filter");
-        FindIterable<BsonDocument> iterable = collection.find(filter);
+        FindIterable<BsonDocument> iterable = session == null
```
jyemin, line 140: Needed this change for one of the tests that uses sessions with find.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ClientSessionPublisherImpl.java
```diff
@@ -74,6 +74,13 @@ public boolean notifyMessageSent() {
         }
     }
 
+    @Override
+    public void notifyNonCommitOperationInitiated() {
+        if (transactionState == TransactionState.COMMITTED) {
```
stIncMale, line 79: I understand that we cannot rely solely on `transactionState` and need additional knowledge that the operation is not a `CommitTransactionOperation` because commit may be called multiple times. However, I do not understand why we check for `transactionState == TransactionState.COMMITTED` instead of checking for `transactionState != TransactionState.IN` to determine whether an operation is executed in a transaction. @jyemin, could you please clarify this?

driver-core/src/test/resources/unified-test-format/transactions/java-only-mongos-unpin.json
rozza: Agreed, lets make it less controversial and add it to repo. Other drivers may benefit from it also.

rozza (commented):
Looks good agree with @stIncMale about the test and interested in the `notifyNonCommitOperationInitiated` conversation.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ClientSessionPublisherImpl.java
jyemin: It's because if state is IN or ABORTED, no server will be pinned, so it would be a no-op.  I thought this way would make the intent clearer.  I see three choices

1. Leave it as is
2. Add an assertion that nothing is pinned if state is ABORTED or NONE (there should already be tests that assert this)
3. Change conditional to `transactionState != IN`

What's the preference?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/ClientSession.java
```diff
@@ -46,6 +46,14 @@
      */
     boolean notifyMessageSent();
 
+    /**
+     * Notify the client session that command execution is being initiated. This should be called before server selection occurs.
+     * <p>
+     * For internal use only
```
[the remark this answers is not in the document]
jyemin: Ack

jyemin:
@divjotarora any objection to adding this test to the specs repo?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ClientSessionPublisherImpl.java
rozza: Thanks @jyemin makes sense, I'm happy for it to be kept as is. I think part of the confusion is the name of the method, versus what it does. Could we make that clearer?

jyemin: I was trying to keep it in the spirit of the existing `notifyMessageSent`.  What do you suggest as an alternative name?

rozza: `unPinServerIfNotCommitted` ?

rozza: Happy for it to be left as well.

rozza (approved):

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ClientSessionPublisherImpl.java
stIncMale: > keep it in the spirit of the existing notifyMessageSent

It seems best to have a method `notifyOperationInitiated` with all the logic that decides which operations to react and how in the method. Other approaches with necessity move pieces of this logic out of the method to the surrounding code, and I do not see a good reason to break the encapsulation this way.

> Add an assertion that nothing is pinned if state is ABORTED or NONE

Yes, having
```java
if (transactionState != IN) {
  assertTrue(pinnedServerAddress == null || (transactionState != ABORTED && transactionState != NONE));//optional
  pinnedServerAddress = null;
}
```
makes the intent of the `if` statement much clearer. Relying here on the not obvious knowledge _"if state is IN or ABORTED, no server will be pinned"_ makes the code hard to understand.

I also noticed that there is a method `hasActiveTransaction`. Is it possible that if we use this method, we can stop checking whether or not an action is a commit action?

driver-core/src/test/resources/unified-test-format/transactions/java-only-mongos-unpin.json
jyemin: https://jira.mongodb.org/browse/DRIVERS-1685

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ClientSessionPublisherImpl.java
jyemin: Added the requested assertion logic.

The `hasActiveTransaction` method, while tempting, does not serve the required purpose, as the first time commitTransaction is executed the state is `IN`, but we don't want to unpin.

Given this, I'm not sure if you would like to see further changes, so please let me know.  Happy to go further if it makes sense.

driver-core/src/test/resources/unified-test-format/transactions/java-only-mongos-unpin.json
stIncMale: Thank you for creating the ticket.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ClientSessionPublisherImpl.java
stIncMale: I have two proposals, which may require a discussion.

**1. Improve encapsulation**
Why do we move the logic of checking the operation type outside of `ClientSessionPublisherImpl` by introducing method `notifyNonCommitOperationInitiated` instead of having `notifyOperationInitiated` that would check the operation type itself? I propose having `notifyOperationInitiated` method instead.

**2. Use `!hasActiveTransaction()` instead of `transactionState != IN`**
> The `hasActiveTransaction` method, while tempting, does not serve the required purpose, as the first time commitTransaction is executed the state is `IN`, but we don't want to unpin.

With `if (!hasActiveTransaction())` we cannot unpin in the `IN` state because the method returns `true` for this state.

If we compare `if (transactionState != IN)` vs `if (!hasActiveTransaction())`, the comparison is equivalent to `transactionState != IN` vs `transactionState != IN && !(transactionState == COMMITTED && commitInProgress)`. The latter is a more strict condition, which does not match situations `transactionState == TransactionState.COMMITTED && commitInProgress`. `commitInProgress` is only true while a commit operation is being executed, which means that no other operation may be initiated concurrently with `commitInProgress` being true. Therefore, for the purpose of the `notifyNonCommitOperationInitiated` (or `notifyOperationInitiated`) method the checks `if (transactionState != IN)` and `if (!hasActiveTransaction())` are equivalent, with the latter being more expressive of its intent and using a check that already exists instead of introducing another very similar one.

jyemin: Agreed on all points.  I was thinking if it were to become operation-aware, I would have to to add two overloads, one that took a ReadOperation and one that took a WriteOperation, and even worse those interfaces are not in the public API.  The thing I didn't see, which is obvious now, is that it's ok for this new method to take an Object.  

Just to be safe, I added an assertion that the Object is either an instance of ReadOperation or WriteOperation.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ClientSessionPublisherImpl.java
```diff
@@ -74,6 +77,16 @@ public boolean notifyMessageSent() {
         }
     }
 
+    @Override
+    public void notifyOperationInitiated(final Object operation) {
+        assertTrue(operation instanceof ReadOperation || operation instanceof WriteOperation);
```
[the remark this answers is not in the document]
jyemin: Yep.  The assertion isn't firing because all (I think) operations implement both the sync and async interfaces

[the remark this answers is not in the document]
stIncMale: Thanks. Yes, that appears to be the case.

stIncMale (approved):

PR #699, opened 2021-04-25
Fix deadlock and couple more problems in `DefaultConnectionPool`

stIncMale:
I discovered 3 problems in `DefaultConnectionPool` and `ConcurrentPool` introduced by me in https://github.com/mongodb/mongo-java-driver/pull/685.

1. [deadlock] See this [commit message](https://github.com/mongodb/mongo-java-driver/pull/699/commits/9e7348281b27be21288783330e6d0e9cdeac9e7d) for the details.
2. [not expecting an exception where it may happen] Removing from `OpenConcurrencyLimiter.desiredConnectionSlots` despite not putting anything there previously. See this [commit message](https://github.com/mongodb/mongo-java-driver/pull/699/commits/9e7348281b27be21288783330e6d0e9cdeac9e7d) for the details.
3. [double release on exception] The method `ConcurrentPool.ensureMinSize` explicitly requires a caller to release resources if `initialize` throws an exception, but then releases the permit for `newItem` itself, which leads to double release. This was by far the hardest problem to investigate and find the cause. See this [commit message](https://github.com/mongodb/mongo-java-driver/pull/699/commits/5ec1763d8e403b96db8d0fd3b3bf2206b6bb87d7) for more details.

Commits in this PR can be reviewed one by one.

JAVA-3927

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
@@ -150,7 +150,7 @@ public class ConcurrentPool<T> implements Pool<T> {
 
         T t = available.pollLast();
         if (t == null) {
-            t = createNewAndReleasePermitIfFailure(noInit -> {});
+            t = createNewAndReleasePermitIfFailure();
         }
 
         return t;
@@ -197,28 +197,27 @@ public class ConcurrentPool<T> implements Pool<T> {
      * The {@code postCreate} action throwing a exception causes this method to stop and re-throw that exception.
      *
      * @param initialize An action applied to non-{@code null} new items.
-     *                   If this action throws an {@link Exception}, it must release resources associated with the item.
+     *                   If an exception is thrown by the action, the action must treat the provided item as if obtained via
+     *                   a {@link #get(long, TimeUnit) get…} method, {@linkplain #release(Object, boolean) releasing} it
+     *                   if an exception is thrown; otherwise the action must not release the item.
      */
     public void ensureMinSize(final int minSize, final Consumer<T> initialize) {
         while (getCount() < minSize) {
             if (!acquirePermit(0, TimeUnit.MILLISECONDS)) {
                 break;
             }
-            T newItem = createNewAndReleasePermitIfFailure(initialize);
+            T newItem = createNewAndReleasePermitIfFailure();
+            initialize.accept(newItem);
             release(newItem);
         }
     }
 
-    /**
-     * @param initialize See {@link #ensureMinSize(int, Consumer)}.
-     */
-    private T createNewAndReleasePermitIfFailure(final Consumer<T> initialize) {
+    private T createNewAndReleasePermitIfFailure() {
         try {
             T newMember = itemFactory.create();
             if (newMember == null) {
                 throw new MongoInternalException("The factory for the pool created a null item");
             }
-            initialize.accept(newMember);
             return newMember;
         } catch (RuntimeException e) {
             permits.release();
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -733,10 +733,10 @@ class DefaultConnectionPool implements ConnectionPool {
          * one becomes available while waiting for a permit.
          * The first phase has one of the following outcomes:
          * <ol>
-         *     <li>A {@link MongoTimeoutException} or a different {@link Exception} is thrown.</li>
+         *     <li>A {@link MongoTimeoutException} or a different {@link Exception} is thrown,
+         *     and the specified {@code connection} is {@linkplain PooledConnection#closeSilently() silently closed}.</li>
          *     <li>An opened connection different from the specified one is returned,
-         *     and the specified {@code connection} is {@linkplain PooledConnection#closeSilently() silently closed}.
-         *     </li>
+         *     and the specified {@code connection} is {@linkplain PooledConnection#closeSilently() silently closed}.</li>
          *     <li>A permit is acquired, {@link #connectionCreated(ConnectionPoolListener, ConnectionId)} is reported
          *     and an attempt to open the specified {@code connection} is made. This is the second phase in which
          *     the {@code connection} is {@linkplain PooledConnection#open() opened synchronously}.
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -62,6 +62,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.StampedLock;
 
 import static com.mongodb.assertions.Assertions.assertFalse;
 import static com.mongodb.assertions.Assertions.assertNotNull;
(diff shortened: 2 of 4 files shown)


driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
```diff
@@ -236,18 +241,30 @@ public void shouldPruneAfterMaintenanceTaskRuns() throws InterruptedException {
...
                 .build());
-        assertUseConcurrently(provider, 2 * maxAvailableConnections, cachedExecutor, SECONDS.toNanos(15));
+        assertUseConcurrently(provider, concurrentUsersCount, checkoutSync, checkoutAsync, invalidateProb,
+                cachedExecutor, SECONDS.toNanos(10));
+    }
+
+    private static Stream<Arguments> concurrentUsageArguments() {
+        return Stream.of(
+                Arguments.of(0, 1, 8, true, false, 0.02f),
+                Arguments.of(0, 1, 8, false, true, 0.02f),
```
stIncMale, line 264: Of all 5 parameter sets, only this one reproduced the deadlock.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -710,7 +698,7 @@ public Prune shouldPrune(final UsageTrackingInternalConnection usageTrackingConn
     }
 
     /**
-     * Package-private methods are thread-safe,
+     * Package-access methods are thread-safe,
```
stIncMale, line 701: "package-private" is not a term the JLS uses, while ["package-access"](https://docs.oracle.com/javase/specs/jls/se16/html/jls-6.html#d5e10927) is.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -874,7 +864,7 @@ private PooledConnection acquirePermitOrGetAvailableOpenedConnection(final boole
                 return availableConnection;
             } finally {
                 try {
-                    if (tryGetAvailable && availableConnection == null) {//the desired connection slot has not yet been removed
+                    if (expressedDesireToGetAvailableConnection && availableConnection == null) {
```
stIncMale, line 867: This fixes the second problem I discovered while implementing pausable pool. `getPooledConnectionImmediately` there may throw an exception, which in turn may result in this code calling `giveUpOnTryingToGetAvailableConnection` despite `expressDesireToGetAvailableConnection` not being called.

This is an example of how one should try not to rely if possible on methods not throwing exceptions.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -968,17 +958,15 @@ private void tryLock(final Timeout timeout) throws MongoTimeoutException {
...
-         * Returns {@code timeoutNanos} if {@code timeoutNanos} is negative, otherwise returns 0 or a positive value.
-         *
-         * @param timeoutNanos Use a negative value for an infinite timeout,
-         *                     in which case {@link Condition#awaitNanos(long)} is called with {@link Long#MAX_VALUE}.
+         * @param timeoutNanos See {@link Timeout#startNow(long)}.
+         * @return The remaining duration as per {@link Timeout#remainingOrInfinite(TimeUnit)} if waiting ended early either
+         * spuriously or because of receiving a signal.
          */
-        private long awaitNanos(final long timeoutNanos) {
+        private long awaitNanos(final Condition condition, final long timeoutNanos) throws MongoInterruptedException {
```
stIncMale, line 965: This change makes the behavior of this method consistent with `Timeout` and allows to see explicitly what condition it waits for on the calling side.

driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
```diff
@@ -197,28 +197,27 @@ public void prune() {
...
      */
     public void ensureMinSize(final int minSize, final Consumer<T> initialize) {
         while (getCount() < minSize) {
             if (!acquirePermit(0, TimeUnit.MILLISECONDS)) {
                 break;
             }
-            T newItem = createNewAndReleasePermitIfFailure(initialize);
+            T newItem = createNewAndReleasePermitIfFailure();
+            initialize.accept(newItem);
             release(newItem);
```
rozza, line 211: Should this be in a finally block?

rozza (approved):
One question but LGTM

driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
stIncMale: No, it's important for this method to not always be called:

1. if creation failed, `createNewAndReleasePermitIfFailure` releases the permit;
2. if initialization failed, the initializer (`initialize`) releases the item as per the contract of the `ensureMinSize` method; (this contract seems to be the lesser of all evils, allowing to not further complicate the implementation of `initialize`)
3. `release(newItem)` is called only when both creation of `newItem` and initialization succeeded;

We still can put `release(newItem)` in a `finally` block like this:
```java
T newItem = null;
boolean success = false;
if (!acquirePermit(0, TimeUnit.MILLISECONDS)) {
    break;
}
try {
    newItem = createNewAndReleasePermitIfFailure();
    initialize.accept(newItem);
    success = true;
} finally {
    if (success) {
        release(newItem);
    }
}
```

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -968,17 +958,15 @@ private void tryLock(final Timeout timeout) throws MongoTimeoutException {
...
             try {
-                if (timeoutNanos < 0) {
-                    //noinspection ResultOfMethodCallIgnored
-                    condition.awaitNanos(Long.MAX_VALUE);
-                    return timeoutNanos;
+                if (timeoutNanos < 0 || timeoutNanos == Long.MAX_VALUE) {
+                    condition.await();
+                    return -1;
```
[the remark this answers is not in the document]
stIncMale: > Is there any significance to returning -1 here?

The return value must be consistent with what `Timeout.remainingOrInfinite(TimeUnit)` returns (this is stated in the documentation of the `awaitNanos` method: `@return The remaining duration as per {@link Timeout#remainingOrInfinite(TimeUnit)}`). In this specific branch of execution, the return value needs to represent an infinite timeout, i.e., it must be negative (the absolute value is irrelevant).

> Is it handled specially by the caller?

The way callers handle the return value is the same they handle the return value of [`Condition.awaitNanos`](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/concurrent/locks/Condition.html#awaitNanos(long)) - by checking whether it represents that the time run out and feeding the value back to `awaitNanos` if there is still time left to wait:
```java
long remainingNanos = timeout.remainingOrInfinite(NANOSECONDS);
while (permits == 0) {
  //check the rest of our complex condition
  if (Timeout.expired(remainingNanos)) {
    throw createTimeoutException(timeout);
  }
  remainingNanos = awaitNanos(condition, remainingNanos);
}
```

The peculiarity is that we need to reconcile `Timeout` with `Condition.awaitNanos`, which have an inherent difference: `awaitNanos` does not have a concept of infinite timeout, while `Timeout` does<sup>*</sup>. As a result, our wrapper of the `Condition.awaitNanos` needs to take arguments and return values that are consistent with `Timeout.startNow(long)` and `Timeout.remainingOrInfinite(TimeUnit)` respectively (as is now specified by the documentation of our `awaitNanos`).

---
<sup>*</sup> For infinite timeouts there is a special method `Condition.await`, but if I were to go the same design path here, I would have to have the logic that differentiates between the infinite and finite timeouts on the calling side, which would shift the complexity there, which negatively affects readability and generally speaking requires writing more code (`awaitNanos` is defined once, but may be called from multiple places). I wish Java SE API unified methods for infinite and finite timeouts (similarly how our `awaitNanos` does that), but it does not.

driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
rozza: Thanks for the explanation - no need for any code change.

PR #700, opened 2021-04-27
Add support for IAM role authentication to AWS Elastic Kubernetes Service (EKS)

HoutanSadafi:
Mongodb driver supports authenticating with mongodb in AWS using IAM roles, currently it only supports retrieving the temporary credentials for the role in ECS and EC2. For those running on EKS (amazon managed kubernetes), you need to allow your pods to inherit the IAM role (and policies) of the node it is running on, which goes against the principle of least privilege. AWS recently released [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) which basically assigns an IAM role to the pod, allowing each deployed pod on a node to have its own IAM role, with this developers can explicitly prevent the pod from inheriting the nodes IAM role and instead use the IAM role assigned to it. This pull request adds support for using the IAM role on the pod to retrieve the temporary credentials used by the driver during authentication.

JAVA-4118

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/AwsAuthenticator.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AwsAuthenticator.java
@@ -22,6 +22,11 @@ import com.mongodb.MongoException;
 import com.mongodb.MongoInternalException;
 import com.mongodb.ServerAddress;
 import com.mongodb.ServerApi;
+import com.mongodb.internal.connection.aws.AwsCredentials;
+import com.mongodb.internal.connection.aws.AwsCredentialsMiddleware;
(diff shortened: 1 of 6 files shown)


HoutanSadafi:
There were no tests for the existing aws IAM role authentication flow. I'm not sure how that is tested, is it manually done?

I am happy to write some unit test, though it may require tease apart some of the code to inject dependencies.

jyemin:
Hi @HoutanSadafi thanks for starting to work on this.  At this point we've decided to go with the approach implemented in JAVA-4292, where we allow the application to register a credentials-obtaining callback with the driver, instead of adding explicit support for EKS in the driver.

PR #702, opened 2021-04-29
Simplify `DefaultConnectionPool.OpenConcurrencyLimiter.tryLock`

stIncMale:
TODO: rebase on top of `master` after merging https://github.com/mongodb/mongo-java-driver/pull/699, then request for review.

JAVA-3927

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -827,7 +827,7 @@ class DefaultConnectionPool implements ConnectionPool {
                 throws MongoTimeoutException {
             PooledConnection availableConnection = null;
             boolean expressedDesireToGetAvailableConnection = false;
-            tryLock(timeout);
+            lockInterruptibly(lock);
             try {
                 if (tryGetAvailable) {
                     /* An attempt to get an available opened connection from the pool (must be done while holding the lock)
@@ -935,25 +935,12 @@ class DefaultConnectionPool implements ConnectionPool {
             }
         }
 
-        /**
-         * @param timeout If {@linkplain Timeout#isInfinite() infinite},
-         *                then the lock is {@linkplain Lock#lockInterruptibly() acquired interruptibly}.
-         */
-        private void tryLock(final Timeout timeout) throws MongoTimeoutException {
-            boolean success;
+        private void lockInterruptibly(final Lock lock) throws MongoInterruptedException {
             try {
-                if (timeout.isInfinite()) {
-                    lock.lockInterruptibly();
-                    success = true;
-                } else {
-                    success = lock.tryLock(timeout.remaining(NANOSECONDS), NANOSECONDS);
-                }
+                lock.lockInterruptibly();
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
                 throw new MongoInterruptedException(null, e);
-            }
-            if (!success) {
-                throw createTimeoutException(timeout);
             }
         }
 


driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -838,7 +838,7 @@ void openAsyncOrGetAvailable(
         private PooledConnection acquirePermitOrGetAvailableOpenedConnection(final boolean tryGetAvailable, final Timeout timeout)
                 throws MongoTimeoutException {
             PooledConnection availableConnection = null;
-            tryLock(timeout);
+            lockInterruptibly(lock);
```
jyemin, line 841: Why is it ok to ignore the timeout when acquiring `lock`?

jyemin (commented):
Can you provide some context as to the purpose of this change?  It seems like a behavioral change, but the PR title seems to indicate that it's not.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: To some extent this is answered in the commit message (which I modified several times, which resulted in re-creating the only commit in this PR again and again). But since this is still not obvious, let's think about a situation when locking may take time significant enough for us to want to make the locking itself timed.

The time to acquire this fair reentrant lock is `numberOfContendingThreads * averageTimeBeforeCallingWait`. `numberOfContendingThreads` is the number of threads which are ahead of the considered thread in the imaginary queue of threads that concurrently want to acquire the lock; `averageTimeBeforeCallingWait` is usually the average time to check the condition, but in our case it's more than that.

`numberOfContendingThreads`
Most of the time this lock is uncontended because threads get opened connections, and only after invalidation of the pool (or after creation of the pool) the lock has a chance to be contended.

`averageTimeBeforeCallingWait`
Has a potential to take some time only if the pool has many stale connections, which may happen if it has recently been invalidated and has large `maxSize`. While any individual invocation of `getPooledConnectionImmediately` has a potential to be slightly time-consuming (and, therefore, occasionally acquiring the lock may take comparable time), on average this method is of constant small time because the majority of invocations observe no stale connections.

If we also take into account that
1. all timeout implementations always overtime the specified timeout and that
2. we have a much more significant way to overtime: all asynchronous check outs and all (internal to the pool) asynchronous open requests are queued (open requests are also throttled) because they are executed by a single thread

worrying about the time to acquire the lock here does not seem important enough anymore. These thoughts appeared to me only recently, and not when I was implementing the functionality.

jyemin: Thanks for the more detailed explanation.  And really, the problem was that I didn't even look at the commit message.  Usually with single commit PRs the commit message is copied to the PR description automatically.  Perhaps that didn't happen because, as you said, you changed the message a few times.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -945,26 +945,13 @@ void tryHandOverOrRelease(final UsageTrackingInternalConnection openConnection)
             }
         }
 
-        /**
-         * @param timeout If {@linkplain Timeout#isInfinite() infinite},
-         *                then the lock is {@linkplain Lock#lockInterruptibly() acquired interruptibly}.
-         */
-        private void tryLock(final Timeout timeout) throws MongoTimeoutException {
-            boolean success;
+        private void lockInterruptibly(final Lock lock) throws MongoTimeoutException {
```
jyemin, line 948: Why the change to take the lock as the parameter?  Is there more than lock that we could be using in this context?
Taking the lock as a param it seems more like a utility method and therefore should be static. 

What do you think?

jyemin, line 948: You can remove the throws clause now.

jyemin (changes requested):

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: Thanks, it should have been `MongointerruptedException`. Done.

stIncMale: > Why the change to take the lock as the parameter? Is there more than lock that we could be using in this context?

The change is to remove the weird asymmetry. Previously the code looked like
```java
lockInterruptibly(); // use a method that uses the `lock` field internally
try {
  ...
} finally {
  lock.unlock(); // use the `lock` field explicitly
}
```
now it looks like
```java
lockInterruptibly(lock); // use the `lock` field explicitly
try {
  ...
} finally {
  lock.unlock(); // use the `lock` field explicitly
}
```

> a utility method and therefore should be static

I 100% agree, and I wish we could write static methods in inner classes, but Java syntax forbids that (this restriction seems to be artificial and not helpful).

stIncMale:
> Can you provide some context as to the purpose of this change? It seems like a behavioral change, but the PR title seems to indicate that it's not.

@jyemin Now that you know there is a commit message, is this request still actual?

jyemin (approved):
LGTM!

PR #708, opened 2021-05-12
Improve assertion messages in unified test runner

jyemin:
* Add a context for operation execution.  This provides context when the
  operation itself asserts
* Include the index of the operation in the operations array.  This helps
  when there are multiple operations that are identical and you want to
  be able to determine which one failed

JAVA-4152

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
@@ -97,7 +97,7 @@ abstract class ContextElement {
         }
 
         public String toString() {
-            return "Started Operation Result Context: " + "\n"
+            return "Started Operation Context: " + "\n"
                     + "   Operation:\n"
                     + operation.toJson(JsonWriterSettings.builder().indent(true).build()) + "\n"
                     + "   Operation index: " + index + "\n";
@@ -116,7 +116,7 @@ abstract class ContextElement {
         }
 
         public String toString() {
-            return "Completed Operation Result Context: " + "\n"
+            return "Completed Operation Context: " + "\n"
                     + "   Operation:\n"
                     + operation.toJson(JsonWriterSettings.builder().indent(true).build()) + "\n"
                     + "   Actual result:\n"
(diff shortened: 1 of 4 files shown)


rozza (approved):

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
```diff
@@ -83,21 +87,41 @@ public String toString() {
...
+        private final BsonDocument operation;
+        private final int index;
+
+        StartedOperationContextElement(final BsonDocument operation, final int index) {
+            this.operation = operation;
+            this.index = index;
+        }
+
+        public String toString() {
+            return "Started Operation Result Context: " + "\n"
```
stIncMale, line 100: I do not understand what "Started Operation Result Context" is. Maybe "Started Operation Context" is less confusing given that there is no result?

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
```diff
@@ -83,21 +87,41 @@ public String toString() {
...
-        OperationContextElement(final BsonDocument operation, final OperationResult result) {
+        CompletedOperationContextElement(final BsonDocument operation, final OperationResult result, final int index) {
             this.operation = operation;
             this.result = result;
+            this.index = index;
         }
 
         public String toString() {
-            return "Operation Result Context: " + "\n"
+            return "Completed Operation Result Context: " + "\n"
```
stIncMale, line 119: It appears that this should be either "Completed Operation Context" or "Operation Result Context" (the original text). However, "Completed Operation Context" is more consistent with the class name `CompletedOperationContextElement`.

stIncMale (changes requested):
I messed something up, it is supposed to be a "Request changes" review.

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
jyemin: Done

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
jyemin: Done

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
jyemin: I think part of what makes this confusing is that "Started" (and "Completed" below) are used here as adjectives modifying "Operation", rather that verbs.

stIncMale (approved):

PR #709, opened 2021-05-12
Fix JsonPoweredCrudTestHelper

jyemin:
The change stream cursors were not being closed, which was
causing problems in downstream tests

JAVA-4153

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/JsonPoweredCrudTestHelper.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/JsonPoweredCrudTestHelper.java
@@ -1084,19 +1084,34 @@ public class JsonPoweredCrudTestHelper {
     BsonDocument getClientWatchResult(final BsonDocument collectionOptions, final BsonDocument rawArguments,
                                       @Nullable final ClientSession clientSession) {
         MongoCursor<ChangeStreamDocument<Document>> cursor = mongoClient.watch().iterator();
-        return new BsonDocument("ok", new BsonInt32(1));
+        //noinspection TryFinallyCanBeTryWithResources
+        try {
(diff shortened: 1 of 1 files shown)


rozza (approved):
Confusing but I get that the cursor is needed to create events.

LGTM

driver-sync/src/test/functional/com/mongodb/client/JsonPoweredCrudTestHelper.java
```diff
@@ -1084,19 +1084,34 @@ BsonDocument getUploadResult(final BsonDocument collectionOptions, final BsonDoc
     BsonDocument getClientWatchResult(final BsonDocument collectionOptions, final BsonDocument rawArguments,
                                       @Nullable final ClientSession clientSession) {
         MongoCursor<ChangeStreamDocument<Document>> cursor = mongoClient.watch().iterator();
-        return new BsonDocument("ok", new BsonInt32(1));
+        //noinspection TryFinallyCanBeTryWithResources
```
stIncMale, line 1087: It's a shame that whether we use try-with-resources in this case (which I would prefer seeing here, but approving anyway) or not, we have to either use `//noinspection TryFinallyCanBeTryWithResources` or `@SuppressWarnings("try")`. The latter is a manifestation of how useless the `javac`'s _"warning: [try] auto-closeable resource cursor is never referenced in body of corresponding try statement"_ is, because we can only put `@SuppressWarnings("try")` on the whole method.

stIncMale (approved):

driver-sync/src/test/functional/com/mongodb/client/JsonPoweredCrudTestHelper.java
jyemin: Indeed, it is.

PR #713, opened 2021-05-14
`AsyncChangeStreamBatchCursor` must cache `resumeToken` after completion of the async `cursor.next`/`tryNext` operation

stIncMale:
And also after extracting `resumeToken` from the received change stream documents; at least, this is consistent with the behaviour of `ChangeStreamBatchCursor`.

JAVA-4165

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
@@ -82,7 +82,6 @@ final class AsyncChangeStreamBatchCursor<T> implements AsyncAggregateResponseBat
             public void apply(final AsyncAggregateResponseBatchCursor<RawBsonDocument> cursor,
                               final SingleResultCallback<List<RawBsonDocument>> callback) {
                 cursor.next(callback);
-                cachePostBatchResumeToken(cursor);
             }
         }, convertResultsCallback(callback), false);
     }
@@ -94,7 +93,6 @@ final class AsyncChangeStreamBatchCursor<T> implements AsyncAggregateResponseBat
             public void apply(final AsyncAggregateResponseBatchCursor<RawBsonDocument> cursor,
                               final SingleResultCallback<List<RawBsonDocument>> callback) {
                 cursor.tryNext(callback);
-                cachePostBatchResumeToken(cursor);
             }
         }, convertResultsCallback(callback), true);
     }
@@ -187,8 +185,9 @@ final class AsyncChangeStreamBatchCursor<T> implements AsyncAggregateResponseBat
     }
 
     private void cachePostBatchResumeToken(final AsyncAggregateResponseBatchCursor<RawBsonDocument> queryBatchCursor) {
-        if (queryBatchCursor.getPostBatchResumeToken() != null) {
-            resumeToken = queryBatchCursor.getPostBatchResumeToken();
+        BsonDocument resumeToken = queryBatchCursor.getPostBatchResumeToken();
+        if (resumeToken != null) {
+            this.resumeToken = resumeToken;
         }
     }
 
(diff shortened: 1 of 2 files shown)


driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
```diff
@@ -82,7 +82,6 @@ public void next(final SingleResultCallback<List<T>> callback) {
             public void apply(final AsyncAggregateResponseBatchCursor<RawBsonDocument> cursor,
                               final SingleResultCallback<List<RawBsonDocument>> callback) {
                 cursor.next(callback);
-                cachePostBatchResumeToken(cursor);
```
stIncMale, line 85: This is the main culprit - an obvious race, once you know what line to look at.

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
```diff
@@ -227,23 +226,29 @@ public void onResult(final List<RawBsonDocument> rawDocuments, final Throwable t
         void apply(AsyncAggregateResponseBatchCursor<RawBsonDocument> cursor, SingleResultCallback<List<RawBsonDocument>> callback);
     }
 
-    private void resumeableOperation(final AsyncBlock asyncBlock, final SingleResultCallback<List<RawBsonDocument>> callback,
+    private void resumeableOperation(final AsyncBlock asyncBlock, final SingleResultCallback<List<RawBsonDocument>> convertResultsCallback,
```
stIncMale, line 229: Renamed `callback` to `convertResultsCallback` to make it more clear what the callback does. While one may guess this by staring at the original callback type `SingleResultCallback<List<RawBsonDocument>>` and the callback type `SingleResultCallback<List<T>>`, the variable name in this case does help because the converting callback type does not have a custom name.

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
```diff
@@ -227,23 +226,29 @@ public void onResult(final List<RawBsonDocument> rawDocuments, final Throwable t
...
+                    convertResultsCallback.onResult(result, null);
                 } else if (isRetryableError(t, maxWireVersion)) {
-                    nullifyAndCloseWrapped();
-                    retryOperation(asyncBlock, callback, tryNext);
+                    retry = true;
                 } else {
-                    callback.onResult(null, t);
+                    convertResultsCallback.onResult(null, t);
+                }
+                cachePostBatchResumeToken(wrappedCursor);
```
stIncMale, line 248: Cashing after converting results and calculating `resumeToken` from them just as it's done in the synchronous counterpart (`ChangeStreamBatchCursor`). Note also that we must cache `resumeToken` before we retry.

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
jyemin: Oh my.

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
jyemin, line 248: In the case where this method does call convertResultsCallback with a result, could it be consequential whether the post batch resume token is initialized prior to invoking the callback?  My concern is that the callback might be relying on it being set.  It's certainly strange that the order of operations is now:

1. set resumeToken to the last document's _id
2. invoke the callback
3. set resumeToken to the post batch resume token

stIncMale: - Thanks! You are correct, the `set resumeToken to the post batch resume token` action must happen before we complete the callback just like `set resumeToken to the last document's _id` happens before completing the callback.
- On the other hand, `set resumeToken to the last document's _id` must happen before `set resumeToken to the post batch resume token` (as I described [here|https://github.com/mongodb/mongo-java-driver/pull/713#discussion_r632727156], though I am not sure is this is actually a "must").

These two constraints mean that with the current code I would have to call `cachePostBatchResumeToken` both

- in `convertResultsCallback` (right before calling `callback.onResult(results, null)` and `callback.onResult(null, null)`, i.e. in two places);
- and in `resumeableOperation` for the retry branch.

I will look if the code can be refactored to avoid such a mess.

stIncMale: Done. There is no more callback that converts data, and even with the fix (it adds lines) the number of lines in `AsyncChangeStreamBatchCursor` went from 285 to 267, which I consider a small win :)

driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
```diff
@@ -161,17 +166,24 @@ private void cachePostBatchResumeToken(final AggregateResponseBatchCursor<RawBso
         }
     }
 
-    private List<T> convertResults(final List<RawBsonDocument> rawDocuments) {
+    /**
+     * @param lastIdConsumer Is {@linkplain Consumer#accept(Object) called} iff {@code rawDocuments} is successfully converted
```
rozza, line 170: Nit: typo `iff`

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
rozza: > As a result, postBatchResumeToken always overrides the resumeToken calculated from the received change stream documents if postBatchResumeToken is sent by the server (not sure if this is important, but this is how ChangeStreamBatchCursor is written).

:+1: `postBatchResumeToken` only started being sent by the server in MongoDB 4.0.7. Changes streams for 
rarely updated collections on a busy databases were failing the retryability, as the last seen resumeToken fell out the oplog. So sending a new token each getMore was the workaround.

rozza (approved):
LGTM - one javadoc nit

driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
stIncMale: Actually, it's not a typo but a short for ["if and only if"](https://en.wikipedia.org/wiki/If_and_only_if). It'd be nice if we agreed to adapt this shortening.

Do you still want me to change it to "if and only if"?

rozza: TIL :) I'm happy with if or iff in this case.

driver-core/src/main/com/mongodb/internal/operation/AsyncChangeStreamBatchCursor.java
jyemin: Looks good!

jyemin (approved):
LGTM.  Don't forget to update commit message to refer to JAVA-4165.

PR #716, opened 2021-05-19
Update integration tests to account for new explain command reply format

jyemin:
JAVA-4169

This accounts for four of the tests now failing against latest


driver-core/src/test/functional/com/mongodb/internal/operation/QueryOperationHelper.groovy
```diff
@@ -32,7 +32,12 @@ class QueryOperationHelper {
 
     static BsonDocument getKeyPattern(BsonDocument explainPlan) {
         BsonDocument winningPlan = explainPlan.getDocument('queryPlanner').getDocument('winningPlan')
-        if (winningPlan.containsKey('inputStage')) {
+        if (winningPlan.containsKey('queryPlan')) {
```
jyemin, line 35: This could certainly be DRY-ed up, but it didn't seem worth it.

jyemin:
Tests are failing.  Looking into it.

stIncMale (approved):

jyemin:
Merged with one LGTM, since the change is simple and only to test code

PR #717, opened 2021-05-19
Use "hello" command where possible

jyemin:
There are two scenarios where "hello" command can be used:

1. When the application declares an API version
3. When the response to "ismaster" contains helloOk: true

Additionally, use OP_MSG when the application declares an API version,
even when maxWireVersion is unknown during the handshake

JAVA-4092
JAVA-4157

### DIFF
--- a/driver-core/src/test/unit/com/mongodb/internal/connection/MessageHelper.java
+++ b/driver-core/src/test/unit/com/mongodb/internal/connection/MessageHelper.java
@@ -35,6 +35,7 @@ import java.nio.ByteOrder;
 
 import static com.mongodb.ClusterFixture.getServerApi;
 import static com.mongodb.connection.ConnectionDescription.getDefaultMaxMessageSize;
+import static org.junit.Assert.fail;
 
 final class MessageHelper {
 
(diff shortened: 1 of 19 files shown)


driver-core/src/test/unit/com/mongodb/internal/connection/MessageHelper.java
```diff
@@ -105,6 +114,10 @@ public static String getApiVersionField() {
         return getServerApi() == null ? "" : ", \"apiVersion\": \"" + getServerApi().getVersion().getValue() + "\"";
     }
 
+    public static String getDbField(final String databaseName) {
```
rozza, line 117: Is this related to hello?

rozza (commented):
One question about the db change in auth.

driver-core/src/test/unit/com/mongodb/internal/connection/MessageHelper.java
jyemin: Loosely.  It's related to the change in CommandMessage.getOpCode where we now use OP_MSG if API version is set.  If API version is non-null, we use OP_MSG, and when we use OP_MSG the command document includes a $db field.

rozza (approved):
LGTM

driver-core/src/test/unit/com/mongodb/internal/connection/MessageHelper.java
rozza: ACK - thanks for the clarification

PR #719, opened 2021-05-25
Add one more thread to `DefaultConnectionPool` for handling timeouts of async tasks

stIncMale:
JAVA-4146

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -50,13 +50,19 @@ import org.bson.ByteBuf;
 import org.bson.codecs.Decoder;
 import org.bson.types.ObjectId;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Deque;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -65,7 +71,8 @@ import java.util.concurrent.atomic.LongAdder;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.concurrent.locks.StampedLock;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 import static com.mongodb.assertions.Assertions.assertFalse;
 import static com.mongodb.assertions.Assertions.assertNotNull;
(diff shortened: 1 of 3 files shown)


stIncMale:
Removed reviewers until I fix the OOM problem. Also, @BorisDog proposed a way to potentially simplify the solution.

stIncMale:
Re-added reviewers.

The OOM problem is going to be solved in JAVA-4178.

The approach proposed by Boris (the same we use to implement async timeout in `NettyStream`) does not make things simpler because with `DefaultConnectionPool` we have concurrent async requests queued, and we need to remove them from the queue when they time out.

rozza (commented):
Looking good - one question is merging this ticket blocked by JAVA-4178? 

I'm unclear as to OOM issues caused by this change.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1170,55 +1183,270 @@ int getNumPinnedToTransaction() {
...
+                            task.execute();
+                        }
+                    } catch (RuntimeException e) {
+                        LOGGER.error(null, e);
+                    }
                 }
+            } catch (InterruptedException closed) {
+                // fail the rest of the tasks and stop
             }
+            LOGGER.debug("Stopping");
```
rozza, line 1305: Nit: Logging seems inconsistent.  Suggest either remove the logging or improve the logging message and include it elsewhere if needed.

rozza (changes requested):

stIncMale:
> one question is merging this ticket blocked by JAVA-4178?

Right, I will merge this one after merging JAVA-4178.

> I'm unclear as to OOM issues caused by this change.

The size of `DefaultConnectionPool` was drastically increased in this PR, because we added one more thread. Not only this affects the retained Java heap size of its instances, but also increases the native heap usage for the process (because for each new Java thread a new OS thread is created). This was enough for a different problem (JAVA-4178) to start manifesting itself.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1170,55 +1183,270 @@ int getNumPinnedToTransaction() {
...
+     * <p>
+     * Due to tasks having a timeout, we also must monitor queued tasks to determine which ones timed out and completing them accordingly.
+     * While this logic can be incorporated in {@code create}/{@code open} tasks, doing so would have significantly complicate them further,
+     * and it is especially not obvious how to maintain fairness in that case.
+     */
     @ThreadSafe
-    private static class Workers implements AutoCloseable {
-        private volatile ExecutorService getter;
-        private volatile ExecutorService opener;
+    private static class AsyncWorkManager implements AutoCloseable {
```
jyemin, line 1202: Given all the new synchronization logic in this class, should we add a stress/concurrency unit test to detect subtle bugs or race conditions?

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1170,55 +1183,270 @@ int getNumPinnedToTransaction() {
...
+                if (initWorkersUnlessClosed()) {
+                    queue.add(task);
+                    closedOrTaskAddedCondition.signalAll();
+                    return;
                 }
+            } finally {
+                lock.unlock();
             }
-            return getter;
+            task.failAsClosed();
```
jyemin, line 1242: Is this line of code outside of the lock for correctness, performance, or style? If style, I will note that I prefer to avoid early returns and would move this into an else block inside the try.  But it's personal preference so I don't insist on it.

stIncMale: This is done for correctness: we should avoid completing callbacks while holding locks. It still may happen because a caller holds a lock while calling `DefaultConnectionPool.get`, but the class does the best it can.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: While thinking about execution in which the timeout handling was broken, I realized that things can (and must) be much better that they currently are. I changed this PR to the draft status and opened https://github.com/mongodb/mongo-java-driver/pull/722. The commit messages there explain everything, I am also happy to discuss the matter over a call if that would be simpler.

PR #720, opened 2021-05-25
Close server/pool in `LoadBalancedCluster` and `LoadBalancedServer` respectively

stIncMale:
JAVA-4178

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/LoadBalancedCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/LoadBalancedCluster.java
@@ -74,7 +74,7 @@ final class LoadBalancedCluster implements Cluster {
     private final ClusterListener clusterListener;
     private ClusterDescription description;
     @Nullable
-    private volatile ClusterableServer server;
+    private ClusterableServer server;
     private final AtomicBoolean closed = new AtomicBoolean();
     private final DnsSrvRecordMonitor dnsSrvRecordMonitor;
     private volatile MongoException srvResolutionException;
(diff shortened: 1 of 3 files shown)


driver-core/src/main/com/mongodb/internal/connection/LoadBalancedCluster.java
```diff
@@ -259,6 +259,7 @@ public void close() {
             if (dnsSrvRecordMonitor != null) {
                 dnsSrvRecordMonitor.close();
             }
+            server.close();
```
jyemin, line 262: I think we need a null check for server.

stIncMale: Thanks!  Indeed, not only we need to account for it being `null`-able, but also now we need it to be `volatile`, because `close()` does not call `waitForSrv()` before reading from `server`.

jyemin (approved):
LGTM

driver-core/src/main/com/mongodb/internal/connection/LoadBalancedCluster.java
stIncMale: I am sorry, the previous attempt also was incorrect. Hopefully, now it is.

jyemin (commented):
Good catch.  Did you find the bug just by code inspection?

I think it would be possible to write a unit test asserting this behavior.  What do you think?

driver-core/src/test/unit/com/mongodb/internal/connection/LoadBalancedClusterTest.java
```diff
@@ -317,6 +321,25 @@ public void shouldTimeoutSelectServerAsynchronouslyWhenThereIsSRVLookupException
                 exception.getMessage());
     }
 
+    @Test
```
jyemin, line 324: Could we also have a `shouldCloseServerWhenClosing` test?

stIncMale: Sure, added in fcc97d288137284b86c588955daa3b876c949f5c.

jyemin (approved):
LGTM.  Thanks for adding the tests.

PR #722, opened 2021-05-31
 Refactor `DefaultConnectionPool` to use single worker thread when checking out asynchronously

stIncMale:
This PR is a successor of https://github.com/mongodb/mongo-java-driver/pull/719, and uses a different, much more resource-effective approach: we are back to using a single worker for async check outs, which, given that all check out operations currently have the same timeout, puts us back to a situation when async timeouts are handled correctly. Please see the commit messages for more details.

With CSOT different check out operations will have different timeouts, and we will have to change the `DefaultConnectionPool` code to handle timeouts correctly. However, that's a task for future. We can either do the same thing I did in https://github.com/mongodb/mongo-java-driver/pull/719, or we can mix the logic handling timeouts into existing blocking calls done by the (now single) async worker thread. Because of the changes done in the current PR (removal of the `getAsync` optimization that allowed the method to sometimes be executed synchronously), putting the timeout handling logic in the same async worker thread now does not seem to be a practically impossible task (I'd like to thank @BorisDog for useful conversations we had around this topic) because now it does not break fairness guarantees.

JAVA-4146

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
@@ -141,7 +141,7 @@ public class ConcurrentPool<T> implements Pool<T> {
     @Override
     public T get(final long timeout, final TimeUnit timeUnit) {
         if (closed) {
-            throw new IllegalStateException("The pool is closed");
+            throw poolClosedException();
         }
 
         if (!acquirePermit(timeout, timeUnit)) {
@@ -293,5 +293,9 @@ public class ConcurrentPool<T> implements Pool<T> {
         } catch (RuntimeException e) {
             // ItemFactory.close() really should not throw
         }
+    }
+
+    static IllegalStateException poolClosedException() {
+        return new IllegalStateException("The pool is closed");
     }
 }
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -1185,13 +1185,15 @@ class DefaultConnectionPool implements ConnectionPool {
         }
 
         /**
+         * Invocations of this method must be guarded by {@link #lock}.
+         *
          * @return {@code false} iff the {@link #state} is {@link State#CLOSED}.
          */
         private boolean initUnlessClosed() {
             boolean result = true;
             if (state == State.NEW) {
                 worker = Executors.newSingleThreadExecutor(new DaemonThreadFactory("AsyncGetter"));
-                assertNotNull(worker).submit(() -> runAndLogUncaught(this::workerRun));
+                worker.submit(() -> runAndLogUncaught(this::workerRun));
                 state = State.INITIALIZED;
             } else if (state == State.CLOSED) {
                 result = false;
@@ -1209,9 +1211,10 @@ class DefaultConnectionPool implements ConnectionPool {
             lock.lock();
             try {
                 if (state != State.CLOSED) {
-                    try (UncheckedAutoCloseable ignored = UncheckedAutoCloseable.create(worker)) {
-                        state = State.CLOSED;
-                    } // at this point we interrupt `worker`s thread
+                    state = State.CLOSED;
+                    if (worker != null) {
+                        worker.shutdownNow(); // at this point we interrupt `worker`s thread
+                    }
                 }
             } finally {
                 lock.unlock();
@@ -1268,19 +1271,6 @@ class DefaultConnectionPool implements ConnectionPool {
             NEW,
             INITIALIZED,
             CLOSED
-        }
-
-        private interface UncheckedAutoCloseable extends AutoCloseable {
-            @Override
-            void close();
-
-            static UncheckedAutoCloseable create(@Nullable final ExecutorService ex) {
-                return () -> {
-                    if (ex != null) {
-                        ex.shutdownNow();
-                    }
-                };
-            }
         }
     }
 
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -53,10 +53,12 @@ import org.bson.types.ObjectId;
 import java.util.Deque;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -65,7 +67,8 @@ import java.util.concurrent.atomic.LongAdder;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.concurrent.locks.StampedLock;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 import static com.mongodb.assertions.Assertions.assertFalse;
 import static com.mongodb.assertions.Assertions.assertNotNull;
@@ -91,7 +94,7 @@ class DefaultConnectionPool implements ConnectionPool {
     private final ConnectionPoolSettings settings;
     private final AtomicInteger generation = new AtomicInteger(0);
     private final ScheduledExecutorService sizeMaintenanceTimer;
-    private final Workers workers;
+    private final AsyncWorkManager asyncWorkManager;
     private final Runnable maintenanceTask;
     private final ConnectionPoolListener connectionPoolListener;
     private final ServerId serverId;
@@ -113,7 +116,7 @@ class DefaultConnectionPool implements ConnectionPool {
         sizeMaintenanceTimer = createMaintenanceTimer();
         connectionPoolCreated(connectionPoolListener, serverId, settings);
         openConcurrencyLimiter = new OpenConcurrencyLimiter(MAX_CONNECTING);
-        workers = new Workers();
+        asyncWorkManager = new AsyncWorkManager();
         connectionGenerationSupplier = new ConnectionGenerationSupplier() {
             @Override
             public int getGeneration() {
@@ -172,25 +175,10 @@ class DefaultConnectionPool implements ConnectionPool {
                 errHandlingCallback.onResult(null, checkOutFailed(failure));
             }
         };
-        PooledConnection immediateConnection = null;
-
-        try {
-            immediateConnection = getPooledConnection(Timeout.immediate());
-        } catch (MongoTimeoutException e) {
-            // fall through
-        } catch (RuntimeException e) {
-            eventSendingCallback.onResult(null, e);
-            return;
-        }
-
-        if (immediateConnection != null) {
-            openAsync(immediateConnection, timeout, eventSendingCallback);
-        } else {
-            workers.getter().execute(() -> {
-                if (timeout.expired()) {
-                    eventSendingCallback.onResult(null, createTimeoutException(timeout));
-                    return;
-                }
+        asyncWorkManager.enqueue(new Task(timeout, t -> {
+            if (t != null) {
+                eventSendingCallback.onResult(null, t);
+            } else {
                 PooledConnection connection;
                 try {
                     connection = getPooledConnection(timeout);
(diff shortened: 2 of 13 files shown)


stIncMale:
Without the `getAsync` optimization that allowed this method to be executed synchronously in some cases, the tests that implicitly and incorrectly relied on that started failing as a house of cards. One of such tests I fixed (to the best of what can be done) in 0763d486dbf306189f2bdc9083689e3c9034c2f8, but it is not the only one.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1170,55 +1154,170 @@ int getNumPinnedToTransaction() {
...
+        private volatile BlockingQueue<Task> tasks;
         private final Lock lock;
+        @Nullable
+        private ExecutorService worker;
 
-        Workers() {
+        AsyncWorkManager() {
+            state = State.NEW;
+            tasks = new LinkedBlockingQueue<>();
             lock = new StampedLock().asWriteLock();
```
jyemin, line 1171: I realized I don't understand why we're using StampedLock here.  Can you explain?

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1170,55 +1154,170 @@ int getNumPinnedToTransaction() {
...
-                lock.lock();
-                try {
-                    if (opener == null) {
-                        opener = Executors.newSingleThreadExecutor(new DaemonThreadFactory("AsyncOpener"));
-                    }
-                } finally {
-                    lock.unlock();
-                }
+        /**
+         * @return {@code false} iff the {@link #state} is {@link State#CLOSED}.
```
jyemin, line 1188: It may be obvious to others, but consider adding a comment indicating that this method must be called while holding `lock`

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1170,55 +1154,170 @@ int getNumPinnedToTransaction() {
...
+         */
+        private boolean initUnlessClosed() {
+            boolean result = true;
+            if (state == State.NEW) {
+                worker = Executors.newSingleThreadExecutor(new DaemonThreadFactory("AsyncGetter"));
+                assertNotNull(worker).submit(() -> runAndLogUncaught(this::workerRun));
```
jyemin, line 1194: Nit: I think this assertion is unnecessary.  Can't we assume that the JDK is implemented to its specification?

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1170,55 +1154,170 @@ int getNumPinnedToTransaction() {
...
+         * {@linkplain Task#failAsClosed() fail} asynchronously.
+         */
         @Override
+        @SuppressWarnings("try")
         public void close() {
+            lock.lock();
             try {
-                if (getter != null) {
-                    getter.shutdownNow();
+                if (state != State.CLOSED) {
```
jyemin, line 1211: I would find:

                if (state != State.CLOSED) {
                    state = State.CLOSED;
                    if (worker != null) {
                        worker.shutdownNow();
                    }
                }

easier to understand.  Is there an important functional difference that I'm not seeing?

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -172,35 +176,32 @@ public void getAsync(final SingleResultCallback<InternalConnection> callback) {
...
+                        LOGGER.trace(format("Pooled connection %s to server %s is already open",
+                                getId(connection), serverId));
+                    }
+                    eventSendingCallback.onResult(connection, null);
+                } else {
+                    if (LOGGER.isTraceEnabled()) {
+                        LOGGER.trace(format("Pooled connection %s to server %s is not yet open",
+                                getId(connection), serverId));
+                    }
+                    openConcurrencyLimiter.openAsyncOrGetAvailable(connection, timeout, eventSendingCallback);
```
jyemin, line 201: IIUC, this is ok to do in the same task because the timeout to wait for a permit will still always be shorter than the timeouts for all subsequent timeouts in the queue?

jyemin (changes requested):
This is a nice simplification, and I can't find any logical errors.  Comments are generally about style/design, documentation, and requests for clarification.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/MainTransactionsTest.java
```diff
@@ -24,8 +24,17 @@
...
 import org.bson.BsonDocument;
+import org.junit.After;
+import org.junit.Before;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 
 public class MainTransactionsTest extends AbstractMainTransactionsTest {
+    private static final Set<String> SESSION_CLOSE_TIMING_SENSITIVE_TESTS = new HashSet<>(Collections.singletonList(
```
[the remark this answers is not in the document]
stIncMale: This change is referred to under the item 1 in https://github.com/mongodb/mongo-java-driver/pull/722#issuecomment-851729301.

driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
```diff
@@ -141,7 +141,7 @@ public T get() {
     @Override
     public T get(final long timeout, final TimeUnit timeUnit) {
         if (closed) {
-            throw new IllegalStateException("The pool is closed");
+            throw poolClosedException();
```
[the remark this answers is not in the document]
stIncMale: `DefaultConnectionPool` must report that is is closed the same way `ConcurrentPool` does. Throwing different exceptions about the same condition just because a connection pool is implemented as two classes is confusing. I had to do the same thing in https://github.com/mongodb/mongo-java-driver/pull/701 (pausable pool), and additionally had to introduce `ConcurrentPool.isPoolClosedException(Throwable)`.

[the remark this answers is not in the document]
jyemin: My bad, didn't notice it was called from elsewhere now.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: Hilarious! 😆 I believe I had `worker` marked as `volatile` at some point. Given that it is also `@Nullable`, the assertion was needed to make IDEA happy. Now that the field is plain, IDEA is smart enough to see that it cannot be `null` at this point.

Done.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: Thanks, good point. This code was written while there were two workers and a timeouter. Using try-with-resources made sense at that point, but with a single worker we are better off without it.

Done.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: Done.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: Picking the functionally simplest tool for the task. Alternative would be `ReentrantLock`, which adds the following functionality:

1. fairness
2. reentrancy, which requires maintaining the owner as part of the lock object state
3. conditional variables

Since neither is needed for this use case, `ReentrantLock` is not needed.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: All checkout tasks are added to the single queue with the same timeout `settings.getMaxWaitTime(NANOSECONDS)` started before they are added to the queue. Each task then leaves the queue to be handled by the `AsyncGetter` thread, and the thread always completes the task (even if it has to wait due to both/either `maxSize`/`maxConnecting` limits) before taking another task. Thus, each task (whether queued or being handled) has less time remaining until it times out than any task after it in the queue (of than any queued task if we are talking about the task that is being handled)<sup>1</sup>. Because of this, no queued task can time out while 

----

<sup>1</sup> This is a lie and a simplification, yet we are fine. In reality, because the pair of actions (start counting down time for a task, enqueue the task) is not done atomically with regard to other such pairs, it is possible that the aforementioned guarantee is violated. However, in order to realize that, one needs to have access to the queue, which users of `DefaultConcurrentPool` do not have. Not only this detail does not violate our best effort to not overdue timeouts, but also it is how things were in the `DefaultConcurrentPool` forever.

stIncMale:
The latest [failure](https://spruce.mongodb.com/version/60b7f3803e8e865cdccbc6e4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=tests-jdk8-unsecure__version~5.0_os~linux_topology~replicaset_auth~noauth_ssl~nossl_jdk~jdk8) seem to be related to Evergreen's infrastructure problem: neither logs nor the reported test results specify the problem, and it's not like the tests hanged (test logs span over just 6 minutes).

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: I think I came up with a better explanation for the footnote 2:

1. Pool totally orders `getAsync` operations by putting the corresponding tasks in the queue.
2. For operations that are sequential (i.e. are ordered by `hb`), the queue order is consistent with `hb` and is also consistent with the order by `timeout.remaining`. For such operations a pool's user has an expectation that an earlier operation must timeout before the later operation.
3. For operations that are concurrent, the queue order may not be consistent with the order by `timeout.remaining`. For such operations a pool's user has no expectation on which must timeout first because they are not ordered from the user's perspective. So the fact that the pool may order them not by `timeout.remaining` is irrelevant to a user.

jyemin: Thanks for the detailed explanation!

jyemin (approved):
LGTM

stIncMale:
Finally fixed Scala tests in ec7f1a153aa492e9bcd96065b474003c79337820. Evergeen results are [here](https://spruce.mongodb.com/version/60ba415c7742ae03b32437b7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=scala-tests__version~4.4_os~ubuntu_topology~replicaset_auth~noauth_ssl~nossl_jdk~jdk11_scala~2.13).

driver-scala/src/test/scala/org/mongodb/scala/ApiAliasAndCompanionSpec.scala
```diff
@@ -65,7 +65,18 @@ class ApiAliasAndCompanionSpec extends BaseSpec {
       "SingleResultCallback",
       "SubjectProvider",
       "TransactionExample",
-      "UnixServerAddress"
+      "UnixServerAddress",
```
jyemin, line 68: Was it necessary to add all the primers?  I didn't think it would be.

stIncMale: I added `testImplementation project(':driver-reactive-streams').sourceSets.test.output` to be able to call and reuse `SyncMongoClient.enableSleepAfterSessionClose`/`disableSleep` and `MainTransactionsTest.SESSION_CLOSE_TIMING_SENSITIVE_TESTS` from the Java code. As a result, `ApiAliasAndCompanionSpec` started complaining about all those new classes it started to see. I added only those it was complaining about (and instead of complaining about `PublisherHelpers` it was complaining about non-existent `ObservableHelpers` due to a deceiving line [`getSimpleName.replace("Publisher", "Observable")`](https://github.com/mongodb/mongo-java-driver/pull/722/files?file-filters%5B%5D=.scala#diff-d7d8702101580e51566b39b45c43b6e14fa4aa87263b27468521ae2d0499d025R146)).

driver-scala/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
     testImplementation project(':bson').sourceSets.test.output
     testImplementation project(':driver-sync').sourceSets.test.output
     testImplementation project(':driver-core').sourceSets.test.output
+    testImplementation project(':driver-reactive-streams').sourceSets.test.output
```
rozza, line 29: I'm not sure this is worth it.  All it does is couple the sleep static variable in the reactive streams version of SyncMongoClient to the scala version.

Non of the other code is coupled to it. So I think this coupling should be removed.

rozza (changes requested):

driver-scala/build.gradle
rozza: On reflection we already couple to driver-sync - so no harm

rozza (approved):
LGTM

stIncMale:
@rozza Thank you for all the help with the Scala code! Took me some time to make it work.

PR #723, opened 2021-06-03
Support creation of time-series collections

jyemin:
JAVA-3997

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
+++ b/driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
@@ -19,12 +19,15 @@ package com.mongodb.client.model;
 import com.mongodb.lang.Nullable;
 import org.bson.conversions.Bson;
 
+import java.time.Duration;
+
 import static com.mongodb.assertions.Assertions.notNull;
 
 /**
  * Options for creating a collection
  *
  * @mongodb.driver.manual reference/method/db.createCollection/ Create Collection
+ * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
  * @since 3.0
  */
 public class CreateCollectionOptions {
@@ -35,6 +38,8 @@ public class CreateCollectionOptions {
     private IndexOptionDefaults indexOptionDefaults = new IndexOptionDefaults();
     private ValidationOptions validationOptions = new ValidationOptions();
     private Collation collation;
+    private Duration expireAfter;
+    private TimeSeriesOptions timeSeriesOptions;
 
     /**
      * Gets the maximum number of documents allowed in a capped collection.
@@ -194,6 +199,63 @@ public class CreateCollectionOptions {
         return this;
     }
 
+    /**
+     * Returns the expire-after option.
+     *
+     * @return the expire-after option, which may be null.
+     * @since 4.3
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     */
+    @Nullable
+    public Duration getExpireAfter() {
+        return expireAfter;
+    }
+
+    /**
+     * Sets the expire-after option.
+     *
+     * <p>
+     * A duration indicating after how long old time-series data should be deleted. Partial seconds are ignored.
+     * </p>
+     * <p>
+     * Currently applies only to time-series collections, so if this value is set then so must the time-series options
+     * </p>
+     * @param expireAfter the expire-after duration
+     * @return this
+     * @since 4.3
+     * @see #timeSeriesOptions(TimeSeriesOptions)
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     */
+    public CreateCollectionOptions expireAfter(@Nullable final Duration expireAfter) {
+        this.expireAfter = expireAfter;
+        return this;
+    }
+
+    /**
+     * Gets the time series collection options.
+     *
+     * @return the options for a time-series collection
+     * @since 4.3
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     */
+    @Nullable
+    public TimeSeriesOptions getTimeSeriesOptions() {
+        return timeSeriesOptions;
+    }
+
+    /**
+     * Sets the time-series collection options.
+     *
+     * @param timeSeriesOptions the time-series options
+     * @return this
+     * @since 4.3
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     */
+    public CreateCollectionOptions timeSeriesOptions(final TimeSeriesOptions timeSeriesOptions) {
+        this.timeSeriesOptions = timeSeriesOptions;
+        return this;
+    }
+
     @Override
     public String toString() {
         return "CreateCollectionOptions{"
@@ -204,6 +266,8 @@ public class CreateCollectionOptions {
                 + ", indexOptionDefaults=" + indexOptionDefaults
                 + ", validationOptions=" + validationOptions
                 + ", collation=" + collation
+                + ", expireAfter=" + expireAfter
+                + ", timeSeriesOptions=" + timeSeriesOptions
                 + '}';
     }
 }
--- a/driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
+++ b/driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
@@ -207,7 +207,6 @@ public class CreateCollectionOptions {
      * @since 4.3
      * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
      */
-    @Nullable
     public long getExpireAfter(final TimeUnit timeUnit) {
         notNull("timeUnit", timeUnit);
         return timeUnit.convert(expireAfterSeconds, TimeUnit.SECONDS);
--- a/driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
+++ b/driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
@@ -19,7 +19,7 @@ package com.mongodb.client.model;
 import com.mongodb.lang.Nullable;
 import org.bson.conversions.Bson;
 
-import java.time.Duration;
+import java.util.concurrent.TimeUnit;
 
 import static com.mongodb.assertions.Assertions.notNull;
 
@@ -38,7 +38,7 @@ public class CreateCollectionOptions {
     private IndexOptionDefaults indexOptionDefaults = new IndexOptionDefaults();
     private ValidationOptions validationOptions = new ValidationOptions();
     private Collation collation;
-    private Duration expireAfter;
+    private long expireAfterSeconds;
     private TimeSeriesOptions timeSeriesOptions;
 
     /**
@@ -200,15 +200,17 @@ public class CreateCollectionOptions {
     }
 
     /**
-     * Returns the expire-after option.
+     * Returns the expire-after option.  The default value is 0, which indicates no expiration.
      *
+     * @param timeUnit the time unit
      * @return the expire-after option, which may be null.
      * @since 4.3
      * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
      */
     @Nullable
-    public Duration getExpireAfter() {
-        return expireAfter;
+    public long getExpireAfter(final TimeUnit timeUnit) {
+        notNull("timeUnit", timeUnit);
+        return timeUnit.convert(expireAfterSeconds, TimeUnit.SECONDS);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
```diff
@@ -194,6 +199,63 @@ public CreateCollectionOptions collation(@Nullable final Collation collation) {
...
+     * <p>
+     * Currently applies only to time-series collections, so if this value is set then so must the time-series options
+     * </p>
+     * @param expireAfter the expire-after duration
+     * @return this
+     * @since 4.3
+     * @see #timeSeriesOptions(TimeSeriesOptions)
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     */
+    public CreateCollectionOptions expireAfter(@Nullable final Duration expireAfter) {
```
jyemin, line 229: Prior to 4.0 past we might have used this signature:

    expireAfter(long expireAfter, TimeUnit timeUnit)

I think `Duration` is better, and now we can use it (since it was introduced in Java 8), but there are downsides

1. Inconsistent with existing API, which uses `TimeUnit`
2. We are dropping the nanoseconds portion of `Duration` (though we could it?)

Also, since the actual field is `expireAfterSeconds`, we could instead do:

    expireAfterSeconds(long seconds)

but that seems unfriendly.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -407,7 +426,34 @@ private BsonDocument getCommand(final ConnectionDescription description) {
...
+            TimeSeriesGranularity granularity = timeSeriesOptions.getGranularity();
+            if (granularity != null) {
+                timeSeriesDocument.put("granularity", new BsonString(getGranularityAsString(granularity)));
+            }
+            document.put("timeseries", timeSeriesDocument);
+        }
         return document;
     }
 
+    private String getGranularityAsString(final TimeSeriesGranularity granularity) {
```
jyemin, line 447: We could push this code into the enum, but generally (though there are exceptions) we have preferred to keep the mapping from public API to command protocol out of the public API)

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
stIncMale: Using `Duration` is great. I wish we decided to start using `Optional` in public APIs.

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
```diff
@@ -0,0 +1,108 @@
...
+import static com.mongodb.assertions.Assertions.notNull;
+
+/**
+ * Options related to the creation of time-series collections.
+ *
+ * @since 4.3
+ * @see CreateCollectionOptions
+ * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+ */
+public class TimeSeriesOptions {
```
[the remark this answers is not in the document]
jyemin: No.  We've been inconsistent about this, but most of the options classes are final (though not CreateCollectionOptions).

Done

driver-core/src/main/com/mongodb/client/model/TimeSeriesGranularity.java
```diff
@@ -0,0 +1,46 @@
...
+ * </p>
+ * @since 4.3
+ * @see TimeSeriesOptions
+ * @see CreateCollectionOptions
+ */
+public enum TimeSeriesGranularity {
+    /**
+     * Seconds-level granularity.
+     * <p>
+     * If granularity of a time-series collection is unspecified, this is the default value.
```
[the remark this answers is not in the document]
jyemin: I duplicated it in `TimeSeriesOptions#setGranularity`

Done

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
```diff
@@ -0,0 +1,108 @@
...
+    public TimeSeriesOptions(final String timeField) {
+        this.timeField = notNull("timeField", timeField);
+    }
+
+    /**
+     * Gets the name of the field holding the time value.
+     *
+     * @return the name of the field holding the time value.
+     */
+    public String getTimeField() {
```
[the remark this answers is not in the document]
jyemin: We generally have not thought this was important to do in other options/settings classes, primarily because there is rarely a use case for the application to actually use the getters.  But I don't mind, I guess.

Done

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -407,7 +426,34 @@ private BsonDocument getCommand(final ConnectionDescription description) {
...
+        switch (granularity) {
+            case SECONDS:
+                return "seconds";
+            case MINUTES:
+                return "minutes";
+            case HOURS:
+                return "hours";
+            default:
+                throw new MongoInternalException("Unexpected granularity " + granularity);
```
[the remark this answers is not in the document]
jyemin: Yes, though can't use `Assertions.fail()` here.

Done.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -341,6 +350,16 @@ public CreateCollectionOperation collation(final Collation collation) {
         return this;
     }
 
+    public CreateCollectionOperation expireAfter(@Nullable final Duration expireAfter) {
```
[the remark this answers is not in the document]
jyemin: I've thought about this too, and this is a refactoring we could consider in the future.  It makes more sense now that operations are not part of the public API.

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
rozza: Also why isn't this part of `TimeSeriesOptions`?

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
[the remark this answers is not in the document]
rozza: I think an issue was regarding mutability, when the values where copied from the options it effectively makes the operation immutable from the public API.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -341,6 +350,16 @@ public CreateCollectionOperation collation(final Collation collation) {
...
+        this.expireAfter = expireAfter;
+        return this;
+    }
+
+    public CreateCollectionOperation timeSeriesOptions(@Nullable final TimeSeriesOptions timeSeriesOptions) {
```
rozza, line 357: This breaks the previous immutability rule.

rozza (commented):
Left a couple of comments.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
jyemin: You mean about operations?  I don't think we really have immutability of operations.  Consider that we often do shallow copies of BSON documents, which applications can maintain references to.

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
jyemin: The server specification indicates that it's separate.  There is some thought that expireAfter could apply to other types of collections besides time-series in the future.  Jeremy left a [comment](https://jira.mongodb.org/browse/DRIVERS-1438?focusedCommentId=3854462&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3854462) in the associated DRIVERS ticket about it

jyemin: I'm sympathetic to the view that we shouldn't use `Duration` in just one place and am fine using `TimeUnit` instead.  @stIncMale would you object to this decision?

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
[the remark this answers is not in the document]
stIncMale: @rozza If defensive copying was needed, one could have a copy constructor in `CreateCollectionOptions` and assign a copy of a user-supplied `CreateCollectionOptions` to a field `CreateCollectionOperation.options` which is of type `CreateCollectionOptions`.

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
stIncMale: While I understand the API uniformity sentiment<sup>1</sup>, its disadvantage that the global API refactoring may either never happen, or (the best case scenario), may happen in a distant future, thus depriving the API from the benefits of better approaches for longer than it would have been otherwise due to hard restrictions (neither `Duration` nor `Optional` could have been used until the driver required the Java SE 8 at runtime). That being said, I will greet new APIs that use `Optional`/`java.time`, but I will also **not object** to new APIs that use the old approaches.

----

<sup>1</sup> It is especially relevant when mixing `@Nullable` and `Optional`, because observing `Optional` in some places may give a false expectation that all non-`Optional` return values are not nullable. This is likely one of the the reasons the Java SE API adopted neither `Optional` nor the new `java.time` API.

stIncMale (approved):

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
jyemin: I opted for `long` and `TimeUnit` to match how we did `maxTimeMS`.

jyemin: PTAL, @rozza

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
```diff
@@ -200,15 +200,17 @@ public CreateCollectionOptions collation(@Nullable final Collation collation) {
...
     /**
-     * Returns the expire-after option.
+     * Returns the expire-after option.  The default value is 0, which indicates no expiration.
      *
+     * @param timeUnit the time unit
      * @return the expire-after option, which may be null.
      * @since 4.3
      * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
      */
     @Nullable
```
stIncMale, line 210: THe method is not `@Nullable` anymore.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
```diff
@@ -194,6 +199,72 @@ public CreateCollectionOptions collation(@Nullable final Collation collation) {
...
+    /**
+     * Returns the expire-after option.  The default value is 0, which indicates no expiration.
+     *
+     * @param timeUnit the time unit
+     * @return the expire-after option, which may be null.
+     * @since 4.3
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     */
+    @Nullable
+    public long getExpireAfter(final TimeUnit timeUnit) {
```
stIncMale, line 211: The method is not `@Nullable` anymore.

P.S. My previous comment was placed terribly, moved here.

stIncMale, line 211: Commenting so that GitHub allows me to mark the PR as "reviewed, requested changes" (for some reason previous comment is not enough).

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
stIncMale, line 211: - Commenting so that GitHub allows me to mark the PR as "reviewed, requested changes".
- I cannot remove this comment because then GitHub removes my the "reviewed, requested changes" status. So I will resolve this comment instead. Sorry for the mess, this is so weird.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
jyemin: Whoops!  Done.

(I feel like that should generate a warning of some sort, to mark a method returning a primitive value as Nullable)

stIncMale: IDEA marks the line with a warning, but SpotBugs indeed could have enforced the rule, but it does not have one: https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html

stIncMale (approved):
Approved, but static checks fail, so it seems like one more commit would be needed to fix that.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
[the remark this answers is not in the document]
rozza: Ack

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
rozza: Added JAVA-4191 to track improving the whole API.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
rozza: Agreed

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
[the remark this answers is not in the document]
rozza: Added JAVA-4192 to improve how operations are created.

rozza (approved):

PR #726, opened 2021-06-07
Support let option in aggregation helper

jyemin:
JAVA-4186


driver-core/src/main/com/mongodb/internal/operation/Operations.java
```diff
@@ -220,21 +221,22 @@ public EstimatedDocumentCountOperation estimatedDocumentCount(final EstimatedDoc
                 .batchSize(batchSize)
                 .collation(collation)
                 .hint(hint == null ? null : hint.toBsonDocument(documentClass, codecRegistry))
-                .comment(comment);
-
+                .comment(comment)
+                .let(variables == null ? null : variables.toBsonDocument(documentClass, codecRegistry));
```
rozza, line 225: Use the: `com.mongodb.internal.operation.Operations#toBsonDocument` helper

driver-scala/src/main/scala/org/mongodb/scala/AggregateObservable.scala
```diff
@@ -118,6 +118,33 @@ case class AggregateObservable[TResult](private val wrapped: AggregatePublisher[
     this
   }
 
+  /**
+   * Sets the comment to the aggregation. A null value means no comment is set.
```
rozza, line 122: Copy pasta

No need for `<p>` tags in scaladoc.

rozza (changes requested):
Couple of minor nits - once updated  LGTM

driver-scala/src/main/scala/org/mongodb/scala/AggregateObservable.scala
jyemin: Done

driver-core/src/main/com/mongodb/internal/operation/Operations.java
```diff
@@ -220,21 +221,22 @@ public EstimatedDocumentCountOperation estimatedDocumentCount(final EstimatedDoc
...
+                                                         final Bson variables, final AggregationLevel aggregationLevel) {
         return new AggregateToCollectionOperation(namespace, toBsonDocumentList(pipeline), readConcern, writeConcern, aggregationLevel)
                 .maxTime(maxTimeMS, MILLISECONDS)
                 .allowDiskUse(allowDiskUse)
                 .bypassDocumentValidation(bypassDocumentValidation)
                 .collation(collation)
                 .hint(hint == null ? null : hint.toBsonDocument(documentClass, codecRegistry))
-                .comment(comment);
+                .comment(comment)
+                .let(variables == null ? null : variables.toBsonDocument(documentClass, codecRegistry));
```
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/operation/Operations.java
jyemin: Done

rozza (approved):

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/AggregatePublisherImpl.java
```diff
@@ -110,6 +111,12 @@
         return this;
     }
 
+    @Override
+    public AggregatePublisher<T> let(@Nullable final Bson variables) {
```
stIncMale, line 115: [optional]
Here and in `AggregateIterableImpl.java` we specify `@Nullable` despite this class not being a part of the public API. Should we do the same in other non-public places where the `let` method was added?

stIncMale (approved):

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/AggregatePublisherImpl.java
jyemin: I do it here to match the public API that is being overridden.  I'm less inclined to do the same in driver-core, like AggregateOperation, esp. considering it's not done for any other properties.

PR #732, opened 2021-06-15
Increase `waitQueueTimeoutMS` in some tests to reduce the probability of timing out accidentally

stIncMale:
JAVA-4194


jyemin (approved):
LGTM.  I thought they were 50 already.

PR #733, opened 2021-06-18
Add the primitive type 'void'

robxyy:
Ref: https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Class.java#L765

```java
class Model {
    public static void getExpiredAt$annotations() {
    }
}
```


rozza:
Hi @robxyy,

Thanks for the PR - I'm trying to determine the impact of the change (or not having the change). It would also be great to add a regression test - which I'm happy to do once I understand more about the issue.

Ross

robxyy:
Hi @rozza, I added a regression test. When I used kotlinx.serialization for serialization, the `static void getXXX$annotations()` method was automatically generated, and the corresponding `codec` could not be obtained, so I added the primitive type `void`.

rozza:
Hi @robxyy 

I think the underlying issue of this PR was fixed in [JAVA-4119](https://jira.mongodb.org/browse/JAVA-4119)  3c0f582. That said I added [JAVA-4209](https://jira.mongodb.org/browse/JAVA-4209) and used your initial code to extend the other work in JAVA-4119 see: 56e8e1b

Many thanks for your contribution to the driver!

Ross

PR #734, opened 2021-06-19
Add support for setting `io.netty.handler.ssl.SslContext`

stIncMale:
This may be used as a convenient way to utilize OpenSSL as an alternative to the TLS/SSL protocol implementation in a JDK.

JAVA-4177

### DIFF
--- a/docs/reference/content/driver-scala/tutorials/ssl.md
+++ b/docs/reference/content/driver-scala/tutorials/ssl.md
@@ -10,8 +10,9 @@ title = "TLS/SSL"
 
 ## TLS/SSL
 
-The Java driver supports TLS/SSL connections to MongoDB servers using
-the underlying support for TLS/SSL provided by the JDK. 
+By default the Java driver supports TLS/SSL connections to MongoDB servers using
+the underlying support for TLS/SSL provided by the JDK. This can be changed either by utilizing extensibility
+of the [Java SE API]({{< javaseref "api">}}), or via the [Netty API]({{< nettyapiref >}}).
 You can configure the driver to use TLS/SSL either with [`ConnectionString`]({{< apiref "mongo-scala-driver" "org/mongodb/scala/package$$ConnectionString$.html" >}}) or with
 [`MongoClientSettings`]({{< apiref "mongo-scala-driver" "org/mongodb/scala/MongoClientSettings$.html" >}}).
 
@@ -42,7 +43,7 @@ val settings = MongoClientSettings.builder()
 val client = MongoClients.create(settings)
 ```
 
-### Specify `SSLContext` via `MongoClientSettings`
+### Specify Java SE `SSLContext` via `MongoClientSettings`
 
 ```scala
 import javax.net.ssl.SSLContext
@@ -58,6 +59,33 @@ val settings = MongoClientSettings.builder()
         builder.enabled(true)
         builder.context(sslContext)
     })
+    .build()
+val client = MongoClients.create(settings)
+```
+
+### Specify Netty `SslContext` via `NettyStreamFactoryFactory`
+
+If you use the driver with [Netty](https://netty.io/) for network IO,
+you have an option to plug an alternative TLS/SSL protocol implementation provided by Netty.
+
+```java
+import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.ssl.SslProvider;
+```
+
+To instruct the driver to use [`io.netty.handler.ssl.SslContext`]({{< nettyapiref "io/netty/handler/ssl/SslContext.html" >}}),
+use the method
+[`NettyStreamFactoryFactory.Builder.applyToNettySslContext`]({{< apiref "mongodb-driver-core" "com/mongodb/connection/netty/NettyStreamFactoryFactory.Builder.html#applyToNettySslContext(java.util.function.Consumer)" >}}).
+See the documentation of this method for details on which
+[`io.netty.handler.ssl.SslProvider`]({{< nettyapiref "io/netty/handler/ssl/SslProvider.html" >}})s are supported by the driver
+and implications of using them.
+
+```scala
+val settings = MongoClientSettings.builder()
+    .applyToSslSettings((builder: SslSettings.Builder) => builder.enabled(true))
+    .streamFactoryFactory(NettyStreamFactoryFactory.builder()
+        .applyToNettySslContext((builder: SslContextBuilder) => builder.sslProvider(SslProvider.OPENSSL))
+        .build())
     .build()
 val client = MongoClients.create(settings)
 ```
@@ -81,7 +109,10 @@ val settings = MongoClientSettings.builder()
 ```
 
 ## Common TLS/SSL Configuration Tasks
-<p></p>
+
+This section is based on the documentation for [Oracle JDK](https://www.oracle.com/java/technologies/javase-downloads.html#JDK8),
+so some parts may be inapplicable to your JDK or to the custom TLS/SSL implementation you use.
+
 ### Configure Trust Store and Key Store
 One may either configure trust stores and key stores specific to the client via 
 [`javax.net.ssl.SSLContext.init(KeyManager[] km, TrustManager[] tm, SecureRandom random)`]
--- a/docs/reference/content/driver-scala/tutorials/ssl.md
+++ b/docs/reference/content/driver-scala/tutorials/ssl.md
@@ -75,19 +75,22 @@ import io.netty.handler.ssl.SslProvider;
 
 To instruct the driver to use [`io.netty.handler.ssl.SslContext`]({{< nettyapiref "io/netty/handler/ssl/SslContext.html" >}}),
 use the method
-[`NettyStreamFactoryFactory.Builder.applyToNettySslContext`]({{< apiref "mongodb-driver-core" "com/mongodb/connection/netty/NettyStreamFactoryFactory.Builder.html#applyToNettySslContext(java.util.function.Consumer)" >}}).
+[`NettyStreamFactoryFactory.Builder.nettySslContext`]({{< apiref "mongodb-driver-core" "com/mongodb/connection/netty/NettyStreamFactoryFactory.Builder.html#nettySslContext(io.netty.handler.ssl.SslContext)" >}}).
 See the documentation of this method for details on which
 [`io.netty.handler.ssl.SslProvider`]({{< nettyapiref "io/netty/handler/ssl/SslProvider.html" >}})s are supported by the driver
 and implications of using them.
 
 ```scala
+val nettySslContext = SslContextBuilder.forClient()
+    .sslProvider(SslProvider.OPENSSL)
+    .build();
 val settings = MongoClientSettings.builder()
     .applyToSslSettings((builder: SslSettings.Builder) => builder.enabled(true))
     .streamFactoryFactory(NettyStreamFactoryFactory.builder()
-        .applyToNettySslContext((builder: SslContextBuilder) => builder.sslProvider(SslProvider.OPENSSL))
+        .nettySslContext(nettySslContext)
         .build())
     .build()
-val client = MongoClients.create(settings)
+val client = MongoClient(settings)
 ```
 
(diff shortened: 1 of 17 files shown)


driver-core/src/main/com/mongodb/connection/netty/NettyStreamFactoryFactory.java
```diff
@@ -99,6 +110,42 @@ public Builder eventLoopGroup(final EventLoopGroup eventLoopGroup) {
...
+         * Sets the tuner for a {@linkplain SslContextBuilder#forClient() client-side} {@link SslContext io.netty.handler.ssl.SslContext},
+         * which overrides the standard {@link SslSettings#getContext()}.
+         * By default the tuner is {@code null} and {@link SslSettings#getContext()} is at play.
+         * <p>
+         * This option may be used as a convenient way to utilize
+         * <a href="https://www.openssl.org/">OpenSSL</a> as an alternative to the TLS/SSL protocol implementation in a JDK.
+         * To achieve this, specify {@link SslProvider#OPENSSL}
+         * {@linkplain SslContextBuilder#sslProvider(SslProvider) TLS/SSL protocol provider} via the tuner.
+         * Note that doing so adds a runtime dependency on
+         * <a href="https://netty.io/wiki/forked-tomcat-native.html">netty-tcnative</a>, which you must satisfy.
```
stIncMale, line 123: There are multiple artifacts that may be used to satisfy this dependency. They differ by the target OS, ISA, native dependencies which are linked either dynamically or statically. As a result, I don't think we can specify this dependency in the [published POM](https://search.maven.org/artifact/org.mongodb/mongodb-driver-core/4.3.0-beta3/jar) as an optional runtime dependency.

driver-core/src/main/com/mongodb/connection/netty/NettyStreamFactoryFactory.java
```diff
@@ -99,6 +110,42 @@ public Builder eventLoopGroup(final EventLoopGroup eventLoopGroup) {
...
+         * </ul>
+         *
+         * @param nettySslContextTuner The tuner for an {@link SslContext}.
+         *                             The {@link SslContextBuilder} {@linkplain Consumer#accept(Object) supplied} to the tuner
+         *                             is created via {@linkplain SslContextBuilder#forClient()}.
+         * @return {@code this}.
+         *
+         * @since 4.3
+         */
+        public Builder applyToNettySslContext(final Consumer<? super SslContextBuilder> nettySslContextTuner) {
```
stIncMale, line 144: We accept `Consumer<? super SslContextBuilder>` instead of accepting `SslContext` (as we do for the Java SE `SSLContext`) because there are server- and client-side implementations, and it seems better to create a builder for a client-side `SslContext` and give it to a user than to say in the docs that a user must start creating a context via `SslContextBuilder.forClient()`. We can discuss this and change if needed.

driver-core/build.gradle
```diff
@@ -30,18 +30,26 @@ ext {
...
+}
+
 dependencies {
     api project(path: ':bson', configuration: 'default')
 
     implementation "com.github.jnr:jnr-unixsocket:$jnrUnixsocketVersion", optional
     api "io.netty:netty-buffer:$nettyVersion", optional
     api "io.netty:netty-transport:$nettyVersion", optional
-    implementation "io.netty:netty-handler:$nettyVersion", optional
+    api "io.netty:netty-handler:$nettyVersion", optional
```
stIncMale, line 46: This changes the dependency in the [published POM](https://search.maven.org/artifact/org.mongodb/mongodb-driver-core/4.3.0-beta3/jar) from optional runtime to optional compile time.

docs/reference/content/driver-scala/tutorials/ssl.md
```diff
@@ -62,6 +63,33 @@ val settings = MongoClientSettings.builder()
...
+
+```scala
+val settings = MongoClientSettings.builder()
+    .applyToSslSettings((builder: SslSettings.Builder) => builder.enabled(true))
+    .streamFactoryFactory(NettyStreamFactoryFactory.builder()
+        .applyToNettySslContext((builder: SslContextBuilder) => builder.sslProvider(SslProvider.OPENSSL))
+        .build())
+    .build()
+val client = MongoClients.create(settings)
+```
```
stIncMale, line 91: @rozza Could you please pay attention to my monkey-typing the Scala code here (I did it based on the existing code examples)?

driver-core/src/main/com/mongodb/connection/netty/NettyStream.java
```diff
@@ -191,15 +198,7 @@ private void initializeChannel(final AsyncCompletionHandler<Void> handler, final
...
                     if (sslSettings.isEnabled()) {
-                        SSLEngine engine = getSslContext().createSSLEngine(address.getHost(), address.getPort());
-                        engine.setUseClientMode(true);
-                        SSLParameters sslParameters = engine.getSSLParameters();
-                        enableSni(address.getHost(), sslParameters);
-                        if (!sslSettings.isInvalidHostNameAllowed()) {
-                            enableHostNameVerification(sslParameters);
-                        }
-                        engine.setSSLParameters(sslParameters);
-                        pipeline.addFirst("ssl", new SslHandler(engine, false));
```
stIncMale, line 202: This code was moved to `addSslHandler` almost as it was. Thee only modification is with regard to how we get `SSLEngine`.

driver-core/src/main/com/mongodb/connection/netty/NettyStreamFactoryFactory.java
rozza: Personally I'd rather keep both API's the similar and just provide the SslContext.  The documentation could highlight that it should be a client based context.  It seems that configuring this code is in the weeds already and advanced usage. That said you're better placed to determine the user pain of such an approach.

So, if you do wish to keep this more ergonomic approach then I suggest its updated to:

```
applyToSslContext(final Block<? super SslContextBuilder> sslContextBlock)
```

The `Netty` part is inferred from the type signature and the use of `Block` rather than `Consumer` is to follow suite with the rest of the configuration settings.

rozza (changes requested):
A couple of comments

driver-core/src/main/com/mongodb/connection/netty/NettyStreamFactoryFactory.java
stIncMale: I was on the fence. Since you prefer accepting `SslContext`, I changed the API.

Done.

docs/reference/content/driver-scala/tutorials/ssl.md
stIncMale: Thank you!

Done.

driver-core/src/main/com/mongodb/connection/netty/NettyStreamFactoryFactory.java
```diff
@@ -99,6 +111,43 @@ public Builder eventLoopGroup(final EventLoopGroup eventLoopGroup) {
...
+         *
+         * @param nettySslContext The Netty {@link SslContext}, which must be created via {@linkplain SslContextBuilder#forClient()}.
+         * @return {@code this}.
+         *
+         * @since 4.3
+         */
+        public Builder nettySslContext(final SslContext nettySslContext) {
+            this.nettySslContext = notNull("nettySslContext", nettySslContext);
+            isTrueArgument("nettySslContext must be client-side", nettySslContext.isClient());
+            isTrueArgument("nettySslContext must not be reference counted",
```
rozza, line 145: Should this requirement be in the api docs?

driver-core/src/main/com/mongodb/connection/netty/NettyStreamFactoryFactory.java
```diff
@@ -99,6 +111,43 @@ public Builder eventLoopGroup(final EventLoopGroup eventLoopGroup) {
...
+         *    <li>Only {@link SslProvider#JDK} and {@link SslProvider#OPENSSL} TLS/SSL protocol providers are supported.
+         *    </li>
+         * </ul>
+         *
+         * @param nettySslContext The Netty {@link SslContext}, which must be created via {@linkplain SslContextBuilder#forClient()}.
+         * @return {@code this}.
+         *
+         * @since 4.3
+         */
+        public Builder nettySslContext(final SslContext nettySslContext) {
```
rozza, line 142: Naming convention question: Why `nettySslContext` and not just `SslContext`?

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
```diff
@@ -396,32 +400,41 @@ public static StreamFactory getStreamFactory() {
     }
 
     public static StreamFactory getAsyncStreamFactory() {
-        String streamType = System.getProperty("org.mongodb.test.async.type", "nio2");
-
-        if (streamType.equals("netty")) {
-            return new NettyStreamFactory(getSocketSettings(), getSslSettings());
-        } else if (streamType.equals("nio2")) {
+        StreamFactoryFactory overriddenStreamMetafactory = getOverriddenStreamFactoryFactory();
```
rozza, line 403: Naming convention question: why `Metafactory` and not `FactoryFactory`?

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
```diff
@@ -396,32 +400,41 @@ public static StreamFactory getStreamFactory() {
...
 
     @Nullable
     public static StreamFactoryFactory getOverriddenStreamFactoryFactory() {
         String streamType = System.getProperty("org.mongodb.test.async.type", "nio2");
 
-        if (streamType.equals("netty")) {
-            if (nettyStreamFactoryFactory == null) {
-                nettyStreamFactoryFactory = NettyStreamFactoryFactory.builder().build();
+        if (nettyStreamFactoryFactory == null) {
+            if (streamType.equals("netty")) {
```
rozza, line 420: this double if statement could be joined.

rozza (commented):
Looks really good, just a couple of questions for my edification.

driver-core/src/main/com/mongodb/connection/netty/NettyStreamFactoryFactory.java
stIncMale: It is there, formulated as _"Only {@link SslProvider#JDK} and {@link SslProvider#OPENSSL} TLS/SSL protocol providers are supported."_ The constraint is implemented as is because checking the provided context against a blacklist is shorter and saves us from relying on specific classes implementing allowed values of `SslProvider` (if Netty starts implementing a forbidden value with a different class - not a big deal, our check becomes useless; but if Netty starts implementing an allowed value with a different class - our users will not be able to use it).

stIncMale: I changed the error message for this check to _"sslContext must use either SslProvider.JDK or SslProvider.OPENSSL TLS/SSL protocol provider"_.

Done.

driver-core/src/main/com/mongodb/connection/netty/NettyStreamFactoryFactory.java
stIncMale: I wanted to emphasize that Netty `SslContext` is not of type `javax.net.ssl.SSLContext`, however I was not sure whether it is a good idea to make names longer just for that purpose. I interpret your question as a sign that it was not a good idea.

Removed the `netty` prefix.

Done.

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
stIncMale: Thought I would use "metafactory" as this is what should have been used instead of "factoryFactory". Replaced with "factoryFactory".

Done.

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
stIncMale: Thanks! Done.

rozza (approved):
LGTM

PR #737, opened 2021-06-28
Snapshot read support

jyemin:
JAVA-4203
JAVA-4213


jyemin:
Putting this back into draft mode: there's a new requirement (to not allow writes in snapshot sessions) that is forcing a change in implementation strategy.

jyemin:
Superceded.

PR #738, opened 2021-06-29
Add builder for $count accumulator for $group stage

jyemin:
JAVA-4026

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/Accumulators.java
+++ b/driver-core/src/main/com/mongodb/client/model/Accumulators.java
@@ -37,6 +37,23 @@ import static java.util.stream.Collectors.toList;
 public final class Accumulators {
 
     /**
+     * Gets a field name for a $group operation representing the number of documents in the group.
+     *
+     * <p>
+     * {@code Accumulators.count("count")} has equivalent semantics to {@code Accumulators.sum("count", 1)}.
+     * </p>
+     *
+     * @param fieldName the field name
+     * @return the field
+     * @since 4.3
+     * @mongodb.server.release 5.0
+     * @mongodb.driver.manual reference/operator/aggregation/count/ $count
+     */
+    public static BsonField count(final String fieldName) {
+        return accumulatorOperator("$count", fieldName, new BsonDocument());
+    }
+
+    /**
      * Gets a field name for a $group operation representing the sum of the values of the given expression when applied to all members of
      * the group.
      *
--- a/driver-core/src/main/com/mongodb/client/model/Accumulators.java
+++ b/driver-core/src/main/com/mongodb/client/model/Accumulators.java
@@ -47,7 +47,7 @@ public final class Accumulators {
      * @return the field
      * @since 4.3
      * @mongodb.server.release 5.0
-     * @mongodb.driver.manual reference/operator/aggregation/count/ $count
+     * @mongodb.driver.manual reference/operator/aggregation/count-accum/  $count
      */
     public static BsonField count(final String fieldName) {
         return accumulatorOperator("$count", fieldName, new BsonDocument());
--- a/driver-core/src/test/functional/com/mongodb/client/model/AggregatesFunctionalSpecification.groovy
+++ b/driver-core/src/test/functional/com/mongodb/client/model/AggregatesFunctionalSpecification.groovy
@@ -174,6 +174,8 @@ class AggregatesFunctionalSpecification extends OperationFunctionalSpecification
 
         aggregate([group('$z')]).containsAll([new Document('_id', true), new Document('_id', false)])
 
+        aggregate([group(null, Accumulators.count('acc'))]) == [new Document('_id', null).append('acc', 3)]
+
         aggregate([group(null, sum('acc', '$x'))]) == [new Document('_id', null).append('acc', 6)]
 
         aggregate([group(null, avg('acc', '$x'))]) == [new Document('_id', null).append('acc', 2)]
--- a/driver-core/src/test/functional/com/mongodb/client/model/AggregatesFunctionalSpecification.groovy
+++ b/driver-core/src/test/functional/com/mongodb/client/model/AggregatesFunctionalSpecification.groovy
@@ -174,8 +174,6 @@ class AggregatesFunctionalSpecification extends OperationFunctionalSpecification
 
         aggregate([group('$z')]).containsAll([new Document('_id', true), new Document('_id', false)])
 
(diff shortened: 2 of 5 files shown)


driver-core/src/test/functional/com/mongodb/client/model/AggregatesFunctionalSpecification.groovy
```diff
@@ -174,6 +174,8 @@ class AggregatesFunctionalSpecification extends OperationFunctionalSpecification
 
         aggregate([group('$z')]).containsAll([new Document('_id', true), new Document('_id', false)])
 
+        aggregate([group(null, Accumulators.count('acc'))]) == [new Document('_id', null).append('acc', 3)]
```
jyemin, line 177: Note that because there is already a `Aggregates.count(String)` method, we (and our users) can't just say `count('acc`)` here like we can with the other accumulators.  This naming conflict is unfortunate.

driver-core/src/main/com/mongodb/client/model/Accumulators.java
```diff
@@ -36,6 +36,23 @@
...
+     *
+     * <p>
+     * {@code Accumulators.count("count")} has equivalent semantics to {@code Accumulators.sum("count", 1)}.
+     * </p>
+     *
+     * @param fieldName the field name
+     * @return the field
+     * @since 4.3
+     * @mongodb.server.release 5.0
+     * @mongodb.driver.manual reference/operator/aggregation/count/ $count
```
rozza, line 50: This link is the same as `Aggregates.count` which doesn't reference $group at all? And says server version 3.4 (which Aggregates.count also says). Do you know if this will be updated for 5.0 ?
(https://docs.mongodb.com/v5.0/reference/operator/aggregation/count/)

Its a shame and becomes confusing when the same operator can be used in multiple stages which require different syntax.

For example: $sum can now be used in $group, $project, $addFields, $set, $replaceRoot, $replaceWith and $match when used with $expr - not sure if it always has the same syntax.

rozza: I suggest adding the json form of what this count produces, top help highlight why its different to `Aggregates.count`:

rozza (changes requested):
One suggestion for the API docs to help users differentiate between the two different $count methods and the differing syntax.

driver-core/src/main/com/mongodb/client/model/Accumulators.java
jyemin: I don't know if it will be updated.  I can open a ticket though.  We could link to https://docs.mongodb.com/v5.0/reference/operator/aggregation/count-accum/#mongodb-group-grp.-count, but from what I can see $count is the only accumulator that has a URI of this form.

rozza: Good spot - yeah linking to https://docs.mongodb.com/v5.0/reference/operator/aggregation/count-accum/ makes sense

jyemin: Updated docs links.

jyemin: > I suggest adding the json form of what this count produces, top help highlight why its different to Aggregates.count

I considered that but then it would be the only builder method that had such an example.  I think the docs link that shows it should be sufficient.

rozza: Ack

rozza (approved):
LGTM

jyemin:
> Question for reviewers. Is the naming conflict with Aggregates.count(String) sufficient reason not to add the new Accumulators.count(String) method. Is there an alternative we could consider?

What if, instead of adding `Accumulators.count()`, we overload `Accumulators.sum()` with a method that takes only a single argument (the field name)? That would avoid the breakage in source compatibility.

rozza:
> What if, instead of adding Accumulators.count, we overload Accumulators.sum with a method that takes only a single argument (the field name)? That would avoid the breakage in source compatibility.

That just swaps one problem for another as its not discoverable for users.

jyemin:
I agree it's a swap, but it may be a favorable one.  I'm still concerned that anyone following our reference docs and statically importing both Aggregates and Accumulators will hit this compilation error on upgrade and be very confused by it.  I know we don't guarantee source compatibility, only binary compatibility, but we do try pretty hard not to break source.

rozza:
> Another option is to just not do this at all.
I'd be in favour of that over renaming the method such that it becomes undiscoverable.  

Notice docs had to differentiate it by using `count-accum` so perhaps an alternative would be `countAccum` ? I'm more likely to find that method when looking for a count helper.

rozza:
I wouldn't find `sum(1)` when looking for $count - so it might as well not be there.

stIncMale:
> I wouldn't find `sum(1)` when looking for $count - so it might as well not be there.

It can be located be searching the javadoc page for "$count" (because of the `@mongodb.driver.manual` link), but otherwise is, indeed, undiscoverable.

I would not attempt to preserve source backwards compatibility at the cost of messing up method names for those who use `*` imports: using them is generally a bad idea, and everyone using them should be aware of the potential risks and endure the consequences when they happen.

stIncMale (approved):

stIncMale:
> Which looks nicer: `countAccum()` or `sum(1)`

We may also consider naming the method `countDocs` or `countDocuments` and maybe even explain in its docs why the method name is not just `count`, if we go this way.

jyemin:
I could live with `countDocuments`.  I could also live with not doing anything.

rozza:
To summize, the options seem to be:

| Solution | pros | cons | worse case 
-------------|--------|--------|----------------
| `count`  | matches the actual operator name | risk blowing up peoples code | Angry users, support cases and code changes needed 
| `sum`  | doesn't break anyone | discoverability | Confused users, tickets to add support 
| nothing | doesn't break anyone | no helper | Confused users, tickets to add support, potentially try to use the `Aggregates.count` method.
| `countDocuments`  | partially matches the actual operator name so has some measure of discoverablilty | Ugly, counter intuitive and a special case | Mildly confused users

rozza (changes requested):
Given that I'm for either doing nothing or adding a  `count<something>` method. 

So I'd be happy with any of: `countDocuments`, `countAccumulator` or  `countInGroup` or nothing.

jyemin:
I vote for nothing, and we can reconsider if any users notice and request support.

PR #740, opened 2021-06-30
Add URI options spec tests for minPoolSize/maxPoolSize of 0 

jyemin:
Two commits:

1. Implement the new spec tests
2. Clean up the numerous warnings in the test class 

JAVA-4160

This PR is pretty low priority, as it's not required for the 4.3 release

### DIFF
--- a/driver-core/src/test/resources/uri-options/connection-pool-options.json
+++ b/driver-core/src/test/resources/uri-options/connection-pool-options.json
@@ -2,13 +2,15 @@
   "tests": [
     {
       "description": "Valid connection pool options are parsed correctly",
-      "uri": "mongodb://example.com/?maxIdleTimeMS=50000",
+      "uri": "mongodb://example.com/?maxIdleTimeMS=50000&maxPoolSize=5&minPoolSize=3",
       "valid": true,
       "warning": false,
       "hosts": null,
       "auth": null,
       "options": {
-        "maxIdleTimeMS": 50000
+        "maxIdleTimeMS": 50000,
+        "maxPoolSize": 5,
+        "minPoolSize": 3
       }
     },
     {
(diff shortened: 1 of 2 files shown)


driver-core/src/test/resources/uri-options/connection-pool-options.json
```diff
@@ -28,6 +30,28 @@
       "hosts": null,
       "auth": null,
       "options": {}
+    },
+    {
+      "description": "maxPoolSize=0 does not error",
+      "uri": "mongodb://example.com/?maxPoolSize=0",
```
stIncMale, line 36: I don't think our `DefaultConnectionPool` interprets `maxPoolSize=0` as "unlimited":

1. `DefaultConnectionPool.createTimeoutException` works incorrectly in this case;
2. `ConcurrentPool` is created with 0 semaphore permits in this case.

There also appears to be a bug in the constructor of the `ConcurrentPool`: it claims _"if < 0 then no limit"_ but instead creates a semaphore with negative initial permits, which means connections must be released before they can be created.

P.S. I also need to take the above into account in https://github.com/mongodb/mongo-java-driver/pull/701.

stIncMale (changes requested):

driver-core/src/test/resources/uri-options/connection-pool-options.json
jyemin: I see.  Can you report bugs on the above behavior?  I don't see a reason not to push this PR as is, since it's not changing any driver behavior.

stIncMale: Done: https://jira.mongodb.org/browse/JAVA-4216

stIncMale (approved):

rozza (approved):

PR #741, opened 2021-06-30
Disable `align_multiline_parameters` in `.editorconfig`

stIncMale:
Not only this setting is ignored by IDEA when the alignment has to change
as a result of parameter/type/method name change,
but also it leaves much less space for a parameter till the end of line.

Generally, having any alignment that is based not on the line start & nesting level,
but on a specific position in the text on the previous line is neither enforceable
nor a good idea, because it results in spending too much time trying to align the code.


jyemin (approved):
LGTM

rozza (approved):

PR #742, opened 2021-07-01
Support snapshot reads

jyemin:
Add support for snapshot reads to ClientSession

JAVA-4203
JAVA-4212
JAVA-4212

### DIFF
--- a/driver-core/src/main/com/mongodb/ClientSessionOptions.java
+++ b/driver-core/src/main/com/mongodb/ClientSessionOptions.java
@@ -210,7 +210,7 @@ public final class ClientSessionOptions {
 
     private ClientSessionOptions(final Builder builder) {
         if (builder.causallyConsistent != null && builder.causallyConsistent && builder.snapshot != null && builder.snapshot) {
-            throw new IllegalArgumentException("TODO"); // TODO
+            throw new IllegalArgumentException("A session can not be both a snapshot and causally consistent");
         }
         this.causallyConsistent = builder.causallyConsistent != null || builder.snapshot == null
                 ? builder.causallyConsistent
--- a/driver-core/src/main/com/mongodb/ClientSessionOptions.java
+++ b/driver-core/src/main/com/mongodb/ClientSessionOptions.java
@@ -21,6 +21,8 @@ import com.mongodb.annotations.NotThreadSafe;
 import com.mongodb.lang.Nullable;
 import com.mongodb.session.ClientSession;
 
+import java.util.Objects;
+
 import static com.mongodb.assertions.Assertions.notNull;
 
 /**
@@ -54,7 +56,7 @@ public final class ClientSessionOptions {
      * Whether read operations using this session should all share the same snapshot.
      *
      * @return whether read operations using this session should all share the same snapshot. A null value indicates to use the global
-     * default, which is currently false.
+     * default, which is false.
      * @since 4.3
      * @mongodb.server.release 5.0
      * @mongodb.driver.manual  reference/read-concern-snapshot/#read-concern-and-atclustertime Snapshot reads
@@ -86,15 +88,14 @@ public final class ClientSessionOptions {
 
         ClientSessionOptions that = (ClientSessionOptions) o;
 
-        if (causallyConsistent != null ? !causallyConsistent.equals(that.causallyConsistent) : that.causallyConsistent != null) {
+        if (!Objects.equals(causallyConsistent, that.causallyConsistent)) {
             return false;
         }
 
-        if (snapshot != null ? !snapshot.equals(that.snapshot) : that.snapshot != null) {
+        if (!Objects.equals(snapshot, that.snapshot)) {
             return false;
         }
-        if (defaultTransactionOptions != null ? !defaultTransactionOptions.equals(that.defaultTransactionOptions)
-                : that.defaultTransactionOptions != null) {
+        if (!Objects.equals(defaultTransactionOptions, that.defaultTransactionOptions)) {
             return false;
         }
 
--- a/driver-core/src/main/com/mongodb/ClientSessionOptions.java
+++ b/driver-core/src/main/com/mongodb/ClientSessionOptions.java
@@ -35,6 +35,7 @@ import static com.mongodb.assertions.Assertions.notNull;
 public final class ClientSessionOptions {
 
     private final Boolean causallyConsistent;
+    private final Boolean snapshot;
     private final TransactionOptions defaultTransactionOptions;
 
     /**
@@ -47,6 +48,20 @@ public final class ClientSessionOptions {
     @Nullable
     public Boolean isCausallyConsistent() {
         return causallyConsistent;
+    }
+
+    /**
+     * Whether read operations using this session should all share the same snapshot.
+     *
+     * @return whether read operations using this session should all share the same snapshot. A null value indicates to use the global
+     * default, which is currently false.
+     * @since 4.3
+     * @mongodb.server.release 5.0
+     * @mongodb.driver.manual  reference/read-concern-snapshot/#read-concern-and-atclustertime Snapshot reads
+     */
+    @Nullable
+    public Boolean isSnapshot() {
+        return snapshot;
     }
 
     /**
@@ -74,6 +89,10 @@ public final class ClientSessionOptions {
         if (causallyConsistent != null ? !causallyConsistent.equals(that.causallyConsistent) : that.causallyConsistent != null) {
             return false;
         }
+
+        if (snapshot != null ? !snapshot.equals(that.snapshot) : that.snapshot != null) {
+            return false;
+        }
         if (defaultTransactionOptions != null ? !defaultTransactionOptions.equals(that.defaultTransactionOptions)
                 : that.defaultTransactionOptions != null) {
             return false;
@@ -85,6 +104,7 @@ public final class ClientSessionOptions {
     @Override
     public int hashCode() {
         int result = causallyConsistent != null ? causallyConsistent.hashCode() : 0;
+        result = 31 * result + (snapshot != null ? snapshot.hashCode() : 0);
         result = 31 * result + (defaultTransactionOptions != null ? defaultTransactionOptions.hashCode() : 0);
         return result;
     }
@@ -93,6 +113,7 @@ public final class ClientSessionOptions {
     public String toString() {
         return "ClientSessionOptions{"
                 + "causallyConsistent=" + causallyConsistent
+                + "snapshot=" + snapshot
                 + ", defaultTransactionOptions=" + defaultTransactionOptions
                 + '}';
     }
@@ -117,6 +138,7 @@ public final class ClientSessionOptions {
         notNull("options", options);
         Builder builder = new Builder();
         builder.causallyConsistent = options.isCausallyConsistent();
+        builder.snapshot = options.isSnapshot();
         builder.defaultTransactionOptions = options.getDefaultTransactionOptions();
         return builder;
     }
@@ -127,6 +149,7 @@ public final class ClientSessionOptions {
     @NotThreadSafe
     public static final class Builder {
         private Boolean causallyConsistent;
+        private Boolean snapshot;
         private TransactionOptions defaultTransactionOptions = TransactionOptions.builder().build();
 
         /**
@@ -138,6 +161,24 @@ public final class ClientSessionOptions {
          */
         public Builder causallyConsistent(final boolean causallyConsistent) {
             this.causallyConsistent = causallyConsistent;
+            return this;
+        }
+
+        /**
+         * Sets whether read operations using the session should share the same snapshot.
+         *
+         * <p>
+         * The default value is unset, in which case the driver will use the global default value, which is currently false.
+         * </p>
+         *
+         * @param snapshot true for snapshot reads, false otherwise
+         * @return this
+         * @since 4.3
+         * @mongodb.server.release 5.0
+         * @mongodb.driver.manual  reference/read-concern-snapshot/#read-concern-and-atclustertime Snapshot reads
+         */
+        public Builder snapshot(final boolean snapshot) {
+            this.snapshot = snapshot;
             return this;
         }
 
@@ -168,7 +209,13 @@ public final class ClientSessionOptions {
     }
 
     private ClientSessionOptions(final Builder builder) {
-        this.causallyConsistent = builder.causallyConsistent;
+        if (builder.causallyConsistent != null && builder.causallyConsistent && builder.snapshot != null && builder.snapshot) {
+            throw new IllegalArgumentException("TODO"); // TODO
+        }
+        this.causallyConsistent = builder.causallyConsistent != null || builder.snapshot == null
+                ? builder.causallyConsistent
+                : Boolean.valueOf(!builder.snapshot);
+        this.snapshot = builder.snapshot;
         this.defaultTransactionOptions = builder.defaultTransactionOptions;
     }
 }
--- a/driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
+++ b/driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
@@ -230,6 +230,7 @@ public class ClientSessionBinding implements AsyncReadWriteBinding {
         }
 
         @Override
+        @Nullable
         public BsonTimestamp getSnapshotTimestamp() {
             return clientSession.getSnapshotTimestamp();
         }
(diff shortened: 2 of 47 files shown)


driver-core/src/main/com/mongodb/ClientSessionOptions.java
```diff
@@ -49,6 +50,20 @@ public Boolean isCausallyConsistent() {
...
     }
 
+    /**
+     * Whether read operations using this session should all share the same snapshot.
+     *
+     * @return whether read operations using this session should all share the same snapshot. A null value indicates to use the global
+     * default, which is currently false.
+     * @since 4.3
+     * @mongodb.server.release 5.0
+     * @mongodb.driver.manual  reference/read-concern-snapshot/#read-concern-and-atclustertime Snapshot reads
```
jyemin, line 60: This is not a great reference.  Waiting on docs team for something better.

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
```diff
@@ -133,6 +134,21 @@ public void advanceClusterTime(final BsonDocument newClusterTime) {
         this.clusterTime = greaterOf(newClusterTime);
     }
 
+    @Override
+    public void setSnapshotTimestamp(final BsonTimestamp snapshotTimestamp) {
+        isTrue("open", !closed);
+        if (snapshotTimestamp == null) {
+            return;
```
jyemin, line 141: Don't overwrite existing value with null, since not all responses contain a snapshot timestamp.

driver-core/src/main/com/mongodb/internal/operation/OperationReadConcernHelper.java
```diff
@@ -23,15 +23,20 @@
...
+                                           final BsonDocument commandDocument) {
         notNull("commandDocument", commandDocument);
         notNull("sessionContext", sessionContext);
 
         if (sessionContext.hasActiveTransaction()) {
             return;
         }
 
-        BsonDocument readConcernDocument = getReadConcernDocument(sessionContext);
+        if (sessionContext.isSnapshot()) {
```
jyemin, line 35: As with transactions (just above), we don't append read concern at this point, instead deferring to lower-level code in CommandMessage.  This is brittle, but it works.

driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java
```diff
@@ -146,6 +161,25 @@ private static BsonValue getField(final BsonReader bsonReader, final String fiel
         return null;
     }
 
+    @SuppressWarnings("SameParameterValue")
+    private static BsonValue getNestedFieldValue(final ResponseBuffers responseBuffers, final String topLevelFieldName,
```
jyemin, line 165: Only handles one level of nesting, but that's all we need for this scenario

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -286,6 +289,8 @@ private BsonDocument getCommandToEncode() {
                 addReadConcernDocument(extraElements, sessionContext);
             }
             extraElements.add(new BsonElement("autocommit", BsonBoolean.FALSE));
+        } else if (sessionContext.isSnapshot()) {
+            addReadConcernDocument(extraElements, sessionContext);
```
jyemin, line 293: Same pattern we use for transactions (just above).  Higher level code defers adding read concern for snapshot reads, allowing it to be appended in a central location (here).

driver-core/src/main/com/mongodb/ClientSessionOptions.java
```diff
@@ -168,7 +209,13 @@ private Builder() {
     }
 
     private ClientSessionOptions(final Builder builder) {
-        this.causallyConsistent = builder.causallyConsistent;
+        if (builder.causallyConsistent != null && builder.causallyConsistent && builder.snapshot != null && builder.snapshot) {
+            throw new IllegalArgumentException("TODO"); // TODO
```
rozza, line 213: What should the error message be and when is this TODO planned?

driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java
```diff
@@ -112,17 +116,28 @@ static BsonDocument getClusterTime(final BsonDocument response) {
...
     static BsonDocument getRecoveryToken(final ResponseBuffers responseBuffers) {
-        return getFieldDocument(responseBuffers, "recoveryToken");
+        BsonValue value = getFieldValue(responseBuffers, "recoveryToken");
+        if (value == null) {
+            return null;
+        }
+        return value.asDocument();
     }
 
-    private static BsonDocument getFieldDocument(final ResponseBuffers responseBuffers, final String fieldName) {
```
rozza, line 119: I'd probably have kept this method just delegated to the getFieldValue method.

driver-core/src/main/com/mongodb/internal/connection/ReadConcernHelper.java
```diff
@@ -35,12 +38,22 @@ public static BsonDocument getReadConcernDocument(final SessionContext sessionCo
             readConcernDocument.append("level", new BsonString(level.getValue()));
         }
 
+        assertFalse(sessionContext.isSnapshot() && sessionContext.isCausallyConsistent());
+        if (sessionContext.isSnapshot() && maxWireVersion < FIVE_DOT_ZERO_WIRE_VERSION) {
+            throw new MongoClientException("Snapshot reads require MongoDB 5.0 or later");
```
rozza, line 43: Add test to ensure this is handled in reactive streams

driver-core/src/main/com/mongodb/internal/operation/OperationReadConcernHelper.java
rozza: Ack

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
rozza: Simplification:
```
 if (snapshotTimestamp != null) {
    this.snapshotTimestamp = snapshotTimestamp;
 }
 ```

**Review by rozza — CHANGES_REQUESTED** [review]

File: driver-core/src/main/com/mongodb/internal/connection/ReadConcernHelper.java (hunk 8 above)
Comment (jyemin) [reply driver-core/src/main/com/mongodb/internal/connection/ReadConcernHelper.java:43]: There is a spec test for this in "snapshot-sessions-not-supported-client-error", and a reactive streams runner for sessions tests.  So I think it's covered already?

File: driver-core/src/main/com/mongodb/ClientSessionOptions.java (hunk 6 above)
Comment (jyemin) [reply driver-core/src/main/com/mongodb/ClientSessionOptions.java]: Sorry, just missed this.  Added error messages are removed TODO.

File: driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java (hunk 7 above)
Comment (jyemin) [reply driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java:119]: Agreed.  I added it back, and also did the same for similar timestamp one above

File: driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java (hunk 2 above)
Comment (jyemin) [reply driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java]: I'm partial to early returns, but in this case not really useful.

Done

File: driver-core/src/main/com/mongodb/internal/connection/ReadConcernHelper.java (hunk 8 above)
Comment (rozza) [reply driver-core/src/main/com/mongodb/internal/connection/ReadConcernHelper.java:43]: I just double checked and it was all good.

**Review by rozza — APPROVED** [review]
LGTM

File: driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java [hunk 9]
```diff
@@ -89,40 +89,48 @@ static MongoException createSpecialWriteConcernException(final ResponseBuffers r
...
-        BsonValue clusterTime = response.get("$clusterTime");
-        if (clusterTime == null) {
-            return null;
+    static BsonTimestamp getSnapshotTimestamp(final ResponseBuffers responseBuffers) {
+        BsonValue atClusterTimeValue = getNestedFieldValue(responseBuffers, "cursor", "atClusterTime");
+        if (atClusterTimeValue == null) {
+            atClusterTimeValue = getFieldValue(responseBuffers, "atClusterTime");
+        }
+        if (atClusterTimeValue != null && atClusterTimeValue.getBsonType() == BsonType.TIMESTAMP) {
+            return atClusterTimeValue.asTimestamp();
```
stIncMale, line 105: I would expect that if `atClusterTimeValue.isTimestamp()` is true, then `atClusterTimeValue.asTimestamp()` can be called safely. If this is the case, maybe we should replace `atClusterTimeValue.getBsonType() == BsonType.TIMESTAMP` with ``atClusterTimeValue.isTimestamp()`?

stIncMale (changes requested):

driver-core/src/main/com/mongodb/ClientSessionOptions.java
```diff
@@ -49,6 +50,20 @@ public Boolean isCausallyConsistent() {
         return causallyConsistent;
     }
 
+    /**
+     * Whether read operations using this session should all share the same snapshot.
+     *
+     * @return whether read operations using this session should all share the same snapshot. A null value indicates to use the global
+     * default, which is currently false.
```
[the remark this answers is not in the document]
jyemin: There is precedent in the drivers for changing the default from one release to another.  In particular, we changed the default for retryWrites from false to true.  But in this case it would change semantics to change the default, so I don't think we would ever change the default.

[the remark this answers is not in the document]
jyemin: We actually have the same language just above for the isCausallyConsistent method.

driver-core/src/main/com/mongodb/ClientSessionOptions.java
```diff
@@ -74,6 +89,10 @@ public boolean equals(final Object o) {
         if (causallyConsistent != null ? !causallyConsistent.equals(that.causallyConsistent) : that.causallyConsistent != null) {
             return false;
         }
+
+        if (snapshot != null ? !snapshot.equals(that.snapshot) : that.snapshot != null) {
+            return false;
```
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
```diff
@@ -217,6 +218,22 @@ public boolean notifyMessageSent() {
...
+            return snapshot != null && snapshot;
+        }
+
+        @Override
+        public void setSnapshotTimestamp(final BsonTimestamp snapshotTimestamp) {
+            clientSession.setSnapshotTimestamp(snapshotTimestamp);
+        }
+
+        @Override
+        public BsonTimestamp getSnapshotTimestamp() {
```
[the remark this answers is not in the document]
jyemin: I did it on all the interface methods, but not on the overrides.  I think this is correct.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -278,6 +279,8 @@ private BsonDocument getCommandToEncode() {
             extraElements.add(new BsonElement("lsid", sessionContext.getSessionId()));
         }
         boolean firstMessageInTransaction = sessionContext.notifyMessageSent();
+
+            assertFalse(sessionContext.hasActiveTransaction() && sessionContext.isSnapshot());
```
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java
jyemin: Done

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
```diff
@@ -133,6 +134,20 @@ public void advanceClusterTime(final BsonDocument newClusterTime) {
         this.clusterTime = greaterOf(newClusterTime);
     }
 
+    @Override
+    public void setSnapshotTimestamp(final BsonTimestamp snapshotTimestamp) {
+        isTrue("open", !closed);
+        if (snapshotTimestamp != null) {
+            this.snapshotTimestamp = snapshotTimestamp;
```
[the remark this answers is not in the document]
jyemin: So we could add an extra sanity check like:

    else if (snapshotTimestamp.equals(this.snapshotTimestamp) { 
         throw new MongoClientException(...) // or maybe an assertEquals
    }

but I don't except it to happen in practice by design of the protocol.

driver-core/src/main/com/mongodb/internal/operation/OperationReadConcernHelper.java
stIncMale: So here we simply want to omit adding `readConcern` to `commandDocument` if either `sessionContext.hasActiveTransaction` or `sessionContext.isSnapshot`. Then I cannot understand why do we add `readConcern` here at all? This is especially unclear given that `CommandMessage.getExtraElements` also adds `readConcern`. Is this because `OperationReadConcernHelper.appendReadConcernToCommand` is called only for messages that are not `OP_MSG` (i.e., for the old wire protocol, before all messages became `OP_MSG`), and `CommandMessage.getExtraElements` is called only `OP_MSG` (i.e., for the new wire protocol)?

P.S. I realize that this question is not about the PR changes anymore, but it would be useful for me to understand the situation.

jyemin: No, it's not about OP_MSG at all.  Probably easier to explain the design over Slack though.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -642,6 +643,7 @@ private ServerAddress getServerAddress() {
     private void updateSessionContext(final SessionContext sessionContext, final ResponseBuffers responseBuffers) {
         sessionContext.advanceOperationTime(getOperationTime(responseBuffers));
         sessionContext.advanceClusterTime(getClusterTime(responseBuffers));
+        sessionContext.setSnapshotTimestamp(getSnapshotTimestamp(responseBuffers));
```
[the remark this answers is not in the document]
stIncMale: Right, I should have realized this.

driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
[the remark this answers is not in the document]
stIncMale: I think it should be added on all methods. Here is what the documentation of `com.mongodb.lang.Nullable` says: _"Methods override should repeat parent `@Nullable annotations unless they behave differently."_

This requirement is consistent with return types and declared checked exceptions in overriden methods: if an overriding method returns a more specific type than its ancestor, or does not throw a checked exception allowed to be thrown by its ancestor, then a program can rely on this more specific behaviors provided that the compiler can detect that the overriding methods is used. It is only natural to expect the same from the `@Nullable` annotation: if an overriding method does not specify it, then, just like with checked exceptions, it does not return `null` values even though the ancestor allows returning them; and the program should be able to rely on this.

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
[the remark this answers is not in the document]
stIncMale: > I don't except it to happen in practice by design of the protocol

I see, but this fact is not obvious from reading the code. While I think we should not have checks whose sole purpose is to verify that the server behaves correctly, the check you mention would primarily serve as a self-documented code stating "even though this method is allowed to be called more than once, it must always be called with the same argument", from which one can infer that the server must not reply with different `snapshotTimestamp` (though is allowed to not return any) to requests belonging with the same session.

So I am in favour of adding such a check, or, at least, documenting the above in the method.

driver-core/src/main/com/mongodb/ClientSessionOptions.java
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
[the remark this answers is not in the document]
jyemin: Check added

driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/operation/OperationReadConcernHelper.java
stIncMale: Jeff, thank you for explaining this matter (and beyond) to me over Zoom in details.

stIncMale (approved):

PR #743, opened 2021-07-01
Support serviceId in reply from load balanced cluster

jyemin:
The current load balancing code doesn't support an actual reply from a load-balanced cluster that contains a serviceId, as required by the specification. This patch adds it. 

JAVA-4218

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/ConnectionDescription.java
+++ b/driver-core/src/main/com/mongodb/connection/ConnectionDescription.java
@@ -25,6 +25,7 @@ import org.bson.types.ObjectId;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 import static com.mongodb.assertions.Assertions.notNull;
 import static com.mongodb.connection.ServerDescription.getDefaultMaxDocumentSize;
@@ -274,6 +275,9 @@ public class ConnectionDescription {
 
         ConnectionDescription that = (ConnectionDescription) o;
 
+        if (maxWireVersion != that.maxWireVersion) {
+            return false;
+        }
         if (maxBatchCount != that.maxBatchCount) {
             return false;
         }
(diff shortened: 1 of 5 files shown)


driver-core/src/main/com/mongodb/connection/ConnectionDescription.java
```diff
@@ -274,6 +275,9 @@ public boolean equals(final Object o) {
 
         ConnectionDescription that = (ConnectionDescription) o;
 
+        if (maxWireVersion != that.maxWireVersion) {
```
jyemin, line 278: Re-auto-generated equals and hashCode since we needed to add serviceId

driver-core/src/main/com/mongodb/internal/connection/DescriptionHelper.java
```diff
@@ -77,14 +77,22 @@ static ConnectionDescription createConnectionDescription(final ClusterConnection
...
-                if (topologyVersion != null) {
-                    connectionDescription = connectionDescription.withServiceId(topologyVersion.getProcessId());
-                }
+            ObjectId serviceId = getServiceId(isMasterResult);
+            if (serviceId != null) {
+                connectionDescription = connectionDescription.withServiceId(serviceId);
             } else {
-                throw new MongoClientException("Driver attempted to initialize in load balancing mode, but the server does not support "
-                        + "this mode");
+                if (manufactureServiceId) {
```
rozza, line 84: Nit: would be cleaner as `} else if (manufactureServiceId) {`

rozza (approved):
One nit then LGTM

driver-core/src/main/com/mongodb/internal/connection/DescriptionHelper.java
jyemin: Yeah, definitely.  There's even an IntelliJ auto-fix for this.

While looking at the code, I also realized I could simplify the manufactureServiceId branch by using an assertion instead of a conditional.  This is appropriate as it's test-only code.  PTAL.

Done.

rozza (approved):
LGTM

DmitryLukyanov (approved):
LGTM :)

jyemin:
Another change I missed.  Found this by testing against an actual load balanced cluster in cloud dev.

driver-core/src/main/com/mongodb/internal/connection/DescriptionHelper.java
```diff
@@ -144,6 +147,11 @@ private static TopologyVersion getTopologyVersion(final BsonDocument isMasterRes
                 ? new TopologyVersion(isMasterResult.getDocument("topologyVersion")) : null;
     }
 
+    private static ObjectId getServiceId(final BsonDocument isMasterResult) {
+        return isMasterResult.containsKey("serviceId") && isMasterResult.get("serviceId").isObjectId()
+                ? isMasterResult.getObjectId("serviceId").getValue() : null;
```
stIncMale, line 152: [optional]

This code does 3 map lookups instead of 1. This does not affect correctness, though.

stIncMale (approved):

PR #744, opened 2021-07-02
Ensure MongoClient#close only actions on first call

rozza:
JAVA-4215

### DIFF
--- a/driver-legacy/src/main/com/mongodb/MongoClient.java
+++ b/driver-legacy/src/main/com/mongodb/MongoClient.java
@@ -119,6 +119,7 @@ public class MongoClient implements Closeable {
     private final ConcurrentLinkedQueue<ServerCursorAndNamespace> orphanedCursors = new ConcurrentLinkedQueue<>();
     private final ExecutorService cursorCleaningService;
     private final MongoClientImpl delegate;
+    private boolean closed;
 
     /**
      * Gets the default codec registry.  It includes the following providers:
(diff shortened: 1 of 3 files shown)


jyemin (approved):
LGTM, but add Valentin to review as well.  He may have reason to prefer an AtomicBoolean over synchronization.

stIncMale:
> He may have reason to prefer an AtomicBoolean over synchronization.

It is likely safe to say that `close` is usually not called often on the same object, and it is definitely not called often on the same object where it doing so would have observable performance effects, so performance-wise it does not matter whether we use `AtomicBoolean` or a `synchronized` method. However, `synchronized` methods (I think, the whole idea of associating a monitor with each object) is a mistake that somehow was accepted and implemented in Java:

- Absolute majority of objects are used neither for exclusive locking nor for conditional variables, so why in the world would one associate a monitor with each of them? Having a monitor conceptually associated with any object either results in higher resource consumption or further complicates the runtime, which now has to create an illusion of having a monitor without actually associating it with an object, unless it is actually used.
- `synchronized` methods use synchronization primitives which are publicly accessible, which is unacceptable (most engineers consider this to be fine for some reason).

stIncMale (approved):
I would implement idempotency via `if (!closed.getAndSet(true)) { /* close */ }`, but unless the language designers start blaming `synchronized`, I doubt it is possible to convince engineers to stop using it.

stIncMale (approved):

rozza:
Thanks @stIncMale for your feedback and explanations regarding synchronized

PR #745, opened 2021-07-02
Fix regression in MongoWriteException error code

jyemin:
Ensure that the error code from the single WriteError is propagated to the
exception error code.

JAVA-4219


rozza (approved):
LGTM

PR #746, opened 2021-07-06
Implemented unified test runner for GridFS

jyemin:
Abstract GridFSBucket creation in test runner so that the reactive implementation can be wrapped
by a new SyncGridFSBucket implementation.

Introduced abstract base classes for sync and reactive unified test runners to reduce code
duplication.

Also, removed legacy GridFS test runner

JAVA-4109


driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncGridFSBucket.java
```diff
@@ -0,0 +1,376 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.mongodb.ClusterFixture.TIMEOUT_DURATION;
+import static java.util.Objects.requireNonNull;
+
+public class SyncGridFSBucket implements GridFSBucket {
```
jyemin, line 49: This is needed because otherwise we are just testing the synchronous GridFSBucket implementation on top of SyncMongoCollection.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncGridFSBucket.java
```diff
@@ -0,0 +1,376 @@
...
+    public void drop() {
+        Mono.from(wrapped.drop()).block(TIMEOUT_DURATION);
+    }
+
+    @Override
+    public void drop(final ClientSession clientSession) {
+        Mono.from(wrapped.drop(unwrap(clientSession))).block(TIMEOUT_DURATION);
+    }
+
+    private void toOutputStream(final GridFSDownloadPublisher downloadPublisher, final OutputStream destination) {
```
jyemin, line 344: These two methods use Project Reactor Flux to go from synchronous to asynchronous world.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncMongoCollection.java
```diff
@@ -673,7 +673,7 @@ public String createIndex(final ClientSession clientSession, final Bson keys, fi
 
     @Override
     public ListIndexesIterable<Document> listIndexes() {
-        throw new UnsupportedOperationException();
+        return listIndexes(Document.class);
```
jyemin, line 676: Needed because GridFS uses it.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/UnifiedReactiveStreamsTest.java
```diff
@@ -0,0 +1,47 @@
...
+import com.mongodb.client.unified.UnifiedTest;
+import com.mongodb.reactivestreams.client.MongoClients;
+import com.mongodb.reactivestreams.client.gridfs.GridFSBuckets;
+import com.mongodb.reactivestreams.client.syncadapter.SyncGridFSBucket;
+import com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient;
+import com.mongodb.reactivestreams.client.syncadapter.SyncMongoDatabase;
+import org.bson.BsonArray;
+import org.bson.BsonDocument;
+
+public abstract class UnifiedReactiveStreamsTest extends UnifiedTest {
```
jyemin, line 32: Extracted this common superclass for all Reactive spec tests.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedSyncTest.java
```diff
@@ -0,0 +1,43 @@
...
+import com.mongodb.MongoClientSettings;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.gridfs.GridFSBucket;
+import com.mongodb.client.gridfs.GridFSBuckets;
+import org.bson.BsonArray;
+import org.bson.BsonDocument;
+
+public abstract class UnifiedSyncTest extends UnifiedTest{
```
jyemin, line 28: Same for unified tests of the sync driver: extracted a common abstract base class before adding the new abstract method to UnifiedTest

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -123,6 +125,8 @@ public Entities getEntities() {
 
     protected abstract MongoClient createMongoClient(MongoClientSettings settings);
 
+    protected abstract GridFSBucket createGridFSBucket(MongoDatabase database);
```
jyemin, line 128: This is the abstraction that makes it possible to create an alternate implementation of GridFSBucket, similar to what we already do for creating a MongoClient.

jyemin (commented):
Added comments.  Feel free to let this one sit for a while.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedGridFSHelper.java
```diff
@@ -117,4 +161,8 @@ public OperationResult executeUpload(final BsonDocument operation) {
             return OperationResult.of(e);
         }
     }
+
+    Document asDocument(final BsonDocument bsonDocument) {
```
stIncMale, line 165: [A question not directly related to the PR changes]

We need this method because `GridFSUploadOptions.metadata` accepts `Document` instead of accepting `Bson`. Was there a reason why accepting `Bson` instead was considered a worse idea, or is this an unnecessary API limitation in retrospect?

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncGridFSBucket.java
```diff
@@ -0,0 +1,376 @@
...
+        Flux.from(downloadPublisher).toStream().forEach(byteBuffer -> {
+            try {
+                byte[] bytes = new byte[byteBuffer.remaining()];
+                byteBuffer.get(bytes);
+                destination.write(bytes);
+            } catch (IOException e) {
+                throw new MongoGridFSException("IOException when reading from the OutputStream", e);
+            }
+        });
```
stIncMale, line 353: [A bug in the implementation of our `downloadToStream` methods?]

The `toOutputStream` method is used to implement `com.mongodb.client.gridfs.GridFSBucket.downloadToStream` methods. Those methods do not specify whether the destination `OutputStream` is flushed or not, but the [example in our documentation](https://docs.mongodb.com/drivers/java/sync/current/fundamentals/gridfs/#download-a-file-to-an-output-stream) clearly does not flush the stream, implying that users do not need to flush it, i.e., it is flushed by `downloadToStream` methods. I looked at `com.mongodb.client.gridfs.GridFSBucketImpl.downloadToStream` and discovered that our real implementation does not flush the stream just like the `SyncGridFSBucket` does not. This is disturbing, because users expect streams to be flushed based on the aforementioned example, which means that depending on the `OutputStream` implementation, a part of the read data may be lost.

@jyemin I would like to know your opinion before filing a bug about our `downloadToStream` methods.

stIncMale (approved):
Just a couple of questions and a nit, but nothing that would prevent merging the PR.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedGridFSHelper.java
jyemin: It seems like an unnecessary limitation.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncGridFSBucket.java
jyemin: I'm not sure it should be the driver's responsibility to flush the stream.  Regarding the documentation, it's true it doesn't flush the stream, but it does close it.  There seems to be a lot of conflicting opinions about whether closing a file outputstream stream is sufficient to also flush it (it would seem weird if it didn't, but it's not explicitly stated in the javadoc).

Anyway, let's discuss in a different channel than this PR.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedSyncTest.java
jyemin: Done

PR #747, opened 2021-07-06
Add builders for window functions

stIncMale:
JAVA-3866


docs/reference/content/builders/aggregation.md
```diff
@@ -10,8 +10,9 @@ title = "Aggregates"
 
 ## Aggregates
 
-The [`Aggregates`]({{< apiref "mongodb-driver-core" "com/mongodb/client/model/Aggregates" >}}) class provides static factory methods that build [aggregation
-pipeline operators]({{< docsref "reference/operator/aggregation/" >}}).  Each method returns an instance of the
+The [`Aggregates`]({{< apiref "mongodb-driver-core" "com/mongodb/client/model/Aggregates" >}}) class provides static factory methods
+that build [aggregation pipeline stages]({{< docsref "meta/aggregation-quick-reference/#stages" >}}).
```
stIncMale, line 14: The link to aggregation pipeline stages was incorrect. Also filed https://jira.mongodb.org/browse/DOCSP-17428 to fix the link in the new Java driver docs.

bson/src/main/org/bson/BsonNumber.java
```diff
@@ -19,7 +19,7 @@
 import org.bson.types.Decimal128;
 
 /**
- * Base class for the three numeric BSON types.  This class mirrors the functionality provided by {@code java.lang.Number}.
+ * Base class for the numeric BSON types. This class mirrors the functionality provided by {@code java.lang.Number}.
```
stIncMale, line 22: Actually, there are four numeric BSON types. I removed the number here because we forgot to update it previously (when `BsonDecimal128` was added) and may forget in the future if more types are added.

driver-core/src/main/com/mongodb/client/model/MongoTimeUnit.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+package com.mongodb.client.model;
+
+/**
+ * Units for specifying time-based bounds for {@linkplain Window windows} and output units for some time-based
+ * {@linkplain WindowedComputation windowed computations}.
+ *
+ * @since 4.3
+ */
+public enum MongoTimeUnit {
```
stIncMale, line 24: `Unit` clashes with `scala.Unit` and is also not descriptive enough, `TimeUnit` clashes with `java.util.concurrent.Unit`. So I settled on `MongoTimeUnit`.

I also considered using `java.time.temporal.ChronoUnit` instead, but the downsides are by far greater than the small upside of reusing an existing class:

1. We would not support all its instances, and would have to document the subset that is supported. This is weird and not user-friendly (the `java.time` API does this, but there it is at least can be explained by the complexity and extensibility of the API).
2. This approach is incompatible with server starting to support time units that do not have `ChronoUnit` counterparts. It is possible to work around this limitation of `ChronoUnit` by using the `TemporalUnit` interface instead, but such an approach gives us no benefits compared to introducing a new `Enum` class.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
```diff
@@ -0,0 +1,772 @@
...
+ * {@link Aggregates#setWindowFields(Object, Bson, List) $setWindowFields} pipeline stage
+ * of an aggregation pipeline. Each windowed computation is a triple:
+ * <ul>
+ *     <li>A window function. Some functions require documents in a window to be sorted
+ *     (see {@code sortBy} in {@link Aggregates#setWindowFields(Object, Bson, List)}).</li>
+ *     <li>An optional {@linkplain Window window}, a.k.a. frame.
+ *     Specifying {@code null} window is equivalent to specifying [{@link Bound#UNBOUNDED}, {@link Bound#UNBOUNDED}].</li>
+ *     <li>A path to an output field to be computed by the window function over the window.</li>
+ * </ul>
+ * A windowed computation is similar to an {@linkplain Accumulators accumulator} but does not result in folding documents constituting
```
stIncMale, line 57: While `BsonField`s representing windowed computations are similar to those produced by the `Accumulators` methods, they are different at least because the field value documents include windows. Thus, we need a new set of builders, which are provided by the `WindowedComputations` class.

docs/reference/content/driver-scala/builders/aggregation.md
```diff
@@ -129,9 +130,9 @@ expression and outputs to the next stage a document for each distinct grouping.
 expression on which to group, and zero or more 
 [accumulators]({{< docsref "reference/operator/aggregation/group/#accumulator-operator" >}}) which are evaluated for each 
 grouping.  To simplify the expression of accumulators, the driver includes an 
-[`Accumulators`]({{< apiref "mongo-scala-driver" "org/mongodb/scala/model/Aggregates$" >}}) class with static factory methods for each of the supported 
-accumulators. In the example below, it's assumed that the `sum` and `avg` methods of the `Accumulators` class have been statically 
-imported. 
+[`Accumulators`]({{< apiref "mongo-scala-driver" "org/mongodb/scala/model/Accumulators$" >}}) singleton object with factory methods 
+for each of the supported accumulators.
+In the example below, it's assumed that the `sum` and `avg` methods of the `Accumulators` class have been statically imported.
```
stIncMale, line 135: - The link was not to the right program element.
- `org.mongodb.scala.model.Accumulators` is not a class but a singleton object.
- Changed line breaks such that lines comply with our `max_line_length = 140` limit.

driver-core/src/test/unit/com/mongodb/client/model/TestWindowedComputations.java
```diff
@@ -0,0 +1,281 @@
...
+    private static final String NO_EXPRESSION = "{}";
+    private static final String PATH = "newField";
+    private static final Map.Entry<Integer, BsonValue> INT_EXPR = new AbstractMap.SimpleImmutableEntry<>(1, new BsonInt32(1));
+    private static final Map.Entry<String, BsonValue> STR_EXPR =
+            new AbstractMap.SimpleImmutableEntry<>("$fieldToRead", new BsonString("$fieldToRead"));
+    private static final Window POSITION_BASED_WINDOW = documents(1, 2);
+    private static final Window RANGE_BASED_WINDOW = range(1, 2);
+
+    @Test
+    void simpleWindowFunctions() {
```
rozza, line 53: Question: Whats the rationale / motivation for these tests? Do we do similar for the other builders or do they just have output style tests as in AggregatesSpecification?

driver-scala/src/main/scala/org/mongodb/scala/model/package.scala
```diff
@@ -768,6 +769,73 @@ package object model {
...
 
   }
+
+  /**
+   * Units for specifying time-based bounds for [[Window windows]] and output units for some time-based
+   * [[WindowedComputation windowed computations]].
+   *
+   * @since 4.3
+   */
+  object MongoTimeUnit {
```
rozza, line 779: :+1: Sometimes Scala makes me sad!

rozza (commented):
Looks really good - only a couple of comments for my edification.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
```diff
@@ -0,0 +1,765 @@
...
+     *
+     * @param path The output field path.
+     * @param expression The expression.
+     * @param window The window.
+     * @param <TExpression> The expression type.
+     * @return The constructed {@link WindowedComputation}.
+     * @mongodb.driver.dochub core/window-functions-sum $sum
+     * @mongodb.server.release 5.0
+     */
+    public static <TExpression> WindowedComputation sum(final String path, final TExpression expression, @Nullable final Window window) {
```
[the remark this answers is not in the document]
stIncMale: In the new documentation example for `$setWindowFields` I used `WindowedComputations.sum` without using static import.

It may be worth pointing out that having the names of the `WindowedComputations` methods clashing with those in the `Accumulators` class does not break compilation for existing code, as it was the case in https://github.com/mongodb/mongo-java-driver/pull/738.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
```diff
@@ -0,0 +1,765 @@
...
+public final class WindowedComputations {
+    /**
+     * Builds a computation of the sum of the evaluation results of the {@code expression} over the {@code window}.
+     *
+     * @param path The output field path.
+     * @param expression The expression.
+     * @param window The window.
+     * @param <TExpression> The expression type.
+     * @return The constructed {@link WindowedComputation}.
+     * @mongodb.driver.dochub core/window-functions-sum $sum
```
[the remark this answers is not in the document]
stIncMale: Yes, see https://jira.mongodb.org/browse/DOCSP-17321.

driver-core/src/test/unit/com/mongodb/client/model/TestWindowedComputations.java
stIncMale: Yes, this class contains unit tests which are conceptually similar to those in `AggregatesSpecification`: we know what the builders are supposed to build, and we check that the results are correct or that building fails for illegal arguments.

Since the signatures of many builder methods differ only in the method name, the unit tests for them were combined to omit repeating the same test logic.

docs/reference/content/whats-new.md
```diff
@@ -28,6 +28,9 @@ New features of the 4.3 Java driver release include:
   [Netty](https://netty.io/) [`io.netty.handler.ssl.SslContext`]({{< nettyapiref "io/netty/handler/ssl/SslContext.html" >}}),
   which may be used as a convenient way to utilize [OpenSSL](https://www.openssl.org/) as an alternative
   to the TLS/SSL protocol implementation in a JDK.
+* Added [builders]({{< apiref "mongodb-driver-core" "com/mongodb/client/model/Aggregates.html#setWindowFields(TExpression,org.bson.conversions.Bson,java.util.List)" >}})
+  for a new [`$setWindowFields`](https://dochub.mongodb.org/core/window-functions-set-window-fields)
```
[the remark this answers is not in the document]
stIncMale: Thanks! Done.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
```diff
@@ -0,0 +1,765 @@
...
+
+/**
+ * Builders for {@linkplain WindowedComputation windowed computations} used in the
+ * {@link Aggregates#setWindowFields(Object, Bson, List) $setWindowFields} pipeline stage
+ * of an aggregation pipeline. Each windowed computation is a triple:
+ * <ul>
+ *     <li>A window function. Some functions require documents in a window to be sorted
+ *     (see {@code sortBy} in {@link Aggregates#setWindowFields(Object, Bson, List)}).</li>
+ *     <li>An optional {@linkplain Window window}, a.k.a. frame.
+ *     Specifying {@code null} window is equivalent to specifying [{@link Bound#UNBOUNDED}, {@link Bound#UNBOUNDED}].</li>
```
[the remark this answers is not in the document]
stIncMale: Yes. However, it is not the same thing as an unbounded interval in math. Even though a window is called unbounded, there is always a bound: `Bound.UNBOUNDED` means the_"window bound is the same as the corresponding bound of the partition encompassing it"_(this is documented on `Bound.UNBOUNDED`) , and all partitions have finite ends.

I changed the wording, please tell if the docs are still confusing.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
```diff
@@ -0,0 +1,767 @@
...
+import static com.mongodb.client.model.MongoTimeUnit.MONTH;
+import static com.mongodb.client.model.MongoTimeUnit.YEAR;
+import static org.bson.assertions.Assertions.notNull;
+
+/**
+ * Builders for {@linkplain WindowedComputation windowed computations} used in the
+ * {@link Aggregates#setWindowFields(Object, Bson, List) $setWindowFields} pipeline stage
+ * of an aggregation pipeline. Each windowed computation is a triple:
+ * <ul>
+ *     <li>A window function. Some functions require documents in a window to be sorted
```
stIncMale, line 44: While the whole world and the internal server docs use the word "window function", the official MongoDB docs refer to it as "[window operator](https://docs-mongodbcom-staging.corp.mongodb.com/docs/docsworker-xlarge/window-functions/reference/operator/aggregation/setWindowFields/#window-operators)". I mentioned this name in the docs of the method `Aggregates.setWindowFields`, but continued using "window function" in other places. Please tell if you think this is wrong and I should replace "function" with "operator" everywhere.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -594,6 +597,61 @@ public static Bson sample(final int size) {
...
+     *               of the output documents.
+     * @param output A nonempty array of {@linkplain WindowedComputation windowed computations}.
+     * @param <TExpression> The {@code partitionBy} expression type.
+     * @return The {@code $setWindowFields} pipeline stage.
+     * @mongodb.driver.dochub core/window-functions-set-window-fields $setWindowFields
+     * @mongodb.server.release 5.0
+     * @since 4.3
+     */
+    @Beta
+    public static <TExpression> Bson setWindowFields(@Nullable final TExpression partitionBy, @Nullable final Bson sortBy,
```
jyemin, line 621: I think we could safely remove this overload.  It's not that helpful, and arguably misleading, since it's required that there be at least one windowed computation.  With group, by contrast, it's not required that there be any accumulators.  And with `List.of` it's quite easy to create a list of one (or more) items.

driver-core/src/main/com/mongodb/client/model/MongoTimeUnit.java
jyemin: I'm happy with introducing our own class, but should we limit it to the window use-case by using the less generic "Window" prefix on the name instead of "Mongo"?  So it could be "WindowTimeUnit" or "WindowUnit"

driver-core/src/main/com/mongodb/client/model/MongoTimeUnit.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Units for specifying time-based bounds for {@linkplain Window windows} and output units for some time-based
+ * {@linkplain WindowedComputation windowed computations}.
+ *
+ * @since 4.3
+ */
+@Beta
+public enum MongoTimeUnit {
+    /**
+     * @mongodb.server.release 5.0
```
jyemin, line 29: Can we just move these annotations to the class level?

driver-core/src/main/com/mongodb/client/model/Window.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * A subset of documents within a partition in the {@link Aggregates#setWindowFields(Object, Bson, List) $setWindowFields} pipeline stage
+ * of an aggregation pipeline (see {@code partitionBy} in {@link Aggregates#setWindowFields(Object, Bson, List)}).
+ *
+ * @see Windows
+ * @since 4.3
+ */
+@Beta
+public interface Window extends Bson {
```
jyemin, line 31: I'm not clear that this interface buys us anything.  It has no methods, so it just seems like a way to protect a user from themselves in a way that may prove inconvenient.  It's also arguably inconsistent with the rest of the builder API, which doesn't enforce many type constraints (for example, the sortBy field in setWindowFields is of type Bson, not of type SortBy)

driver-core/src/main/com/mongodb/client/model/WindowedComputation.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * The core part of the {@link Aggregates#setWindowFields(Object, Bson, List) $setWindowFields} pipeline stage of an aggregation pipeline.
+ * A triple of a window function, a {@linkplain Window window} and a path to a field to be computed by the window function over the window.
+ *
+ * @see WindowedComputations
+ * @since 4.3
+ */
+@Beta
+public interface WindowedComputation {
```
jyemin, line 31: I'm also not sure about this interface.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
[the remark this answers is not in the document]
jyemin: It's only a problem if the method signature is identical.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
```diff
@@ -0,0 +1,769 @@
...
+     *
+     * @param path The output field path.
+     * @param expression The expression.
+     * @param window The window.
+     * @param <TExpression> The expression type.
+     * @return The constructed {@link WindowedComputation}.
+     * @mongodb.driver.dochub core/window-functions-sum $sum
+     * @mongodb.server.release 5.0
+     */
+    public static <TExpression> WindowedComputation sum(final String path, final TExpression expression, @Nullable final Window window) {
```
jyemin, line 72: It's strange to me that for all of these windowed computations, the window itself is optional.  Can you try to explain that in the Javadoc at the class level, i.e. what the computations mean when no window is provided?

Also, I wonder how common it will be to have a null window?  Would it be better to overload the methods instead of requiring a null third parameter (that will introduce conflicts with Accumulators if we did that)?

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
```diff
@@ -0,0 +1,769 @@
...
+ *     <li>A path to an output field to be computed by the window function over the window.</li>
+ * </ul>
+ * A windowed computation is similar to an {@linkplain Accumulators accumulator} but does not result in folding documents constituting
+ * the window into a single document.
+ *
+ * @mongodb.driver.manual meta/aggregation-quick-reference/#field-paths Field paths
+ * @since 4.3
+ */
+@Beta
+public final class WindowedComputations {
```
jyemin, line 60: Did you have any other ideas for the name of this class?  It's a mouthful, and I don't love using the word computations because, in the end, isn't pretty much everything a computation?  Maybe there's something in the server scope or design doc that would suggest an alternate name?

jyemin (changes requested):
I'm going to give this another round of close review, but wanted to provide some initial general feedback on some of the higher-level API choice and names.

Overall, it looks great.

driver-core/src/test/unit/com/mongodb/client/model/TestWindowedComputations.java
rozza: Ack

driver-core/src/main/com/mongodb/client/model/Window.java
rozza: I think my main question is will it hamper users in the future? When potentially new window operators are added.  

The magic and lack of safety with Bson is that most things can be Bson (Documents etc..) So methods taking Bson are future flexible because of that.

driver-core/src/main/com/mongodb/client/model/MongoTimeUnit.java
stIncMale: I thought about it, but decided to not go with that name for the following reasons:

1. These values are also used by the derivative/integral window functions when the variable is of BSON Date type, i.e., they are used not only for defining windows.
2. The [document specifying the syntax](https://docs.google.com/document/d/1IDbgaoEN9ttB9DNPk5mbRIT6S935QaL0HKlKGwOcSHE/edit#) specifies that BSON Date bounds for range windows are computed as per the [`$dateAdd`](https://docs.mongodb.com/v5.0/reference/operator/aggregation/dateAdd/) aggregation pipeline operation. This means that if we add builders for `$dateAdd`/`$dateDiff` operations we should use the same enum there, and having  it named `WindowTimeUnit` would be confusing. Nicholas Zolnierz (the author of the aforementioned document) confirmed that windows bounds should allow the same units allowed by `$dateAdd`.

P.S. I have just noticed that `$dateAdd` allows `quarter` as a unit, but range windows do not explicitly specify this unit among allowed units. The server, nonetheless, allows `quarter` as a unit for windows, and Nicholas Zolnierz confirmed that `quarter` is supposed to be allowed. I will add it to the enum.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
stIncMale: I added this overload because I saw similar overloads for other aggregates, and anticipated a request to add it. Other than that, I do not see this overload being useful regardless of whether the `output` is allowed to be empty or not. So I would gladly remove the overload, but I went to read the syntax again, and it does say _"We will consider it a no-op rather than an error if no fields are provided."_ So I will remove the check `isTrueArgument("output must not be empty", !output.isEmpty())`.

Given the above, do you still want the overload to be removed or to stay?

jyemin: For consistency with group we should probably keep the overload then.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
stIncMale: > It's strange to me that for all of these windowed computations, the window itself is optional

This is with accordance to the MQL syntax: explicit window is optional for the majority of functions.

> Can you try to explain that in the Javadoc at the class level, i.e. what the computations mean when no window is provided?

Yes, the documentation of `WindowedComputations` states:

```
Specifying null window is equivalent to specifying an unbounded window,
i.e., a window with both ends specified as Windows.Bound.UNBOUNDED. 
Some window functions require to specify an explicit unbounded window instead of specifying null.
```

> Would it be better to overload the methods instead of requiring a null third parameter 

I am not sure this is actually useful, but since adding such overloads is a backward-compatible change, we could do it if there is an interest expressed by users. Our approach of duplicating documentation (including documentation of overloaded methods) instead of linking to a single place that could be used to document multiple program elements also results in an additional burden per each overload, and makes overload less favourable.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
stIncMale: I have not seen any special name for this construct in the server documents, and in MQL each "windowed computation" is a field of a document representing the value of the `output` field, so one might say this is an "output field", which is even how it is called in https://docs-mongodbcom-staging.corp.mongodb.com/docs/docsworker-xlarge/window-functions/reference/operator/aggregation/setWindowFields/#syntax. However, this is a non-descriptive and misleading term, because the output field path is the smallest part of the construct. Besides, we cannot call it simply `OutputField` because such a name implies nothing about the relation with window functions, so the tern would need to be made more specific somehow.

I think that "windowed computation" is quite a descriptive term and not that long. I can also imagine it being called "window function call", because it represents both the function to be called and its arguments. However, it also includes the field path where to put the result, and this may not be a good fit for the term "function call". That said, I am open to consider other names and in no way imply that this name is the best there can be, but it is the best I was able to come up with.

driver-core/src/main/com/mongodb/client/model/Window.java
stIncMale: ### General thoughts

In Java an interface/class is a language concept that allows us to define a subset of the following: a type, a container for data, a container for instructions. From the perspective of a programmer, a type merely restricts how one may use an object. In spite of the potentially negative connotation of the word "restriction", in programming such restrictions help programmers to know in advance, i.e., before running the program, what can be done with an object and in which context the object can be used. Types provide a guide for a programmer, guarding him from making some mistakes, which is also what any API, including a builder API should strive for.

I did notice that our MQL builder API uses `Bson` and `BsonField` as construction blocks, which is no different from an imaginary SQL builder API that uses `String` as a construction block. Not only such API allows combining blocks that are not combinable, but is also not self-evident, requiring to provide a user with a manual explaining what blocks can be used in what context, because there is no way for a user to extract such information from the API itself.

### Addressing specific concerns

#### Flexibility

> The magic and lack of safety with Bson is that most things can be Bson (Documents etc..) So methods taking Bson are future flexible because of that.

This seems a lot like arguments between weakly- and strongly-typed programming languages. The Java driver is already written in a strongly-typed language, it would be odd to destroy types in the driver API and use only `Object`; even stopping to use generics would be odd, despite being less drastic. What is different about the builder API when it comes to using types?

Note that it is not the case that `Window` "has no methods"--it is `Bson` and it has all the methods `Bson` has. The API implementation does not rely on any arcane knowledge when using a `Window`, which means that even if the `Windows` API would lag behind the server functionality, a user would still be able to construct a `BsonDocument` representing a window and trivially wrap into the `Window` type to use anywhere where the API requires a `Window`. If we will find ourselves constantly lagging behind the server, we may even provide an implementation of a `Window` that wraps a `BsonDocument.

Hopefully it can now be seen that having the `Window` type does not reduces the flexibility compared to using `Bson`/`BsonDocument` instead.

#### API Consistency

While it is true that the API for window functions is less weakly-typed that the rest of the builder API, this fact does not result in different usage patterns and, therefore, does not compel a user to adopt a new style. Indeed, if one were to look at the code that uses this API without storing intermediate building blocks in variables, one would not tell the difference:

```java
group("$localityId",
    Accumulators.sum("totalRainfall", "$rainfall"),
    Accumulators.avg("avgTemp", "$temperature"));

setWindowFields("$localityId", ascending("measurementDateTime"),
        WindowedComputations.sum("monthlyRainfall", "$rainfall", timeRange(-1, CURRENT, MONTH)),
        WindowedComputations.avg("monthlyAvgTemp", "$temperature", timeRange(-1, CURRENT, MONTH)));
```

### Conclusion

Given that moving towards strongly-typed API is beneficial as explained in the "General thoughts", and that it does not even reduce flexibility or introduces new usage patterns, as argued above, I see nothing even remotely sufficient to justify removing the new types.

driver-core/src/main/com/mongodb/client/model/WindowedComputation.java
stIncMale: Everything I said [here](https://github.com/mongodb/mongo-java-driver/pull/747#discussion_r667257438) is applicable fully to the `WindowedComputation` interface. It is safe to say that we can discuss them in a single thread.

driver-core/src/main/com/mongodb/client/model/MongoTimeUnit.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/Windows.java
```diff
@@ -0,0 +1,403 @@
...
+     * Creates a dynamically-sized range window whose bounds are determined by a range of possible values around
+     * the BSON {@link BsonType#DATE_TIME Date} value of the {@link Aggregates#setWindowFields(Object, Bson, List) sortBy}
+     * field in the current document.
+     *
+     * @param lower A value based on which the lower bound of the window is calculated.
+     * @param upper A value based on which the upper bound of the window is calculated.
+     * @param unit A time unit in which {@code lower} is specified.
+     * @return A range window.
+     */
+    public static Window timeRange(final long lower, final Bound upper, final MongoTimeUnit unit) {
```
stIncMale, line 306: Do you think it is worth rearranging the parameters to be `lower`, `unit`, `upper` for this method? I am asking because currently the code looks like this

```java
timeRange(Windows.Bound.CURRENT, 1, , MongoTimeUnit.MONTH); // fine
timeRange(-1, Windows.Bound.CURRENT, MongoTimeUnit.MONTH); // a bit weird
```

if we change the order, the code would look like

```java
timeRange(Windows.Bound.CURRENT, 1, , MongoTimeUnit.MONTH); // fine
timeRange(-1, MongoTimeUnit.MONTH, Windows.Bound.CURRENT); // also fine
```

jyemin: Yes, that makes a lot of sense.  Go for it.

driver-core/src/main/com/mongodb/client/model/WindowedComputations.java
jyemin: Thanks for the explanation.  Resolving the thread.

driver-core/src/main/com/mongodb/client/model/Window.java
stIncMale: I introduced the `of` builder methods. Spent quite a few time trying to write tiny Scala examples with accordance to how our Scala API is meant to be used, but I still cannot be sure (the code works, that much I am sure). If @rozza were to check the Scala examples, it would be great.

driver-core/src/main/com/mongodb/client/model/Windows.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/WindowedComputation.java
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Render into {@link BsonField}.
+     *
+     * @return A {@link BsonField} representation.
+     */
+    BsonField toBsonField();
```
stIncMale, line 37: Renamed `asBsonField` -> `toBsonField`. I used "as" after looking at `BsonValue` methods, e.g., `asArray`. However, there is `Bson.toBsonDocument` inherited by `Window`, so it seems consistent to use "to" in `WindowedComputation.toBsonField`.

driver-scala/src/main/scala/org/mongodb/scala/model/Windows.scala
```diff
@@ -58,6 +59,24 @@ import org.bson.types.Decimal128
...
+   *  val pastWeek1: Window = Windows.timeRange(-1, MongoTimeUnit.WEEK, Windows.Bound.CURRENT)
+   *  val pastWeek2: Window = Windows.of(
+   *      Document("range" -> BsonArray(-1, "current"),
+   *          "unit" -> BsonString("week")))
+   * }}}
+   *
+   * @param window A document representing the required window.
+   * @return The constructed window.
+   */
+  def of(window: Document): Window = JWindows.of(window)
```
stIncMale, line 78: There is only one `Windows.of` method in the Scala API because as far as I understand from the docs saying [_"The underlying implementations of the scala Document use the type safe BsonDocument class"_](), there is no direct counterpart of `org.bson.Document` in the Scala API, and there is no split as opposed to `org.bson.BsonDocument` and `org.bson.Document` as in the Java API.

Is my understanding correct and we do not need the overload here?

driver-core/src/main/com/mongodb/client/model/Windows.java
```diff
@@ -85,12 +87,52 @@
...
+     *  Window pastWeek1 = Windows.timeRange(-1, MongoTimeUnit.WEEK, Windows.Bound.CURRENT);
+     *  Window pastWeek2 = Windows.of(
+     *          new BsonDocument("range", new BsonArray(Arrays.asList(new BsonInt32(-1), new BsonString("current"))))
+     *                  .append("unit", new BsonString("week")));
+     * }</pre>
+     *
+     * @param window A document representing the required window.
+     * @return The constructed window.
+     */
+    public static Window of(final BsonDocument window) {
```
jyemin, line 126: I think you could just have a single `of` method with a `window` parameter of type `Bson`.  This would correspond more closely to the rest of the driver API.

jyemin (changes requested):

driver-core/src/main/com/mongodb/client/model/Windows.java
stIncMale: Indeed. Done.

driver-scala/src/main/scala/org/mongodb/scala/model/Windows.scala
stIncMale: This comment is not relevant anymore.

jyemin (approved):
I think it looks good!

driver-scala/src/main/scala/org/mongodb/scala/model/WindowedComputations.scala
```diff
@@ -0,0 +1,448 @@
...
+
+  /**
+   * Creates a windowed computation from a document field in situations when there is no builder method that better satisfies your needs.
+   * This method cannot be used to validate the document field syntax.
+   *
+   * {{{
+   *  val pastWeek: Window = Windows.timeRange(-1, MongoTimeUnit.WEEK, Windows.Bound.CURRENT)
+   *  val pastWeekExpenses1: WindowedComputation = WindowedComputations.sum("pastWeekExpenses", "$expenses", pastWeek)
+   *  val pastWeekExpenses2: WindowedComputation = WindowedComputations.of(
+   *      BsonField("pastWeekExpenses", Document("$sum" -> "$expenses",
```
stIncMale, line 51: Note that the Gradle task `:driver-scala:scaladoc` produces warnings, e.g.,
```
WindowedComputations.scala:49: warning: Variable expenses undefined in comment for method of in object WindowedComputations
   *  val pastWeekExpenses1: WindowedComputation = WindowedComputations.sum("pastWeekExpenses", "$expenses", pastWeek)
```
But they do not to make sense to me, and escaping `$` as `\$` does not help as it results in `\$` being rendered in the documentation.

driver-scala/src/main/scala/org/mongodb/scala/model/Windows.scala
```diff
@@ -0,0 +1,265 @@
...
+ * @see [[http://docs.mongodb.org/manual/reference/operator/aggregation/dateAdd/ \$dateAdd aggregation pipeline stage]]
+ * @see [[http://docs.mongodb.org/manual/reference/operator/aggregation/dateSubtract/ \$dateSubtract aggregation pipeline stage]]
+ * @since 4.3
+ * @note Requires MongoDB 5.0 or greater.
+ */
+@Beta
+object Windows {
+
+  /**
+   * Creates a window from [[org.mongodb.scala.bson.conversions.Bson]]
```
stIncMale, line 63: The Gradle task `:driver-scala:aggregatedScalaDoc` produces the following warning
```
Windows.scala:62: warning: Could not find any member to link for "org.mongodb.scala.bson.conversions.Bson"
```

Links to fully-qualified names are supposed to work according to https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html#markup. @rozza Do you know what could be wrong?

driver-scala/src/main/scala/org/mongodb/scala/model/WindowedComputations.scala
rozza: Its to do with scala templating - I'll add a ticket to clean up any scaladoc warnings post 4.3

driver-scala/src/main/scala/org/mongodb/scala/model/Windows.scala
rozza: Probably because its a type alias and can;t find `org.bson.conversions.Bson`

stIncMale: So it is not possible to create a scaladoc link to a program element in a different package?

driver-scala/src/main/scala/org/mongodb/scala/model/WindowedComputations.scala
rozza:
```
   * {{{
   *  val pastWeek: Window = Windows.timeRange(-1, MongoTimeUnit.WEEK, Windows.Bound.CURRENT)
   *  val pastWeekExpenses1: WindowedComputation = WindowedComputations.sum("pastWeekExpenses", "\$expenses", pastWeek)
   *  val pastWeekExpenses2: WindowedComputation = WindowedComputations.of(
   *      BsonField("pastWeekExpenses", Document("\$sum" -> "\$expenses",
   *          "window" -> pastWeek.toBsonDocument)))
   * }}}
```

Works for me I get a code block:

```
val pastWeek: Window = Windows.timeRange(-1, MongoTimeUnit.WEEK, Windows.Bound.CURRENT)
val pastWeekExpenses1: WindowedComputation = WindowedComputations.sum("pastWeekExpenses", "$expenses", pastWeek)
val pastWeekExpenses2: WindowedComputation = WindowedComputations.of(
    BsonField("pastWeekExpenses", Document("$sum" -> "$expenses",
        "window" -> pastWeek.toBsonDocument)))
```

driver-scala/src/main/scala/org/mongodb/scala/model/Windows.scala
rozza: > So it is not possible to create a scaladoc link to a program element in a different package?
Probably but its not going to be a quick fix for today.

stIncMale: Replaced with `Bson`. Done.

PR #748, opened 2021-07-09
Remove Beta annotation from remaining client side encryption-related classes

jyemin:
JAVA-4226

I grepped for "beta" in the entire project so I think I found everything.


docs/reference/content/driver-scala/tutorials/client-side-encryption.md
```diff
@@ -17,9 +17,6 @@ With field level encryption, developers can encrypt fields client side without a
 configuration or directives. Client-side field level encryption supports workloads where applications must guarantee that 
 unauthorized parties, including server administrators, cannot read the encrypted data.
 
-{{% note class="important" %}}
```
rozza, line 20: The note tag is still needed.  (But no longer class="important").

Added in a extra commit.

rozza (changes requested):
I noticed the hugo note tag was also removed - so added a commit to put it back.

rozza (approved):

PR #749, opened 2021-07-09
Test redaction of replies to security-sensitive commands

jyemin:
JAVA-4207


driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -280,6 +285,10 @@ public void init(final BsonArray entitiesArray, final Function<MongoClientSettin
 
     private void initClient(final BsonDocument entity, final String id,
                             final Function<MongoClientSettings, MongoClient> mongoClientSupplier) {
+        if (!SUPPORTED_CLIENT_ENTITY_OPTIONS.containsAll(entity.keySet())) {
```
jyemin, line 288: Added this so we notice when new options are added that are not supported.  In this case "observeSensitiveCommands" was added.  It just happens that we don't need to do anything to support this option, since java driver's command monitoring already does not generate events for commands executed as part of the handshake.

jyemin:
Not required for the release.

rozza (approved):

PR #759, opened 2021-07-16
Relax synchronized block in LazyPropertyModelCodec when a codec is retrieved

eshishkin:
To avoid race conditions `LazyPropertyModelCodec#getPropertyModelCodec` is made synchronized earlier. But all that it does is caching a codec on a first invocation. On subsequent calls it just returns the cached value. In case when the same POJO is requested from Mongo DB very often by different threads, having the whole method synchronized might be a bit excessive or rather wasteful (too many context switches, time spent in locks, etc). 

A simple solution with double check locking should help here and can improve total performance of the method in case concurrent requests

Here is just an example from JFR report for one of our applications
mongo_locks


rozza:
Hi @eshishkin thanks for the PR, looks good. 

I've added [JAVA-4243](https://jira.mongodb.org/projects/JAVA/issues/JAVA-4243) to track this.

stIncMale (approved):
Thank you, @eshishkin. Attached profiling results is a nice touch.

PR #761, opened 2021-07-17
fix race condition in FlatMapObservable completion handler

tkroman:
[JAVA-4241](https://jira.mongodb.org/browse/JAVA-4241)

Introducing a state machine that should take care of race conditions between child & parent's `onComplete` calls.

### DIFF
--- a/driver-scala/src/main/scala/org/mongodb/scala/internal/FlatMapObservable.scala
+++ b/driver-scala/src/main/scala/org/mongodb/scala/internal/FlatMapObservable.scala
@@ -18,72 +18,99 @@ package org.mongodb.scala.internal
 
 import org.mongodb.scala._
 
+import java.util.concurrent.atomic.AtomicReference
+
+sealed trait State
+case object Init extends State
+case class WaitingOnChild(s: Subscription) extends State
+case object LastChildNotified extends State
+case object LastChildResponded extends State
+case object Done extends State
+case object Error extends State
+
 private[scala] case class FlatMapObservable[T, S](observable: Observable[T], f: T => Observable[S])
     extends Observable[S] {
-
   // scalastyle:off cyclomatic.complexity method.length
   override def subscribe(observer: Observer[_ >: S]): Unit = {
     observable.subscribe(
       SubscriptionCheckingObserver(
         new Observer[T] {
-
-          @volatile
-          private var outerSubscription: Option[Subscription] = None
-          @volatile
-          private var nestedSubscription: Option[Subscription] = None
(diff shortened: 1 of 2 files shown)


driver-scala/src/test/scala/org/mongodb/scala/internal/FlatMapObservableTest.scala
```diff
@@ -0,0 +1,38 @@
...
+            })
+          }
+      )
+      .subscribe(
+        _ => (),
+        p.failure,
+        () => {
+          completedCounter.incrementAndGet()
+          Thread.sleep(100)
+          p.trySuccess(())
```
tkroman, line 30: without this change this would fail here because there would be 2 calls to the "root" `onComplete`, which would mean we can't reliably block until the observable is exhausted if done in this fashion (deriving `Future` from `Observable`). Also, this would fail if we did `p.success(())` instead of `trySuccess` because we don't expect to get repeated `onComplete` events.

driver-scala/src/main/scala/org/mongodb/scala/internal/FlatMapObservable.scala
```diff
@@ -18,72 +18,99 @@ package org.mongodb.scala.internal
...
                 }
 
                 override def onSubscribe(subscription: Subscription): Unit = {
-                  nestedSubscription = Some(subscription)
+                  state.set(WaitingOnChild(subscription))
                   if (demand > 0) subscription.request(demand)
                 }
 
                 override def onComplete(): Unit = {
-                  nestedSubscription = None
```
tkroman, line 80: for the sake of completeness, race condition (threads A & B):

A: `Observer[S]#onComplete`
A: `nestedSubscription = None`
B: `Observer[T]#onComplete`
B: `onCompleteCalled == false` => B: `onCompleteCalled = true` 
B: `nestedSubscription.isEmpty == true` => B: `observer.onComplete()`
A: `onCompleteCalled == true` => A: `observer.onComplete`

At least this is an explanation I've come up with while trying to debug the failing test (under original code)

tkroman:
@jyemin could you please take a look at this?

rozza:
Merged in 7033c6c2800773b5a9864e88aa2912f4adcfa696 and slight tweak for the CI in e5e2ff243f55c04e6307b57f2ae61aa1fe06c3eb

rozza:
Thanks @tkroman I manually merged and this fix will be available in the next release.

tkroman:
Thanks a lot @rozza!

PR #763, opened 2021-07-21
Run tests against Atlas serverless instance

jyemin:
Skip tests as necessary

JAVA-3987


.evergreen/run-serverless-tests.sh
```diff
@@ -0,0 +1,34 @@
...
+SINGLE_HOST=${MONGODB_URI_SINGLE_HOST:10}
+
+MONGODB_URI_FINAL="mongodb://${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${SINGLE_HOST}/?tls=true"
+
+echo ${MONGODB_URI_FINAL}
+
+./gradlew -version
+
+./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI_FINAL} -Dorg.mongodb.test.serverless=true \
+   --stacktrace --info --continue driver-sync:test
```
jyemin, line 34: Running the driver-sync tests meets the testing requirements defined in the spec.  I suppose we could run reactive streams as well but it seems redundant.

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
```diff
@@ -102,6 +102,7 @@
     public static final String MONGODB_URI_SYSTEM_PROPERTY_NAME = "org.mongodb.test.uri";
     public static final String MONGODB_API_VERSION = "org.mongodb.test.api.version";
     public static final String MONGODB_TRANSACTION_URI_SYSTEM_PROPERTY_NAME = "org.mongodb.test.transaction.uri";
+    public static final String SERVERLESS_TEST_SYSTEM_PROPERTY_NAME = "org.mongodb.test.serverless";
```
jyemin, line 105: Introduced this property since there is no way to figure out from the URI that it's serverless instance

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
```diff
@@ -515,6 +521,11 @@ public static boolean isAuthenticated() {
         return getConnectionString().getCredential() != null;
     }
 
+    public static boolean isClientSideEncryptionTest() {
```
jyemin, line 524: I had to exclude some additional CSFLE tests and realized we had this snippet of code duplicated all over the place, so abstracted into this helper method and updated all the tests that had this snippet.

driver-core/src/test/functional/com/mongodb/JsonTestServerVersionChecker.java
```diff
@@ -43,6 +43,10 @@ public static boolean skipTest(final BsonDocument testDocument, final BsonDocume
     }
 
     private static boolean canRunTest(final BsonDocument document, final ServerVersion serverVersion) {
+        if (document.containsKey("serverless") && (!serverlessMatches(document.getString("serverless").getValue()))) {
```
jyemin, line 46: Handle serverless restrictions in legacy tests.

driver-core/src/test/functional/com/mongodb/JsonTestServerVersionChecker.java
```diff
@@ -51,9 +55,8 @@ private static boolean canRunTest(final BsonDocument document, final ServerVersi
                 && serverVersion.compareTo(getMaxServerVersionForField("maxServerVersion", document)) > 0) {
             return false;
         }
-        if (document.containsKey("topology")) {
-            BsonArray topologyTypes = document.getArray("topology");
-            return topologyMatches(topologyTypes);
```
jyemin, line 56: This was a bug.  We only on want to return if we know we're skipping.  Otherwise we don't execute any of the checks below, one of which may be false.

rozza (approved):

driver-sync/src/test/functional/com/mongodb/client/unified/RunOnRequirementsMatcher.java
```diff
@@ -89,6 +93,19 @@ public static boolean runOnRequirementsMet(final BsonArray runOnRequirements, fi
         return false;
     }
 
+    private static boolean serverlessMatches(final String serverlessRequirement) {
```
stIncMale, line 96: I propose reusing `JsonTestServerVersionChecker.serverlessMatches` here. Note that the same has already been done for `JsonTestServerVersionChecker.topologyMatches`.

stIncMale (changes requested):
Just a few minor things

driver-core/src/test/functional/com/mongodb/JsonTestServerVersionChecker.java
jyemin: I went with

        if ((!serverlessMatches(document.getString("serverless", new BsonString("allow")).getValue()))) {
            return false;
        }

instead, as I prefer not to pass BsonString parameters to places that shouldn't have to care about BSON.

jyemin: Done

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
```diff
@@ -515,6 +521,11 @@ public static boolean isAuthenticated() {
...
+        return System.getProperty("org.mongodb.test.awsAccessKeyId") != null
+                && !System.getProperty("org.mongodb.test.awsAccessKeyId").isEmpty();
```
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
```diff
@@ -236,6 +237,11 @@ public static synchronized ConnectionString getMultiMongosConnectionString() {
         return getConnectionStringFromSystemProperty(MONGODB_TRANSACTION_URI_SYSTEM_PROPERTY_NAME);
     }
 
+    public static boolean isServerlessTest() {
+        String isServerlessSystemProperty = System.getProperty(SERVERLESS_TEST_SYSTEM_PROPERTY_NAME);
+        return isServerlessSystemProperty != null && isServerlessSystemProperty.equals("true");
```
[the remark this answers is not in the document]
jyemin: Done

driver-sync/src/test/functional/com/mongodb/client/unified/RunOnRequirementsMatcher.java
jyemin: Agreed.  Done.

stIncMale (approved):

PR #765, opened 2021-07-24
Support calling `QueryBatchCursor.close` concurrently with other `QueryBatchCursor` methods

stIncMale:
Run `QueryBatchCursor` integration tests (`QueryBatchCursorFunctionalSpecification`) with load balancer.

The high-level implementation idea is described in the documentation of the class `QueryBatchCursor.ResourceManager`. Only `QueryBatchCursor` required a change as per this [analysis](https://jira.mongodb.org/browse/JAVA-4183?focusedCommentId=3950629&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3950629).

JAVA-4183

### DIFF
--- a/.evergreen/run-load-balancer-tests.sh
+++ b/.evergreen/run-load-balancer-tests.sh
@@ -74,11 +74,22 @@ echo $first
   --tests UnifiedTransactionsTest \
   --tests InitialDnsSeedlistDiscoveryTest
 second=$?
+echo $second
+
+./gradlew -PjdkHome=/opt/java/${JDK} \
+  -Dorg.mongodb.test.uri=${SINGLE_MONGOS_LB_URI} \
+  -Dorg.mongodb.test.transaction.uri=${MULTI_MONGOS_LB_URI} \
+  ${GRADLE_EXTRA_VARS} --stacktrace --info --continue driver-core:test \
+  --tests QueryBatchCursorFunctionalSpecification
+third=$?
+echo $third
 
 if [ $first -ne 0 ]; then
    exit $first
 elif [ $second -ne 0 ]; then
    exit $second
+elif [ $third -ne 0 ]; then
+   exit $third
 else
    exit 0
 fi
--- a/driver-core/src/main/com/mongodb/assertions/Assertions.java
+++ b/driver-core/src/main/com/mongodb/assertions/Assertions.java
@@ -179,7 +179,7 @@ public final class Assertions {
 
     /**
      * @throws AssertionError Always
-     * @return Never completes normally. The return type is {@link AssertionError} to allow writing {@code throws fail()}.
+     * @return Never completes normally. The return type is {@link AssertionError} to allow writing {@code throw fail()}.
      * This may be helpful in non-{@code void} methods.
      */
     public static AssertionError fail() throws AssertionError {
@@ -189,7 +189,7 @@ public final class Assertions {
     /**
      * @param msg The failure message.
      * @throws AssertionError Always
-     * @return Never completes normally. The return type is {@link AssertionError} to allow writing {@code throws fail("failure message")}.
+     * @return Never completes normally. The return type is {@link AssertionError} to allow writing {@code throw fail("failure message")}.
      * This may be helpful in non-{@code void} methods.
      */
     public static AssertionError fail(final String msg) throws AssertionError {
--- a/driver-core/src/main/com/mongodb/assertions/Assertions.java
+++ b/driver-core/src/main/com/mongodb/assertions/Assertions.java
@@ -179,9 +179,21 @@ public final class Assertions {
 
     /**
      * @throws AssertionError Always
+     * @return Never completes normally. The return type is {@link AssertionError} to allow writing {@code throws fail()}.
+     * This may be helpful in non-{@code void} methods.
      */
-    public static void fail() throws AssertionError {
+    public static AssertionError fail() throws AssertionError {
         throw new AssertionError();
+    }
+
+    /**
+     * @param msg The failure message.
+     * @throws AssertionError Always
+     * @return Never completes normally. The return type is {@link AssertionError} to allow writing {@code throws fail("failure message")}.
+     * This may be helpful in non-{@code void} methods.
+     */
+    public static AssertionError fail(final String msg) throws AssertionError {
+        throw new AssertionError(assertNotNull(msg));
     }
 
     private Assertions() {
--- a/driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
@@ -16,6 +16,7 @@
 
 package com.mongodb.internal.operation;
 
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import com.mongodb.MongoChangeStreamException;
 import com.mongodb.MongoException;
@@ -44,7 +45,7 @@ final class ChangeStreamBatchCursor<T> implements AggregateResponseBatchCursor<T
 
     private AggregateResponseBatchCursor<RawBsonDocument> wrapped;
     private BsonDocument resumeToken;
-    private volatile boolean closed;
+    private final AtomicBoolean closed;
 
     ChangeStreamBatchCursor(final ChangeStreamOperation<T> changeStreamOperation,
                             final AggregateResponseBatchCursor<RawBsonDocument> wrapped,
@@ -56,6 +57,7 @@ final class ChangeStreamBatchCursor<T> implements AggregateResponseBatchCursor<T
         this.wrapped = wrapped;
         this.resumeToken = resumeToken;
         this.maxWireVersion = maxWireVersion;
+        closed = new AtomicBoolean();
     }
 
     AggregateResponseBatchCursor<RawBsonDocument> getWrapped() {
@@ -108,8 +110,7 @@ final class ChangeStreamBatchCursor<T> implements AggregateResponseBatchCursor<T
 
     @Override
     public void close() {
-        if (!closed) {
-            closed = true;
+        if (!closed.getAndSet(true)) {
             wrapped.close();
             binding.release();
         }
--- a/driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
@@ -22,12 +22,16 @@ import com.mongodb.MongoNamespace;
 import com.mongodb.MongoSocketException;
 import com.mongodb.ReadPreference;
 import com.mongodb.ServerAddress;
+import com.mongodb.ServerApi;
 import com.mongodb.ServerCursor;
+import com.mongodb.annotations.ThreadSafe;
 import com.mongodb.connection.ServerType;
 import com.mongodb.internal.binding.ConnectionSource;
 import com.mongodb.internal.connection.Connection;
 import com.mongodb.internal.connection.QueryResult;
+import com.mongodb.internal.session.SessionContext;
 import com.mongodb.internal.validator.NoOpFieldNameValidator;
+import com.mongodb.lang.Nullable;
 import org.bson.BsonArray;
 import org.bson.BsonDocument;
 import org.bson.BsonInt32;
@@ -40,8 +44,13 @@ import org.bson.codecs.Decoder;
 
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.StampedLock;
 
 import static com.mongodb.assertions.Assertions.assertNotNull;
+import static com.mongodb.assertions.Assertions.assertNull;
+import static com.mongodb.assertions.Assertions.assertTrue;
+import static com.mongodb.assertions.Assertions.fail;
 import static com.mongodb.assertions.Assertions.isTrueArgument;
 import static com.mongodb.assertions.Assertions.notNull;
 import static com.mongodb.internal.operation.CursorHelper.getNumberToReturn;
@@ -62,17 +71,13 @@ class QueryBatchCursor<T> implements AggregateResponseBatchCursor<T> {
     private final Decoder<T> decoder;
     private final long maxTimeMS;
     private int batchSize;
-    private ConnectionSource connectionSource;
-    private Connection connection;
-    private ServerCursor serverCursor;
     private List<T> nextBatch;
     private int count;
-    private volatile boolean closed;
     private BsonDocument postBatchResumeToken;
     private BsonTimestamp operationTime;
     private final boolean firstBatchEmpty;
     private int maxWireVersion = 0;
-    private boolean killCursorOnClose = true;
+    private final CloseableState closeableState;
 
     QueryBatchCursor(final QueryResult<T> firstQueryResult, final int limit, final int batchSize, final Decoder<T> decoder) {
         this(firstQueryResult, limit, batchSize, decoder, null);
@@ -102,37 +107,43 @@ class QueryBatchCursor<T> implements AggregateResponseBatchCursor<T> {
             this.operationTime = result.getTimestamp(OPERATION_TIME, null);
             this.postBatchResumeToken = getPostBatchResumeTokenFromResponse(result);
         }
-        if (firstQueryResult.getCursor() != null) {
+        ServerCursor serverCursor = initFromQueryResult(firstQueryResult);
+        if (serverCursor != null) {
             notNull("connectionSource", connectionSource);
         }
-        if (connectionSource != null) {
-            this.connectionSource = connectionSource.retain();
-        }
-
-        initFromQueryResult(firstQueryResult);
         firstBatchEmpty = firstQueryResult.getResults().isEmpty();
-
+        Connection connectionToPin = null;
+        boolean releaseServerAndResources = false;
         if (connection != null) {
             this.maxWireVersion = connection.getDescription().getMaxWireVersion();
             if (limitReached()) {
-                killCursor(connection);
+                releaseServerAndResources = true;
             } else {
                 assertNotNull(connectionSource);
                 if (connectionSource.getServerDescription().getType() == ServerType.LOAD_BALANCER) {
-                    this.connection = connection.retain();
-                    this.connection.markAsPinned(Connection.PinningMode.CURSOR);
+                    connectionToPin = connection;
                 }
             }
         }
-        releaseConnectionAndSourceIfNoServerCursor();
+        closeableState = new CloseableState(namespace, connectionSource, connectionToPin, serverCursor);
+        if (releaseServerAndResources) {
+            closeableState.releaseServerAndClientResources(assertNotNull(connection));
+        }
     }
 
     @Override
     public boolean hasNext() {
-        if (closed) {
+        if (!closeableState.startOperation()) {
             throw new IllegalStateException("Cursor has been closed");
         }
+        try {
+            return doHasNext();
+        } finally {
+            closeableState.endOperation();
+        }
+    }
 
+    private boolean doHasNext() {
         if (nextBatch != null) {
             return true;
         }
@@ -141,9 +152,11 @@ class QueryBatchCursor<T> implements AggregateResponseBatchCursor<T> {
             return false;
         }
 
+        ServerCursor serverCursor = closeableState.serverCursor();
         while (serverCursor != null) {
-            getMore();
-            if (closed) {
+            getMore(serverCursor);
+            serverCursor = closeableState.serverCursor();
+            if (!closeableState.operable()) {
                 throw new IllegalStateException("Cursor has been closed");
             }
             if (nextBatch != null) {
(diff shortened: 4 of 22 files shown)


driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
```diff
@@ -108,8 +110,8 @@ public Boolean apply(final AggregateResponseBatchCursor<RawBsonDocument> queryBa
 
     @Override
     public void close() {
-        if (!closed) {
-            closed = true;
+        if (!closed.getAndSet(true)) {
```
stIncMale, line 113: Made idempotent as per the specification of `BatchCursor.close`. Now the `close` method in all our cursors is idempotent.

driver-core/src/test/functional/com/mongodb/internal/connection/ServerHelper.java
```diff
@@ -62,15 +61,25 @@ public static void waitForLastRelease(final ServerAddress address, final Cluster
...
         if (pool.getInUseCount() > 0) {
             throw new IllegalStateException("Connection pool in use count is " + pool.getInUseCount());
         }
     }
 
+    private static ConcurrentPool<UsageTrackingInternalConnection> connectionPool(final Server server) {
+        ConnectionPool connectionPool;
+        if (server instanceof DefaultServer) {
+            connectionPool = ((DefaultServer) server).getConnectionPool();
+        } else if (server instanceof LoadBalancedServer) {
```
stIncMale, line 75: All the changes in `ServerHelper` are done to support running `QueryBatchCursorFunctionalSpecification` and `AsyncQueryBatchCursorFunctionalSpecification` with load balancer.

driver-sync/src/main/com/mongodb/client/MongoChangeStreamCursor.java
```diff
@@ -24,17 +25,18 @@
...
  *         System.out.println(cursor.next());
  *     }
  * }
- * </pre></blockquote>
+ * }</pre>
  *
  * @since 3.11
  * @param <TResult> The type of documents the cursor contains
  */
+@NotThreadSafe
```
stIncMale, line 39: `MongoCursor` (the super-interface) has already been annotated with `@NotThreadSafe`. Added the annotation here to avoid user confusion. I can see how absence of the annotation can be argued for. Accordingly, I will undo this change without arguing in the event of being asked to do so.

.evergreen/run-load-balancer-tests.sh
```diff
@@ -74,11 +74,22 @@ echo $first
   --tests UnifiedTransactionsTest \
   --tests InitialDnsSeedlistDiscoveryTest
 second=$?
+echo $second
+
+./gradlew -PjdkHome=/opt/java/${JDK} \
+  -Dorg.mongodb.test.uri=${SINGLE_MONGOS_LB_URI} \
+  -Dorg.mongodb.test.transaction.uri=${MULTI_MONGOS_LB_URI} \
+  ${GRADLE_EXTRA_VARS} --stacktrace --info --continue driver-core:test \
+  --tests QueryBatchCursorFunctionalSpecification
```
stIncMale, line 83: I was unable to add {{AsyncQueryBatchCursorFunctionalSpecification}} here because some tests fail with load balancer. Filed https://jira.mongodb.org/browse/JAVA-4259.

driver-core/src/test/unit/com/mongodb/internal/operation/AsyncQueryBatchCursorSpecification.groovy
```diff
@@ -27,6 +27,7 @@ import com.mongodb.connection.ServerConnectionState
 import com.mongodb.connection.ServerDescription
 import com.mongodb.connection.ServerType
 import com.mongodb.connection.ServerVersion
+import com.mongodb.internal.async.SingleResultCallback
```
stIncMale, line 30: The changes in `AsyncQueryBatchCursorSpecification` were done to make unit tests related to calling `close` concurrently with other operations more in line with the corresponding unit tests in `QueryBatchCursorSpecification`.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -102,37 +107,43 @@
...
     @Override
     public boolean hasNext() {
-        if (closed) {
+        if (!closeableState.startOperation()) {
             throw new IllegalStateException("Cursor has been closed");
         }
+        try {
+            return doHasNext();
+        } finally {
+            closeableState.endOperation();
```
stIncMale, line 142: This comment elaborates on  `startOperation`/`endOperation` and the extracted `do...` methods.

`QueryBatchCursor` has five public methods that access releasable resources:

- `close`
- `getServerCursor`
- `hasNext`
- `next`
- `tryNext`

`getServerCursor` is a special trivial case that reads a reference to a releasable resource and returns it, so its implementation is trivial: it can either see the resource (if it was there in the first place and it was not killed) or not. We are not going to discuss this method anymore.

The rest four methods are less trivial. Methods `hasNext`, `next`, `tryNext` are supposed to be called sequentially because `QueryBatchCursor` is not thread-safe, but the method `close` can be called concurrently with either of them and with itself. We want all the releasable resources to stay available until `hasNext`/`next`/`tryNext` is running, as otherwise they would have to be prepared to lose any resource at any point, which significantly complicates matters. Any of the three aforementioned operations signify that it is running by executing `startOperation`, and signifies that it stopped running `endOperation`.

The demarkation described above is complicated by the fact that operations may have mutual calls, e.g., `next` calls `hasNext`. This means that if `hasNext` signifies that it has completed, we cannot assume that the resources are not used anymore as the `next` that called `hasNext` may still be running. Surely this can be handled by counting the number of active operations, but a simpler approach is to get rid of the root cause: one operation calling another. This is what I did in this PR: the actual logic of the methods that use `startOperation`/`endOperation` was extracted to methods with the prefix `do` (`hasNext` -> `doHasNext`) where needed, such that the `do` methods can continue calling each other, while we can rely on resources not being used after `endOperation` returning and before `startOperation` starting. The `startOperation`/`endOperation` operations are now totally ordered with `start` always followed by `end`. These methods now enforce this order internally, which also allows us to detect invalid concurrent usage of the `hasNext`/`next`/`tryNext` methods.

If the resources are released after `endOperation` returning and before `startOperation` starting (which is the only logical time at which they could be released by the deferred close action), then the next `startOperation` returns `false`, signifying that the `QueryBatchCursor` has been closed.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -102,37 +107,43 @@
...
+        closeableState = new CloseableState(namespace, connectionSource, connectionToPin, serverCursor);
+        if (releaseServerAndResources) {
+            closeableState.releaseServerAndClientResources(assertNotNull(connection));
+        }
     }
 
     @Override
     public boolean hasNext() {
-        if (closed) {
+        if (!closeableState.startOperation()) {
```
rozza, line 136: I think the naming here could be worked on. It reads if not started the cursor has been closed. Which doesn't make sense.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -156,11 +169,18 @@ public boolean hasNext() {
 
     @Override
     public List<T> next() {
-        if (closed) {
+        if (!closeableState.startOperation()) {
             throw new IllegalStateException("Iterator has been closed");
```
rozza, line 173: Same comment as before (error message is wierdly different).

Perhaps a loan method could dry this logic up:

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -102,37 +107,43 @@
...
                 assertNotNull(connectionSource);
                 if (connectionSource.getServerDescription().getType() == ServerType.LOAD_BALANCER) {
-                    this.connection = connection.retain();
-                    this.connection.markAsPinned(Connection.PinningMode.CURSOR);
+                    connectionToPin = connection;
                 }
             }
         }
-        releaseConnectionAndSourceIfNoServerCursor();
+        closeableState = new CloseableState(namespace, connectionSource, connectionToPin, serverCursor);
```
rozza, line 128: I think we could do with a better name for this class / variable. To reflect its use - essentially its the manager of the cursor / connection etc.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -217,25 +233,27 @@ boolean tryHasNext() {
...
     @Override
+    @Nullable
     public ServerCursor getServerCursor() {
-        if (closed) {
+        if (!closeableState.operable()) {
             throw new IllegalStateException("Iterator has been closed");
         }
 
-        return serverCursor;
+        return closeableState.serverCursor();
```
rozza, line 251: Can these IllegalStateExceptions be dry'd up?

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
rozza: I think you can clean and encapsulate this pattern to inside the closeable state class:

Something like this (my naming isn't great):
```
        <A> A inOperation(final Supplier<A> supplier) {
            if (!startOperation()) {
                throw new IllegalStateException("Iterator has been closed");
            }
            try {
                return supplier.get();
            } finally {
                endOperation();
            }
        }
```

would allow for:

```
    @Override
    public boolean hasNext() {
        return closeableState.inOperation(this::doHasNext);
    }
    
    public List<T> next() {
        return closeableState.inOperation(this::doNext);
    }

```

I think something like that clarifies the intent more.

rozza (changes requested):
I like the changes, so only have some nits and suggestions.  I think there are some opportunities to DRY the code up.  For example where state is checked can that logic be added to the CloseableState class.

I'm not sold on the name of the variable closableState, I think it could be renamed to reflect how its used.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
stIncMale: We have different error messages depending on what methods of the `QueryBatchCursor` detected that it is closed. However, I tried removing the repetition by moving the code that throws `IllegalStateException` to `ResourceManager.execute`, which is essentially the `inOperation` you suggested.

Since most of your concerns are similar, I will not reply to all the threads to avoid duplication. What do you think about the changed code?

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
stIncMale: Renamed to `ResourceManager`.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -320,83 +322,311 @@ private BsonDocument asGetMoreCommandDocument() {
...
+        return initFromQueryResult(queryResult);
     }
 
     private boolean limitReached() {
         return Math.abs(limit) != 0 && count >= Math.abs(limit);
     }
 
-    private void killCursor() {
-        if (serverCursor != null && killCursorOnClose) {
+    private BsonDocument getPostBatchResumeTokenFromResponse(final BsonDocument result) {
```
stIncMale, line 344: This is not a new method. Its position has changed as a result of removing multiple methods (their logic went to `ResourceManger`), and the diff tool decided it is new.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -320,83 +322,311 @@ private BsonDocument asGetMoreCommandDocument() {
...
-    }
+        void setServerCursor(@Nullable final ServerCursor serverCursor) {
+            assertTrue(state.inProgress());
+            assertNotNull(this.serverCursor);
+            // without `connectionSource` we will not be able to kill `serverCursor` later
+            assertNotNull(connectionSource);
+            this.serverCursor = serverCursor;
+        }
 
-    private BsonDocument getPostBatchResumeTokenFromResponse(final BsonDocument result) {
```
stIncMale, line 395: This method was actually neither removed nor changed.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -320,83 +322,311 @@ private BsonDocument asGetMoreCommandDocument() {
...
+                ServerCursor localServerCursor = serverCursor;
+                if (localServerCursor != null) {
+                    killServerCursor(namespace, localServerCursor, sessionContext(), serverApi, assertNotNull(connection));
+                }
+            } finally {
+                serverCursor = null;
+            }
+        }
+
+        private static void killServerCursor(final MongoNamespace namespace, final ServerCursor serverCursor,
```
stIncMale, line 576: The logic of this method came mostly from the former `killCursor` method.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -320,83 +322,311 @@ private BsonDocument asGetMoreCommandDocument() {
         return document;
     }
 
-    private void initFromQueryResult(final QueryResult<T> queryResult) {
-        serverCursor = queryResult.getCursor();
+    @Nullable
+    private ServerCursor initFromQueryResult(final QueryResult<T> queryResult) {
```
stIncMale, line 326: The changes in `initFromQueryResult` and `initFromCommandResult` were made to move the code changing the reference to `ServerCursor` to `getMore`, such that it is more explicit and, what is more important, to reduce the number of methods that use releasable resources managed by `ResourceManager` either directly or indirectly.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
rozza: Like the changes :+1:

rozza (approved):
LGTM

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -263,49 +262,47 @@ public int getMaxWireVersion() {
...
-            if (this.connection != null) {
-                killCursorOnClose = false;
+            try {
+                resourceManager.onCorruptedConnection(connection);
+            } catch (RuntimeException suppressed) {
+                e.addSuppressed(suppressed);
             }
             throw e;
         } finally {
             connection.release();
```
jyemin, line 298: This seems like it should be the responsibility of ResourceManager to release the Connection, since it's the one that has retained it.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -320,83 +317,313 @@ private BsonDocument asGetMoreCommandDocument() {
...
+     * This total order induces the happens-before order.
+     * <p>
+     * The deferred close action does not violate externally observable idempotence of {@link QueryBatchCursor#close()},
+     * because {@link QueryBatchCursor#close()} is allowed to release resources "eventually".
+     * <p>
+     * Only methods explicitly documented as thread-safe are thread-safe,
+     * others are not and rely on the total order mentioned above.
+     */
+    @ThreadSafe
+    private static final class ResourceManager {
```
jyemin, line 361: Can we consider making this a non-static inner class and removing namespace property, and perhaps moving serverCursor back to the top-level?  I find it a bit confusing that a class that "maintains all resources that must be released" also contains these two properties, both of which were already available as properties of QueryBatchCursor.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -320,83 +317,313 @@ private BsonDocument asGetMoreCommandDocument() {
...
+        @Nullable
+        private volatile Connection pinnedConnection;
+        @Nullable
+        private volatile ServerCursor serverCursor;
+        private volatile boolean skipReleasingServerResourcesOnClose;
+
+        ResourceManager(final MongoNamespace namespace, @Nullable final ConnectionSource connectionSource,
+                @Nullable final Connection connectionToPin, @Nullable final ServerCursor serverCursor) {
+            this.namespace = namespace;
+            lock = new StampedLock().asWriteLock();
```
jyemin, line 376: For my edification: why is this preferable to ReentrantLock?

stIncMale: Linking my previous [answer to this question](https://github.com/mongodb/mongo-java-driver/pull/722/files#r644208261). I am happy to discuss if needed.

jyemin: Nah, I just forgot.  Thanks!

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
jyemin: Update: I'm sure you already considered it, so perhaps start by explaining the rationale.  Perhaps it's because making it an inner class gives it unnecessary access to many other fields that it doesn't need access to?

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
stIncMale: When we use the method `ResourceManager.connection`, the `ResourceManager` acts as a `ConnectionSource`, and just like with the `ConnectionSource`, the object that provides a connection retains it, and caller must release it<sup>*</sup>. However, this indeed can be changed if a method `ResourceManager.doWithConnection` is introduced. Will do.

----
<sup>*</sup> This is a consequence of the `ReferenceCounted` contract, which states that a newly created object has count 1 and must be released even if the user of the object did not retained the object explicitly. I do not think there is another way `ReferenceCounted` can be specified.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
stIncMale: I think `serverCursor` should stay part of the `ResourceManager` specifically because it is effectively a representation of a releasable resource (the resource is server-side, so we cannot have the resource itself in the `ResourceManager`).

Making `ResourceManager` an inner class will indeed reduce the isolation, but I am not concerned with it much within a single source file, so will happily do the change, though this is indeed the reason I tend to prefer static nested classes to inner classes.

What is a problem I was afraid someone would point out, is the method `ResourceManager.serverApi` :), which seems completely out of place. This is something `QueryBatchCursor` may extract in the constructor store as its field, instead of calling `ResourceManager.serverApi`. However, when I tried doing this previously, I discovered there is a number of unit tests that start failing, because Spock fails to mock the `final ServerApi` class (which does not mean the class should not be final). Those tests require `ConnectionSource.getServerApi` to be mocked explicitly. I will look at this situation again.

driver-core/src/main/com/mongodb/assertions/Assertions.java
```diff
@@ -179,11 +179,23 @@ public static boolean assertFalse(final boolean value) throws AssertionError {
 
     /**
      * @throws AssertionError Always
+     * @return Never completes normally. The return type is {@link AssertionError} to allow writing {@code throws fail()}.
```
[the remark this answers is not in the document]
stIncMale: This seems to be the only way to tell the compiler that a specific method call (`fail`) always completes abruptly. So far we were able to live without it, but this seemed like a convenient thing when I was writing `tryStartOperation`: the method returns `boolean`, which means that the compiler requires either returning or explicitly throwing something from the branch `else { fail(); }`.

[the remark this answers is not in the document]
jyemin: Yep, I got it.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
jyemin: Just to be clear: I'm not necessarily advocating for this change.  I'm just asking you to consider it, and if you decide it's better to leave it as is I will be ok with that.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
stIncMale: Done. I think the code does look simpler now.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
stIncMale: Right. I think changing this is likely beneficial.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
jyemin: I like it!

jyemin (approved):
QueryBatchCursor is much more legible now.

LGTM

PR #768, opened 2021-07-27
Support aggregation with $merge as a string

jyemin:
JAVA-4258


rozza (approved):
LGTM

Shame there is still lots of logic duplicated and mixed up in the Publisher / Iterable implementations.

PR #771, opened 2021-08-04
Fix import of javax.security.auth.login

stbischof:
The bundle does not import the package of "javax.security.auth.login.LoginException".

This is just a "quick fix". I recommend to use it this way:

```
        '!foo.bar.*', // ! signals to exclude from importing 
        'io.netty.*;resolution:=optional',
        'org.xerial.snappy.*;resolution:=optional',
        'com.github.luben.zstd.*;resolution:=optional',
        'org.slf4j.*;resolution:=optional',
        'jnr.unixsocket.*;resolution:=optional',
        'com.mongodb.crypt.capi.*;resolution:=optional',
        '*' // import all that is not excluded or modified before
    ].join(',')
```


jyemin:
Hi @stbischof,

I like your recommendation.  If you wouldn't mind changing the PR to reflect it and then testing it (unfortunately we don't have an OSGi testing environment set up), we'd be happy to merge it.

Thanks,
Jeff

rozza:
Hi @stbischof I've created #779 applying your suggested fix.  Closing this PR in favour of that PR.

stbischof:
Thx

PR #774, opened 2021-08-19
Sync latest JSON specification test changes

jyemin:
This also pulls in changes for other, unrelated tickets.  It's hard to avoid that.

JAVA-4217


driver-core/src/test/resources/server-discovery-and-monitoring-integration/rediscover-quickly-after-step-down.json
```diff
@@ -45,27 +45,14 @@
...
-          "command_name": "replSetFreeze",
-          "arguments": {
-            "command": {
-              "replSetFreeze": 0
-            },
-            "readPreference": {
-              "mode": "Secondary"
-            }
-          }
-        },
```
stIncMale: These changes came from https://jira.mongodb.org/browse/SPEC-1708 (https://jira.mongodb.org/browse/DRIVERS-1290). They are indeed unrelated to the https://jira.mongodb.org/browse/JAVA-3988 epic, but it appears that we have the same problem with regard to them: the JAVA ticket branched from DRIVERS-1290 is marked as closed, yet it was not implemented. It appears that diverging our copies of the spec tests with regard to DRIVERS-1290 was a conscious decision. What was the reason behind the decision and do we have plans to remove the discrepancy, or, on the opposite, we plan to continue operating in a way that causes the discrepancies to grow?

jyemin: From Esha's comment on JAVA-3765, it was because we weren't seeing the test fail.

Can we leave the general discussion about divergence to the next retrospective?  It doesn't seem directly related to whether this PR should be merged.

jyemin: P.S. I did actually try to merge this change originally, but the test failed, and fixing it would have required unrelated test runner changes.  That's the main reason I rolled it back.

stIncMale: > From Esha's comment on JAVA-3765, it was because we weren't seeing the test fail.

Right, that's how I realized that it was decided to diverge the spec tests. However, since I do not think this justifies the divergence, I hoped there was something else.

Will raise this concern during the retrospective meeting.

stIncMale (approved):

PR #775, opened 2021-08-20
Deprecate mapReduce

jyemin:
Deprecate all public API related to the mapReduce command, which has been
deprecated in MongoDB server as superseded by the aggregate command.

See https://docs.mongodb.com/manual/reference/command/mapReduce/ for more
information.

JAVA-4235

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/Operations.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/Operations.java
@@ -36,7 +36,6 @@ import com.mongodb.client.model.IndexModel;
 import com.mongodb.client.model.InsertManyOptions;
 import com.mongodb.client.model.InsertOneModel;
 import com.mongodb.client.model.InsertOneOptions;
-import com.mongodb.client.model.MapReduceAction;
 import com.mongodb.client.model.RenameCollectionOptions;
 import com.mongodb.client.model.ReplaceOneModel;
 import com.mongodb.client.model.ReplaceOptions;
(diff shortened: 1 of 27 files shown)


driver-core/src/main/com/mongodb/internal/operation/Operations.java
```diff
@@ -36,7 +36,6 @@
 import com.mongodb.client.model.InsertManyOptions;
 import com.mongodb.client.model.InsertOneModel;
 import com.mongodb.client.model.InsertOneOptions;
-import com.mongodb.client.model.MapReduceAction;
```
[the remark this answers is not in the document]
jyemin: I think we now understand: it's because source compatibility is still Java 8, but JEP-211 applies to Java 9-compatible source.

driver-sync/src/main/com/mongodb/client/internal/MapReduceIterableImpl.java
```diff
@@ -90,99 +90,100 @@ public void toCollection() {
...
 
     @Override
-    public MapReduceIterable<TResult> maxTime(final long maxTime, final TimeUnit timeUnit) {
+    public com.mongodb.client.MapReduceIterable<TResult> maxTime(final long maxTime, final TimeUnit timeUnit) {
         notNull("timeUnit", timeUnit);
         this.maxTimeMS = TimeUnit.MILLISECONDS.convert(maxTime, timeUnit);
         return this;
     }
 
+    @SuppressWarnings("deprecation")
```
stIncMale, line 148: Do we really need this annotation here and one on the private `action` field above, given that the whole class is annotated with `@SuppressWarnings("deprecation")`? I expect we can remove them.

stIncMale (changes requested):

driver-sync/src/main/com/mongodb/client/internal/MapReduceIterableImpl.java
jyemin: No, we don't.

Done.

PR #778, opened 2021-09-01
POC for Atlas search builder API

jyemin:
JAVA-4190

https://docs.atlas.mongodb.com/atlas-search/

This is a draft PR to gather feedback.  So as not to waste time, it only implements a small subset of a complete API that would cover all operators and options.  I think it's enough to get feedback on the following questions

1. Does this API substantially improve the development experience for users of Atlas search?
2. Are there improvements we can make on the proposed API?  For example, is immutability the right choice for the operators?

To give a flavor of what the API usage looks like with a realistic query, I translated an actual customer query from JSON to this API.  JSON looks like this:

```
{
    "$search": {
        "index": "shopUsed",
        "compound": {
            "filter": [
                {"text": {"query": "ACTIVE", "path": "status"}},
                {"equals": {"value": false, "path": "manuallySuppressed"}},
                {"equals": {"value": false,"path": "purchasePending"}},
                {
                    "range": {
                        "path": "ymmt.year",
                        "gte": 2012,
                        "lte": 2021
                    }
                },
            ],
            "must": [
                {
                    "compound": {
                        "should": [
                            {
                                "text": {
                                    "query": "Honda",
                                    "path": "ymmt.make"
                                }
                            },
                            {
                                "compound": {
                                    "should": [
                                        {
                                            "text": {
                                                "query": "Toyota",
                                                "path": "ymmt.make"
                                            }
                                        },
                                        {
                                            "text": {
                                                "query": "Camry",
                                                "path": "ymmt.model"
                                            }
                                        }
                                    ],
                                    "minimumShouldMatch": 2
                                }
                            },
                            {
                                "compound": {
                                    "should": [
                                        {
                                            "text": {
                                                "query": "Ford",
                                                "path": "ymmt.make"
                                            }
                                        },
                                        {
                                            "text": {
                                                "query": ["Focus", "Taurus", "F-150"],
                                                "path": "ymmt.model"
                                            }
                                        },
                                    ],
                                    "minimumShouldMatch": 2
                                }
                            }
                        ],
                        "minimumShouldMatch": 1,
                        "must": {"text": {
                            "query": "sedan",
                            "path": [
                                "bodyStyle"
                            ],
                            "fuzzy": {
                                "maxEdits": 1,
                                "maxExpansions": 100
                            }
                        }}
                    }
                }
            ]
        }
    }
}
```

Using this builder API (plus a lot of static imports), the same query looks like this:

```
                search(compound(
                        Map.of(FILTER, List.of(
                                        text("ACTIVE", path("status")),
                                        equal(false, path("manuallySuppressed")),
                                        equal(false, path("purchasePending")),
                                        range(path("ymmt.year")).gte(2012).lte(2021)),
                                MUST, List.of(
                                        compound(Map.of(
                                                SHOULD, List.of(
                                                        text("Honda", path("ymmt.make")),
                                                        compound(Map.of(SHOULD,
                                                                List.of(
                                                                        text("Toyota", path("ymmt.make")),
                                                                        text("Camry", path("ymmt.model"))
                                                                ))).minimumShouldMatch(2),
                                                        compound(Map.of(SHOULD,
                                                                List.of(
                                                                        text("Ford", path("ymmt.make")),
                                                                        text(asList("Focus", "Taurus", "F-150"), asList(path("ymmt.model")))
                                                                ))).minimumShouldMatch(2)),
                                                MUST, List.of(
                                                        text("sedan", path("bodyStyle"))
                                                ))).minimumShouldMatch(1))))
                        .index("shopUsed"))
```

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/search/RangeSearchOperator.java
+++ b/driver-core/src/main/com/mongodb/client/model/search/RangeSearchOperator.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.search;
+
+import com.mongodb.annotations.Immutable;
+import com.mongodb.lang.Nullable;
+import org.bson.BsonDecimal128;
+import org.bson.BsonDocument;
+import org.bson.BsonDouble;
+import org.bson.BsonInt32;
+import org.bson.BsonInt64;
+import org.bson.BsonValue;
+import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.types.Decimal128;
+
+import java.util.List;
+
+import static com.mongodb.assertions.Assertions.notNull;
+
+@Immutable
+public final class RangeSearchOperator extends SearchOperator {
+
+    private final List<SearchPath> path;
+    @Nullable private final BsonValue gt;
+    @Nullable private final BsonValue gte;
+    @Nullable private final BsonValue lt;
+    @Nullable private final BsonValue lte;
+
+    RangeSearchOperator(final List<SearchPath> path, @Nullable final String index,
+            @Nullable final BsonValue gt, @Nullable final BsonValue gte, @Nullable final BsonValue lt, @Nullable final BsonValue lte) {
+        super(index);
+        this.path = notNull("path", path);
+        this.gt = gt;
+        this.gte = gte;
+        this.lt = lt;
+        this.lte = lte;
+    }
+
+    public List<SearchPath> getPath() {
+        return path;
+    }
+
+    public RangeSearchOperator gt(Number value) {
+        return new RangeSearchOperator(path, getIndex(), numberToBsonValue(value), gte, lt, lte);
+    }
+
+    public RangeSearchOperator gte(Number value) {
+        return new RangeSearchOperator(path, getIndex(), gt, numberToBsonValue(value), lt, lte);
+    }
+
+    public RangeSearchOperator lt(Number value) {
+        return new RangeSearchOperator(path, getIndex(), gt, gte, numberToBsonValue(value), lte);
+    }
+
+    public RangeSearchOperator lte(Number value) {
+        return new RangeSearchOperator(path, getIndex(), gt, gte, lt, numberToBsonValue(value));
+    }
+
+    @Override
+    public RangeSearchOperator index(final String index) {
+        return new RangeSearchOperator(path, index, gt, gte, lt, lte);
+    }
+
+    @Override
+    public <TDocument> BsonDocument toBsonDocument(final Class<TDocument> documentClass, final CodecRegistry codecRegistry) {
+        BsonDocument searchStageDocument = new BsonDocument();
+        appendCommonFields(searchStageDocument);
+        BsonDocument rangeSearchOperator = new BsonDocument();
+
+        appendPath(rangeSearchOperator, path);
+
+        if (gt != null) {
+            rangeSearchOperator.append("gt", gt);
+        }
+
+        if (gte != null) {
+            rangeSearchOperator.append("gte", gte);
+        }
+
+        if (lt != null) {
+            rangeSearchOperator.append("lt", lt);
+        }
+
+        if (lte != null) {
+            rangeSearchOperator.append("lte", lte);
+        }
+
+        searchStageDocument.append("range", rangeSearchOperator);
+
+        return searchStageDocument;
+    }
+
+    private static BsonValue numberToBsonValue(final Number value) {
+        if (value instanceof Integer) {
+            return new BsonInt32((Integer) value);
+        } else if (value instanceof Long) {
+            return new BsonInt64((Long) value);
+        } else if (value instanceof Double) {
+            return new BsonDouble((Double) value);
+        } else if (value instanceof Decimal128) {
+            return new BsonDecimal128((Decimal128) value);
+        } else {
+            throw new IllegalArgumentException("Unsupported number subclass: " + value.getClass());
+        }
+    }
+}
--- a/driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
+++ b/driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.search;
+
+import com.mongodb.annotations.Immutable;
+import com.mongodb.lang.Nullable;
+import org.bson.BsonArray;
+import org.bson.BsonDocument;
+import org.bson.BsonString;
+import org.bson.conversions.Bson;
+
+import java.util.List;
+
+/**
+ *
+ * @since 4.4
+ */
+@Immutable
+public abstract class SearchOperator implements Bson {
+    @Nullable
+    private final String index;
+
+    SearchOperator(@Nullable final String index) {
+        this.index = index;
+    }
+
+    @Nullable
+    public String getIndex() {
+        return index;
+    }
+
+    abstract public SearchOperator index(final String index);
+
+    void appendCommonFields(final BsonDocument searchStageDocument) {
+        if (index != null) {
+            searchStageDocument.append("index", new BsonString(index));
+        }
+    }
+
+    void appendPath(final BsonDocument searchOperatorDocument, final List<SearchPath> path) {
+        if (path.size() == 1) {
+            searchOperatorDocument.append("path", path.get(0).toBsonValue());
+        } else {
+            searchOperatorDocument.append("path",
+                    path.stream().map(SearchPath::toBsonValue).collect(BsonArray::new, BsonArray::add, BsonArray::addAll));
+        }
+    }
+
+    void appendScore(final BsonDocument searchOperatorDocument, @Nullable final SearchScore score) {
+        if (score != null) {
+            searchOperatorDocument.append("score", score.toBsonValue());
+        }
+    }
+}
--- a/driver-core/src/main/com/mongodb/client/model/search/SearchOperators.java
+++ b/driver-core/src/main/com/mongodb/client/model/search/SearchOperators.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.search;
+
+import org.bson.BsonBoolean;
+import org.bson.BsonObjectId;
+import org.bson.types.ObjectId;
+
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Collections.singletonList;
+
+/**
+ * Builder API for Atlas search.
+ *
+ * @since 4.4
+ * @mongodb.driver.manual.atlas/reference/atlas-search/query-syntax/
+ */
+public interface SearchOperators {
+
+    /**
+     * A builder for the {@code text} search operator.
+     *
+     * @param query the query 
+     * @param path the path
+     * @return an instance of {code TextSearchOperator} with the given query and path
+     * @mongodb.driver.manual.atlas reference/atlas-search/text/
+     */
+    static TextSearchOperator text(final String query, final SearchPath path) {
+        return new TextSearchOperator(singletonList(query), singletonList(path), null, null);
+    }
+
+    /**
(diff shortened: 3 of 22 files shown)


driver-core/src/main/com/mongodb/client/model/search/RangeSearchOperator.java
```diff
@@ -0,0 +1,120 @@
...
+    @Nullable private final BsonValue gt;
+    @Nullable private final BsonValue gte;
+    @Nullable private final BsonValue lt;
+    @Nullable private final BsonValue lte;
+
+    RangeSearchOperator(final List<SearchPath> path, @Nullable final String index,
+            @Nullable final BsonValue gt, @Nullable final BsonValue gte, @Nullable final BsonValue lt, @Nullable final BsonValue lte) {
+        super(index);
+        this.path = notNull("path", path);
+        this.gt = gt;
```
jyemin, line 47: We could do error checking here to ensure that only one of gt and gte and one of lt and lte are set.  Or we could just allow the server to error since it's a runtime error anyway.

driver-core/src/main/com/mongodb/client/model/search/RangeSearchOperator.java
```diff
@@ -0,0 +1,120 @@
...
+        this.gte = gte;
+        this.lt = lt;
+        this.lte = lte;
+    }
+
+    public List<SearchPath> getPath() {
+        return path;
+    }
+
+    public RangeSearchOperator gt(Number value) {
```
jyemin, line 57: Or... we could null out gte if gt is set, and vice versa.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -0,0 +1,68 @@
...
+import org.bson.conversions.Bson;
+
+import java.util.List;
+
+/**
+ *
+ * @since 4.4
+ */
+@Immutable
+public abstract class SearchOperator implements Bson {
```
jyemin, line 33: Base class for all search operators.  It doesn't pull a lot of weight...  but I want it to be an abstract class rather than an interface so that we can lock down the ability to subclass by keeping the constructor non-public.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -0,0 +1,68 @@
...
+    @Nullable
+    private final String index;
+
+    SearchOperator(@Nullable final String index) {
+        this.index = index;
+    }
+
+    @Nullable
+    public String getIndex() {
```
jyemin, line 42: Could it bite us to assume that all operators accept an index?

driver-core/src/main/com/mongodb/client/model/search/SearchOperators.java
```diff
@@ -0,0 +1,96 @@
...
+
+import static java.util.Collections.singletonList;
+
+/**
+ * Builder API for Atlas search.
+ *
+ * @since 4.4
+ * @mongodb.driver.manual.atlas/reference/atlas-search/query-syntax/
+ */
+public interface SearchOperators {
```
jyemin, line 34: The main entry point of the API.

driver-core/src/main/com/mongodb/client/model/search/SearchOperators.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * A builder for the {@code text} search operator.
+     *
+     * @param query the query 
+     * @param path the path
+     * @return an instance of {code TextSearchOperator} with the given query and path
+     * @mongodb.driver.manual.atlas reference/atlas-search/text/
+     */
+    static TextSearchOperator text(final String query, final SearchPath path) {
```
jyemin, line 44: The basic pattern is that the factory method(s) for each operator take only the required values as parameters, and optional values are set as methods on each operator.  The one awkward thing about it is that both query and path can be either a single value or a list, which would require four overloads for this operator.  I chose instead to just offer two of the four and require use of the second overload (below) if either query or path require a list.

driver-core/src/main/com/mongodb/client/model/search/SearchOperators.java
```diff
@@ -0,0 +1,96 @@
...
+        return new TextSearchOperator(singletonList(query), singletonList(path), null, null);
+    }
+
+    /**
```
jyemin, line 48: This gives a taste of what I'm thinking the Javadoc will look like for each operator.

driver-core/src/main/com/mongodb/client/model/search/SearchScore.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.mongodb.client.model.search;
+
+import com.mongodb.annotations.Immutable;
+import org.bson.BsonValue;
+
+@Immutable
+public abstract class SearchScore {
+    public static BoostSearchScore boost(final double value) {
```
jyemin, line 24: I chose to create a subclass for each score option rather than have conditional logic.

driver-core/src/main/com/mongodb/client/model/search/TextSearchOperator.java
```diff
@@ -0,0 +1,128 @@
...
+import org.bson.BsonDocument;
+import org.bson.BsonString;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import java.util.Collections;
+import java.util.List;
+
+import static com.mongodb.assertions.Assertions.notNull;
+
+/**
```
jyemin, line 31: This class gives a flavor of what I expect the Javadoc to look like for each operator class.

driver-core/src/test/functional/com/mongodb/client/model/SearchOperatorsIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+public class SearchOperatorsIntegrationTest {
+    private static Cluster cluster;
+    private static CollectionHelper<BsonDocument> collectionHelper;
+
+    @BeforeAll
+    public static void beforeAll() {
+        assumeTrue(System.getProperties().containsKey("org.mongodb.test.search.uri"));
+        cluster = createCluster(new ConnectionString(System.getProperty("org.mongodb.test.search.uri")));
```
jyemin, line 55: Create a cluster using the new system property instead of the default one.

driver-core/src/test/unit/com/mongodb/client/model/search/SearchOperatorsTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.bson.BsonDouble;
+import org.bson.BsonString;
+import org.junit.jupiter.api.Test;
+
+import static com.mongodb.client.model.search.SearchPath.path;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class SearchOperatorsTest {
+    @Test
+    public void testTextSearchOperator() {
```
jyemin, line 30: This gives a flavor of the unit tests we should have for all the options of all the operators.  The assertions are that each operator produce the correct BSON document.   I start with an assertion based on only the required values, then I configure all the options and assert again.  I think this should cover all the null checks required in `toBsonDocument`.

driver-core/src/main/com/mongodb/client/model/search/SearchOperators.java
```diff
@@ -0,0 +1,96 @@
...
+    static AutoCompleteSearchOperator autoComplete(final String query, final SearchPath path) {
+        return new AutoCompleteSearchOperator(singletonList(query), singletonList(path), null, null, null, null);
+
+    }
+
+    static AutoCompleteSearchOperator autoComplete(final List<String> query, final List<SearchPath> path) {
+        return new AutoCompleteSearchOperator(query, path, null, null, null, null);
+    }
+
+    static CompoundSearchOperator compound(final Map<CompoundSearchOperator.Clause, List<SearchOperator>> clauses) {
```
jyemin, line 93: This API is awkward without Java 9's `Map.of` methods, but it's usable.

rozza (commented):
After a quick review I have some design questions because the API and feature seem relatively complex.  I haven't used it before so have no familiarity to fall back on.

Main question, is the API expressive enough for a user to use concisely and equally important is it easy to translate between the Document style command and this API ?

If I take a subsection from the example provided in the initial comment:

```
{
"compound": {
    "filter": [
        {
            "text": {
                "query": "Ford",
                "path": "ymmt.make"
            }
        },
        {
            "compound": {
                "should": [
                    {
                        "text": {
                            "query": "Focus",
                            "path": "ymmt.model"
                        }
                    },
                    {
                        "text": {
                            "query": "Taurus",
                            "path": "ymmt.model"
                        }
                    },
                    {
                        "text": {
                            "query": "F-150",
                            "path": "ymmt.model"
                        }
                    }
                ],
                "minimumShouldMatch": 1
            }
        }
    ]
}
```
I think it equates to:

```compound(Map.of(SHOULD,
  List.of(
    text("Ford", path("ymmt.make")),
    text(asList("Focus", "Taurus", "F-150"), asList(path("ymmt.model")))
  ))).minimumShouldMatch(2)),
```

The challenge I have is I see different terms in the JSON - essentially I read it something like:

```
compound
  .filter([  
    text,  
    compound
       .should([ text, text, text])
       .minimumShouldMatch(..)
])
```

A couple of other notes, you call out a cost of the immutable API but is there a benefit for an immutable builder API?  Doesn't seem to have much benefits, especially when it is chainable.

Do all the getters need to be public - are they for end users or just for the API? Can we reduce the size of the API by removing them?

Ross

jyemin (commented):
First of all, apologies on the PR description:  I was provided with both an original and improved JSON version of the query, but I copied the original version into the description, and then translated the improved version into Java.  So they don't match.  I update the JSON query in the description to match the one I translated from.  Please have a look and see if that clears up your question about the translation.

> Is there a benefit for an immutable builder API?

I'm not completely sold on it, but it does open up some use cases that wouldn't be available if the operators were mutable.  For instance, an application could create a template for a query and then pass it somewhere else to be modified, without having to worry about the original being modified.  Alternatively, we could implement a custom copy/clone method (like we have with BsonDocument) to help users that want to modify a query without messing with the original.  Or we can just say we don't support use cases like that.  I opted for immutability because cloning trees is error prone, and the cost of immutability doesn't seem high.  If you're chaining you probably wouldn't even notice.

> Do all the getters need to be public - are they for end users or just for the API? 

Not sure what you mean about being for end users or just for the API.  That seems like the same thing to me.  Can you clarify?

The getters don't need to be public in order to implement the toBsonDocument methods, which is all that the driver needs.  But it felt weird to define public classes and then not provide getters for the properties.  On the other hand, we have done that in other places: e.g. FindIterable doesn't have public accessors for its properties and no one has ever asked for them.  The downside of the getters is that they show up in auto-complete.

driver-core/src/main/com/mongodb/client/model/search/SearchOperators.java
rozza: I'm still not used to it, feels a miss use of an interface as its not designed to be implemented.

rozza:
Sorry I wasn't clear, I meant to ask if the getters were expected to be used by the end users and therefore need to be public. As far as I can tell we don't generally add getters to builders.

> The getters don't need to be public in order to implement the toBsonDocument methods, which is all that the driver needs. But it felt weird to define public classes and then not provide getters for the properties. On the other hand, we have done that in other places: e.g. FindIterable doesn't have public accessors for its properties and no one has ever asked for them. The downside of the getters is that they show up in auto-complete.

I think if there isn't a clear usecase for them, then they should be removed to help with the readability and consistency of the builders API.

---

Looking at the updated Json & code - it does work as it reads eg:

```
{
    "compound": {
        "should": [
            {
                "text": {
                    "query": "Ford",
                    "path": "ymmt.make"
                }
            },
            {
                "text": {
                    "query": ["Focus", "Taurus", "F-150"],
                    "path": "ymmt.model"
                }
            },
        ],
        "minimumShouldMatch": 2
    }
}
```

Equates to:

```
compound(
  Map.of(SHOULD, 
    List.of(
        text("Ford", path("ymmt.make")),
        text(asList("Focus", "Taurus", "F-150"), asList(path("ymmt.model"))) // <- I think the asList(path(...)) is incorrect?
    ))
).minimumShouldMatch(2)),
```

That makes much more sense :+1: and it was just the json / code being out of sync. That said I still think: 

```
compound()
  .should(asList(
    text("Ford", path("ymmt.make")),
    text(asList("Focus", "Taurus", "F-150"), path("ymmt.model")))
  )
  .minimumShouldMatch(2)
```

Is an improvement - as auto completion would be there for the methods as opposed to being lost when using a Map.

driver-core/src/main/com/mongodb/client/model/search/SearchOperators.java
jyemin: I can't find the JEP for this feature, but I think that this use case was one of the motivations.

rozza: :+1:

jyemin:
@rozza replaced the Map with explicit methods for must, mustNot, should, and filter.  PTAL.

Waiting for @stIncMale to weigh in on the question of whether to include getters.  I'm torn.

rozza:
Think this is looking good - was surprised that there were no test changes for the last change.

driver-core/src/main/com/mongodb/client/model/search/SearchOperators.java
stIncMale: > improvement over a final class in that it expresses intent more clearly

The intent is to have a place with a bunch of `static` methods. A `final` class with no constructor clearly expresses this intent, while an interface does not, because an interface defines a type and expresses an intent to for this type to be implemented.

I had a similar conversation with @rozza here https://github.com/mongodb/mongo-spark/pull/54#discussion_r716296848, and instead of copying it, I am linking it.

stIncMale:
> Waiting for @stIncMale to weigh in on the question of whether to include getters. I'm torn.

@jyemin 
I concur with the idea that unless a program element (including methods) is intended to be used by the API uses, we should try to not expose it as API. This idea sometimes conflicts with evolvability of the API in a backward-compatible way, so finding the balance is not necessary straightforward. But in our case, I think that not only getters, but the concrete classes, e.g., `ConstantSearchScore`, that have them should be not exposed as API.

You may see that for `Window`/`WindowedComputation` neither implementations nor ant getters are exposed.

stIncMale:
> @stIncMale: I think if we decide to stick with this approach, we should additionally improve it by not exposing specific types that don't currently have any optional fields. An example is SearchScore: neither `BoostSearchScore` nor `ConstantSearchScore` have any optional fields, which means we do not need to expose these types. If any of them gains an optional field in the future, we can expose the type without breaking compatibility.

The _"we can expose the type without breaking compatibility"_ part is incorrect, doing so [breaks binary compatibility](https://stackoverflow.com/a/47478105/1285873).

---

> @jyemin: This proposal has a significant downside though: a final call to .build() for every builder ... I'm not sure that will be appreciated.

I agree.

---

> > @stIncMale: One positive side of using specific builders, though, is that one will not be able to create an invalid specific SearchOperator, while with the current approach it is possible.

> @jyemin: Can you clarify why you think so? With the current approach that is not possible, as the factory method for each operator takes all required parameters. Chaining is only for the optional ones.

Optional parameters may be allowed or not allowed based on the presence of other optional parameters. With the chaining approach we don't have a place similar to a `build` method where we could check the rules before creating an object. We still may check for such rules after an object has been created, e.g., in the `Aggregates.search` method, but by that time the process could have spread an invalid object throughout its memory, e.g., cached, which makes the checks useless. That's not a big deal, unless our goal is to guarantee that user cannot build an invalid request, and there is no such goal.

jyemin:
This PR has been superseded by https://github.com/mongodb/mongo-java-driver/pull/891.

PR #779, opened 2021-09-03
Fix manifest import-package imports

rozza:
Refs: #771 Using the suggested method by @stbischof

Fixes imports in `mongo-scala-bson`, `mongodb-driver-sync`, `mongodb-driver-reactivestreams`, `mongodb-driver-core`.

JAVA-4274

### DIFF
--- a/driver-core/build.gradle
+++ b/driver-core/build.gradle
@@ -64,20 +64,14 @@ afterEvaluate {
     jar.manifest.attributes['Automatic-Module-Name'] = 'org.mongodb.driver.core'
     jar.manifest.attributes['Bundle-SymbolicName'] = 'org.mongodb.driver-core'
     jar.manifest.attributes['Import-Package'] = [
-        'org.bson.*', // unfortunate that this is necessary, but if it's left out then it's not included
-        'javax.crypto.*',
-        'javax.management.*',
-        'javax.naming.*',
-        'javax.net.*',
-        'jdk.net.*',
-        'javax.security.sasl.*',
-        'javax.security.auth.callback.*',
-        'org.ietf.jgss.*',
-        'io.netty.*;resolution:=optional',
-        'org.xerial.snappy.*;resolution:=optional',
-        'com.github.luben.zstd.*;resolution:=optional',
-        'org.slf4j.*;resolution:=optional',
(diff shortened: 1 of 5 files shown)


driver-core/build.gradle
```diff
@@ -64,20 +64,14 @@ afterEvaluate {
...
-        'javax.security.sasl.*',
-        'javax.security.auth.callback.*',
-        'org.ietf.jgss.*',
-        'io.netty.*;resolution:=optional',
-        'org.xerial.snappy.*;resolution:=optional',
-        'com.github.luben.zstd.*;resolution:=optional',
-        'org.slf4j.*;resolution:=optional',
-        'jnr.unixsocket.*;resolution:=optional',
-        'com.mongodb.crypt.capi.*;resolution:=optional'
+            '!sun.misc.*',  // Used by DirectBufferDeallocator only for java 8
```
jyemin, line 67: Even given the comment, I'm not clear on why these need to be explicitly excluded.

jyemin (commented):
Can you post a comment that shows the diff between the generated old and new manifests?

driver-core/build.gradle
rozza: I removed them because they weren't in the original and due to the forking logic between jvm versions. Java 8 and Java 9+ work differently.

rozza:
Added imports to import packages:

* mongodb-driver-core
  - javax.annotation
  - javax.annotation.meta
  - javax.security.auth
  - javax.security.auth.kerberos
  - javax.security.auth.login

* mongodb-driver-sync:
  -  javax.naming
  - javax.naming.spi 
  - javax.net.ssl

* mongodb-driver-reactivestreams:
  - javax.net.ssl
  - org.reactivestreams

* mongodb-driver-scala:
  - org.mongodb.scala.bson
  - org.mongodb.scala.bson.codecs
  - org.mongodb.scala.bson.collection.immutable
  - org.reactivestreams

jyemin:
Huh, do you know why these missing imports didn't also break OSGi apps?

jyemin (approved):
LGTM

stbischof:
It does break my app. But the one that use OSGi know how to quick fix this. E.g. fragment hosts or other bad tricks.

PR #780, opened 2021-09-04
Allow to update the whole bson document

yarosman:


jyemin:
Apologies for letting this sit for so long with no response.  It slipped through the crack due to the lack of an associated Jira issue.

The intention of the API is to rely on `Updates.combine` for updates to different fields, e.g.

```
combine(set("x",  6), set("y", 4))
```

I would prefer not to introduce another way of doing the same thing.

I'm going to close this, but if you'd like to make your case for it please open a Jira issue to discuss the merits.

yarosman:
> Apologies for letting this sit for so long with no response. It slipped through the crack due to the lack of an associated Jira issue.
> 
> The intention of the API is to rely on `Updates.combine` for updates to different fields, e.g.
> 
> ```
> combine(set("x",  6), set("y", 4))
> ```
> 
> I would prefer not to introduce another way of doing the same thing.
> 
> I'm going to close this, but if you'd like to make your case for it please open a Jira issue to discuss the merits.

But if you have object with 20 fields, and manually write set for each field looks very strange

PR #786, opened 2021-09-22
Test for null byte in BSON cstring

jyemin:
The driver already has the necessary checks. This is just adding tests

JAVA-4236

### DIFF
--- a/bson/src/main/org/bson/BsonDocument.java
+++ b/bson/src/main/org/bson/BsonDocument.java
@@ -735,10 +735,6 @@ public class BsonDocument extends BsonValue implements Map<String, BsonValue>, C
         if (value == null) {
             throw new IllegalArgumentException(format("The value for key %s can not be null", key));
         }
-        if (key.contains("\0")) {
-            throw new BSONException(format("BSON cstring '%s' is not valid because it contains a null character at index %d", key,
-                                           key.indexOf('\0')));
-        }
         return map.put(key, value);
     }
 
--- a/bson/src/test/unit/org/bson/BsonBinaryWriterTest.java
+++ b/bson/src/test/unit/org/bson/BsonBinaryWriterTest.java
@@ -33,6 +33,7 @@ import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.fail;
 
 public class BsonBinaryWriterTest {
@@ -358,12 +359,7 @@ public class BsonBinaryWriterTest {
     public void testNullByteInTopLevelName() {
         writer.writeStartDocument();
         writer.writeName("a\u0000b");
-        try {
-            writer.writeBoolean(true);
-            fail();
-        } catch (BsonSerializationException e) {
-            // expected
-        }
+        assertThrows(BsonSerializationException.class, () -> writer.writeBoolean(true));
     }
 
     @Test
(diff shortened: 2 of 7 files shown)


bson/src/test/unit/org/bson/GenericBsonTest.java
```diff
@@ -261,7 +262,10 @@ private void runParseError() {
             }
         } else if (testDefinitionDescription.startsWith("Top-level") || testDefinitionDescription.startsWith("Binary type")) {
             try {
-                parse(str);
+                BsonDocument document = parse(str);
+                if (isTestOfNullByteInCString(description)) {
```
jyemin, line 266: This is kind of gross, but required to get the tests to pass for drivers whose JSON parsers don't include the null byte check: https://github.com/mongodb/specifications/commit/ff5d91fea80b9dcb40fb2fb3533fd1bbcee818b0

bson/src/test/unit/org/bson/BsonBinaryWriterTest.java
```diff
@@ -354,6 +354,56 @@ public void testWriteDBPointer() {
...
 
+    @Test
+    public void testNullByteInTopLevelName() {
+        writer.writeStartDocument();
+        writer.writeName("a\u0000b");
+        try {
+            writer.writeBoolean(true);
+            fail();
+        } catch (BsonSerializationException e) {
+            // expected
```
stIncMale, line 365: Let's use `@Test(expected = BsonSerializationException.class)` instead.

bson/src/test/unit/org/bson/GenericBsonTest.java
stIncMale: Would any tests fail if we were to call `encodeToHex(document)` in all of them? If no, then we could remove the condition here and leave it only where the error is checked.

stIncMale (changes requested):
Mostly questions to better understand what we are doing and whether it is actually what we want to do.

bson/src/test/unit/org/bson/BsonBinaryWriterTest.java
stIncMale: Resolving, because when `Test.expected` is used, an exception may come from a line different than that being tested. @jyemin pointed this out, and I agree.

jyemin: Talked it over, decide on `assertThrows` instead.

Done.

bson/src/test/unit/org/bson/GenericBsonTest.java
jyemin: All tests passes either way.  I had it this way more to be clear on the intentions. But it seems like it just added to the confusion so I removed it.

Done.

bson/src/main/org/bson/BsonDocument.java
```diff
@@ -735,10 +735,6 @@ public BsonValue put(final String key, final BsonValue value) {
         if (value == null) {
             throw new IllegalArgumentException(format("The value for key %s can not be null", key));
         }
-        if (key.contains("\0")) {
```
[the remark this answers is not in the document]
jyemin: > My understanding is that with this change it is now allowed to create a BsonDocument that violate the BSON specification.

Yes

> Have this already been the case regarding some other checks, of was it previously the case that if I have a BsonDocument, it is guaranteed to be writable to BSON?

It wasn't the case, since the checks are not include in BsonRegularExpression

> If there was an explicit decision to allow creating BsonDocuments that violate the BSON specification, then it's all fine, of course.

Essentially there was, since many drivers just rely on native dictionaries to represent BSON documents.

Including the check doesn't improve safety, since we have the check where it's really needed.  And there is a small performance cost to including it since we have to scan every key to check. It arguably improves things for users to include the check since they will find their mistake earlier, but it's still a runtime error any way you slice it.

For all these reasons I'm comfortable with this change.  I wouldn't have done it just to make it so we can run the new tests, but it's a nice benefit.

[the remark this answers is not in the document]
stIncMale: Thanks for explaining 👍

stIncMale (approved):

PR #788, opened 2021-09-24
[JAVA-4311] Change ToSingleObservableVoid -> ToSingleObservableUnit so that operations returning Publisher[Void] can be composed using monadic operations

scottrice10:
[JAVA-4311](https://jira.mongodb.org/browse/JAVA-4311)


rozza:
Hi @scottrice10 

Thanks for the PR - as this is a breaking change it will have to wait until the 5.0 release. I added a `completeWithUnit` helper for `Observable[T]` to convert them to `Observable[Unit]` until 5.0 is released please use that helper, so that composition can happen.

Ross

PR #790, opened 2021-09-27
Support callback for AWS credentials

jyemin:
JAVA-4310

### DIFF
--- a/.evergreen/run-mongodb-aws-test.sh
+++ b/.evergreen/run-mongodb-aws-test.sh
@@ -35,4 +35,5 @@ set -x
 
 echo "Running tests with ${JDK}"
 ./gradlew -version
-./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI} --stacktrace --debug --info driver-core:test --tests AwsAuthenticationSpecification
+./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI} --stacktrace --debug --info --no-build-cache \
+driver-core:cleanTest driver-core:test --tests AwsAuthenticationSpecification
--- a/.evergreen/run-mongodb-aws-test.sh
+++ b/.evergreen/run-mongodb-aws-test.sh
@@ -35,5 +35,8 @@ set -x
 
 echo "Running tests with ${JDK}"
 ./gradlew -version
+
+# As this script may be executed multiple times in a single task, with different values for MONGODB_URI, it's necessary
+# to run cleanTest to ensure that the test actually executes each run
 ./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI} --stacktrace --debug --info --no-build-cache \
 driver-core:cleanTest driver-core:test --tests AwsAuthenticationSpecification
--- a/driver-core/src/main/com/mongodb/MongoCredential.java
+++ b/driver-core/src/main/com/mongodb/MongoCredential.java
@@ -16,6 +16,7 @@
 
 package com.mongodb;
 
+import com.mongodb.annotations.Beta;
 import com.mongodb.annotations.Immutable;
 import com.mongodb.lang.Nullable;
 
@@ -174,6 +175,7 @@ public final class MongoCredential {
      * @see AwsCredential
      * @since 4.4
      */
+    @Beta
     public static final String AWS_CREDENTIAL_PROVIDER_KEY = "AWS_CREDENTIAL_PROVIDER";
 
     /**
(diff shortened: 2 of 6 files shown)


.evergreen/run-mongodb-aws-test.sh
```diff
@@ -35,4 +35,5 @@ set -x
 
 echo "Running tests with ${JDK}"
 ./gradlew -version
-./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI} --stacktrace --debug --info driver-core:test --tests AwsAuthenticationSpecification
+./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI} --stacktrace --debug --info --no-build-cache \
+driver-core:cleanTest driver-core:test --tests AwsAuthenticationSpecification
```
jyemin, line 39: Necessary because the evergreen config calls this script multiple times in successive functions, with different URIs, and without some sort of clean here all but the first call becomes a no-op.  This isn't a new problem, but I noticed it when debugging this new feature.

driver-core/src/main/com/mongodb/internal/connection/AwsAuthenticator.java
```diff
@@ -77,12 +83,24 @@ protected SaslClient createSaslClient(final ServerAddress serverAddress) {
 
     private static class AwsSaslClient implements SaslClient {
         private final MongoCredential credential;
+        @Nullable
+        private final AwsCredential credentialFromSupplier;
         private final byte[] clientNonce = new byte[RANDOM_LENGTH];
         private int step = -1;
         private String httpResponse;
 
         AwsSaslClient(final MongoCredential credential) {
```
stIncMale, line 92: For the new mechanism based on `Supplier<AwsCredential>` to work, a new `AwsSaslClient` must be created each time authentication is being done. However, it seems that at least `ScramShaAuthenticator` stores it in its field `speculativeSaslClient` and reuses in `createSpeculativeAuthenticateCommand`. I may misunderstand the situation.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/connection/AwsAuthenticator.java
jyemin: It's not strictly required that we get new credentials for every single authentication attempt.  These are not one-time passwords.  So even if this is true, it's a short-lived situation that only lasts the length of a single handshake.

driver-core/src/main/com/mongodb/MongoCredential.java
```diff
@@ -287,14 +314,13 @@ public static MongoCredential createGSSAPICredential(final String userName) {
     /**
      * Creates a MongoCredential instance for the MONGODB-AWS mechanism.
      *
-     * @param userName the user name
-     * @param password the user password
+     * @param userName the user name, which may be null.  This maps to the AWS accessKeyId
+     * @param password the user password, which may be null if the userName is also null.  This maps to the AWS secretAccessKey.
```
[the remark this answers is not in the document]
jyemin: I'd rather not make this change now.  The way AwsAuthenticator is written, it could actually work the way it is, and I don't want to risk introducing a breaking change.  But I'll check the spec and open a separate ticket if we're missing validation.

driver-core/src/main/com/mongodb/MongoCredential.java
```diff
@@ -287,14 +314,13 @@ public static MongoCredential createGSSAPICredential(final String userName) {
...
      * @return the credential
      * @since 4.1
      * @mongodb.server.release 4.4
-     * @see #createCredential(String, String, char[])
      */
-    public static MongoCredential createAwsCredential(final String userName, final char[] password) {
+    public static MongoCredential createAwsCredential(@Nullable final String userName, @Nullable final char[] password) {
```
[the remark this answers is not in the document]
jyemin: I added `@see` javadoc for `withMechanismProperty` as well as the two mechanism keys that you can use with an AWS credential.  The Javadoc for the `AWS_CREDENTIAL_PROVIDER_KEY` in turn links to `AwsCredential`, which seems a more appropriate place for that link since it's the only way to actually use `AwsCredential`.

.evergreen/run-mongodb-aws-test.sh
jyemin: Good idea. Done.

driver-core/src/main/com/mongodb/internal/connection/AwsAuthenticator.java
stIncMale: > it's a short-lived situation that only lasts the length of a single handshake

👍

stIncMale (approved):

PR #791, opened 2021-09-30
Execute aggregate with $out/$merge on secondaries

jyemin:
JAVA-3994

This implementation uses the recommended approach from the specification:

> Drivers SHOULD use a custom server selector to consider server/wire version when matching a read preference and, if a pre-5.0 secondary would be selected, fall back to selecting a primary. With this approach, only a single server selection attempt is needed.

This required some minor surgery on the ReadBinding and ConnectionSource internal APIs.

### DIFF
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/VoidWriteOperationThenCursorReadOperation.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/VoidWriteOperationThenCursorReadOperation.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.reactivestreams.client.internal;
+
+import com.mongodb.internal.async.AsyncBatchCursor;
+import com.mongodb.internal.async.SingleResultCallback;
+import com.mongodb.internal.binding.AsyncReadBinding;
+import com.mongodb.internal.binding.AsyncWriteBinding;
+import com.mongodb.internal.operation.AsyncReadOperation;
+import com.mongodb.internal.operation.AsyncWriteOperation;
+
+class VoidWriteOperationThenCursorReadOperation<T> implements AsyncReadOperation<AsyncBatchCursor<T>> {
+    private final AsyncWriteOperation<Void> writeOperation;
+    private final AsyncReadOperation<AsyncBatchCursor<T>> cursorReadOperation;
+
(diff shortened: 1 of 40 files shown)


driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/VoidWriteOperationThenCursorReadOperation.java
```diff
@@ -23,32 +23,23 @@
 import com.mongodb.internal.operation.AsyncReadOperation;
 import com.mongodb.internal.operation.AsyncWriteOperation;
 
-class WriteOperationThenCursorReadOperation<T> implements AsyncReadOperation<AsyncBatchCursor<T>> {
-    private final AsyncWriteOperation<Void> aggregateToCollectionOperation;
-    private final AsyncReadOperation<AsyncBatchCursor<T>> readOperation;
+class VoidWriteOperationThenCursorReadOperation<T> implements AsyncReadOperation<AsyncBatchCursor<T>> {
```
jyemin, line 26: Did a little cleanup here to match the new class that was added just above.

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -413,13 +413,18 @@ private void initClient(final BsonDocument entity, final String id,
     private void initDatabase(final BsonDocument entity, final String id) {
         MongoClient client = clients.get(entity.getString("client").getValue());
         MongoDatabase database = client.getDatabase(entity.getString("databaseName").getValue());
-        if (entity.containsKey("collectionOptions")) {
-            for (Map.Entry<String, BsonValue> entry : entity.getDocument("collectionOptions").entrySet()) {
-                //noinspection SwitchStatementWithTooFewBranches
+        if (entity.containsKey("databaseOptions")) {
```
jyemin, line 416: This was a straight-up bug.  We were looking at the wrong options.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -116,11 +117,14 @@ static WriteConcern asWriteConcern(final BsonDocument writeConcernDocument) {
     }
 
     public static ReadPreference asReadPreference(final BsonDocument readPreferenceDocument) {
-        if (readPreferenceDocument.size() > 1) {
-            throw new UnsupportedOperationException("Unsupported read preference properties");
+        if (readPreferenceDocument.size() == 1) {
```
jyemin, line 120: For some reason the new tests use `maxStalenessSeconds` so had to add support for it.

jyemin:
> I don't like the conflation of the ReadOperation being used to perform Writes. Seems to me to be a bad code smell.

I don't like it either.  It's a new requirement, and one that I'm not sure will ever be repeated.  Though in principle you could imagine a change to the server where a client can send a write operation to any replica set member, and it will just "do the right thing".  But for now, it's niche.

> I realise that its for a single Operation but without the documentation of why it is in that operation, I can imagine it's going to be confusing when visiting the code in n months time.

I'm not sure what you mean.  Is there something you think should be done here? LIke write some documentation?  I see one specific request for docs, which I will do.

> Did you consider adding a new custom operation type?'

I entertained a bunch of alternatives, but a custom operation type was not one of them.  I was trying to avoid something too invasive.  The one thing I did consider was to add another implementation of ReadBinding that was written specifically for AggregateToCollectionOperation.  That way, we could just make that implementation do what's needed to get the ConnectionSource.  Could also consider doing that instead for WriteBinding, and leave AggregateToCollectionOperation alone completely.  The complication is that we'd then need to add some sort of Read/WriteBinding factory that was dependent on the specific Operation sub-class it's being used with.  That's where I got stuck and decided to do it this way.  

Let me know if you want me to spike something like that.

PR #792, opened 2021-10-02
JAVA-4028 Provide explicit guidance on handling command errors that occur before the handshake completes during operation execution

stIncMale:
This PR synchronizes [`server-discovery-and-monitoring/tests/integration`](https://github.com/mongodb/specifications/tree/master/source/server-discovery-and-monitoring/tests/integration) tests (as per [JAVA-4028](https://jira.mongodb.org/browse/JAVA-4028)), which also includes changes from JAVA-4139, with accordance to the "has to be finished together with" link in Jira. The only test excluded from the synchronization is [`rediscover-quickly-after-step-down.json`](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/tests/integration/rediscover-quickly-after-step-down.json). The reason is that there was made a conscious decision to introduce a discrepancy between the original specification tests and their copy in the Java driver repo by ignoring changes in [DRIVERS-1290](https://jira.mongodb.org/browse/DRIVERS-1290) (see [JAVA-3765](https://jira.mongodb.org/browse/JAVA-3765)), and we will have to live with and continue bumping into it.

I did not discover new behavior that we need to add to the Java driver as a result of the changes in [DRIVERS-1476](https://jira.mongodb.org/browse/DRIVERS-1476) (a task that caused [JAVA-4028](https://jira.mongodb.org/browse/JAVA-4028)).

JAVA-4028

### DIFF
--- a/driver-core/src/test/functional/com/mongodb/JsonTestServerVersionChecker.java
+++ b/driver-core/src/test/functional/com/mongodb/JsonTestServerVersionChecker.java
@@ -25,6 +25,7 @@ import java.util.List;
 
 import static com.mongodb.ClusterFixture.getServerVersion;
 import static com.mongodb.ClusterFixture.getVersionList;
+import static com.mongodb.ClusterFixture.isAuthenticated;
 import static com.mongodb.ClusterFixture.isDiscoverableReplicaSet;
 import static com.mongodb.ClusterFixture.isLoadBalanced;
 import static com.mongodb.ClusterFixture.isSharded;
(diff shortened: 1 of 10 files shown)


jyemin (approved):

stIncMale:
@rozza A reminder about this PR, just in case the review request was lost in your notifications.

jyemin:
Removed @rozza.  This one doesn't require two reviewers.

PR #793, opened 2021-10-04
Ensure that either CommandFailedEvent or CommandSucceededEvent is published, but not both

jyemin:
This fixed a bug in the synchronous driver in which the driver sends a CommandSucceededEvent,
followed by a CommandFailedEvent (for the same command) in the case were the server replied
successfully but the Decoder configured to decode the response throws an exception.

JAVA-4321

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
@@ -333,15 +333,16 @@ public class InternalStreamConnection implements InternalConnection {
 
         try {
             sendCommandMessage(message, bsonOutput, sessionContext);
-            if (message.isResponseExpected()) {
-                return receiveCommandMessageResponse(decoder, commandEventSender, sessionContext, 0);
-            } else {
-                commandEventSender.sendSucceededEventForOneWayCommand();
-                return null;
-            }
         } catch (RuntimeException e) {
             commandEventSender.sendFailedEvent(e);
             throw e;
+        }
+
+        if (message.isResponseExpected()) {
+            return receiveCommandMessageResponse(decoder, commandEventSender, sessionContext, 0);
+        } else {
+            commandEventSender.sendSucceededEventForOneWayCommand();
+            return null;
         }
     }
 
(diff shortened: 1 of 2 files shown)


driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -410,26 +411,36 @@ private void sendCommandMessage(final CommandMessage message,
...
+        }
+
+        try {
             updateSessionContext(sessionContext, responseBuffers);
             if (!isCommandOk(responseBuffers)) {
-                throw getCommandFailureException(responseBuffers.getResponseDocument(responseTo, new BsonDocumentCodec()),
-                        description.getServerAddress());
+                MongoException commandFailureException = getCommandFailureException(responseBuffers.getResponseDocument(responseTo,
+                                new BsonDocumentCodec()), description.getServerAddress());
+                commandEventSender.sendFailedEvent(commandFailureException);
```
stIncMale, line 427: `sendFailedEvent` being present twice in the code does not smell good. I propose a simpler approach (it also requires less changes to the existing code): https://github.com/stIncMale/mongo-java-driver/commit/58f950c031ca3a772acaa70584618c2044570ff6.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -410,26 +411,36 @@ private void sendCommandMessage(final CommandMessage message,
     private <T> T receiveCommandMessageResponse(final Decoder<T> decoder,
                                                 final CommandEventSender commandEventSender, final SessionContext sessionContext,
                                                 final int additionalTimeout) {
-        try (ResponseBuffers responseBuffers = receiveMessageWithAdditionalTimeout(additionalTimeout)) {
+        ResponseBuffers responseBuffers;
+        try {
+            responseBuffers = receiveMessageWithAdditionalTimeout(additionalTimeout);
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
jyemin: I agree that the complexity of the additional boolean local variable is worth the simplification.  Thanks for the suggestion.  

Done

stIncMale (approved):

PR #795, opened 2021-10-05
Run serverless tests against a load balancer

jyemin:
JAVA-4279

I disabled a few prose spec tests that don't properly run in serverless.  I think some of them could be made to run given the time none of them seem like they are testing anything that is necessary to prove serverless is working.

There is also one change to the unified test runner, which required a change to the Fixture class.  I'll discuss that in an inline comment.

### DIFF
--- a/.evergreen/run-serverless-tests.sh
+++ b/.evergreen/run-serverless-tests.sh
@@ -21,14 +21,11 @@ echo "Running serverless tests with ${JDK}"
 
 export JAVA_HOME="/opt/java/jdk11"
 
-MONGODB_URI_SINGLE_HOST=${MONGODB_URI%%,*}
-SINGLE_HOST=${MONGODB_URI_SINGLE_HOST:10}
-
-MONGODB_URI_FINAL="mongodb://${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${SINGLE_HOST}/?tls=true"
-
-echo ${MONGODB_URI_FINAL}
+# assume "mongodb" protocol for single server, and "mongodb+srv" protocol for multi server
+SINGLE_SERVER_URI="mongodb://${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${SINGLE_ATLASPROXY_SERVERLESS_URI:10}"
+MULTI_SERVER_URI="mongodb+srv://${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${MULTI_ATLASPROXY_SERVERLESS_URI:14}"
 
 ./gradlew -version
 
-./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI_FINAL} -Dorg.mongodb.test.serverless=true \
-   --stacktrace --info --continue driver-sync:test
+./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${SINGLE_SERVER_URI}  -Dorg.mongodb.test.transaction.uri=${MULTI_SERVER_URI} \
+   -Dorg.mongodb.test.serverless=true --stacktrace --info --continue driver-sync:test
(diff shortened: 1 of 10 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -290,16 +290,17 @@ private void initClient(final BsonDocument entity, final String id,
...
-        MongoClientSettings.Builder clientSettingsBuilder = getMongoClientSettingsBuilder();
-        if (entity.getBoolean("useMultipleMongoses", BsonBoolean.FALSE).getValue()) {
+        MongoClientSettings.Builder clientSettingsBuilder;
+        if (entity.getBoolean("useMultipleMongoses", BsonBoolean.FALSE).getValue() && (isSharded() || isLoadBalanced())) {
             assumeTrue("Multiple mongos connection string not available for sharded cluster",
                     !isSharded() || getMultiMongosConnectionString() != null);
             assumeTrue("Multiple mongos connection string not available for load-balanced cluster",
                     !isLoadBalanced() || getMultiMongosConnectionString() != null);
-            if (isSharded() || isLoadBalanced()) {
-                clientSettingsBuilder.applyConnectionString(requireNonNull(getMultiMongosConnectionString()));
```
jyemin, line 300: This line of code is the one that precipitated the change to the method and the associated change to the Fixture class.  The problem is that the `ClusterSettings.Builder#srvHost` method has the following check:

```
    if (this.hosts != DEFAULT_HOSTS) {
         throw new IllegalArgumentException("Can not set both hosts and srvHost");
     }
```

That exception was being thrown when executing this line of code.  The reason is because on line 293 we are initializing the MongoClientSettings.Builder from the single-mongos URI, and thus `hosts != DEFAULT_HOSTS` by the time we get here.  Moreover, because it's a reference equality check and not an `equals` check, it doesn't even work to first re-set hosts to `singletonList(new ServerAddress())`

I considered a number of solutions, including changing ClusterSettings, but instead opted for this change instead, which I think is actually correct, regardless of whether we change ClusterSettings.

.evergreen/run-serverless-tests.sh
```diff
@@ -21,14 +21,11 @@ echo "Running serverless tests with ${JDK}"
 
 export JAVA_HOME="/opt/java/jdk11"
 
-MONGODB_URI_SINGLE_HOST=${MONGODB_URI%%,*}
-SINGLE_HOST=${MONGODB_URI_SINGLE_HOST:10}
-
-MONGODB_URI_FINAL="mongodb://${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${SINGLE_HOST}/?tls=true"
-
-echo ${MONGODB_URI_FINAL}
+# assume "mongodb" protocol for single server, and "mongodb+srv" protocol for multi server
```
jyemin, line 24: This is a bit of a hack, but it allows me not to have to write too much bash string manipulation code, which I despise.

rozza: Ack - and agree.

.evergreen/run-serverless-tests.sh
```diff
@@ -21,14 +21,11 @@ echo "Running serverless tests with ${JDK}"
...
+SINGLE_SERVER_URI="mongodb://${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${SINGLE_ATLASPROXY_SERVERLESS_URI:10}"
+MULTI_SERVER_URI="mongodb+srv://${SERVERLESS_ATLAS_USER}:${SERVERLESS_ATLAS_PASSWORD}@${MULTI_ATLASPROXY_SERVERLESS_URI:14}"
 
 ./gradlew -version
 
-./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${MONGODB_URI_FINAL} -Dorg.mongodb.test.serverless=true \
-   --stacktrace --info --continue driver-sync:test
+./gradlew -PjdkHome=/opt/java/${JDK} -Dorg.mongodb.test.uri=${SINGLE_SERVER_URI}  -Dorg.mongodb.test.transaction.uri=${MULTI_SERVER_URI} \
```
rozza, line 30: Should changing the name of this property be considered? It might be worth adding a ticket to either rename it to a more generic name and / or to create a custom test uri property which is more obvious in intent.

rozza (commented):
One question.

Also can you confirm the failing evergreen tests are not related to this change.

.evergreen/run-serverless-tests.sh
jyemin: Yes, definitely.

jyemin:
The test failures are not related

rozza (approved):

rozza:
LGTM

.evergreen/run-serverless-tests.sh
rozza: Cool - added JAVA-4333 to track that.

PR #797, opened 2021-10-06
Update performance benchmarks to use perf.send

jyemin:
Evergreen had deprecated json.send in favor of perf.send. See
https://github.com/evergreen-ci/evergreen/wiki/Project-Commands#perf-send.

JAVA-4323

### DIFF
--- a/driver-benchmarks/src/main/com/mongodb/benchmark/framework/EvergreenBenchmarkResultWriter.java
+++ b/driver-benchmarks/src/main/com/mongodb/benchmark/framework/EvergreenBenchmarkResultWriter.java
@@ -22,40 +22,44 @@ import org.bson.json.JsonWriterSettings;
 
 import java.io.FileWriter;
 import java.io.IOException;
-import java.io.Writer;
+import java.io.StringWriter;
 
 public class EvergreenBenchmarkResultWriter implements BenchmarkResultWriter {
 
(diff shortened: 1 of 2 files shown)


driver-benchmarks/src/main/com/mongodb/benchmark/framework/EvergreenBenchmarkResultWriter.java
```diff
@@ -22,53 +22,61 @@
...
-        System.out.println("DONE");
+
+        try (FileWriter fileWriter = new FileWriter(System.getProperty(OUTPUT_FILE_SYSTEM_PROPERTY))) {
+            fileWriter.write(getJsonResultsArrayFromJsonDocument());
+        }
     }
 
-    public String getResults() {
-        return writer.toString();
+    private String getJsonResultsArrayFromJsonDocument() {
```
jyemin, line 76: This is the only awkward part.  JsonWriter does not allow writing a top-level array, so we have to write the array as the value of a field in a document.  This piece of code just strips off the outer document and field name and returns the JSON array.

jyemin:
Evergreen won't generate the graphs in a patch build.  I'm trying to find out if there's a way to ensure that it's working.  All I can see so far is that `perf.send` succeeds.

PR #800, opened 2021-10-13
Resume change stream on any MongoClientException

jyemin:
In order to handle the reported scenario, several changes are made:

1. DefaultConnectionPool throws a MongoServerUnavailableException if the pool is closed
2. DefaultServer throws a MongoServerUnavailableException if the server is closed
3. ChangeStreamBatchCursorHelper.isResumableError returns true for any MongoClientException
   (MongoServerUnavailableException is a subclass of MongoClientException)

JAVA-4289

AFAIK this can't be integration-tested with failpoints , so unit tests are used to test the behavior.


stIncMale:
> @rozza I couldn't figure out the right way to fix the failing Scala test in ApiCompanionSpec. Can you advise?

I believe, you need to do the same that was done for `MongoConnectionPoolClearedException`: https://github.com/mongodb/mongo-java-driver/blob/master/driver-scala/src/main/scala/org/mongodb/scala/package.scala#L352-L356.

PR #804, opened 2021-10-15
Prestart connection pool async work manager when running `wait-queue-timeout.json`, `wait-queue-timeouts.json`

stIncMale:
I anticipate a push back on the approach taken in the PR, but as far as my imagination goes, it's either this approach or an approach via system properties. The main complication is that the LB test runner creates not just a pool, where we could have passed a special parameter via pool's constructor, but the whole client, and pool settings is the only way to affect a pool in this case.

Note that the new methods added to `ConnectionPoolSettings`, `ConnectionPoolSettings.Builder` are not accessible to our users and are, therefore, not part the driver API. The method `ConnectionPoolSettings.toString` was also left unmodified to prevent any leaks of the internal settings to a user.

JAVA-4318


stIncMale:
I am going to try and change the approach: instead of introducing an internal setting in `ConnectionPoolSettings`, `com.mongodb.reactivestreams.client.internal.MongoClientImpl` will always create pools that pre-start the async pool worker, and the tests that actually need this will sleep for a predefined amount of time to emulate awaiting for the pre-start to complete (this part has already been implemented).

This approach is fine despite it affects even the async pools in production: instead of starting async workers lazily, async pools will start them eagerly. The approach is deemed fine because situations when a pool is created and never used are considered unusual.

jyemin (approved):
Nice.  LGTM.

rozza (approved):

PR #808, opened 2021-10-20
Write the type of ADTs with only one subclass

rozza:
Original PR: mongodb/mongo-scala-driver#67

JAVA-4300


bson-scala/src/main/scala/org/mongodb/scala/bson/codecs/macrocodecs/MacroCodec.scala
```diff
@@ -80,7 +80,7 @@ trait MacroCodec[T] extends Codec[T] {
    * The field used to save the class name when saving sealed case classes.
    */
   val classFieldName = "_t"
-  lazy val hasClassFieldName: Boolean = caseClassesMap.size > 1
+  lazy val hasClassFieldName: Boolean = caseClassesMapInv.keySet != Set(encoderClass)
```
jyemin, line 83: I'm not sure I understand this new condition.  When would it evaluate to false?  Are we testing that case?

jyemin (approved):

PR #809, opened 2021-10-20
Java 4301

rozza:
Add support for case objects (the equivalent to enums in scala)


jyemin (approved):
I don't fully understand the implementing code, but the tests look good, so I'm going to LGTM this.

PR #811, opened 2021-10-22
Use CodecRegistry in DocumentCodec for Iterable and Map instances

jyemin:
* Instead of using private methods within DocumentCodec to encode instances
of Iterable and Map, instead look up the Codec in the registry as with
any other instance.
* Instead of using a private method within DocumentCodec to decode BSON arrays,
instead lookup up the Codec in the registry according to the type mapped to
array in the BsonTypeClassMap.

JAVA-3925

I added a second commit as well to deduplicate the horrendous UUID-related code that exists in the three affected Codec implementations.


rozza (approved):
LGTM

PR #812, opened 2021-10-25
Test with Java 17

jyemin:
* Mostly testing with Java 8 and 17 now
* Limited testing with Java 11

JAVA-4358

Full patch build: https://spruce.mongodb.com/version/6176b137a4cf47118a95b640/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/run-ocsp-test.sh
+++ b/.evergreen/run-ocsp-test.sh
@@ -31,7 +31,7 @@ provision_ssl () {
 
 echo "Running OCSP tests"
 
-export JAVA_HOME="/opt/java/${JDK}"
+export JAVA_HOME="/opt/java/jdk11"
 
 # show test output
 set -x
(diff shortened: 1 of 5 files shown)


rozza (approved):

.evergreen/run-ocsp-test.sh
```diff
@@ -40,5 +40,5 @@ provision_ssl
 
 echo "Running OCSP tests with ${JDK}"
 ./gradlew -version
-./gradlew -PjdkHome=${JAVA_HOME} ${GRADLE_EXTRA_VARS} --stacktrace --debug --info driver-sync:test --tests OcspTest
+./gradlew -PjdkHome=/opt/java/${JDK} ${GRADLE_EXTRA_VARS} --stacktrace --debug --info driver-sync:test --tests OcspTest
```
stIncMale, line 43: So we run Gradle with JDK 11, but run tests with JDK 17. I guess this is because of [_"Java 17 and later versions are not yet supported."_](https://docs.gradle.org/current/userguide/compatibility.html).

stIncMale (approved):

.evergreen/run-ocsp-test.sh
jyemin: That's right.  We have a separate issue for upgrading to Gradle 7.3.

jyemin: Or, we could do it with toolchains: https://docs.gradle.org/7.3-rc-3/userguide/toolchains.html

PR #817, opened 2021-11-04
Log cursor results at debug level

jyemin:
* The logger name is "org.mongodb.driver.operation"
* The message is "Received batch of %s documents with cursorId %d from server %s"
* It logs the initial results and results from each getMore

JAVA-4377

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
@@ -53,6 +53,7 @@ import static com.mongodb.internal.operation.OperationHelper.LOGGER;
 import static com.mongodb.internal.operation.OperationHelper.getMoreCursorDocumentToQueryResult;
 import static com.mongodb.internal.operation.QueryHelper.translateCommandException;
 import static com.mongodb.internal.operation.ServerVersionHelper.serverIsAtLeastVersionThreeDotTwo;
+import static java.lang.String.format;
 import static java.util.Collections.singletonList;
 
 class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
@@ -120,6 +121,7 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
             }
         }
         this.maxWireVersion = connection == null ? 0 : connection.getDescription().getMaxWireVersion();
+        logQueryResult(firstBatch);
     }
 
     /**
@@ -337,8 +339,10 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
         }
     }
 
+
     private void handleGetMoreQueryResult(final AsyncConnection connection, final SingleResultCallback<List<T>> callback,
                                           final QueryResult<T> result) {
+        logQueryResult(result);
         cursor.set(result.getCursor());
         if (isClosePending) {
             try {
@@ -371,6 +375,11 @@ class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
                 callback.onResult(result.getResults(), null);
             }
         }
+    }
+
+    private void logQueryResult(final QueryResult<T> result) {
+        LOGGER.debug(format("Received batch of %d documents with cursorId %d from server %s", result.getResults().size(),
+                result.getCursorId(), result.getAddress()));
     }
 
     private class CommandResultSingleResultCallback implements SingleResultCallback<BsonDocument> {
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
@@ -22,6 +22,8 @@ import com.mongodb.MongoNamespace;
 import com.mongodb.ReadPreference;
 import com.mongodb.ServerCursor;
 import com.mongodb.connection.ServerType;
+import com.mongodb.diagnostics.logging.Logger;
+import com.mongodb.diagnostics.logging.Loggers;
 import com.mongodb.internal.async.AsyncAggregateResponseBatchCursor;
 import com.mongodb.internal.async.SingleResultCallback;
 import com.mongodb.internal.binding.AsyncConnectionSource;
@@ -49,7 +51,6 @@ import static com.mongodb.assertions.Assertions.isTrueArgument;
 import static com.mongodb.assertions.Assertions.notNull;
 import static com.mongodb.internal.async.ErrorHandlingResultCallback.errorHandlingCallback;
 import static com.mongodb.internal.operation.CursorHelper.getNumberToReturn;
-import static com.mongodb.internal.operation.OperationHelper.LOGGER;
 import static com.mongodb.internal.operation.OperationHelper.getMoreCursorDocumentToQueryResult;
 import static com.mongodb.internal.operation.QueryHelper.translateCommandException;
 import static com.mongodb.internal.operation.ServerVersionHelper.serverIsAtLeastVersionThreeDotTwo;
@@ -57,6 +58,7 @@ import static java.lang.String.format;
 import static java.util.Collections.singletonList;
 
 class AsyncQueryBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
+    private static final Logger LOGGER = Loggers.getLogger("operation");
     private static final FieldNameValidator NO_OP_FIELD_NAME_VALIDATOR = new NoOpFieldNameValidator();
     private static final String CURSOR = "cursor";
     private static final String POST_BATCH_RESUME_TOKEN = "postBatchResumeToken";
--- a/driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
@@ -27,6 +27,8 @@ import com.mongodb.ServerApi;
 import com.mongodb.ServerCursor;
 import com.mongodb.annotations.ThreadSafe;
 import com.mongodb.connection.ServerType;
+import com.mongodb.diagnostics.logging.Logger;
+import com.mongodb.diagnostics.logging.Loggers;
 import com.mongodb.internal.binding.ConnectionSource;
 import com.mongodb.internal.connection.Connection;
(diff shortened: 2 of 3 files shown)


driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -60,9 +62,11 @@
 import static com.mongodb.internal.operation.OperationHelper.getMoreCursorDocumentToQueryResult;
 import static com.mongodb.internal.operation.QueryHelper.translateCommandException;
 import static com.mongodb.internal.operation.ServerVersionHelper.serverIsAtLeastVersionThreeDotTwo;
+import static java.lang.String.format;
 import static java.util.Collections.singletonList;
 
 class QueryBatchCursor<T> implements AggregateResponseBatchCursor<T> {
+    private static final Logger LOGGER = Loggers.getLogger("operation");
```
stIncMale, line 69: `AsyncQueryBatchCursor` uses `com.mongodb.internal.operation.OperationHelper.LOGGER`. For consistency, we should either use the same object here, or create its own instance of a logger in `AsyncQueryBatchCursor`.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
```diff
@@ -373,6 +377,11 @@ private void handleGetMoreQueryResult(final AsyncConnection connection, final Si
         }
     }
 
+    private void logQueryResult(final QueryResult<T> result) {
+        LOGGER.debug(format("Received batch of %d documents with cursorId %d from server %s", result.getResults().size(),
+                result.getCursorId(), result.getAddress()));
```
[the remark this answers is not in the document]
jyemin: I don't think this is a good use of a toString method.  Note that the Javadoc also says: "The string output is not necessarily stable over time". And that's precisely the problem.  It's not uncommon for developers to change the implementation of toString when the class changes (I've often deleted toString and then used the IDEA auto-generation support to recreate it).  In contrast, we want to be very careful about the stability of anything that the driver logs.

As an alternative, I could move this to a helper method, maybe in `OperationHelper` since that's where the logger is also defined. What do you think?

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
jyemin: > For consistency

What is being kept consistent?  I don't think it's a problem to repeat loggers like this.  We do it all over the place. (search for `Loggers.getLogger` and you'll see).

> or create its own instance of a logger in AsyncQueryBatchCursor

What do you mean?  With a different name? I'm trying not to introduce a different name since we document all the logger name that we use.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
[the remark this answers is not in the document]
stIncMale: The phrase about stability is there to discourage non-human usage of the return strings. Anyway, regarding our case (the driver), if we want to keep log messages stable<sup>1</sup>, we should keep `toString` methods stable to the same extent. The only reason why we may not be able to keep `toString` consistent to the same degree we want log messages to be consistent is if we use `toString` for something else in the driver. As far as I know, we don't. Do we?

Note also that if we decide to log something additional about the state of `QueryResult` in the future, it will not matter whether we need to change its `toString` or to explicitly change the message format in the code - we will identically change the log message format as seen by a user either way. That is, again, unless the driver uses `toString` for something different than logging, nothing prevents us from using it for logging<sup>2</sup>.

> It's not uncommon for developers to change the implementation of toString when the class changes
> As an alternative, I could move this to a helper method

It's also not uncommon to have unstable logs and to use `toString` for logging. This practice alone cannot be used as a justification for not using `toString` for stable logging. Instead, we should stop using the practice.

----
<sup>1</sup> The more stable the logging, the more message-specific automation users will introduce to analyze them. When the amount of such automation reaches a critical point, we may end up having to support not only the API but also the arbitrary logging format we will have by that time. If this is what we are trying to achieve, then we are fine, but I would expect that we don't want this to happen. Some libraries intentionally violate properties that are not guaranteed just to reduce the chances of users relying on such properties, even though it is not necessary and the current state of the library manifests the property at runtime otherwise.

<sup>2</sup> I think, logging is what the method is supposed to be used. Do you think there are other more appropriate usages?

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
stIncMale: > What is being kept consistent?
The way `AsyncQueryBatchCursor` and `QueryBatchCursor` access a logger object. There is no `AsyncQueryBatchCursor.LOGGER`, instead this class uses `com.mongodb.internal.operation.OperationHelper.LOGGER`. `QueryBatchCursor` is its synchronous counterpart, so it should also use `com.mongodb.internal.operation.OperationHelper.LOGGER` instead of having `QueryBatchCursor. LOGGER`.

> or create its own instance of a logger in AsyncQueryBatchCursor
> With a different name?

No, I meant a different instance with the same name (though I was incorrect, because `Loggers.getLogger` does not specify whether it returns a reference to the same instance or not, though the library we use creates a new instance). Anyway, please disregard this and only pay attention to the part related to consistency of the code:)

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
```diff
@@ -373,6 +377,11 @@ private void handleGetMoreQueryResult(final AsyncConnection connection, final Si
         }
     }
 
+    private void logQueryResult(final QueryResult<T> result) {
+        LOGGER.debug(format("Received batch of %d documents with cursorId %d from server %s", result.getResults().size(),
```
rozza, line 381: Suggestion: Add an if statement to see if debug is enabled before composing the string and passing it to debug.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -318,6 +322,8 @@ private BsonDocument asGetMoreCommandDocument(final ServerCursor serverCursor) {
     private ServerCursor initFromQueryResult(final QueryResult<T> queryResult) {
         nextBatch = queryResult.getResults().isEmpty() ? null : queryResult.getResults();
         count += queryResult.getResults().size();
+        LOGGER.debug(format("Received batch of %d documents with cursorId %d from server %s", queryResult.getResults().size(),
```
rozza, line 325: Note this will always format a string regardless of debug being enabled.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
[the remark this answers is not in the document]
jyemin: The toString implementations that we typically provide are nearly all auto-generated, and useful only for debugging, not logging.  And worse, because toString is often called implicitly by `String.format`, you can't easily find usages of the method to see how it's being used. 

I would be happy to encapsulate this pretty much anywhere else than toString.

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
jyemin: I considered that, but generally we don't do that for cases where the formatting is relatively cheap, as it is here, and where the messages are only logged once per I/O activity, as it is here.  The cost is far exceeded by the I/O.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
jyemin: See above.

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
jyemin: Oh... I just copy and pasted that line of code from sync to async, and IntelliJ auto-imported the LOGGER. I didn't even notice.  

Done

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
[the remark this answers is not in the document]
stIncMale: > The toString implementations that we typically provide are nearly all auto-generated (and typically contain newlines)

I don't think this is the case, as looking through about 100 implementations in IDE with find filter "String toString()" shows otherwise - there are no explicit newline characters in the returned strings. IntelliJ Idea also does not insert such characters when it generates `toString`.

> intended for debugging, not logging

Logging exists for debugging purposes. But you are probably talking about the process of debugging via an interactive debugger in IDE, for example. This process is sensitive neither to the format of the `toString` result nor to its completeness (when the result is too long, one looks for the specific field of interest by means provided by a debugger). Given the two above statements, `toString` used for logging is equally useful for interactive debugging.

> because toString is often called implicitly by String.format, you can't easily find usages of the method to see where it's being called from

One don't necessary need to find usages, because the answer is supposed to be "it is called when we log messages".

Besides verbosity of the logging code, the problem with deconstructing objects for logging instead of using `toString` is that it significantly increases chances (comparing to when `toString` is used) of having different representations of the same object when logging the object from different places.

I presented all the arguments I can think of at the moment, which is one of the reasons I am approving the PR.

stIncMale (approved):

driver-core/src/main/com/mongodb/internal/operation/AsyncQueryBatchCursor.java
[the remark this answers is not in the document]
jyemin: Sorry, you are totally right about the newlines.  I was remembering incorrectly because of the way the toString methods are typically formatted with one field per line.  But still, something like this is not what I would want to log from a library like ours:

```
      "AcknowledgedBulkWriteResult{"
               + "insertedCount=" + insertedCount
               + ", matchedCount=" + matchedCount
               + ", removedCount=" + removedCount
               + ", modifiedCount=" + modifiedCount
               + ", upserts=" + upserts
               + '}'
```

Thank you for thinking about this as carefully as you have.  We should probably continue the more general discussion in a more interactive forum.

rozza (approved):

PR #819, opened 2021-11-08
JAVA 4357

rozza:
I tried to keep all related changes in separate commits.

Summary of changes: 

* Gradle updated to 6.9.1
* Java toolchain added to the build
* Plugins updated to support compiling with later versions of Java.
  * Configurations moved to directories - standardized the (less config needed for the plugins)
  * Some configurations had to be updated to support changes in the plugins - eg Spotbugs exclusions for random added. We may want to address those errors in the future.
* Evergreen Updated:
  *  `javaVersion` instead of jdk
  * Added a `javaConfig.bash` that is sourced by the shell files and sets: 
    * Locations of jdks - also see gradle.properties
    * Sets `JAVA_HOME` for gradle to use
    * Sets `JAVA_VERSION` for the java version the tests use.
 * ~Updated the Bson project to compile using JAVA 17 - so record support can be added.~   Some more work needed to support JAVA 17 compiling for Bson.  Just need to get the util project / doctags happy with release target 8.


rozza:
Had to remove the bson compile with Java 17 as getting the following error when trying to check the whole project.

```
No matching variant of project :util was found. The consumer was configured to find an API of a library compatible with Java 8, preferably not packaged as a jar, and its dependencies declared externally but:
          - Variant 'apiElements' capability mongo-java-driver:util:unspecified declares an API of a library, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares a component compatible with Java 11 and the consumer needed a component compatible with Java 8
          - Variant 'runtimeElements' capability mongo-java-driver:util:unspecified declares a runtime of a library, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares a component compatible with Java 11 and the consumer needed a component compatible with Java 8

```

gradle.properties
```diff
@@ -19,3 +19,5 @@ org.gradle.jvmargs=-Duser.country=US -Duser.language=en
 scalaVersions=2.11.12,2.12.13,2.13.4
 defaultScalaVersions=2.13.4
 runOnceTasks=clean,release
+org.gradle.java.installations.auto-download=false
+org.gradle.java.installations.fromEnv=JDK8,JDK11,JDK17
```
jyemin, line 23: What's the behavior when this property is unset?

.evergreen/compile.sh
```diff
@@ -6,10 +6,9 @@ set -o errexit  # Exit the script with error if any of the commands fail
 ############################################
 #            Main Program                  #
 ############################################
+source "${BASH_SOURCE%/*}/javaConfig.bash"
```
jyemin, line 9: Nice

gradle.properties
rozza: Tries for auto detection: https://docs.gradle.org/current/userguide/toolchains.html#sec:auto_detection

jyemin (approved):
LGTM! We're on our way.

PR #822, opened 2021-11-10
Make maxConnecting configurable

stIncMale:
The changes in this PR that affect public API are conditional on the reviewers of https://github.com/mongodb/specifications/pull/1098 accepting the `maxConnection` option name. This PR also does not include specification tests, which may appear in the aforementioned PR as a result of the review process. The PR is submitter prematurely in case we want to include it in release 4.4 despite the corresponding [DRIVERS-1943](https://jira.mongodb.org/browse/DRIVERS-1943) not being ready for implementation.

JAVA-4390

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/ConnectionPoolSettings.java
+++ b/driver-core/src/main/com/mongodb/connection/ConnectionPoolSettings.java
@@ -19,7 +19,9 @@ package com.mongodb.connection;
 import com.mongodb.ConnectionString;
 import com.mongodb.annotations.Immutable;
 import com.mongodb.annotations.NotThreadSafe;
+import com.mongodb.event.ConnectionCreatedEvent;
 import com.mongodb.event.ConnectionPoolListener;
+import com.mongodb.event.ConnectionReadyEvent;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -46,6 +48,7 @@ public class ConnectionPoolSettings {
     private final long maxConnectionIdleTimeMS;
     private final long maintenanceInitialDelayMS;
     private final long maintenanceFrequencyMS;
+    private final int maxConnecting;
 
     /**
      * Gets a Builder for creating a new ConnectionPoolSettings instance.
@@ -80,6 +83,7 @@ public class ConnectionPoolSettings {
         private long maxConnectionIdleTimeMS;
         private long maintenanceInitialDelayMS;
         private long maintenanceFrequencyMS = MILLISECONDS.convert(1, MINUTES);
+        private int maxConnecting = 2;
 
         Builder() {
         }
@@ -103,6 +107,7 @@ public class ConnectionPoolSettings {
             maxConnectionIdleTimeMS = connectionPoolSettings.maxConnectionIdleTimeMS;
             maintenanceInitialDelayMS = connectionPoolSettings.maintenanceInitialDelayMS;
             maintenanceFrequencyMS = connectionPoolSettings.maintenanceFrequencyMS;
+            maxConnecting = connectionPoolSettings.maxConnecting;
             return this;
         }
 
@@ -211,6 +216,19 @@ public class ConnectionPoolSettings {
         }
 
         /**
+         * The maximum number of connections a pool may be establishing concurrently.
+         *
+         * @param maxConnecting The maximum number of connections a pool may be establishing concurrently. Must be positive.
+         * @return {@code this}.
+         * @see ConnectionPoolSettings#getMaxConnecting()
+         * @since 4.4
+         */
+        public Builder maxConnecting(final int maxConnecting) {
+            this.maxConnecting = maxConnecting;
+            return this;
+        }
+
+        /**
          * Creates a new ConnectionPoolSettings object with the settings initialised on this builder.
          *
          * @return a new ConnectionPoolSettings object
@@ -249,6 +267,11 @@ public class ConnectionPoolSettings {
             Integer maxConnectionLifeTime = connectionString.getMaxConnectionLifeTime();
             if (maxConnectionLifeTime != null) {
                 maxConnectionLifeTime(maxConnectionLifeTime, MILLISECONDS);
+            }
+
+            Integer maxConnecting = connectionString.getMaxConnecting();
+            if (maxConnecting != null) {
+                maxConnecting(maxConnecting);
             }
 
             return this;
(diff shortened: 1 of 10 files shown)


driver-core/src/main/com/mongodb/connection/ConnectionPoolSettings.java
```diff
@@ -19,7 +19,9 @@
 import com.mongodb.ConnectionString;
 import com.mongodb.annotations.Immutable;
 import com.mongodb.annotations.NotThreadSafe;
+import com.mongodb.event.ConnectionCreatedEvent;
```
rozza, line 22: Comment: As these imports are just for javadocs I'm surprised checkstyle doesn't complain

rozza (changes requested):
`MongoClientOptionsSpecification.should_only_have_the_following_fields_in_the_builder` needs updating but once done LGTM.

driver-core/src/main/com/mongodb/connection/ConnectionPoolSettings.java
```diff
@@ -343,6 +366,21 @@ public long getMaintenanceFrequency(final TimeUnit timeUnit) {
         return connectionPoolListeners;
     }
 
+    /**
+     * The maximum number of connections a pool may be establishing concurrently.
+     * Establishment of a connection is a part of its life cycle
+     * starting after a {@link ConnectionCreatedEvent} and ending before a {@link ConnectionReadyEvent}.
+     * <p>
+     * Default is 2.
+     *
```
jyemin, line 375: Nit: No close p tag is required?  IntelliJ always add it for me.

driver-legacy/src/main/com/mongodb/MongoClientOptions.java
```diff
@@ -142,7 +144,8 @@ private MongoClientOptions(final Builder builder) {
                 .maxSize(getConnectionsPerHost())
                 .maxWaitTime(getMaxWaitTime(), MILLISECONDS)
                 .maxConnectionIdleTime(getMaxConnectionIdleTime(), MILLISECONDS)
-                .maxConnectionLifeTime(getMaxConnectionLifeTime(), MILLISECONDS);
+                .maxConnectionLifeTime(getMaxConnectionLifeTime(), MILLISECONDS)
+                .maxConnecting(builder.maxConnecting);
```
jyemin, line 148: Why is this the only one that comes from the builder rather than a getter?  Does it need to be different?

jyemin (approved):
Just a couple of nits. LGTM.

driver-legacy/src/main/com/mongodb/MongoClientOptions.java
stIncMale: While it seems off, the right question is "why does the `MongoClientOptions` duplicates `maxConnectionLifeTime` and other fields from `MongoClientOptions.connectionPoolSettings`?" As a result of duplicating them, the constructor of `MongoClientOptions` at first initializes the duplicated fields from the `builder`, and then initializes the fields in its `connectionPoolSettings` by reading the duplicated fields via methods like `MongoClientOptions.getMaxConnectionLifeTime`.

Instead of doing the same duplication for `maxConnecting`, I use `connectionPoolSettings.maxConnecting`.

driver-core/src/main/com/mongodb/connection/ConnectionPoolSettings.java
stIncMale: In some places our static checkers complained (which is wrong in my opinion), but here they fortunately did not.

driver-legacy/src/main/com/mongodb/MongoClientOptions.java
jyemin: Maybe a philosophical difference.  I find it confusing as a reader when two styles are mixed in the same class like this. I start wondering if there is a functional difference between the two approaches, and it wastes time.  But if you feel strongly about it, I won't insist. Perhaps a comment explaining the difference would be helpful for future readers.

stIncMale:
> MongoClientOptionsSpecification.should_only_have_the_following_fields_in_the_builder needs updating

@rozza The problem was no that I did not update the test, but that the order of fields was incorrect: https://evergreen.mongodb.com/test_log/mongo_java_driver_tests_jdk8_unsecure_standalone__version~4.4_os~linux_topology~standalone_auth~noauth_ssl~nossl_jdk~jdk8_test_patch_eada8a28acc0769a2e0ad57327be0fb8bfc52018_618b5754a4cf477851acc0fe_21_11_10_05_23_33/0?test_name=5f156141c6036f7032f2ed887bf8afb9&group_id=&text=true

My understanding is that originally the field was called `maxPoolConnecting` and was placed in the right position. Then I renamed it to `maxConnecting`, which changed its position in a sorted list, but did not move it in the expected list and created the PR without re-running tests locally. In order to avoid similar situations in the future, I changed the test so that it operates sets, not lists.

driver-core/src/main/com/mongodb/connection/ConnectionPoolSettings.java
stIncMale: I guess, I have never given it proper thought. [Examples posted by Oracle](https://www.oracle.com/ca-en/technical-resources/articles/java/javadoc-tool.html#format) do not close the tag, so I blindly followed. This is similar to the situation we discussed yesterday about examples violating best practices because "it's simpler this way".

- The standard doclet specification states [_"HTML constructs should be written in HTML 5"_](https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html#html-content)
- HTML 5 states [_"A p element's end tag can be omitted if the p element is immediately followed by an address, article, aside, blockquote, details, div, dl, fieldset, figcaption, figure, footer, form, h1, h2, h3, h4, h5, h6, header, hgroup, hr, main, menu, nav, ol, p, pre, section, table, or ul element, or if there is no more content in the parent element and the parent element is an HTML element that is not an a, audio, del, ins, map, noscript, or video element, or an autonomous custom element."_](https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element).

I don't think a human can reasonably apply the above omission rule, and since omitting the end tag for the `p` element is optional, we should just not do it. Thank you for pointing this out, I will stop omitting the end tag, I also added closing tags in the PR.

driver-legacy/src/main/com/mongodb/MongoClientOptions.java
stIncMale: Duplicated `maxConnecting` in `MongoClientOptions` similarly to other fields and filed [Refactor `MongoClientOptions` so that it wraps `MongoClientSettings`](https://jira.mongodb.org/browse/JAVA-4404) as you suggested in Slack.

PR #823, opened 2021-11-11
Fix id reporting bug with _ids containing arrays

rozza:
JAVA-4403


driver-core/src/main/com/mongodb/internal/connection/IdHoldingBsonWriter.java
```diff
@@ -55,6 +55,7 @@
     private BasicOutputBuffer outputBuffer;
     private String currentFieldName;
     private BsonValue id;
+    private boolean idFieldIsNested = false;
```
jyemin, line 58: Not sure what "nested" means in this context.  Can you clarify?  When I first read it, I was thinking it meant that there was an _id field nested inside a subdocument, but that clearly is not right. Maybe we can come up with a better name.

driver-core/src/test/unit/com/mongodb/internal/connection/IdHoldingBsonWriterSpecification.groovy
```diff
@@ -88,6 +88,32 @@ class IdHoldingBsonWriterSpecification extends Specification {
...
+        when:
+        new BsonDocumentCodec().encode(idTrackingBsonWriter, document, EncoderContext.builder()
+                .isEncodingCollectibleDocument(true).build())
+        def encodedDocument = getEncodedDocument(bsonBinaryWriter.getBsonOutput())
+
+        then:
+        encodedDocument == document
+
+        where:
+        json << ['{"_id": {"a": []}, "b": 123}',
```
jyemin, line 105: Did all of these fail before the fix?

rozza: No I added extra for regression purposes.

These tests failed (where the array is a sub field of a document of the _id):

```
{"_id": {"a": [1, 2]}, "b": 123},
{"_id": {"a": [[[[1]]]]}, "b": 123},
{"_id": {"a": []}, "b": 123},
{"_id": {"a": [{"a": [1, 2]}]}, "b": 123},

driver-core/src/main/com/mongodb/internal/connection/IdHoldingBsonWriter.java
rozza: Its tricky - it really means the id field is a top level array and has to be nested in a document to decode.  Happy for any suggestions.

jyemin (approved):

PR #824, opened 2021-11-12
Java 4357

rozza:


.evergreen/javaConfig.bash
```diff
@@ -0,0 +1,16 @@
+# Java configurations for evergreen
+# Defaults to Java 11
+
+export JDK8="/opt/java/jdk8"
+export JDK11="/opt/java/jdk11"
+export JDK17="/opt/java/jdk17"
+
+if [ -d "$JDK11" ]; then
```
jyemin, line 8: What's the intention here?  It seems that since JDK11 is defined to a constance on line 5, any conditional based on its value is going to always evaluate to the same thing.

.evergreen/javaConfig.bash
```diff
@@ -0,0 +1,16 @@
...
+  export JAVA_HOME=$JDK11
+fi
+
+export JAVA_VERSION=${JDK:-11}
```
jyemin, line 12: Which scripts define the JDK variable?

.evergreen/publish.sh
```diff
@@ -7,23 +7,23 @@ set -o errexit  # Exit the script with error if any of the commands fail
...
-export JAVA_HOME="/opt/java/jdk11"
-
 if [ "$RELEASE" == "true" ]; then
   TASK="publishArchives"
 else
   TASK="publishSnapshots"
 fi
 
-SYSTEM_PROPERTIES="-PjdkHome=/opt/java/jdk17 -Dorg.gradle.internal.publish.checksums.insecure=true -Dorg.gradle.internal.http.connectionTimeout=120000 -Dorg.gradle.internal.http.socketTimeout=120000"
+SYSTEM_PROPERTIES="-Dorg.gradle.internal.publish.checksums.insecure=true -Dorg.gradle.internal.http.connectionTimeout=120000 -Dorg.gradle.internal.http.socketTimeout=120000"
```
jyemin, line 26: Is `-PjavaVersion=${JAVA_VERSION}` needed here?

.evergreen/run-kms-tls-tests.sh
```diff
@@ -10,11 +10,10 @@ set -o errexit  # Exit the script with error if any of the commands fail
 ############################################
 #            Main Program                  #
 ############################################
-
+RELATIVE_DIR_PATH="$(dirname "${BASH_SOURCE:-$0}")"
```
jyemin, line 13: Is `-PjavaVersion=${JAVA_VERSION}` needed in the gradle invocations below?

.evergreen/run-mongodb-aws-ecs-test.sh
```diff
@@ -34,6 +34,10 @@ if ! which git ; then
 fi
 
 cd src
+
+RELATIVE_DIR_PATH="$(dirname "${BASH_SOURCE:-$0}")"
```
jyemin, line 38: Is `-PjavaVersion=${JAVA_VERSION}` needed below, or are we just relying on the default?

build.gradle
```diff
@@ -354,21 +333,7 @@ def getGitVersion() {
...
 apply from: 'gradle/javadoc.gradle'
-
-//////////////////////////////////////////
-// Root project configuration           //
-//////////////////////////////////////////
-wrapper {
-    gradleVersion = '6.0.1'
-}
-
-if (!JavaVersion.current().isJava9Compatible()) {
```
jyemin, line 368: We don't need this anymore?

driver-scala/build.gradle
```diff
@@ -21,7 +21,7 @@ dependencies {
     compile project(path: ':bson-scala', configuration: 'default')
     compile project(path: ':driver-reactive-streams', configuration: 'default')
 
-    testImplementation project(':util')
+    testImplementation project(':util:spock')
```
jyemin, line 24: Scala uses Spock?

gradle/javaToolchain.gradle
```diff
@@ -0,0 +1,45 @@
...
+def javaMainProjects = subprojects - project(":util:taglets")
+
+allprojects {
+    afterEvaluate { project ->
+        def javaVersion = (findProperty("javaVersion") ?: DEFAULT_JDK_VERSION.toString()).toInteger()
+        project.pluginManager.withPlugin("java") {
+            logger.info("Compiling ${project.name} using JDK${DEFAULT_JDK_VERSION}")
+            java {
+                toolchain {
+                    languageVersion = JavaLanguageVersion.of(DEFAULT_JDK_VERSION)
```
jyemin, line 27: So this is saying that we always compile with Java 11?  Does that include compiling the test code?

gradle/javaToolchain.gradle
```diff
@@ -0,0 +1,45 @@
...
+        if (project in javaMainProjects) {
+            tasks.withType(JavaCompile) {
+                options.encoding = "UTF-8"
+                options.release.set(8)
+            }
+        }
+
+        tasks.withType(Test) {
+            javaLauncher.set(javaToolchains.launcherFor {
+                languageVersion = JavaLanguageVersion.of(javaVersion)
```
jyemin, line 41: And we always execute the tests with javaVersion, after compilation is complete?

gradle/javadoc.gradle
```diff
@@ -17,11 +17,11 @@ import static org.gradle.util.CollectionUtils.single
  */
 
 
-def javaMainProjects = subprojects.findAll { !['util'].contains(it.name) }
+def nonDocsProjects = project(":util").allprojects + project(":driver-benchmarks") + project("driver-workload-executor")
```
jyemin, line 20: This variable name confused me.  I was thinking it would include driver-core, bson, etc, based on the name.  It's more like nonDocsOrMainProjects?    Maybe we should just inline it...

gradle/publish.gradle
```diff
@@ -69,7 +69,8 @@ ext {
     }
 }
 
-def publishedProjects = subprojects.findAll {!['util', 'driver-benchmarks', 'driver-workload-executor'].contains(it.name) }
+def nonPublishProjects = project(":util").allprojects + project(":driver-benchmarks") + project("driver-workload-executor")
```
jyemin, line 72: nonPublishProjects => nonPublishedProjects

jyemin (changes requested):

.evergreen/javaConfig.bash
rozza: Checks that the directory actually exists. If not don't set JAVA_HOME.

.evergreen/javaConfig.bash
rozza: Good catch, thats a bug. Should be JAVA_VERSION as set in .evg.yml

.evergreen/publish.sh
rozza: No because JavaVersion is only used for testing purposes.

.evergreen/run-kms-tls-tests.sh
rozza: No this has the same behaviour as before - JAVA_HOME is set to the default java 11.

.evergreen/run-mongodb-aws-ecs-test.sh
rozza: Previously, this used the default JAVA_HOME - and potentially installs JAVA 11.  The directory check in the javaConfig.bash ensures that the default (/opt/java/jdk11) exists - if it doesn't set JAVA_HOME

build.gradle
rozza: Because java toolchain functionality will error if the java version being used isn't compatible.

driver-scala/build.gradle
rozza: No - I'll double check that nothing is marked as `@Slow`

gradle/javaToolchain.gradle
rozza: Correct

gradle/javaToolchain.gradle
rozza: Correct

gradle/javadoc.gradle
rozza: These are all the projects for which we don't publish javaDoc or ScalaDocs for.

Changed to: `projectsThatDoNotPublishJavaDocs`

gradle/publish.gradle
rozza: Changed to `projectsNotPublishedToMaven`

rozza (commented):
Replied to comments, fixed a bug (good catch) and clarified some variable names.

rozza:
Put back into draft.

Seems like the fixes aren't working as expected as groovy code is compiled without the java 8 release target. So can't run using java 8. Investigating why and hopefully a fix.

jyemin (approved):
I'm not sure I understand the latest changes, but assuming the outcome is correct then LGTM

rozza:
Rebased on the command line and merged.

PR #827, opened 2021-11-18
Support "com.sun.security.jgss.initiate" as Kerberos login context name

jyemin:
JAVA-4391

Straightforward change except for testing it.

### DIFF
--- a/.evergreen/run-gssapi-auth-test.sh
+++ b/.evergreen/run-gssapi-auth-test.sh
@@ -11,21 +11,21 @@ set -o errexit  # Exit the script with error if any of the commands fail
 #       REALM                   The realm
 #       KEYTAB_BASE64           The BASE64-encoded keytab
 #       PROJECT_DIRECTORY       The project directory
+#       LOGIN_CONTEXT_NAME      The login context name to use to look up the GSSAPI Subject
 
 ############################################
 #            Main Program                  #
 ############################################
 RELATIVE_DIR_PATH="$(dirname "${BASH_SOURCE:-$0}")"
 . "${RELATIVE_DIR_PATH}/javaConfig.bash"
-
-echo "Running GSSAPI authentication tests"
+echo "Running GSSAPI authentication tests with login context name '${LOGIN_CONTEXT_NAME}'"
 
 echo ${KEYTAB_BASE64} | base64 -d > ${PROJECT_DIRECTORY}/.evergreen/drivers.keytab
 
 trap "rm ${PROJECT_DIRECTORY}/.evergreen/drivers.keytab; exit" EXIT HUP
 
 cat << EOF > .evergreen/java.login.drivers.config
-com.sun.security.jgss.krb5.initiate {
+${LOGIN_CONTEXT_NAME} {
     com.sun.security.auth.module.Krb5LoginModule required
           doNotPrompt=true useKeyTab=true keyTab="${PROJECT_DIRECTORY}/.evergreen/drivers.keytab" principal=drivers;
 };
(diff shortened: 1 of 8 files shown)


.evergreen/run-gssapi-auth-test.sh
```diff
@@ -11,21 +11,21 @@ set -o errexit  # Exit the script with error if any of the commands fail
...
-echo "Running GSSAPI authentication tests"
+echo "Running GSSAPI authentication tests with login context name '${LOGIN_CONTEXT_NAME}'"
 
 echo ${KEYTAB_BASE64} | base64 -d > ${PROJECT_DIRECTORY}/.evergreen/drivers.keytab
 
 trap "rm ${PROJECT_DIRECTORY}/.evergreen/drivers.keytab; exit" EXIT HUP
 
 cat << EOF > .evergreen/java.login.drivers.config
-com.sun.security.jgss.krb5.initiate {
+${LOGIN_CONTEXT_NAME} {
```
jyemin, line 28: Now we pass this in so we can test different values

.evergreen/run-gssapi-auth-test.sh
```diff
@@ -34,7 +34,7 @@ EOF
 echo "Running tests with Java ${JAVA_VERSION}"
 ./gradlew -version
 ./gradlew -PjavaVersion=${JAVA_VERSION} --stacktrace --info \
--Dorg.mongodb.test.uri=${MONGODB_URI} \
--Pgssapi.enabled=true -Psun.security.krb5.debug=true -Pauth.login.config=file://${PROJECT_DIRECTORY}/.evergreen/java.login.drivers.config \
+-Dorg.mongodb.test.uri=${MONGODB_URI} -Pgssapi.enabled=true -Pgssapi.login.context.name=${LOGIN_CONTEXT_NAME} \
```
jyemin, line 37: Also pass it in as a property so that tests know which one to use.

build.gradle
```diff
@@ -201,6 +201,7 @@ configure(javaAndScalaTestedProjects) {
 
         if (project.buildingWith('gssapi.enabled')) {
             systemProperties(
+                    'org.mongodb.test.gssapi.login.context.name': project.getProperty('gssapi.login.context.name'),
```
jyemin, line 204: I was just following the pattern for `gssapi.enbabled` by making this a project property.

driver-core/src/main/com/mongodb/KerberosSubjectProvider.java
```diff
@@ -52,23 +54,30 @@
...
+        this("com.sun.security.jgss.krb5.initiate", "com.sun.security.jgss.initiate");
     }
 
     /**
      * Construct an instance with the specified login context name
      *
      * @param loginContextName the login context name
      */
-    public KerberosSubjectProvider(@NonNull final String loginContextName) {
+    public KerberosSubjectProvider(final String loginContextName) {
```
jyemin, line 74: This is already a NonNullApi so these annotations are not needed.

driver-core/src/main/com/mongodb/KerberosSubjectProvider.java
```diff
@@ -80,16 +89,32 @@ public KerberosSubjectProvider(@NonNull final String loginContextName) {
...
+            subject = createNewSubject();
         }
         return subject;
     }
 
-    private static boolean needNewSubject(@NonNull final Subject subject) {
+    private Subject createNewSubject() throws LoginException {
+        LoginContext loginContext;
+        try {
+            LOGGER.debug(format("Creating LoginContext with name '%s'", loginContextName));
```
jyemin, line 100: Should never have been info logging

driver-core/src/main/com/mongodb/KerberosSubjectProvider.java
```diff
@@ -80,16 +89,32 @@ public KerberosSubjectProvider(@NonNull final String loginContextName) {
...
+            loginContext = new LoginContext(loginContextName);
```
jyemin, line 101: Take advantage of the fact that it's the constructor that throws if the name is not found, not the login method

driver-core/src/main/com/mongodb/KerberosSubjectProvider.java
```diff
@@ -80,16 +89,32 @@ public KerberosSubjectProvider(@NonNull final String loginContextName) {
...
+        } catch (LoginException e) {
+            if (fallbackLoginContextName == null) {
+                throw e;
+            }
+            LOGGER.debug(format("Creating LoginContext with fallback name '%s'", fallbackLoginContextName));
+            loginContext = new LoginContext(fallbackLoginContextName);
+            loginContextName = fallbackLoginContextName;
```
jyemin, line 108: Once we fall back once, always fall back

rozza (approved):
LGTM

driver-core/src/main/com/mongodb/KerberosSubjectProvider.java
stIncMale: If we want this behavior, then we should execute
```java
loginContextName = fallbackLoginContextName;
fallbackLoginContextName = null;
```
before attempting to construct a `LoginContext` within the `catch` block: if this attempt also fails, then the next execution of the `getSubject`method (they are totally ordered as a result of `synchronized`) uses the initial `loginContextName` instead of using the fallback name.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/KerberosSubjectProvider.java
jyemin: Sorry, I was imprecise.  I meant to say that once we fallback to calling the `_login_` method, then we always fall back.  I don't actually think it matters much what we do if the `LoginContext` constructor throws for both names.

stIncMale (approved):

PR #828, opened 2021-11-19
Java 4388

rozza:
Gradle now requires Java 17 to build.

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -1527,11 +1527,11 @@ axes:
       - id: "2.12"
         display_name: "Scala 2.12"
         variables:
-          SCALA: "2.12.10"
+          SCALA: "2.12.15"
       - id: "2.13"
         display_name: "Scala 2.13"
         variables:
-          SCALA: "2.13.0"
+          SCALA: "2.13.6"
 
   # Choice of MongoDB storage engine
   - id: storage-engine
--- a/.evergreen/run-mongodb-aws-ecs-test.sh
+++ b/.evergreen/run-mongodb-aws-ecs-test.sh
@@ -25,7 +25,10 @@ apt update
 
 if ! which java ; then
     echo "Installing java..."
-    apt install openjdk-11-jdk -y
+    # Ubuntu 18.04 ca-certificates-java and opendjdk-17 bug work around
+    dpkg --purge --force-depends ca-certificates-java
+    apt install ca-certificates-java -y
+    apt install openjdk-17-jdk -y
 fi
 
 if ! which git ; then
--- a/.evergreen/run-scala-tests.sh
+++ b/.evergreen/run-scala-tests.sh
@@ -34,4 +34,4 @@ fi
 echo "Running scala tests with Scala $SCALA"
 
 ./gradlew -version
-./gradlew -PjavaVersion=${JAVA_VERSION} -PscalaVersion=$SCALA --stacktrace --info :bson-scala:test :driver-scala:test :driver-scala:integrationTest -Dorg.mongodb.test.uri=${MONGODB_URI} ${MULTI_MONGOS_URI_SYSTEM_PROPERTY}
+./gradlew -PscalaVersion=$SCALA --stacktrace --info :bson-scala:test :driver-scala:test :driver-scala:integrationTest -Dorg.mongodb.test.uri=${MONGODB_URI} ${MULTI_MONGOS_URI_SYSTEM_PROPERTY}
--- a/bson-scala/build.gradle
+++ b/bson-scala/build.gradle
@@ -24,11 +24,10 @@ dependencies {
 sourceSets {
     main {
         scala {
-            srcDirs += ["src/main/scala"]
             if(scalaVersion.startsWith("2.13")) {
-                srcDirs += ["src/main/scala-2.13+"]
+                srcDirs = ["src/main/scala", "src/main/scala-2.13+"]
             } else {
-                srcDirs += ["src/main/scala-2.13-"]
+                srcDirs = ["src/main/scala", "src/main/scala-2.13-"]
             }
         }
     }
--- a/bson-scala/build.gradle
+++ b/bson-scala/build.gradle
@@ -18,7 +18,7 @@ description = "A Scala wrapper / extension to the bson library"
 archivesBaseName = 'mongo-scala-bson'
 
 dependencies {
-    compile project(path: ':bson', configuration: 'default')
+    implementation project(path: ':bson', configuration: 'default')
 }
 
 sourceSets {
--- a/build.gradle
+++ b/build.gradle
@@ -133,12 +133,6 @@ configure(scalaProjects) {
         scalaCompileOptions.deprecation = false
     }
 
-    if (project.scalaVersion.startsWith("2.13")) {
-        scaladoc {
-            scalaClasspath += files("$rootDir/gradle/scala/lib/scala-ant-2.13.1.jar")
-        }
-    }
-
     tasks.withType(GenerateModuleMetadata) {
         enabled = false
     }
--- a/build.gradle
+++ b/build.gradle
@@ -72,10 +72,6 @@ configure(coreProjects) {
         mavenLocal()
         google()
         mavenCentral()
-        maven {
-            url 'https://oss.sonatype.org/content/repositories/snapshots'
-        }
-
     }
 }
 
@@ -330,7 +326,7 @@ def getGitVersion() {
     describeStdOut.toString().substring(1).trim()
 }
 
+apply from: 'gradle/javaToolchain.gradle'
 apply from: 'gradle/publish.gradle'
 apply from: 'gradle/deploy.gradle'
 apply from: 'gradle/javadoc.gradle'
-apply from: 'gradle/javaToolchain.gradle'
--- a/build.gradle
+++ b/build.gradle
@@ -24,12 +24,12 @@ buildscript {
         maven { url "https://plugins.gradle.org/m2/" }
     }
     dependencies {
-        classpath 'com.netflix.nebula:gradle-extra-configurations-plugin:3.0.3'
+        classpath 'com.netflix.nebula:gradle-extra-configurations-plugin:7.0.0'
         classpath "com.github.spotbugs.snom:spotbugs-gradle-plugin:4.7.9"
         classpath 'biz.aQute.bnd:biz.aQute.bnd.gradle:5.1.2'
 
         // Scala plugins
-        classpath "com.adtran:scala-multiversion-plugin:1.0.36"
+        classpath "com.adtran:scala-multiversion-plugin:2.0.4"
         classpath "com.diffplug.spotless:spotless-plugin-gradle:3.27.1"
 
         // Test logging plugin
@@ -73,7 +73,7 @@ configure(coreProjects) {
         google()
         mavenCentral()
         maven {
-            url 'http://oss.sonatype.org/content/repositories/snapshots'
+            url 'https://oss.sonatype.org/content/repositories/snapshots'
         }
 
     }
@@ -103,8 +103,8 @@ configure(scalaProjects) {
     group = 'org.mongodb.scala'
 
     dependencies {
-        compile ('org.scala-lang:scala-library:%scala-version%')
-        compile ('org.scala-lang:scala-reflect:%scala-version%')
+        implementation ('org.scala-lang:scala-library:%scala-version%')
+        implementation ('org.scala-lang:scala-reflect:%scala-version%')
 
         testImplementation(platform("org.junit:junit-bom:$junitBomVersion"))
         testImplementation("org.junit.vintage:junit-vintage-engine")
@@ -152,6 +152,9 @@ configure(javaMainProjects) {
         compileOnly 'com.google.code.findbugs:jsr305:1.3.9'
         api 'org.slf4j:slf4j-api:1.7.6', optional
         testImplementation 'com.google.code.findbugs:jsr305:1.3.9'
+
+        // https://issues.apache.org/jira/browse/GROOVY-10194
+        testImplementation 'org.codehaus.groovy:groovy-all:3.0.9'
     }
 
     /* Compiling */
@@ -294,8 +297,8 @@ configure(javaCodeCheckedProjects) {
     /* Code quality */
     tasks.withType(Checkstyle) {
         reports {
-            xml.enabled true
-            html.enabled true
+            xml.required = true
+            html.required = true
         }
     }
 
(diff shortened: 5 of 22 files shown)


.evergreen/.evg.yml
```diff
@@ -1527,11 +1527,11 @@ axes:
       - id: "2.12"
         display_name: "Scala 2.12"
         variables:
-          SCALA: "2.12.10"
+          SCALA: "2.12.15"
```
rozza, line 1530: Use latest minor versions of scala.

.evergreen/run-mongodb-aws-ecs-test.sh
```diff
@@ -25,7 +25,10 @@ apt update
 
 if ! which java ; then
     echo "Installing java..."
-    apt install openjdk-11-jdk -y
+    # Ubuntu 18.04 ca-certificates-java and opendjdk-17 bug work around
+    dpkg --purge --force-depends ca-certificates-java
+    apt install ca-certificates-java -y
+    apt install openjdk-17-jdk -y
```
rozza, line 31: Just calling `apt install openjdk-17-jdk -y` failed to install due to a dependency / packaging bug. The purge and install fixes the issue.

.evergreen/run-scala-tests.sh
```diff
@@ -34,4 +34,4 @@ fi
 echo "Running scala tests with Scala $SCALA"
 
 ./gradlew -version
-./gradlew -PjavaVersion=${JAVA_VERSION} -PscalaVersion=$SCALA --stacktrace --info :bson-scala:test :driver-scala:test :driver-scala:integrationTest -Dorg.mongodb.test.uri=${MONGODB_URI} ${MULTI_MONGOS_URI_SYSTEM_PROPERTY}
+./gradlew -PscalaVersion=$SCALA --stacktrace --info :bson-scala:test :driver-scala:test :driver-scala:integrationTest -Dorg.mongodb.test.uri=${MONGODB_URI} ${MULTI_MONGOS_URI_SYSTEM_PROPERTY}
```
rozza, line 37: Don't set the javaVersion for scala

build.gradle
```diff
@@ -24,12 +24,12 @@ buildscript {
         maven { url "https://plugins.gradle.org/m2/" }
     }
     dependencies {
-        classpath 'com.netflix.nebula:gradle-extra-configurations-plugin:3.0.3'
+        classpath 'com.netflix.nebula:gradle-extra-configurations-plugin:7.0.0'
```
rozza, line 27: Update plugins to work with Java 17 and Gradle 7.3

build.gradle
```diff
@@ -152,6 +146,9 @@ configure(javaMainProjects) {
         compileOnly 'com.google.code.findbugs:jsr305:1.3.9'
         api 'org.slf4j:slf4j-api:1.7.6', optional
         testImplementation 'com.google.code.findbugs:jsr305:1.3.9'
+
+        // https://issues.apache.org/jira/browse/GROOVY-10194
+        testImplementation 'org.codehaus.groovy:groovy-all:3.0.9'
```
rozza, line 151: Force the latest version of groovy - allows the Spock tests to compile and then run! (was a pain to find and fix this).

build.gradle
```diff
@@ -294,8 +291,8 @@ configure(javaCodeCheckedProjects) {
     /* Code quality */
     tasks.withType(Checkstyle) {
         reports {
-            xml.enabled true
-            html.enabled true
+            xml.required = true
```
rozza, line 294: Changed the variable name in checkstyle.

driver-core/build.gradle
```diff
@@ -16,7 +16,7 @@
 
 
 plugins {
-    id 'de.fuerstenau.buildconfig' version '1.1.8'
+    id 'com.github.gmazzo.buildconfig' version '3.0.3'
```
rozza, line 19: `de.fuerstenau.buildconfig` doesn't support Java 17. Found a replacement that does.

driver-scala/src/it/scala/org/mongodb/scala/gridfs/GridFSObservableSpec.scala
```diff
@@ -39,7 +39,7 @@ class GridFSObservableSpec extends RequiresMongoDBISpec with FuturesSpec with Be
   private var _filesCollection: Option[MongoCollection[GridFSFile]] = None
   private var _chunksCollection: Option[MongoCollection[Document]] = None
   private val singleChunkString = "GridFS"
-  private val multiChunkString = singleChunkString.formatted(s"%${1024 * 255 * 5}s")
+  private val multiChunkString = f"${singleChunkString}%1305600s"
```
rozza, line 42: This is bonkers - a change in scala behaviour across minor releases!

The purpose of the code is to pad the string. Depending on the **point version** of scala `singleChunkString.formatted(s"%${1024 * 255 * 5}s")` may or may not pad the string. This caused a local test to fail but I never understood why, when it passed on evergreen.

I finally found the cause when I changed all the point versions in the build and in gradle.properties and then evergreen failed also. Updated the code to use the format string literal `f"` and explicitly set the amount of padding.

driver-scala/src/test/scala/org/mongodb/scala/MongoDriverInformationSpec.scala
```diff
@@ -28,7 +28,7 @@ class MongoDriverInformationSpec extends BaseSpec {
       MongoDriverInformationClass.getDeclaredFields.filter(f => isStatic(f.getModifiers)).map(_.getName).toSet
     val wrappedMethods =
       MongoDriverInformationClass.getDeclaredMethods.filter(f => isStatic(f.getModifiers)).map(_.getName).toSet
-    val exclusions = Set("$VALUES", "valueOf", "values")
+    val exclusions = Set("$VALUES", "$values", "valueOf", "values")
```
rozza, line 31: Another change - depending on scala version there maybe a $VALUES or $values synthetic method.

gradle/javaToolchain.gradle
```diff
@@ -13,18 +13,20 @@
...
 
 allprojects {
 
     def javaVersion = (findProperty("javaVersion") ?: DEFAULT_JDK_VERSION.toString()).toInteger()
     project.pluginManager.withPlugin("java") {
-        logger.info("Compiling ${project.name} using JDK${DEFAULT_JDK_VERSION}")
-        java {
-            toolchain {
-                languageVersion = JavaLanguageVersion.of(DEFAULT_JDK_VERSION)
+        if (!project.pluginManager.hasPlugin("scala")) {
```
rozza, line 24: Don't force java version for Scala projects.  Scala 2.11 doesn't support Java 17. It EOL'd in 2017 - its worth determining our support plan for older scala versions (2.11 & 2.12)

build.gradle
jyemin: Wow, we were very far behind.  Do you recall what this plugin is used for?

bson-scala/build.gradle
```diff
@@ -18,17 +18,16 @@ description = "A Scala wrapper / extension to the bson library"
...
 
 dependencies {
-    compile project(path: ':bson', configuration: 'default')
+    implementation project(path: ':bson', configuration: 'default')
 }
 
 sourceSets {
     main {
         scala {
-            srcDirs += ["src/main/scala"]
```
jyemin, line 27: Did something stop working here?   It looks the same to me.

build.gradle
```diff
@@ -73,7 +73,7 @@ configure(coreProjects) {
         google()
         mavenCentral()
         maven {
-            url 'http://oss.sonatype.org/content/repositories/snapshots'
+            url 'https://oss.sonatype.org/content/repositories/snapshots'
```
jyemin, line 76: Do we depend on any snapshots, or could this just be removed?

build.gradle
jyemin: Oh, just needed the latest patch release.  So once Spock re-releases we can remove this?

jyemin: Should we have a ticket to update Spock dependency, and mention this?

build.gradle
```diff
@@ -333,7 +330,7 @@ def getGitVersion() {
     describeStdOut.toString().substring(1).trim()
 }
 
-apply from: 'gradle/javaToolchain.gradle'
 apply from: 'gradle/publish.gradle'
 apply from: 'gradle/deploy.gradle'
 apply from: 'gradle/javadoc.gradle'
+apply from: 'gradle/javaToolchain.gradle'
```
jyemin, line 336: There's a significance to the order?  Another thing I didn't know...

build.gradle
jyemin: Did we pick up a newer version of Checkstyle somewhere?  I didn't see it in the diffs.

driver-scala/build.gradle
```diff
@@ -29,22 +29,21 @@ dependencies {
...
         scala {
-            compileClasspath += main.output + test.output
-            runtimeClasspath += main.output + test.output
+            compileClasspath += main.output
+            runtimeClasspath += main.output
         }
     }
 }
 
+configurations {
```
jyemin, line 42: Curious why order matters here.  It has to go below sourceSets?

driver-scala/build.gradle
```diff
@@ -29,22 +29,21 @@ dependencies {
...
-    integrationTestRuntime.extendsFrom testRuntime
-}
-
 sourceSets {
     integrationTest {
         scala.srcDir file('src/it/scala')
         scala {
-            compileClasspath += main.output + test.output
-            runtimeClasspath += main.output + test.output
+            compileClasspath += main.output
```
jyemin, line 36: Curious why this changed.

gradle/javaToolchain.gradle
jyemin: How does the underlying JVM get chosen when we specify the Scala version?

jyemin (commented):
Your ability to figure out all things Gradle astounds me.  I am grateful and a bit envious.

I just have a lot of questions about all the things I don't understand.

build.gradle
rozza: optional dependencies

build.gradle
rozza: Can do - https://repo1.maven.org/maven2/org/spockframework/spock-core/2.0-groovy-3.0/spock-core-2.0-groovy-3.0.pom requires 3.0.8 :/

Spock 2.1 will require Groovy 3.0.9 but isn't released yet.
Spock 2.2 will possibly add a Groovy  4.0 release

build.gradle
rozza: Yes, the checkstyle plugin is packaged with gradle - so it was implicitly updated.

jyemin:
> Gradle now requires Java 17 to build.

That doesn't seem to be true, as I can build with Java 11.  Is it the intention that Java 11 will be rejected?  Once we start testing our support of records, it will be true.  Do we need to short-circuit the build if we detect a Java < 17?

gradle/javaToolchain.gradle
rozza: I think its via the java toolchain support mechanism.

build.gradle
rozza: I don't believe we do - it could be removed. It just had to be changed to support https as http is no longer supported.

build.gradle
rozza: I just changed it as part of debugging and getting everything done - it might not actually matter.

driver-core/build.gradle
```diff
@@ -53,11 +53,11 @@ dependencies {
...
-    appName = 'mongo-java-driver'
-    version = project.gitVersion
-
-    clsName = 'MongoDriverVersion'
-    packageName = 'com.mongodb.internal.build'
+    className('MongoDriverVersion')
+    packageName('com.mongodb.internal.build')
+    useJavaOutput()
+    buildConfigField('String', 'NAME', '"mongo-java-driver"')
+    buildConfigField('String', 'VERSION', "\"${gitVersion}\"")
```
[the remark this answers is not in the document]
rozza: Something to do with the way the config works it has to be wrapped in a quotes.

`${gitversion}` just uses the template syntax to process the string.

driver-scala/build.gradle
rozza: test source sets were not needed.

driver-scala/build.gradle
rozza: Order not important

rozza:
> > Gradle now requires Java 17 to build.
> 
> That doesn't seem to be true, as I can build with Java 11. Is it the intention that Java 11 will be rejected? Once we start testing our support of records, it will be true. Do we need to short-circuit the build if we detect a Java < 17?

gradlew will use `JAVA_HOME` to run but to build the driver it does require Java 17. The java toolchain feature will do [auto discovery](https://docs.gradle.org/current/userguide/toolchains.html#sec:auto_detection) for Java 17. Try running it with that disabled:

```
./gradlew :bson:compileJava -Porg.gradle.java.installations.auto-detect=false
> Task :bson:compileJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':bson:compileJava'.
> Error while evaluating property 'javaCompiler' of task ':bson:compileJava'
   > Failed to calculate the value of task ':bson:compileJava' property 'javaCompiler'.
      > No compatible toolchains found for request filter: {languageVersion=17, vendor=any, implementation=vendor-specific} (auto-detect true, auto-download false)

```

bson-scala/build.gradle
rozza: It didn't like the `srcDirs +=` - so now we just set it once.

build.gradle
rozza: Reverted this change back

driver-scala/build.gradle
rozza: Moved this back in the config and fixed the intellij issue

driver-scala/build.gradle
```diff
@@ -17,6 +17,13 @@
 description = "A Scala wrapper of the MongoDB Reactive Streams Java driver"
 archivesBaseName = 'mongo-scala-driver'
 
+configurations {
+    integrationTestImplementation {
+        extendsFrom testImplementation
+        canBeResolved = true
```
rozza, line 23: This should fix intellij - also it seems extending from testRuntime wasn't needed.

rozza:
Extra test coverage: https://spruce.mongodb.com/version/6197d036d6d80a78c57cb6c4

driver-scala/build.gradle
jyemin: Indeed, it did.

jyemin (approved):
LGTM.  On to Gradle 8!

rozza:
Manually merged into master

PR #829, opened 2021-11-19
Build: Fix Manifest for jdk.net

rozza:
JAVA-4399

Simple fix - I checked all other calls to `Class.forName` and didn't find any others that should be removed from the manifest.

### DIFF
--- a/driver-core/build.gradle
+++ b/driver-core/build.gradle
@@ -66,6 +66,7 @@ afterEvaluate {
     jar.manifest.attributes['Import-Package'] = [
             '!sun.misc.*',  // Used by DirectBufferDeallocator only for java 8
             '!sun.nio.ch.*',  // Used by DirectBufferDeallocator only for java 8
+            '!jdk.net.*', // Used by SocketStreamHelper depends on JDK version
             'io.netty.*;resolution:=optional',
             'org.xerial.snappy.*;resolution:=optional',
             'com.github.luben.zstd.*;resolution:=optional',


driver-core/build.gradle
```diff
@@ -66,6 +66,7 @@ afterEvaluate {
     jar.manifest.attributes['Import-Package'] = [
             '!sun.misc.*',  // Used by DirectBufferDeallocator only for java 8
             '!sun.nio.ch.*',  // Used by DirectBufferDeallocator only for java 8
+            '!jdk.net.*', // Used by SocketStreamHelper depends on JDK version
```
jyemin, line 69: It is kind of nuts that the osgi build tool is looking at the constant strings that are passed to `Class.forName`.  I played around a bit and found that you can get around it, e.g.:

```
            Class extendedSocketOptionsClass = Class.forName(new StringBuilder()
                    .append(JDK_NET_EXTENDED_SOCKET_OPTIONS_PACKAGE_NAME)
                    .append('.')
                    .append(JDK_NET_EXTENDED_SOCKET_OPTIONS_CLASS_NAME).toString());
```

but this is obviously better.

jyemin (approved):
LGTM.  I'm just not sure what the effect is on OSGi users that are actually on Java 11+.  Will the `Class.forName("jdk.net.ExtendedSocketOptions")` throw a `ClassNotFoundException` because we don't declare it here?  Even if that's true, this is still the right thing to do, but it would be unfortunate.

driver-core/build.gradle
stIncMale: > Will the Class.forName("jdk.net.ExtendedSocketOptions") throw a ClassNotFoundException because we don't declare it here?

My understanding is that it will, unless the bundle the uses our bundle `org.mongodb.driver-core` imports the `jdk.net` package by itself (see [OSGi 3.5 Class Loading Architecture](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3174728):

> Each bundle has a single class loader.

> A class space is then all classes reachable from a given bundle's class loader. Thus, a class space for a given bundle can contain classes from:
> - The parent class loader (normally java.* packages from the boot class path)
> - Imported packages
> - Required bundles
> - The bundle's class path (private packages)
> - Attached fragments

rozza: We've had no feedback regarding that there is an issue for Java 8 and osgi users relating to DirectBufferDeallocator. Which I would have expected had it been an issue. Possibly included via the parent class loader and the boot class path?

rozza: The alternative would be to mark it as optional - be we've only done that for our optional dependencies and not jdk dependencies.

stIncMale: > We've had no feedback regarding that there is an issue for Java 8 and osgi users relating to DirectBufferDeallocator. 

It's because `DirectBufferDeallocator` uses `sun.nio.ch.DirectBuffer`, `sun.misc.Cleaner`, `sun.misc.Unsafe`, which are all in [`sun.*`, which is in `JavaSE-1.8.profile`](https://git.eclipse.org/c/equinox/rt.equinox.framework.git/tree/bundles/org.eclipse.osgi/JavaSE-1.8.profile#n244). However, `jdk.net` is not there.

> Possibly included via the parent class loader and the boot class path?

Yes, it seems that classes in `sun.*` are part of the class space of any bundle via the parent class loader because of the above.

stIncMale (approved):

driver-core/build.gradle
rozza: Because its not in the profile - marking it as optional seems the correct way to handle it.

rozza:
Based on the feedback, rather than not declare it, I've marked it optional. So it should work for all users now.

https://www.mail-archive.com/user@karaf.apache.org/msg22444.html - also seems to think optional is the correct way to handle it.

PR #830, opened 2021-11-24
Ensure MongoClient instances are closed in Scala tests

jyemin:
JAVA-4412

### DIFF
--- a/driver-scala/src/it/scala/org/mongodb/scala/ClientSideEncryptionBypassAutoEncryptionSpec.scala
+++ b/driver-scala/src/it/scala/org/mongodb/scala/ClientSideEncryptionBypassAutoEncryptionSpec.scala
@@ -60,27 +60,30 @@ class ClientSideEncryptionBypassAutoEncryptionSpec extends RequiresMongoDBISpec 
       .codecRegistry(DEFAULT_CODEC_REGISTRY)
       .build
 
-    val clientEncrypted = MongoClient(clientSettings)
+    withTempClient(
+      clientSettings,
+      clientEncrypted => {
 
-    val fieldValue = BsonString("123456789")
+        val fieldValue = BsonString("123456789")
 
-    val dataKeyId = clientEncryption.createDataKey("local", DataKeyOptions()).head().futureValue
+        val dataKeyId = clientEncryption.createDataKey("local", DataKeyOptions()).head().futureValue
 
-    val encryptedFieldValue = clientEncryption
-      .encrypt(fieldValue, EncryptOptions("AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic").keyId(dataKeyId))
-      .head()
-      .futureValue
+        val encryptedFieldValue = clientEncryption
+          .encrypt(fieldValue, EncryptOptions("AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic").keyId(dataKeyId))
+          .head()
+          .futureValue
 
-    val collection: MongoCollection[Document] =
-      clientEncrypted.getDatabase(databaseName).getCollection[Document]("test")
+        val collection: MongoCollection[Document] =
+          clientEncrypted.getDatabase(databaseName).getCollection[Document]("test")
 
-    collection.insertOne(Document("encryptedField" -> encryptedFieldValue)).futureValue
+        collection.insertOne(Document("encryptedField" -> encryptedFieldValue)).futureValue
 
-    val result = collection.find().first().head().futureValue
+        val result = collection.find().first().head().futureValue
 
-    result.get[BsonString]("encryptedField") should equal(Some(fieldValue))
+        result.get[BsonString]("encryptedField") should equal(Some(fieldValue))
 
-    clientEncrypted.close()
+      }
+    )
   }
 
 }
--- a/driver-scala/src/it/scala/org/mongodb/scala/ClientSideEncryptionBypassAutoEncryptionSpec.scala
+++ b/driver-scala/src/it/scala/org/mongodb/scala/ClientSideEncryptionBypassAutoEncryptionSpec.scala
@@ -79,6 +79,8 @@ class ClientSideEncryptionBypassAutoEncryptionSpec extends RequiresMongoDBISpec 
     val result = collection.find().first().head().futureValue
 
     result.get[BsonString]("encryptedField") should equal(Some(fieldValue))
+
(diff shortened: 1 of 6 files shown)


driver-scala/src/it/scala/org/mongodb/scala/ClientSideEncryptionBypassAutoEncryptionSpec.scala
```diff
@@ -79,6 +79,8 @@ class ClientSideEncryptionBypassAutoEncryptionSpec extends RequiresMongoDBISpec
     val result = collection.find().first().head().futureValue
 
     result.get[BsonString]("encryptedField") should equal(Some(fieldValue))
+
+    clientEncrypted.close()
```
jyemin, line 83: There's probably a way to ensure that this happens even when an exception is thrown, but I suppose it's not so important.

driver-scala/src/it/scala/org/mongodb/scala/RequiresMongoDBISpec.scala
```diff
@@ -28,6 +28,25 @@ import scala.util.{ Properties, Try }
 
 trait RequiresMongoDBISpec extends BaseSpec with BeforeAndAfterAll {
 
+  private object GlobalState {
```
jyemin, line 31: This may be the completely wrong Scala idiom for what I'm trying to accomplish, which is to make all these values singletons.  It does seem to work, but happy for any suggestions to make it more idiomatic.

driver-scala/src/it/scala/org/mongodb/scala/RequiresMongoDBISpec.scala
```diff
@@ -53,30 +72,21 @@ trait RequiresMongoDBISpec extends BaseSpec with BeforeAndAfterAll {
...
-    Try(Await.result(MongoClient(mongoClientSettings).listDatabaseNames().toFuture(), WAIT_DURATION)).isSuccess
+    Try(Await.result(GlobalState.mongoClient.listDatabaseNames().toFuture(), WAIT_DURATION)).isSuccess
   }
 
   def hasSingleHost(): Boolean = {
-    new ConnectionString(mongoClientURI).getHosts.size() == 1
+    GlobalState.connectionString.getHosts.size() == 1
   }
 
   def checkMongoDB() {
```
rozza, line 92: Move to the global helper?

driver-scala/src/it/scala/org/mongodb/scala/gridfs/GridFSObservableSpec.scala
```diff
@@ -307,6 +307,8 @@ class GridFSObservableSpec extends RequiresMongoDBISpec with FuturesSpec with Be
     val fileAsDocument = filesCollection.find[BsonDocument]().head().futureValue
     fileAsDocument.getDocument("metadata").getBinary("uuid").getType should equal(4.toByte)
     fileAsDocument.getDocument("metadata").getBinary("uuid").asUuid() should equal(uuid)
+
+    client.close()
```
rozza, line 311: If the trait had a loan method that took a settings then we could wrap the close method.  

```
  def withClient(mongoClientSettings: MongoClientSettings, testCode: MongoClient => Any): Unit = {
    val client = MongoClient(mongoClientSettings)
    try testCode(client) // loan the client
    finally {
      client.close()
    }
  }
```

That could be used for `clientEncrypted` as well.

rozza (changes requested):
I think move the global state helper out into its own file.

Also some small suggestions as well.

driver-scala/src/it/scala/org/mongodb/scala/gridfs/GridFSObservableSpec.scala
jyemin: I didn't quite understand this.  If you think it should be changed, can you add a commit showing how to do it?

driver-scala/src/it/scala/org/mongodb/scala/RequiresMongoDBISpec.scala
```diff
@@ -123,15 +129,21 @@ trait RequiresMongoDBISpec extends BaseSpec with BeforeAndAfterAll {
     false
   } else {
     Await
-      .result(mongoClient().getDatabase("admin").runCommand(Document("isMaster" -> 1)).toFuture(), WAIT_DURATION)
```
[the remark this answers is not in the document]
rozza: Sorry - thats where I realised we were leaking a client

driver-scala/src/it/scala/org/mongodb/scala/gridfs/GridFSObservableSpec.scala
rozza: Done

driver-scala/src/it/scala/org/mongodb/scala/TestMongoClientHelper.scala
```diff
@@ -51,4 +51,12 @@ object TestMongoClientHelper {
   def hasSingleHost: Boolean = {
     TestMongoClientHelper.connectionString.getHosts.size() == 1
   }
+
+  Runtime.getRuntime.addShutdownHook(new ShutdownHook())
+
+  private[mongodb] class ShutdownHook extends Thread {
+    override def run() {
+      TestMongoClientHelper.mongoClient.close()
```
jyemin, line 59: @rozza did you intentionally remove the drop of the database?  Not sure it matters, but want to make sure it is intentional.

rozza: Yeah because its already dropped in the `afterAll` method of the spec test

jyemin: Yeah, I think the reason we have both is in case you Ctrl-C the tests in the middle of the run.

jyemin, line 59: Nit: I think this can become `mongoClient.close()`

rozza: I think in that scenario its fine to leave the database in a undefined state

rozza: :+1:

PR #832, opened 2021-12-01
Iteration should throw if cursor is closed

jyemin:
This changes the behavior of the cursor implementations in the CRUD API
such that attempts to iterate the cursor with hasNext/next/tryNext will
throw IllegalStateException instead of returning already-cached results
from previous iteration.

JAVA-4368

### DIFF
--- a/driver-sync/src/main/com/mongodb/client/internal/MongoBatchCursorAdapter.java
+++ b/driver-sync/src/main/com/mongodb/client/internal/MongoBatchCursorAdapter.java
@@ -34,6 +34,7 @@ public class MongoBatchCursorAdapter<T> implements MongoCursor<T> {
     private final BatchCursor<T> batchCursor;
     private List<T> curBatch;
     private int curPos;
+    private boolean isClosed;
 
     public MongoBatchCursorAdapter(final BatchCursor<T> batchCursor) {
         this.batchCursor = batchCursor;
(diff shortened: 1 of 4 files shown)


stIncMale:
`MongoChangeStreamCursorImpl` and `MongoBatchCursorAdapter` both implement `MongoCursor`. `MongoCursor.close` may be called concurrently with other methods of `MongoCursor`, even though `MongoCursor` is not thread-safe. See [`MongoCursor.close`](https://mongodb.github.io/mongo-java-driver/4.4/apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html#close()). This is currently not taken into account in the PR changes.

jyemin:
What do you suggest?  Does it require that all iteration methods become synchronized?

stIncMale (changes requested):
`MongoChangeStreamCursorImpl` and `MongoBatchCursorAdapter` both implement `MongoCursor`. `MongoCursor.close` may be called concurrently with other methods of `MongoCursor`, even though `MongoCursor` is not thread-safe. See [`MongoCursor.close`](https://mongodb.github.io/mongo-java-driver/4.4/apidocs/mongodb-driver-sync/com/mongodb/client/MongoCursor.html#close()). This is currently not taken into account in the PR changes.

stIncMale:
> What do you suggest? Does it require that all iteration methods become synchronized?

The requirement adds noticeable burden to not otherwise thread-safe code. Serializing executions of all methods that access state accessed by `close`, is the most straightforward approach. We can use `synchronized` with a `private` object, or `StampedLock.asWriteLock()`.

jyemin:
Decided not to merge.  The problem is that `close()` is allowed to be called concurrently with iteration methods, which requires at minimum that `isClosed` be volatile.  There is some cost to that, and it's not clear that it's worth the benefit that we get from throwing while iterating a closed cursor, given how rare that should be in practice.

PR #833, opened 2021-12-01
Add command monitoring unified tests for serverConnectionId

jyemin:
The driver already supports the feature, so this is just test changes.
It includes support for version 1.6 of the unified test format.

JAVA-4238

### DIFF
--- a/driver-core/src/test/resources/unified-test-format/command-monitoring/pre-42-server-connection-id.json
+++ b/driver-core/src/test/resources/unified-test-format/command-monitoring/pre-42-server-connection-id.json
@@ -1,0 +1,101 @@
+{
+  "description": "pre-42-server-connection-id",
+  "schemaVersion": "1.6",
+  "runOnRequirements": [
+    {
+      "maxServerVersion": "4.0.99"
+    }
+  ],
+  "createEntities": [
+    {
+      "client": {
+        "id": "client",
+        "observeEvents": [
+          "commandStartedEvent",
+          "commandSucceededEvent",
+          "commandFailedEvent"
+        ]
+      }
+    },
+    {
+      "database": {
+        "id": "database",
+        "client": "client",
+        "databaseName": "server-connection-id-tests"
+      }
+    },
+    {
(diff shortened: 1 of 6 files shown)


driver-core/src/test/resources/unified-test-format/command-monitoring/pre-42-server-connection-id.json
```diff
@@ -0,0 +1,101 @@
+{
```
jyemin, line 1: This test is skipped entirely in code, but per previous conversation better to still include it in the repository

driver-sync/src/test/functional/com/mongodb/client/unified/CommandMonitoringTest.java
```diff
@@ -38,6 +38,9 @@ public CommandMonitoringTest(@SuppressWarnings("unused") final String fileDescri
                                  final BsonDocument definition) {
         super(schemaVersion, runOnRequirements, entities, initialData, definition);
         assumeFalse(isServerlessTest());
+        // The driver has a hack where getLastError command is executed as part of the handshake in order to get a connectionId
```
jyemin, line 41: This code duplication between reactive and sync tests is unfortunate, but don't want to deal with that in this PR.

driver-sync/src/test/functional/com/mongodb/client/unified/EventMatcher.java
```diff
@@ -68,6 +68,16 @@ public void assertCommandEventsEquality(final String client, final BsonArray exp
                 }
             }
 
+            if (expected.containsKey("hasServerConnectionId")) {
+                boolean hasServerConnectionId = expected.getBoolean("hasServerConnectionId").getValue();
+                Integer serverConnectionId = actual.getConnectionDescription().getConnectionId().getServerValue();
```
jyemin, line 73: Make sure we're checking that there is a server value, since that's the thing that could be null.

stIncMale (approved):

PR #834, opened 2021-12-02
Add error labels to MongoWriteConcernException

jyemin:
The error labels are at the top level of the response
document, and the code was assuming that they were
embedded in the writeConcernError document.  This fix
ensures that error labels at the top level are properly
added to MongoWriteConcernException

JAVA-4419

Note: this change is tested in scope of JAVA-4382 spec test changes, which will follow in a separate PR

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/WriteConcernHelper.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/WriteConcernHelper.java
@@ -61,9 +61,12 @@ public final class WriteConcernHelper {
     }
 
     public static MongoWriteConcernException createWriteConcernException(final BsonDocument result, final ServerAddress serverAddress) {
-        return new MongoWriteConcernException(
+        MongoWriteConcernException writeConcernException = new MongoWriteConcernException(
                 createWriteConcernError(result.getDocument("writeConcernError")),
                 WriteConcernResult.acknowledged(0, false, null), serverAddress);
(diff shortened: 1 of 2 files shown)


driver-core/src/main/com/mongodb/internal/operation/WriteConcernHelper.java
```diff
@@ -61,9 +61,12 @@ public static boolean hasWriteConcernError(final BsonDocument result) {
     }
 
     public static MongoWriteConcernException createWriteConcernException(final BsonDocument result, final ServerAddress serverAddress) {
-        return new MongoWriteConcernException(
+        MongoWriteConcernException writeConcernException = new MongoWriteConcernException(
                 createWriteConcernError(result.getDocument("writeConcernError")),
                 WriteConcernResult.acknowledged(0, false, null), serverAddress);
+        result.getArray("errorLabels", new BsonArray()).stream().map(i -> i.asString().getValue())
+                .forEach(writeConcernException::addLabel);
+        return writeConcernException;
```
jyemin, line 69: It would be natural to consolidate the code in the two methods changed in this PR, but the best place for that it probably in MongoWriteConcernException, and that would require an API changes.  But I want to get this fix into a patch release so am avoiding API changes.  

I think we could do the refactoring as part of a larger consideration of how write concern error are handled, but let me know your thoughts on it.

stIncMale (approved):

PR #835, opened 2021-12-03
Use `@VisibleForTesting` instead of writing verbose comments

stIncMale:


driver-legacy/src/main/com/mongodb/DBCursor.java
```diff
@@ -616,6 +618,7 @@ public int length() {
      * @throws MongoException if failed
      * @see #count()
      */
+    @VisibleForTesting(otherwise = PRIVATE)
```
jyemin, line 621: I'm not convinced about this one.  DBCursor is a public class and this method has been part of the API forever.  I think of VisibleForTesting as an annotation we apply to driver-internal classes as a way to signal to _us_ the intent, not something that we want consumers of our API to be aware of.  AFAIK this would be the first and only API element to be annotated as such, so we're be setting a precedent here that I don't think we'd ever want to repeat.

jyemin (changes requested):

driver-legacy/src/main/com/mongodb/DBCursor.java
stIncMale: Yes, I apologize. `VisibleForTesting` is indeed internal, and it is weird to add it to public API program elements.

Done.

jyemin (approved):

PR #836, opened 2021-12-03
Always specify the known cause when calling `ConnectionPool.invalidate`

stIncMale:
After this [Slack discussion](https://mongodb.slack.com/archives/G016S9LLKU2/p1638376906000700), I realized that there is no reason for us to have both `ConnectionPool.invalidate()` and `ConnectionPool.invalidate(Throwable)`. Moreover, I noticed that there are situations when `DefaultSdamServerDescriptionManager` calls `invalidate` without specifying the cause despite it being known. This PR fixes the discovered issue and removes the unnecessary method.

### DIFF
--- a/driver-core/src/test/unit/com/mongodb/internal/connection/DefaultConnectionPoolSpecification.groovy
+++ b/driver-core/src/test/unit/com/mongodb/internal/connection/DefaultConnectionPoolSpecification.groovy
@@ -152,7 +152,7 @@ class DefaultConnectionPoolSpecification extends Specification {
         connectionFactory.createdConnections.get(0).opened()  // if the first one is opened, they all should be
 
         when: 'the pool is invalidated and the maintenance tasks runs'
-        pool.invalidate()
+        pool.invalidate(null)
         pool.ready()
         pool.doMaintenance()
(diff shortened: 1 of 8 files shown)


driver-core/src/test/unit/com/mongodb/internal/connection/DefaultConnectionPoolSpecification.groovy
```diff
@@ -152,7 +152,7 @@ class DefaultConnectionPoolSpecification extends Specification {
         connectionFactory.createdConnections.get(0).opened()  // if the first one is opened, they all should be
 
         when: 'the pool is invalidated and the maintenance tasks runs'
-        pool.invalidate()
+        pool.invalidate(null)
```
jyemin, line 155: Does production code ever pass null as the argument?  Should it be allowed?

stIncMale: Yes, we do pass `null` here not only in tests. See `DefaultSdamServerDescriptionManager.handleException`, where `SdamIssue` may be created via `SdamIssue.unspecified` and not have an associated exception.

jyemin (approved):

PR #837, opened 2021-12-06
Allow hint for unacknowledged writes if supported by server

jyemin:
JAVA-4275


driver-core/src/main/com/mongodb/internal/operation/OperationHelper.java
```diff
@@ -161,17 +162,27 @@ private static void validateWriteRequestHint(final ConnectionDescription connect
...
+            }
+            if (request instanceof DeleteRequest && serverIsLessThanVersionFourDotFour(connectionDescription)) {
+                throw new IllegalArgumentException(format("Hint not supported by wire version: %s",
+                        connectionDescription.getMaxWireVersion()));
+            }
         }
     }
 
-    static void validateHint(final ConnectionDescription connectionDescription, final WriteConcern writeConcern) {
+    static void validateHintForFindAndModify(final ConnectionDescription connectionDescription, final WriteConcern writeConcern) {
```
jyemin, line 178: Changed the method name because the logic only applies to findAndModify

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/UnifiedCrudTest.java
```diff
@@ -37,6 +39,12 @@ public UnifiedCrudTest(final String fileDescription, final String testDescriptio
         super(schemaVersion, runOnRequirements, entities, initialData, definition);
         this.fileDescription = fileDescription;
         this.testDescription = testDescription;
+        assumeFalse(testDescription.equals("Unacknowledged findOneAndReplace with hint string on 4.4+ server"));
```
jyemin, line 42: Skipping these tests.  See downstream changes comment on JAVA-4275 for an explanation.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudTest.java
```diff
@@ -25,13 +25,21 @@
...
+import static org.junit.Assume.assumeFalse;
+
 public class UnifiedCrudTest extends UnifiedSyncTest {
 
     public UnifiedCrudTest(@SuppressWarnings("unused") final String fileDescription,
                            @SuppressWarnings("unused") final String testDescription,
                            final String schemaVersion, @Nullable final BsonArray runOnRequirements, final BsonArray entities,
                            final BsonArray initialData, final BsonDocument definition) {
         super(schemaVersion, runOnRequirements, entities, initialData, definition);
+        assumeFalse(testDescription.equals("Unacknowledged findOneAndReplace with hint string on 4.4+ server"));
```
jyemin, line 37: Given the class structure there is currently no straightforward way to share these assumptions.  I'm thinking about opening a JAVA ticket to address this shortcoming.

jyemin:
Tests are failing on 3.4, so hold off on reviewing while I take a look

driver-core/src/main/com/mongodb/internal/operation/OperationHelper.java
```diff
@@ -161,17 +162,27 @@ private static void validateWriteRequestHint(final ConnectionDescription connect
         if (serverIsLessThanVersionThreeDotFour(connectionDescription)) {
             throw new IllegalArgumentException(format("Hint not supported by wire version: %s",
                     connectionDescription.getMaxWireVersion()));
-        } else if ((request instanceof DeleteRequest || request instanceof UpdateRequest) && !writeConcern.isAcknowledged()) {
-            throw new MongoClientException("Specifying hints with an unacknowledged WriteConcern is not supported");
+        }
+        if (!writeConcern.isAcknowledged()) {
+            if (request instanceof UpdateRequest && serverIsLessThanVersionFourDotTwo(connectionDescription)) {
+                throw new IllegalArgumentException(format("Hint not supported by wire version: %s",
```
jyemin, line 168: I can't see any rationale why this was a MongoClientException and not an IllegalArgumentException, so I changed it to be consistent.  We never documented one or the other so I don't think it's a breaking change.

jyemin:
OK, should be all set now.

driver-core/src/main/com/mongodb/internal/operation/OperationHelper.java
stIncMale: It's difficult to tell what exception should be used based on the usage of `MongoClientException`. We have `throw new MongoClientException("Read preference in a transaction must be primary")` and `throw new MongoClientException("Unsupported server API version: " + value)`. So `MongoClientException` also seems fine here given its usage in other places.

stIncMale (approved):

PR #838, opened 2021-12-06
Add #available method to legacy cursor API

jyemin:
The new methods are

* com.mongodb.DBCursor#available
* com.mongodb.Cursor#available

with the same semantics as the recently added

* com.mongodb.client.MongoCursor#available

JAVA-4422

### DIFF
--- a/driver-legacy/src/main/com/mongodb/Cursor.java
+++ b/driver-legacy/src/main/com/mongodb/Cursor.java
@@ -28,7 +28,12 @@ import java.util.Iterator;
 public interface Cursor extends Iterator<DBObject>, Closeable {
 
     /**
-     * Gets the number of results available locally without blocking, which may be 0, or 0 when the cursor is exhausted or closed.
+     * Gets the number of results available locally without blocking, which may be 0.
+     *
+     * <p>
+     * If the cursor is known to be exhausted, returns 0.  If the cursor is closed before it's been exhausted, it may return a non-zero
+     * value.
+     * </p>
      *
      * @return the number of results available locally without blocking
      * @since 4.5
--- a/driver-legacy/src/main/com/mongodb/Cursor.java
+++ b/driver-legacy/src/main/com/mongodb/Cursor.java
@@ -28,7 +28,7 @@ import java.util.Iterator;
 public interface Cursor extends Iterator<DBObject>, Closeable {
 
     /**
-     * Gets the number of results available locally without blocking,which may be 0, or 0 when the cursor is exhausted or closed.
+     * Gets the number of results available locally without blocking, which may be 0, or 0 when the cursor is exhausted or closed.
      *
      * @return the number of results available locally without blocking
      * @since 4.5
--- a/driver-legacy/src/main/com/mongodb/Cursor.java
+++ b/driver-legacy/src/main/com/mongodb/Cursor.java
@@ -28,6 +28,14 @@ import java.util.Iterator;
 public interface Cursor extends Iterator<DBObject>, Closeable {
 
     /**
+     * Gets the number of results available locally without blocking,which may be 0, or 0 when the cursor is exhausted or closed.
+     *
+     * @return the number of results available locally without blocking
+     * @since 4.5
+     */
+    int available();
+
+    /**
      * Gets the server's identifier for this Cursor.
      *
      * @return the cursor's ID, or 0 if there is no active cursor.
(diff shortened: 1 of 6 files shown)


driver-legacy/src/main/com/mongodb/Cursor.java
```diff
@@ -27,6 +27,14 @@
  */
 public interface Cursor extends Iterator<DBObject>, Closeable {
 
+    /**
+     * Gets the number of results available locally without blocking,which may be 0, or 0 when the cursor is exhausted or closed.
```
dunkyboy, line 31: (Typo? `which may be 0, or 0 when...`)

jyemin: Not sure what you mean.  (There is a missing space after the comma, but I don't think that's what you're referring to.) Can you write the sentence the way you think it should be? 

Note that I cribbed this from `java.io.InputStream#available`, which says:

> Returns an estimate of the number of bytes that can be read (or skipped over) from this input stream without blocking, which may be 0, or 0 when end of stream is detected.

dunkyboy: Ah, it's just the language is a little confusing - think I get it now. It could return 0 in two scenarios: because there are no results available locally, or because the cursor is closed (the first clause covers the cursor being exhausted, I think?)

Don't feel strongly, just found it a bit tricky to parse.

jyemin: No, the second clause covers both exhausted and closed.  The first clause covers the case where the next call to next/hasNext would have to do I/O to fetch more results, i.e. by executing a getMore, as opposed to the non-zero case where there are still documents in the currently fetched batch (either from the initial find/aggregate or the previous getMore)

rozza (changes requested):
Only nit I have is I'm unsure why the two functional tests differ - would be good to see the DBCollectionAggregationTest.java return 0 when none are available and when the cursor is closed.

jyemin:
> would be good to see the DBCollectionAggregationTest.java return 0 when none are available and when the cursor is closed 

Yes, this is unfortunate.  The problem is discussed in https://github.com/mongodb/mongo-java-driver/pull/832, where @stIncMale  pointed out that applications are supposed to be able to call the close method concurrently with iteration, and iteration should eventually stop.  But there is currently no synchronization between close and iteration, which means that the close method can't modify the state of the object.

Turns out that DBCursor already violates that by setting the wrapped cursor to null in its close method, so we can take advantage of that in the available method by checking for a null cursor.  But MongoCursorAdapter does not (yet) do the same.  The cursor field is currently final.  And the cursor that it wraps will actually report a non-zero value from available even after its closed, and we decided not to fix that due to the concurrency issues.

Given prior decisions, perhaps its best to change the Javadoc to something like this:

> Gets the number of results available locally without blocking, which may be 0, or 0 when the cursor is exhausted.  If the cursor is closed, the behavior is undefined.

rozza:
I think updating the javadoc makes sense - as the behavior is undefined do we need to define it as such?

> Gets the number of results available locally without blocking, which may be 0.

Is the simplest version of the text.

driver-legacy/src/main/com/mongodb/Cursor.java
stIncMale: I agree that the _"may be 0, or 0"_ part is difficult to understand.

stIncMale (changes requested):
> applications are supposed to be able to call the close method concurrently with iteration
> DBCursor already violates that

Note that this requirement is specified only for `MongoCursor.close` and `BatchCursor.close`, the legacy `Cursor.close` does not have it specified.

> But MongoCursorAdapter does not (yet) do the same. The cursor field is currently final. And the cursor that it wraps will actually report a non-zero value from available even after its closed

I agree that we should not specify the behavior of `Cursor.available` for closed objects. Furthermore, the current specification of `MongoCursor.available` is also violated by implementations: `MongoChangeStreamCursorImpl.available` and `MongoBatchCursorAdapter.available` may both return a positive value despite the objects being closed. I.e., we should also remove the specification of the closed behavior from `MongoCursor.available`.

Only the specification of `BatchCursor.available` seems to not be violated: the only actual implementation `QueryBatchCursor` follows it. However, for consistency, I am proposing to change this method's specification similarly to `Cursor.available` and `MongoCursor.available`.

jyemin:
I updated the Javadoc for all three methods.  PTAL.  

@rozza after a quick Slack discussion, I decided to leave in something about the `close` method weirdness.

rozza (approved):

PR #839, opened 2021-12-09
JAVA-4317: Add methods to override listeners in ClusterSettings, ServerSettings and ConnectionPoolSettings

ramasai1:
[JAVA-4317](https://jira.mongodb.org/browse/JAVA-4317)

This PR adds the ability to override listeners in `ClusterSettings`, `ServerSettings` and `ConnectionPoolSettings` by allowing the user to provide a list of listeners. These changes have been modeled after `MongoClientSettings.Builder#commandListenerList`.

QA Performed: Amended unit tests where appropriate.

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/ClusterSettings.java
+++ b/driver-core/src/main/com/mongodb/connection/ClusterSettings.java
@@ -269,6 +269,19 @@ public final class ClusterSettings {
         }
 
         /**
+         * Sets the cluster listeners.
+         *
+         * @param clusterListeners list of cluster listeners
+         * @return this
+         * @since 4.5
+         */
+        public Builder clusterListenerList(final List<ClusterListener> clusterListeners) {
+            notNull("clusterListeners", clusterListeners);
+            this.clusterListeners = new ArrayList<>(clusterListeners);
+            return this;
+        }
+
+        /**
          * Takes the settings from the given {@code ConnectionString} and applies them to the builder
          *
          * @param connectionString the connection string containing details of how to connect to MongoDB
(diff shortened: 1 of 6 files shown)


stIncMale (approved):
Thanks, @ramasai1!

driver-core/src/main/com/mongodb/connection/ClusterSettings.java
```diff
@@ -268,6 +268,19 @@ public Builder addClusterListener(final ClusterListener clusterListener) {
...
         }
 
+        /**
+         * Sets the cluster listeners.
+         *
+         * @param clusterListeners list of cluster listeners
+         * @return this
+         * @since 4.5
+         */
+        public Builder clusterListenerList(final List<ClusterListener> clusterListeners) {
```
jyemin, line 278: @stIncMale I'm wondering if, for consistency with the getter, this method should be named `clusterListeners` instead of `clusterListenerList`.  The "List" is implied by the type.  What do you think?

stIncMale: > The "List" is implied by the type

I agree, but the PR description states that this method was modelled after [`MongoClientSettings.Builder.commandListenerList`](https://mongodb.github.io/mongo-java-driver/4.4/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#commandListenerList(java.util.List)) (there is also [MongoClientSettings.Builder.compressorList](https://mongodb.github.io/mongo-java-driver/4.4/apidocs/mongodb-driver-core/com/mongodb/MongoClientSettings.Builder.html#compressorList(java.util.List))). If you are OK with breaking this consistency, then I have no objections to naming new `*List` methods differently.

jyemin: Ah, I see.  Let's leave as is then.

PR #840, opened 2021-12-09
JAVA-4421: Remove errorLabels from WriteConcernError

ramasai1:
[JAVA-4421](https://jira.mongodb.org/browse/JAVA-4421)

### DIFF
--- a/driver-core/src/main/com/mongodb/bulk/WriteConcernError.java
+++ b/driver-core/src/main/com/mongodb/bulk/WriteConcernError.java
@@ -35,7 +35,7 @@ public class WriteConcernError {
     private final String codeName;
     private final String message;
     private final BsonDocument details;
-    private final Set<String> errorLabels;
+    @Deprecated private final Set<String> errorLabels;
 
 
     /**
--- a/driver-core/src/main/com/mongodb/bulk/WriteConcernError.java
+++ b/driver-core/src/main/com/mongodb/bulk/WriteConcernError.java
@@ -59,7 +59,9 @@ public class WriteConcernError {
      * @param details any details
      * @param errorLabels any error labels
      * @since 4.1
+     * @deprecated Prefer using error labels included in the top level response document
      */
+    @Deprecated
     public WriteConcernError(final int code, final String codeName, final String message, final BsonDocument details,
                              final Set<String> errorLabels) {
         this.code = code;
@@ -113,7 +115,9 @@ public class WriteConcernError {
      * @param errorLabel the non-null error label to add to the exception
      *
      * @since 4.1
+     * @deprecated Prefer using error labels included in the top level response document
      */
+    @Deprecated
     public void addLabel(final String errorLabel) {
         notNull("errorLabel", errorLabel);
         errorLabels.add(errorLabel);
@@ -124,7 +128,6 @@ public class WriteConcernError {
      *
      * @return the error labels, which may not be null but may be empty
      * @since 4.1
-     *
      * @deprecated Prefer using error labels included in the top level response document
      */
     @NonNull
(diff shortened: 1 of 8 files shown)


stIncMale:
> Currently I am unable to run tests. I am running ./gradlew test and nothing happens.

Just in case, I am throwing everything that comes to mind regarding running the tests locally:

- JDK17.
- A accessible MongoDB listening on localhost:27017 and with enabled test commands at the very least. I create mine via [mtools](https://blog.rueckstiess.com/mtools/) with either of the three following commands:
  - `mlaunch init --single --port 27017 --storageEngine=wiredTiger --setParameter enableTestCommands=1 -v`
  - `mlaunch init --replicaset --nodes 3 --port 27017 --storageEngine=wiredTiger --setParameter enableTestCommands=1 -v`
  - `mlaunch init --sharded 2 --mongos 1 --config 1 --replicaset --nodes 3 --port 27017 --storageEngine=wiredTiger --setParameter enableTestCommands=1 -v`
- Compile and run tests:
   - `./gradlew check`

ramasai1:
Thanks! I was running my DB on a different port and was under the impression that it was on 27017. 
Due to 2 constructors being annotated with `@Deprecated` in `MongoWriteConcernException`, there are a couple of deprecation warnings that are part of building the driver. These are in [`MongoOperationPublisher`](https://github.com/mongodb/mongo-java-driver/blob/master/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/MongoOperationPublisher.java#L488) and [`MongoCollectionImpl`](https://github.com/mongodb/mongo-java-driver/blob/master/driver-sync/src/main/com/mongodb/client/internal/MongoCollectionImpl.java#L1014). Should these be marked with `@SuppressWarnings`? Or should I be passing in an empty error labels `BsonArray`? @jyemin @stIncMale

driver-core/src/main/com/mongodb/bulk/WriteConcernError.java
```diff
@@ -35,7 +35,7 @@
     private final String codeName;
     private final String message;
     private final BsonDocument details;
-    private final Set<String> errorLabels;
+    @Deprecated private final Set<String> errorLabels;
```
jyemin, line 38: Moved this to the getter and add an `@deprecated` Javadoc tag.

driver-core/src/main/com/mongodb/internal/operation/WriteConcernHelper.java
```diff
@@ -61,21 +59,18 @@ public static boolean hasWriteConcernError(final BsonDocument result) {
...
     }
 
     public static WriteConcernError createWriteConcernError(final BsonDocument writeConcernErrorDocument) {
         return new WriteConcernError(writeConcernErrorDocument.getNumber("code").intValue(),
                 writeConcernErrorDocument.getString("codeName", new BsonString("")).getValue(),
                 writeConcernErrorDocument.getString("errmsg").getValue(),
-                writeConcernErrorDocument.getDocument("errInfo", new BsonDocument()),
-                writeConcernErrorDocument.getArray("errorLabels", new BsonArray()).stream().map(i -> i.asString().getValue())
-                        .collect(Collectors.toSet()));
+                writeConcernErrorDocument.getDocument("errInfo", new BsonDocument()));
```
jyemin, line 73: What was the rationale for changing this?  The property is deprecated, but until it's removed it should continue to work as it does.

jyemin (changes requested):
> Or should I be passing in an empty error labels BsonArray

Yes, might as well, since right after that the errorLabels are added

ramasai1 (commented):
Added `@Deprecated` tags where necessary.

driver-core/src/main/com/mongodb/bulk/WriteConcernError.java
```diff
@@ -124,8 +124,11 @@ public void addLabel(final String errorLabel) {
      *
      * @return the error labels, which may not be null but may be empty
      * @since 4.1
+     *
+     * @deprecated Prefer using error labels included in the top level response document
```
ramasai1, line 128: Copied this from the ticket because I could not think of a better note.

PR #843, opened 2021-12-16
Support srvServiceName option for SRV lookups

jyemin:
JAVA-4239

### DIFF
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -224,7 +224,7 @@ import static java.util.Collections.unmodifiableList;
  * </ul>
  * <p>SRV configuration:</p>
  * <ul>
- * <li>{@code srvServiceName=string}: The SRV service name.</li>
+ * <li>{@code srvServiceName=string}: The SRV service name. See {@link ClusterSettings#getSrvServiceName()} for details.</li>
  * <li>{@code srvMaxHosts=number}: The maximum number of hosts from the SRV record to connect to.</li>
  * </ul>
  * <p>General configuration:</p>
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -16,6 +16,7 @@
 
 package com.mongodb;
 
+import com.mongodb.connection.ClusterSettings;
 import com.mongodb.connection.ConnectionPoolSettings;
 import com.mongodb.diagnostics.logging.Logger;
 import com.mongodb.diagnostics.logging.Loggers;
@@ -220,6 +221,11 @@ import static java.util.Collections.unmodifiableList;
  * <li>{@code zlibCompressionLevel=integer}: Integer value from -1 to 9 representing the zlib compression level. Lower values will make
  * compression faster, while higher values will make compression better.</li>
  * </ul>
+ * <p>SRV configuration:</p>
+ * <ul>
+ * <li>{@code srvServiceName=string}: The SRV service name.</li>
+ * <li>{@code srvMaxHosts=number}: The maximum number of hosts from the SRV record to connect to.</li>
+ * </ul>
  * <p>General configuration:</p>
  * <ul>
  * <li>{@code retryWrites=true|false}. If true the driver will retry supported write operations if they fail due to a network error.
@@ -254,6 +260,7 @@ public class ConnectionString {
     private final String connectionString;
 
     private Integer srvMaxHosts;
+    private String srvServiceName;
     private Boolean directConnection;
     private Boolean loadBalanced;
     private ReadPreference readPreference;
@@ -404,6 +411,10 @@ public class ConnectionString {
             throw new IllegalArgumentException("srvMaxHosts can only be specified with mongodb+srv protocol");
         }
 
+        if (!isSrvProtocol && srvServiceName != null) {
+            throw new IllegalArgumentException("srvServiceName can only be specified with mongodb+srv protocol");
+        }
+
         if (directConnection != null && directConnection) {
             if (isSrvProtocol) {
                 throw new IllegalArgumentException("Direct connections are not supported when using mongodb+srv protocol");
@@ -482,6 +493,7 @@ public class ConnectionString {
         GENERAL_OPTIONS_KEYS.add("loadbalanced");
 
         GENERAL_OPTIONS_KEYS.add("srvmaxhosts");
+        GENERAL_OPTIONS_KEYS.add("srvservicename");
 
         COMPRESSOR_KEYS.add("compressors");
         COMPRESSOR_KEYS.add("zlibcompressionlevel");
@@ -594,6 +606,8 @@ public class ConnectionString {
                 if (srvMaxHosts < 0) {
                     throw new IllegalArgumentException("srvMaxHosts must be >= 0");
                 }
+            } else if (key.equals("srvservicename")) {
+                srvServiceName = value;
             }
         }
 
@@ -1135,6 +1149,20 @@ public class ConnectionString {
     }
 
     /**
+     * Gets the SRV service name according to RFC 6335, with the exception that it may exceed 15 characters as long as the 63rd (62nd with
+     * prepended underscore) character DNS query limit is not surpassed.
+     *
+     * @return the SRV service name.  Defaults to null in the connection string, but defaults to {@code "mongodb"} in
+     * {@link ClusterSettings}.
+     * @since 4.5
+     * @see ClusterSettings#getSrvServiceName()
+     */
+    @Nullable
+    public String getSrvServiceName() {
+        return srvServiceName;
+    }
+
+    /**
      * Gets the list of hosts
      *
      * @return the host list
@@ -1474,7 +1502,9 @@ public class ConnectionString {
                 && Objects.equals(heartbeatFrequency, that.heartbeatFrequency)
                 && Objects.equals(applicationName, that.applicationName)
                 && Objects.equals(compressorList, that.compressorList)
-                && Objects.equals(uuidRepresentation, that.uuidRepresentation);
+                && Objects.equals(uuidRepresentation, that.uuidRepresentation)
+                && Objects.equals(srvServiceName, that.srvServiceName)
+                && Objects.equals(srvMaxHosts, that.srvMaxHosts);
     }
 
     @Override
@@ -1483,6 +1513,6 @@ public class ConnectionString {
                 writeConcern, retryWrites, retryReads, readConcern, minConnectionPoolSize, maxConnectionPoolSize, maxWaitTime,
                 maxConnectionIdleTime, maxConnectionLifeTime, maxConnecting, connectTimeout, socketTimeout, sslEnabled,
                 sslInvalidHostnameAllowed, requiredReplicaSetName, serverSelectionTimeout, localThreshold, heartbeatFrequency,
-                applicationName, compressorList, uuidRepresentation);
+                applicationName, compressorList, uuidRepresentation, srvServiceName, srvMaxHosts);
     }
 }
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -1150,8 +1150,7 @@ public class ConnectionString {
     }
 
     /**
-     * Gets the SRV service name according to RFC 6335, with the exception that it may exceed 15 characters as long as the 63rd (62nd with
-     * prepended underscore) character DNS query limit is not surpassed.
+     * Gets the SRV service name.
      *
      * @return the SRV service name.  Defaults to null in the connection string, but defaults to {@code "mongodb"} in
      * {@link ClusterSettings}.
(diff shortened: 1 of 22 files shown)


driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -1474,7 +1502,9 @@ public boolean equals(final Object o) {
                 && Objects.equals(heartbeatFrequency, that.heartbeatFrequency)
                 && Objects.equals(applicationName, that.applicationName)
                 && Objects.equals(compressorList, that.compressorList)
-                && Objects.equals(uuidRepresentation, that.uuidRepresentation);
+                && Objects.equals(uuidRepresentation, that.uuidRepresentation)
+                && Objects.equals(srvServiceName, that.srvServiceName)
+                && Objects.equals(srvMaxHosts, that.srvMaxHosts);
     }
```
jyemin, line 1508: srvMaxHosts wasn't added before, so adding both here

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -1134,6 +1148,20 @@ public Integer getSrvMaxHosts() {
         return srvMaxHosts;
     }
 
+    /**
+     * Gets the SRV service name according to RFC 6335, with the exception that it may exceed 15 characters as long as the 63rd (62nd with
+     * prepended underscore) character DNS query limit is not surpassed.
```
stIncMale, line 1153: I found where the 15 characters limit comes from (https://www.rfc-editor.org/rfc/rfc6335.html#section-5.1), but failed to find anything about the 62/63 character limit in both https://www.rfc-editor.org/rfc/rfc6335 and https://www.rfc-editor.org/rfc/rfc2782. Where is this limit specified? It may also be useful to express the limits as hyperlinks to the places where they are specified.

Another thing that confuses me is the "rd" and "nd" ordinal indicators. I read those as "sixty-thi**rd**" and "sixty-seco**nd**" ordinals, i.e., numbers that express position, not size/length/cardinality. And we need a cardinal number (not an ordinal one) when expressing the limit. Am I interpreting the "rd" and "nd" incorrectly here?

P.S. This comment is equally applicable to the corresponding methods in `ClusterSettings`.

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -220,6 +221,11 @@
  * <li>{@code zlibCompressionLevel=integer}: Integer value from -1 to 9 representing the zlib compression level. Lower values will make
  * compression faster, while higher values will make compression better.</li>
  * </ul>
+ * <p>SRV configuration:</p>
+ * <ul>
+ * <li>{@code srvServiceName=string}: The SRV service name.</li>
+ * <li>{@code srvMaxHosts=number}: The maximum number of hosts from the SRV record to connect to.</li>
```
stIncMale, line 227: Should `MongoClientURI`/`MongoClientOptions` also support these options, or are we not adding support for any new options to those classes?

driver-core/src/main/com/mongodb/connection/ClusterSettings.java
```diff
@@ -156,6 +161,19 @@ public Builder srvMaxHosts(final Integer srvMaxHosts) {
             return this;
         }
 
+        /**
+         * Sets the SRV service name according to RFC 6335, with the exception that it may exceed 15 characters as long as the 63rd (62nd
+         * with prepended underscore) character DNS query limit is not surpassed..
```
stIncMale, line 166: An excessive comma in _"surpassed.."_

stIncMale (changes requested):
The class `SrvPollingProseTests` appears to have 12 tests, i.e., the [13th test](https://github.com/mongodb/specifications/commit/153d3b1c96137eed1765c76af1ff3e6f68cbe07d) is missing. It was supposed to be added by [JAVA-4349](https://jira.mongodb.org/browse/JAVA-4349), and it [was decided](https://jira.mongodb.org/browse/JAVA-4349?focusedCommentId=4233802&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4233802) that this ticket should be done together with [JAVA-4239](https://jira.mongodb.org/browse/JAVA-4239).

jyemin (commented):
Addressed all comments except one where I asked a followup question.

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
```diff
@@ -52,16 +52,16 @@
...
         List<String> srvHostDomainParts = asList(srvHostDomain.split("\\."));
         List<String> hosts = new ArrayList<String>();
         InitialDirContext dirContext = createDnsDirContext();
         try {
-            Attributes attributes = dirContext.getAttributes("_mongodb._tcp." + srvHost, new String[]{"SRV"});
+            Attributes attributes = dirContext.getAttributes("_" + srvServiceName + "._tcp." + srvHost, new String[]{"SRV"});
             Attribute attribute = attributes.get("SRV");
             if (attribute == null) {
-                throw new MongoConfigurationException("No SRV records available for host " + srvHost);
+                throw new MongoConfigurationException("_" + "No SRV records available for " + "_" + srvServiceName + "._tcp." + srvHost);
```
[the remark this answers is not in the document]
stIncMale: Based on the [resource record (RR) format](https://www.rfc-editor.org/rfc/rfc1035#section-3.2.1), this triplet is the part of a resource record called "name", which matches its meaning from the perspective of the `DirContext.getAttributes` method. So it is definitely correct to call the triplet `resourceRecordName`<sup>1</sup>.

---
<sup>1</sup> The "resourceRecord" prefix is useful to avoid the ambiguity caused by the fact that RRs of the [`type`](https://www.rfc-editor.org/rfc/rfc1035#section-3.2.2) 33 (`SRV`) introduce structure to the `name` (both the type value and the structure are specified in [RFC2782](https://www.rfc-editor.org/rfc/rfc2782)), and call one of its parts "name", which overloads the term.

driver-core/src/main/com/mongodb/ConnectionString.java
stIncMale: I see. It's here [here|https://github.com/stIncMale/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#srvservicename] in our spec, and the same "rd" and "nd" ordinal indicators are used there. Since you did not say anything about them, I deduce that you see no problem with the way the limit is expressed?

As for the actual source of the limit, I found that the [preferred name syntax](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1) states that the name is a sequence of labels joined with `.`. Each label is limited to 63 characters, and the whole name is limited to 255 characters according to the [size limits](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.4)<sup>1</sup>. For an SRV record, the `_Service` part of the RR name is a label from the standpoint of the RR name, i.e., it is limited to 63 characters.

Now we have the links to the actual source of the limit, i.e., the situation has changed, and so I again propose to include the links in the docs to explain where the limit comes from.

---
<sup>1</sup> The RFC uses both terms "character" and "octet" when mentions the limit in different places, but given what characters are [allowed to be used as part of the name](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1), they are all encoded with a single octet.

stIncMale (commented):
Some discussions are still ongoing, and I am not sure if you noticed or missed the [comment](https://github.com/mongodb/mongo-java-driver/pull/843#pullrequestreview-838037660) about the missing prose test. It is also possible that the test is not missing, and I misunderstand the situation.

driver-core/src/main/com/mongodb/ConnectionString.java
jyemin: I'm not seeing anything about "rd" or "nd" in the driver spec.  Where exactly are you seeing it?

I'm still unsure why we should say anything more than what's in the spec:

> This option specifies a valid SRV service name according to RFC 6335, with the exception that it may exceed 15 characters as long as the 63rd (62nd with prepended underscore) character DNS query limit is not surpassed. This option requires a string value and defaults to "mongodb". 

but if it's insufficient can you suggest the exact change you'd like to see?

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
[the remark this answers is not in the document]
jyemin: Done

jyemin:
> The class SrvPollingProseTests appears to have 12 tests, i.e., the 13th test is missing. It was supposed to be added by JAVA-4349, and it was decided that this ticket should be done together with JAVA-4239.

The test says this:

> The following test MUST setup a MongoClient using the ``srvServiceName`` option
and the ``test22.test.build.10gen.cc`` SRV record. The test MUST mock the described
DNS changes and then verify that the new list of hosts is present.

The Java driver is not able to mock the DNS resolver with a MongoClient.

We do, however, have the test `DefaultDnsSrvRecordMonitorSpecification#should discover new resolved hosts` which, I think, serves the same purpose.

What do you think?  Can you see a better way to test this?

stIncMale:
> We do, however, have the test DefaultDnsSrvRecordMonitorSpecification#should discover new resolved hosts which, I think, serves the same purpose.

Thanks for pointing out the test. I agree, it is a fine substitute. Can we leave a comment in [JAVA-4349](https://jira.mongodb.org/browse/JAVA-4349), stating that this test is used in lieu of the prose test required by the ticket?

jyemin:
Sure: done.

stIncMale:
[_"make do"_](https://jira.mongodb.org/browse/JAVA-4349?focusedCommentId=4264879&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4264879) - I learned a new phrase 😄

driver-core/src/main/com/mongodb/ConnectionString.java
stIncMale: > I'm not seeing anything about "rd" or "nd" in the driver spec. Where exactly are you seeing it?

It's in _"63rd"_ and _"62nd"_. My interpretation and confusion are described in the second paragraph in the [first comment](https://github.com/mongodb/mongo-java-driver/pull/843#discussion_r773590353).

> why we should say anything more than what's in the spec

- At first, since we decided to specify the limits, I wanted to have links to the relevant part of relevant RFCs where the limits are specified.
- When I found those links (took time, and I don't want to burden users with the exercise, it's enough that our specification burdens us) and read parts of RFCs, I ended up having more questions:
  - Why do we suggest users to disregard the 15 characters limit (applicable to SRV records) and instead care about a more generic 63 characters limit (applicable to all DNS resource records)?
  - Why do we not mention the 255 characters limit on the whole DNS resource record name? Given that users supply both the service name and the host name parts of the SRV resource record name (specified by [RFC2782](https://www.rfc-editor.org/rfc/rfc2782)), this limit appears to be at least as relevant as the other two limits (arguably, more relevant than the 63 characters limit overridden by the 15 characters limit).

The links to all limits:
  - https://www.rfc-editor.org/rfc/rfc6335#section-5.1 for the 15 characters limit on the SRV service name;
  - https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1 for the 63 characters limit on the label part of a resource record name (the SRV service name is a label in terms of this RFC);
  - https://www.rfc-editor.org/rfc/rfc1035#section-2.3.4 for the 255 characters limit on the whole resource record name.

> can you suggest the exact change you'd like to see?

For example:
The [SRV resource record](https://www.rfc-editor.org/rfc/rfc2782) ([RFC 2782](https://www.rfc-editor.org/rfc/rfc2782)) [service name, which is limited to 15 characters](https://www.rfc-editor.org/rfc/rfc6335#section-5.1) ([RFC 6335 section 5.1](https://www.rfc-editor.org/rfc/rfc6335#section-5.1)). If specified, it is combined with the single host name specified by [`ConnectionString.getHosts`](https://mongodb.github.io/mongo-java-driver/4.4/apidocs/mongodb-driver-core/com/mongodb/ConnectionString.html#getHosts()) as follows: `_srvServiceName._tcp.hostName`. The combined string is an SRV [resource record name](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1) ([RFC 1035 section 2.3.1](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1)), which is [limited to 255 characters](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.4) ([RFC 1035 section 2.3.4](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.4)).

jyemin: Done.  I duplicated it in four places (getter and setter for MongoClientOptions/Settings) but not ConnectionString, where I just linked to the ClusterSettings docs.  That seems to fit the style better (brevity).

stIncMale: Thanks 👍 The previous description is left on `MongoClientURI.getSrvServiceName`. If you remove it, there is no need to re-request review, the PR can then be merged right away.

stIncMale (approved):

PR #844, opened 2021-12-20
Support codec for Java records

jyemin:
JAVA-3567

This is a draft pull request to get feedback on the approach.  Of note:

1. It's implemented as a separate module, bson-record-codec, with a minimum release dependency of Java 17
2. It is not dependent on any POJO Codec support classes.  
3. There is no support for conventions.  The provider is implicitly automatic.  It supports all record types.
4. There is no automatic conversion of "id" to "_id".  You have to either name the field "_id" or annotate with BsonId.  
5. It redefines the annotations so that they can be added to the record component and only the record component.  Also, I can't think of a straightforward way to add `ElementType.RECORD_COMPONENT` to `@Target` of the existing annotations, since that requires Java 17. 
6. It doesn't support generics at all, even for nested records.  It's not clear if they are needed, but probably some users will have a use case for it.
7. It doesn't support discriminators at all,  It's not clear if they are needed, but again, they probably are.  It's just a bit weird because records are final and so the only base type would be an interface they is implemented by multiple record types.
8. It does not serialize null values. It just leaves the field out entirely (I think PojoCodec is the same).
9. The tests don't run properly in Evergreen yet because it requires Java 17.
10. Javadoc generation is failing on Evergreen for some reason I don't understand.  It works locally.

### DIFF
--- a/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
+++ b/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
@@ -165,9 +165,10 @@ final class RecordCodec<T extends Record> implements Codec<T> {
     }
 
     private static <T> List<ComponentModel> getComponentModels(final Class<T> clazz, final CodecRegistry codecRegistry) {
-        var componentModels = new ArrayList<ComponentModel>(clazz.getRecordComponents().length);
-        for (int i = 0; i < clazz.getRecordComponents().length; i++) {
-            componentModels.add(new ComponentModel(clazz.getRecordComponents()[i], codecRegistry, i));
+        var recordComponents = clazz.getRecordComponents();
+        var componentModels = new ArrayList<ComponentModel>(recordComponents.length);
+        for (int i = 0; i < recordComponents.length; i++) {
+            componentModels.add(new ComponentModel(recordComponents[i], codecRegistry, i));
         }
         return componentModels;
     }
--- a/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
+++ b/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
@@ -1,0 +1,217 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bson.codecs.record;
+
+import org.bson.BsonReader;
+import org.bson.BsonType;
+import org.bson.BsonWriter;
+import org.bson.codecs.Codec;
+import org.bson.codecs.DecoderContext;
+import org.bson.codecs.EncoderContext;
+import org.bson.codecs.RepresentationConfigurable;
+import org.bson.codecs.configuration.CodecConfigurationException;
+import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.codecs.record.annotations.BsonProperty;
+import org.bson.codecs.record.annotations.BsonId;
+import org.bson.codecs.record.annotations.BsonRepresentation;
+
+import javax.annotation.Nullable;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.RecordComponent;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+import static org.bson.assertions.Assertions.notNull;
+
+final class RecordCodec<T extends Record> implements Codec<T> {
+    private final Class<T> clazz;
+    private final Constructor<?> canonicalConstructor;
+    private final List<ComponentModel> componentModels;
+    private final ComponentModel componentModelForId;
+    private final Map<String, ComponentModel> fieldNameToComponentModel;
+
+    private static final class ComponentModel {
+        private final RecordComponent component;
+        private final Codec<?> codec;
+        private final int index;
+        private final String fieldName;
+
+        private ComponentModel(final RecordComponent component, final CodecRegistry codecRegistry, final int index) {
+            this.component = component;
+            this.codec = computeCodec(component, codecRegistry);
+            this.index = index;
+            this.fieldName = computeFieldName(component);
+        }
+
+        String getComponentName() {
+            return component.getName();
+        }
+
+        String getFieldName() {
+            return fieldName;
+        }
+
+        Object getValue(final Record record) throws InvocationTargetException, IllegalAccessException {
+            return component.getAccessor().invoke(record);
+        }
+
+        private static Codec<?> computeCodec(final RecordComponent component, final CodecRegistry codecRegistry) {
+            var codec = codecRegistry.get(toWrapper(component.getType()));
+            var bsonRepresentationAnnotation = component.getAnnotation(BsonRepresentation.class);
+            if (bsonRepresentationAnnotation != null) {
+                if (codec instanceof RepresentationConfigurable<?> representationConfigurable) {
+                    codec = representationConfigurable.withRepresentation(bsonRepresentationAnnotation.value());
+                } else {
+                    throw new CodecConfigurationException(
+                            format("Codec for %s must implement RepresentationConfigurable to support BsonRepresentation",
+                                    codec.getEncoderClass()));
+                }
+            }
+            return codec;
+        }
+
+        private static String computeFieldName(final RecordComponent component) {
+            if (component.isAnnotationPresent(BsonId.class)) {
+                return "_id";
+            } else if (component.isAnnotationPresent(BsonProperty.class)) {
+                return component.getAnnotation(BsonProperty.class).value();
+            }
+            return component.getName();
+        }
+    }
+
+    RecordCodec(final Class<T> clazz, final CodecRegistry codecRegistry) {
+        this.clazz = notNull("class", clazz);
+        canonicalConstructor = notNull("canonicalConstructor", getCanonicalConstructor(clazz));
+        componentModels = getComponentModels(clazz, codecRegistry);
+        fieldNameToComponentModel = componentModels.stream()
+                .collect(Collectors.toMap(ComponentModel::getFieldName, Function.identity()));
+        componentModelForId = getComponentModelForId(clazz, componentModels);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public T decode(final BsonReader reader, final DecoderContext decoderContext) {
+        reader.readStartDocument();
+
+        Object[] constructorArguments = new Object[componentModels.size()];
+        while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
+            var fieldName = reader.readName();
+            var componentModel = fieldNameToComponentModel.get(fieldName);
+            constructorArguments[componentModel.index] = decoderContext.decodeWithChildContext(componentModel.codec, reader);
+        }
+        reader.readEndDocument();
+
+        try {
+            return (T) canonicalConstructor.newInstance(constructorArguments);
+        } catch (ReflectiveOperationException e) {
+            throw new CodecConfigurationException(format("Unable to invoke canonical constructor of record class %s", clazz.getName()), e);
+        }
+    }
+
+    @Override
+    public void encode(final BsonWriter writer, final T record, final EncoderContext encoderContext) {
+        writer.writeStartDocument();
+        if (componentModelForId != null) {
+            writeComponent(writer, record, componentModelForId);
+        }
+        for (var componentModel : componentModels) {
+            if (componentModel == componentModelForId) {
+                continue;
+            }
+            writeComponent(writer, record, componentModel);
+        }
+        writer.writeEndDocument();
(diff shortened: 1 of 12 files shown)


bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -0,0 +1,206 @@
...
+            var fieldName = reader.readName();
+            var componentModel = fieldNameToComponentModel.get(fieldName);
+            constructorArguments[componentModel.index] = decoderContext.decodeWithChildContext(componentModel.codec, reader);
+        }
+        reader.readEndDocument();
+
+        try {
+            return (T) canonicalConstructor.newInstance(constructorArguments);
+        } catch (ReflectiveOperationException e) {
+            throw new CodecConfigurationException(format("Unable to invoke canonical constructor of record class %s", clazz.getName()), e);
```
jyemin, line 128: Seems like the wrong exception, but not sure what the right one is.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
+    @Override
+    public void encode(final BsonWriter writer, final T record, final EncoderContext encoderContext) {
+        writer.writeStartDocument();
+        if (componentModelForId != null) {
```
jyemin, line 135: Write _id first, as with other document codecs.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -0,0 +1,206 @@
...
+    @Override
+    public Class<T> getEncoderClass() {
+        return clazz;
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private void writeComponent(final BsonWriter writer, final T record, final ComponentModel componentModel) {
+        try {
+            Object componentValue = componentModel.getValue(record);
+            if (componentValue != null) {
```
jyemin, line 157: Skips nulls entirely

bson-record-codec/src/main/org/bson/codecs/record/annotations/BsonId.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * An annotation that configures the record component as the _id field of the document
+ *
+ * @since 4.XXX
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.RECORD_COMPONENT})
```
jyemin, line 32: To avoid confusion, only allow the annotations on the record component itself, since that's the only place it's going to be looked for.

bson-record-codec/src/test/unit/org/bson/codecs/record/samples/TestRecord.java
```diff
@@ -0,0 +1,35 @@
...
+package org.bson.codecs.record.samples;
+
+import org.bson.BsonType;
+import org.bson.codecs.record.annotations.BsonId;
+import org.bson.codecs.record.annotations.BsonProperty;
+import org.bson.codecs.record.annotations.BsonRepresentation;
+
+import java.util.List;
+
+public record TestRecord(String name,
```
jyemin, line 26: Weirdly, this one record allows most everything to be tested.

gradle/javaToolchain.gradle
```diff
@@ -31,7 +31,12 @@ allprojects {
         }
     }
 
-    if (project in javaMainProjects) {
+    if (project == project(":bson-record-codec")) {
```
jyemin, line 34: This could probably be cleaner.  And I don't know how to make sure the tests only run on Java 17+

jyemin: @rozza probably will need your assistance on this one.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -0,0 +1,217 @@
...
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public T decode(final BsonReader reader, final DecoderContext decoderContext) {
+        reader.readStartDocument();
+
+        Object[] constructorArguments = new Object[componentModels.size()];
+        while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
+            var fieldName = reader.readName();
+            var componentModel = fieldNameToComponentModel.get(fieldName);
```
stIncMale, line 120: [performance-related mumbling, no action is expected]

As far as I remember (and I may be completely wrong here), some of the optimizations that @BorisDog prototyped for POCOs in the C# driver are:
- Rely on the order in which fields are serialized when decoding: instead of looking up in a map, one may simply pick the next component model from `componentModels` (this implies that `componentModelForId` is excluded form `componentModels`).
- Since `null` values are not serialized, the previous optimization may result in the need to skip some of the `componentModels`. This means that the `ComponentModel.getFieldName` still needs to be checked agains the `fieldName` from the `reader`, however, such a comparison is less resource-consuming than `fieldNameToComponentModel.get`, because the latter does the same check in addition to other activities. Such a comparison may be further speed up by storing a BSON binary representation of the field name in `ComponentModel`, and then comparing it with a binary representation of the field name from `BsonReader` instead of decoding that binary representation into a `String` and then comparing strings (our `BsonReader` does not seem to support doing so).

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -0,0 +1,217 @@
...
+                ((Codec) componentModel.codec).encode(writer, componentValue, EncoderContext.builder().build());
+            }
+        } catch (ReflectiveOperationException e) {
+            throw new CodecConfigurationException(
+                    format("Unable to access value of component %s for record %s", componentModel.getComponentName(), clazz.getName()), e);
+        }
+    }
+
+    private static <T> List<ComponentModel> getComponentModels(final Class<T> clazz, final CodecRegistry codecRegistry) {
+        var componentModels = new ArrayList<ComponentModel>(clazz.getRecordComponents().length);
```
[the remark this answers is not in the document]
jyemin: I don't think we care much about performance here because this is called once per record class, and not for every encode/decode.  But aesthetically it seems more pleasing to store in a local variable rather than call it three times.

bson-record-codec/src/main/org/bson/codecs/record/annotations/BsonProperty.java
```diff
@@ -0,0 +1,47 @@
...
+@Documented
+@Target({ElementType.RECORD_COMPONENT})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface BsonProperty {
+    /**
+     * The field name of the record component.
+     *
+     * @return the field name to use for the record component
+     */
+    String value() default "";
```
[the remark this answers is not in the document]
jyemin: Empty string is actually allowed as a field name, so I don't think it's necessary or correct to have this check.  The bigger problem is if multiple components end up with the same name.  It would be trivial to check for that, but I don't think POJO codec does either.

bson-record-codec/src/test/unit/org/bson/codecs/record/samples/TestRecord.java
```diff
@@ -0,0 +1,35 @@
...
+                         @BsonProperty("a") int age,
+                         List<String> hobbies,
+                         @BsonRepresentation(BsonType.OBJECT_ID) @BsonId String identifier) {
+
+    // To test that the default constructor is always used for decoding
```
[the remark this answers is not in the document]
jyemin: Done.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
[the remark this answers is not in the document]
jyemin: Done

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
rozza: `CodecConfigurationException` is the grab bag for all manner of codec exceptions

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -0,0 +1,217 @@
...
+        String getFieldName() {
+            return fieldName;
+        }
+
+        Object getValue(final Record record) throws InvocationTargetException, IllegalAccessException {
+            return component.getAccessor().invoke(record);
+        }
+
+        private static Codec<?> computeCodec(final RecordComponent component, final CodecRegistry codecRegistry) {
+            var codec = codecRegistry.get(toWrapper(component.getType()));
```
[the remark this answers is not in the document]
rozza: Yes, the PojoCodec uses the `TypeData` which has a `boxType` method and a `isPrimitive` check

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
rozza: This is different to POJOs that check if values should be serialized or not via `PropertySerialization`

bson-record-codec/src/main/org/bson/codecs/record/annotations/BsonId.java
rozza: Ack

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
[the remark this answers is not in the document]
jyemin: I don't think it's worth refactoring to try to share this, but we can discuss again in final review

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
jyemin: Ah, I see.  Seems like the default behavior for POJOs is the same as this, as there is only one implementation of `PropertySerialization` in the driver (`PropertyModelSerializationImpl`), and it has the same behavior:

```
    public boolean shouldSerialize(final T value) {
        return value != null;
    }
```

If we end up adding convention support for records, then something like `PropertySerialization` would likely be included.

bson-record-codec/src/main/org/bson/codecs/record/annotations/BsonProperty.java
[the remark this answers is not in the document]
stIncMale: > Empty string is actually allowed as a field name

Unexpected, but cool) Nevertheless, do you think that specifying the default value for `BsonProperty.value` is actually useful? To me it seems that the API should push users into providing a field name, rather than tolerating no explicit name. The latter allows to specify multiple `@BsonProperty` annotations with no explicit value, i.e., with the same empty field name.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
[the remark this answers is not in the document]
stIncMale: If we move `TypeData` to an internal package and make `public` (it is currently package-access), then we can use `TypeData.builder(component.getType()).build().getType()` instead of `toWrapper(component.getType())`.

gradle/javaToolchain.gradle
rozza: No problems will take a look - this is another good reason to fix our gradle build and make it more configurable per project and explicit rather than this older approach of an uber gradle configuration.

rozza (commented):
Its hard to review, as the type support is exceptionally simple at present which is good in that its standalone but is it too simple for our users?

We would need to add enum support, `record EnumRecord(ObjectId _id, SimpleEnum simpleEnum){}` would fail but the pojo equivalent wouldn't.
The lack of generic support is concerning to me as users will require concrete generic fields and probably will require implementations of generic interfaces for the record itself.

I think this new feature should be scoped so we can look at what should be supported and what shouldn't. If the goal is to support users current using the pojos to be able to migrate to the records then there's much more complexity to be added.

bson-record-codec/src/test/unit/org/bson/codecs/record/RecordCodecTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.bson.conversions.Bson;
+import org.bson.types.ObjectId;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class RecordCodecTest {
+    private final RecordCodec<TestRecord> codec = new RecordCodec<>(TestRecord.class, Bson.DEFAULT_CODEC_REGISTRY);
```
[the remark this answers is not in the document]
jyemin: Ack

bson-record-codec/src/main/org/bson/codecs/record/annotations/BsonProperty.java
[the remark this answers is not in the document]
jyemin: I think that it was done this way on the existing BsonProperty annotation so that if you just wanted to specify `useDiscriminator` you could do so, and empty string for `value` was just ignored.  I can remove the default for now, but we might want it in the future if we add more properties to the annotation.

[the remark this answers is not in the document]
stIncMale: > I can remove the default for now, but we might want it in the future if we add more properties to the annotation.

Adding the empty default in the future and start ignoring it (if we don't then it defeats the purpose in the scenario you described) will be a breaking change.

> the existing BsonProperty annotation ... empty string for value was just ignored

Since the POJO codec does not allow using an empty string (ignores the value and uses the name of the Java field), maybe `RecordCodec` should have the same behavior? I know I mentioned not allowing empty names earlier, and you said they are allows, but with the new information you provided, it seems like `RecordCodec` behaves differently from the POJO codec, and I am not sure it's a good thing.

jyemin:
This PR has been superseded by https://github.com/mongodb/mongo-java-driver/pull/910 and https://github.com/mongodb/mongo-java-driver/pull/911, which have both been merged and will be out soon in the 4.6 release of the driver.  Thanks everyone for your patience.

PR #846, opened 2021-12-26
Add support for attaching elements to the DecoderContext

evanchooly:
I would like to discuss adding something like this to the DecoderContext.  This would allow me to attach a cache in Morphia so that when I'm fetching referenced documents, I can cache them and reuse the hydrated objects rather than reprocessing them all over and over.  The precise shape of this API could no doubt use some fine tuning but this is the general idea.

### DIFF
--- a/bson/src/main/org/bson/codecs/DecoderContext.java
+++ b/bson/src/main/org/bson/codecs/DecoderContext.java
@@ -42,9 +42,7 @@ public final class DecoderContext {
      * @return the new DecoderContext
      */
     public DecoderContext attach(String key, Object object) {
-        return builder()
-                .attachments(attachments)
-                .checkedDiscriminator(checkedDiscriminator)
+        return builder(this)
                 .attach(key, object)
                 .build();
     }
@@ -66,6 +64,7 @@ public final class DecoderContext {
     public <T> T attachment(String key) {
         return (T) attachments.get(key);
     }
+
     /**
      * Create a builder.
      *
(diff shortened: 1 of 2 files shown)


bson/src/main/org/bson/codecs/DecoderContext.java
```diff
@@ -29,14 +32,40 @@
...
+     * Attaches an item to this context with the given name.  Returns a new modified context.
+     *
+     * @param key the key for the attachment
+     * @param object the item to attach
+     * @return the new DecoderContext
+     */
+    public DecoderContext attach(String key, Object object) {
+        return builder()
+                .attachments(attachments)
+                .checkedDiscriminator(checkedDiscriminator)
```
evanchooly, line 47: these two lines, e.g., could be in a private "copy constructor" on `Builder` that handles copying the state.

evanchooly:
bump.  any comments?  Even making these classes non-final would work for.  I could use my own implementations instead.

rozza:
Hi @evanchooly !

Thanks for the bump - I've added [JAVA-4444](https://jira.mongodb.org/browse/JAVA-4444) so we can triage this.

Cheers,

Ross

jyemin:
@evanchooly on JAVA-4444 you said you would play around with this a bit more and get back to us.  Since we haven't heard from you yet, I'm going to close this out, bearing in mind that we can always re-open it if necessary.

PR #850, opened 2022-01-06
Ensure insertedIds contain ids from all batches

jyemin:
The previous code was incorrect because it was comparing absolute write indexes
with indexes that are relative to the current batch. This patch avoids that
by using the insertedId map from SplittablePayload directly, which already
contains absolute write indexes.

JAVA-4436

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
@@ -53,12 +53,11 @@ import org.bson.codecs.Decoder;
 import org.bson.codecs.configuration.CodecRegistry;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static com.mongodb.internal.bulk.WriteRequest.Type.DELETE;
 import static com.mongodb.internal.bulk.WriteRequest.Type.INSERT;
@@ -282,21 +281,11 @@ public final class BulkWriteBatch {
     }
 
     private List<BulkWriteInsert> getInsertedItems(final BsonDocument result) {
-        if (payload.getPayloadType() == SplittablePayload.Type.INSERT) {
-
-            Stream<WriteRequestWithIndex> writeRequests = payload.getWriteRequestWithIndexes().stream();
-            List<Integer> writeErrors = getWriteErrors(result).stream().map(BulkWriteError::getIndex).collect(Collectors.toList());
-            if (!writeErrors.isEmpty()) {
-                writeRequests = writeRequests.filter(wr -> !writeErrors.contains(wr.getIndex()));
-            }
-            if (payload.getPosition() < payload.size()) {
-                writeRequests = writeRequests.filter(wr -> wr.getIndex() < payload.getPosition());
-            }
-            return writeRequests
-                    .map(wr -> new BulkWriteInsert(wr.getIndex(), payload.getInsertedIds().get(wr.getIndex())))
-                    .collect(Collectors.toList());
-        }
-        return Collections.emptyList();
+        Set<Integer> writeErrors = getWriteErrors(result).stream().map(BulkWriteError::getIndex).collect(Collectors.toSet());
+        return payload.getInsertedIds().entrySet().stream()
+                .filter(entry -> !writeErrors.contains(entry.getKey()))
+                .map(entry -> new BulkWriteInsert(entry.getKey(), entry.getValue()))
+                .collect(Collectors.toList());
     }
 
 
--- a/driver-core/src/test/unit/com/mongodb/internal/operation/BulkWriteBatchSpecification.groovy
+++ b/driver-core/src/test/unit/com/mongodb/internal/operation/BulkWriteBatchSpecification.groovy
@@ -16,6 +16,7 @@
 
 package com.mongodb.internal.operation
 
+import com.mongodb.MongoBulkWriteException
 import com.mongodb.MongoNamespace
 import com.mongodb.ReadConcern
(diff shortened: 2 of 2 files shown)


driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
```diff
@@ -282,21 +281,11 @@ private BulkWriteResult getBulkWriteResult(final BsonDocument result) {
     }
 
     private List<BulkWriteInsert> getInsertedItems(final BsonDocument result) {
-        if (payload.getPayloadType() == SplittablePayload.Type.INSERT) {
```
jyemin, line 285: I removed the conditional just to simplify the code.  If the payload type is anything but INSERT, then `payload.getInsertedIds` will be empty anyway.

driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
```diff
@@ -282,21 +281,11 @@ private BulkWriteResult getBulkWriteResult(final BsonDocument result) {
...
-
-            Stream<WriteRequestWithIndex> writeRequests = payload.getWriteRequestWithIndexes().stream();
-            List<Integer> writeErrors = getWriteErrors(result).stream().map(BulkWriteError::getIndex).collect(Collectors.toList());
```
jyemin, line 288: Changed this to a Set to make the `contains` check more efficient.

driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
```diff
@@ -282,21 +281,11 @@ private BulkWriteResult getBulkWriteResult(final BsonDocument result) {
...
-            if (!writeErrors.isEmpty()) {
```
jyemin, line 289: Removed this conditional, as it seems an unnecessary optimization.  The contains check will be fast enough on any empty set.

driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
```diff
@@ -282,21 +281,11 @@ private BulkWriteResult getBulkWriteResult(final BsonDocument result) {
...
-            if (payload.getPosition() < payload.size()) {
-                writeRequests = writeRequests.filter(wr -> wr.getIndex() < payload.getPosition());
-            }
-            return writeRequests
-                    .map(wr -> new BulkWriteInsert(wr.getIndex(), payload.getInsertedIds().get(wr.getIndex())))
-                    .collect(Collectors.toList());
-        }
-        return Collections.emptyList();
+        Set<Integer> writeErrors = getWriteErrors(result).stream().map(BulkWriteError::getIndex).collect(Collectors.toSet());
+        return payload.getInsertedIds().entrySet().stream()
```
jyemin, line 285: It's safe to use SplittablePayload`s insertedIds map directly, since it will only contain the ids from the previously inserted batch.

driver-core/src/test/unit/com/mongodb/internal/operation/BulkWriteBatchSpecification.groovy
```diff
@@ -228,23 +229,71 @@ class BulkWriteBatchSpecification extends Specification {
         !bulkWriteBatch.hasAnotherBatch()
     }
 
-    def 'should only map inserts up to the payload position'() {
+    def 'should map all inserted ids'() {
```
jyemin, line 232: I might have made this test more complicated that it needs to be, as I changed it first just to reproduce the bug, but before I settled on the simplified approach in BulkWriteBatch

driver-core/src/test/unit/com/mongodb/internal/operation/BulkWriteBatchSpecification.groovy
```diff
@@ -228,23 +229,71 @@ class BulkWriteBatchSpecification extends Specification {
...
         bulkWriteBatch.addResult(BsonDocument.parse('{"n": 1, "ok": 1.0}'))
 
         then:
-        bulkWriteBatch.getResult().inserts == [new BulkWriteInsert(0, null), new BulkWriteInsert(1, null)]
+        bulkWriteBatch.getResult().inserts == [new BulkWriteInsert(0, new BsonInt32(0)),
+                                               new BulkWriteInsert(1, new BsonInt32(1)),
+                                               new BulkWriteInsert(2, new BsonInt32(2))]
+    }
+
+    def 'should not map inserted id with a write error'() {
```
jyemin, line 273: There wasn't a test for this so added one.

driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
rozza: So much neater and easier to read.

driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
rozza: Ack

driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
rozza: Ack

driver-core/src/test/unit/com/mongodb/internal/operation/BulkWriteBatchSpecification.groovy
rozza: Good stuff

rozza (approved):
Much easier to read now.

I think you mentioned checking upsertedIds as well - does anything need happen with them?

jyemin:
> I think you mentioned checking upsertedIds as well - does anything need happen with them?

No, I was wrong about that, as upsertedIds come directly from the server's response and wouldn't suffer from the same mapping issue.

PR #853, opened 2022-01-11
Only override Context.PROVIDER_URL if the default isn't supported

jyemin:
JAVA-4423

### DIFF
--- a/driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
+++ b/driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.internal.dns;
+
+import com.mongodb.MongoClientException;
+import com.mongodb.MongoConfigurationException;
+import org.junit.jupiter.api.Test;
+
+import javax.naming.Context;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class DefaultDnsResolverTest {
+    private static final String TEST_HOST = "test1.test.build.10gen.cc";
+
+    @Test
+    public void nonDnsProviderUrlShouldBeIgnored() {
+        String currentValue = System.getProperty(Context.PROVIDER_URL);
+        try {
+            System.setProperty(Context.PROVIDER_URL, "file:///tmp/provider.txt");
+            new DefaultDnsResolver().resolveHostFromSrvRecords(TEST_HOST, "mongodb");
+        } catch (MongoClientException e) {
+            fail("Resolution should succeed");
+        }
+        finally {
+            if (currentValue != null) {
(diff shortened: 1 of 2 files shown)


driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Test
+    public void dnsProviderUrlShouldNotBeIgnored() {
+        String currentValue = System.getProperty(Context.PROVIDER_URL);
+        try {
+            System.setProperty(Context.PROVIDER_URL, "dns:///mongodb.unknown.server.com");
+            new DefaultDnsResolver().resolveHostFromSrvRecords(TEST_HOST, "mongodb");
+            fail("Resolution should fail");
+        } catch (MongoConfigurationException e) {
+            // expected
+        }
```
stIncMale, line 55: This test uses JUnit5, so [assertThrows](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertThrows(java.lang.Class,org.junit.jupiter.api.function.Executable)) can be used, and it is both clearer and more robust than a `try`-`catch` with a comment and a `fail` at the end of `try`.

driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Test
+    public void nonDnsProviderUrlShouldBeIgnored() {
+        String currentValue = System.getProperty(Context.PROVIDER_URL);
+        try {
+            System.setProperty(Context.PROVIDER_URL, "file:///tmp/provider.txt");
+            new DefaultDnsResolver().resolveHostFromSrvRecords(TEST_HOST, "mongodb");
+        } catch (MongoClientException e) {
+            fail("Resolution should succeed");
+        }
```
stIncMale, line 38: I suggest passing `e` to `fail`, there is a variant of this method that allows this: https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#fail(java.lang.String,java.lang.Throwable)

stIncMale (changes requested):
Just two suggestions regarding assertions.

driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
jyemin: Done.  Though I wasn't sure if would be better style to just remove the catch clause, since the test will naturally fail if an exception is thrown.  What do you think?  I just find test that have no assertions to look weird, which is why I included it.

driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
jyemin: Ah, much better.  Done

driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
stIncMale: > I just find test that have no assertions to look weird

It's clear that such a test is not supposed to throw, but it may be unclear what part of the test is especially expected not to throw. Fortunately, JUnit5 has [`assertDoesNotThrow`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable)) (I have just found it). I am approving the PR anyway, but if you decide to use the method, it'd be nice.

stIncMale (approved):

driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
stIncMale: Cool, thanks for using `@AfterEach`! I had a thought about proposing it, but given the PR size and the two other comments, decided to omit to avoid being perceived as too picky)

stIncMale (approved):

PR #856, opened 2022-01-21
Add codec provider for enums

jyemin:
JAVA-4440

One question I have is whether to keep `org.bson.codecs.pojo.EnumPropertyCodecProvider` around, or just remove it.  I'm comfortable with leaving it, but wonder what others think.

### DIFF
--- a/bson/src/main/org/bson/conversions/Bson.java
+++ b/bson/src/main/org/bson/conversions/Bson.java
@@ -20,6 +20,7 @@ import org.bson.BsonDocument;
 import org.bson.codecs.BsonCodecProvider;
 import org.bson.codecs.BsonValueCodecProvider;
 import org.bson.codecs.DocumentCodecProvider;
+import org.bson.codecs.EnumCodecProvider;
 import org.bson.codecs.IterableCodecProvider;
 import org.bson.codecs.JsonObjectCodecProvider;
 import org.bson.codecs.MapCodecProvider;
@@ -47,6 +48,7 @@ public interface Bson {
      *     <li>{@link Jsr310CodecProvider}</li>
      *     <li>{@link JsonObjectCodecProvider}</li>
      *     <li>{@link BsonCodecProvider}</li>
+     *     <li>{@link EnumCodecProvider}</li>
      * </ul>
      * <p>
      * Additional providers may be added in a future release.
@@ -63,7 +65,8 @@ public interface Bson {
                     new MapCodecProvider(),
                     new Jsr310CodecProvider(),
                     new JsonObjectCodecProvider(),
-                    new BsonCodecProvider()));
+                    new BsonCodecProvider(),
+                    new EnumCodecProvider()));
 
     /**
      * Render into a BsonDocument.
--- a/bson/src/test/unit/org/bson/codecs/EnumCodecProviderTest.java
+++ b/bson/src/test/unit/org/bson/codecs/EnumCodecProviderTest.java
@@ -17,7 +17,6 @@
 package org.bson.codecs;
 
 import org.bson.codecs.configuration.CodecRegistries;
-import org.bson.codecs.pojo.entities.SimpleEnum;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
--- a/bson/src/test/unit/org/bson/codecs/EnumCodecProviderTest.java
+++ b/bson/src/test/unit/org/bson/codecs/EnumCodecProviderTest.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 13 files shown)


bson/src/main/org/bson/conversions/Bson.java
```diff
@@ -63,7 +65,8 @@
                     new MapCodecProvider(),
                     new Jsr310CodecProvider(),
                     new JsonObjectCodecProvider(),
-                    new BsonCodecProvider()));
+                    new BsonCodecProvider(),
+                    new EnumCodecProvider()));
```
jyemin, line 69: As far as I can see there are no regression tests for the default registries, so I didn't add new tests.

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -71,7 +72,8 @@
                     new GridFSFileCodecProvider(),
                     new Jsr310CodecProvider(),
                     new JsonObjectCodecProvider(),
-                    new BsonCodecProvider()));
+                    new BsonCodecProvider(),
```
rozza, line 75: I kind of feel we should be using the `org.bson.conversions.DEFAULT_CODEC_REGISTRY` as the base and extend it for the local overrides settings.
eg:

```
fromRegistries(
   fromProviders(asList(
       new DBRefCodecProvider(),
       new DBObjectCodecProvider(),
       new DocumentCodecProvider(new DocumentToDBRefTransformer()),
       new IterableCodecProvider(new DocumentToDBRefTransformer()),
       new MapCodecProvider(new DocumentToDBRefTransformer()),
       new GeoJsonCodecProvider(),
       new GridFSFileCodecProvider()
   )),
   org.bson.conversions.DEFAULT_CODEC_REGISTRY
);
```

But determining if that could cause an issue has made my head hurt.

rozza (changes requested):
I think we can remove the `org.bson.codecs.pojo.EnumPropertyCodecProvider` and `org.bson.codecs.pojo.EnumCodec` in favour of the new ones.  

`PropertyCodecRegistryImpl` should just be updated to point to the new provider. This is belt and braces but its just incase a user has a custom codec registry that won't include this change.

bson/src/test/unit/org/bson/codecs/EnumCodecProviderTest.java
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bson.codecs;
+
+import org.bson.codecs.configuration.CodecRegistries;
+import org.bson.codecs.pojo.entities.SimpleEnum;
```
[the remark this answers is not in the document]
jyemin: It's not a new class from this PR, so I didn't think to move it.  What package do you think it should be moved to?  `org.bson.codecs.entities`, perhaps?  This would be a new package, if so.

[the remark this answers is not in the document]
rozza: Yeah, or just `org.bson.codecs` as there is only going to one enum tested.

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
```diff
@@ -352,11 +352,19 @@ public void testUseGettersForSettersConventionNotEmptyMap() {
 
     @Test
     public void testEnumSupportWithCustomCodec() {
-        CodecRegistry registry = fromRegistries(getCodecRegistry(getPojoCodecProviderBuilder(SimpleEnumModel.class)),
-                fromCodecs(new SimpleEnumCodec()));
+        CodecRegistry registry = fromRegistries(fromCodecs(new SimpleEnumCodec()),
```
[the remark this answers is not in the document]
rozza: So this is the only possible subtle issue users will run into, but they will also have to store the none value of the enum.  So I'm pretty sure that will be minimal.

driver-core/src/main/com/mongodb/MongoClientSettings.java
rozza: Ack - will you open a ticket?

jyemin: Now I see why we didn't do this.  The MongoClientSettings one has a bunch of somewhat subtle alterations because of its support for DBRef.  So I don't see how it could be done straightforwardly.

bson/src/test/unit/org/bson/codecs/EnumCodecProviderTest.java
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/MongoClientSettings.java
rozza: I think that is fine, as the document codec provider with the changes are added after the bson defaults. It does mean there are two DocumentCodecProviders registered - which is a pain.

rozza: I think we can resolve this as won't fix :)

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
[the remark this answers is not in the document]
rozza: Resolving as this should be documented as part of the change.

rozza (changes requested):
Just the previous suggestion:

I think we can remove the org.bson.codecs.pojo.EnumPropertyCodecProvider and org.bson.codecs.pojo.EnumCodec in favour of the new ones.

PropertyCodecRegistryImpl should just be updated to point to the new provider. This is belt and braces but its just incase a user has a custom codec registry that won't include this change.

rozza (approved):
LGTM!

PR #858, opened 2022-01-25
Fix the `await` condition in `OpenConcurrencyLimiter.acquirePermitOrGetAvailableOpenedConnection`

stIncMale:
JAVA-4451


driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -992,12 +992,10 @@ private PooledConnection acquirePermitOrGetAvailableOpenedConnection(final boole
...
-                while (permits == 0) {
-                    stateAndGeneration.throwIfClosedOrPaused();
-                    availableConnection = tryGetAvailable ? tryGetAvailableConnection() : null;
-                    if (availableConnection != null) {
-                        break;
-                    }
+                while (permits == 0
+                        // the absence of short-circuiting is of importance
+                        & !stateAndGeneration.throwIfClosedOrPaused()
+                        & (availableConnection = tryGetAvailable ? tryGetAvailableConnection() : null) == null) {
```
stIncMale, line 998: I know we don't usually use assignment expressions as expressions, but this is the situation when making an exception is worth it. The bug mentioned in the description of the PR was caused by how the `while` condition was specified. It's a condition that the thread waits for, so it has to be checked before calling `awaitNanos`, and after returning from this method. Having the full condition specified in the `while (...)` part not only achieves this, but also achieves this without code duplication.

Originally, the condition was split: part was inside `while (...)`, and part was inside the `while` body. The second part should have been duplicated after the `while` loop in order to make sure that the whole condition is checked again after returning from the last `awaitNanos` call, but it wasn't there.

I could have fixed the code by duplicating the second part of the condition
```java
stateAndGeneration.throwIfClosedOrPaused();
availableConnection = tryGetAvailable ? tryGetAvailableConnection() : null;
```
after the `while` loop, but it only makes the code more complex and leaves opportunities for future mistakes if the condition is ever changed. I think that putting the whole condition inside `while (...)` is very much the right thing to do, even if it looks unconventional compared to the rest of the driver code.

driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
```diff
@@ -384,8 +384,9 @@ boolean acquirePermitFair(final long timeout, final TimeUnit unit) throws MongoI
                 throw new MongoInterruptedException(null, e);
             }
             try {
-                while (permits == 0) {
-                    throwIfClosedOrPaused();
+                while (permits == 0
+                        // the absence of short-circuiting is of importance
+                        & !throwIfClosedOrPaused()) {
```
jyemin, line 389: This isn't fixing a bug, right? If not, consider holding off on it until the next PR, where we're modifying this loop anyway.

jyemin (changes requested):

driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
stIncMale: This change just applies the same style as in `DefaultConnectionPool`. This change has at least some relevance to the  current PR, but will be completely irrelevant und weird in the next PR. Why do you think it is better to do it there?

stIncMale: We discussed this over Zoom, and decided that the style change in `ConcurrentPool` will be committed separately.

stIncMale: I force-pushed changes, it seemed appropriate here. Done.

jyemin (approved):
LGTM

PR #860, opened 2022-01-25
Ensure server discovery and monitoring events are published in order

jyemin:
* Ensure events are ordered according to how they are applied to driver state
* Publish all events to application listeners one at a time, on a separate thread

JAVA-4449

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
@@ -26,6 +26,7 @@ import com.mongodb.connection.ClusterType;
 import com.mongodb.connection.ServerDescription;
 import com.mongodb.diagnostics.logging.Logger;
 import com.mongodb.diagnostics.logging.Loggers;
+import com.mongodb.event.ClusterListener;
 import com.mongodb.event.ServerDescriptionChangedEvent;
 import org.bson.types.ObjectId;
 
@@ -68,8 +69,9 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
         }
     }
 
-    AbstractMultiServerCluster(final ClusterId clusterId, final ClusterSettings settings, final ClusterableServerFactory serverFactory) {
-        super(clusterId, settings, serverFactory);
+    AbstractMultiServerCluster(final ClusterId clusterId, final ClusterSettings settings, final ClusterListener clusterListener,
+            final ClusterableServerFactory serverFactory) {
+        super(clusterId, settings, clusterListener, serverFactory);
         isTrue("connection mode is multiple", settings.getMode() == ClusterConnectionMode.MULTIPLE);
         clusterType = settings.getRequiredClusterType();
         replicaSetName = settings.getRequiredReplicaSetName();
@@ -98,8 +100,8 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
                 addServer(serverAddress);
             }
             newDescription = updateDescription();
+            fireChangeEvent(newDescription, currentDescription);
         }
-        fireChangeEvent(newDescription, currentDescription);
     }
 
     @Override
@@ -221,9 +223,9 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
                 oldClusterDescription = getCurrentDescription();
                 newClusterDescription = updateDescription();
             }
-        }
-        if (shouldUpdateDescription) {
-            fireChangeEvent(newClusterDescription, oldClusterDescription);
+            if (shouldUpdateDescription) {
+                fireChangeEvent(newClusterDescription, oldClusterDescription);
+            }
         }
     }
 
--- a/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
@@ -26,7 +26,6 @@ import com.mongodb.connection.ClusterType;
 import com.mongodb.connection.ServerDescription;
 import com.mongodb.diagnostics.logging.Logger;
 import com.mongodb.diagnostics.logging.Loggers;
-import com.mongodb.event.ClusterListener;
 import com.mongodb.event.ServerDescriptionChangedEvent;
 import org.bson.types.ObjectId;
 
@@ -69,9 +68,8 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
         }
     }
 
-    AbstractMultiServerCluster(final ClusterId clusterId, final ClusterSettings settings, final ClusterListener clusterListener,
-            final ClusterableServerFactory serverFactory) {
-        super(clusterId, settings, clusterListener, serverFactory);
+    AbstractMultiServerCluster(final ClusterId clusterId, final ClusterSettings settings, final ClusterableServerFactory serverFactory) {
+        super(clusterId, settings, serverFactory);
         isTrue("connection mode is multiple", settings.getMode() == ClusterConnectionMode.MULTIPLE);
         clusterType = settings.getRequiredClusterType();
         replicaSetName = settings.getRequiredReplicaSetName();
--- a/driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
@@ -162,20 +162,16 @@ final class AsynchronousClusterEventListener implements ClusterListener, ServerL
     }
 
     private void publishEvents() {
-        try {
-            while (true) {
+        while (true) {
+            try {
                 Supplier<Boolean> eventPublisher = eventPublishers.take();
-                try {
-                    boolean isLastEvent = eventPublisher.get();
-                    if (isLastEvent) {
-                        break;
-                    }
-                } catch (Exception e) {
-                    // ignore exceptions thrown from listeners
+                boolean isLastEvent = eventPublisher.get();
+                if (isLastEvent) {
+                    break;
                 }
+            } catch (RuntimeException | InterruptedException e) {
+                // ignore exceptions thrown from listeners, also ignore interrupts that user code may cause
             }
-        } catch (InterruptedException e) {
-            // ignore
         }
     }
 }
--- a/driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.internal.connection;
+
+import com.mongodb.connection.ClusterId;
+import com.mongodb.event.ClusterClosedEvent;
+import com.mongodb.event.ClusterDescriptionChangedEvent;
+import com.mongodb.event.ClusterListener;
+import com.mongodb.event.ClusterOpeningEvent;
+import com.mongodb.event.ServerClosedEvent;
+import com.mongodb.event.ServerDescriptionChangedEvent;
+import com.mongodb.event.ServerHeartbeatFailedEvent;
+import com.mongodb.event.ServerHeartbeatStartedEvent;
+import com.mongodb.event.ServerHeartbeatSucceededEvent;
+import com.mongodb.event.ServerListener;
+import com.mongodb.event.ServerMonitorListener;
+import com.mongodb.event.ServerOpeningEvent;
+import com.mongodb.internal.VisibleForTesting;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.function.Supplier;
+
+import static com.mongodb.assertions.Assertions.notNull;
+import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
+
+/**
+ * An implementation of a listener for all cluster-related events.  Its purpose is the following:
+ *
+ * 1. To ensure that cluster-related events are delivered one at a time, with happens-before semantics
+ * 2. To ensure that application-provided event listener methods do not execute within critical sections of the driver
+ *
+ * This is done by adding all events to an unbounded blocking queue, and then publishing them from a dedicated thread by taking
+ * them off the queue one at a time.
+ *
+ * There is an assumption that the last event that should be published is the {@link ClusterClosedEvent}.  Once that event is published,
+ * the publishing thread is allowed to die.
+ */
+final class AsynchronousClusterEventListener implements ClusterListener, ServerListener, ServerMonitorListener {
+    private final BlockingQueue<Supplier<Boolean>> eventPublishers = new LinkedBlockingQueue<>();
+    private final ClusterListener clusterListener;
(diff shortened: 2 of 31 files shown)


driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -221,9 +223,9 @@ private void onChange(final ServerDescriptionChangedEvent event) {
                 oldClusterDescription = getCurrentDescription();
                 newClusterDescription = updateDescription();
             }
-        }
-        if (shouldUpdateDescription) {
-            fireChangeEvent(newClusterDescription, oldClusterDescription);
+            if (shouldUpdateDescription) {
```
jyemin, line 226: Move into synchronized block to ensure events are published in the order they are applied.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -81,11 +79,12 @@
     private volatile boolean isClosed;
     private volatile ClusterDescription description;
 
-    BaseCluster(final ClusterId clusterId, final ClusterSettings settings, final ClusterableServerFactory serverFactory) {
+    BaseCluster(final ClusterId clusterId, final ClusterSettings settings, final ClusterListener clusterListener,
```
jyemin, line 82: The bulk of the changes are due to the fact that we can no longer get the listener directly from the settings.  It has to be wrapped at a higher level.

driver-core/src/main/com/mongodb/internal/connection/DefaultClusterFactory.java
```diff
@@ -70,26 +74,33 @@ public Cluster createCluster(final ClusterSettings clusterSettings, final Server
 
         DnsSrvRecordMonitorFactory dnsSrvRecordMonitorFactory = new DefaultDnsSrvRecordMonitorFactory(clusterId, serverSettings);
 
+        AsynchronousClusterEventListener clusterEventListener =
```
jyemin, line 77: This is where the rubber meets the road: we wrap the listeners from cluster and server settings into a single listener.

driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
```diff
@@ -83,28 +84,30 @@ public void close() {
...
-                                .exception(new MongoConfigurationException(
-                                        format("Replica set name '%s' does not match required replica set name of '%s'",
-                                                newDescription.getSetName(), getSettings().getRequiredReplicaSetName())))
-                                .type(ServerType.UNKNOWN)
-                                .setName(null)
-                                .ok(false)
-                                .build();
-                        publishDescription(ClusterType.UNKNOWN, newDescription);
-                        return;
+            synchronized (this) {
```
jyemin, line 87: Just whitespace changes except for the addition of the synchronized block

driver-core/src/main/com/mongodb/internal/connection/ClusterableServerFactory.java
```diff
@@ -18,11 +18,10 @@
 
 import com.mongodb.ServerAddress;
 import com.mongodb.connection.ServerSettings;
-import com.mongodb.event.ServerListener;
 
 public interface ClusterableServerFactory {
     ClusterableServer create(ServerAddress serverAddress, ServerDescriptionChangedListener serverDescriptionChangedListener,
-                             ServerListener serverListener, ClusterClock clusterClock);
```
jyemin, line 25: There is no need to pass the server listener here, and doing so made it more difficult to wrap the listener.  So removed the listener from this interface and made it the responsibility of implementers of this interface to have the appropriate (wrapped) listener available.

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
```diff
@@ -125,10 +125,10 @@ private void updateDescription(final ServerDescription newDescription) {
         description = newDescription;
         ServerDescriptionChangedEvent serverDescriptionChangedEvent = new ServerDescriptionChangedEvent(
                 serverId, newDescription, previousDescription);
-        serverDescriptionChangedListener.serverDescriptionChanged(serverDescriptionChangedEvent);
         if (!wouldDescriptionsGenerateEquivalentEvents(newDescription, previousDescription)) {
             serverListener.serverDescriptionChanged(serverDescriptionChangedEvent);
         }
+        serverDescriptionChangedListener.serverDescriptionChanged(serverDescriptionChangedEvent);
```
jyemin, line 131: It makes more sense to publish the ServerDescriptionChangedEvent before the ClusterDescriptionChangedEvent (which this line will indirectly cause), so moving this down.

driver-core/src/test/unit/com/mongodb/internal/connection/AsynchronousClusterEventListenerTest.java
```diff
@@ -0,0 +1,171 @@
...
+        ClusterClosedEvent clusterClosedEvent = new ClusterClosedEvent(clusterId);
+        listener.clusterClosed(clusterClosedEvent);
+        assertEquals(clusterClosedEvent, targetListener.take());
+
+        // The thread should die after publishing the ClusterClosedEvent
+        listener.getPublishingThread().join(5000);
+    }
+
+    private static final class AllClusterEventListener implements ClusterListener, ServerListener, ServerMonitorListener {
+        private final BlockingQueue<Object> lastEvent = new ArrayBlockingQueue<>(1);
```
stIncMale, line 111: If we use `SynchronousQueue` here instead of `ArrayBlockingQueue` (it's also a `BlockingQueue`), then `testEventsPublished` will also verify that `AsynchronousClusterEventListener` delivers events asynchronously: if it were to deliver synchronously, then the test thread would have hanged on sending events via `AsynchronousClusterEventListener`.

driver-core/src/test/unit/com/mongodb/internal/connection/AsynchronousClusterEventListenerTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+        Object take() throws InterruptedException {
+            return lastEvent.poll(5, TimeUnit.SECONDS);
+        }
+
+        void addEvent(final Object event) {
+            try {
+                lastEvent.put(event);
+            } catch (InterruptedException e) {
+                // ignore
```
stIncMale, line 121: [optional]
We have not started the test thread, and if it is interrupted, that's likely because the test runner did that. Therefore, we should not erase the information about the thread being interrupted. Preserving this information may be done either by throwing `InterruptedException`, or by setting the interrupted flag back on via `Thread.currentThread().interrupt()`.

driver-core/src/main/com/mongodb/internal/connection/DefaultClusterFactory.java
```diff
@@ -68,6 +73,18 @@ public Cluster createCluster(final ClusterSettings clusterSettings, final Server
 
         ClusterId clusterId = new ClusterId();
 
+        AsynchronousClusterEventListener clusterEventListener =
+                new AsynchronousClusterEventListener(clusterId, getClusterListener(originalClusterSettings),
+                        getServerListener(originalServerSettings), getServerMonitorListener(originalServerSettings));
+
+        ClusterSettings clusterSettings = ClusterSettings.builder(originalClusterSettings)
+                .clusterListenerList(singletonList(clusterEventListener))
```
jyemin, line 81: These new methods are already proving valuable!

jyemin (commented):
Simplified based on in-person conversation.  PTAL.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/connection/DefaultClusterFactory.java
stIncMale: Nice! Pinging @ramasai1 so that he sees the fruits of labor :)

stIncMale:
I like the new approach for decorating listeners 👍 

[optional]
I propose to assert that there is exactly 1 listener here and then just use it. We know that it must be the case for all clusters and servers because of how `DefaultClusterFactory` works.

If you agree, the same can be done also in

- LoadBalancedCluster
- DefaultClusterableServerFactory
- LoadBalancedClusterableServerFactory
- DefaultServerMonitor

As a result, the only place where `EventListenerHelper.getClusterListener`/`getServerListener`/`getServerMonitorListener` are called will be `DefaultClusterFactory`. Additionally, we could even move the methods there and make them private to avoid using them accidentally in a new server/cluster/factory implementation.

stIncMale:
I like the new approach for decorating listeners 👍 

[optional]
I propose to assert that there is exactly 1 listener in the constructor of `BaseCluster` and then just use it, instead of calling `getClusterListener(settings)`. We know that it must be the case for all clusters and servers because of how `DefaultClusterFactory` works.

If you agree, the same can be done also in

- LoadBalancedCluster
- DefaultClusterableServerFactory
- LoadBalancedClusterableServerFactory
- DefaultServerMonitor

As a result, the only place where `EventListenerHelper.getClusterListener`/`getServerListener`/`getServerMonitorListener` are called will be `DefaultClusterFactory`. Additionally, we could even move the methods there and make them private to avoid using them accidentally in a new server/cluster/factory implementation.

driver-core/src/test/unit/com/mongodb/internal/connection/AsynchronousClusterEventListenerTest.java
stIncMale: > This method is called by the listener methods

Ah, sorry :(, it's indeed never called by the thread that runs the test method. In that case, `Assertions.fail()` would have also been an appropriate way to handle the exception.

driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
```diff
@@ -0,0 +1,172 @@
...
+                    boolean isLastEvent = eventPublisher.get();
+                    if (isLastEvent) {
+                        break;
+                    }
+                } catch (Exception e) {
+                    // ignore exceptions thrown from listeners
+                }
+            }
+        } catch (InterruptedException e) {
+            // ignore
```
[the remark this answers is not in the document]
stIncMale: > any InterruptedException thrown by the listener will be ignored by the inner catch
`eventPublisher.get()` cannot throw `InterruptedException`, unless, of course, the ["sneaky throw"](https://www.baeldung.com/java-sneaky-throws) approach is used, but there is no point in writing software that tries to work around this exploit.

> The outer catch of InterruptedException should never be hit because publisherThread is private
> an application could enumerate all threads and find this one, and interrupt it
`publishingThread` runs user code, so all it takes is to call `Thread.currentThread().interrupt()` in a user-supplied event handler, and `publishingThread` exits the `while` loop and terminates. The code proposed above handles `InterruptedException` thrown by `eventPublishers.take()` as a result of user code interrupting `publishingThread`.

driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
```diff
@@ -0,0 +1,172 @@
...
+ * 1. To ensure that cluster-related events are delivered one at a time, with happens-before semantics
+ * 2. To ensure that application-provided event listener methods do not execute within critical sections of the driver
+ *
+ * This is done by adding all events to an unbounded blocking queue, and then publishing them from a dedicated thread by taking
+ * them off the queue one at a time.
+ *
+ * There is an assumption that the last event that should be published is the {@link ClusterClosedEvent}.  Once that event is published,
+ * the publishing thread is allowed to die.
+ */
+final class AsynchronousClusterEventListener implements ClusterListener, ServerListener, ServerMonitorListener {
```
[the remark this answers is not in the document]
stIncMale: Sure, we can write javadocs explaining that implementations are not required to be thread-safe.

What do you think about the rest of the proposal? It seems wasteful to implement these guarantees, but not explain them to users, which means that users won't be able to rely on them.

driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
```diff
@@ -0,0 +1,172 @@
...
+            eventPublisher.apply(serverMonitorListener);
+            return false;
+        });
+    }
+
+    private void addEvent(final Supplier<Boolean> supplier) {
+        // protect against rogue publishers
+        if (!publishingThread.isAlive()) {
+            return;
+        }
```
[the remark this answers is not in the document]
stIncMale: > server monitor events can be published after cluster is closed because DefaultServerMonitor#close only interrupts the monitorThread, but does not join it

Thanks! That means that the calling code is racy and we are fine with ignoring the events that arrive after the thread is terminated.

The check `if (!publishingThread.isAlive())` is still there, what purpose does it serve given the observations from my previous comment?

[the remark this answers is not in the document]
jyemin: I guess it's a little bit defensive programming, which I generally avoid.  The idea is that if for whatever reason the monitor thread doesn't stop producing events even after the cluster closes, we won't keep adding to the queue with nothing to empty it.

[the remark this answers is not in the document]
jyemin: Given that, do you want me to remove it?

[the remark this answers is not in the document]
stIncMale: > if for whatever reason the monitor thread doesn't stop producing events even after the cluster closes

I see. I would not think about doing this, but since defending against that is trivial and the code has already been written, let's leave it as is.

driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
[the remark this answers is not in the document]
jyemin: Ok, took me a while to understand, but I get it now.

Done.

driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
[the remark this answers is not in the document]
jyemin: Yeah, totally agree we should document.  Can you take first crack at coming up with the right wording for it?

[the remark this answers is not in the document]
stIncMale: Maybe something like this: https://github.com/stIncMale/mongo-java-driver/commit/7e18a2f453067154a71448126b75fe0ddecb443b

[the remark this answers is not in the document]
jyemin: Thanks.  Done.

driver-core/src/main/com/mongodb/internal/connection/DefaultClusterFactory.java
```diff
@@ -67,6 +78,30 @@ public Cluster createCluster(final ClusterSettings clusterSettings, final Server
                                  final List<MongoCompressor> compressorList, final @Nullable ServerApi serverApi) {
 
         ClusterId clusterId = new ClusterId();
+        ClusterSettings clusterSettings;
+        ServerSettings serverSettings;
+
+        if (noClusterEventListeners(originalClusterSettings, originalServerSettings)) {
```
jyemin, line 84: Since the vast number of clients never register a listener, avoiding thread/queue creation for that case.

driver-core/src/main/com/mongodb/internal/connection/DefaultClusterFactory.java
```diff
@@ -96,4 +131,28 @@ public Cluster createCluster(final ClusterSettings clusterSettings, final Server
...
         }
     }
+
+    private boolean noClusterEventListeners(final ClusterSettings clusterSettings, final ServerSettings serverSettings) {
+        return clusterSettings.getClusterListeners().isEmpty()
+                && serverSettings.getServerListeners().isEmpty()
+                && serverSettings.getServerMonitorListeners().isEmpty();
+    }
+
+    private static ClusterListener getClusterListener(final ClusterSettings clusterSettings) {
```
jyemin, line 141: Small behavioral change here.  Using multicaster when there is one listener registered, because the muilticaster will log exceptions thrown by the listeners, which should happen in all cases.  Didn't seem worth it to add a non-multicasting exception-logging implementation.

stIncMale (approved):

PR #862, opened 2022-01-26
Alternate fair/unfair lock modes in the connection pool to improve throughput

stIncMale:
Another tiny change is moving `pool.release(openConnection)` out of the guarded section
in `DefaultConnectionPool.OpenConcurrencyLimiter.tryHandOverOrRelease`:
it does not have to be guarded, and when not only fair locking is used,
this change has a significant positive measurable effect
on the connection cycle (check out/check in) throughput.

JAVA-4452

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
@@ -28,8 +28,10 @@ import com.mongodb.lang.Nullable;
 
 import java.util.Iterator;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Condition;
-import java.util.concurrent.locks.ReadWriteLock;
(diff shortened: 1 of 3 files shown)


driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
```diff
@@ -501,4 +537,45 @@ boolean closed() {
...
+            lock.lock();
+        }
+    }
+
+    private static void lockUnfair(final ReentrantReadWriteLock.WriteLock lock) {
+        // `WriteLock.tryLock` is unfair
+        if (!lock.tryLock()) {
+            lock.lock();
+        }
+    }
```
stIncMale, line 574: See the discussion [here](https://github.com/mongodb/mongo-java-driver/pull/857#discussion_r791658469) about the reason behind introducing two methods as opposed to having one that accepts `Lock`.

jyemin (approved):
LGTM!

rozza (approved):
LGTM

PR #866, opened 2022-02-02
Signal all threads (monitor and RTT) from DefaultServerMonitor.connect

stIncMale:
JAVA-4467

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
@@ -112,7 +112,7 @@ class DefaultServerMonitor implements ServerMonitor {
     public void connect() {
         lock.lock();
         try {
-            condition.signal();
+            condition.signalAll();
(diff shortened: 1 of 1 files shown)


driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -112,7 +112,7 @@ public void start() {
     public void connect() {
         lock.lock();
         try {
-            condition.signal();
+            condition.signalAll();
```
jyemin, line 115: From what I can see, the RTT thread doesn't await this condition.  It calls `DefaultServerMonitor#waitForNext`, which just sleeps.  It's confusing because there are two waitForNext methods in the same file (also `DefaultServerMonitor.ServerMonitorRunnable#waitForNext`).

stIncMale: Indeed, good catch!

PR #867, opened 2022-02-03
Fix minor typo in java comments

haoch:


jyemin:
@haoch thanks for catching the copy and paste error!

PR #869, opened 2022-02-04
Introduce and use `Cluster.withLock` to achieve cluster-wide locking

stIncMale:
See this [comment](https://jira.mongodb.org/browse/JAVA-4471?focusedCommentId=4336655&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4336655) for the details on what this PR fixes. The idea is to guard the action of pausing/unpausing a pool with the same cluster-wide lock that is used to guard the cluster state.

There will be one more PR that tries to get rid of some fields and params that duplicate the information that can be accessed via `Cluster`.

JAVA-4471

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
@@ -29,6 +29,9 @@ import com.mongodb.diagnostics.logging.Logger;
 import com.mongodb.diagnostics.logging.Loggers;
 import com.mongodb.event.ServerDescriptionChangedEvent;
 
+import java.util.concurrent.atomic.AtomicReference;
+
+import static com.mongodb.assertions.Assertions.assertNotNull;
 import static com.mongodb.assertions.Assertions.isTrue;
 import static com.mongodb.connection.ServerConnectionState.CONNECTING;
 import static java.lang.String.format;
@@ -41,7 +44,7 @@ import static java.util.Collections.singletonList;
 public final class SingleServerCluster extends BaseCluster {
     private static final Logger LOGGER = Loggers.getLogger("cluster");
 
-    private final ClusterableServer server;
+    private final AtomicReference<ClusterableServer> server;
 
     public SingleServerCluster(final ClusterId clusterId, final ClusterSettings settings, final ClusterableServerFactory serverFactory) {
         super(clusterId, settings, serverFactory);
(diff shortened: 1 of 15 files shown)


driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
```diff
@@ -52,38 +55,39 @@ public SingleServerCluster(final ClusterId clusterId, final ClusterSettings sett
...
+            assertNotNull(server.get()).close();
             super.close();
         }
     }
 
     private class DefaultServerDescriptionChangedListener implements ServerDescriptionChangedListener {
         @Override
         public void serverDescriptionChanged(final ServerDescriptionChangedEvent event) {
-            synchronized (this) {
+            withLock(() -> {
```
stIncMale, line 90: Note that the monitor used here for synchronization (`DefaultServerDescriptionChangedListener.this`) was different than the one used as a result of changes in this PR (`SingleServerCluster.this`). However, I don't think there was a reason to use a different one. Note that `AbstractMultiServerCluster` used and continues to use the same monitor `AbstractMultiServerCluster.this` both to handle events in its `AbstractMultiServerCluster.DefaultServerDescriptionChangedListener` and in other places.

driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
```diff
@@ -41,7 +44,7 @@
 public final class SingleServerCluster extends BaseCluster {
     private static final Logger LOGGER = Loggers.getLogger("cluster");
 
-    private final ClusterableServer server;
+    private final AtomicReference<ClusterableServer> server;
```
stIncMale, line 47: Introducing this indirection is the only way to do the same thing that was done via the `synchronized` block below. This code weirdness stems from the approach when constructors do mo work than purely constructing objects.

stIncMale:
`tests-load-balancer` failures are caused by https://github.com/mongodb-labs/drivers-evergreen-tools/pull/183 (see also this [comment](https://mongodb.slack.com/archives/C01C3KRQZ9P/p1643165533004100)).

driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
jyemin: Ack

driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
jyemin: I don't understand what the difference is between withLock and synchronized that requires this change.  Can you explain a bit more?

stIncMale: It is impossible to assign a value to `SingleServerCluster.server` from within a lambda expression, but was possible from a `synchronized` block inside the body of a constructor.

jyemin (approved):
LGTM

rozza:
Looks good - one question does adding this lock add any extra contention, which in turn impacts performance?

driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
stIncMale: It also may seem like I could have just made the `server` field non-`final` instead of making it `final AtomicReference<ClusterableServer>`. However it is not the case:

- `AtomicReference` is needed only as a mutable container for `ClusterableServer`. The fact that it stores its field internally as `volatile` has no significance to us.
- `final` makes sure that even if a reference to `SingleServerCluster` is published via a data race, `SingleServerCluster.connect` (and other methods reading the `server` field) are guaranteed to observe this it properly initialized: not only `server` is not `null`, but also `server.get()` is guaranteed to return non-`null`.

stIncMale:
> does adding this lock add any extra contention, which in turn impacts performance?

We removed a per-server lock, and now all servers use a cluster-wide lock. The same lock is used when cluster state (e.g., server membership) is updated. While formally this must mean that more code now will be guarded by the same cluster-wide lock, it is unlikely that this may affect performance (if changes to server and cluster state happen so often that there is a lock contention, then there is a much more serious problem with the MongoDB cluster) characteristics other eliminating deadlocks.

driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
jyemin: Got it.  Thanks.

PR #871, opened 2022-02-07
Refactor internal APIs to avoid requiring duplicate data in method/constructor params

stIncMale:
This is a follow-up for https://github.com/mongodb/mongo-java-driver/pull/869, Since this PR contains only refactoring, there is no need to port it to older versions.

JAVA-4471


driver-core/src/main/com/mongodb/internal/connection/Cluster.java
```diff
@@ -99,4 +95,11 @@
      * @param action The action to {@linkplain Runnable#run() do}.
      */
     void withLock(Runnable action);
+
+    /**
+     * This method allows {@link Server}s to notify the {@link Cluster} about changes in their state as per the
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst">
+     * Server Discovery And Monitoring</a> specification.
+     */
+    void onChange(ServerDescriptionChangedEvent event);
```
stIncMale, line 104: This method replaces `ServerDescriptionChangedListener`.

jyemin (approved):
I've always found that internal listener class to be confusing.  Glad to see it gone.

LGTM

PR #873, opened 2022-02-07
Fix ServerDiscoveryAndMonitoringProseTests.testConnectionPoolManagement and disable it

stIncMale:
The changes in this PR change the test such that it does not rely on non-existing ordering relations. However, we still have to disable the test because sone ordering relations expected by the corresponding SDAM prose scenario are not expressed in the specs and do not currently exist in the driver. See JAVA-4483 for more details.

JAVA-4483

P.S. I run the modified test before disabling it, and it mostly works, unless I introduce an artificial delay that makes it fail as a result of the aforementioned lack of ordering that is checked by the test scenario.


jyemin (approved):
LGTM

PR #874, opened 2022-02-08
Make ClusterConnectionMode required in CommandMessage

jyemin:
JAVA-4447

This is a pre-cursor to the work to actually ensure that OP_MSG is used when the connection mode is load-balanced.  Before we can do that we have to ensure that ClusterConnectionMode is consistently passed all the way down to CommandMessage, which is where that decision is made.

The majority of this work was done using Change Message Signature refactoring in IntelliJ.


driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -106,7 +107,7 @@
         this.exhaustAllowed = exhaustAllowed;
         this.payload = payload;
         this.payloadFieldNameValidator = payloadFieldNameValidator;
-        this.clusterConnectionMode = clusterConnectionMode;
+        this.clusterConnectionMode = notNull("clusterConnectionMode", clusterConnectionMode);
```
jyemin, line 110: No longer optional!

rozza (approved):

PR #876, opened 2022-02-10
Make reference count check atomic with release

jyemin:
JAVA-4490

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
+++ b/driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
@@ -45,8 +45,7 @@ public class ClientSessionBinding extends AbstractReferenceCounted implements As
 
     public  ClientSessionBinding(final AsyncClientSession session, final boolean ownsSession,
                                  final AsyncClusterAwareReadWriteBinding wrapped) {
-        this.wrapped = notNull("wrapped", (wrapped));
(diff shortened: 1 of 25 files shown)


driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
```diff
@@ -131,15 +128,15 @@ public void getReadConnectionSource(final int minWireVersion, final ReadPreferen
     }
 
     @Override
-    public void release() {
-        wrapped.release();
-        closeSessionIfCountIsZero();
-    }
-
-    private void closeSessionIfCountIsZero() {
-        if (getCount() == 0 && ownsSession) {
```
jyemin, line 140: Stop using `getCount` and instead give this class its own reference count so that `release` can properly close the session.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerConnection.java
```diff
@@ -61,16 +61,17 @@ public DefaultServerConnection retain() {
...
+        int count = super.release();
+        if (count == 0) {
             wrapped.close();
         }
+        return count;
     }
 
     @Override
     public ConnectionDescription getDescription() {
-        isTrue("open", getCount() > 0);
```
jyemin, line 73: Just get rid of superfluous use of `getCount`

PR #878, opened 2022-02-15
Ensure callback using cluster description is guarded

rozza:
Getting the description from a cluster can throw a
MongoTimeoutException or a MongoInterruptedException.
So in that case the exception should be passed to
the callback.

JAVA-4432

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
+++ b/driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
@@ -16,6 +16,7 @@
 
 package com.mongodb.internal.async.client;
 
+import com.mongodb.MongoException;
 import com.mongodb.ReadConcern;
 import com.mongodb.ReadPreference;
 import com.mongodb.RequestContext;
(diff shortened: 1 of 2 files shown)


rozza:
Looks like I broke some tests - putting this into draft, will review and double check the root cause.

rozza:
Ok this is now ready for review.

The bug was introduced in a82ce71d and is extremely subtle. The lack of guarding of sync code that can throw (unchecked) 
 errors and the non chainable pattern of SingleResultCallback make this type of bug both too easy to write and hard to diagnose.
 
Adopting react and its functional publishers should mitigate this style of bug. The `Mono` wrapper guards the code used in `onNext` or any of the functional transformations (map, flatmap, etc..)

driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
```diff
@@ -58,19 +59,27 @@ public ReadPreference getReadPreference() {
...
-        }
+        isConnectionSourcePinningRequired((isConnectionSourcePinningRequired, t) -> {
+            if (t != null) {
+                callback.onResult(null, t);
+            } else if (isConnectionSourcePinningRequired) {
+                getPinnedConnectionSource(true, callback);
+            } else {
+                wrapped.getReadConnectionSource(new WrappingCallback(callback));
+            }
+        });
```
jyemin, line 70: I had a hard time parsing this code, maybe because of the method name and variable name being the same, combined with the extra layer of callbacks.  What do you think about just wrapping it like this?

```suggestion

        try {
            if (isConnectionSourcePinningRequired()) {
                getPinnedConnectionSource(true, callback);
            } else {
                wrapped.getReadConnectionSource(new WrappingCallback(callback));
            }
        } catch (Exception e) {
            callback.onResult(null, e);
        }
```

jyemin (commented):
Looks good, just a stylistic change request.

driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
rozza: I tried that and also tried abstracting the pattern  eg:

```
import java.util.function.Consumer;

public final class CallbackHelper {

    public static <T> void guardCallback(
            final Consumer<SingleResultCallback<T>> callbackConsumer,
            final SingleResultCallback<T> callback) {
        try {
            callbackConsumer.accept(callback);
        } catch (Throwable t) {
            callback.onResult(null, t);
        }
    }

    private CallbackHelper() {
    }
}
```

However, I discounted it as I wanted to be explicit that `isConnectionSourcePinningRequired()` can produce an error.  Also I didnt want to wrap the same method call in try catches twice.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/async/client/ClientSessionBinding.java
jyemin: ok, i can live with it.

jyemin (approved):
LGTM

jyemin (approved):

PR #879, opened 2022-02-16
Refactor AwsAuthenticator to extract code that obtains credentials from the environment

jyemin:
JAVA-4501


driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
```diff
@@ -0,0 +1,211 @@
...
+     * Gets the session token.
+     *
+     * @return the sessionToken, which may not be null
+     */
+    @Nullable
+    public String getSessionToken() {
+        return sessionToken;
+    }
+
+    public static AwsInternalCredential obtainFromEnvironment(final AwsInternalCredential startingCredential,
```
jyemin, line 87: I opted for this static factory method.  It's a bit strange because it both takes and returns an instance of itself.  

Happy to discuss alternative designs.

driver-core/src/main/com/mongodb/internal/connection/AwsAuthenticator.java
```diff
@@ -181,11 +155,16 @@ public void dispose() {
...
-            final AuthorizationHeader authorizationHeader = AuthorizationHeader.builder()
-                    .setAccessKeyID(getUserName())
-                    .setSecretKey(getPassword())
-                    .setSessionToken(token)
+            AwsInternalCredential awsInternalCredential = AwsInternalCredential.obtainFromEnvironment(new AwsInternalCredential(
+                                credential.getUserName(), getSecretAccessKey(),
+                                credential.getMechanismProperty(AWS_SESSION_TOKEN_KEY, null)),
+                                credential.getMechanismProperty(AWS_CREDENTIAL_PROVIDER_KEY, null));
+
+            String sessionToken = awsInternalCredential.getSessionToken();
```
jyemin, line 163: To make SpotBugs stop complaining

driver-core/src/main/com/mongodb/internal/connection/AwsAuthenticator.java
```diff
@@ -194,13 +173,17 @@ public void dispose() {
...
-            if (token != null) {
-                ret.append("t", new BsonString(token));
+            if (sessionToken != null) {
+                ret.append("t", new BsonString(sessionToken));
             }
 
             return toBson(ret);
         }
 
+        private String getSecretAccessKey() {
```
jyemin, line 183: SpotBugs...

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
```diff
@@ -0,0 +1,211 @@
...
+            final Supplier<AwsCredential> awsCredentialSupplier) {
+
+        if (awsCredentialSupplier != null) {
+            return obtainFromSupplier(startingCredential, awsCredentialSupplier);
+        }
+
+        BsonDocument ec2OrEcsResponse = null;
+
+        String accessKeyId;
+        if (startingCredential.getAccessKeyId() != null) {
```
markbenvenuto, line 97: A style difference is that I do not build the aws credential in pieces. If a given credential source has the credentials, I get it all or nothing from a given source. Your code allows one to pass `accessKeyId` in `startingCredential` and `secretAcccessKeyId` from the environment variable.

I think of the credential as a single thing, not a thing that can be pieced together from different sources.
_Example pseudo code_
```
if (startingCredential != null) {
  return startCredential
if System.getenv("AWS_ACCESS_KEY_ID") != null {
  .. get other env
  return
}
Do the getEc2OrEcsResponse stuff
...
```
This is how the server code checks credentials - https://github.com/mongodb/mongo/blob/c92b95b22c3290a6235f7952c598c8c5feae669e/src/mongo/client/sasl_aws_client_conversation.cpp#L94-L117

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
```diff
@@ -0,0 +1,211 @@
...
+            if (System.getenv("AWS_SECRET_ACCESS_KEY") != null) {
+                secretAccessKey = System.getenv("AWS_SECRET_ACCESS_KEY");
+            } else {
+                ec2OrEcsResponse = ec2OrEcsResponse == null ? getEc2OrEcsResponse() : ec2OrEcsResponse;
+                secretAccessKey = ec2OrEcsResponse.getString("SecretAccessKey").getValue();
+            }
+        }
+
+        String sessionToken;
+        if (startingCredential.getAccessKeyId() != null) {
```
markbenvenuto, line 120: Do you mean to call `getAccessKeyId()` here? I think it is right but I am surprised.

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
```diff
@@ -0,0 +1,211 @@
...
+            }
+        }
+
+        return new AwsInternalCredential(accessKeyId, secretAccessKey, sessionToken);
+    }
+
+    private static AwsInternalCredential obtainFromSupplier(final AwsInternalCredential startingCredential, final Supplier<AwsCredential> awsCredentialSupplier) {
+        isTrueArgument("accessKeyId is null", startingCredential.getAccessKeyId() == null);
+        isTrueArgument("secret access key is null", startingCredential.getSecretAccessKey() == null);
+        isTrueArgument("session token is null", startingCredential.getSessionToken() == null);
```
markbenvenuto, line 145: Session token can be null. User could theoretically provide a "regular" aws credential instead of a temporary one.

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
jyemin: No, I think I meant this to be `getSessionToken`.

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
jyemin: This is just saying that if you provide a credential supplier, _nothing_ should have been provided in the connection string, e.g. `mongodb://localhost/?authMechanism=MONGODB-AWS`

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
jyemin: Yeah, this definitely confused me too.  But for this PR I was trying to not change any existing behavior, and just refactor.  And I think this is also how the driver currently behaves.

jyemin: If it's wrong, I will open a separate ticket for that.

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
jyemin: I'm wrong.  I transcribed this correctly from the original.  It's equivalent to this code from `AwsAuthenticator#getSessionToken`:

```
            String token = credential.getMechanismProperty(AWS_SESSION_TOKEN_KEY, null);
            if (credential.getUserName() != null) {
                return token;
            }
            if (token != null) {
                throw new IllegalArgumentException("The connection string contains a session token without credentials");
            }
```

jyemin: The intention as I understand it is that a session token isn't allowed without an accessKeyId.  Is that correct?

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
jyemin: Or rather:

```
MongoCredential.createAwsCredential(null, null)
   .withMechanismProperty(AWS_SESSION_TOKEN_KEY, nulll)
   .withMechanismProperty(AWS_CREDENTIAL_PROVIDER_KEY, supplier)
```

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
markbenvenuto: Yeah, session token is not allowed without a accessKeyId.

AWS Credentials are either
(accessKeyId, secretAccessKey) - aka regular, non expiring, must be allocated in AWS console, per user
OR
(accessKeyId, secretAccessKey, sessionToken) - aka temporary, expiring, retrieved from ECS/EKS/EC2/assume role

Any single element is no enough to authenticate. Also combinations like (secretAccessKey, sessionToken) and (accessKeyId, sessionToken) simply do not work.

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
markbenvenuto: Sounds good

driver-core/src/main/com/mongodb/internal/authentication/AwsInternalCredential.java
markbenvenuto: Thanks for the clarification.

jyemin:
I thought more about [this comment thread](https://github.com/mongodb/mongo-java-driver/pull/879#discussion_r808642871), and realized that the code as currently organized does a very poor job of showing the actual intent, which I didn't understand until Mark explained it in the comments.  Given that, I decided to re-organize and refactor to make it clearer.  PTAL.

driver-core/src/main/com/mongodb/AwsCredential.java
```diff
@@ -38,12 +39,12 @@
      *
      * @param accessKeyId the non-null access key ID that identifies the temporary security credentials.
      * @param secretAccessKey the non-null secret access key that can be used to sign requests
-     * @param sessionToken the non-null session token
+     * @param sessionToken the session token, which may be null
```
jyemin, line 42: This is a small but backwards-compatible API change which allows us to use this class internally as well as externally, and doesn't really harm the external users.

driver-core/src/main/com/mongodb/internal/authentication/AwsCredentialHelper.java
```diff
@@ -0,0 +1,115 @@
...
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class AwsCredentialHelper {
+
+    public static AwsCredential obtainFromEnvironment() {
```
jyemin, line 35: Now this is just a helper class that obtains credentials from the environment, either from variables or web servers at known IP addresses.

driver-core/src/main/com/mongodb/internal/authentication/AwsCredentialHelper.java
```diff
@@ -0,0 +1,115 @@
...
+        if (System.getenv("AWS_ACCESS_KEY_ID") != null) {
+            return obtainFromEnvironmentVariables();
+        } else {
+            return obtainFromEc2OrEcsResponse();
+        }
+    }
+
+    private static AwsCredential obtainFromEnvironmentVariables() {
```
jyemin, line 43: This makes it clear that all the credential pieces come from variables, or none do.

driver-core/src/main/com/mongodb/internal/authentication/AwsCredentialHelper.java
```diff
@@ -0,0 +1,115 @@
...
+        return new AwsCredential(
+                System.getenv("AWS_ACCESS_KEY_ID"),
+                System.getenv("AWS_SECRET_ACCESS_KEY"),
+                System.getenv("AWS_SESSION_TOKEN"));
+    }
+
+    private static AwsCredential obtainFromEc2OrEcsResponse() {
```
jyemin, line 50: This makes it clear that all credential pieces come from the EC2/ECS response, or none do.

driver-core/src/main/com/mongodb/internal/connection/AwsAuthenticator.java
```diff
@@ -194,13 +173,37 @@ public void dispose() {
...
-            if (token != null) {
-                ret.append("t", new BsonString(token));
+            if (sessionToken != null) {
+                ret.append("t", new BsonString(sessionToken));
             }
 
             return toBson(ret);
         }
 
+        private AwsCredential createAwsCredential() {
```
jyemin, line 183: Centralized the logic for creating the AwsCredential.  This code can't be shared with CSFLE, but I think that's ok since the first two branches are going to be slightly different in that context anyway.

markbenvenuto (approved):

stIncMale (approved):
I like the code after refactoring much better that the original code, which was convoluted.

PR #881, opened 2022-02-18
fix: abort transaction in all cases

emasab:
catching RuntimeException all other subclasses of Exception are not caught and don't abort the transaction

### DIFF
--- a/driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
+++ b/driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
@@ -209,7 +209,7 @@ final class ClientSessionImpl extends BaseClientSessionImpl implements ClientSes
             try {
                 startTransaction(options);
                 retVal = transactionBody.execute();
-            } catch (RuntimeException e) {
+            } catch (Throwable e) {
                 if (transactionState == TransactionState.IN) {
                     abortTransaction();
                 }


esabellico-facephi:
a bit of context: maybe this can happen only in Kotlin you can actually compile code that throws checked exceptions or calling java code that throws checked exceptions without declaring them with throws in the TransactionBody interface

emasab:
here is some code to reproduce the error https://github.com/emasab/mongo-client-fix-881

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
```diff
@@ -209,7 +209,7 @@ public <T> T withTransaction(final TransactionBody<T> transactionBody, final Tra
             try {
                 startTransaction(options);
                 retVal = transactionBody.execute();
-            } catch (RuntimeException e) {
+            } catch (Throwable e) {
```
stIncMale, line 212: Such a tiny change, and so many things come to mind when thinking about it.

**`Error`s in Java the way they are make no sense**

My view on this is expressed [here](https://stackoverflow.com/questions/66961896/why-does-java-expose-error-and-java-se-api-callback-methods-accept-throwable-ins). A consequence of this view is that we should not catch `Error`s, which means we should catch `Exception`s instead of catching `Throwable`s each time we want to catch "everything" that makes sense catching (surely, there are exceptions from this rule, like catching assertion errors in test frameworks, or catching specific `Error`s that should have been `Exception`s, but were made `Error`s instead).

Unlike Java, Rust uses two drastically different mechanisms to represent recoverable and unrecoverable errors:

- Recoverable errors are represented via return values. Unlike in C, the compiler forces a caller to either handle them or to explicitly propagate as the caller's return value via the [`?` operator](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator).
- Unrecoverable errors are called panics, which can be [implemented via unwinding or aborting](https://doc.rust-lang.org/rustc/codegen-options/index.html#panic). [Unwinding](https://doc.rust-lang.org/nomicon/unwinding.html) is similar to throwing an exception in Java, and an unwinding panic can even be [caught](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html).

It is worth pointing out that the ability to catch unwinding panics in Rust should not be interpreted as "see, even in Rust catching unrecoverable errors is possible, which suggests that trying to handle unrecoverable errors (whether in Java or in Rust) may make sense!". The reason Rust allows catching unwinding panics is that [_"Rust's unwinding strategy is not specified to be fundamentally compatible with any other language's unwinding. As such, unwinding into Rust from another language, or unwinding into another language from Rust is Undefined Behavior. You must absolutely catch any panics at the FFI boundary!"_](https://doc.rust-lang.org/nomicon/unwinding.html#:~:text=Rust%27s%20unwinding%20strategy%20is%20not%20specified%20to%20be%20fundamentally%20compatible%20with%20any%20other%20language%27s%20unwinding.%20As%20such%2C%20unwinding%20into%20Rust%20from%20another%20language%2C%20or%20unwinding%20into%20another%20language%20from%20Rust%20is%20Undefined%20Behavior.%20You%20must%20absolutely%20catch%20any%20panics%20at%20the%20FFI%20boundary!) This point leads us directly to the issue that this PR is trying to address.

**Kotlin's exception handling is not compatible with Java's exception handling**

Unlike Java, Kotlin does not have the concept of checked exceptions (good for Kotlin). However, this means that Kotlin's exception mechanism is not compatible with Java's exception mechanism, and just like Rust code must catch any unwinding panics at the FFI boundary, Kotlin generally should catch all exceptions that are considered checked in Java (the check is simple: they are of the `Exception` type and not of the `RuntimeException` type) at the Kotlin-Java boundary. Unfortunately, the Kotlin's [interop guide](https://kotlinlang.org/docs/java-to-kotlin-interop.html#checked-exceptions) erroneously does not require anything like this. It partially addresses the incompatibility for a subset of cases, but that subset does not include the case addressed in this PR.

**Conclusion**

1. Even though I consider catching `Throwable`s instead of `Exception`s an undesirable practice, the code of the driver contains many places where this is done, which is why introducing one more such catch changes nothing.
2. Even though having unchecked exception thrown from Kotlin code that implements a Java API method that does not declare checked exceptions is a problem in the Kotlin code, we better work around it as suggested in the PR. Also, from now on we should remember to consider any method as being able to throw checked exception even it is not declared<sup>1</sup> (despite this not being true for all methods, for simplicity and robustness it is better to treat all of them this way).

---

<sup>1</sup> One may argue that even without calling Kotlin code, [sneaky throws](https://www.baeldung.com/java-sneaky-throws) are possible in Java due to the [bug-feature](https://stackoverflow.com/a/31316879) it introduced. However, in this case it is easy to argue that whoever uses the technique is fully responsible for the consequences. The same is much harder to say about Kotlin code because it is more abundant and Kotlin's documentation fails to require the right thing to be done for interoperability.

stIncMale (approved):

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
emasab: Nice reasoning. I add that if you decompile a Kotlin class file you see that a checked exception is wrapped inside a Throwable and then thrown, I've not tested it but I'd guess that it's done in Scala and Groovy too.

stIncMale: Interesting, it is not obvious to me why Kotlin compiler wraps checked exceptions in `Throwable` given that an exception of the `Throwable` type is considered checked by Java compiler.

stIncMale:
Thanks, @emasab, @esabellico-facephi, for informing the driver team about the problem. [JAVA-4575 Driver code should catch Exception instead of RuntimeException even if no checked exception is declared](https://jira.mongodb.org/browse/JAVA-4575) will address the problem in other places in the code.

emasab:
Not at all @stIncMale.  Correction about the Kotlin bytecode: it's not wrapped, it's casted to Throwable. And as it's checked in Java, it should be the Kotlin compiler that, not doing the checks, allows to compile to bytecode. If you decompile the Kotlin bytecode to Java it gives the error: you have to declare it with throws.

stIncMale:
@emasab I have not been familiar with this code, but it appears that the code is written with the following idea behind: a client may try to either commit or rollback a transaction, but cannot try both:

1. Both [`commitTransaction`](https://github.com/emasab/mongo-java-driver/blob/1294de8a3daf0bcc0403d127f8c932511a51ea18/driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java#L156-L157) and [`abortTransaction`](https://github.com/emasab/mongo-java-driver/blob/1294de8a3daf0bcc0403d127f8c932511a51ea18/driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java#L185-L187) change `transactionState` to `COMMITTED`/`ABORTED` respectively regardless of whether they succeed or fail.
2. `commitTransaction` [fails](https://github.com/emasab/mongo-java-driver/blob/1294de8a3daf0bcc0403d127f8c932511a51ea18/driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java#L133-L134) if `transactionState` is `ABORTED`. `abortTransaction` [fails](https://github.com/emasab/mongo-java-driver/blob/1294de8a3daf0bcc0403d127f8c932511a51ea18/driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java#L167-L168) if `transactionState` is `COMMITTED`.

`clearTransactionContextOnError` [called](https://github.com/emasab/mongo-java-driver/blob/1294de8a3daf0bcc0403d127f8c932511a51ea18/driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java#L230) when `commitTransaction` fails must be doing the client-side cleanup needed to either retry `commitTransaction` or abandon the transaction; and the server must cleanup resources related to an abandoned transaction eventually.

> we're still experiencing a case where the same sessionId is used with two different transaction ids even if the session is started and closed with the transaction

`startTransaction` [may be executed multiple times](https://github.com/emasab/mongo-java-driver/blob/1294de8a3daf0bcc0403d127f8c932511a51ea18/driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java#L207-L210) per a `withTransaction` execution. This method [advances](https://github.com/emasab/mongo-java-driver/blob/1294de8a3daf0bcc0403d127f8c932511a51ea18/driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java#L119) the transaction number. That is, it appears to be normal to see more than one `txnNumber` for a session even if there is only one `withTransaction` for that session.

emasab:
@stIncMale thanks for the explanation it's much clearer now, about the txnId it's giving this error `Caused by: com.mongodb.MongoQueryException: Query failed with error code 251 and error message 'cannot continue txnId 28 for session f1949373-d84d-468e-8667-98389bc480b9 - W/Ao6W6j9bCrE7yhAkO0wEm49NrHTolA71DjF3jfPsw= with txnId 29` but it's using a version that is copied from this one but adapted to Kotlin's coroutines so if debugging further I find something that applies to this code too, I'm sending some code to reproduce it.

PR #883, opened 2022-03-02
ByteBuffer is not released when compression is enabled, causing pool leakage

kulabun:
Hey Mongo team.

My team experienced an increased number of failures caused by latency recently. Further investigation showed that InternalStreamConnection doesn't close buffers returned by PowerOfTwoBufferPool when compression is enabled. It leads to permits leakage. After INTEGER_MAX buffer retrievals buffer pool is fully exhausted and all following requests would be blocked indefinitely unless the thread gets interrupted. 

JAVA-4510

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
@@ -319,23 +319,18 @@ public class InternalStreamConnection implements InternalConnection {
     @Override
     public <T> T sendAndReceive(final CommandMessage message, final Decoder<T> decoder, final SessionContext sessionContext,
             final RequestContext requestContext) {
-        ByteBufferBsonOutput bsonOutput = new ByteBufferBsonOutput(this);
         CommandEventSender commandEventSender;
 
-        try {
+        try (ByteBufferBsonOutput bsonOutput = new ByteBufferBsonOutput(this)) {
             message.encode(bsonOutput, sessionContext);
             commandEventSender = createCommandEventSender(message, bsonOutput, requestContext);
             commandEventSender.sendStartedEvent();
-        } catch (RuntimeException e) {
-            bsonOutput.close();
-            throw e;
-        }
-
-        try {
-            sendCommandMessage(message, bsonOutput, sessionContext);
-        } catch (RuntimeException e) {
-            commandEventSender.sendFailedEvent(e);
-            throw e;
+            try {
+                sendCommandMessage(message, bsonOutput, sessionContext);
+            } catch (RuntimeException e) {
+                commandEventSender.sendFailedEvent(e);
+                throw e;
+            }
         }
 
         if (message.isResponseExpected()) {
@@ -592,6 +587,7 @@ public class InternalStreamConnection implements InternalConnection {
         writeAsync(byteBuffers, errorHandlingCallback(callback, LOGGER));
     }
 
+    @SuppressWarnings("try")
     private void writeAsync(final List<ByteBuf> byteBuffers, final SingleResultCallback<Void> callback) {
         try {
             stream.writeAsync(byteBuffers, new AsyncCompletionHandler<Void>() {
@@ -607,8 +603,14 @@ public class InternalStreamConnection implements InternalConnection {
                 }
             });
         } catch (Throwable t) {
-            close();
-            callback.onResult(null, t);
+            try (AutoCloseable second = () -> callback.onResult(null, t);
+                 AutoCloseable first = this::close) {
+                releaseAllBuffers(byteBuffers);
+            } catch (RuntimeException e) {
+                throw e;
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


kulabun:
An alternative way to address the issue can be buffer release as part of `Compressor.ByteBufInputStream.close()` method. For example, SocketInputStream works this way and closes the underlying socket as part of `close` call. But I think it's better to be explicit and close it within the same section that used it.

stIncMale:
Hi @kulabun,

Thank you for the PR. I looked through the class `InternalStreamConnection` trying to see whether there are other places where we may not release buffers, and pushed a new commit.

Using your code in JAVA-4510 I confirmed that after the changes threads are not blocked due to `ConcurrentPool.maxSize` being exhausted. This can be done quickly by changing the `ConcurrentPool.INFINITE_SIZE` constant from `Integer.MAX_VALUE` to something small.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -319,23 +319,18 @@ public boolean isClosed() {
     @Override
     public <T> T sendAndReceive(final CommandMessage message, final Decoder<T> decoder, final SessionContext sessionContext,
             final RequestContext requestContext) {
-        ByteBufferBsonOutput bsonOutput = new ByteBufferBsonOutput(this);
         CommandEventSender commandEventSender;
 
-        try {
+        try (ByteBufferBsonOutput bsonOutput = new ByteBufferBsonOutput(this)) {
```
jyemin, line 324: I'm surprised that this works.  `bsonOutput` is passed to `sendCommandMessage`, which takes responsibility for calling `close`.  So aren't we double-closing here?  It doesn't look like `ByteBufferBsonOutput` protects against that.

jyemin: There's a good reason for this, too.  If the message needs to be compressed, we want to release the buffers holding the uncompressed message once we have compressed it into a new set of buffers.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -607,8 +603,14 @@ public void failed(final Throwable t) {
                 }
             });
         } catch (Throwable t) {
-            close();
-            callback.onResult(null, t);
+            try (AutoCloseable second = () -> callback.onResult(null, t);
+                 AutoCloseable first = this::close) {
+                releaseAllBuffers(byteBuffers);
```
jyemin, line 608: I don't think this is correct.  The callback that is passed to this method already has responsibility for releasing the buffers.   See line 490 in this file.

jyemin (changes requested):

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
stIncMale: > `bsonOutput` is passed to `sendCommandMessage`, which takes responsibility for calling close

Right, one of the examples where non-local knowledge is required to reason about correctness. I paid attention to this on my own, and fortunately, `ByteBufferBsonOutput.close` must be idempotent because `Closeable.close` is specified as such: _"If the stream is already closed then invoking this method has no effect."_ (the Java SE API is clearly documented incorrectly by implying that any `Closeable` implementation is a stream, but the idempotence is nonetheless there). Indeed, `ByteBufferBsonOutput.close` clears its `bufferList` to avoid releasing buffers more than once.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
stIncMale: Thanks you! 👏  It is great that you spotted this. Rolled back the changes to this method.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
jyemin: OK, I totally overlooked the buffer list clearing when I glanced at the close method.  I was expecting something like `if (!closed)`.

Still I wonder if we should modify this method in the same commit.  Is this fixing an actual bug, or clarifying intent?  If the latter, maybe we should just document the intent about who is responsible for closing the instance.

jyemin (commented):
Added a comment.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
stIncMale: Since I don't think that trying to handle `Error`s is reasonable (an we don't do that in many cases), this change does not fix a bug: try-finally is not better than try-catch(RuntimeException) in this case. So it only clarifies the intent. However, the method `send(final CommandMessage message, final Decoder<T> decoder, final SessionContext sessionContext)` located right after this one also uses try-with-resources block to close `bsonOutput` and also calls `sendCommandMessage`. The new code not only makes the intent cleaner, it makes the code consistent with that method.

If at this point you still want me to roll the changes in this method back, I will.

jyemin: I would prefer it, yes.  I think we should open a separate PR to refactor the code to clarify intent.

stIncMale: Done. The changes are not in https://github.com/mongodb/mongo-java-driver/pull/884.

jyemin (approved):
LGTM

PR #884, opened 2022-03-03
Refactor buffer cleaning in `InternalStreamConnection.sendAndReceive`

stIncMale:
This change was extracted from https://github.com/mongodb/mongo-java-driver/pull/883.


jyemin (commented):
Somehow it confuses me.  Maybe because I've never written code like this that takes advantage of Closeable idempotency so explicitly.  

But it looks correct so if you're confident about this pattern I will LGTM.

stIncMale:
> Somehow it confuses me. Maybe because I've never written code like this that takes advantage of Closeable idempotency so explicitly.

Actually, you [did here](https://github.com/mongodb/mongo-java-driver/commit/c8b028a509c825c0de30c67fa1eb6a5871d8bea9#diff-b4c0702b1d188ac6a1328c1a0807e4ea4235781735aff239169e6d37f167d0bdR292-R300) the same exact thing :ъ

jyemin (approved):
Well, it must be a good idea then ;)

PR #886, opened 2022-03-07
Prioritize electionId over setVersion when updating topology

jyemin:
JAVA-4375

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
@@ -228,9 +228,7 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
         }
 
         if (newDescription.getType() == REPLICA_SET_GHOST) {
-            if (LOGGER.isInfoEnabled()) {
-                LOGGER.info(format("Server %s does not appear to be a member of an initiated replica set.", newDescription.getAddress()));
-            }
+            LOGGER.info(format("Server %s does not appear to be a member of an initiated replica set.", newDescription.getAddress()));
             return true;
         }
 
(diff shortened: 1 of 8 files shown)


driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -251,46 +249,25 @@ private boolean handleReplicaSetMemberChanged(final ServerDescription newDescrip
         if (newDescription.getCanonicalAddress() != null
                 && !newDescription.getAddress().equals(new ServerAddress(newDescription.getCanonicalAddress()))
                 && !newDescription.isPrimary()) {
-            if (LOGGER.isInfoEnabled()) {
-                LOGGER.info(format("Canonical address %s does not match server address.  Removing %s from client view of cluster",
-                                   newDescription.getCanonicalAddress(), newDescription.getAddress()));
-            }
+            LOGGER.info(format("Canonical address %s does not match server address.  Removing %s from client view of cluster",
```
rozza, line 252: This will always format a string even if Logger.isInfoEnabled is false. I recall this pattern being added due to the cost of such operations. Is that desired.

SLF4J uses an varargs method eg: `public void info(String format, Object... arguments);` which prevents this and encompasses the check.  Added JAVA-4521

rozza (commented):
One question about the removal of logging info check.

Other than that LGTM

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
jyemin: In this context it's an unnecessary optimization as it's executing on background threads.  But it's a drive-by change and can be reverted if you prefer.

rozza: Would rather fix it in JAVA-4521 - it would improve all the code.

rozza (approved):

stIncMale (approved):

PR #887, opened 2022-03-07
Make buffer pool shared and support pruning

jyemin:
JAVA-4511
JAVA-4518

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
@@ -38,18 +38,14 @@ import static com.mongodb.internal.connection.ConcurrentPool.INFINITE_SIZE;
  *
  * <p>This class should not be considered a part of the public API.</p>
  */
-public class PowerOfTwoBufferPool implements BufferProvider, AutoCloseable {
+public class PowerOfTwoBufferPool implements BufferProvider {
 
     /**
      * The global default pool.  Pruning is enabled on this pool. Idle buffers are pruned after one minute.
      */
-    public static final PowerOfTwoBufferPool DEFAULT = new PowerOfTwoBufferPool();
+    public static final PowerOfTwoBufferPool DEFAULT = new PowerOfTwoBufferPool().enablePruning();
 
     private static final long DEFAULT_MAX_IDLE_TIME_MINUTES = 1;
-
-    static {
-        DEFAULT.enablePruning();
-    }
 
     private static final class IdleTrackingByteBuffer {
         private final long lastUsedNanos;
@@ -87,7 +83,7 @@ public class PowerOfTwoBufferPool implements BufferProvider, AutoCloseable {
 
         @Override
         public Prune shouldPrune(final IdleTrackingByteBuffer idleTrackingByteBuffer) {
-            return idleTrackingByteBuffer.getLastUsedNanos() < System.nanoTime() + maxIdleTimeNanos
+            return System.nanoTime() - idleTrackingByteBuffer.getLastUsedNanos() >= maxIdleTimeNanos
                     ? Prune.YES : Prune.STOP;
         }
     }
@@ -100,7 +96,7 @@ public class PowerOfTwoBufferPool implements BufferProvider, AutoCloseable {
     /**
      * Construct an instance with a highest power of two of 24.
      */
-    public PowerOfTwoBufferPool() {
+    PowerOfTwoBufferPool() {
         this(24);
     }
 
@@ -109,7 +105,7 @@ public class PowerOfTwoBufferPool implements BufferProvider, AutoCloseable {
      *
      * @param highestPowerOfTwo the highest power of two buffer size that will be pooled
      */
-    public PowerOfTwoBufferPool(final int highestPowerOfTwo) {
+    PowerOfTwoBufferPool(final int highestPowerOfTwo) {
         this(highestPowerOfTwo, DEFAULT_MAX_IDLE_TIME_MINUTES, TimeUnit.MINUTES);
     }
 
@@ -120,7 +116,7 @@ public class PowerOfTwoBufferPool implements BufferProvider, AutoCloseable {
      * @param maxIdleTime max idle time when pruning is enabled
      * @param timeUnit time unit of maxIdleTime
      */
-    public PowerOfTwoBufferPool(final int highestPowerOfTwo, final long maxIdleTime, final TimeUnit timeUnit) {
+    PowerOfTwoBufferPool(final int highestPowerOfTwo, final long maxIdleTime, final TimeUnit timeUnit) {
         int powerOfTwo = 1;
         for (int i = 0; i <= highestPowerOfTwo; i++) {
             int size = powerOfTwo;
@@ -134,12 +130,12 @@ public class PowerOfTwoBufferPool implements BufferProvider, AutoCloseable {
     /**
      * Call this method at most once to enable a background thread that prunes idle buffers from the pool
      */
-    public void enablePruning() {
+    PowerOfTwoBufferPool enablePruning() {
         pruner.scheduleAtFixedRate(this::prune, maxIdleTimeNanos, maxIdleTimeNanos / 2, TimeUnit.NANOSECONDS);
+        return this;
     }
 
-    @Override
-    public void close() {
+    void disablePruning() {
         pruner.shutdownNow();
     }
 
--- a/driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
@@ -45,8 +45,6 @@ public class PowerOfTwoBufferPool implements BufferProvider {
      */
     public static final PowerOfTwoBufferPool DEFAULT = new PowerOfTwoBufferPool().enablePruning();
 
-    private static final long DEFAULT_MAX_IDLE_TIME_MINUTES = 1;
-
     private static final class IdleTrackingByteBuffer {
         private final long lastUsedNanos;
         private final ByteBuffer buffer;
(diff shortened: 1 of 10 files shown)


driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
```diff
@@ -34,9 +38,64 @@
...
+        }
+
+        @Override
+        public void close(final IdleTrackingByteBuffer idleTrackingByteBuffer) {
+        }
+
+        @Override
+        public Prune shouldPrune(final IdleTrackingByteBuffer idleTrackingByteBuffer) {
+            return idleTrackingByteBuffer.getLastUsedNanos() < System.nanoTime() + maxIdleTimeNanos
+                    ? Prune.YES : Prune.STOP;
```
jyemin, line 91: Stop when we find the first non-idle buffer, since they should be roughly ordered in the pool by last use time (though no lock strictly enforces that invariant)

driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
```diff
@@ -51,27 +110,37 @@ public PowerOfTwoBufferPool() {
...
+    /**
+     * Call this method at most once to enable a background thread that prunes idle buffers from the pool
+     */
+    public void enablePruning() {
+        pruner.scheduleAtFixedRate(this::prune, maxIdleTimeNanos, maxIdleTimeNanos / 2, TimeUnit.NANOSECONDS);
+    }
+
+    @Override
+    public void close() {
+        pruner.shutdownNow();
```
jyemin, line 143: Note that the default pool is never closed, so this daemon thread hangs around until JVM shutdown

driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
```diff
@@ -34,9 +38,64 @@
...
+            return new IdleTrackingByteBuffer(createNew(size));
+        }
+
+        @Override
+        public void close(final IdleTrackingByteBuffer idleTrackingByteBuffer) {
+        }
+
+        @Override
+        public Prune shouldPrune(final IdleTrackingByteBuffer idleTrackingByteBuffer) {
+            return idleTrackingByteBuffer.getLastUsedNanos() < System.nanoTime() + maxIdleTimeNanos
```
stIncMale, line 90: Values returned by `System.nanoTime()`  are not wall-clock readings, and adding anything to them is meaningless. This is specified by the API: [_"This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time ... To compare elapsed time against a timeout, use `if (System.nanoTime() - startTime >= timeoutNanos) ...` instead of `if (System.nanoTime() >= startTime + timeoutNanos) ...` because of the possibility of numerical overflow."_](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html#nanoTime()) We must express the condition as `System.nanoTime() - idleTrackingByteBuffer.getLastUsedNanos() >= maxIdleTimeNanos`.

P.S. Even if `System.nanoTime()` represented wall-clock, the proposed comparison would have been different: `System.nanoTime() >= idleTrackingByteBuffer.getLastUsedNanos() + maxIdleTimeNanos`. But since `System.nanoTime()` does not return wall-clock time, the following values are possible and lead to an incorrect decision:

- `idleTrackingByteBuffer.getLastUsedNanos() == Long.MAX_VALUE - 10`;
- `System.nanoTime() == Long.MAX_VALUE`, i.e., the buffer was last used 10 ns ago;
- `idleTrackingByteBuffer.getLastUsedNanos() + maxIdleTimeNanos == Long.MAX_VALUE - 10 + TimeUnit.MINUTES.toNanos(1) == -9223371976854775819` , which is negative and, therefore, less than `Long.MAX_VALUE`, i.e., less than `System.nanoTime()`. This means we would have pruned just after 10 ns of idling.

driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
```diff
@@ -51,27 +110,37 @@ public PowerOfTwoBufferPool() {
...
             powerOfTwo = powerOfTwo << 1;
         }
+        maxIdleTimeNanos = timeUnit.toNanos(maxIdleTime);
+        pruner = Executors.newSingleThreadScheduledExecutor(new DaemonThreadFactory("BufferPoolPruner"));
+    }
+
+    /**
+     * Call this method at most once to enable a background thread that prunes idle buffers from the pool
+     */
+    public void enablePruning() {
```
stIncMale, line 137: If we start returning `this` here, we could write `DEFAULT = new PowerOfTwoBufferPool().enablePruning()` and get rid of the `static` block.

driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
```diff
@@ -51,27 +110,37 @@ public PowerOfTwoBufferPool() {
...
+        pruner.scheduleAtFixedRate(this::prune, maxIdleTimeNanos, maxIdleTimeNanos / 2, TimeUnit.NANOSECONDS);
```
stIncMale, line 138: Using `maxIdleTimeNanos / 2` instead of `maxIdleTimeNanos` as the period reduces the practically achievable "max idle time" from roughly `2 * maxIdleTimeNanos` to roughly `1.5 * maxIdleTimeNanos`. Pointing this out just to make sure we have the same understanding: despite `maxIdleTimeNanos` is set to 1 minute, idle buffers may stay in the pool for up to 1.5 minutes even without any unexpected freezes caused by the GC or the OS scheduler.

driver-core/src/test/unit/com/mongodb/internal/connection/PowerOfTwoBufferPoolTest.java
```diff
@@ -73,4 +74,18 @@ public void testHugeBufferRequest() {
...
+    @Test
+    public void testPruning() throws InterruptedException {
+        try (PowerOfTwoBufferPool pool = new PowerOfTwoBufferPool(10, 5, TimeUnit.MILLISECONDS)) {
+            pool.enablePruning();
+            ByteBuf byteBuf = pool.getBuffer(256);
+            ByteBuffer wrappedByteBuf = byteBuf.asNIO();
+            byteBuf.release();
+            Thread.sleep(50);
+            ByteBuf newByteBuf = pool.getBuffer(256);
+            assertNotSame(wrappedByteBuf, newByteBuf.asNIO());
```
stIncMale, line 88: It may be worth also checking `assertSame`:
```java
ByteBuf byteBuf = pool.getBuffer(256);
ByteBuffer wrappedByteBuf = byteBuf.asNIO();
byteBuf.release();

byteBuf = pool.getBuffer(256);
assertSame(wrappedByteBuf, byteBuf.asNIO());
byteBuf.release();

Thread.sleep(50);
byteBuf = pool.getBuffer(256);
assertNotSame(wrappedByteBuf, byteBuf.asNIO());
```

driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
```diff
@@ -51,27 +110,37 @@ public PowerOfTwoBufferPool() {
...
+    }
+
+    /**
+     * Construct an instance.
+     *
+     * @param highestPowerOfTwo the highest power of two buffer size that will be pooled
+     * @param maxIdleTime max idle time when pruning is enabled
+     * @param timeUnit time unit of maxIdleTime
+     */
+    public PowerOfTwoBufferPool(final int highestPowerOfTwo, final long maxIdleTime, final TimeUnit timeUnit) {
```
stIncMale, line 123: It appears that this constructor can be package-access. If that's the case, let's remove `public`.

driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
```diff
@@ -51,27 +110,37 @@ public PowerOfTwoBufferPool() {
      * @param highestPowerOfTwo the highest power of two buffer size that will be pooled
      */
     public PowerOfTwoBufferPool(final int highestPowerOfTwo) {
+        this(highestPowerOfTwo, DEFAULT_MAX_IDLE_TIME_MINUTES, TimeUnit.MINUTES);
```
stIncMale, line 113: Given that

- this is the only place where `DEFAULT_MAX_IDLE_TIME_MINUTES` is used;
- the meaning of a literal here would be clear because it is used as the value of the `maxIdleTime` parameter

introducing a constant instead of using a literal 1 does not seem helpful here.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
jyemin: Yes, that sounds right.  I'm ok with it since 1 minute is not a publicly settable value, so anything we do here is arbitrary.  We could always run the thread more often.

driver-core/src/test/unit/com/mongodb/internal/connection/PowerOfTwoBufferPoolTest.java
jyemin: I didn't do this because that behavior is already tested in `testReuse`

driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
jyemin: This led me down a small rabbit hole.  Since besides the test, all clients must use `DEFAULT`, then those clients should not be able to enable pruning or close `DEFAULT`.  So I also made the pruning methods package-access, and removed `AutoCloseable` from the interface list.

driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
jyemin: Done

driver-core/src/test/unit/com/mongodb/internal/connection/PowerOfTwoBufferPoolTest.java
stIncMale: Right, sorry for not checking that we have a test for this behavior.

stIncMale (approved):
Approved, but the tests are red.

stIncMale (approved):

PR #889, opened 2022-03-10
Support NEED_KMS_CREDENTIALS state for CSFLE

jyemin:
Links useful for reviewers:

- [Summary of changes for bindings and drivers to add support for on-demand credentials in AWS](https://jira.mongodb.org/browse/MONGOCRYPT-382?focusedCommentId=4401229&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4401229).
- [New mongodb-crypt API](https://github.com/mongodb/libmongocrypt/commit/a6a17a052e86a21910d9e924923c183919797fc7).

JAVA-4503

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ ext {
     nettyTcnativeBoringsslVersion = '2.0.46.Final'
     snappyVersion = '1.1.8.4'
     zstdVersion = '1.5.0-4'
-    mongoCryptVersion = '1.4.0-SNAPSHOT'
+    mongoCryptVersion = '1.4.0-alpha0'
     projectReactorVersion = 'Californium-SR23'
     junitBomVersion = '5.8.1'
     gitVersion = getGitVersion()
@@ -73,7 +73,6 @@ configure(coreProjects) {
         mavenLocal()
         google()
         mavenCentral()
-        maven { url "https://oss.sonatype.org/content/repositories/snapshots/"}
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ ext {
     nettyTcnativeBoringsslVersion = '2.0.46.Final'
     snappyVersion = '1.1.8.4'
     zstdVersion = '1.5.0-4'
-    mongoCryptVersion = '1.3.0'
+    mongoCryptVersion = '1.4.0-SNAPSHOT'
     projectReactorVersion = 'Californium-SR23'
     junitBomVersion = '5.8.1'
     gitVersion = getGitVersion()
--- a/build.gradle
+++ b/build.gradle
@@ -73,6 +73,7 @@ configure(coreProjects) {
         mavenLocal()
         google()
         mavenCentral()
+        maven { url "https://oss.sonatype.org/content/repositories/snapshots/"}
     }
 }
 
--- a/driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
+++ b/driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
@@ -16,6 +16,7 @@
 
 package com.mongodb.internal.capi;
 
+import com.mongodb.AwsCredential;
 import com.mongodb.Block;
 import com.mongodb.ConnectionString;
 import com.mongodb.MongoClientException;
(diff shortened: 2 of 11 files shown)


build.gradle
```diff
@@ -73,6 +73,7 @@ configure(coreProjects) {
         mavenLocal()
         google()
         mavenCentral()
+        maven { url "https://oss.sonatype.org/content/repositories/snapshots/"}
```
jyemin, line 76: Temporary change until we get a 1.4.0 beta published

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
```diff
@@ -40,13 +43,36 @@ public static MongoCryptOptions createMongoCryptOptions(final Map<String, Map<St
                                                             final Map<String, BsonDocument> namespaceToLocalSchemaDocumentMap) {
         MongoCryptOptions.Builder mongoCryptOptionsBuilder = MongoCryptOptions.builder();
 
+        BsonDocument bsonKmsProviders = getKmsProvidersAsBsonDocument(kmsProviders);
+        mongoCryptOptionsBuilder.kmsProviderOptions(bsonKmsProviders);
+        mongoCryptOptionsBuilder.localSchemaMap(namespaceToLocalSchemaDocumentMap);
+        mongoCryptOptionsBuilder.needsKmsCredentialsStateEnabled(true);
```
jyemin, line 49: This enables the new state in the state machine

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
```diff
@@ -40,13 +43,36 @@ public static MongoCryptOptions createMongoCryptOptions(final Map<String, Map<St
...
+        return mongoCryptOptionsBuilder.build();
+    }
+
+    public static BsonDocument fetchCredentials(final Map<String, Map<String, Object>> kmsProviders) {
```
jyemin, line 53: So the logic can be shared between sync and reactive, we put it here

driver-sync/src/main/com/mongodb/client/internal/Crypt.java
```diff
@@ -59,23 +62,27 @@
      * @param keyRetriever the key retriever
      * @param keyManagementService the key management service
      */
-    Crypt(final MongoCrypt mongoCrypt, final KeyRetriever keyRetriever, final KeyManagementService keyManagementService) {
-        this(mongoCrypt, null, null, keyRetriever, keyManagementService, false, null);
+    Crypt(final MongoCrypt mongoCrypt, final Map<String, Map<String, Object>> kmsProviders, final KeyRetriever keyRetriever,
```
rozza, line 65: Nit: add javadoc

rozza (approved):
LGTM - one javadoc nit

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
```diff
@@ -40,13 +43,36 @@ public static MongoCryptOptions createMongoCryptOptions(final Map<String, Map<St
...
+        BsonDocument kmsProvidersDocument = MongoCryptHelper.getKmsProvidersAsBsonDocument(kmsProviders);
+        if (kmsProvidersDocument.containsKey("aws") && kmsProvidersDocument.get("aws").asDocument().isEmpty()) {
+            AwsCredential awsCredential = AwsCredentialHelper.obtainFromEnvironment();
```
stIncMale, line 56: `AwsCredentialHelper.obtainFromEnvironment` may call `AwsCredentialHelper.obtainFromEc2OrEcsResponse`, which is a blocking call because it does IO. As a result, the method `com.mongodb.reactivestreams.client.internal.crypt.Crypt.fetchCredentials` is blocking even though it is also asynchronous. Consequently, `com.mongodb.reactivestreams.client.internal.crypt.Crypt.encrypt`/`decrypt` are now blocking, though previously they were both asynchronous and non-blocking. This, in turn, correspondingly affects the properties of the driver reactive API.

I don't have good enough understanding to say whether it's a problem or not (maybe our reactive API blocks in some situations even without this change), but it is worth thinking about or discussing.

build.gradle
```diff
@@ -48,7 +48,7 @@ ext {
     nettyTcnativeBoringsslVersion = '2.0.46.Final'
     snappyVersion = '1.1.8.4'
     zstdVersion = '1.5.0-4'
-    mongoCryptVersion = '1.3.0'
+    mongoCryptVersion = '1.4.0-alpha0'
```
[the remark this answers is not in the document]
jyemin: There's no process for it, but since we do the mongodb-crypt releases we're not likely to forget.  A ticket wouldn't hurt though.

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
jyemin: It's a problem that I'm aware of.  We already do the same thing in a couple of `SaslClient` implementations, including the one in AwsAuthenticator which uses the same helper method.

PR #891, opened 2022-03-13
Add builders for the `$search` pipeline stage

stIncMale:
**Superseded by https://github.com/mongodb/mongo-java-driver/pull/952.**

The API follows the approach proposed by @jyemin in https://github.com/mongodb/mongo-java-driver/pull/778 with the difference that it exposes types as interfaces and refrains from introducing separate classes-containers for static builder methods. Instead, static builder methods are placed in the interfaces, which is exactly the use case for which static methods were allowed in interfaces (for more of my thoughts see [here](https://github.com/mongodb/mongo-java-driver/pull/778#discussion_r718651160)).

Another significant difference is in the implementation. You will not find many classes implementing the interfaces and having various fields, instead all interfaces are implemented with a handful of classes, which in turn extend either `AbstractConstructibleBson` or `AbstractConstructibleBsonElement`.

Note that the API related to [counting](https://docs.atlas.mongodb.com/atlas-search/counting/) and [`facet`](https://docs.atlas.mongodb.com/atlas-search/facet/) is marked as `@Beta` because [_"Atlas Search `facet` and `count` are in preview, but can be used in production applications"_](https://docs.atlas.mongodb.com/atlas-search/query-syntax/#-searchmeta). All API is marked with a new annotation `@Evolving` to make it clear that users should not implement/extend it. We implicitly rely on the restrictions specified by this annotation in some APIs, e.g. in `MongoDatabase`, `MongoCollection`, and it seems useful to have a way to make the restriction explicit.

Progress status:

- API
  - API for [`$search`](https://docs.atlas.mongodb.com/atlas-search/query-syntax/#-search) is done and supports both [operators](https://docs.atlas.mongodb.com/atlas-search/operators/) and [collectors](https://docs.atlas.mongodb.com/atlas-search/collectors/).
  - API for operators is done only for the [`exists`](https://docs.atlas.mongodb.com/atlas-search/exists/) operator, the rest will be done in [JAVA-4190](https://jira.mongodb.org/browse/JAVA-4190).
  - API for all collectors (currently it's only [`facet`](https://docs.atlas.mongodb.com/atlas-search/facet/)) is fully done.
  - API for [highlighting](https://docs.atlas.mongodb.com/atlas-search/highlighting/) and [counting](https://docs.atlas.mongodb.com/atlas-search/counting/) search options is fully done.
  - Generic API for all types that allows constructing an instance when there is no satisfying builder method is fully done.
  - Scala API is done.
  - API for [`$searchMeta`](https://docs.atlas.mongodb.com/atlas-search/query-syntax/#-searchmeta) will be done in [JAVA-4315](https://jira.mongodb.org/browse/JAVA-4315). It will reuse the counting and `facet` API that is already done.
- Tests
  - Unit tests are done.
  - Integration tests are done.

Code examples:
See tests in `driver-core/src/test/unit/com/mongodb/client/model/search`, `AggregatesSearchIntegrationTest`, `AggregatesSpecification` for API examples.

JAVA-4415


stIncMale:
## Choosing between `ToBsonField` and `Bson`

### Intro

`Bson` is an interface allowing an object to render itself into a `BsonDocument`, similarly `ToBsonField` allows to an object to render itself into a `BsonField. Currently, `ToBsonField` types are:

- `WindowedComputation` (pre-existing type),
- `SearchOperator`,
- `SearchCollector`,
- `SearchFacet`.

### Concern

@jyemin raised a concern that `BsonField` is less versatile than `BsonDocument` (after all, it can represent only one name/value pair, while a document can represent multiple pairs), and, thus, if the search syntax ever changes such that  `SearchOperator`, `SearchCollector`, `SearchFacet` is a document, we would have a problem.

### Response

#### Versatility

### Internal

Replacing `ToBsonField` with `Bson` means that we need to wrap a single name/value pair in a document and then unwrap it when we need the name/value. Similarly, if what currently is a name/value pair becomes a document in the future as a result of a syntax change, we can wrap such a document in a name/value by using a dummy name and the document as a value, and then unwrap the document back when we need it. Thus, `ToBsonField` and `Bson` are of equal versatility from the standpoint of implementation.

### External

If we provide no API to construct a specific search collector, a user may use the `SearchCollector.of` method to construct what he needs. Here is an example of a user creating a facet collector with an operator for which we have API, and three facets, such that we don't have API for one of the facets:
```java
SearchCollector.of(new BsonField("facet",
        new BsonDocument("operator",
                exists(fieldPath("fieldName")).appendTo(new BsonDocument()))
                .append("facets",
                        numberFacet("numberFacetName",
                                fieldPath("numberFieldName"), asList(1, 2)).appendTo(
                        stringFacet("stringFacetName",
                                fieldPath("stringFieldName")).appendTo(new BsonDocument()))
                        .append("customFacetName",
                                new BsonDocument("type", new BsonString("customFacetType")))
                )));
```
We can see here that if `SearchOperator` were `Bson` instead of being `ToBsonField`, the call `appendTo(new BsonDocument())` after `exists(fieldPath("fieldName"))` would have been replaced with a `toBsonDocument()` call. Not a big difference.

If `SearchOperator` were `Bson` instead of being `ToBsonField`, the user would have had to unwrap each facet from its wrapping document and then create a document from all the unwrapped facets. We again need an `appendTo` method, and we can't have it on the `Bson` interface. We will need to add it to `SearchFacet`. As a result, the part of the example related to facets will not change at all<sup>1</sup>.

Thus, it appears that from the user perspective having `ToBsonField` or `Bson` hardly changes anything, at least within things that came to my mind.

#### Decision

If versatility does not seem be improved by using `Bson` instead of `ToBsonField`, is there a reason to choose `Bson` over `ToBsonField`? The answer is yes - the fewer interfaces we use, the better, unless using more interfaces gives us some benefits. Unfortunately, using `ToBsonField` in the search API does not seem to have benefits when compared to using `Bson`. This means, that I made a mistake, and should have been using `Bson` for all search API types from the very beginning.

---
<sup>1</sup> An alternative to having `ToBsonField.appendTo(BsonDocument)`/`SearchFacet.appendTo(BsonDocument)` instance methods is to have the `SearchFacet.combineToBsonDocument(SearchFacet...)` static method. Using it allows to avoid having a nasty hierarchy of `appendTo` calls that requires tricky formatting (pay attention how code is formatted in the example) to still look linearly, and spawns tons of `)` characters. I should probably go with this approach.

stIncMale:
Refactored

- `SearchPath`, it is `Bson` instead of `ToBsonValue`;
- `SearchFacet`, `SearchOperator`, `SearchCollector`, they are now `Bson` instead of `ToBsonField`.

This resulted in removing the `ToBsonValue`, `ToBsonField` interfaces and some implementations. I.e., reduced the amount of code in addition to making the types more consistent.

.evergreen/run-atlas-search-tests.sh
```diff
@@ -0,0 +1,19 @@
...
+############################################
+RELATIVE_DIR_PATH="$(dirname "${BASH_SOURCE[0]:-$0}")"
+source "${RELATIVE_DIR_PATH}/javaConfig.bash"
+
+echo "Running Atlas Search tests"
+./gradlew -version
+#./gradlew --stacktrace --info \
+#    -Dorg.mongodb.test.atlas.search=true \
+#    -Dorg.mongodb.test.uri=${MONGODB_URI} \
+#    driver-core:test --tests AggregatesSearchIntegrationTest
```
stIncMale, line 19: Commented out until [CLOUDP-116195 Atlas cluster for drivers to test search APIs](https://jira.mongodb.org/browse/CLOUDP-116195) is done.

One may run Atlas search integration tests locally by uncommenting this code and replacing `${MONGODB_URI}` with `mongodb+srv://driver-testing:JIbS2Flc30ENAOL0@free-cluster-search.ujz9d.mongodb.net`.

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchFacet.scala
```diff
@@ -0,0 +1,102 @@
...
+   * in the search results by breaking the results into separate ranges.
+   *
+   * @param name The facet name.
+   * @param path The path to facet on.
+   * @param boundaries Bucket boundaries in ascending order. Must contain at least two boundaries.
+   * The supported `Number` types are the same as those supported by org.bson.BasicBSONEncoder.putNumber`.
+   * @return The requested `SearchFacet`.
+   * @see [[https://www.mongodb.com/docs/atlas/atlas-search/facet/#numeric-facets Numeric facet definition]]
+   */
+  def numberFacet(name: String, path: FieldSearchPath, boundaries: Iterable[_ <: Number]): NumericSearchFacet =
```
stIncMale, line 58: Using `Iterable[_ <: Number]` here is a problem, because `Number` is not a Scala type. As a result, users have to write `List[Number](10, 20, 30)` instead of writing `List(10, 20, 30)`, for example.

A similar problem is with `Iterable[Instant]` (`Instant` is not specified as an upper boundary because this class is `final`).

@rozza Could you please help with figuring out what are the right types to use in Scala, and how to convert from them to Java's `Iterable<? extends Number>` and `Iterable<Instant>`?

driver-core/src/main/com/mongodb/client/model/search/StringSearchFacet.java
```diff
@@ -0,0 +1,35 @@
...
+@Beta
+@Evolving
+public interface StringSearchFacet extends SearchFacet {
+    /**
+     * Creates a new {@link StringSearchFacet} that explicitly limits the number of facet categories.
+     *
+     * @param max The maximum number of facet categories to return in the results.
+     * @return A new {@link StringSearchFacet}.
+     */
+    StringSearchFacet numBuckets(int max);
```
stIncMale, line 34: There is no Scala wrapper for the `StringSearchFacet.numBuckets` method because this method can be called directly from Scala without any inconvenience. Some other new interfaces in `driver-core/src/main/com/mongodb/client/model/search` also have similar instance methods that are not wrapped for the same reason.

I talked with @ross about this, and so far this approach seems to be fine. However, it seems concerning to me that if in the future we have instance methods that can't be conveniently called from Scala, we won't be able to wrap them because for that we would need to have a Scala type that wraps `com.mongodb.client.model.search.StringSearchFacet` and Scala API that always uses that type instead of `com.mongodb.client.model.search.StringSearchFacet`. It appears that if we don't wrap the type now, we won't be able to wrap it in the future. Ross, could you please confirm this? If this is the case, then we have to either wrap all types now, or accept that we are not going to wrap them in the future regardless of what new methods they grow.

driver-core/src/main/com/mongodb/internal/client/model/AbstractConstructibleBsonField.java
```diff
@@ -0,0 +1,83 @@
...
+ * A {@link Bson} that contains exactly one name/value pair
+ * and allows constructing new instances via {@link #newWithAppendedValue(String, BsonValue)} instead of mutating {@code this}.
+ * The value must itself be a {@code Bson}.
+ * While instances are not {@link Immutable immutable},
+ * instances with {@link BsonDocument#isEmpty() empty} values are treated specially and are immutable.
+ *
+ * @param <S> A type introduced by the concrete class that extends this abstract class.
+ * @see AbstractConstructibleBson
+ */
+public abstract class AbstractConstructibleBsonField<S extends AbstractConstructibleBsonField<S>> implements Bson {
```
rozza, line 36: This is essentially a Constructable BsonElement - could we use that name as both represent a bson key and value.

driver-core/src/main/com/mongodb/internal/client/model/AbstractConstructibleBson.java
```diff
@@ -0,0 +1,96 @@
...
+import com.mongodb.annotations.Immutable;
+import org.bson.BsonDocument;
+import org.bson.BsonValue;
+import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.conversions.Bson;
+
+import java.util.Objects;
+
+/**
+ * A {@link Bson} that allows constructing new instances via {@link #newAppended(String, BsonValue)} instead of mutating {@code this}.
```
rozza, line 27: Why not  make mutable?

driver-core/src/main/com/mongodb/client/model/search/ConstructibleBsonFieldToManifoldAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+package com.mongodb.client.model.search;
+
+import com.mongodb.internal.client.model.AbstractConstructibleBsonField;
+import org.bson.BsonInt32;
+import org.bson.BsonString;
+import org.bson.conversions.Bson;
+
+import static org.bson.assertions.Assertions.notNull;
+
+final class ConstructibleBsonFieldToManifoldAdapter extends AbstractConstructibleBsonField<ConstructibleBsonFieldToManifoldAdapter>
```
rozza, line 25: I think the name should relate to its usage - eg search.  

`SearchConstructibleBsonField` is clearer to me.

driver-core/src/main/com/mongodb/client/model/search/ConstructibleBsonToManifoldAdapter.java
```diff
@@ -0,0 +1,85 @@
...
+import com.mongodb.internal.client.model.AbstractConstructibleBson;
+import org.bson.BsonDocument;
+import org.bson.BsonInt32;
+import org.bson.BsonString;
+import org.bson.BsonValue;
+import org.bson.conversions.Bson;
+
+import static com.mongodb.assertions.Assertions.notNull;
+
+final class ConstructibleBsonToManifoldAdapter extends AbstractConstructibleBson<ConstructibleBsonToManifoldAdapter> implements
```
rozza, line 27: Naming again - relate it to the domain .

driver-core/src/main/com/mongodb/client/model/search/StringSearchFacet.java
rozza: With scala I think we'd be able to wrap it in the future.  Also we could always implicitly wrap an instance to provide extension methods that have scala types eg: Seq vs java Lists in the future.

driver-core/src/test/unit/com/mongodb/client/model/AggregatesSpecification.groovy
```diff
@@ -544,6 +554,99 @@ class AggregatesSpecification extends Specification {
         }''')
     }
 
+    def 'should render $search'() {
+        when:
+        BsonDocument searchDoc = toBson(
+                search(
+                        (SearchOperator) exists(fieldPath('fieldName')),
```
rozza, line 561: Is this casting needed due to Groovy weirdness?

rozza (commented):
I've taken a first review and have some questions, before I continue.

Do you think we are missing a trick by not having a `Bson.toBsonValue` method? That would allow for all `BsonValue` types to be generated by `Bson` types. It may be that the moment to add that has passed.

I like the builder style Constructible classes, although - I'm unsure as to why they are immutable when appending new data. I think regardless of adding a `toBsonValue` to `bson` this approach to building the search operators is good.

One other observation is the examples in the java docs use `BsonDocument` do you think using `Document` would be better - its our default `MongoCollection` type so more users will be familiar with that.

driver-core/src/test/unit/com/mongodb/client/model/AggregatesSpecification.groovy
stIncMale: Yes, this is due to how Groovy dispatches methods. Our implementations of `SearchOperator` and `SearchCollector` has the same run-time type, which makes both `Aggregates.search` methods valid candidates, and the dispatching code fails to pick the right one, because it completely disregards the compile-time types of the arguments passed to a method. Yet, it does not disregard the compile-time type of the argument when an explicit cast is added... This language is insane.

driver-core/src/main/com/mongodb/internal/client/model/AbstractConstructibleBson.java
```diff
@@ -0,0 +1,96 @@
...
+ * While instances are not {@link Immutable immutable},
+ * {@link BsonDocument#isEmpty() empty} instances are treated specially and are immutable.
+ *
+ * @param <S> A type introduced by the concrete class that extends this abstract class.
+ */
+public abstract class AbstractConstructibleBson<S extends AbstractConstructibleBson<S>> implements Bson {
```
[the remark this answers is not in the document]
stIncMale: Currently no, and there is no technicality that prevents me from moving it into `com.mongodb.client.model.search` and making package-access. However, if we are going to use the same approach for other query building APIs as the one this PR uses for Atlas search, we will have to make the class public, and, consequently, we will not be able to keep it in `com.mongodb.client.model.search`, so the class will end up living in the same internal package where it is currently located.

Do you want the class to me moved into `com.mongodb.client.model.search` for now?

driver-core/src/main/com/mongodb/annotations/Beta.java
```diff
@@ -32,7 +32,7 @@
  * about the quality or performance of the API in question, only the fact that
  * it is not "API-frozen."
  *
- * <p>It is generally safe for <i>applications</i> to depend on beta APIs, at
```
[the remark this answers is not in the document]
stIncMale: Sorry, apparently I accidentally pressed a space bar there. Done.

driver-core/src/main/com/mongodb/internal/client/model/AbstractConstructibleBsonField.java
stIncMale: Done. Renamed to `AbstractConstructibleBsonElement`.

driver-core/src/main/com/mongodb/client/model/search/ConstructibleBsonFieldToManifoldAdapter.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/search/ConstructibleBsonToManifoldAdapter.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/search/StringSearchFacet.java
jyemin: bq. My understanding is that binary compatibility is about the Java bytecode rules, and is not a concept that exists on the level of JVM programming language, but this understanding may be incorrect

That is my understanding as well.

stIncMale:
> One other observation is the examples in the java docs use BsonDocument do you think using Document would be better - its our default MongoCollection type so more users will be familiar with that.

`Document` may be convenient to represent BSON read from the server that a user is planning to use in Java. `BsonDocument`, on the other hand, appears to be more convenient when preparing BSON for sending it to the server, because `BsonDocument` allows you to use BSON data types directly, leaving no ambiguity with regard to what BSON will be sent.

stIncMale:
Meanwhile, a new search option was added: https://www.mongodb.com/docs/atlas/atlas-search/return-stored-source/. I am going to leave the API as is, as chasing the changes does not seem to be the right thing to do.

driver-core/src/main/com/mongodb/internal/client/model/AbstractConstructibleBson.java
rozza: Ah I see this pattern was adopted from the previous poc.  Personally I see no real benefit to making it clone each time.  Given that the main Document / BsonDocument classes are all mutable I'd opt for pushing immutability to the point of need rather than add a new pattern without an expressed need. I see Bson types generally being more akin to our builders and the toBsonDocument method being used as a final build stage.

So given that I'd rather a MutableBson abstract class, and a MutableBsonElement class - I see the implementation would be slightly simpler. This is no blocker for the Search API itself just an observation. 

Any thoughts @jyemin / @stIncMale

driver-core/src/main/com/mongodb/client/model/search/StringSearchFacet.java
rozza: In Scala implicits can be used to extend and add functionality to a type. So essentially an implicit could be used to add support for scala types to an existing class. For example, this class adds a `withoutVowels` method to any String instance.

```
implicit class SuperString(val str: String) extends AnyVal {
    def withoutVowels: String = str.replaceAll("[aeiou]", "")
}

// Usage:
println("Hello world".withoutVowels) // Hll wrld
```

That wouldn't break binary compatibility. 

So only static methods need wrapping and the sub types can be aliased.

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchFacet.scala
rozza: Just use: `Iterable[Number]` and then the converter will handle the types correctly eg:

```
println(SearchFacet.numberFacet("a", fieldPath("numberFieldName"), Seq(10, 20.0, 30L)))
```

Also: Instant seems the appropriate type in scala :+1:

rozza:
> Document may be convenient to represent BSON read from the server that a user is planning to use in Java. BsonDocument, on the other hand, appears to be more convenient when preparing BSON for sending it to the server, because BsonDocument allows you to use BSON data types directly, leaving no ambiguity with regard to what BSON will be sent.

Aren't the API docs for consumption for a user about how to use the API. If they rarely use BsonDocument it seems the wrong type to use as an example.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -0,0 +1,64 @@
...
+@Evolving
+public interface SearchOperator extends Bson {
+    /**
+     * Returns a {@link SearchOperator} that tests if the {@code path} exists in a document.
+     *
+     * @param path The path to test.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/exists/ exists operator
+     */
+    static ExistsSearchOperator exists(final FieldSearchPath path) {
```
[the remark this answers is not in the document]
stIncMale: There are two reasons at least one operator built in the API is needed:

- To be used in tests for `Aggregates.search` and `SearchFacet`. We can just write tests in an uglier manner by using `SearchOperator.of`, but I would like not to do that.
- To be used in the documentation of the `SearchOperator.of` method: all such methods include examples of constructing functionally equivalent objects via factory methods and typed built-in builders, and via a generic API.

There seems to be no problem in having factory methods for only a subset of operators (even if that subset contains a single element) because generally this will be the case as Atlas search evolves: Atlas adds a new operator/collector/etc., the driver API still does not support it via factory methods, as time goes by, the driver API evolves to support the new Atlas search feature.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Creates a {@link SearchOperator} from a {@link Bson} in situations when there is no builder method that better satisfies your needs.
+     * This method cannot be used to validate the syntax.
+     * <p>
+     * <i>Example</i><br>
+     * The following code creates two functionally equivalent {@link SearchOperator}s,
+     * though they may not be {@linkplain Object#equals(Object) equal}.
+     * <pre>{@code
```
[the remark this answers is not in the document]
stIncMale: We may start using them as soon as we modify the build process such that it uses Java SE 18.

driver-core/src/main/com/mongodb/annotations/Evolving.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Signifies that the annotated program element is subject to incompatible changes by means of adding abstract methods.
+ * This, in turn, means that implementing interfaces or extending classes annotated with {@code @}{@link Evolving} bears the risk
```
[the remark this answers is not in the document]
stIncMale: Why? Try rendering it in the IDE (F1 should work in Idea), it renders just fine.

bson/src/main/org/bson/BasicBSONEncoder.java
```diff
@@ -281,10 +281,22 @@ protected void putDate(final String name, final Date date) {
      * Encodes any number field.
      *
      * @param name   the field name
-     * @param number the value
+     * @param number The value. The following types are supported:
```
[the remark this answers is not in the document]
stIncMale: I needed to see a set of specific `Number` types that we automatically convert to BSON types, to document those types as supported by `SearchFacet.numberFacet`. If `BasicBSONEncoder.putNumber` is the wrong method to use as a reference, do you know what method/class we may suggest instead?

driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -1612,6 +1658,65 @@ public String toString() {
...
+            BsonDocumentWriter writer = new BsonDocumentWriter(new BsonDocument());
+            writer.writeStartDocument();
+            writer.writeStartDocument("$search");
+            BsonDocument operatorOrCollectorDoc = operatorOrCollector.toBsonDocument(tDocumentClass, codecRegistry);
+            assertTrue(operatorOrCollectorDoc.size() == 1);
+            Map.Entry<String, BsonValue> operatorOrCollectorEntry = operatorOrCollectorDoc.entrySet().iterator().next();
+            writer.writeName(operatorOrCollectorEntry.getKey());
+            BuildersHelper.encodeValue(writer, operatorOrCollectorEntry.getValue(), codecRegistry);
+            if (options != null) {
+                options.toBsonDocument().forEach((optionName, optionValue) -> {
```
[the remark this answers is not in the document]
stIncMale: Here I am appending name-value pairs from options to a BSON document, which seems different to just writing a value via `BuildersHelper.encodeValue(writer, options, codecRegistry)`.

If I replace this `forEach` call with `BuildersHelper.encodeValue(writer, options, codecRegistry)`, the `AggregatesSpecification.'should render $search'` test fails with `org.bson.BsonInvalidOperationException: writeStartDocument can only be called when State is INITIAL or VALUE or SCOPE_DOCUMENT or DONE, not when State is NAME`.

bson/src/main/org/bson/BsonNumber.java
```diff
@@ -23,7 +23,8 @@
  *
  * @since 3.0
  */
-public abstract class BsonNumber extends BsonValue {
+public abstract class
```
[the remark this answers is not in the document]
stIncMale: Done. Sorry, Ross also found a place where the space bar was pressed accidentally (though here it was Enter).

.evergreen/run-atlas-search-tests.sh
stIncMale: Configured a cluster for `AggregatesSearchIntegrationTest` here https://cloud-dev.mongodb.com/v2/5dd7496c7a3e5a648454341b#clusters and uncommented the Bash script. Let's see if the tests pass on Evergreen (they passed when I started them locally).

stIncMale: And [it worked](https://spruce.mongodb.com/task/mongo_java_driver_atlas_search_test_atlas_search_test_patch_3928344ab5738f62d563bbb6bc86191e82b1fa80_624b36097742ae73b4aae2c7_22_04_04_18_16_43/tests?execution=0&sortBy=STATUS&sortDir=ASC). Resolving.

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchFacet.scala
stIncMale: > Instant seems the appropriate type in scala

Thanks!, at least that part is done :)

> Just use: Iterable[Number]

In Java `Iterable<Double>` is not of type `Iterable<Number>`, but I would like to allow users to pass `Iterable<Double>` as well as any other `Iterable<N>` where `N extends Number`. This can be done in one of the following ways:

1. `<N extends Number> NumericSearchFacet numberFacet(..., Iterable<N> boundaries)`
2. `NumericSearchFacet numberFacet(..., Iterable<? extends Number> boundaries)`, which is almost exactly the asme thing as the first option, only instead of having of the type parameter `N`, we don't have any name and cannot refer to this type parameter in the body of the method.

Based on https://docs.scala-lang.org/tour/upper-type-bounds.html, my understanding is that in Scala 2 `Iterable[_ <: Number]` does exactly the same thing as #2 in Java. If that is correct, then using `Iterable[Number]` is not equivalent to `Iterable[_ <: Number]`, and is unjustifiably restrictive.

If I use `Iterable[Number]`, then `SearchFacet.numberFacet("a", SearchPath.fieldPath("numberFieldName"), Seq(10, 20.0, 30L))` indeed compiles. However, the following code does not compile:
```scala
val boundaries: Seq[scala.Double] = Seq(1.0, 2.0)
// `var boundaries = Seq(1.0, 2.0)` also fails with the same error
val facet: SearchFacet = SearchFacet.numberFacet("facetName", SearchPath.fieldPath("fieldName"), boundaries)
```
```
type mismatch;
 found   : Seq[Double]
 required: Iterable[Number]
```

However, the above code does not compile even if I use `Iterable[_ <: Number]`, which implies that "transparent" convertors from `collection.JavaConverters._` are not exactly "transparent" and break type expectations. Because of this, I will use `Iterable[Number]`, but the fact that users are not allowed to pass `Seq[scala.Double]`/`Seq[scala.Int]`/etc. is disheartening.

driver-core/src/main/com/mongodb/client/model/search/SearchFacet.java
```diff
@@ -0,0 +1,147 @@
...
+        Iterator<? extends SearchFacet> facetIterator = facets.iterator();
+        if (!facetIterator.hasNext()) {
+            throw new IllegalArgumentException("facets must not be empty");
+        }
+        BsonDocument result = new BsonDocument();
+        while (facetIterator.hasNext()) {
+            BsonDocument doc = facetIterator.next().toBsonDocument();
+            assertTrue(doc.size() == 1);
+            Map.Entry<String, BsonValue> entry = doc.entrySet().iterator().next();
+            result.append(entry.getKey(), entry.getValue());
```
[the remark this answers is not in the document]
stIncMale: That is a good idea, thanks. Done.

driver-core/src/main/com/mongodb/annotations/Evolving.java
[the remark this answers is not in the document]
jyemin: Ah, I see what you mean.  But I don't really see the point of linking to itself.  Wouldn't `{@code Evolving}` be sufficient (and also avoid the possibility of a line break after the `@`)?

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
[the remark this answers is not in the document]
jyemin: Even though the API will inevitably fall out of date as time passes, releasing an initial version of it that supports only a single operator will be confusing, since at the time of the release it won't even have methods all the current ones.  But I'm ok leaving it as is for this PR.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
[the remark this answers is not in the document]
jyemin: oh, because we're flattening the options into another document.  I guess I still find it confusing that SearchOptions implements Bson.

bson/src/main/org/bson/BasicBSONEncoder.java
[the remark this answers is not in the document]
jyemin: Maybe `org.bson.codecs.ValueCodecProvider`?  The issue is that the set of supported types is not fixed, but rather based on the configuration of the codec registry.

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchFacet.scala
rozza: There are quirks to scala and its compiler :)

Why is removing bounds helpful? Normally it wouldn't be if you were using Scala types - it would be the correct thing to do. However, `Number` is not a Scala type and it has its own Numeric types.

So usage becomes the key issue here and making it as simple as possible for the user.

```
val boundaries: Seq[Double] = Seq(10, 20.0, 30L)
```

Will never work as `scala.Double` is nothing to do with `Number`. So you force the user to cast to a `Number` type. Where as not setting the type inline will work as it allows the compiler to inspect the type and automatically cast to Java `Number`.

By restricting the types, you allow users boundaries inline eg:

```
SearchFacet.numberFacet("a", fieldPath("numberFieldName"), Seq(10, 20.0, 30L))
```

You are getting out of the users way. Other wise by setting the bounds you are forcing the user to declare their types, either inline or by values eg:

```
val boundaries: Seq[Number] = Seq(10, 20.0, 30L)
SearchFacet.numberFacet("a", fieldPath("numberFieldName"), boundaries)

val boundaries: Seq[java.lang.Double] = Seq(10, 20.0, 30L)
SearchFacet.numberFacet("a", fieldPath("numberFieldName"), boundaries)
```

The downside is you also force inline usage to declare the type:

```
SearchFacet.numberFacet("a", fieldPath("numberFieldName"), Seq[NUMBER TYPE NEEDED](10, 20.0, 30L))

```

So bounds are technically correct, as Scala has its own number types, it actually gets in the way
for users in all scenarios. So thats why I recommend not setting them.

jyemin (changes requested):
A few more comments, one of then particularly substantive I think.

driver-core/src/main/com/mongodb/internal/client/model/AbstractConstructibleBson.java
```diff
@@ -0,0 +1,96 @@
...
+    private final BsonDocument doc;
+
+    protected AbstractConstructibleBson(final Bson doc) {
+        this.doc = doc.toBsonDocument();
```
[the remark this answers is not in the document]
stIncMale: This issue is related to https://github.com/mongodb/mongo-java-driver/pull/891#discussion_r842255622. I agree, that's the right thing to do, and it appears to be doable. I will refactor `AbstractConstructibleBson`, `AbstractConstructibleBsonElement`, `SearchFacet.numberFacet`/`dateFacet`, `SearchPath.toBsonValue` and the places where it is used.

driver-core/src/main/com/mongodb/annotations/Evolving.java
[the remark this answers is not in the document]
stIncMale: Removed `{@code @}`, left only `{@link Evolving}`.

> But I don't really see the point of linking to itself.

The approach I am using has at least the following benefits compared to alternating between `@code` and `@link` based on the context:

- Tools can "understand" that the linked thing is a program element. Therefore, they can check that the referred program element is available (both Idea and `javadoc` can do that), furthermore tools can use this information when doing automatic refactoring.
-  Always using hyperlinks when mentioning program elements in documentation (provided that a link to the element is available) is a consistent way to mention program elements that does not require the author to write documentation differently based on where it is written and what elements are being mentioned. Also, this approach allows copying/moving documentation around without the need to check if in the new place some program elements need to be `@link`ed or `@code`d.

driver-legacy/src/main/com/mongodb/QueryBuilder.java
```diff
@@ -374,6 +376,7 @@ public QueryBuilder withinPolygon(final List<Double[]> points) {
      *
      * @param search the search terms to apply to the text index.
      * @return {@code this}
+     * @see Filters#text(String)
```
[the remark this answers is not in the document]
stIncMale: Rolled back.

driver-core/src/test/functional/com/mongodb/client/model/search/AggregatesSearchIntegrationTest.java
```diff
@@ -0,0 +1,211 @@
...
+                arguments(
+                        search(
+                                exists(fieldPath("tomatoes")),
+                                defaultSearchOptions()
+                                        .highlight(paths(
+                                                singleton(wildcardPath("pl*t")))
+                                                .maxCharsToExamine(100))
+                        ),
+                        asList(limit(1), project(metaSearchHighlights("highlights"))),
+                        Asserters.nonEmpty())
```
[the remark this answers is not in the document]
stIncMale: The type of the score is not specified in the docs. We could have tried inferring the type based on the type of values that can be used to [boost the score](https://www.mongodb.com/docs/atlas/atlas-search/scoring/#fields), but even there the type specified is "float", which is not a valid [BSON type](https://www.mongodb.com/docs/manual/reference/bson-types/). I assume that "float" in this context must be interpreted as a floating-point number, which according to [BSON specification](https://bsonspec.org/spec.html) may be either `double` or `decimal128 `.

Regardless of the type that the server uses in responses, the test may use any numeric type that can be retrieved from `BsonNumber`. Changed to `double`.

driver-core/src/test/functional/com/mongodb/client/model/search/AggregatesSearchIntegrationTest.java
```diff
@@ -0,0 +1,211 @@
...
+import static java.util.Arrays.asList;
+import static java.util.Collections.singleton;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+
+/**
+ * These tests require the following Atlas Search indices:
```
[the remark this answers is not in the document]
stIncMale: Yes, there is a requirement regarding the data in the namespace. I added it to the note.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -650,6 +656,46 @@ public static Bson sample(final int size) {
...
+     * @param options Optional {@code $search} pipeline stage fields.
+     * Specifying {@code null} is equivalent to specifying {@link SearchOptions#defaultSearchOptions()}.
+     * @return The {@code $search} pipeline stage.
+     *
+     * @mongodb.atlas.manual atlas-search/query-syntax/#-search $search
+     * @mongodb.atlas.manual atlas-search/operators-and-collectors/#operators Search operators
+     * @mongodb.atlas.manual atlas-search/scoring/ Scoring
+     * @since 4.6
+     */
+    public static Bson search(final SearchOperator operator, @Nullable final SearchOptions options) {
```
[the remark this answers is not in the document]
stIncMale: Method overloads are nice, but only if the documentation does not have to be duplicated. In our case, the decision was to always duplicate the docs, which I see as being harmful to both us (we pollute the sources with duplicates that are problematic to maintain in a consistent state) and users (the user has no way of quickly seeing what is the same and what is different in the documentation of two program elements when that documentation is either fully duplicated, or is different only for a single parameter, as would have been in this case).

This is the reason I do not want to create overloads here.

driver-core/src/main/com/mongodb/internal/client/model/AbstractConstructibleBson.java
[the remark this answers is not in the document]
stIncMale: Changing all new types to be rendered into `BsonDocument` lazily, which allows us to wait and use `documentClass` and `codecRegistry` once they are available, is not exactly straightforward, and I have not completed it yet.

Also, this cannot be done for `SearchPath` at all because we can only render lazily `Bson`, and turns out, that using the form `"fieldPath"` (BSON String) instead of `{"value": "fieldPath"}` is mandatory, not optional. This means that `SearchPath` is not `Bson` (some of its forms are documents, but some are just strings), i.e., cannot be rendered lazily.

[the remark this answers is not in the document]
stIncMale: Done, everything is now encoded to `BsonDocument` lazily except for `SearchPath`, because it can't always be represented as a document (sometimes it must be represented as a string), and the driver's codec system supports lazy encoding only for documents via `Bson`.

bson/src/main/org/bson/BasicBSONEncoder.java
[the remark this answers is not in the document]
stIncMale: Rolled back the changes related to `BasicBSONEncoder`. The search API does not need any of that anymore because it now utilizes the codec system and does not need to know how to encode `Number`.

driver-core/src/main/com/mongodb/client/model/search/StringSearchFacet.java
stIncMale: Ross, thanks for the explanations. Resolving this discussion as my concerns are addressed now.

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchCollector.scala
```diff
@@ -0,0 +1,85 @@
...
+   *    Document("operator" -> SearchOperator.exists(
+   *      SearchPath.fieldPath("fieldName")).toBsonDocument,
+   *      "facets" -> SearchFacet.combineToBson(Seq(
+   *        SearchFacet.stringFacet(
+   *          "stringFacetName",
+   *          SearchPath.fieldPath("stringFieldName")),
+   *        SearchFacet.numberFacet(
+   *          "numberFacetName",
+   *          SearchPath.fieldPath("numberFieldName"),
+   *          Seq(10, 20, 30)))).toBsonDocument)))
```
stIncMale, line 77: Looks like Scala's `Document` does not really allow to append `Bson`, which is why I have to call `toBsonDocument` twice in this example, thus encoding `Bson` eagerly without having access to user-specified codecs. My understanding is that there is no other way for Scala API, but @rozza, could you please confirm this is indeed the case?

stIncMale:
> One other observation is the examples in the java docs use BsonDocument do you think using Document would be better - its our default MongoCollection type so more users will be familiar with that.

@rozza I changed all the examples and corresponding tests such that they use `Document`.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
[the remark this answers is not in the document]
rozza: I would go for consistency with the CRUD API and add the overload. 

This API is about being easier for users, by being less verbose than just using documents. Being convenient for users is important and we'll have to handle the extra documentation.

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchCollector.scala
rozza: The `Document` class is a scala wrapper for `BsonDocument` and as such is type safe and only allows {{BsonValue}} values to be added.

stIncMale: Thank you, Ross, for confirming.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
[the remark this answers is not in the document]
stIncMale: > Being convenient for users is important 

I am arguing that by copying documentation in multiple places we are not providing convenient documentation for users, because users have not quick and easy way to see the difference / equivalence of multiple methods that copy each other's documentation and potentially (but not necessary) change something there.

> we'll have to handle the extra documentation

Not only maintaining duplicated documentation is resource-consuming, but we also cannot preserve the integrity: if there is no mechanism that enforces it (and there is none), it will eventually be violated. When users look at two seemingly similar methods such that for one of them we updated the docs, but for the other we forgot, it creates a new problem for the users: which one of the methods has the correct documentation?

Since both you and Jeff requested the change, I will go ahead and introduce 2 more copies to the existing 2 copies.

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchCollector.scala
```diff
@@ -0,0 +1,85 @@
...
+   * for each of those groups.
+   *
+   * @param operator A search operator to use.
+   * @param facets Non-empty facet definitions.
+   * @return The requested `SearchCollector`.
+   * @see [[https://www.mongodb.com/docs/atlas/atlas-search/facet/ facet collector]]
+   */
+  @Beta
+  def facet(operator: SearchOperator, facets: Iterable[_ <: SearchFacet]): FacetSearchCollector =
+    JSearchCollector.facet(operator, facets.asJava)
```
stIncMale, line 46: After [fixing](https://github.com/mongodb/mongo-java-driver/pull/891/commits/50eee20077f35b308d9a9d800e63906034fe0299) another problem with Scala compiler, it now does not like having static methods in interfaces: _"SearchCollector.scala:46:27: Static methods in interface require -target:jvm-1.8"_. The error happens only for Scala 2.11 (`./gradlew -PscalaVersion=2.11.12 --stacktrace --info :bson-scala:test :driver-scala:test :driver-scala:integrationTest`). I tried specifying
```gradle
if(scalaVersion.startsWith("2.11")) {
    scalaCompileOptions.additionalParameters = [
            "-target:jvm-1.8"
    ]
}
```
in `build.gradle` inside the `tasks.withType(ScalaCompile) { ... }` block that we already have, but it did not help. @rozza, can you think of what I did wrong and what is the right way to pass `-target:jvm-1.8` to the Scala compiler?

driver-core/src/main/com/mongodb/internal/client/model/AbstractConstructibleBson.java
```diff
@@ -70,27 +81,39 @@ public final boolean equals(final Object o) {
...
 
     @Override
     public String toString() {
-        return doc.toString();
+        return "{base=" + base
+                + ", appended=" + appended
+                + '}';
+    }
+
+    private static BsonDocument newAppended(final BsonDocument base, final BsonDocument appended) {
```
jyemin, line 99: I initially got confused by the existence of a `newAppended` non-static method and this new static one.  Can you think of a different name for one of them?  They seem semantically different so this kind-of overload is confusing.

jyemin (approved):
LGTM (minus the Scala API, which I didn't review)

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchCollector.scala
rozza: Investigating

driver-core/src/main/com/mongodb/internal/client/model/AbstractConstructibleBson.java
stIncMale: Renamed to `newMerged`.

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchCollector.scala
stIncMale: Thank you, Ross! I could have realized that it is worth checking if there is something in configuration of other modules that may overwrite parameters specified in `build.gradle`. Alas!

At first I tried moving the code you quoted from `drivers-scala/build.gradle` to `build.gradle`, but it resulted in failures of the `MacrosSpec.support tagged types in case classes` test. Because of this, I had to leave the code in `drivers-scala/build.gradle`, but change it so that it adds to parameters, instead of overwriting them. I added a similar piece of code to `build.gradle` in order to specify `-target:jvm-1.8`: it seems useful to allow static methods in interfaces for all our Scala code, not just that in the `drivers-scala` module.

PR #892, opened 2022-03-13
Introduce DnsResolver SPI

gastaldi:
DefaultDnsResolver is now a singleton and DnsResolverFactory was introduced

Fixes JAVA-4512

### DIFF
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -18,9 +18,10 @@ package com.mongodb;
 
 import com.mongodb.connection.ClusterSettings;
 import com.mongodb.connection.ConnectionPoolSettings;
+import com.mongodb.connection.dns.DnsResolverFactory;
 import com.mongodb.diagnostics.logging.Logger;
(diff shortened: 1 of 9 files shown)


jyemin:
Hi @gastaldi thanks for getting the ball rolling on this. I've had a quick look and I see a couple of things that I think we can improve on.

# There is a fair amount of MongoDB-specific behavior baked into the current `DefaultDnsResolver`, and we've had more than one bug/feature request reported against it that have required changes to its implementation and even its API.  I would like to avoid making change requests like those cascade down into alternate implementations of `DnsResolver`.  Instead, I think we should define a lower-level provider API that encapsulates only the functionality currently provided by JNDI.
# As mentioned in JAVA-4512, the driver could leverage the JDK's `ServiceLoader` API to look up alternative implementations of DNS for the driver to use instead of defining its own domain-specific mechanism.

I took a stab at implementing these ideas in [this PR](https://github.com/mongodb/mongo-java-driver/pull/893). Let me know if you think it would suit your needs.

gastaldi:
@jyemin that looks like a better (or less breaking) solution. I'll close this PR on behalf of that one 👍🏻

PR #896, opened 2022-03-21
Added support for comment for operations

rozza:
JAVA-4369


driver-core/src/main/com/mongodb/client/model/InsertManyOptions.java
```diff
@@ -65,6 +68,8 @@ public Boolean getBypassDocumentValidation() {
     /**
      * Sets the bypass document level validation flag.
      *
+     * <p>For bulk operations use: {@link BulkWriteOptions#bypassDocumentValidation(Boolean)}</p>
```
rozza, line 71: Took the opportunity to document this is ignored in bulk operations.

driver-core/src/main/com/mongodb/client/model/BulkWriteOptions.java
```diff
@@ -76,11 +79,50 @@ public BulkWriteOptions bypassDocumentValidation(@Nullable final Boolean bypassD
...
+
+    /**
+     * Sets the comment for this operation. A null value means no comment is set.
+     *
+     * @param comment the comment
+     * @return this
+     * @since 4.6
+     * @mongodb.server.release 4.4
+     */
+    public BulkWriteOptions comment(@Nullable final String comment) {
```
rozza, line 102: Opted to add both `comment(String)` and `comment(BsonValue)` throughout - to keep the api's the consistent. Also, I took the opinion that most users won't be familiar with BsonValues.

driver-core/src/main/com/mongodb/internal/operation/AggregateOperation.java
```diff
@@ -208,23 +208,37 @@ public Collation getCollation() {
     /**
      * Returns the comment to send with the aggregate. The default is not to include a comment with the aggregation.
      *
+     * <p>The comment can be any valid BSON type for server versions 4.4 and above.
```
jyemin, line 211: This seems like it was written for public consumption, but this is an internal class.

Can you clarify your intent?

driver-core/src/main/com/mongodb/internal/operation/FindAndUpdateOperation.java
```diff
@@ -70,6 +71,7 @@
     private List<BsonDocument> arrayFilters;
     private Bson hint;
     private String hintString;
+    private BsonValue comment;
```
jyemin, line 74: Could this go in BaseFindAndModifyOperation?

driver-core/src/main/com/mongodb/internal/operation/OperationHelper.java
```diff
@@ -479,15 +480,14 @@ static boolean canRetryRead(final ServerDescription serverDescription, final Con
                                                           final ConnectionSource source, final Connection connection, final int batchSize) {
         return new QueryBatchCursor<T>(OperationHelper.<T>cursorDocumentToQueryResult(cursorDocument,
                 source.getServerDescription().getAddress()),
-                0, batchSize, 0, decoder, source, connection);
+                0, batchSize, 0, decoder, null, source, connection);
```
jyemin, line 483: Why `null` instead of `comment`, like it is for async just below?

driver-core/src/main/com/mongodb/internal/operation/FindAndUpdateOperation.java
rozza: I followed the existing conventions - like with hint

driver-core/src/main/com/mongodb/internal/operation/OperationHelper.java
rozza: Good catch :+1:

driver-core/src/main/com/mongodb/internal/operation/AggregateOperation.java
rozza: Happy to remove if required.

driver-core/src/main/com/mongodb/internal/operation/FindAndUpdateOperation.java
rozza: Added JAVA-4544 which I'll do after this ticket :+1:

driver-core/src/main/com/mongodb/internal/operation/AggregateOperation.java
rozza: I had it to document the behavior of strings vs other types. Will remove.

rozza:
Note some tests are failing - I am speaking with the authors as it seems the expected pre 4.4 failures don't fail on 3.2, 3.0 and 2.6

driver-core/src/main/com/mongodb/internal/operation/FindAndUpdateOperation.java
jyemin: No worries.

jyemin (commented):
I'm unable to comment on some of the threads.  I don't know why.

I'll wait to LGTM until we figure out why tests are failing.

driver-core/src/main/com/mongodb/internal/operation/OperationHelper.java
jyemin: Can I deduce that there is no test coverage for this scenario?

rozza: No - it would be for a getMore on a list collection / list indexes operation.

jyemin (approved):
LGTM if tests are all passing.

I don't think this needs a second review.

PR #898, opened 2022-03-22
Reduced duplication in the FindAndModify operations

rozza:
JAVA-4544

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/BaseFindAndModifyOperation.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/BaseFindAndModifyOperation.java
@@ -18,23 +18,34 @@ package com.mongodb.internal.operation;
 
 import com.mongodb.MongoNamespace;
 import com.mongodb.WriteConcern;
-import com.mongodb.internal.async.SingleResultCallback;
+import com.mongodb.client.model.Collation;
 import com.mongodb.connection.ConnectionDescription;
-import com.mongodb.connection.ServerDescription;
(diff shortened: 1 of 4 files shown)


driver-core/src/main/com/mongodb/internal/operation/BaseFindAndModifyOperation.java
```diff
@@ -122,22 +140,231 @@ public boolean isRetryWrites() {
...
+     * @mongodb.server.release 4.4
+     */
+    public BaseFindAndModifyOperation<T> comment(final BsonValue comment) {
+        this.comment = comment;
+        return this;
     }
 
     protected abstract FieldNameValidator getFieldNameValidator();
+
+    protected abstract void specializeCommand(BsonDocument initialCommand, ConnectionDescription connectionDescription);
```
jyemin, line 330: Nit: since this is an internal API, it's safe to remove the currently-unused `connectionDescription` parameter.

jyemin: You could also make it not abstract since 2 of 3 subclasses don't require it.

jyemin (approved):

driver-core/src/main/com/mongodb/internal/operation/BaseFindAndModifyOperation.java
jyemin: Never mind.  I mis-read the diffs and thought that two of the three implementations were empty methods.

PR #899, opened 2022-03-23
Ensure getMore comments only sent if server >= 4.4.0

rozza:
JAVA-4507
DRIVERS-2214
DRIVERS-2243


rozza:
These tests or similar will eventually make it into the specs repo, but I thought it best to get ahead and implement the changes sooner rather than later.

jyemin (approved):
LGTM

PR #900, opened 2022-03-25
Added support for let in operations

rozza:
JAVA-4366

Note: I followed the existing convention in naming the let variable as `variables`.


jyemin (commented):
You haven't requested review yet, but I noticed it in GitHub so took a look.  It looks good to me.

rozza:
Thanks I was just waiting on the test suite to run before requesting the review :)

PR #865, opened 2022-02-01
Check out an implicit session only after checking out a connection

jyemin:
JAVA-4365

PR #903, opened 2022-03-28
Add client session re-use prose test

jyemin:
JAVA-4531

This prose test was originally reviewed as part of the [JAVA-4365 PR](https://github.com/mongodb/mongo-java-driver/pull/865/commits/49f6b172a3fd390589e375d9af73b57277e8376f), but was then removed as it was found to be flaky.  The updated prose test is intended to minimize the flakiness of the previous test.

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/AbstractSessionsProseTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/AbstractSessionsProseTest.java
@@ -26,7 +26,7 @@ import org.bson.BsonDocument;
 import org.bson.Document;
 import org.junit.jupiter.api.Test;
 
-import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
@@ -36,6 +36,8 @@ import java.util.concurrent.TimeUnit;
 import static com.mongodb.ClusterFixture.getConnectionString;
 import static com.mongodb.ClusterFixture.getDefaultDatabaseName;
 import static com.mongodb.ClusterFixture.serverVersionAtLeast;
+import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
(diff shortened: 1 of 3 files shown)


driver-sync/src/test/functional/com/mongodb/client/AbstractSessionsProseTest.java
```diff
@@ -0,0 +1,111 @@
...
+                        @Override
+                        public void commandStarted(final CommandStartedEvent event) {
+                            lsidSet.add(event.getCommand().getDocument("lsid"));
+                        }
+                    })
+                    .build());
+            MongoCollection<Document> collection = client.getDatabase(getDefaultDatabaseName()).getCollection(getClass().getName());
+
+            // when executing 8 operations concurrently
+            ExecutorService executor = Executors.newFixedThreadPool(3);
```
stIncMale, line 71: Is there a reason we don't have 8 threads here? Using 3 with a synchronous driver means that we definitely cannot have all 8 operations executing concurrently.

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/client/AbstractSessionsProseTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @Test
+    public void shouldCreateServerSessionOnlyAfterConnectionCheckout() throws InterruptedException {
+        assumeTrue(serverVersionAtLeast(3, 6));
+
+        int minLsidSetSize = Integer.MAX_VALUE;
+
+        for (int i = 0; i < 5; i++) {
+            // given
+            Set<BsonDocument> lsidSet = ConcurrentHashMap.newKeySet();
+            MongoClient client = getMongoClient(MongoClientSettings.builder()
```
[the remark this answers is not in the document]
jyemin: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractSessionsProseTest.java
jyemin: No, that is a bug. Done

jyemin:
While fixing the issues that you mentioned, I found a number of other improvements to aid in code clarity and prevent future bugs from being introduced.

* I realized there is no need for a new MongoClient each time.  Even if one iteration fails to create only one session and the server session pool has more than one item in it, the pool is still LIFO so on subsequent iterations the same session should be pulled out
* I created a `List<Runnable>` and use its size to determine the number of concurrent operations, rather than just assuming there are 8 jobs submitted
* I break out of the loop as soon as I know the assertion will succeed.

stIncMale (approved):
👏 
Not sure how I feel about depending on the LIFO behavior, but reusing the client must noticeably speed the test up.

PR #906, opened 2022-03-31
Tolerate `DefaultConnectionPool.invalidate` being called after or concurrently with `close`

stIncMale:
JAVA-4551

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -1417,7 +1417,7 @@ class DefaultConnectionPool implements ConnectionPool {
         void start() {
             if (maintainer != null) {
                 assertTrue(cancellationHandle == null);
-                cancellationHandle = doIgnoringRejectedExecution(() -> maintainer.scheduleAtFixedRate(
+                cancellationHandle = ignoreRejectedExectution(() -> maintainer.scheduleAtFixedRate(
                         DefaultConnectionPool.this::doMaintenance,
                         initialStart ? settings.getMaintenanceInitialDelay(MILLISECONDS) : 0,
                         settings.getMaintenanceFrequency(MILLISECONDS), MILLISECONDS));
@@ -1431,7 +1431,7 @@ class DefaultConnectionPool implements ConnectionPool {
                     cancellationHandle.cancel(false);
                     cancellationHandle = null;
                 }
-                doIgnoringRejectedExecution(() -> maintainer.execute(DefaultConnectionPool.this::doMaintenance));
+                ignoreRejectedExectution(() -> maintainer.execute(DefaultConnectionPool.this::doMaintenance));
             }
         }
 
@@ -1442,15 +1442,15 @@ class DefaultConnectionPool implements ConnectionPool {
             }
         }
 
-        private void doIgnoringRejectedExecution(final Runnable action) {
-            doIgnoringRejectedExecution(() -> {
+        private void ignoreRejectedExectution(final Runnable action) {
+            ignoreRejectedExectution(() -> {
                 action.run();
                 return null;
             });
         }
 
         @Nullable
-        private <T> T doIgnoringRejectedExecution(final Supplier<T> action) {
+        private <T> T ignoreRejectedExectution(final Supplier<T> action) {
             try {
                 return action.get();
             } catch (RejectedExecutionException ignored) {
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -69,6 +69,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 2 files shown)


driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1437,6 +1441,23 @@ public void close() {
                 maintainer.shutdownNow();
             }
         }
+
+        private void doIgnoringRejectedExecution(final Runnable action) {
```
rozza, line 1445: This method name is hard to read - "do ignoring rejected execution" doesn't read right. Can you think of a clearer name for it? `ignoreRejectedExectution` seems clearer in intent and is more readable.

rozza (approved):
LGTM, only one naming nit which might just be me.

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
```diff
@@ -443,6 +443,41 @@ public void checkoutHandOverMechanism() throws InterruptedException, TimeoutExce
...
+                mockSdamProvider());
+        provider.ready();
+        provider.close();
+        provider.invalidate(null);
+    }
+
+    @Test
+    public void invalidateConcurrentWithCloseMustNotThrow() throws ExecutionException, InterruptedException {
+        Future<?> backgroundTask = null;
+        for (int i = 0; i < 10_000; i++) {
```
jyemin, line 461: Does the number of iterations need to be quite so high to reliably generate a failure if there is a regression?  I was able to get it to fail (after commenting out the bug fix) with just one iteration.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1416,18 +1417,21 @@ private BackgroundMaintenanceManager() {
...
                 initialStart = false;
             }
         }
 
         void runOnceAndStop() {
             if (maintainer != null) {
-                assertNotNull(cancellationHandle).cancel(false);
-                cancellationHandle = null;
-                maintainer.execute(DefaultConnectionPool.this::doMaintenance);
+                if (cancellationHandle != null) {
```
jyemin, line 1430: What's the execution path where cancellationHandle can be null?  I'm not sure why it's no longer an assertion.

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
```diff
@@ -443,6 +443,41 @@ public void checkoutHandOverMechanism() throws InterruptedException, TimeoutExce
...
+                connectionFactory,
+                ConnectionPoolSettings.builder().maxSize(1).build(),
+                mockSdamProvider());
+        provider.ready();
+        provider.close();
+        provider.invalidate(null);
+    }
+
+    @Test
+    public void invalidateConcurrentWithCloseMustNotThrow() throws ExecutionException, InterruptedException {
```
jyemin, line 459: I don't understand how this method tests that both start _and_ runOnceAndStop wrap their calls to execute in `doIgnoringRejectedExecution`.  I proved to myself that it does (since if I comment out either one the test fails), but I don't understand why it's the case.  Can you explain?

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
```diff
@@ -443,6 +443,41 @@ public void checkoutHandOverMechanism() throws InterruptedException, TimeoutExce
...
+        for (int i = 0; i < 10_000; i++) {
+            provider = new DefaultConnectionPool(
+                    SERVER_ID,
+                    connectionFactory,
+                    ConnectionPoolSettings.builder().maxSize(1).build(),
+                    mockSdamProvider());
+            try {
+                backgroundTask = cachedExecutor.submit(() -> {
+                    provider.ready();
+                    provider.invalidate(null);
```
jyemin, line 470: Does this line just increase the likelihood of failure in the case of a regression?  I was able to get the test to fail (after commenting out the bug fix, that is) after one iteration even with it commented out.

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
stIncMale: `BackgroundMaintenanceManager.start` is called by `DefaultConnectionPool.ready`, and `BackgroundMaintenanceManager.runOnceAndStop` is called by `DefaultConnectionPool.invalidate`. However, this is not a very important implementation detail. The `invalidateConcurrentWithCloseMustNotThrow` test checks that neither `ready` nor `invalidate` throws an exception when are called concurrently with `close`, because this is the behavior we care about. It does so by running them concurrently and checking that `ready`/`invalidate` do not throw by calling `backgroundTask.get()`. I should rename `backgroundTask` to `readyAndInvalidateResult`.

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
stIncMale: This test checks that nether `ready` nor `invalidate` throw when called concurrently with `close`. Renamed the test to `readyInvalidateConcurrentWithCloseMustNotThrow`.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: Done.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
stIncMale: When `BackgroundMaintenanceManager.start` is called concurrently with `close`, it may fail to schedule `doMaintenance`, writing `null` to `cancellationHandle` as a consequence. `BackgroundMaintenanceManager.runOnceAndStop` called after `start` (these methods must be called sequentially as per the documentation of `BackgroundMaintenanceManager`) observes `null` in `cancellationHandle`.

jyemin (approved):

PR #908, opened 2022-04-06
Add unified change stream test for comment

jyemin:
Requires support for absent saveResultAsEntity, which is optional

JAVA-4369

This test was missed in the initial PR for comment support, and requires a few changes in the runner

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
@@ -91,7 +91,7 @@ import static java.util.stream.Collectors.toList;
 
 final class UnifiedCrudHelper {
     private final Entities entities;
-    private final AtomicInteger uniqueId = new AtomicInteger();
+    private final AtomicInteger uniqueIdGenerator = new AtomicInteger();
 
     private final Codec<ChangeStreamDocument<BsonDocument>> changeStreamDocumentCodec = ChangeStreamDocument.createCodec(
             BsonDocument.class,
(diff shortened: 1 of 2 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -226,7 +229,8 @@ OperationResult executeFind(final BsonDocument operation) {
     OperationResult createFindCursor(final BsonDocument operation) {
         FindIterable<BsonDocument> iterable = createFindIterable(operation);
         return resultOf(() -> {
-            entities.addCursor(operation.getString("saveResultAsEntity").getValue(), iterable.cursor());
+            entities.addCursor(operation.getString("saveResultAsEntity", new BsonString(createRandomEntityId())).getValue(),
```
jyemin, line 232: The advantage of storing the cursor in `entities` with a random identifier is that it will ensure that the cursor is closed at the end of the test.

rozza (approved):
LGTM

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -90,6 +91,8 @@
 
 final class UnifiedCrudHelper {
     private final Entities entities;
+    private final AtomicInteger uniqueId = new AtomicInteger();
```
stIncMale, line 94: [optional]
It's an ID generator, not an ID. Let's rename it to facilitate readability.

stIncMale (approved):
Just a small optional thing. Please feel free to merge either as is, or after doing the proposed change.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
jyemin: Done

PR #911, opened 2022-04-08
Add Java record support to the default registry

rozza:
Add dependency in core to bson-record-codec
Wrap the RecordCodecProvider with a canary check so support is only added if the JDK version supports it.

JAVA-3567

### DIFF
--- a/driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
+++ b/driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
@@ -33,7 +33,7 @@ import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
 public class Jep395RecordCodecProvider implements CodecProvider {
 
     @Nullable
-    private static final CodecProvider recordCodecProvider;
+    private static final CodecProvider RECORD_CODEC_PROVIDER;
     static {
 
         CodecProvider possibleCodecProvider;
@@ -44,18 +44,18 @@ public class Jep395RecordCodecProvider implements CodecProvider {
             // No JEP-395 support
             possibleCodecProvider = null;
         }
-        recordCodecProvider = possibleCodecProvider;
+        RECORD_CODEC_PROVIDER = possibleCodecProvider;
     }
 
     @Override
     @Nullable
     public <T> Codec<T> get(final Class<T> clazz, final CodecRegistry registry) {
-        return recordCodecProvider != null ? recordCodecProvider.get(clazz, registry) : null;
+        return RECORD_CODEC_PROVIDER != null ? RECORD_CODEC_PROVIDER.get(clazz, registry) : null;
     }
 
     @VisibleForTesting(otherwise = PRIVATE)
     public boolean hasRecordSupport() {
-        return recordCodecProvider != null;
+        return RECORD_CODEC_PROVIDER != null;
     }
 }
 
--- a/driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
+++ b/driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
@@ -29,6 +29,8 @@ import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
  * A CodecProvider for Java Records.
  *
  * <p>Requires java.lang.Record support - eg Java 17 or greater.</p>
+ *
+ * @since 4.6
  */
 public class Jep395RecordCodecProvider implements CodecProvider {
 
--- a/driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
+++ b/driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


driver-core/src/test/unit/com/mongodb/Jep395RecordCodecProviderTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class Jep395RecordCodecProviderTest {
+
+    @Test
+    @EnabledForJreRange(min = JRE.JAVA_17)
+    void canSupportJavaRecordsOnJava17() {
+        assertTrue(new Jep395RecordCodecProvider().hasRecordSupport());
```
rozza, line 31: Testing: 

```
        Jep395RecordCodecProvider jep395RecordCodecProvider = new Jep395RecordCodecProvider();
        try {
            Class<?> recordClass = Class.forName("java.lang.Record");
            assertNotNull(jep395RecordCodecProvider.get(recordClass, MongoClientSettings.getDefaultCodecRegistry()));
        } catch (ClassNotFoundException e) {
            fail("Should not fail on JAVA 17");
        }
```

Failed because using `recordClass`always returned null  - it failed the clazz.isRecord in the actual `RecordCodecProvider`.

jyemin: Oh, interesting.

jyemin (approved):
So the only thing we can't test is that Jep395RecordCodecProvider is actually added to the default codec registry.  Unfortunate, but for now I think it's good enough.

LGTM

driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
```diff
@@ -0,0 +1,61 @@
...
+
+import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
+
+
+/**
+ * A CodecProvider for Java Records.
+ *
+ * <p>Requires java.lang.Record support - eg Java 17 or greater.</p>
+ */
+public class Jep395RecordCodecProvider implements CodecProvider {
```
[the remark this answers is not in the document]
cescoffier: Yes, we would need to have version specific substitutions to remove that class on java 11. Also, while  there is work in progress, records are not native friendly yet.

driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
```diff
@@ -0,0 +1,61 @@
...
+import org.bson.codecs.record.RecordCodecProvider;
+
+import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
+
+
+/**
+ * A CodecProvider for Java Records.
+ *
+ * <p>Requires java.lang.Record support - eg Java 17 or greater.</p>
+ */
```
jyemin, line 32: I was going to suggest that this class be non-public, but given @cescoffier's comment below, it might be best if it's part of the API.  

If that's the case, add an `@since 4.6` to the Javadoc

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -73,7 +73,8 @@
                     new Jsr310CodecProvider(),
                     new JsonObjectCodecProvider(),
                     new BsonCodecProvider(),
-                    new EnumCodecProvider()));
+                    new EnumCodecProvider(),
+                    new Jep395RecordCodecProvider()));
```
jyemin, line 77: This should be added to the `getDefaultCodecRegistry` Javadoc below.

driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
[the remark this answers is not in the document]
jyemin: @christophstrobl, @mp911de, this may be a Spring Boot issue as well.

It's not clear to me what the way forward is for this and similar issues.  Do all of these substitutions (the one for DNS provider as well) have to done separately for every framework that supports native images?

jyemin (changes requested):

driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
rozza: Done

driver-core/src/main/com/mongodb/MongoClientSettings.java
rozza: Done

jyemin (approved):
LGTM!

driver-core/src/main/com/mongodb/Jep395RecordCodecProvider.java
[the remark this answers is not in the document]
cescoffier: Yeah, BUT it's precisely what we want to avoid doing. We want to know almost everything at build time to optimize and initialize during the build and not during the runtime. I know it can be seen as a different way to execute Java (it is), but the benefits are quite high.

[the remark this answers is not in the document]
rozza: Thanks for the clarification @cescoffier :+1:

PR #912, opened 2022-04-08
Various changes required to update unified change stream tests

jyemin:
I'm doing all these as one PR because the unified change stream tests are what required the first two changes, but it was difficult to split up do to the overlapping commits to the unified tests.  So submitting this as together but will not squash the commits.

JAVA-4565: this one was necessary because one of the new tests ends up throwing a MongoQueryException instead of a MongoCommandException, and MongoQueryException didn't have the errorCodeName property needed to make assertion.

JAVA-4470: this one is not fully done.  It requires a bunch of things if we want to use ChangeStreamDocument for all the unified tests.  The one that's now handled properly is an operationType for which there is no associated enum value.  But for a couple of the new tests we're escaping to BsonDocument for the entire change stream document so that the assertions pass.

JAVA-4555: ChangeStreamDocument already handles the "to" field, so nothing needed to be done there.  So it's just updates to the unified tests.

### DIFF
--- a/driver-core/src/test/resources/unified-test-format/change-streams/change-streams.json
+++ b/driver-core/src/test/resources/unified-test-format/change-streams/change-streams.json
@@ -3,6 +3,7 @@
   "schemaVersion": "1.0",
   "runOnRequirements": [
     {
+      "minServerVersion": "3.6",
       "topologies": [
         "replicaset",
         "sharded-replicaset"
@@ -168,7 +169,6 @@
       "description": "Test with document comment - pre 4.4",
       "runOnRequirements": [
         {
-          "minServerVersion": "3.6.0",
           "maxServerVersion": "4.2.99"
         }
       ],
@@ -213,11 +213,6 @@
     },
     {
       "description": "Test with string comment",
-      "runOnRequirements": [
-        {
-          "minServerVersion": "3.6.0"
-        }
-      ],
       "operations": [
         {
           "name": "createChangeStream",
@@ -347,7 +342,6 @@
       "description": "Test that comment is not set on getMore - pre 4.4",
       "runOnRequirements": [
         {
-          "minServerVersion": "3.6.0",
           "maxServerVersion": "4.3.99",
           "topologies": [
             "replicaset"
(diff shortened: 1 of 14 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -1182,4 +1197,72 @@ public OperationResult executeEstimatedDocumentCount(final BsonDocument operatio
...
         return "random-entity-id" + uniqueIdGenerator.getAndIncrement();
     }
+
+    /**
+     * The tests in this list can not currently pass when using {@link ChangeStreamDocument} because there is information less when
+     * decoding into an instance of that class.  So for the offending tests, we just decode directly into {@link BsonDocument}.  For all
+     * others, we decode into {@link ChangeStreamDocument} and from there to {@link BsonDocument} so that there are integration tests of
+     * {@link ChangeStreamDocument}.
+     */
+    private static final List<String> OFFENDING_CHANGE_STREAM_TESTS = asList(
```
rozza, line 1207: Nit: Possibly change to `BSON_DOCUMENT_CHANGE_STREAM_TESTS` - not sure they are offending me ;)

rozza (approved):
LGTM

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
jyemin: OFFENDING not OFFENSIVE ;)

rozza: I know but in colloquial English it amounts to the same thing 🇬🇧

rozza (approved):

stIncMale (approved):

jyemin:
Now that PRs are executing a larger set of variants, I'm seeing many of them fail.  

I will investigate and update the PR when I understand the root cause(s).

stIncMale (approved):
Despite the force pushes, GitHub showed me that the only file I need to review / re-review is `SyncMongoCollection.java`, and that's what I did.

jyemin:
Tests are still failing, so this can't be merged yet.

jyemin:
Tests are finally green, due to two changes:

1. `enableSleepAfterCursorOpen` is called for all change stream tests.  This fixes an issue in many of the tests that are relying on the change stream being established before executing a crud operation, which is then expected to generate a change stream event. But if the crud operation ends up executing before the change stream is established, then it will not generate an event.  By sleeping for a bit, we decrease the chances (hopefully close to zero) that the change stream is established first
2. Skip one test that is failing on latest sharded clusters, due to SERVER-65497.  This can be reverted once the fix for the sever issue makes it into the latest build.

driver-core/src/test/resources/unified-test-format/change-streams/change-streams.json
```diff
@@ -559,6 +559,21 @@
...
+        },
+        {
+          "minServerVersion": "3.6",
+          "maxServerVersion": "5.0",
+          "topologies": [
+            "sharded-replicaset"
+          ]
+        }
+      ],
       "operations": [
```
stIncMale, line 577: This code is different from the one in the spec repo: https://github.com/mongodb/specifications/blob/master/source/change-streams/tests/unified/change-streams.json#L622-L623

Is there a PR pending to be merged there, and the change was taken from that PR, or is it a temporary change that we will have to undo once the latest server is fixed? If it is the latter, can we have a Jira task reminding us about undoing this change?

jyemin: It's a temporary change that should be undone once [SERVER-65497](https://jira.mongodb.org/browse/SERVER-65497) issue is fixed.  Opened [JAVA-4579](https://jira.mongodb.org/browse/JAVA-4579) to track

stIncMale: Thanks!

jyemin:
Merged on command line.

PR #914, opened 2022-04-11
Update REAME.md to adapt the latest version of IDEA

luozhenyu:
# Description
1. Replace "Settings" with "Settings/Preferences"
   * Reference: https://www.jetbrains.com/help/idea/settings-preferences-dialog.html
2. Update the procedure of generating a build config
   * Reference: https://www.jetbrains.com/help/idea/work-with-gradle-projects.html#delegate_build_gradle


rozza:
Thanks @luozhenyu

PR #915, opened 2022-04-11
Update the requirements to build and compile the source

luozhenyu:
1. Java 17+ is required rather than Java 9
   * Reference: https://github.com/mongodb/mongo-java-driver/blob/master/gradle/javaToolchain.gradle#L16
2. git is required
   * Reference: https://github.com/mongodb/mongo-java-driver/blob/master/build.gradle#L334

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Example for Maven:
 
 ## Build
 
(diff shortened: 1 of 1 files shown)


rozza:
Thanks @luozhenyu :)

PR #916, opened 2022-04-11
Fix or skip new failing unified change stream tests

jyemin:
* Add support for ignoreExtraEvents in unified test runner. (note: this is a new option and not yet out of spec review, but looks like this is what it's going to look like)
* Update change stream tests to specify ignoreExtraEvents
* Update reactive change stream test runner to skip tests where extra getMore events occur between required events
* Update reactive change stream test runner to skip tests that require change stream initialization to report an error
* Enable artificial delay for change stream tests that expect an event when change stream is initialized

JAVA-4369

### DIFF
--- a/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/ChangeStreamsTest.java
+++ b/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/ChangeStreamsTest.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 
 import static com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient.disableSleep;
@@ -34,9 +35,8 @@ import static org.junit.Assume.assumeFalse;
 
 public final class ChangeStreamsTest extends UnifiedReactiveStreamsTest {
 
-    @SuppressWarnings("ArraysAsListWithZeroOrOneArgument")
     private static final List<String> ERROR_REQUIRED_FROM_CHANGE_STREAM_INITIALIZATION_TESTS =
-            Arrays.asList(
+            Collections.singletonList(
                     "Test with document comment - pre 4.4"
             );
 
--- a/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/ChangeStreamsTest.java
+++ b/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/ChangeStreamsTest.java
@@ -19,18 +19,59 @@ package com.mongodb.reactivestreams.client.unified;
 import com.mongodb.lang.Nullable;
 import org.bson.BsonArray;
 import org.bson.BsonDocument;
+import org.junit.After;
 import org.junit.runners.Parameterized;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
+
+import static com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient.disableSleep;
+import static com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient.enableSleepAfterCursorOpen;
+import static org.junit.Assume.assumeFalse;
 
 public final class ChangeStreamsTest extends UnifiedReactiveStreamsTest {
+
+    @SuppressWarnings("ArraysAsListWithZeroOrOneArgument")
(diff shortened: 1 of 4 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -188,17 +188,20 @@ public void shouldPassAllOutcomes() {
         }
     }
 
-    private void compareEvents(final BsonDocument operation) {
```
jyemin, line 191: Drive by change.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/ChangeStreamsTest.java
```diff
@@ -19,18 +19,59 @@
...
+
+import static com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient.disableSleep;
+import static com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient.enableSleepAfterCursorOpen;
+import static org.junit.Assume.assumeFalse;
 
 public final class ChangeStreamsTest extends UnifiedReactiveStreamsTest {
+
+    @SuppressWarnings("ArraysAsListWithZeroOrOneArgument")
+    private static final List<String> ERROR_REQUIRED_FROM_CHANGE_STREAM_INITIALIZATION_TESTS =
+            Arrays.asList(
```
stIncMale, line 39: We can just use `Collections.singletonList`. If in the future we need to specify multiple tests here, we will be forced to replace it with `Arrays.asList`. On the other hand, if we specify multiple array elements in `Arrays.asList` in the future, nothing will force us to remove the `@SuppressWarnings("ArraysAsListWithZeroOrOneArgument")` annotation, which will become useless. That's why I propose to use `Collections.singletonList`.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/ChangeStreamsTest.java
```diff
@@ -19,18 +19,59 @@
...
+                    "Test that comment is set on getMore",
+                    "Test that comment is not set on getMore - pre 4.4"
+            );
+
+    private static final List<String> CURSOR_OPEN_TIMING_SENSITIVE_TESTS =
+            Arrays.asList(
+                    "Test with document comment",
+                    "Test with document comment - pre 4.4",
+                    "Test with string comment",
+                    "Test that comment is set on getMore"
```
stIncMale, line 54: Does this mean that tests mentioned here and in `ERROR_REQUIRED_FROM_CHANGE_STREAM_INITIALIZATION_TESTS`, `EVENT_SENSITIVE_TESTS` are incorrect and we ignore them until they are fixed?

stIncMale (changes requested):
Jeff, you mentioned in the description that `ignoreExtraEvents` _"is a new option and not yet out of spec review"_. I see this option specified [here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectedeventsforclient). Is there something in addition to this piece that is being reviewed?

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/ChangeStreamsTest.java
jyemin: I don't consider the tests in this list to be incorrect.  `enableSleepAfterCursorOpen` is a workaround that we already apply to other reactive unified test runners, and it's just necessary due to the way that reactive streams work.  There may be other enhancements we could make to the unified test format definition that would make this unnecessary, but for now this is what we have.  

The tests in the other two lists are not incorrect per se, but we don't have a workaround right now that would let them run, so we just skip them for now.  There may be enhancements that we could request in the unified test format, or perhaps other workarounds that we could add into the reactive test runner.  I separate them into two lists as a way to (sort of) document the two different problems that we're encountering.

I will open JAVA ticket(s) to track the possibility of un-skipping both classes of tests.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/ChangeStreamsTest.java
jyemin: Fair enough.  Done.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/ChangeStreamsTest.java
stIncMale: > I don't consider the tests in this list to be incorrect. enableSleepAfterCursorOpen is a workaround

Right, sorry, I meant to mention only skipped tests, but then accidentally mixed everything together.

Thank you for explaining.

stIncMale (approved):

PR #922, opened 2022-04-18
Use $$unsetOrMatches for upsert and multi in spec tests

jyemin:
* Modifications made in a previous commit of bulkWrite-comment.json
  had to be preserved, so this is not an exact copy of the spec tests
* Modifications to require minServerVersion of 3.6 also had to be preserved,
  since Java driver still supports older server versions

JAVA-4568


jyemin:
Merging without review since it's only test file changes

PR #924, opened 2022-04-21
Added GET_AND_SET_FIELDS_CONVENTION

rozza:
Ignores all non field properties

JAVA-4545

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/Conventions.java
+++ b/bson/src/main/org/bson/codecs/pojo/Conventions.java
@@ -73,6 +73,7 @@ public final class Conventions {
      * A convention that <em>only uses fields</em> for setting and getting values to be serialized. Automatically,
      * removes any properties that are not explicitly backed by a field.
      *
+     * <p>For a field to be discoverable by the PojoCodec and therefore this convention it must have a public: field, getter or setter.</p>
      * <p>Note: This convention is not part of the {@code DEFAULT_CONVENTIONS} list and must explicitly be set.</p>
      *
      * @since 4.7
--- a/bson/src/main/org/bson/codecs/pojo/Conventions.java
+++ b/bson/src/main/org/bson/codecs/pojo/Conventions.java
@@ -69,8 +69,17 @@ public final class Conventions {
      */
     public static final Convention USE_GETTERS_FOR_SETTERS = new ConventionUseGettersAsSettersImpl();
 
-
     /**
+     * A convention that <em>only uses fields</em> for setting and getting values to be serialized. Automatically,
+     * removes any properties that are not explicitly backed by a field.
+     *
+     * <p>Note: This convention is not part of the {@code DEFAULT_CONVENTIONS} list and must explicitly be set.</p>
+     *
+     * @since 4.7
+     */
+    public static final Convention GET_AND_SET_FIELDS_CONVENTION = new ConventionGetAndSetFieldsImpl();
+
+        /**
      * A convention that sets the IdGenerator if the id property is either a {@link org.bson.types.ObjectId} or
      * {@link org.bson.BsonObjectId}.
      *
--- a/bson/src/main/org/bson/codecs/pojo/Conventions.java
+++ b/bson/src/main/org/bson/codecs/pojo/Conventions.java
@@ -70,10 +70,8 @@ public final class Conventions {
     public static final Convention USE_GETTERS_FOR_SETTERS = new ConventionUseGettersAsSettersImpl();
 
     /**
-     * A convention that <em>only uses fields</em> for setting and getting values to be serialized. Automatically,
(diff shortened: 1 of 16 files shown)


bson/src/main/org/bson/codecs/pojo/Conventions.java
```diff
@@ -69,8 +69,17 @@ public final class Conventions {
      */
     public static final Convention USE_GETTERS_FOR_SETTERS = new ConventionUseGettersAsSettersImpl();
 
-
     /**
```
jyemin, line 72: Does this convention assume that the `CLASS_AND_PROPERTY_CONVENTION` (or equivalent) precedes it in the list of conventions?  It seems so, but if so perhaps it's worth stating, as it would seem reasonable to assume that this convention would discover all the fields and add them as properties rather than removing fields that have already been discovered by a previous convention.

bson/src/main/org/bson/codecs/pojo/Conventions.java
```diff
@@ -69,8 +69,17 @@ public final class Conventions {
...
+     * A convention that <em>only uses fields</em> for setting and getting values to be serialized. Automatically,
```
jyemin, line 73: Let's also document whether this convention will encode/decode fields that _don't_ have a corresponding getter and/or setter.  I'm not sure whether it should or not, actually.

jyemin (changes requested):

bson/src/main/org/bson/codecs/pojo/FieldPropertyAccessor.java
```diff
@@ -32,9 +32,15 @@
         }
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     public <S> T get(final S instance) {
-        return wrapped.get(instance);
+        try {
+            return (T) wrapped.getPropertyMetadata().getField().get(instance);
```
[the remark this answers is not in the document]
rozza: This is a field property accessor so ensures that the field and not the getter method (if available) is accessed.

bson/src/main/org/bson/codecs/pojo/Conventions.java
rozza: Its documented as only using fields - so wont use any methods for encoding / decoding.

jyemin: There are two pieces to it:

1. Choosing which properties to include, based on existence of field/getter/setter
2. Once properties are included/excluded, choosing how to get/set the value

The second is clear to me, but the first is not.

bson/src/main/org/bson/codecs/pojo/Conventions.java
rozza: It doesn't require any other conventions but can be used in combination with them.

bson/src/test/unit/org/bson/codecs/pojo/entities/GetAndSetMutatesModel.java
```diff
@@ -0,0 +1,74 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bson.codecs.pojo.entities;
+
+import java.util.Objects;
+
+public final class GetAndSetMutatesModel {
```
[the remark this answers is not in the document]
rozza: Working on this, I will probably add another commit so these new models can be reused with other conventions.

bson/src/main/org/bson/codecs/pojo/Conventions.java
rozza: Ack - will clarify

bson/src/test/unit/org/bson/codecs/pojo/entities/GetAndSetMutatesModel.java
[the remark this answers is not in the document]
rozza: Done

bson/src/main/org/bson/codecs/pojo/Conventions.java
rozza: Added some text to clarify.

jyemin: I'm still finding this hard to interpret. Perhaps we can wordsmith it together in person..

rozza:
Marking as draft - awaiting user feedback on the design.

lbenc135:
Hey @rozza, this looks like exactly what we need as well. Any estimates when can it be released? Let me know if I can help. I'll try to implement a custom convention on my side for now, but would like to use the official version if possible.

jyemin:
Closing as this has gone stale.

PR #925, opened 2022-04-21
Support global configuration of UuidRepresentation for legacy encoder…

jyemin:
…/decoder

Both BasicBSONEncoder and BSONCallbackAdapter assume a UuidRepresentation of
JAVA_LEGACY in a way that is not configurable. Allow it to be configured process-wide
using a static method for each class to set the default.

JAVA-4587

### DIFF
--- a/bson/src/main/org/bson/BSONCallbackAdapter.java
+++ b/bson/src/main/org/bson/BSONCallbackAdapter.java
@@ -16,10 +16,13 @@
 
 package org.bson;
 
+import org.bson.internal.UuidHelper;
 import org.bson.types.Decimal128;
 import org.bson.types.ObjectId;
 
-import static org.bson.Bits.readLong;
+import java.util.UUID;
+
+import static org.bson.BasicBSONDecoder.getDefaultUuidRepresentation;
 
 class BSONCallbackAdapter extends AbstractBsonWriter {
 
@@ -83,12 +86,21 @@ class BSONCallbackAdapter extends AbstractBsonWriter {
 
     @Override
     protected void doWriteBinaryData(final BsonBinary value) {
-        if (value.getType() == BsonBinarySubType.UUID_LEGACY.getValue()) {
-            bsonCallback.gotUUID(getName(),
-                    readLong(value.getData(), 0),
-                    readLong(value.getData(), 8));
+        UuidRepresentation defaultUuidRepresentation = getDefaultUuidRepresentation();
+        if (defaultUuidRepresentation == UuidRepresentation.STANDARD) {
+            if (value.getType() == BsonBinarySubType.UUID_LEGACY.getValue()) {
+                bsonCallback.gotBinary(getName(), value.getType(), value.getData());
+            } else {
+                UUID uuid = UuidHelper.decodeBinaryToUuid(value.getData(), value.getType(), defaultUuidRepresentation);
+                bsonCallback.gotUUID(getName(), uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
+            }
         } else {
-            bsonCallback.gotBinary(getName(), value.getType(), value.getData());
+            if (value.getType() == BsonBinarySubType.UUID_LEGACY.getValue()) {
+                UUID uuid = UuidHelper.decodeBinaryToUuid(value.getData(), value.getType(), defaultUuidRepresentation);
+                bsonCallback.gotUUID(getName(), uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
+            } else {
+                bsonCallback.gotBinary(getName(), value.getType(), value.getData());
+            }
         }
     }
 
--- a/bson/src/main/org/bson/BSONCallbackAdapter.java
+++ b/bson/src/main/org/bson/BSONCallbackAdapter.java
@@ -86,21 +86,20 @@ class BSONCallbackAdapter extends AbstractBsonWriter {
 
     @Override
     protected void doWriteBinaryData(final BsonBinary value) {
-        UuidRepresentation defaultUuidRepresentation = getDefaultUuidRepresentation();
-        if (defaultUuidRepresentation == UuidRepresentation.STANDARD) {
-            if (value.getType() == BsonBinarySubType.UUID_LEGACY.getValue()) {
-                bsonCallback.gotBinary(getName(), value.getType(), value.getData());
-            } else {
-                UUID uuid = UuidHelper.decodeBinaryToUuid(value.getData(), value.getType(), defaultUuidRepresentation);
-                bsonCallback.gotUUID(getName(), uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
-            }
+        if (BsonBinarySubType.isUuid(value.getType())) {
+            doWriteUuid(value);
         } else {
-            if (value.getType() == BsonBinarySubType.UUID_LEGACY.getValue()) {
-                UUID uuid = UuidHelper.decodeBinaryToUuid(value.getData(), value.getType(), defaultUuidRepresentation);
-                bsonCallback.gotUUID(getName(), uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
-            } else {
-                bsonCallback.gotBinary(getName(), value.getType(), value.getData());
-            }
+            bsonCallback.gotBinary(getName(), value.getType(), value.getData());
+        }
+    }
+
+    private void doWriteUuid(final BsonBinary value) {
+        UuidRepresentation defaultUuidRepresentation = getDefaultUuidRepresentation();
+        if (value.getType() == defaultUuidRepresentation.getSubtype().getValue()) {
+            UUID uuid = UuidHelper.decodeBinaryToUuid(value.getData(), value.getType(), defaultUuidRepresentation);
+            bsonCallback.gotUUID(getName(), uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
+        } else {
+            bsonCallback.gotBinary(getName(), value.getType(), value.getData());
         }
     }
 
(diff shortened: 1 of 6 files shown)


bson/src/main/org/bson/BasicBSONDecoder.java
```diff
@@ -22,11 +22,47 @@
...
 /**
  * Basic implementation of BSONDecoder interface that creates BasicBSONObject instances
  */
 public class BasicBSONDecoder implements BSONDecoder {
 
+    /**
+     * Sets the default {@link UuidRepresentation} to use when decoding BSON binary values.
+     *
+     * <p>
+     * If unset, the default is {@link UuidRepresentation#JAVA_LEGACY}.
```
rozza, line 36: Nit: Should this just be: "Defaults to {@link UuidRepresentation#JAVA_LEGACY}."

Due to the notNull I cant see how its unset?

bson/src/main/org/bson/BasicBSONDecoder.java
```diff
@@ -22,11 +22,47 @@
...
 
+import static org.bson.assertions.Assertions.notNull;
+
 /**
  * Basic implementation of BSONDecoder interface that creates BasicBSONObject instances
  */
 public class BasicBSONDecoder implements BSONDecoder {
 
+    /**
+     * Sets the default {@link UuidRepresentation} to use when decoding BSON binary values.
```
rozza, line 33: Perhaps add warning that this globally sets the default for all users of `BasicBSONDecoder` within the JVM.

rozza (commented):
Couple of javadoc change suggestions - to make users aware this is a global change.

bson/src/main/org/bson/BSONCallbackAdapter.java
```diff
@@ -83,12 +86,21 @@ protected void doWriteEndArray() {
...
+                bsonCallback.gotUUID(getName(), uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
+            }
         } else {
-            bsonCallback.gotBinary(getName(), value.getType(), value.getData());
+            if (value.getType() == BsonBinarySubType.UUID_LEGACY.getValue()) {
+                UUID uuid = UuidHelper.decodeBinaryToUuid(value.getData(), value.getType(), defaultUuidRepresentation);
+                bsonCallback.gotUUID(getName(), uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
+            } else {
+                bsonCallback.gotBinary(getName(), value.getType(), value.getData());
+            }
```
stIncMale, line 103: I cannot understand neither the old nor the new code. I took a pause and tried again in a day, but I still can't make any sense even of the old code. Jeff, could you please help me?

- The old code was decoding UUID values if they had the legacy format, and was omitting the decoding them if they were in the standard format. Why was it done this way, and how/when the values that had not been decoded were then decoded?
- The new code brings at least one more things that I fail to understand: the meaning of `BasicBSONDecoder.getDefaultUuidRepresentation`. The interpretation I can think of is "the format of the binary data that is being decoded". However, that format is specified in `value.getType`, and indeed, `BasicBSONDecoder.getDefaultUuidRepresentation` is not used as a specifier of the decoded data format. Instead, it is used to either decode the data as `UUID` or omit decoding it, and depending on the `getDefaultUuidRepresentation` value, the code omits decoding in different cases. Even looking at the `should decode UUID according to default uuid representation` test did not help me to understand what's happening.

bson/src/test/unit/org/bson/BasicBSONEncoderSpecification.groovy
```diff
@@ -179,4 +188,28 @@ class BasicBSONEncoderSpecification extends Specification {
...
+        def bytes = bsonEncoder.encode(new BasicBSONObject(document))
+        def decodedDocument = new BsonDocumentCodec().decode(new BsonBinaryReader(ByteBuffer.wrap(bytes)),
+                DecoderContext.builder().build())
+
+        then:
+        defaultUuidRepresentation == JAVA_LEGACY
+        decodedDocument.getBinary('u').asUuid(uuidRepresentation) == uuid
+
+        cleanup:
+        setDefaultUuidRepresentation(defaultUuidRepresentation)
```
stIncMale, line 210: Shouldn't `defaultUuidRepresentation == JAVA_LEGACY` be moved (or copied, depending on how the framework executes this test) in the `cleanup` block? I worry that the current code may not always return the static field to its default `JAVA_LEGACY` value.

bson/src/main/org/bson/BSONCallbackAdapter.java
jyemin: It's all part of the accidents of history...

In the beginning, there was only binary subtype 3 for UUIDs and a bunch of drivers spent their days happily decoding them into platform UUID objects (like java.util.UUID in Java) and encoding them back into binary subtype 3.

But lo, all was not well in the world of UUIDs.  It turned out that all the drivers twiddled the UUID bits in different ways when encoding them, and twiddled them back in the same way when decoding them.  So a binary subtype 3 sitting in a MongoDB document on some server could be decoding in three different ways (the java way, the C# way, and the python way).

And thus was born binary subtype 4.  A new subtype to rule them all.  For subtype 4 we picked the one true way to twiddle the bits (the python way, which is also the way it's specified in [the holy book](https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.2).  

But lo (again, lo) that solved one problem but created more problems.  Now each UUID-supporting driver has to decide how to encode a UUID.  Should it use the unholy format that it had been using, or the new format?  And how to decode?  Should it turn all subtype 3 and 4 into platform UUIDs?  And if so, how to round-trip the UUIDs back to the database? A true mess.

So it was decided that the binary subtype 3's would continue to be decoded into platform UUIDs, and platform UUIDs would be encoding into binary subtype 3's.  And binary subtype 4 would be decoding into Binary, which no bit twiddling done at all.  And thus round-tripping was preserved.  But oh, still what a mess.

And that's how it remained until around 2019, when a sage engineer (Bernie Hackett) decided that we could no longer live in such confusion, and thus a [new holy text](https://github.com/mongodb/specifications/blob/master/source/uuid.rst) was born.  And the 4.0 Java driver obeyed the letter of this law, for the most part.  It turns out that there were a few corners of the legacy Java driver API that remained in the dark.  And here they are. 

So what is defaultUuidRepresentation and why does it do what it does?  Hopefully [this section](https://github.com/mongodb/specifications/blob/master/source/uuid.rst#implicit-decoding-and-encoding) will explain it.

bson/src/test/unit/org/bson/BasicBSONEncoderSpecification.groovy
jyemin: `defaultUuidRepresentation == JAVA_LEGACY` is an assertion, not an assignment.  It's just saying that the original value of the field is JAVA_LEGACY.  But it's weird to put that assertion in this test.  There should just be a separate test, like:

```
def 'default value of defaultUuidRepresentation is JAVA_LEGACY'() {
    expect:
    getDefaultUuidRepresentation() == JAVA_LEGACY
}
```

bson/src/main/org/bson/BSONCallbackAdapter.java
```diff
@@ -83,12 +86,21 @@ protected void doWriteEndArray() {
 
     @Override
     protected void doWriteBinaryData(final BsonBinary value) {
-        if (value.getType() == BsonBinarySubType.UUID_LEGACY.getValue()) {
-            bsonCallback.gotUUID(getName(),
-                    readLong(value.getData(), 0),
-                    readLong(value.getData(), 8));
+        UuidRepresentation defaultUuidRepresentation = getDefaultUuidRepresentation();
+        if (defaultUuidRepresentation == UuidRepresentation.STANDARD) {
+            if (value.getType() == BsonBinarySubType.UUID_LEGACY.getValue()) {
```
ramasai1, line 91: I think this needs to be changed to:
```suggestion
            if (value.getType() == BsonBinarySubType.UUID_STANDARD.getValue()) {
                UUID uuid = UuidHelper.decodeBinaryToUuid(value.getData(), value.getType(), defaultUuidRepresentation);
      bsonCallback.gotUUID(getName(), uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
      } else {
          bsonCallback.gotBinary(getName(), value.getType(), value.getData());
```
Without this change, all subtypes other than 3 will go through the `decodeBinary` function and will throw an exception.

jyemin: I fixed the bug in a slightly different way, by refactoring the code to extract common logic in both branches, and introduce a new top-level branch for non-UUID case.

bson/src/main/org/bson/BasicBSONDecoder.java
jyemin: Done

bson/src/main/org/bson/BSONCallbackAdapter.java
ramasai1: Changes make sense to me.

bson/src/main/org/bson/BSONCallbackAdapter.java
stIncMale: Nice writing style! Thank you for the detailed explanation and the link to the spec. Now it all makes sense to me, including the implicit decoding table from the spec. And the current code complies to that table.

bson/src/test/unit/org/bson/BasicBSONEncoderSpecification.groovy
stIncMale: Sorry, I was clearly blind/stupid yesterday.

stIncMale (approved):

bson/src/test/unit/org/bson/BasicBSONEncoderSpecification.groovy
jyemin: Not at all.  As I was writing that assertion I knew it was confusing, and was being a bit lazy.

rozza (approved):

PR #923, opened 2022-04-20
Add query building API for the `$searchMeta` aggregation pipeline stage

stIncMale:
**Superseded by https://github.com/mongodb/mongo-java-driver/pull/952.**

The main thing to know about the [`$searchMeta`](https://www.mongodb.com/docs/atlas/atlas-search/query-syntax/#-searchmeta) stage is that any `$search` stage is and will be a valid `$searchMeta` stage. This is not documented, but so I was told by Evan Plotkin.

Evergreen patch: https://spruce.mongodb.com/version/6260459232f41728ef1e93c5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC.

In order to merge this PR we need to:

- merge https://github.com/mongodb/mongo-java-driver/pull/891;
- change the target branch from `JAVA-4415` to `master`;
- delete the `JAVA-4415` branch;
- merge this PR.

JAVA-4315

PR #928, opened 2022-04-28
Add query building API for Atlas search operators

stIncMale:
**Superseded by https://github.com/mongodb/mongo-java-driver/pull/952.**

The PR is submitted early to give an opportunity to review it piece by piece, and also to allow discussing some approaches early. However, this is up to you guys, and it is of course fine to wait until the work is done.

**Evergreen patch:** https://spruce.mongodb.com/version/62857cb55623430561712dc1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC.

**Current state:**

1. Integration tests for highlighting and alternate analyzers.
2. [Score modifiers](https://www.mongodb.com/docs/atlas/atlas-search/scoring/) are implemented.
3. API for the following operators is implemented / in progress:
    - [`exists`](https://www.mongodb.com/docs/atlas/atlas-search/exists/) (was implemented in https://github.com/mongodb/mongo-java-driver/pull/891)
    - [`text`](https://www.mongodb.com/docs/atlas/atlas-search/text/)
    - [`compound`](https://www.mongodb.com/docs/atlas/atlas-search/compound/)
    - [`autocomplete`](https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/)
    - [`range`](https://www.mongodb.com/docs/atlas/atlas-search/range/)
    - [`near`](https://www.mongodb.com/docs/atlas/atlas-search/near/)
    - next planned [`phrase`](https://www.mongodb.com/docs/atlas/atlas-search/phrase/)
    - next planned [`wildcard`](https://www.mongodb.com/docs/atlas/atlas-search/wildcard/)
    - next planned [`regex`](https://www.mongodb.com/docs/atlas/atlas-search/regex/)

One may use `AggregatesSearchIntegrationTest`/`SearchOperatorSpec`  to either play around with Java/Scala API, or see how it looks like in practice.

In order to merge this PR we need to:

- merge https://github.com/mongodb/mongo-java-driver/pull/923;
- change the target branch from `JAVA-4315` to `master`;
- delete the `JAVA-4315` branch;
- merge this PR.

JAVA-4190


stIncMale:
Item 2 in the description of this PR may be worth discussing depending on how you feel about the approach taken for the `compound` operator API.

driver-core/src/main/com/mongodb/client/model/search/TextSearchOperator.java
```diff
@@ -0,0 +1,45 @@
...
+@Evolving
+public interface TextSearchOperator extends SearchOperator {
+    /**
+     * Creates a new {@link TextSearchOperator} that uses fuzzy search
+     * and does not use {@linkplain #synonyms(String) synonyms}.
+     *
+     * @param options Fuzzy search options.
+     * @return A new {@link TextSearchOperator}.
+     */
+    TextSearchOperator fuzzy(FuzzySearchOptions options);
```
jyemin, line 33: What would the implementation approach be if in the future we need to add a method to another interface that extends SearchOperator that has the same name and parameters as one in another interface that extends SearchOperator?  SearchConstructibleBsonElement couldn't implement both methods.

I guess we would just split SearchConstructibleBsonElement into multiple classes?

driver-core/src/main/com/mongodb/client/model/search/FuzzySearchOptions.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+package com.mongodb.client.model.search;
+
+import com.mongodb.annotations.Evolving;
+import org.bson.conversions.Bson;
+
+/**
+ * Fuzzy search options that may be used with some {@link SearchOperator}s.
+ *
+ * @mongodb.atlas.manual atlas-search/autocomplete/ autocomplete operator
```
jyemin, line 24: Why the link to autocomplete?

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchOperator.scala
```diff
@@ -26,6 +28,15 @@ import org.mongodb.scala.bson.conversions.Bson
  */
 object SearchOperator {
 
+  /**
+   * Returns a base for a `SearchOperator`` that may combine multiple `SearchOperator`s.
```
jyemin, line 32: Remove extra backquote after SearchOperator

driver-core/src/main/com/mongodb/client/model/search/CompoundSearchOperatorBase.java
```diff
@@ -0,0 +1,66 @@
...
+package com.mongodb.client.model.search;
+
+import com.mongodb.annotations.Evolving;
+
+/**
+ * @see SearchOperator#compound()
+ * @since 4.7
+ */
+@Evolving
+public interface CompoundSearchOperatorBase {
```
jyemin, line 25: I don't see a need for it in the current design (maybe it is a vestige of a previous one).  On my local copy I was able to pull up all the methods from this interface into CompoundSearchOperator and (after a few small changes) remove this interface entirely.

driver-core/src/main/com/mongodb/client/model/search/CompoundSearchOperatorBase.java
```diff
@@ -0,0 +1,66 @@
...
+     * This method may be called multiple times.</p>
+     *
+     * @param clauses Non-empty clauses.
+     * @return A new {@link CompoundSearchOperator}.
+     */
+    ShouldCompoundSearchOperator should(Iterable<? extends SearchOperator> clauses);
+
+    /**
+     * Creates a new {@link CompoundSearchOperator} by adding to it {@code clauses} that, similarly to {@link #must(Iterable)},
+     * must all be satisfied. The difference is that {@link #filter(Iterable)} does not affect the relevance score.
```
jyemin, line 58: IntelliJ warns that this is a link to itself. Perhaps replace the link with "this method"?

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -30,6 +35,17 @@
...
 @Evolving
 public interface SearchOperator extends Bson {
+    /**
+     * Returns a base for a {@link SearchOperator} that may combine multiple {@link SearchOperator}s.
+     * Combining {@link SearchOperator}s affects calculation of the relevance score.
+     *
+     * @return A base for a {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/compound/ compound operator
+     */
+    static CompoundSearchOperatorBase compound() {
```
jyemin, line 45: Is an empty compound a legal search, or is at least one clause required?  Probably not a big deal either way, but maybe document it if it's not legal.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -41,6 +57,24 @@ static ExistsSearchOperator exists(final FieldSearchPath path) {
...
 
+    /**
+     * Returns a {@link SearchOperator} that performs a full-text search.
+     *
+     * @param queries Non-empty terms to search for.
+     * @param paths Non-empty document fields to be searched.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/text/ text operator
+     */
+    static TextSearchOperator text(final Iterable<String> queries, final Iterable<? extends SearchPath> paths) {
```
jyemin, line 68: I know you're not fond of redundant overloads, but this signature could get annoying if the common case is to have a single query and path.

jyemin (changes requested):
It's looking great.  I haven't completed my review, but wanted to get you some initial feedback on the API.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
stIncMale: It's not legal, and this is the answer to the [previous question](https://github.com/mongodb/mongo-java-driver/pull/928#discussion_r861356847) about `CompoundSearchOperatorBase`. We need that interface specifically because `SearchOperator.compound()` must not return a `SearchOperator` because such operator would not have been a valid one.

driver-core/src/main/com/mongodb/client/model/search/FuzzySearchOptions.java
stIncMale: It's because the `autocomplete` operator also supports fuzzy search with the same syntax. I.e., this class is relevant for both `text` and `autocomplete` operators.

driver-core/src/main/com/mongodb/client/model/search/TextSearchOperator.java
stIncMale: Exactly. In fact, this has already happened with `FuzzySearchOptions`: methods `FuzzySearchOptions.option` and `SearchOptions.option` return `FuzzySearchOptions` and `SearchOptions` respectively. So I had to implement `FuzzySearchOptions` in `FuzzySearchOptionsConstructibleBson` instead of implementing it in `SearchConstructibleBson`.

driver-core/src/main/com/mongodb/client/model/search/CompoundSearchOperatorBase.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * @see SearchOperator#compound()
+ * @since 4.7
+ */
+@Evolving
+public interface CompoundSearchOperatorBase {
+    /**
+     * Creates a new {@link CompoundSearchOperator} by adding to it {@code clauses} that must all be satisfied.
+     * <p>
+     * This method may be called multiple times.</p>
```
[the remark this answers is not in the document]
stIncMale: I documented the behavior, the _"by adding to it clauses that ..."_ is the key. Do you have in mind a better way of expressing the behavior?

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchOperator.scala
stIncMale: Thanks! Done.

driver-core/src/main/com/mongodb/client/model/search/CompoundSearchOperatorBase.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
stIncMale: > The Should part suggests that it represents only the should clause of the compound operator

`Mustable`, `MustNotable`, `Shoudable`, `Filterable` ? :trollface:

stIncMale: >  I will point out examples of where I think that could be done in separate comments.

I added those docs in https://github.com/mongodb/mongo-java-driver/pull/928/commits/8aeed1fc9fe35bd5f68c775c123d7375b3b4cb75.

stIncMale:
@jyemin, I addressed things we discuss. Please see the description of 889fbea7392857b1e0353ea5d6f3eb62752849c4 for the details on what was done and what stayed as is.

jyemin (approved):
Really well done.  LGTM.

driver-core/src/test/functional/com/mongodb/client/model/search/AggregatesSearchIntegrationTest.java
```diff
@@ -316,25 +414,125 @@ private static Stream<Arguments> args() {
...
+                                        exists(fieldPath("fieldName6"))
+                                                .score(function(pathExpression(fieldPath("expressionFieldName")))),
+                                        exists(fieldPath("fieldName7"))
+                                                .score(function(pathExpression(fieldPath("expressionFieldName"))
+                                                        .undefined(-1))),
+                                        exists(fieldPath("fieldName8"))
+                                                .score(function(constantExpression(-1.2f))),
+                                        exists(fieldPath("fieldName9"))
+                                                .score(function(
+                                                        gaussExpression(-10, pathExpression(fieldPath("gaussianFeldName")), Double.MAX_VALUE / -2))),
```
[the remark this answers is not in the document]
stIncMale: Thanks, done in all three places.

stIncMale:
@jyemin, you may want to review the two new commits 074e69848afff34307747570889c43aae37ba076 and 93e934d9af02bea7c525c5fef2e724316050e1dc. Really, it's mostly the first one, specifically `Beta.java` that is worth looking at.

rozza (changes requested):
Looks really good.

Had a couple of comments on the annotations - which should narrow down the size of the PR for existing Beta code.  Is the Evolving annotation needed? Are there any uses outside of the Beta code?

Also I noticed a few methods weren't covered by any tests eg: RangeConstructibleBsonElement.score 
SearchScore.of

Not sure if coverage is needed but may help prevent any regressions.

driver-core/src/main/com/mongodb/annotations/Beta.java
```diff
@@ -47,6 +47,25 @@
         ElementType.PACKAGE,
         ElementType.TYPE })
 @Documented
-@Beta
+@Beta(Beta.Reason.CLIENT)
 public @interface Beta {
+    /**
+     * @return The reason an API element is marked with {@link Beta}.
+     */
+    Reason[] value();
```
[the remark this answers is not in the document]
stIncMale: I don't think that having a default value here is a good idea, because it has a good chance of causing us to just use `@Beta` in the future forgetting that we need to think about the reason. And each time we see just `@Beta` we would be guessing "is it really `CLIENT`, or did we forget to specify `SERVER`?".

An additional, but a much lesser point (just a technicality): some of the existing APIs are actually in `CLIENT` beta, like `AwsCredential`, while others, like `WindowedComputation` are in `SERVER` beta. Well, they were marked as beta because of the server, but we need to get rid of those annotations. So we can't technically rely on the default value until we remove `@Beta` from the API related to window functions.

driver-core/src/main/com/mongodb/client/model/search/FieldSearchPath.java
```diff
@@ -15,13 +15,18 @@
...
+import org.bson.conversions.Bson;
+
+import static com.mongodb.internal.client.model.Util.SEARCH_PATH_VALUE_KEY;
 
 /**
  * @see SearchPath#fieldPath(String)
  * @since 4.7
  */
 @Evolving
+@Beta(Beta.Reason.CLIENT)
```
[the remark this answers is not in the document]
stIncMale: It is specified in the `@Evolving` docs:

```java
/**
 * Signifies that the annotated program element is subject to incompatible changes by means of adding abstract methods.
 * This, in turn, means that implementing interfaces or extending classes annotated with {@link Evolving} bears the risk
 * of doing extra work during upgrades.
 * Using such program elements is no different from using ordinary unannotated program elements.
 * Note that the presence of this annotation implies nothing about the quality or performance of the API in question.
 */
```

`@Evolving` was introduced in https://github.com/mongodb/mongo-java-driver/pull/891, and the description of that PR contains an additional explanation for reviewers: _"API is marked with a new annotation `@Evolving` to make it clear that users should not implement/extend it. We implicitly rely on the restrictions specified by this annotation in some APIs, e.g. in `MongoDatabase`, `MongoCollection`, and it seems useful to have a way to make the restriction explicit."_

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchFacet.scala
```diff
@@ -16,7 +16,7 @@
 package org.mongodb.scala.model.search
 
 import com.mongodb.annotations.Beta
-import com.mongodb.client.model.search.{ SearchFacet => JSearchFacet }
+import com.mongodb.client.model.search.{ NumberSearchFacet, SearchFacet => JSearchFacet }
```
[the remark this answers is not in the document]
stIncMale: Thanks for spotting this. It must have been the IDE that did the import, and I did not notice. Done.

driver-core/src/main/com/mongodb/annotations/Beta.java
```diff
@@ -47,6 +47,25 @@
...
+    /**
+     * @see Beta#value()
+     */
+    enum Reason {
+        /**
+         * The driver API is in preview, but the corresponding server API is finalized.
+         */
+        CLIENT,
+        /**
+         * We feel good about the driver API, but the corresponding server API is in preview.
```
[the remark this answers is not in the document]
stIncMale: It is better the way you propose 👍  Done.

driver-core/src/main/com/mongodb/annotations/Beta.java
```diff
@@ -47,6 +47,25 @@
...
+     * @return The reason an API element is marked with {@link Beta}.
+     */
+    Reason[] value();
+
+    /**
+     * @see Beta#value()
+     */
+    enum Reason {
+        /**
+         * The driver API is in preview, but the corresponding server API is finalized.
```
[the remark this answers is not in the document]
stIncMale: Done.

driver-core/src/main/com/mongodb/annotations/Beta.java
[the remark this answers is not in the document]
rozza: I suppose explicit over implicit is better its just kind of ugly.

driver-core/src/main/com/mongodb/client/model/search/FieldSearchPath.java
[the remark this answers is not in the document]
rozza: Ack :)

rozza (approved):
Thanks for your responses - LGTM!

PR #929, opened 2022-05-03
Always use count command for estimatedDocumentCount

jyemin:
JAVA-4530


rozza (approved):
LGTM - certainly simpler!

PR #931, opened 2022-05-04
Support comment in estimatedDocumentCount

jyemin:
JAVA-4529

Hopefully the last comment on comments.


driver-core/src/main/com/mongodb/internal/operation/EstimatedDocumentCountOperation.java
```diff
@@ -105,6 +118,9 @@ private CommandCreator getCommandCreator(final SessionContext sessionContext) {
             BsonDocument document = new BsonDocument("count", new BsonString(namespace.getCollectionName()));
             appendReadConcernToCommand(sessionContext, connectionDescription.getMaxWireVersion(), document);
             putIfNotZero(document, "maxTimeMS", maxTimeMS);
+            if (comment != null) {
```
rozza, line 121: Nit: Could also use the helper: `putIfNotNull(document, "comment", comment);`

jyemin: I copied this from another operation, so at least it's consistent.

PR #934, opened 2022-05-09
Support clustered index on a collection

jyemin:
JAVA-4576

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/ClusteredIndexOptions.java
+++ b/driver-core/src/main/com/mongodb/client/model/ClusteredIndexOptions.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model;
+
+import com.mongodb.lang.Nullable;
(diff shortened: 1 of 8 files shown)


driver-core/src/main/com/mongodb/client/model/ClusteredIndexOptions.java
```diff
@@ -0,0 +1,88 @@
...
+        return name;
+    }
+
+    /**
+     * Sets the index name
+     * @param name the index name
+     */
+    public void name(final String name) {
+        this.name = name;
+    }
```
stIncMale, line 78: Would it be better to return `this` here?

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -356,6 +359,21 @@ public CreateCollectionOperation timeSeriesOptions(@Nullable final TimeSeriesOpt
...
+        this.clusteredIndexKey = clusteredIndexKey;
+        return this;
+    }
+
+    public CreateCollectionOperation clusteredIndexUnique(final boolean clusteredIndexUnique) {
+        this.clusteredIndexUnique = clusteredIndexUnique;
+        return this;
+    }
+
+    public CreateCollectionOperation clusteredIndexName(final String clusteredIndexName) {
```
stIncMale, line 372: Let's mark `clusteredIndexName` as `@Nullable`, because we indeed may pass `null` when calling this method in `MongoOperationPublisher`.

driver-core/src/main/com/mongodb/client/model/ClusteredIndexOptions.java
```diff
@@ -0,0 +1,88 @@
...
+    private String name;
+
+    /**
+     * Construct an instance with the required options.
+     *
+     * @param key the index key, which currently must be {@code {_id: 1}}
+     * @param unique whether the index entries must be unique, which currently must be true
+     */
+    public ClusteredIndexOptions(final Bson key, final boolean unique) {
+        this.key = key;
```
stIncMale, line 41: I propose `this.key = notNull("key", key)`.

stIncMale (changes requested):

stIncMale (approved):

rozza (approved):

PR #939, opened 2022-05-17
Use applicationName in ClusterId

jyemin:
JAVA-4619

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/MongoClientTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/MongoClientTest.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


driver-sync/src/test/functional/com/mongodb/client/MongoClientTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test
+    public void shouldIncludeApplicationNameInClusterId() throws InterruptedException {
+        CountDownLatch latch = new CountDownLatch(1);
+        AtomicReference<ClusterId> clusterId = new AtomicReference<>();
+        ClusterListener clusterListener = new ClusterListener() {
+            @Override
+            public void clusterOpening(final ClusterOpeningEvent event) {
+                clusterId.set(event.getClusterId());
+                latch.countDown();
```
stIncMale, line 43: This code (as well the code below that waits and then reads the value) reinvents poor man's `CompletableFuture` and is more complex than it needs be. Let's just use `CompletableFuture` instead.

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/client/MongoClientTest.java
jyemin: One of these years I'll remember this advice before putting up the PR.

stIncMale (approved):

PR #943, opened 2022-05-19
RecordCodecProvider < JDK 17 canary test

rozza:
The RecordCodecProvider uses JDK 17 LTS to compile.
Users maybe using older (non LTS) JDK's where records are supported.

JAVA-4605


jyemin (approved):
LGTM.

PR #944, opened 2022-05-19
Support the existing pojo annotations on records

jyemin:
JAVA-4622

Pros:

1. Less confusion for application writers.  There is no way to rely on the type system to prevent use of the existing annotations on records, so our options to restrict use are documentation, logging, or runtime exceptions 
2. Less namespace pollution with multiple attributes with the same name but different packages

Cons:

1. It's less clear which annotations are supported on records.  Currently only three are.  More could be, but some don't make sense for records
2. It relies on access to private fields in the record, which may be prevented in some deployments.  We could eventually fix this but not until Java 17 is our minimum

Decision:

Based on internal discussion, we've decided to deprecate the new annotations, support the existing ones when applied to the record component itself, and throw an exception if we detect any unsupported annotation or supported annotations applied to an unsupported place.

### DIFF
--- a/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
+++ b/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
@@ -143,6 +143,7 @@ final class RecordCodec<T extends Record> implements Codec<T> {
         private static void validateAnnotations(final RecordComponent component, final int index) {
             validateAnnotationNotPresentOnType(component.getDeclaringRecord(), org.bson.codecs.pojo.annotations.BsonDiscriminator.class);
             validateAnnotationNotPresentOnConstructor(component.getDeclaringRecord(), org.bson.codecs.pojo.annotations.BsonCreator.class);
+            validateAnnotationNotPresentOnMethod(component.getDeclaringRecord(), org.bson.codecs.pojo.annotations.BsonCreator.class);
             validateAnnotationNotPresentOnFieldOrAccessor(component, org.bson.codecs.pojo.annotations.BsonIgnore.class);
             validateAnnotationNotPresentOnFieldOrAccessor(component, org.bson.codecs.pojo.annotations.BsonExtraElements.class);
             validateAnnotationOnlyOnField(component, index, org.bson.codecs.pojo.annotations.BsonId.class);
@@ -165,6 +166,17 @@ final class RecordCodec<T extends Record> implements Codec<T> {
                     throw new CodecConfigurationException(
                             format("Annotation '%s' not supported on record constructors, but found on constructor of '%s'",
                             annotation, clazz.getName()));
+                }
+            }
+        }
+
+        private static void validateAnnotationNotPresentOnMethod(final Class<?> clazz,
+                @SuppressWarnings("SameParameterValue") final Class<BsonCreator> annotation) {
+            for (var method : clazz.getMethods()) {
+                if (method.isAnnotationPresent(annotation)) {
+                    throw new CodecConfigurationException(
+                            format("Annotation '%s' not supported on methods, but found on method '%s' of '%s'",
+                                    annotation, method.getName(), clazz.getName()));
                 }
             }
         }
--- a/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
+++ b/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
@@ -25,6 +25,7 @@ import org.bson.codecs.EncoderContext;
 import org.bson.codecs.RepresentationConfigurable;
 import org.bson.codecs.configuration.CodecConfigurationException;
 import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.codecs.pojo.annotations.BsonCreator;
 import org.bson.codecs.record.annotations.BsonId;
 import org.bson.codecs.record.annotations.BsonProperty;
 import org.bson.codecs.record.annotations.BsonRepresentation;
@@ -61,6 +62,7 @@ final class RecordCodec<T extends Record> implements Codec<T> {
         private final String fieldName;
 
         private ComponentModel(final RecordComponent component, final CodecRegistry codecRegistry, final int index) {
+            validateAnnotations(component, index);
             this.component = component;
             this.codec = computeCodec(component, codecRegistry);
             this.index = index;
@@ -101,21 +103,17 @@ final class RecordCodec<T extends Record> implements Codec<T> {
             }
             return codec;
         }
+
         @SuppressWarnings("deprecation")
         private static String computeFieldName(final RecordComponent component) {
             if (component.isAnnotationPresent(BsonId.class)) {
                 return "_id";
-            } else if (isAnnotationPresentOnField(component,
-                    org.bson.codecs.pojo.annotations.BsonId.class)) {
+            } else if (isAnnotationPresentOnField(component, org.bson.codecs.pojo.annotations.BsonId.class)) {
                 return "_id";
             } else if (component.isAnnotationPresent(BsonProperty.class)) {
                 return component.getAnnotation(BsonProperty.class).value();
-            } else {
-                if (isAnnotationPresentOnField(component,
-                        org.bson.codecs.pojo.annotations.BsonProperty.class)) {
-                    return getAnnotationOnField(component,
-                            org.bson.codecs.pojo.annotations.BsonProperty.class).value();
-                }
+            } else if (isAnnotationPresentOnField(component, org.bson.codecs.pojo.annotations.BsonProperty.class)) {
+                return getAnnotationOnField(component, org.bson.codecs.pojo.annotations.BsonProperty.class).value();
             }
             return component.getName();
         }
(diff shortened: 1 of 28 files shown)


bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -78,12 +79,20 @@ Object getValue(final Record record) throws InvocationTargetException, IllegalAc
...
         }
 
+        @SuppressWarnings("deprecation")
         private static Codec<?> computeCodec(final RecordComponent component, final CodecRegistry codecRegistry) {
             var codec = codecRegistry.get(toWrapper(component.getType()));
-            var bsonRepresentationAnnotation = component.getAnnotation(BsonRepresentation.class);
-            if (bsonRepresentationAnnotation != null) {
+            BsonType bsonRepresentationType = null;
+
+            if (component.isAnnotationPresent(BsonRepresentation.class)) {
```
rozza, line 87: I'd probably DRY this up and use a single method:

```
isAnnotationPresent(component, BsonRepresentation.class, 
 org.bson.codecs.pojo.annotations.BsonRepresentation.class)
```

bson-record-codec/src/test/unit/org/bson/codecs/record/samples/TestRecordWithPojoAnnotations.java
```diff
@@ -0,0 +1,57 @@
...
+import org.bson.codecs.pojo.annotations.BsonRepresentation;
+
+import java.util.List;
+
+public record TestRecordWithPojoAnnotations(String name,
+                                            @BsonProperty("a") int age,
+                                            List<String> hobbies,
+                                            @BsonRepresentation(BsonType.OBJECT_ID) @BsonId String identifier) {
+
+    public TestRecordWithPojoAnnotations(final String name, final int age, final List<String> hobbies, final String identifier) {
```
rozza, line 31: Can I annotate things in the constructor?

bson-record-codec/src/test/unit/org/bson/codecs/record/samples/TestRecordWithPojoAnnotations.java
```diff
@@ -0,0 +1,57 @@
...
+        this.name = name;
+        this.age = age;
+        this.hobbies = hobbies;
+        this.identifier = identifier;
+    }
+
+    @Override
+    public String name() {
```
rozza, line 39: Is the getter used when serializing? Can I annotate this?

rozza (commented):
I'm not sold on this. I think we're asking for trouble because we don't support all annotations in:
`org.bson.codecs.pojo.annotations` - and that could equally cause user pain. Does it imply you can use the annotations in the same way when using records. 

I'd rather users learn not to use pojo annotations for records by throwing an error. It decouples the pojo code to the records code, and allows the features to work and expand separately.

bson-record-codec/src/test/unit/org/bson/codecs/record/samples/TestRecordWithPojoAnnotations.java
jyemin: The compiler won't prevent it but in this WIP those annotations will be ignored.

bson-record-codec/src/test/unit/org/bson/codecs/record/samples/TestRecordWithPojoAnnotations.java
jyemin: The accessor/getter is used when serializing, and, as above, the compiler won't (can't really) prevent it from being annotated, but in this WIP those annotations will be ignored.  The other choices are:

1. Honor them (but then we have to deal with precedence, like if a different annotation is applied to the accessor, constructor parameter, and in the future, the record component.
2. Detect and throw

jyemin: Decided to detect and throw

bson-record-codec/src/test/unit/org/bson/codecs/record/samples/TestRecordWithPojoAnnotations.java
jyemin: Now it detects and throws.

bson-record-codec/src/main/org/bson/codecs/record/annotations/BsonId.java
```diff
@@ -26,7 +26,9 @@
  * An annotation that configures the record component as the _id field of the document
  *
  * @since 4.6
+ * @deprecated Prefer {@link org.bson.codecs.pojo.annotations.BsonId}
```
stIncMale, line 29: I propose updating the documentation for all annotations in `org.bson.codecs.pojo.annotations` to reflect the following:

- Whether the annotation may be used with a record.
- How to use it with a record (the only valid place for currently supported annotations is a record component).

Without such docs, runtime exceptions is the only way for users to learn.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -125,16 +123,78 @@ private static <T extends Annotation> boolean isAnnotationPresentOnField(final R
...
                 return component.getDeclaringRecord().getDeclaredField(component.getName()).getAnnotation(annotation);
             } catch (NoSuchFieldException e) {
-                throw new CodecConfigurationException("Unexpectedly missing the declared field", e);
+                throw new AssertionError(format("Unexpectedly missing the declared field for recordComponent %s", component), e);
+            }
+        }
+
+        private static void validateAnnotations(final RecordComponent component, final int index) {
+            validateAnnotationNotPresentOnType(component.getDeclaringRecord(), org.bson.codecs.pojo.annotations.BsonDiscriminator.class);
+            validateAnnotationNotPresentOnConstructor(component.getDeclaringRecord(), org.bson.codecs.pojo.annotations.BsonCreator.class);
```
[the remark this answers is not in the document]
jyemin: That does seem absurd.  Also, it wouldn't prevent `BsonCreator` from being added to any other methods.  Do you think we should add checks for all methods?

bson-record-codec/src/main/org/bson/codecs/record/annotations/BsonId.java
jyemin: Done

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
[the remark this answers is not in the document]
jyemin: I added a validation preventing BsonCreator on any method (static or otherwise, as it was easier that way)

stIncMale (approved):

bson/src/main/org/bson/codecs/pojo/annotations/BsonProperty.java
```diff
@@ -27,7 +27,8 @@
 /**
  * An annotation that configures a property.
  *
- * <p>Note: Requires the {@link org.bson.codecs.pojo.Conventions#ANNOTATION_CONVENTION}</p>
+ * <p>For POJOs, requires the {@link org.bson.codecs.pojo.Conventions#ANNOTATION_CONVENTION}</p>
+ * <p>For Java records, the annotation is only supported on the record component.</p>
  *
  * @since 3.5
  * @see org.bson.codecs.pojo.Conventions#ANNOTATION_CONVENTION
```
rozza, line 34: The value() method documentation should also be updated in relation to records.

bson-record-codec/src/test/unit/org/bson/codecs/record/samples/TestRecord.java
```diff
@@ -23,6 +23,7 @@
 
 import java.util.List;
 
+@SuppressWarnings("deprecation")
 public record TestRecord(String name,
```
rozza, line 27: Suggest rename to `TestRecordWithDeprecatedAnnotations`

bson/src/main/org/bson/codecs/pojo/annotations/BsonProperty.java
jyemin: I'm not sure what to say on the `value` method.  The explanatory paragraph would only apply to POJOs because there can't be asymmetrical-ness on records.  

Can you make a suggestion so I know what you're looking for?

rozza: Something like this?

```
    /**
     * The name of the property.
     *
     * <p>
     *     <strong>Note:</strong> Regarding POJOs:<br />
     *     For asymmetrical property names, the context of the {@code BsonProperty} can be important.
     *     For example, when used with {@code @BsonCreator} the value will relate to the read name.
     *     When used directly on a field it will set both the read name if unset and the write name if unset.
     * </p>
     *
     * @return the name to use for the property
     * @see For POJOs see: PropertyModel#getWriteName()
     * @see For POJOs see: PropertyModel#getReadName()
     */
```

bson-record-codec/src/test/unit/org/bson/codecs/record/samples/TestRecord.java
jyemin: Good idea.  Done.  Also replaced most usage of it with `TestRecordWithPojoAnnotations`

PR #945, opened 2022-05-20
Fix Scala case class macro codec lookup bug

rozza:
Fixes an issue when determining the codec to use for a
case class that contains a Map type that has a value that
contains the same key name and field name.

JAVA-4611


rozza:
Once the typeArgs list is gathered in the `decode` method the `fieldTypeArgsMap` are no longer required and in fact caused this bug.

jyemin (approved):
LGTM

stIncMale (approved):
I barely understand the description of the problem, and definitely don't understand the code change (I am sure with regard to the code change, the Java experience would have been the same, as I am not familiar with the codec-related code). Fortunately, the changes only delete the existing code:)

PR #946, opened 2022-05-20
Add wallTime property to ChangeStreamDocument

jyemin:
JAVA-4590


driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -71,6 +73,7 @@
      * @param updateDescription the update description
      * @param txnNumber the transaction number
      * @param lsid the identifier for the session associated with the transaction
+     * @param wallTime the wall time of the server at the moment the change occurred
```
jyemin, line 76: I wasn't quite sure how to phrase this.  I like it better than what's in the spec though:

> The wall time from the mongod that the change event originated from

stIncMale (approved):

PR #947, opened 2022-05-20
Remove unnecessary buffer allocation in IdHoldingBsonWriter

jyemin:
JAVA-4608

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/IdHoldingBsonWriter.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/IdHoldingBsonWriter.java
@@ -82,11 +82,11 @@ public class IdHoldingBsonWriter extends LevelCountingBsonWriter {
     @Override
     public void writeEndDocument() {
         if (isWritingId()) {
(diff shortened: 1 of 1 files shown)


driver-core/src/main/com/mongodb/internal/connection/IdHoldingBsonWriter.java
```diff
@@ -414,23 +414,27 @@ private void setCurrentFieldName(final String name) {
...
     }
 
+    private int getIdBsonWriterCurrentLevel() {
+        return idBsonBinaryWriter == null ? -1 : idBsonBinaryWriter.getCurrentLevel();
+    }
+
     private LevelCountingBsonWriter getIdBsonWriter() {
         if (idBsonBinaryWriter == null) {
-            outputBuffer = new BasicOutputBuffer();
+            outputBuffer = new BasicOutputBuffer(128);
```
stIncMale, line 437: This has the potential to affect performance both positively and negatively. But I agree that 128 bytes is likely a better estimation for the size of the `_id` value than 1024 bytes.

stIncMale (approved):

PR #948, opened 2022-05-20
Use Cleaner for DBCursor cleanup when runtime is Java 9+

jyemin:
JAVA-4612

I don't see a way to regression test this, given the nature of Java garbage collection.  I've proven to myself that it works in the debugger.

### DIFF
--- a/driver-legacy/src/main/com/mongodb/Java8DBCursorCleaner.java
+++ b/driver-legacy/src/main/com/mongodb/Java8DBCursorCleaner.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


driver-legacy/src/main/com/mongodb/Java8DBCursorCleaner.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * A {@link Object#finalize()}-based implementation of {@link DBCursorCleaner}.
+ * {@link DBCursorCleaner#create(MongoClient, MongoNamespace, ServerCursor)} is responsible for ensuring
+ * that this class is only used if the {@code java.lang.ref.Cleaner} class is not available
+ * (i.e. the runtime is Java 8).
+ */
+final class Java8DBCursorCleaner extends DBCursorCleaner {
+    private final MongoClient mongoClient;
+    private final MongoNamespace namespace;
+    private ServerCursor serverCursor;
```
stIncMale, line 30: I think that both here and in `CleanerState`, the field `serverCursor` must be `volatile`. The finalization mechanism / `Cleaner` does not guarantee that writes to the object that is being finalized / the object representing cleaning state & action happen before the execution of `finalize()` / cleaning action. At first I was afraid that our current implementation has the same problem, but it does not: `OptionalFinalizer.serverCursor` is `volatile`.

stIncMale (changes requested):

driver-legacy/src/main/com/mongodb/Java8DBCursorCleaner.java
jyemin: I think you're right.  That was careless of me.

Done.

stIncMale (approved):
👍

PR #949, opened 2022-05-21
Support extra elements in ChangeStreamDocument

jyemin:
JAVA-4580

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
+++ b/driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
@@ -25,6 +25,7 @@ import org.bson.BsonTimestamp;
 import org.bson.codecs.Codec;
 import org.bson.codecs.configuration.CodecRegistry;
 import org.bson.codecs.pojo.annotations.BsonCreator;
+import org.bson.codecs.pojo.annotations.BsonExtraElements;
 import org.bson.codecs.pojo.annotations.BsonId;
 import org.bson.codecs.pojo.annotations.BsonIgnore;
 import org.bson.codecs.pojo.annotations.BsonProperty;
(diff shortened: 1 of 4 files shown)


driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -58,6 +59,8 @@
     private final BsonInt64 txnNumber;
     private final BsonDocument lsid;
     private final BsonDateTime wallTime;
+    @BsonExtraElements
```
jyemin, line 62: There are a couple of annotations that aren't allowed to target constructor parameters.  I'm not sure why, but that's why this annotation is here rather than on the constructor

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -89,7 +93,8 @@ public ChangeStreamDocument(@BsonProperty("operationType") final String operatio
             @Nullable @BsonProperty("updateDescription") final UpdateDescription updateDescription,
             @Nullable @BsonProperty("txnNumber") final BsonInt64 txnNumber,
             @Nullable @BsonProperty("lsid") final BsonDocument lsid,
-            @Nullable @BsonProperty("wallTime") final BsonDateTime wallTime) {
+            @Nullable @BsonProperty("wallTime") final BsonDateTime wallTime,
+            @Nullable @BsonProperty final BsonDocument extraElements) {
```
jyemin, line 97: I was a bit surprised that this worked to combine this annotation with no name here with the one on the field, but it does.

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
rozza: Feel free to fix or add a ticket so they can be.

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
jyemin: I'm not sure.  I just know that every parameter to the `BsonCreator`-annotated constructor has to be annotated with `BsonProperty` or else it throws.

PR #951, opened 2022-05-25
Improve efficiency of ServerSessionPool

jyemin:
* Replace use of ConcurrentPool with direct use of ConcurrentLinkedDeque
* Only prune on ServerSessionPool#get
* Cache ClusterDescription#logicalSessionTimeoutMinutes

JAVA-4629

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/ClusterDescription.java
+++ b/driver-core/src/main/com/mongodb/connection/ClusterDescription.java
@@ -44,6 +44,7 @@ public class ClusterDescription {
     private final ClusterSettings clusterSettings;
     private final ServerSettings serverSettings;
     private final MongoException srvResolutionException;
+    private final Integer logicalSessionTimeoutMinutes;
 
     /**
      * Creates a new ClusterDescription.
@@ -97,6 +98,7 @@ public class ClusterDescription {
         this.serverDescriptions = new ArrayList<ServerDescription>(serverDescriptions);
         this.clusterSettings = clusterSettings;
         this.serverSettings = serverSettings;
+        this.logicalSessionTimeoutMinutes = calculateLogicalSessionTimeoutMinutes();
     }
 
     /**
@@ -237,22 +239,7 @@ public class ClusterDescription {
      */
     @Nullable
     public Integer getLogicalSessionTimeoutMinutes() {
-        Integer retVal = null;
-
-        for (ServerDescription cur : getServersByPredicate(this, serverDescription ->
-                serverDescription.isPrimary() || serverDescription.isSecondary())) {
-
-            Integer logicalSessionTimeoutMinutes = cur.getLogicalSessionTimeoutMinutes();
-            if (logicalSessionTimeoutMinutes == null) {
-                return null;
-            }
-            if (retVal == null) {
-                retVal = logicalSessionTimeoutMinutes;
-            } else {
-                retVal = Math.min(retVal, logicalSessionTimeoutMinutes);
-            }
-        }
-        return retVal;
+        return logicalSessionTimeoutMinutes;
     }
 
     @Override
@@ -332,5 +319,24 @@ public class ClusterDescription {
         }  else {
             return format("{type=%s, srvResolutionException=%s, servers=[%s]", type, srvResolutionException, serverDescriptions);
         }
+    }
+
+    private Integer calculateLogicalSessionTimeoutMinutes() {
+        Integer retVal = null;
+
+        for (ServerDescription cur : getServersByPredicate(this, serverDescription ->
+                serverDescription.isPrimary() || serverDescription.isSecondary())) {
+
+            Integer logicalSessionTimeoutMinutes = cur.getLogicalSessionTimeoutMinutes();
+            if (logicalSessionTimeoutMinutes == null) {
+                return null;
+            }
+            if (retVal == null) {
+                retVal = logicalSessionTimeoutMinutes;
+            } else {
+                retVal = Math.min(retVal, logicalSessionTimeoutMinutes);
+            }
+        }
+        return retVal;
     }
 }
--- a/driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
@@ -22,8 +22,6 @@ import com.mongodb.ServerApi;
 import com.mongodb.connection.ServerDescription;
 import com.mongodb.internal.IgnorableRequestContext;
 import com.mongodb.internal.connection.Cluster;
-import com.mongodb.internal.connection.ConcurrentPool;
-import com.mongodb.internal.connection.ConcurrentPool.Prune;
 import com.mongodb.internal.connection.Connection;
 import com.mongodb.internal.connection.NoOpSessionContext;
 import com.mongodb.internal.selector.ReadPreferenceServerSelector;
@@ -43,23 +41,20 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static com.mongodb.assertions.Assertions.isTrue;
-import static com.mongodb.internal.connection.ConcurrentPool.INFINITE_SIZE;
 import static java.util.concurrent.TimeUnit.MINUTES;
 
 public class ServerSessionPool {
-    private static final int END_SESSIONS_BATCH_SIZE = 10000;
-
-    private final ConcurrentPool<ServerSessionImpl> serverSessionPool =
-            new ConcurrentPool<>(INFINITE_SIZE, new ServerSessionItemFactory());
+    private final ConcurrentLinkedDeque<ServerSessionImpl> available = new ConcurrentLinkedDeque<>();
     private final Cluster cluster;
     private final ServerSessionPool.Clock clock;
-    private volatile boolean closing;
     private volatile boolean closed;
-    private final List<BsonDocument> closedSessionIdentifiers = new ArrayList<>();
     @Nullable
     private final ServerApi serverApi;
+    private final AtomicInteger inUseCount = new AtomicInteger();
 
     interface Clock {
         long millis();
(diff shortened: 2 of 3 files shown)


driver-core/src/main/com/mongodb/connection/ClusterDescription.java
```diff
@@ -97,6 +98,7 @@ public ClusterDescription(final ClusterConnectionMode connectionMode, final Clus
         this.serverDescriptions = new ArrayList<ServerDescription>(serverDescriptions);
         this.clusterSettings = clusterSettings;
         this.serverSettings = serverSettings;
+        this.logicalSessionTimeoutMinutes = calculateLogicalSessionTimeoutMinutes();
```
jyemin, line 101: Otherwise this is calculated over, and over, and over...

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -43,23 +41,20 @@
...
-            new ConcurrentPool<>(INFINITE_SIZE, new ServerSessionItemFactory());
+    private final ConcurrentLinkedDeque<ServerSessionImpl> available = new ConcurrentLinkedDeque<>();
     private final Cluster cluster;
     private final ServerSessionPool.Clock clock;
-    private volatile boolean closing;
     private volatile boolean closed;
-    private final List<BsonDocument> closedSessionIdentifiers = new ArrayList<>();
     @Nullable
     private final ServerApi serverApi;
+    private final AtomicInteger inUseCount = new AtomicInteger();
```
jyemin, line 57: ConcurrentPool used to do this for us.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -77,57 +72,46 @@ public ServerSessionPool(final Cluster cluster, @Nullable final ServerApi server
 
     public ServerSession get() {
         isTrue("server session pool is open", !closed);
-        ServerSessionImpl serverSession = serverSessionPool.get();
-        while (shouldPrune(serverSession)) {
-            serverSessionPool.release(serverSession, true);
-            serverSession = serverSessionPool.get();
+        ServerSessionImpl serverSession = available.pollLast();
```
jyemin, line 75: As discussed, keep pruning until there is a non-stale session or the pool is empty

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -77,57 +72,46 @@ public ServerSessionPool(final Cluster cluster, @Nullable final ServerApi server
...
+        inUseCount.incrementAndGet();
         return serverSession;
     }
 
     public void release(final ServerSession serverSession) {
-        serverSessionPool.release((ServerSessionImpl) serverSession);
-        serverSessionPool.prune();
+        inUseCount.decrementAndGet();
+        ServerSessionImpl serverSessionImpl = (ServerSessionImpl) serverSession;
+        if (serverSessionImpl.isMarkedDirty()) {
```
jyemin, line 90: Dirtiness is the only thing we should check here.   The likelihood of a just-release session already being stale is basically zero.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -77,57 +72,46 @@ public ServerSessionPool(final Cluster cluster, @Nullable final ServerApi server
...
+        if (serverSession == null) {
+            serverSession = new ServerSessionImpl();
         }
+        inUseCount.incrementAndGet();
         return serverSession;
     }
 
     public void release(final ServerSession serverSession) {
-        serverSessionPool.release((ServerSessionImpl) serverSession);
-        serverSessionPool.prune();
```
jyemin, line 90: No more pruning in release

driver-core/src/test/unit/com/mongodb/internal/session/ServerSessionPoolSpecification.groovy
```diff
@@ -253,35 +207,22 @@ class ServerSessionPoolSpecification extends Specification {
...
         for (def cur : sessions) {
             pool.release(cur)
         }
-        pool.release(oneOverBatchSizeSession)
 
         when:
         pool.close()
 
         then:
-        // first batch is the first 10K sessions, final batch is the last one
```
jyemin, line 272: The whole batch size thing is silly.  10K sessions doesn't even come close to reaching 16M command document size limit.  And it's a SHOULD in the spec, so I think it's ok to rip it out.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -77,57 +72,46 @@ public ServerSessionPool(final Cluster cluster, @Nullable final ServerApi server
...
+        inUseCount.decrementAndGet();
+        ServerSessionImpl serverSessionImpl = (ServerSessionImpl) serverSession;
+        if (serverSessionImpl.isMarkedDirty()) {
+            serverSessionImpl.close();
+        } else {
+            available.addLast(serverSessionImpl);
```
rozza, line 93: Does adding to last make it more likely that an earlier session in the deque will be left in a should be pruned state but never pruned?

rozza (commented):
Just one question

driver-core/src/main/com/mongodb/connection/ClusterDescription.java
jyemin: I considered that, but new ClusterDescription instances are always created by the monitor threads so they don't affect the critical path.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
jyemin: Yes.  But not just for unit tests, also for the assertion we make after every integration test to insure that no sessions are leaked.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
jyemin: This is not a change in behavior.  ConcurrentPool also polls last and adds to last. 

What makes it more likely that an earlier session will not be pruned  is the fact that we're no longer pruning in #release, but only in #get.  But there is an upper bound on the size of the pool, as @stIncMale explained to me: it's bounded by the total number of available connections (because we only allocate a new session after checking out a connection).  I think this is ok because each item in the pool is quite small, so even if there are thousands of stale ones it's not likely to be noticed.

rozza: That's fine - so by adding it to the end of the queue you make it more likely a user has a non-stale session. And as soon as they do they will all be cleaned.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -77,57 +72,46 @@ public ServerSessionPool(final Cluster cluster, @Nullable final ServerApi server
...
-        } finally {
+        if (!closed) {
             closed = true;
+            List<BsonDocument> identifiers = new ArrayList<>(available.size());
+            ServerSessionImpl nextSession = available.pollFirst();
+            while (nextSession != null) {
+                nextSession.close();
+                identifiers.add(nextSession.getIdentifier());
+                nextSession = available.pollFirst();
+            }
```
stIncMale, line 106: It appears that this code may be replaced with
```java
List<BsonDocument> identifiers = new ArrayList<>(available.size());
available.forEach(session -> {
    session.close();
    identifiers.add(session.getIdentifier());
});
```
`forEach` has a potential to be more efficient, and also it is shorter :)

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -77,57 +72,46 @@ public ServerSessionPool(final Cluster cluster, @Nullable final ServerApi server
...
+            if (identifiers.size() > 0) {
+                endClosedSessions(identifiers);
```
stIncMale, line 108: `endClosedSessions` already has a similar check, which makes it redundant here.

P.S. While for `ArrayList` the approaches `identifiers.size() > 0` and `!identifiers.isEmpty()` are identical, in general, `identifiers.size() > 0` is never better than `!identifiers.isEmpty()`. For example, `ConcurrentLinkedDeque.size()` is linear, while `ConcurrentLinkedDeque.isEmpty` is constant (both statements are true only when there are no concurrent modifications).

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -232,32 +223,4 @@ public boolean isMarkedDirty() {
...
-
-    private final class ServerSessionItemFactory implements ConcurrentPool.ItemFactory<ServerSessionImpl> {
-        @Override
-        public ServerSessionImpl create() {
-            return new ServerSessionImpl(createNewServerSessionIdentifier());
-        }
-
-        @Override
-        public void close(final ServerSessionImpl serverSession) {
-            closeSession(serverSession);
```
stIncMale, line 244: Previously, we were requesting the server to release its resources for closed sessions (not on every `close`, only when we reach `END_SESSIONS_BATCH_SIZE`, but nonetheless). Now we don't do that at all while the pool is used, and only ask the server to release resources for those session that are still in the pool when the pool is being closed. I don't think that this is really what we want, unless I am misunderstanding the situation.

Unless we are going to get rid of `endClosedSessions`, we also need to make sure that sessions released to the pool concurrently with it being closed do not stay unclosed, and that the server also gets notified about them. The version of the pool before the PR does this, but I have not tried to assess the correctness.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -77,57 +72,46 @@ public ServerSessionPool(final Cluster cluster, @Nullable final ServerApi server
...
-
-            List<BsonDocument> identifiers;
-            synchronized (this) {
-                identifiers = new ArrayList<>(closedSessionIdentifiers);
-                closedSessionIdentifiers.clear();
-            }
-            endClosedSessions(identifiers);
-        } finally {
+        if (!closed) {
             closed = true;
```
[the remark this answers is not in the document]
jyemin: I don't expect it to be called more than once, let alone concurrently, so I supposed this is just defensive coding and can be removed.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
jyemin: The previous code was very confusing and you may be misinterpreting it.  Note that `closeSession` was a no-op unless the pool itself was in the process of being closed (see the comment `only track closed sessions when pool is in the process of closing`).

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
jyemin: Done

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
jyemin: Removed redundant check

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
stIncMale: True, I confused myself. Thanks to explaining to me what is done and why over Zoom.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
```diff
@@ -77,60 +72,40 @@ public ServerSessionPool(final Cluster cluster, @Nullable final ServerApi server
...
-        }
+    public void close() {
+        closed = true;
+        endClosedSessions();
     }
 
-    private void endClosedSessions(final List<BsonDocument> identifiers) {
+    private void endClosedSessions() {
+        List<BsonDocument> identifiers = drainPool();
+        available.clear();
```
stIncMale, line 108: Given that we have just drained `available`, we don't need to clear it.

stIncMale (commented):
A tiny thing in the new code, otherwise approved. This is a nice simplification.

driver-core/src/main/com/mongodb/internal/session/ServerSessionPool.java
jyemin: Done

PR #954, opened 2022-05-31
Add builders for pick-accumulators / window functions

stIncMale:
Note that the links to server docs from API docs are not currently working. They will start working when MongoDB 6.0 become the "current" version. Depending when we anticipate 4.7 driver release and 6.0 MongoDB release, this may either be fine, or we may need to create a DOCSP request to prepare temporary https://dochub.mongodb.org/ URIs.

JAVA-4094

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/Accumulators.java
+++ b/driver-core/src/main/com/mongodb/client/model/Accumulators.java
@@ -497,7 +497,7 @@ public final class Accumulators {
         return new BsonField(fieldName, new Document(accumulatorName, new Document("input", inExpression).append("n", nExpression)));
     }
 
-    private static <OutExpression, NExpression> BsonField sortingPickAccumulator(
+    private static <OutExpression> BsonField sortingPickAccumulator(
             final String fieldName, final String accumulatorName, final Bson sort, final OutExpression outExpression) {
         return new BsonField(fieldName, new Document(accumulatorName, new Document("sortBy", sort).append("output", outExpression)));
     }
--- a/driver-core/src/main/com/mongodb/client/model/Accumulators.java
+++ b/driver-core/src/main/com/mongodb/client/model/Accumulators.java
@@ -20,10 +20,13 @@ import com.mongodb.lang.Nullable;
 import org.bson.BsonArray;
 import org.bson.BsonDocument;
 import org.bson.BsonString;
+import org.bson.Document;
+import org.bson.conversions.Bson;
 
 import java.util.List;
 
 import static java.util.stream.Collectors.toList;
+import static org.bson.assertions.Assertions.notNull;
 
 /**
  * Builders for accumulators used in the group pipeline stage of an aggregation pipeline.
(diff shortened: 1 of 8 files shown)


driver-core/src/main/com/mongodb/client/model/Accumulators.java
```diff
@@ -321,6 +492,24 @@ private static <TExpression> BsonField accumulatorOperator(final String name, fi
         return new BsonField(fieldName, new SimpleExpression<TExpression>(name, expression));
     }
 
+    private static <InExpression, NExpression> BsonField pickNAccumulator(
+            final String fieldName, final String accumulatorName, final InExpression inExpression, final NExpression nExpression) {
+        return new BsonField(fieldName, new Document(accumulatorName, new Document("input", inExpression).append("n", nExpression)));
+    }
+
+    private static <OutExpression, NExpression> BsonField sortingPickAccumulator(
```
jyemin, line 500: `NExpression` is unused and can be removed.

driver-core/src/main/com/mongodb/client/model/Accumulators.java
```diff
@@ -321,6 +492,24 @@ private static <TExpression> BsonField accumulatorOperator(final String name, fi
         return new BsonField(fieldName, new SimpleExpression<TExpression>(name, expression));
     }
 
+    private static <InExpression, NExpression> BsonField pickNAccumulator(
+            final String fieldName, final String accumulatorName, final InExpression inExpression, final NExpression nExpression) {
+        return new BsonField(fieldName, new Document(accumulatorName, new Document("input", inExpression).append("n", nExpression)));
```
jyemin, line 497: For builders and operations, we typically have preferred `BsonDocument` to  `Document`, since the former is less susceptible to user-controlled changes to the codec registry.

We should use `BsonDocument` here unless there is a good reason not to, even though doing so adds a bit of  implementation complexity (see `Aggregates.SimplePipelineStage` for an example of this).

jyemin (changes requested):
Looks great.  Just one significant change request.

driver-core/src/main/com/mongodb/client/model/Accumulators.java
stIncMale: Right, we can and probably should use `BsonDocument` when the types we have are compatible with it. However, that's not the case here: all three new methods `pickNAccumulator`, `sortingPickAccumulator`, `sortingPickNAccumulator` have to put any type (an expression) inside a `BsonField`, so they have to rely on codec registry. `Aggregates.SimplePipelineStage` does the same [here](https://github.com/mongodb/mongo-java-driver/blob/9fcdefd48faaa9c7ab43ee16c8c215ce05453c2e/driver-core/src/main/com/mongodb/client/model/Aggregates.java#L835). The approach I used in this PR is no different from the one I used in search API.

driver-core/src/main/com/mongodb/client/model/Accumulators.java
stIncMale: Thanks, done.

driver-core/src/main/com/mongodb/client/model/Accumulators.java
jyemin: I agree that it's the same as in Search API, but I don't think it's the same as in SimplePipelineStage.   In that case the `Bson` instance came from the application.  In this case the driver is creating the `Document`.

Since we're already doing this elsewhere, I won't object further here.

jyemin (approved):
LGTM

rozza (approved):
LGTM!

PR #955, opened 2022-05-31
Added support for client side encryption shared library

rozza:
JAVA-4459


driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
```diff
@@ -41,25 +40,40 @@
...
-            final boolean bypassQueryAnalysis) {
+            @Nullable final Map<String, BsonDocument> encryptedFieldsMap) {
         MongoCryptOptions.Builder mongoCryptOptionsBuilder = MongoCryptOptions.builder();
-
-        BsonDocument bsonKmsProviders = getKmsProvidersAsBsonDocument(kmsProviders);
-        mongoCryptOptionsBuilder.kmsProviderOptions(bsonKmsProviders);
-        mongoCryptOptionsBuilder.needsKmsCredentialsStateEnabled(true);
+        mongoCryptOptionsBuilder.kmsProviderOptions(getKmsProvidersAsBsonDocument(kmsProviders));
+        mongoCryptOptionsBuilder.bypassQueryAnalysis(bypassQueryAnalysis);
+        mongoCryptOptionsBuilder.searchPaths(searchPaths);
```
jyemin, line 72: Are these two lines the only functional changes in this class?  It's hard to tell with all the refactoring mixed in.

jyemin (changes requested):
- I had some trouble distinguishing between the functional and non-functional changes included in this PR.  In the future I'd prefer that refactoring be done separately (even without a PR in some cases).  It makes the review process much more straightforward.
- Regarding testing, I'm not clear how we are ensuring that both the shared library and mongocryptd usage are both tested.  Can you clarify?
- Also regarding testing, I'm not clear whether/how we are testing both the search path and the search path override.
-

driver-core/src/main/com/mongodb/AutoEncryptionSettings.java
```diff
@@ -480,6 +515,7 @@ private AutoEncryptionSettings(final Builder builder) {
         this.bypassAutoEncryption = builder.bypassAutoEncryption;
         this.encryptedFieldsMap = builder.encryptedFieldsMap;
         this.bypassQueryAnalysis = builder.bypassQueryAnalysis;
+        this.searchPaths = !bypassAutoEncryption && builder.searchPaths.isEmpty() ? singletonList("$SYSTEM") : builder.searchPaths;
```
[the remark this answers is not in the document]
rozza: Done

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
rozza: Functionally these are the two new options passed to mongoCryptOptionsBuilder.  

The refactorings were mainly to support this:

`toBsonDocument` was extracted to handle multiple wrapping of maps - which then could be used in fetchCredentials.

The other refactorting is simplifying the public `createMongoCryptOptions` options by passing in the actual settings and configuring the defaults here - and making a private createMongoCryptOptions method with all the arguments that can be used.

The only drive by refactoring was using closures in the creation of MongoClientSettings.

driver-core/src/main/com/mongodb/AutoEncryptionSettings.java
```diff
@@ -229,6 +241,15 @@ public Builder bypassQueryAnalysis(final boolean bypassQueryAnalysis) {
             return this;
         }
 
+        /**
+         * Internal API may be removed at any time.
+         */
+        @VisibleForTesting(otherwise = VisibleForTesting.AccessModifier.PRIVATE)
```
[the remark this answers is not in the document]
jyemin: I see now, after reading the relevant section of the specification.   But in that case can we do without it (as our test runner can just as easily use the override path in extraOptions to specify the full path)?

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
jyemin: OK, I see your point.

driver-core/src/main/com/mongodb/AutoEncryptionSettings.java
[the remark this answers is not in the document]
rozza: Yeah, I think so let me have a look at doing that.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionTest.java
```diff
@@ -258,8 +258,12 @@ public void setUp() {
                         .addCommandListener(commandListener);
 
         if (!kmsProvidersMap.isEmpty()) {
-            String csflePath = System.getProperty("org.mongodb.test.csfle.path", "");
-            List<String> searchPath = csflePath.isEmpty() ? emptyList() : singletonList(csflePath);
+            String cryptSharedLibPath = System.getProperty("org.mongodb.test.crypt.shared.lib.path", "");
+            if (!cryptSharedLibPath.isEmpty()) {
+                extraOptions = new HashMap<>(extraOptions);
```
jyemin, line 263: Seems wrong to throw away any existing extraOptions by creating a new map here.
I'd be surprised if this doesn't cause a bunch of tests to fail.

jyemin:
> Here I only test the shared library in instances of AbstractClientSideEncryptionTest so all other auto-encryption tests remain testing against mongocryptd

I think that's just the prose tests.  Do you think that's enough?  I'm not sure.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionTest.java
jyemin: Oh, missed the copy.

PR #956, opened 2022-05-31
Remove copy of ConcurrentLinkedDeque

jyemin:
* Rely on java.util.concurrent.ConcurrentLinkedDeque
* Revert ConcurrentPool#prune to previous algorithm that doesn't
  depend on changes to our ConcurrentLinkedDeque copy

JAVA-4630


driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
```diff
@@ -192,23 +179,21 @@ T getImmediateUnfair() {
     }
 
     public void prune() {
-        for (RemovalReportingIterator<T> iter = available.iterator(); iter.hasNext();) {
```
jyemin, line 195: This is the only functional behavior change.  It reverts to the previous behavior of this method.  I think it's still correct according to the requirements of the connection pool specification, but want to confirm that with the team.

stIncMale: The spec says [_"Remove and close perished available Connections"_](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#:~:text=Remove%20and%20close%20perished%20available%20Connections.), and nothing else about this. So the change must be fine.

driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
```diff
@@ -192,23 +179,21 @@ T getImmediateUnfair() {
...
-            T cur = iter.next();
-            Prune shouldPrune = itemFactory.shouldPrune(cur);
-
-            if (shouldPrune == Prune.STOP) {
+        int currentAvailableCount = getAvailableCount();
+        for (int numAttempts = 0; numAttempts < currentAvailableCount; numAttempts++) {
+            if (!stateAndPermits.acquirePermit(10, TimeUnit.MILLISECONDS)) {
```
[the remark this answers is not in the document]
jyemin: Done

stIncMale (approved):
The custom dequeue is no more 🎉

driver-core/src/main/com/mongodb/internal/connection/ConcurrentPool.java
jyemin: Thinking more about this, I'm uncomfortable that this violates the spirit of the spec, which is strangely silent on whether the pool is LIFO or FIFO and how maxIdleTime is actually supposed to work in practice.  And even if the spec provides this latitude, it's still a significant behavior change for the Java driver.

To keep roughly the same behavior, and still use the standard `ConcurrentLinkedDeque` class, what about something like this:

```
        for (T cur : available) {
            if (itemFactory.shouldPrune(cur) && (available.remove(cur)) {
                close(cur);
            }
        }
```

Each remove becomes linear instead of constant, but on the other hand pruning should be rare in practice.

stIncMale (approved):

PR #960, opened 2022-06-06
Annotate Queryable Encryption API as Beta

rozza:
It is in public technical preview and may change before
being finalized.

JAVA-4639


jyemin (approved):

rozza:
Manually merged

PR #962, opened 2022-06-13
Corrected successSubscriber to documentSubscriber for 2 awaits() in the QuickTour.

wbleonard:
The QuickTour was calling await on the wrong subscriber.


jyemin:
Thanks for spotting this.  Merging

PR #963, opened 2022-06-15
Accept `Iterable` instead of `List` in `Aggregates.setWindowFields`

stIncMale:
We have just removed the `@Beta` from this API an have not yet released it without `@Beta`, which means, that regardless of whether this change breaks binary compatibility or not (I don't know and have not tried to find this out), the change is allowed.

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/Aggregates.java
+++ b/driver-core/src/main/com/mongodb/client/model/Aggregates.java
@@ -648,7 +648,7 @@ public final class Aggregates {
      * @since 4.3
      */
     public static <TExpression> Bson setWindowFields(@Nullable final TExpression partitionBy, @Nullable final Bson sortBy,
-                                                     final List<WindowedComputation> output) {
+                                                     final Iterable<? extends WindowedComputation> output) {
         notNull("output", output);
         return new SetWindowFieldsStage<>(partitionBy, sortBy, output);
     }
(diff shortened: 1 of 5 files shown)


driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -1706,9 +1706,12 @@ public String toString() {
...
         @Nullable
         private final Bson sortBy;
-        private final List<WindowedComputation> output;
+        private final Iterable<? extends WindowedComputation> output;
 
-        SetWindowFieldsStage(@Nullable final TExpression partitionBy, @Nullable final Bson sortBy, final List<WindowedComputation> output) {
+        SetWindowFieldsStage(
+                @Nullable final TExpression partitionBy,
+                @Nullable final Bson sortBy,
+                final Iterable<? extends WindowedComputation> output) {
```
stIncMale, line 1714: All changes in this PR except for these were done automatically by the IDE.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -648,7 +648,7 @@ public static <TExpression> Bson setWindowFields(@Nullable final TExpression par
      * @since 4.3
      */
     public static <TExpression> Bson setWindowFields(@Nullable final TExpression partitionBy, @Nullable final Bson sortBy,
-                                                     final List<WindowedComputation> output) {
+                                                     final Iterable<? extends WindowedComputation> output) {
```
stIncMale, line 651: This is the only thing I wanted to change. The rest is just a consequence of this change.

jyemin (commented):
Why is it important that this is an Iterable rather than a List?  Are there use cases you're aware of where it wouldn't be a List?

Asking because we do use List in other places, like MongoCollection#aggregate.

I think the change is fine regardless of the answer, but wanted to get the answer.

stIncMale:
It is trivially better: the API is capable of accepting `Iterable` instead of `List`, and `Iterable` is a wider type, which means, the changed method may be called with arguments whose set of types is a superset of the set of types the original method was accepting. I am not aware of specific cases.

> Asking because we do use List in other places, like MongoCollection#aggregate.

That's true, and I pay attention to style. However, since in this case preserving the style goes against making the API more powerful, the style argument can't have an automatic win (it still may win as a result of a discussion). When I thought whether to accept `List`, `Collection`, `Iterable` in the search API, I picked `Iterable` because it was the widest type the API is capable of accepting. Now I realized that I have a chance to use the same approach with `setWindowFIelds` because of its `@Beta` status.

jyemin (approved):

jyemin:
For the record: it does break binary compatibility, but not source compatibility

stIncMale:
Jeff, thanks for the info, that's useful to remember.

PR #965, opened 2022-06-16
Add query building API for the `$fill` aggregation pipeline stage

stIncMale:
The new API:

- `WindowedComputations.locf` for [`$locf`](https://www.mongodb.com/docs/v6.0/reference/operator/aggregation/locf/)
- `WindowedComputations.linearFill` for [`$linearFill`](https://www.mongodb.com/docs/v6.0/reference/operator/aggregation/linearFill/)
- `Aggregates.fill` for [`$fill`](https://www.mongodb.com/docs/v6.0/reference/operator/aggregation/fill/)

The server-side syntax document is [here](https://docs.google.com/document/d/1pwIBc5XzbZkMnZH-FAoVYf2OOUzQM2VT8wbtoC4FMTg/edit#heading=h.pkrftzkil1l1).

You may see how the API usage looks like, or play with it in

- `AggregatesFunctionalSpecification.'$setWindowFields'`
- `AggregatesFunctionalSpecification.'$fill'`

JAVA-4394

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/Aggregates.java
+++ b/driver-core/src/main/com/mongodb/client/model/Aggregates.java
@@ -662,7 +662,7 @@ public final class Aggregates {
     }
 
     /**
-     * Creates a {@code $fill} pipeline stage, which sets values to fields when they are {@link BsonType#NULL Null} or missing.
+     * Creates a {@code $fill} pipeline stage, which assigns values to fields when they are {@link BsonType#NULL Null} or missing.
      *
      * @param options The fill options.
      * @param output The {@link FillComputation}.
@@ -677,7 +677,7 @@ public final class Aggregates {
     }
 
     /**
-     * Creates a {@code $fill} pipeline stage, which sets values to fields when they are {@link BsonType#NULL Null} or missing.
+     * Creates a {@code $fill} pipeline stage, which assigns values to fields when they are {@link BsonType#NULL Null} or missing.
      *
      * @param options The fill options.
      * @param output The non-empty {@link FillComputation}s.
--- a/driver-core/src/main/com/mongodb/client/model/Aggregates.java
+++ b/driver-core/src/main/com/mongodb/client/model/Aggregates.java
@@ -17,6 +17,8 @@
 package com.mongodb.client.model;
 
 import com.mongodb.MongoNamespace;
+import com.mongodb.client.model.fill.FillComputation;
+import com.mongodb.client.model.fill.FillOptions;
 import com.mongodb.client.model.search.SearchOperator;
 import com.mongodb.client.model.search.SearchCollector;
 import com.mongodb.client.model.search.SearchOptions;
@@ -26,6 +28,7 @@ import org.bson.BsonDocument;
 import org.bson.BsonDocumentWriter;
 import org.bson.BsonInt32;
 import org.bson.BsonString;
+import org.bson.BsonType;
 import org.bson.BsonValue;
 import org.bson.codecs.configuration.CodecRegistry;
 import org.bson.conversions.Bson;
@@ -35,9 +38,11 @@ import java.util.Map;
 import java.util.Objects;
 
 import static com.mongodb.assertions.Assertions.assertTrue;
+import static com.mongodb.assertions.Assertions.isTrueArgument;
+import static com.mongodb.assertions.Assertions.notNull;
 import static com.mongodb.internal.Iterables.concat;
+import static com.mongodb.internal.client.model.Util.sizeAtLeast;
 import static java.util.Arrays.asList;
-import static org.bson.assertions.Assertions.notNull;
 
 /**
  * Builders for aggregation pipeline stages.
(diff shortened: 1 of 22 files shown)


stIncMale:
@katcharov, I added you in case you'd like to express your thoughts about the "debatable" part explained in the description of this PR. Feel free to ignore)

stIncMale:
Moved back to "Ready for review".

stIncMale:
The PR was refactored. A single pipeline stage specification us no longer used. Instead
- the API uses an object that represents optional fields (similarly to `Aggregates.bucket`, `Aggregates.search`; the API of this object has the same style as the API of `SearchOptions`);
- the API uses varargs to represents `output`, similarly to `Aggregates.setWindowFields`.

rozza (commented):
This is a tricky API as the syntax seems to suggest optional values first - which is different to our (Java driver) general approach.

I found the name FillComputation surprising as it represents the output field in the syntax. Did you try any other names? I can only suggest FillOutput then type name is more closely linked to the field it represents.

stIncMale:
> This is a tricky API as the syntax seems to suggest optional values first - which is different to our (Java driver) general approach.

This is only true if we want to accept `output` as varargs. This is not a new situation for us, and we had exactly the same situation in `Aggregates.setWindowFields` which, in turn, was modelled after `Aggregates.group`, that accepts its output via `varargs` (this part is not called `output` in the MQL, which only tells us that the MQL is also inconsistent with itself). One may also observe that `Aggregates.bucket` accepts its output via `BucketOptions.output`, which is different from `Aggregates.group`.

> I found the name FillComputation surprising as it represents the output field in the syntax. Did you try any other names? I can only suggest FillOutput then type name is more closely linked to the field it represents.

We had this exact discussion for `Aggregates.setWindowFields`, because `WindowedComputation` also seemed weird, yet `WindowedOutput`/`WindowOutput` appears to be even weirder. I constructed the name `FillComputation` to be similar to `WindowedComputation`. If we rename `FillComputation` to `FillOutput`, then we should also rename `WindowedComputation` to `WindowedOutput` (or `WindowOutput` 🤷‍♂️). The upcoming release is the last point where we can do this, because `Aggregates.setWindowFields` goes out of beta in 4.7.0.

Ross, is your proposal to rename `WindowedComputation`/`FillComputation` to `WindowedOutput`/`FillOutput`?

rozza:
I agree the naming is difficult / problematic - sorry to rehash old ground. I noticed `setWindowFields` uses the param name `output` for `WindowedComputation`.

Looking at the naming:

* `WindowedComputation` - Relates to the context but uses new language.
* `WindowFieldOutput` - Relates to MQL but awkward to read.
* `WindowOutputField` - Relates to MQL
* `WindowOutput` - Relates to MQL but is very vague (meaningless?).

Personally I think I prefer: `FillOutputField` to `FillComputation` and `WindowOutputField`to `WindowedComputation`.

Regardless of naming I'll rereview in the morning. I'm also happy to delay naming decisions (as a separate ticket) until Jeff is back to see if we can get consensus.

driver-core/src/test/functional/com/mongodb/client/model/AggregatesFunctionalSpecification.groovy
```diff
@@ -1275,4 +1288,59 @@ class AggregatesFunctionalSpecification extends OperationFunctionalSpecification
...
+                new Document('partition', new Document('id', 10))
+                        .append('_id', 2)
+                        .append('doc', null),
+                new Document('partition', new Document('id', 20))
+                        .append('_id', 3)
+                        .append('field1', 3)
+                        .append('doc', new Document('field2', 3))]
+        getCollectionHelper().insertDocuments(docs)
+        String resultField = 'result'
+        LinkedList<Bson> stages = [
```
rozza, line 1310: Q - why linkedlist over just List?

rozza (approved):
LGTM!

I had one question that is just for my learning - no action needed.

driver-core/src/test/functional/com/mongodb/client/model/AggregatesFunctionalSpecification.groovy
stIncMale: Indeed, I could have get by with the `List` interface. Done.

katcharov (approved):
lgtm

Just a trivial doc suggestion.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -656,6 +661,60 @@ public static <TExpression> Bson setWindowFields(@Nullable final TExpression par
...
+     * @mongodb.driver.manual reference/operator/aggregation/fill/ $fill
+     * @mongodb.server.release 5.3
+     * @since 4.7
+     */
+    public static Bson fill(final FillOptions options, final FillComputation output, final FillComputation... moreOutput) {
+        return fill(options, concat(notNull("output", output), moreOutput));
+    }
+
+    /**
+     * Creates a {@code $fill} pipeline stage, which sets values to fields when they are {@link BsonType#NULL Null} or missing.
```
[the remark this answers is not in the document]
stIncMale: Done. This change was also needed in the Scala code.

PR #966, opened 2022-06-17
Use varargs instead of single-value parameters in `$search`/`$setWindowFields` query building API

stIncMale:
This PR applies the following approach:

- Instead of accepting `T` as a single-value parameter in an overloaded method, accept `T...`. This approach supports all usages supported by `T`, and more, i.e., it is more powerful.
  - When at least one value must be provided, use `T first, T... more`.
  - When omitting all values is allowed, use `T...`.

Note that overloads accepting `Iterable<? extends T>` are still present in the API.

Commits are well-organized (don't tend to override each other) and may be reviewed one by one if that's more convenient.

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/Aggregates.java
+++ b/driver-core/src/main/com/mongodb/client/model/Aggregates.java
@@ -35,6 +35,7 @@ import java.util.Map;
 import java.util.Objects;
 
 import static com.mongodb.assertions.Assertions.assertTrue;
+import static com.mongodb.internal.Iterables.concat;
 import static java.util.Arrays.asList;
 import static org.bson.assertions.Assertions.notNull;
 
@@ -614,7 +615,8 @@ public final class Aggregates {
      *               Sorting is required by certain functions and may be required by some windows (see {@link Windows} for more details).
      *               Sorting is used only for the purpose of computing window functions and does not guarantee ordering
      *               of the output documents.
-     * @param output A nonempty array of {@linkplain WindowedComputation windowed computations}.
+     * @param output A {@linkplain WindowedComputation windowed computation}.
+     * @param moreOutput More {@linkplain WindowedComputation windowed computations}.
      * @param <TExpression> The {@code partitionBy} expression type.
      * @return The {@code $setWindowFields} pipeline stage.
      * @mongodb.driver.dochub core/window-functions-set-window-fields $setWindowFields
@@ -622,9 +624,8 @@ public final class Aggregates {
      * @since 4.3
      */
     public static <TExpression> Bson setWindowFields(@Nullable final TExpression partitionBy, @Nullable final Bson sortBy,
-                                                     final WindowedComputation... output) {
-        notNull("output", output);
-        return setWindowFields(partitionBy, sortBy, asList(output));
+            final WindowedComputation output, final WindowedComputation... moreOutput) {
+        return setWindowFields(partitionBy, sortBy, concat(notNull("output", output), moreOutput));
     }
 
     /**
@@ -640,7 +641,8 @@ public final class Aggregates {
      *               Sorting is required by certain functions and may be required by some windows (see {@link Windows} for more details).
      *               Sorting is used only for the purpose of computing window functions and does not guarantee ordering
      *               of the output documents.
-     * @param output A nonempty list of {@linkplain WindowedComputation windowed computations}.
+     * @param output A list of {@linkplain WindowedComputation windowed computations}.
+     * Specifying an empty list is not an error, but the resulting stage does not do anything useful.
      * @param <TExpression> The {@code partitionBy} expression type.
      * @return The {@code $setWindowFields} pipeline stage.
      * @mongodb.driver.dochub core/window-functions-set-window-fields $setWindowFields
--- a/driver-core/src/main/com/mongodb/client/model/search/SearchConstructibleBsonElement.java
+++ b/driver-core/src/main/com/mongodb/client/model/search/SearchConstructibleBsonElement.java
@@ -15,9 +15,7 @@
  */
 package com.mongodb.client.model.search;
 
-import com.mongodb.internal.client.model.AbstractConstructibleBson;
 import com.mongodb.internal.client.model.AbstractConstructibleBsonElement;
-import com.mongodb.lang.Nullable;
 import org.bson.BsonInt32;
 import org.bson.conversions.Bson;
 
@@ -27,9 +25,8 @@ import java.util.stream.StreamSupport;
 
 import static com.mongodb.assertions.Assertions.isTrueArgument;
 import static com.mongodb.assertions.Assertions.notNull;
-import static com.mongodb.internal.client.model.Util.combine;
+import static com.mongodb.client.model.search.FuzzySearchOptions.fuzzySearchOptions;
 import static com.mongodb.internal.client.model.Util.sizeAtLeast;
-import static java.util.Arrays.asList;
 
 final class SearchConstructibleBsonElement extends AbstractConstructibleBsonElement<SearchConstructibleBsonElement> implements
         CompoundSearchOperatorBase, CompoundSearchOperator,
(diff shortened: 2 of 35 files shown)


stIncMale:
Tests fail because `AggregatesFunctionalSpecification.'$setWindowFields with empty output'` fails. The server requires that the `output` document is specified, but allows it to be empty. It is weird, and I can't think of how would specifying empty `output` be useful. I suspect that back when the API was introduced, I added the test just because I observed this server behavior, but I doubt I figured out why the server behaves this way. Tomorrow (Jun 20), when the US employees come back to work, I'll ask the server API authors about this behavior.

PR can be reviewed regardless of the outcome: at the worst, I will roll back the last two commits that touch `setWindowFields` (39e2a219a8bc17ab3700178b6e4d157cdd502235, 3f61318c7f5fc2aa7fa7810aef125c3b1d307237).

driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -681,127 +664,68 @@ public static Bson search(final SearchOperator operator) {
...
-     * or calling {@link #search(SearchOperator)}.
      * @return The {@code $search} pipeline stage.
      *
      * @mongodb.atlas.manual atlas-search/query-syntax/#-search $search
      * @mongodb.atlas.manual atlas-search/operators-and-collectors/#operators Search operators
      * @mongodb.atlas.manual atlas-search/scoring/ Scoring
      * @since 4.7
      */
-    public static Bson search(final SearchOperator operator, @Nullable final SearchOptions options) {
+    public static Bson search(final SearchOperator operator, final SearchOptions... options) {
```
rozza, line 674: This feels hacky and less clear, would you ever expect more that one options here?

stIncMale: This is not to allow specifying `SearchOptions` more than once, but to allow specifying it 0 or 1 times. However, the implementation still processes multiple `SearchOptions` in the unlikely case that they are specified.

rozza: We'd still have to handle it in Scala and as an option type - which is the appropriate idiomatic scala type in this scenario.

I just don't think its very clear that there should be 0 or 1 options. I definitely think it should be an error if multiple options are supplied. 

I think here I'm of the opinion just because you can doesn't mean you should. I don't see a benefit over two explicit methods from the perspective of a user of the API.  I also think the overloads aren't very costly and its us that should be baring that cost as developers and not adding a cognitive cost to users.

stIncMale: > We'd still have to handle it in Scala and as an option type

Scala has varargs, and I am not sure that wrapping a vararg in `Option` is an idiomatic approach.

As for the cost of overloads, I know, I am alone in the camp of those who thinks they are bad because of the duplication in Scala API, and because of the duplicated documentation.

What is your opinion regarding the other thing done in the PR: replacing `T single` with `T first, T... more`?

rozza: > Scala has varargs, and I am not sure that wrapping a vararg in Option is an idiomatic approach.
The idiomatic approach is to look at the correct type for SearchOptions eg: which is expected type: `Seq[SearchOptions]` or `Option[SearchOptions]`

rozza: > What is your opinion regarding the other thing done in the PR: replacing `T single` with `T first, T... more`?

That pattern is more comment and is clearer in the requirement of being 1 or more.

stIncMale: Right, if the type is `Seq[T]`/`Option[T]`, but what if the method uses neither as the parameter type, and the parameter is expressed as `method(args: T*)` instead? If no values are provided, i.e., the method is called as `method()`, then `args` is empty (not `null`) in Scala (in Java it is `null`). So not specifying any values of a vararg parameter in Scala does not introduce `null`s. If we say that expressing a parameter that may have no values as varargs is not idiomatic in either Java or Scala, we are effectively saying that a varargs parameter must always have at least one value. However, that begs the question: why does the syntax of both languages allow specifying 0 vararg values?

rozza: > However, that begs the question: why does the syntax of both languages allow specifying 0 vararg values?

In Scala `args: T*` is syntactic sugar for `Seq[T]` - which I suppose is the same as Java's varargs being syntactic sugar `String.... args` and `String[] args`. 

So I suppose as arrays can be empty it defaults to that :man_shrugging: ?

rozza: In the scenario where its 0 or 1 - I think we'd need to clarify:

* That more than 1 is an error.
* That it is an optional parameter (just java doc).
* That by introducing yet another way of doing things we aren't adding confusion to the API but improving it.

stIncMale: Jeff expressed the opinion similar to your original impression - unusual/don't like. I will undo the changes related to using `T...` for representing optional paramaters.

stIncMale: Rolled back the changes related to expressing optional parameters as varargs. If the team changes its mind regarding this approach before the release of 4.7.0, we can do that part in a separate PR. For now, without that change, I hope that the PR is no longer controversial and has a potential to be merged.

Done.

driver-core/src/main/com/mongodb/client/model/search/SearchConstructibleBsonElement.java
```diff
@@ -68,10 +72,16 @@ public SearchConstructibleBsonElement defaultBucket(final String name) {
     }
 
     @Override
-    public SearchConstructibleBsonElement fuzzy(final SearchFuzzy option) {
+    public SearchConstructibleBsonElement fuzzy(@Nullable final FuzzySearchOptions... options) {
```
rozza, line 75: Should this have been rolled back - or should it be `fuzzy(FuzzySearchOptions, FuzzySearchOptions...)` ?

rozza (commented):
I think I like it,  `(T t, T... ts)` is a good way to enforce 1 or more values are required. 

It does offer a different way of doing things to the API where previously we'd have:

```
    public static Bson orderBy(final Bson... sorts) {
    public static Bson orderBy(final List<? extends Bson> sorts) {
```

This is preferring writing APIs as: 

```
    public static Bson orderBy(final Bson sort, Bson... sorts) {
    public static Bson orderBy(final Iterable<? extends Bson> sorts) {
```

I'm not massively convinced that one approach has much benefits over the other.  So my only concern is this approach adds yet another way of doing things in my API and if using it would confuse users. I don't think it would as the use of the API would be the same from a users perspective. 

So the question is it beneficial to us developing it and should the 5.0 version of the driver adopt this approach over the other?

driver-core/src/main/com/mongodb/client/model/search/SearchConstructibleBsonElement.java
stIncMale: Sorry, I missed that one. Rolled back and searched through the rest of the code to make sure there are no other changes related to expressing optional parameters as varargs.

stIncMale:
> my only concern is this approach adds yet another way of doing things in my API and if using it would confuse users. I don't think it would as the use of the API would be the same from a users perspective.

Yes, the using code looks exactly as if varargs were used without dedicating a parameter for the first value.

> So the question is it beneficial to us developing it and should the 5.0 version of the driver adopt this approach over the other?

It does: `method(T, T...)` supports the same usages as `method(T)`, and more. And the development price for that is negligible in my opinion (certainly smaller than the price of an overload).

If we decide that we should not use varargs at all in our API (which I would be totally fine with), then we should not have `method(T, T...)`. So far there has been no discussions about forbidding them.

PR #969, opened 2022-06-22
Fix `tlschannel.util.Util.getJavaMajorVersion`

stIncMale:
This is a manual cherry-pick and squash of the following commits:
- https://github.com/marianobarrios/tls-channel/commit/3b5f8a0bb592dd918b6641d771fc83d9a5c3d1ca
- https://github.com/marianobarrios/tls-channel/commit/f4b780e10c42c8a49e39f92d43fcb0179131124d

We don't test against JDK 18, and definitely not against early access builds. A manual test confirms that `18-ea` mentioned is the ticket is parsed successfully by the updated code.

The ticket does not specify the version the client uses. I am planning to backport the fix only to 4.6.x.

JAVA-4656

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/tlschannel/util/Util.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/tlschannel/util/Util.java
@@ -52,7 +52,7 @@ public class Util {
   }
 
   @VisibleForTesting(otherwise = PRIVATE)
-  public static int getJavaMajorVersion(final String javaVersion) {
+  static int getJavaMajorVersion(final String javaVersion) {
       String version = javaVersion;
       if (version.startsWith("1.")) {
           version = version.substring(2);
--- a/driver-core/src/main/com/mongodb/internal/connection/tlschannel/util/Util.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/tlschannel/util/Util.java
@@ -19,7 +19,11 @@
 
 package com.mongodb.internal.connection.tlschannel.util;
 
+import com.mongodb.internal.VisibleForTesting;
+
 import javax.net.ssl.SSLEngineResult;
+
+import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
 
 public class Util {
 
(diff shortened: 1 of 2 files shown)


rozza (approved):

driver-core/src/main/com/mongodb/internal/connection/tlschannel/util/Util.java
```diff
@@ -44,15 +48,24 @@ public static String resultToString(SSLEngineResult result) {
...
+      // Allow these formats:
+      // 1.8.0_72-ea
+      // 9-ea
+      // 9
+      // 9.0.1
+      // 17
+      int dotPos = version.indexOf('.');
+      int dashPos = version.indexOf('-');
+      return Integer.parseInt(
+              version.substring(0, dotPos > -1 ? dotPos : dashPos > -1 ? dashPos : version.length()));
```
stIncMale, line 69: The only way we deviate here from the algorithm in tls-channel is that instead of using `1` at the end of this [line](https://github.com/marianobarrios/tls-channel/blob/42acba44f23cf5c8969ced85c8331346ee8d85da/src/main/java/tlschannel/util/Util.java#L44), we use `version.length()`. Without this change `"17"` is parsed as 1, which then causes class loading issues with the root cause being swallowed and not being present in the logs.

I'll create a PR for tls-channel when this PR is approved.

driver-core/src/test/unit/com/mongodb/internal/connection/tlschannel/util/UtilTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+final class UtilTest {
+    @Test
+    void getJavaMajorVersion() {
+        assertAll(
+                () -> assertEquals(8, Util.getJavaMajorVersion("1.8.0_72-ea")),
+                () -> assertEquals(9, Util.getJavaMajorVersion("9-ea")),
+                () -> assertEquals(9, Util.getJavaMajorVersion("9")),
+                () -> assertEquals(9, Util.getJavaMajorVersion("9.0.1")),
+                () -> assertEquals(17, Util.getJavaMajorVersion("17")),
```
jyemin, line 31: Can we add a test for a double digit early access, e.g. `19-ea`?

jyemin (changes requested):

driver-core/src/test/unit/com/mongodb/internal/connection/tlschannel/util/UtilTest.java
stIncMale: When Java 8 dies off, everyone will be able to use `Runtime.version().feature()` instead of writing his own `Util.getJavaMajorVersion`.

Done.

jyemin (approved):

PR #970, opened 2022-06-24
Updated EncryptOptions queryType to be a String

rozza:
Allows for support for future new options and is inline with algorithm
also not being an enum.

JAVA-4643
JAVA-4644

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/vault/EncryptOptions.java
+++ b/driver-core/src/main/com/mongodb/client/model/vault/EncryptOptions.java
@@ -30,20 +30,7 @@ public class EncryptOptions {
     private String keyAltName;
     private final String algorithm;
     private Long contentionFactor;
-    private QueryType queryType;
-
-    /**
-     * The QueryType to use for "Indexed" queries
-     *
-     * @since 4.7
-     */
-    @Beta(Beta.Reason.SERVER)
-    public enum QueryType {
-        /**
-         * Equality query type
-         */
-        EQUALITY
-    }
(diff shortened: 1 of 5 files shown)


driver-core/src/main/com/mongodb/client/model/vault/EncryptOptions.java
```diff
@@ -153,28 +140,28 @@ public Long getContentionFactor() {
...
-     * <p>It is an error to set queryType when algorithm is not "Indexed".</p>
+     * <p>It is an error to set queryType when the algorithm is not "Indexed".</p>
      *
      * @param queryType the query type
      * @return this
      * @since 4.7
      */
     @Beta(Beta.Reason.SERVER)
-    public EncryptOptions queryType(@Nullable final QueryType queryType) {
+    public EncryptOptions queryType(@Nullable final String queryType) {
```
jyemin, line 150: Consider adding Javadoc that notes the currently-supported query type.  So that users don't have to look far to find it.  (We do this in getAlgorithm method, though maybe easier to find if it were on the setters as well)

jyemin (changes requested):

stIncMale (commented):
I agree with Jeff's suggestion ([currently, we support only "equality" `queryType`](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#querytype)), otherwise LGTM.

rozza:
Updated the javadoc and manually merged.

PR #971, opened 2022-06-27
Replace RuntimeException in catch with Exception to support Kotlin

katcharov:
Kotlin does not support checked exceptions, so our handlers will use Exception. All handlers are replaced, even those obviously unaffected, for the sake of a consistent noticeable pattern.

JAVA-4575

### DIFF
--- a/driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
+++ b/driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
@@ -293,7 +293,7 @@ public class DefaultConnectionPoolTest {
             connections.forEach(connection -> {
                 try {
                     connection.close();
-                } catch (RuntimeException e) {
+                } catch (Exception e) {
                     // ignore
                 }
             });
--- a/bson/src/test/unit/org/bson/json/JsonReaderTest.java
+++ b/bson/src/test/unit/org/bson/json/JsonReaderTest.java
@@ -1308,7 +1308,7 @@ public class JsonReaderTest {
                                      final Class<? extends RuntimeException> exClass) {
         try {
             testFunc.apply(new JsonReader(json));
-        } catch (final RuntimeException e) {
+        } catch (final Exception e) {
             if (exClass == null) {
                 throw e;
             }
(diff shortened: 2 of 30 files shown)


rozza (approved):
LGTM!

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
```diff
@@ -293,7 +293,7 @@ void infiniteMaxSize() {
             connections.forEach(connection -> {
                 try {
                     connection.close();
-                } catch (RuntimeException e) {
+                } catch (Exception e) {
                     // ignore
```
stIncMale, line 297: This does not seem like something worth doing in a test: if it so happens that our `connection.close()` throws sneakily inside a test, where no user code can be run, then we have a problem. This change hides thee problem. However, that's not a big deal, provided that we won't start throwing sneaky checked exceptions from our production code.

The reason we blindly apply the change to the production code is that it is difficult to detect whether a specific `catch` block has a chance to stumble upon a sneakily thrown user exception, or not. Furthermore, even if we manage to do this correctly now, future code changes may affect the conclusions we make be looking at the current code, thus opening possibilities to miss sneaky checked exceptions where we decided no such exception may currently happen.

The situation with tests is different: we know that there is no way a sneaky user exception may happen in them. The only reason to change them I can think of is something like "we want to get rid of any code catching `RuntimeException`s to reduce the temptation to ever do this in the production code". If this is the case, we need to agree that in the future, we forbid writing `catch(RuntimeException)` not only in the production code, but also in the test code, and we will request to change such blocks in tests if they appear in PRs (I am fine with it). , @katcharov, @rozza is this what we are agreeing on?

stIncMale (changes requested):

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
katcharov: I agree. Rules like this should be unconditional whenever possible. This reduces the number of unnecessary decisions we must make in the future.

I do not believe that this (or any) test should defend against the introduction of sneaky throws in our production code. I am fine with broadly catching all exceptions here, since we do not seem to be defending against changes to the list of possible exceptions `close` throws.

stIncMale (approved):

PR #974, opened 2022-06-28
Add comment option tests for distinct helper

katcharov:
JAVA-4627

Based on https://github.com/mongodb/mongo-php-library/pull/940/files


stIncMale (changes requested):
Based on the test failures, it looks like we need to modify our unified test runner to support the `comment` option.

stIncMale (approved):

stIncMale:
Static checkers are still unhappy and require changes.

PR #977, opened 2022-07-05
Add showExpandedEvents to change stream pipeline stage

katcharov:
JAVA-4585

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
@@ -54,6 +54,7 @@ import com.mongodb.client.model.IndexOptions;
 import com.mongodb.client.model.InsertManyOptions;
 import com.mongodb.client.model.InsertOneModel;
 import com.mongodb.client.model.InsertOneOptions;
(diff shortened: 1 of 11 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -986,9 +988,9 @@ public OperationResult executeCreateCollection(final BsonDocument operation) {
         ClientSession session = getSession(arguments);
 
         // In Java driver there is a separate method for creating a view, but in the unified test CRUD format
-        // views and collections are both created with the createCollection operation. We branch off the "viewOn" key's
+        // views and collections are both created with the createCollection operation. We branch off of key
```
rozza, line 991: Nit: The sentence is hard to parse

rozza (changes requested):
Looks good.

The Scala API needs updating to add the `showExpandedEvents` to [ChangeStreamObservable](https://github.com/mongodb/mongo-java-driver/blob/master/driver-scala/src/main/scala/org/mongodb/scala/ChangeStreamObservable.scala). It should be simple it just wraps the reactive streams API.  Scaladoc doesn't support custom tags - so just use `@note` for the server version.

rozza (approved):

PR #978, opened 2022-07-11
Add missing showExpandedEvents to .scala

katcharov:
See also 463b753
JAVA-4585


driver-scala/src/it/scala/org/mongodb/scala/syncadapter/SyncChangeStreamIterable.scala
```diff
@@ -94,4 +94,19 @@ case class SyncChangeStreamIterable[T](wrapped: ChangeStreamObservable[T])
     wrapped.comment(comment)
     this
   }
+
```
jyemin, line 97: Please remove the comment, to be consistent with the other methods in this class. (This is a test class, not production).

jyemin (changes requested):

PR #982, opened 2022-07-22
Do not retry a read operation when in a transaction

jyemin:
Fixes a regression introduced in 4.4.0 in scope of JAVA-4034, and
was undetected due to the lack of a specification test for the
required behavior.

JAVA-4684


driver-core/src/main/com/mongodb/internal/operation/OperationHelper.java
```diff
@@ -437,21 +437,12 @@ static boolean canRetryWrite(final ServerDescription serverDescription, final Co
         return true;
     }
 
-    static boolean isRetryableRead(final boolean retryReads, final ServerDescription serverDescription,
```
jyemin, line 440: In JAVA-4034, use of this method was removed with the addition of the `RetryState` class, but that caused us to lost the check for `sessionContext.hasActiveTransaction()`.  Since the `RetryState` class already has the check for whether `retryReads` is enabled, I chose to just remove this unused method and move that check into `canRetryRead` below.

rozza (approved):

PR #984, opened 2022-07-25
Replace synchronized with ReentrantLock

jyemin:
In order to get the benefit for structured concurrency, the driver
must use Lock#lockInterruptibly rather than simply Lock#lock.

JAVA-4642

What use of `synchrnonized` is left out of this PR?

* Used in the bson module: I didn't include this because `MongoInterruptedException` is in driver-core, not bson, so it wasn't clear what exception should be thrown.  So deferring this for now
* Use in driver-legacy. Generally we don't make improvements in legacy
* Use in threads that the driver starts itself.  This can be done later if we ever move these to virtual threads.  But also, these threads are not accessible to applications so can't easily be interrupted.
* Use in asynchronous contexts.  Structured concurrency wouldn't be used with reactive streams, so it's not needed
* Use in tests.  Obviously these are ignorable

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/Locks.java
+++ b/driver-core/src/main/com/mongodb/internal/Locks.java
@@ -25,18 +25,18 @@ import java.util.function.Supplier;
  * This class is not part of the public API and may be removed or changed at any time.
  */
 public final class Locks {
-    public static void runWithLock(final Lock lock, final Runnable action) {
-        supplyWithLock(lock, () -> {
+    public static void withLock(final Lock lock, final Runnable action) {
+        withLock(lock, () -> {
             action.run();
             return null;
         });
     }
 
-    public static <V> V supplyWithLock(final Lock lock, final Supplier<V> supplier) {
-        return checkedSupplyWithLock(lock, supplier::get);
+    public static <V> V withLock(final Lock lock, final Supplier<V> supplier) {
+        return checkedWithLock(lock, supplier::get);
     }
 
-    public static <V, E extends Exception> V checkedSupplyWithLock(final Lock lock, final CheckedSupplier<V, E> supplier) throws E {
+    public static <V, E extends Exception> V checkedWithLock(final Lock lock, final CheckedSupplier<V, E> supplier) throws E {
         try {
             lock.lockInterruptibly();
             try {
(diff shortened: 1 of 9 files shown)


driver-core/src/main/com/mongodb/internal/Locks.java
```diff
@@ -0,0 +1,72 @@
...
+
+import com.mongodb.MongoInterruptedException;
+
+import java.util.concurrent.locks.Lock;
+import java.util.function.Supplier;
+
+/**
+ * This class is not part of the public API and may be removed or changed at any time.
+ */
+public final class Locks {
```
rozza, line 27: Can be simplified down to:

```
public final class Locks {
    public static void withLock(final Lock lock, final Runnable action) {
       checkedWithLock(lock, () -> {
           action.run();
           return null;
       });
    }

    public static <V> V withLock(final Lock lock, final Supplier<V> supplier) {
        return checkedWithLock(lock, supplier::get);
    }

    public static <V, E extends Exception> V checkedWithLock(final Lock lock, final CheckedSupplier<V, E> supplier) throws E {
        try {
            lock.lockInterruptibly();
            try {
                return supplier.get();
            } finally {
                lock.unlock();
            }
            throw new MongoInterruptedException("Interrupted waiting for lock", e);
        }
    }

    private Locks() {
    }
}
```

note: I simplified the method names as well.

rozza (changes requested):
Looks good - one DRY and naming improvement suggestion.

driver-core/src/main/com/mongodb/internal/Locks.java
jyemin: Much nicer.  I kind of liked the old names, but I see the appeal of these, so I changed them as suggested.

I'm not sure how the compiler is inferring that the `supplier::get` method reference is of type `CheckedSupplier<V, RuntimeException>` but that's what's happening so I won't argue.

rozza (approved):
LGTM!

jyemin:
Updated GridFS as well

katcharov (approved):
LGTM

driver-sync/src/main/com/mongodb/client/gridfs/GridFSDownloadStreamImpl.java
```diff
@@ -156,12 +159,12 @@ public void mark() {
     }
 
     @Override
-    public synchronized void mark(final int readlimit) {
+    public void mark(final int readlimit) {
```
jyemin, line 162: This synchronization of mark and reset seems unnecessary, since typical usage of these methods is in a single thread, and also misleading, since currentPosition is used in unsynchronized methods as well.

rozza: Ack

PR #985, opened 2022-07-25
Improve error message for invalid field names

jyemin:
For replacement documents, it's now:
  "Field names in a replacement document can not start with '$' but '%s' does"
For update documents, it's now:
  "All update operators must start with '$', but '%s' does not"

JAVA-2114

### DIFF
--- a/bson/src/main/org/bson/FieldNameValidator.java
+++ b/bson/src/main/org/bson/FieldNameValidator.java
@@ -17,6 +17,7 @@
 package org.bson;
 
 import static java.lang.String.format;
+import static org.bson.assertions.Assertions.isTrue;
 
 /**
  * A field name validator, for use by BSON writers to validate field names as documents are encoded.
@@ -40,9 +41,7 @@ public interface FieldNameValidator {
      * @throws IllegalArgumentException if fieldName is actually valid
      */
     default String getValidationErrorMessage(final String fieldName) {
-        if (validate(fieldName)) {
-            throw new IllegalArgumentException(format("%s is valid", fieldName));
-        }
+        isTrue(fieldName + " is valid", !validate(fieldName));
         return format("Invalid BSON field name %s", fieldName);
     }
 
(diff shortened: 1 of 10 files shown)


rozza (approved):
Looks good - only nit is I don't think you test the `getValidationErrorMessage`if validate is true logic.

bson/src/main/org/bson/FieldNameValidator.java
```diff
@@ -30,6 +32,20 @@ public interface FieldNameValidator {
...
+    /**
+     * Return the validation error message for an invalid field
+     *
+     * @param fieldName the field name
+     * @return the validation error message
+     * @throws IllegalArgumentException if fieldName is actually valid
+     */
+    default String getValidationErrorMessage(final String fieldName) {
+        if (validate(fieldName)) {
+            throw new IllegalArgumentException(format("%s is valid", fieldName));
```
rozza, line 44: Is this tested?

jyemin: Ah, I see what you mean.  No, I didn't test it because `AbstractBsonWriter` guarantees that this will never happen.  Because of that it is more appropriate to use the Assertions class here instead to reflect the intent (and we typically don't test assertions).

rozza (approved):
LGTM

driver-core/src/main/com/mongodb/internal/validator/ReplacingDocumentFieldNameValidator.java
```diff
@@ -34,13 +37,15 @@ public class ReplacingDocumentFieldNameValidator implements FieldNameValidator {
 
     @Override
     public boolean validate(final String fieldName) {
-        if (fieldName == null) {
```
jyemin, line 37: Removed this check since none of the other validators have it, and an assertion didn't seem worth it since the next line will throw anyway if fieldName is null.

rozza (approved):

PR #988, opened 2022-08-16
Remove support for pre-3.6 server releases

jyemin:
JAVA-4295

I divided this up into a commit for each release that we're dropping.  So if we ever need to roll back, it's easy to do it for a single release. 

I suggest enabling Hide Whitespace Only Changes in the UI when you review, as many of the lines changed only whitespace based on the removal of conditional statements.


katcharov:
@jyemin Just curious, what was the process for removing support? Were you able to find all locations by following version checks from some single location? (For example, I see that you have removed methods like `validateReadConcern`, but I did not immediately see why.)

jyemin:
Yes, just following version checks and then removing anything unused as a result of removing the version check and any code branches that were no longer needed.

katcharov (approved):
lgtm. Nice to see the ~7k lines removed.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -177,43 +169,23 @@ protected EncodingMetadata encodeMessageBodyWithMetadata(final BsonOutput bsonOu
...
-
-    private void addDocumentWithPayload(final BsonOutput bsonOutput, final int messageStartPosition) {
-        BsonBinaryWriter bsonBinaryWriter = new BsonBinaryWriter(bsonOutput, getPayloadArrayFieldNameValidator());
-        BsonWriter bsonWriter = new SplittablePayloadBsonWriter(bsonBinaryWriter, bsonOutput, messageStartPosition, getSettings(), payload);
-        BsonDocument commandToEncode = getCommandToEncode();
-        getCodec(commandToEncode).encode(bsonWriter, commandToEncode, EncoderContext.builder().build());
-    }
-
     private int getOpMsgFlagBits() {
         return getOpMsgResponseExpectedFlagBit();
```
rozza, line 190: Nit - getOpMsgResponseExpectedFlagBit() is only used here - no need for two methods.

rozza (approved):
LGTM! Amazing how much code can be removed! :partying_face:

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: Inlined

PR #989, opened 2022-08-19
Replace existing SDAM integration spec tests with unified tests

jyemin:
JAVA-4658


driver-core/src/test/unit/com/mongodb/internal/connection/TestServerListener.java
```diff
@@ -23,31 +23,47 @@
...
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Predicate;
 
-import static com.mongodb.assertions.Assertions.isTrue;
 import static com.mongodb.assertions.Assertions.notNull;
 
-class TestServerListener implements ServerListener {
+public class TestServerListener implements ServerListener {
```
jyemin, line 35: The changes in this file match roughly what's in TestConnectionPoolListener, which lets us wait for events efficiently.

driver-core/src/test/unit/com/mongodb/internal/connection/TestServerListener.java
```diff
@@ -23,31 +23,47 @@
...
     private ServerClosedEvent serverClosedEvent;
     private final List<ServerDescriptionChangedEvent> serverDescriptionChangedEvents = new ArrayList<ServerDescriptionChangedEvent>();
+    private final Lock lock = new ReentrantLock();
+    private final Condition condition = lock.newCondition();
+    private volatile int waitingForEventCount;
+    private Predicate<ServerDescriptionChangedEvent> waitingForEventMatcher;
 
     @Override
     public void serverOpening(final ServerOpeningEvent event) {
-        isTrue("serverOpeningEvent is null", serverOpeningEvent == null);
```
jyemin, line 37: These assertions make it impossible to use this class in a context where there is more than one server (e.g. a replica set).  It seemed fine to remove them.

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
```diff
@@ -73,6 +73,74 @@ public static ContextElement ofConnectionPoolEvent(final BsonDocument expected,
         return new ConnectionPoolEventMatchingContextElement(expected, actual, eventPosition);
     }
 
+    public static ContextElement ofWaitForPrimaryChange() {
```
jyemin, line 76: New operations, new contexts.  Anonymous classes seem fine here, though I don't remember why that wasn't done for many of the existing ones.

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -97,6 +102,8 @@ public final class Entities {
                     "id", "uriOptions", "serverApi", "useMultipleMongoses", "storeEventsAsEntities",
                     "observeEvents", "observeSensitiveCommands", "ignoreCommandMonitoringEvents"));
     private final Set<String> entityNames = new HashSet<>();
+    private final Map<String, ExecutorService> threads = new HashMap<>();
```
jyemin, line 105: This supports the new runOnThread operation

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -84,15 +95,35 @@ public abstract class UnifiedTest {
...
     private final UnifiedCrudHelper crudHelper;
     private final UnifiedGridFSHelper gridFSHelper = new UnifiedGridFSHelper(entities);
     private final UnifiedClientEncryptionHelper clientEncryptionHelper = new UnifiedClientEncryptionHelper(entities);
-    private final ValueMatcher valueMatcher = new ValueMatcher(entities, context);
-    private final ErrorMatcher errorMatcher = new ErrorMatcher(context);
-    private final EventMatcher eventMatcher = new EventMatcher(valueMatcher, context);
     private final List<FailPoint> failPoints = new ArrayList<>();
+    private final UnifiedTestContext rootContext = new UnifiedTestContext();
+
+    private class UnifiedTestContext {
```
jyemin, line 105: This refactoring was required in order to support runOnThread.  The problem is that the context is not really capable of supporting concurrent operations, so what we do now is create a new instance of UnifiedTestContext in runOnThread, and  then pass UnifiedTestContext around everywhere instead of relying on the root context.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -186,74 +219,94 @@ public void shouldPassAllOutcomes() {
...
     }
 
-    private OperationResult executeOperation(final BsonDocument operation, final int operationNum) {
-        context.push(ContextElement.ofStartedOperation(operation, operationNum));
+    private OperationResult executeOperation(final UnifiedTestContext context, final BsonDocument operation, final int operationNum) {
+        context.getAssertionContext().push(ContextElement.ofStartedOperation(operation, operationNum));
         String name = operation.getString("name").getValue();
         try {
             switch (name) {
+                case "createEntities":
```
jyemin, line 292: Lots of new operations!

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -449,6 +502,186 @@ protected boolean terminateLoop() {
...
+        ClusterDescription currentClusterDescription = client.getClusterDescription();
+        long timeoutNanos =
+                TimeUnit.NANOSECONDS.convert(arguments.getNumber("timeoutMS", new BsonInt32(10000)).longValue(), TimeUnit.MILLISECONDS);
+        long startTime = System.nanoTime();
+        while (primaryIsSame(priorClusterDescription, currentClusterDescription) || noPrimary(currentClusterDescription)) {
+            if (System.nanoTime() - startTime > timeoutNanos) {
+                fail(context.getAssertionContext().getMessage("Timed out waiting for primary change"));
+            }
+            try {
+                //noinspection BusyWait
```
jyemin, line 585: We could avoid busy waiting with a more complicated implementation of ServerListener, but it didn't seem worth the effort, so just looping instead with a busy wait.

driver-core/src/test/unit/com/mongodb/internal/connection/TestServerListener.java
rozza: Ack

driver-core/src/test/unit/com/mongodb/internal/connection/TestServerListener.java
rozza: Ack

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
rozza: Ack - sounds good.

rozza (approved):
LGTM

Comment: Given that the unified test code is very abstract, I found that without code familiarity its hard to grok whats going on - especially, when having to implement a new version of Unified Tests. It maybe worth thinking if there is anything that can be done to minimize the steep learning curve/cost in the future. Especially if the code becomes even more complex.

PR #992, opened 2022-08-23
Update netty dependencies to use the netty-bom

rozza:
JAVA-4710


rozza:
Patch: https://spruce.mongodb.com/version/6304c3b6a4cf470c7d90a30d

jyemin:
Same failure with tcnative.  The BOM is pulling in 2.0.53, so that version doesn't work either.  So we have to roll back somewhere between 48 and 52 unless we can figure out why.

PR #993, opened 2022-08-23
Fix double connection release on error for drop and create collection…

jyemin:
… operations

JAVA-4711


driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -552,6 +552,7 @@ class ProcessCommandsCallback implements SingleResultCallback<Void> {
         public void onResult(final Void result, final Throwable t) {
             if (t != null) {
                 finalCallback.onResult(null, t);
+                return;
```
jyemin, line 555: It's also the case that when encryption is enabled, we will continue processing the subsequent commands instead of stopping early.  I don't think that is the intention, but want to confirm that.

rozza (approved):

PR #997, opened 2022-08-30
Fix typo in abstract method name (ServerMonitorListener)

heli-os:
open the PR of typo correction too. However, changes have occurred in the abstraction code and the implementation code.

so i open a new PR

--
I understood it to only open issues for features or bugs.
so,,, should i open an issue? if so, please tell me 🙇


jyemin:
While I'd love to merge this, as I hate spelling errors, the problem is that the error is in the spelling of a public method, and fixing it would break binary compatibility with existing applications that are depending on the current, misspelled name.

So I'm going to have to close this without merging.  But thanks for finding it, and feel free to open PRs for misspellings outside of the public interface of the driver.

heli-os:
@jyemin I know what you mean. thank you 👍

PR #1002, opened 2022-09-23
Support List<T> and Map<String, T> components in Java records

jyemin:
This enhances the support for any record component that is a generic List or Map where the generic type of the List or Map value is a record or POJO.  It used to incorrectly decode to org.bson.Document. Now it respects the generic type of the List or Map value when decoding.

JAVA-4667

### DIFF
--- a/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
+++ b/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
@@ -22,6 +22,7 @@ import org.bson.BsonWriter;
 import org.bson.codecs.Codec;
 import org.bson.codecs.DecoderContext;
 import org.bson.codecs.EncoderContext;
+import org.bson.codecs.Parameterizable;
 import org.bson.codecs.RepresentationConfigurable;
 import org.bson.codecs.configuration.CodecConfigurationException;
 import org.bson.codecs.configuration.CodecRegistry;
@@ -35,6 +36,7 @@ import javax.annotation.Nullable;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.RecordComponent;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -83,6 +85,10 @@ final class RecordCodec<T extends Record> implements Codec<T> {
         @SuppressWarnings("deprecation")
         private static Codec<?> computeCodec(final RecordComponent component, final CodecRegistry codecRegistry) {
             var codec = codecRegistry.get(toWrapper(component.getType()));
+            if (codec instanceof Parameterizable parameterizableCodec
+                    && component.getGenericType() instanceof ParameterizedType parameterizedType) {
+                codec = parameterizableCodec.parameterize(codecRegistry, Arrays.asList(parameterizedType.getActualTypeArguments()));
+            }
             BsonType bsonRepresentationType = null;
 
             if (component.isAnnotationPresent(BsonRepresentation.class)) {
--- a/bson/src/main/org/bson/codecs/AbstractIterableCodec.java
+++ b/bson/src/main/org/bson/codecs/AbstractIterableCodec.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bson.codecs;
+
+import org.bson.BsonReader;
+import org.bson.BsonType;
+import org.bson.BsonWriter;
+
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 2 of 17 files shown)


jyemin:
Particularly looking for feedback on a new `Parameterizable` interface for Codecs as an improved solution to the problem of parameterizing a Codec with generic types. 

If it looks good I'll proceed with added more tests.

bson/src/main/org/bson/codecs/AbstractIterableCodec.java
```diff
@@ -0,0 +1,69 @@
...
+package org.bson.codecs;
+
+import org.bson.BsonReader;
+import org.bson.BsonType;
+import org.bson.BsonWriter;
+
+import java.util.ArrayList;
+import java.util.List;
+
+abstract class AbstractIterableCodec<T> implements Codec<Iterable<T>> {
```
jyemin, line 26: Extracted superclass so that the main logic can be shared between IterableCodec and the new ParameterizedIterableCodec.

bson/src/main/org/bson/codecs/AbstractMapCodec.java
```diff
@@ -0,0 +1,72 @@
...
+package org.bson.codecs;
+
+import org.bson.BsonReader;
+import org.bson.BsonType;
+import org.bson.BsonWriter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public abstract class AbstractMapCodec<T> implements Codec<Map<String, T>> {
```
jyemin, line 26: Same here.

bson/src/main/org/bson/codecs/ContainerCodecHelper.java
```diff
@@ -62,6 +66,23 @@ static Object readValue(final BsonReader reader, final DecoderContext decoderCon
         }
     }
 
+    static Codec<?> getCodec(final CodecRegistry codecRegistry, final Type type) {
```
jyemin, line 69: This is the guts of the implementation: a recursive algorithm for creating a parameterized codec of any depth.

bson/src/main/org/bson/codecs/IterableCodec.java
```diff
@@ -17,25 +17,26 @@
...
+import static org.bson.codecs.ContainerCodecHelper.getCodec;
 
 /**
  * Encodes and decodes {@code Iterable} objects.
  *
  * @since 3.3
  */
 @SuppressWarnings("rawtypes")
-public class IterableCodec implements Codec<Iterable>, OverridableUuidRepresentationCodec<Iterable> {
+public class IterableCodec extends AbstractIterableCodec<Object>
```
jyemin, line 38: `IterableCodec` now implements `Codec<Iterable<Object>>` instead of `Codec<Iterable>`.  This is binary but not source compatibile, which I think is ok given how this class is likely (not) used directly in applications.

bson/src/main/org/bson/codecs/IterableCodec.java
```diff
@@ -79,48 +79,31 @@ public Object transform(final Object objectToTransform) {
 
 
     @Override
-    public Codec<Iterable> withUuidRepresentation(final UuidRepresentation uuidRepresentation) {
```
jyemin, line 82: The diffs are a bit ugly due to the mix of functional changes with refactoring.  I didn't realize I would need to extract a superclass until after the functional work was done.  If it's too hard to read, I can go back and rebase everything, doing the refactoring first in a separate PR.

bson/src/main/org/bson/codecs/ContainerCodecHelper.java
```diff
@@ -62,6 +66,23 @@ static Object readValue(final BsonReader reader, final DecoderContext decoderCon
...
+        if (type instanceof Class) {
+            return codecRegistry.get((Class<?>) type);
+        } else if (type instanceof ParameterizedType) {
+            ParameterizedType parameterizedType = (ParameterizedType) type;
+            Codec<?> rawCodec = codecRegistry.get((Class<?>) parameterizedType.getRawType());
```
jyemin, line 74: `ParameterizedType#getRawType` returns `Type`, not `Class`,  but can the raw type of a parameterized type be anything else except a class? If so, this is an unsafe cast.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -83,6 +85,10 @@ Object getValue(final Record record) throws InvocationTargetException, IllegalAc
         @SuppressWarnings("deprecation")
         private static Codec<?> computeCodec(final RecordComponent component, final CodecRegistry codecRegistry) {
             var codec = codecRegistry.get(toWrapper(component.getType()));
+            if (codec instanceof Parameterizable parameterizableCodec
```
rozza, line 88: What is this magic? assigning variables in an if statement :exploding_head:

bson/src/main/org/bson/codecs/ContainerCodecHelper.java
rozza: Seems familiar to the Pojo codec parameterizedType handling but cleaner.

rozza (commented):
Can you add a `ParameterizedRecordCodec`as well then top level records can contain nested generic records eg:

```
public record TestRecordWithListGenericConcrete(@BsonId ObjectId id, TestRecordWithListGeneric<TestRecordEmbedded> nestedRecords) {
}

public record TestRecordWithListGeneric<T>(List<T> nestedRecords) {
}
```

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
jyemin: Not just one but two!

jyemin:
> Can you add a ParameterizedRecordCodecas well then top level records can contain nested generic records 

Yes, I don't see why not.  The implementation will be a bit more complicated but I think it fits in with the overall design.  I'll open a follow-up ticket for that if you don't mind.

rozza (approved):
LGTM!

jyemin:
Opened JAVA-4740 to track request for support of nested parameterized records.

katcharov (approved):
lgtm!

PR #1003, opened 2022-09-23
use implementation, not compile in Gradle docs

kevinAlbs:
compile was removed in Gradle 7
see: https://docs.gradle.org/current/userguide/upgrading_version_6.html#sec:configuration_removal

### DIFF
--- a/4.7/bson/installation-guide/index.html
+++ b/4.7/bson/installation-guide/index.html
@@ -358,7 +358,7 @@ BSON is short for Binary <a href="http://json.org/">JSON</a>, is a binary-encode
 <section class="gradle hidden">
 <pre><code>
   dependencies {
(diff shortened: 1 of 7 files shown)


ccho-mongodb:
Reassigning review to @jyemin

jyemin:
This is gh-pages.  Are we still even publishing reference docs there, and if so, why?

kevinAlbs:
There is a link to [Reference Documentation](https://mongodb.github.io/mongo-java-driver/4.6/) link on the [MongoDB Java Reactive Streams](https://www.mongodb.com/docs/drivers/reactive-streams/). Is that link out of date?

rozza:
@ccho-mongodb the change is good - but I'm no longer sure of the source location of the docs and the changes will also have to be made there.

rozza (approved):

PR #1004, opened 2022-09-28
Fix check for whether crypt shared lib is available 

jyemin:
JAVA-4739

* Extract method for duplicated logic from sync and reactive CommandMarker
* Fix the bug and and unit tests for the extracted method

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
+++ b/driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
@@ -138,6 +138,13 @@ public final class MongoCryptHelper {
         return new BsonDocumentWrapper<>(new Document(optionsMap), new DocumentCodec());
     }
 
+    public static boolean isMongocryptdSpawningDisabled(@Nullable final String cryptSharedLibVersion,
+            final AutoEncryptionSettings settings) {
+        boolean cryptSharedLibIsAvailable = cryptSharedLibVersion != null && cryptSharedLibVersion.isEmpty();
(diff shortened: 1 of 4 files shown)


driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
```diff
@@ -138,6 +138,13 @@ private static BsonDocument toBsonDocument(final Map<String, Object> optionsMap)
         return new BsonDocumentWrapper<>(new Document(optionsMap), new DocumentCodec());
     }
 
+    public static boolean isMongocryptdSpawningDisabled(@Nullable final String cryptSharedLibVersion,
+            final AutoEncryptionSettings settings) {
+        boolean cryptSharedLibIsAvailable = cryptSharedLibVersion != null && !cryptSharedLibVersion.isEmpty();
+        boolean cryptSharedLibRequired = (boolean) settings.getExtraOptions().getOrDefault("cryptSharedLibRequired", false);
+        return settings.isBypassAutoEncryption() || settings.isBypassQueryAnalysis() || cryptSharedLibRequired || cryptSharedLibIsAvailable;
```
rozza, line 145: I like that this is all in one place but wow what a return statement!!

rozza (approved):
LGTM! Good work on DRY'ing up the code also and adding a regression test. :+1:

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
jyemin: Yeah, I wish it were simpler...

PR #1006, opened 2022-09-29
Add missing condition for whether to sleep in server monitor

jyemin:
There is a missing check in an already-complicated conditinal statement for whether to sleep in the server monitor before executing the next check.

The condition is included in the server discovery and monitoring specification pseudo-code but missing in the driver, and there are currently no tests for it.

This commits adds the missing condition to the conditional statement.  Effectively, this will reduce the number of sockets that the monitor opens on a server that is in the process of shutting down but still accepting connections.

JAVA-4730

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
@@ -170,7 +170,7 @@ class DefaultServerMonitor implements ServerMonitor {
                     sdamProvider.get().update(currentServerDescription);
 
                     if (((connection == null || shouldStreamResponses(currentServerDescription))
-                            && currentServerDescription.getTopologyVersion() != null)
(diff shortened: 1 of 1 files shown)


patrickfreed (approved):
Code change LGTM. I'm not sure how to verify it though (is there an easy way to specify a git branch as a maven dependency?)

Also, it could be good to add a test for this behavior. It could do something like enable a failpoint on hello that returns ShutdownInProgress and verifies that the number of heartbeatFailed events is only 1 per heartbeatFrequencyMS. It may make sense to file a DRIVERS ticket for such a test.

jyemin:
I verified it by manually running your reproduction against this branch.  DRIVERS ticket is a great idea.  Would you mind opening?  I think it's ok to commit this before having a regression test.  I don't think it can be worse.

patrickfreed:
Sounds good, filed DRIVERS-2458.

PR #1007, opened 2022-09-30
Support recursive types in parameterized records

jyemin:
Provide codec support for records whose type declarations contain cycles.

JAVA-4745

Looking for feedback on the general approach.  The codec registry changes could use some more unit tests if we want to proceed with this.

### DIFF
--- a/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
+++ b/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
@@ -270,7 +270,7 @@ final class RecordCodec<T extends Record> implements Codec<T>, Parameterizable {
     }
 
     RecordCodec(final Class<T> clazz, final CodecRegistry codecRegistry, final List<Type> types) {
-        if (types.size() != clazz.getTypeParameters().length) {
+        if (types.size() != clazz.getTypeParameters().length || types.isEmpty()) {
             throw new CodecConfigurationException("Unexpected number of type parameters for record class " + clazz);
         }
         this.clazz = notNull("class", clazz);
--- a/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
+++ b/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
@@ -89,12 +89,11 @@ final class RecordCodec<T extends Record> implements Codec<T>, Parameterizable {
         @SuppressWarnings("deprecation")
         private static Codec<?> computeCodec(final List<Type> typeParameters, final RecordComponent component,
                 final CodecRegistry codecRegistry) {
-            var codec = codecRegistry.get(toWrapper(resolveComponentType(typeParameters, component)));
-            if (codec instanceof Parameterizable parameterizableCodec
-                    && component.getGenericType() instanceof ParameterizedType parameterizedType) {
-                codec = parameterizableCodec.parameterize(codecRegistry,
-                        resolveActualTypeArguments(typeParameters, component.getDeclaringRecord(), parameterizedType));
-            }
+            var rawType = toWrapper(resolveComponentType(typeParameters, component));
+            var codec = component.getGenericType() instanceof ParameterizedType parameterizedType
+                    ? codecRegistry.get(rawType,
+                    resolveActualTypeArguments(typeParameters, component.getDeclaringRecord(), parameterizedType))
+                    : codecRegistry.get(rawType);
             BsonType bsonRepresentationType = null;
 
             if (component.isAnnotationPresent(BsonRepresentation.class)) {
--- a/bson-record-codec/src/test/unit/org/bson/codecs/record/RecordCodecTest.java
+++ b/bson-record-codec/src/test/unit/org/bson/codecs/record/RecordCodecTest.java
@@ -347,17 +347,15 @@ public class RecordCodecTest {
     public void testSelfReferentialRecords() {
         var registry = fromProviders(new RecordCodecProvider(), Bson.DEFAULT_CODEC_REGISTRY);
         var codec = registry.get(TestSelfReferentialHolderRecord.class);
-        var identifier = new ObjectId();
         var testRecord = new TestSelfReferentialHolderRecord("0",
                 new TestSelfReferentialRecord<>("1",
                 new TestSelfReferentialRecord<>("2", null, null),
                 new TestSelfReferentialRecord<>("3", null, null)));
 
         var document = new BsonDocument();
-        var writer = new BsonDocumentWriter(document);
 
         // when
-        codec.encode(writer, testRecord, EncoderContext.builder().build());
+        codec.encode(new BsonDocumentWriter(document), testRecord, EncoderContext.builder().build());
 
         // then
         assertEquals(
--- a/bson-record-codec/src/test/unit/org/bson/codecs/record/RecordCodecTest.java
+++ b/bson-record-codec/src/test/unit/org/bson/codecs/record/RecordCodecTest.java
@@ -50,6 +50,8 @@ import org.bson.codecs.record.samples.TestRecordWithNestedParameterized;
 import org.bson.codecs.record.samples.TestRecordWithNestedParameterizedRecord;
 import org.bson.codecs.record.samples.TestRecordWithParameterizedRecord;
 import org.bson.codecs.record.samples.TestRecordWithPojoAnnotations;
+import org.bson.codecs.record.samples.TestSelfReferentialHolderRecord;
+import org.bson.codecs.record.samples.TestSelfReferentialRecord;
 import org.bson.conversions.Bson;
 import org.bson.types.ObjectId;
 import org.junit.jupiter.api.Test;
@@ -333,6 +335,38 @@ public class RecordCodecTest {
                 .append("name", new BsonString("Felix"))
                 .append("hobbies", new BsonArray(List.of(new BsonString("rugby"), new BsonString("badminton"))))
                 .append("a", new BsonInt32(13));
+
+        // when
+        var decoded = codec.decode(new BsonDocumentReader(document), DecoderContext.builder().build());
+
+        // then
+        assertEquals(testRecord, decoded);
+    }
+
+    @Test
+    public void testSelfReferentialRecords() {
+        var registry = fromProviders(new RecordCodecProvider(), Bson.DEFAULT_CODEC_REGISTRY);
+        var codec = registry.get(TestSelfReferentialHolderRecord.class);
+        var identifier = new ObjectId();
+        var testRecord = new TestSelfReferentialHolderRecord("0",
+                new TestSelfReferentialRecord<>("1",
+                new TestSelfReferentialRecord<>("2", null, null),
+                new TestSelfReferentialRecord<>("3", null, null)));
+
+        var document = new BsonDocument();
+        var writer = new BsonDocumentWriter(document);
+
+        // when
+        codec.encode(writer, testRecord, EncoderContext.builder().build());
+
+        // then
+        assertEquals(
+                new BsonDocument("_id", new BsonString("0"))
+                        .append("selfReferentialRecord",
+                        new BsonDocument("name", new BsonString("1"))
+                        .append("left", new BsonDocument("name", new BsonString("2")))
+                        .append("right", new BsonDocument("name", new BsonString("3")))),
+                document);
 
         // when
         var decoded = codec.decode(new BsonDocumentReader(document), DecoderContext.builder().build());
--- a/bson/src/main/org/bson/codecs/ContainerCodecHelper.java
+++ b/bson/src/main/org/bson/codecs/ContainerCodecHelper.java
@@ -71,12 +71,7 @@ final class ContainerCodecHelper {
             return codecRegistry.get((Class<?>) type);
         } else if (type instanceof ParameterizedType) {
             ParameterizedType parameterizedType = (ParameterizedType) type;
-            Codec<?> rawCodec = codecRegistry.get((Class<?>) parameterizedType.getRawType());
-            if (rawCodec instanceof Parameterizable) {
-                return ((Parameterizable) rawCodec).parameterize(codecRegistry, Arrays.asList(parameterizedType.getActualTypeArguments()));
-            } else {
-                return rawCodec;
-            }
+            return codecRegistry.get((Class<?>) parameterizedType.getRawType(), Arrays.asList(parameterizedType.getActualTypeArguments()));
         } else {
             throw new CodecConfigurationException("Unsupported generic type of container: " + type);
         }
--- a/bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
+++ b/bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
@@ -16,6 +16,7 @@
 
 package org.bson.codecs.configuration;
 
+import org.bson.assertions.Assertions;
 import org.bson.codecs.Codec;
 
 import java.lang.reflect.Type;
@@ -52,22 +53,21 @@ public interface CodecRegistry extends CodecProvider {
      * Gets a Codec for the given parameterized class, after resolving any type variables with the given type arguments.
      *
      * <p>
-     * The default behavior is to throw a {@link CodecConfigurationException}.  Applications are encouraged to either override this
-     * method in their own implementations of this interface, or else use the factory methods in {@link CodecRegistries}.
+     * The default behavior is to throw a {@link AssertionError}, as it's expected that {@code CodecRegistry} implementations are always
+     * provided by this library and will override the method appropriately.
      * </p>
      *
-     * @param clazz the parameterized class
-     * @param typeArguments the type arguments to apply to the parameterized class
+     * @param clazz         the parameterized class
+     * @param typeArguments the type arguments to apply to the parameterized class.  This list may be empty but not null.
+     * @param <T>           the class type
      * @return a codec for the given class, with the given type parameters resolved
-     * @throws CodecConfigurationException by default, if the implementation does not override this method, or if no codec can be found
-     * for the given class and type arguments.
-     * @param <T> the class type
+     * @throws CodecConfigurationException if no codec can be found for the given class and type arguments.
+     * @throws AssertionError              by default, if the implementation does not override this method, or if no codec can be found
+     *                                     for the given class and type arguments.
+     * @see org.bson.codecs.Parameterizable
      * @since 4.8
      */
     default <T> Codec<T> get(Class<T> clazz, List<Type> typeArguments) {
-        throw new CodecConfigurationException("The application is configuring a codec for a parameterized class, but is using an "
-                + "implementation of org.bson.codecs.configuration.CodecRegistry that does not override the default implementation of the"
-                + "method get(java.lang.Class<T>, java.util.List<java.lang.reflect.Type>).  This implementation should be replaced by one "
-                + "that is obtained from one of the CodecRegistry builder methods in org.bson.codecs.configuration.CodecRegistries");
+        throw Assertions.fail("This method should have been overridden but was not.");
     }
 }
--- a/bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
+++ b/bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
@@ -65,6 +65,9 @@ public interface CodecRegistry extends CodecProvider {
      * @since 4.8
      */
     default <T> Codec<T> get(Class<T> clazz, List<Type> typeArguments) {
-        throw new CodecConfigurationException("Make this message really informative");  // TODO
+        throw new CodecConfigurationException("The application is configuring a codec for a parameterized class, but is using an "
+                + "implementation of org.bson.codecs.configuration.CodecRegistry that does not override the default implementation of the"
+                + "method get(java.lang.Class<T>, java.util.List<java.lang.reflect.Type>).  This implementation should be replaced by one "
+                + "that is obtained from one of the CodecRegistry builder methods in org.bson.codecs.configuration.CodecRegistries");
     }
 }
--- a/bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
+++ b/bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
@@ -18,6 +18,9 @@ package org.bson.codecs.configuration;
 
 import org.bson.codecs.Codec;
 
+import java.lang.reflect.Type;
+import java.util.List;
+
 /**
  * A registry of Codec instances searchable by the class that the Codec can encode and decode.
  *
@@ -28,7 +31,11 @@ import org.bson.codecs.Codec;
  *
  * <p>As of the 4.0 release, this class extends the {@code CodecProvider} interface.  This capability was introduced to enable nesting
  * registries inside another registry.</p>
+ *
+ * <p>Applications are encouraged to NOT implement this interface, but rather use the factory methods in {@link CodecRegistries}.</p>
+ *
  * @since 3.0
+ * @see CodecRegistries
  */
 public interface CodecRegistry extends CodecProvider {
     /**
@@ -40,4 +47,24 @@ public interface CodecRegistry extends CodecProvider {
      * @throws CodecConfigurationException if the registry does not contain a codec for the given class.
      */
     <T> Codec<T> get(Class<T> clazz);
+
+    /**
+     * Gets a Codec for the given parameterized class, after resolving any type variables with the given type arguments.
+     *
+     * <p>
+     * The default behavior is to throw a {@link CodecConfigurationException}.  Applications are encouraged to either override this
+     * method in their own implementations of this interface, or else use the factory methods in {@link CodecRegistries}.
+     * </p>
+     *
+     * @param clazz the parameterized class
+     * @param typeArguments the type arguments to apply to the parameterized class
+     * @return a codec for the given class, with the given type parameters resolved
+     * @throws CodecConfigurationException by default, if the implementation does not override this method, or if no codec can be found
+     * for the given class and type arguments.
+     * @param <T> the class type
+     * @since 4.8
+     */
+    default <T> Codec<T> get(Class<T> clazz, List<Type> typeArguments) {
+        throw new CodecConfigurationException("Make this message really informative");  // TODO
+    }
 }
--- a/bson/src/main/org/bson/internal/ChildCodecRegistry.java
+++ b/bson/src/main/org/bson/internal/ChildCodecRegistry.java
@@ -21,8 +21,12 @@ import org.bson.codecs.Codec;
 import org.bson.codecs.configuration.CodecRegistry;
 
 import java.lang.reflect.Type;
-import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
+
+import static java.lang.String.format;
+import static org.bson.assertions.Assertions.isTrueArgument;
+import static org.bson.assertions.Assertions.notNull;
 
 // An implementation of CodecRegistry that is used to detect cyclic dependencies between Codecs
 class ChildCodecRegistry<T> implements CodecRegistry {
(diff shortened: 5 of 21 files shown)


bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -89,12 +89,11 @@ Object getValue(final Record record) throws InvocationTargetException, IllegalAc
         @SuppressWarnings("deprecation")
         private static Codec<?> computeCodec(final List<Type> typeParameters, final RecordComponent component,
                 final CodecRegistry codecRegistry) {
-            var codec = codecRegistry.get(toWrapper(resolveComponentType(typeParameters, component)));
-            if (codec instanceof Parameterizable parameterizableCodec
-                    && component.getGenericType() instanceof ParameterizedType parameterizedType) {
-                codec = parameterizableCodec.parameterize(codecRegistry,
-                        resolveActualTypeArguments(typeParameters, component.getDeclaringRecord(), parameterizedType));
-            }
+            var rawType = toWrapper(resolveComponentType(typeParameters, component));
```
jyemin, line 92: Responsibility for parameterization moves into the registry

bson-record-codec/src/test/unit/org/bson/codecs/record/RecordCodecTest.java
```diff
@@ -341,6 +343,38 @@ public void testRecordWithExtraData() {
         assertEquals(testRecord, decoded);
     }
 
+    @Test
+    public void testSelfReferentialRecords() {
+        var registry = fromProviders(new RecordCodecProvider(), Bson.DEFAULT_CODEC_REGISTRY);
+        var codec = registry.get(TestSelfReferentialHolderRecord.class);
+        var identifier = new ObjectId();
+        var testRecord = new TestSelfReferentialHolderRecord("0",
```
jyemin, line 351: And voila!

bson/src/main/org/bson/codecs/ContainerCodecHelper.java
```diff
@@ -71,12 +71,7 @@ static Codec<?> getCodec(final CodecRegistry codecRegistry, final Type type) {
             return codecRegistry.get((Class<?>) type);
         } else if (type instanceof ParameterizedType) {
             ParameterizedType parameterizedType = (ParameterizedType) type;
-            Codec<?> rawCodec = codecRegistry.get((Class<?>) parameterizedType.getRawType());
```
jyemin, line 74: Responsibility for parameterization moves into the registry

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
```diff
@@ -40,4 +47,24 @@ public interface CodecRegistry extends CodecProvider {
...
+     *
+     * @param clazz the parameterized class
+     * @param typeArguments the type arguments to apply to the parameterized class
+     * @return a codec for the given class, with the given type parameters resolved
+     * @throws CodecConfigurationException by default, if the implementation does not override this method, or if no codec can be found
+     * for the given class and type arguments.
+     * @param <T> the class type
+     * @since 4.8
+     */
+    default <T> Codec<T> get(Class<T> clazz, List<Type> typeArguments) {
```
jyemin, line 67: All the driver-provided implementations override the default implementation.   The only problem is if an application has implemented this interface.  This should be rare, but I've seen a few support cases where people missed the memo about `CodecRegistries`. Since currently it's only records that use this functionality and records are already horribly broken when they have any parameterized types, I think this is ok.  But we'll have to make a decision about whether it's safe to start relying on this for PojoCodec, where we could break existing applications on a driver upgrade.

bson/src/main/org/bson/internal/CodecCache.java
```diff
@@ -19,36 +19,75 @@
...
 import static java.lang.String.format;
 
 final class CodecCache {
-    private final ConcurrentMap<Class<?>, Optional<Codec<?>>> codecCache = new ConcurrentHashMap<>();
 
-    public boolean containsKey(final Class<?> clazz) {
-        return codecCache.containsKey(clazz);
+    static final class CodecCacheKey {
+        private final Class<?> clazz;
+        private final List<Type> types;
```
jyemin, line 35: Now the (possibly null) type arguments are part of the cache key, so we can cache the codec for, e.g., `List<Integer>` as well as `List`.

bson/src/main/org/bson/internal/LazyCodec.java
```diff
@@ -23,14 +23,19 @@
 import org.bson.codecs.EncoderContext;
 import org.bson.codecs.configuration.CodecRegistry;
 
+import java.lang.reflect.Type;
+import java.util.List;
+
 class LazyCodec<T> implements Codec<T> {
     private final CodecRegistry registry;
     private final Class<T> clazz;
+    private final List<Type> types;
```
jyemin, line 32: By making LazyCodec aware of type arguments, it can properly get the parameterized codec when it need to.

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
```diff
@@ -50,18 +58,22 @@ public <T> Codec<T> get(final Class<T> clazz, final CodecRegistry registry) {
...
     public <T> Codec<T> get(final ChildCodecRegistry<T> context) {
-        if (!codecCache.containsKey(context.getCodecClass())) {
+        CodecCacheKey codecCacheKey = new CodecCacheKey(context.getCodecClass(), context.getTypes());
+        if (!codecCache.containsKey(codecCacheKey)) {
             for (CodecProvider provider : codecProviders) {
                 Codec<T> codec = provider.get(context.getCodecClass(), context);
                 if (codec != null) {
-                    return codecCache.putIfMissing(context.getCodecClass(), codec);
+                    if (codec instanceof Parameterizable && context.getTypes() != null) {
+                        codec = (Codec<T>) ((Parameterizable) codec).parameterize(context, context.getTypes());
```
jyemin, line 69: Some dicey casting here.  I think it's ok but I wonder if we can do better.

rozza (commented):
Interesting, would an alternative be to keep this within the RecordCodec implementaiton - so no changes to the CodecRegistry?

For example would just adding a `LazyRecordCodec` be enough to break any cycles eg:

```
class LazyRecordCodec<T> implements Codec<T> {
    private final Supplier<Codec<T>> supplier;
    private volatile Codec<T> wrapped;

    LazyRecordCodec(final Supplier<Codec<T>> supplier) {
        this.supplier = supplier;
    }

    @Override
    public void encode(final BsonWriter writer, final T value, final EncoderContext encoderContext) {
        getWrapped().encode(writer, value, encoderContext);
    }

    @Override
    public Class<T> getEncoderClass() {
        return  getWrapped().getEncoderClass();
    }

    @Override
    public T decode(final BsonReader reader, final DecoderContext decoderContext) {
        return getWrapped().decode(reader, decoderContext);
    }

    private Codec<T> getWrapped() {
        if (wrapped == null) {
            wrapped = supplier.get();
        }
        return wrapped;
    }
}
```

Then the codec could be:

```
this.codec = new LazyRecordCodec<>(() -> computeCodec(typeParameters, component, codecRegistry));
```

The downside is it makes all records codecs lazy but the upside would be no changes required outside the RecordCodec.

bson/src/main/org/bson/internal/CodecCache.java
```diff
@@ -19,36 +19,75 @@
...
     @SuppressWarnings("unchecked")
-    public synchronized <T> Codec<T> putIfMissing(final Class<T> clazz, final Codec<T> codec) {
-        Optional<Codec<?>> cachedCodec = codecCache.computeIfAbsent(clazz, clz -> Optional.of(codec));
+    public synchronized <T> Codec<T> putIfMissing(final CodecCacheKey codecCacheKey, final Codec<T> codec) {
+        Optional<Codec<?>> cachedCodec = codecCache.computeIfAbsent(codecCacheKey, clz -> Optional.of(codec));
         if (cachedCodec.isPresent()) {
             return (Codec<T>) cachedCodec.get();
         }
-        codecCache.put(clazz, Optional.of(codec));
+        codecCache.put(codecCacheKey, Optional.of(codec));
```
stIncMale, line 84: Something appears to be wrong with how this method was implemented before the PR, and how it is implemented now. It likely can be implemented as follows (note that `synchronized` is no longer there):

```java
public <T> Codec<T> putIfMissing(final CodecCacheKey codecCacheKey, final Codec<T> codec) {
    return codecCache.compute(codecCacheKey, (k, v) -> {
        if (v != null && v.isPresent()) {
            return v;
        } else {
            return Optional.of(codec);
        }}).map(c -> (Codec<T>)c).orElseThrow(() -> new AssertionError("unreachable"));
}
```

Tangentially related to the above improvement, not to this PR, methods `OverridableUuidRepresentationCodecRegistry.get`, `ProvidersCodecRegistry.get` do two hash map lookups instead of one on the happy path, because they both use the anti-pattern "check if a binding is there and then find the binding again, this time return the value".

stIncMale (changes requested):

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -89,12 +89,11 @@ Object getValue(final Record record) throws InvocationTargetException, IllegalAc
...
+            var codec = component.getGenericType() instanceof ParameterizedType parameterizedType
```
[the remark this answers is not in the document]
jyemin: Yeah, I love that I can do this in the bson-record-codec module, at the least

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
jyemin: > we should annotate the interface with @Evolving 

Agreed,  though we would need to copy the annotation into either `bson-record-codec` or `bson` to do so.

> and remove the default implementation

You don't buy my argument that it's not really breaking until we start using it in PojoCodec?  I thought that was fairly compelling.

bson/src/main/org/bson/internal/CodecCache.java
stIncMale: Wrong as in both readability and performance could be better. I was initially baffled by seeing a put in a `ConcurrentMap` that is guarded with `synchronized` and is implemented as two puts for the same key.

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
jyemin: > Or we could move @Evolving 
I'd prefer not to add another Maven artifact for such a small thing.  Maybe we should just copy it.

> If this can't happen, could you please help me see why?

Yes, it could happen.  But since it's less likely to actually be the case, the runtime exception still seems ok.

bson/src/main/org/bson/internal/CodecCache.java
jyemin: In that case, let's fix this in a separate PR.

bson/src/main/org/bson/internal/CodecCache.java
```diff
@@ -19,36 +19,75 @@
...
 final class CodecCache {
-    private final ConcurrentMap<Class<?>, Optional<Codec<?>>> codecCache = new ConcurrentHashMap<>();
 
-    public boolean containsKey(final Class<?> clazz) {
-        return codecCache.containsKey(clazz);
+    static final class CodecCacheKey {
+        private final Class<?> clazz;
+        private final List<Type> types;
+
+        CodecCacheKey(final Class<?> clazz, final List<Type> types) {
```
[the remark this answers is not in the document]
stIncMale: This seems like one more reason to introduce a new artifact.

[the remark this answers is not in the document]
jyemin: Not really, as we can't split a package between two modules like that.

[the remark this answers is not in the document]
stIncMale: I did not notice at first: both `TypeWithTypeParameters` and `TypeData` are in `bson`, i.e., are already available to `bson-record-codec` and to `CodecRegistry`, as it is also in `bson`.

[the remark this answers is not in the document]
jyemin: Oh, of course...  Still, it would be weird if CodecRegistry depended on an interface in `org.bson.codecs.pojo`.  I also don't know if that's exactly the interface we want anyway.

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
rozza: Its a breaking change and its not 100% necessary: it can be solved without changes to the codec registry at this time. So should it be added in 4.x rather than 5.x? I'm not sure.

Its only called in 3 places: `ContainerCodecHelper`, `LazyCodec` and `RecordCodec` and implemented in 3 CodecRegistry implementations.

If we are worried about breaking changes then we could always add a new interface and check they type of the codec registry and process from there.

jyemin: > we could always add a new interface and check they type of the codec registry

I think it would amount to the same thing as what's in this PR: a runtime exception when the application actually tries to use the new functionality (currently only in the context of record codec)

rozza: It has the benefit of not breaking any users code and if not available can defer to the none typed version.

This implementation is really internal detail for the RecordCodec and possibly in the future be shared with the PojoCodec (which already has its own implementation that didnt require any breaking changes).

jyemin: > and if not available can defer to the none typed version

Ah, that was the part I was missing!  Yes, we could fallback to the old behavior.  But I'm not sure if that's a better user experience.  How would they find out that they are doing something wrong and correct it?

jyemin: I gave it a shot by adding a new interface--ParameterizationAwareCodecRegistry--that defines the new method, and everything works fine if I change all the CodecRegistry implementations in the driver to implement this interface and add `instancof` checks where required.  

The problem comes if I then replace one of the built-in registries with one that doesn't implement the interface.  Most of the RecordCodecTest cases start to fail, in two different ways:

1. `CodecConfigurationException: Can not decode to a record with type parameters that has not been parameterized`
2. Incorrect decoding, where the type system is violated (e.g. `Document` instances in a list of type `List<FooRecord>` 

The first type of failure is ok, as it manifests as an exception, similar to what would happen with this PR as is.  But the second seems risky, as an application might have more trouble recognizing that decoding is broken, or diagnosing why it's broken when they do notice it.

rozza: Interested in how 2 comes about - can you provide an example of how you configured it to fail?

jyemin: I demonstrated it in a separate draft PR: #1011   PTAL there

jyemin: Added another commit that implements the idea that I described [here](https://github.com/mongodb/mongo-java-driver/pull/1011#discussion_r989273403).  PTAL

jyemin: And I just realized that the change is unnecessary.  Since all application-provided registries are _already_ wrapped in an instance of `OverridableUuidRepresentationCodecRegistry`, we're already protected, since the application-provided registry is always treated as a CodecProvider.

jyemin: It would be clearer, for example, if the application were to provide in the settings not a `CodecRegistry` but rather a `List<CodecProvider>`.

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
rozza: Is it an error to have a parameterizable codec with an empty list of types? - Should that scenario be handled at the codec level? Or at least document that parameterize wont be called if the types list is empty.

rozza (commented):
One question - but its looking good

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
jyemin: I'm considering opening a ticket to deprecate `MongoClientSettings#codecRegistry` and adding `MongoClientSettings#codecProvider`, and add a new `CodecProviders` builder class similar to `CodecRegistries`.  Alternatively, could just document that `MongoClientSettings#codecRegistry` is treated only as a `CodecProvider`.  The former seems better but more disruptive, the latter seems less disruptive but more likely to just be ignored. Thoughts?

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
rozza: So here we should we just pass context.getTypes() and leave it to the instance to handle. Or we should document that `Parameterizable#parameterize` won't be called with an empty list.

rozza (changes requested):

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
jyemin: In the first commit this was a check for null rather than isEmpty, but was changed as a result of the suggestion to make `ChildCodecRegistry#types` non-nullable.  @stIncMale given @rozza 's comment, should we rethink that suggestion?

rozza: I think not null was the right choice - its just that a comment is needed in the javadoc for Parameterizable otherwise 
future implementers could get caught out.

jyemin: ok, so javadoc on Parameterizable, plus a `!types.isEmpty()` check for all implementations of the new CodecRegistry method (and Javadoc change there too)?

rozza: sgtm

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
stIncMale: > And I just realized that the change is unnecessary. Since all application-provided registries are already wrapped in an instance of OverridableUuidRepresentationCodecRegistry, we're already protected, since the application-provided registry is always treated as a CodecProvider. 

> I'm considering opening a ticket to deprecate MongoClientSettings#codecRegistry and adding MongoClientSettings#codecProvider, and add a new CodecProviders builder class similar to CodecRegistries. Alternatively, could just document that MongoClientSettings#codecRegistry is treated only as a CodecProvider. The former seems better but more disruptive, the latter seems less disruptive but more likely to just be ignored. Thoughts?

Is this actually true, though? It seems that `com.mongodb.client.internal.MongoClientImpl.createChangeStreamIterable` does not wrap `settings.getCodecRegistry()` into `OverridableUuidRepresentationCodecRegistry`, and `ChangeStreamIterableImpl` uses this user-provided `CodecRegistry` as `CodecRegistry` by calling `CodecRegistry.get(Class<T> clazz)`.

jyemin: I will have to double-check, but this seems like a bug.

stIncMale: I thought I finally understood the `CodecProvider`/`CodecRegistry`, and started writing why we don't need `<T> Codec<T> get(Class<T> clazz, CodecRegistry registry)`, and only need `<T> Codec<T> get(Class<T> clazz)`. But while doing so, another revelation came and now I believe I see why `CodecProvider` is needed and is indeed what we should ask from user, while asking users to give us a `CodecRegistry` makes no sense. In other words, I am in favour of the first, more disruptive, solution that you proposed.

bson/src/main/org/bson/internal/CodecCache.java
stIncMale: I forgot to link the new ticket here: https://jira.mongodb.org/browse/JAVA-4751. Will do when this PR is merged to avoid conflicts.

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
stIncMale: The `!types.isEmpty()` check is weird: it is usually expected that if a collection is accepted, then an empty collection may be passed. I think it should not be introduced. Also, `RecordCodec` is currently dealing with empty `types` just fine, and itself uses `List.of()` in its `RecordCodec(final Class<T> clazz, final CodecRegistry codecRegistry)` constructor.

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
jyemin: Opened JAVA-4773 to track the bug in MongoClient#watch

jyemin: Opened JAVA-4774 to track the proposed deprecation in MongoClientSettings

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
```diff
@@ -65,6 +65,9 @@ public interface CodecRegistry extends CodecProvider {
      * @since 4.8
      */
     default <T> Codec<T> get(Class<T> clazz, List<Type> typeArguments) {
-        throw new CodecConfigurationException("Make this message really informative");  // TODO
+        throw new CodecConfigurationException("The application is configuring a codec for a parameterized class, but is using an "
+                + "implementation of org.bson.codecs.configuration.CodecRegistry that does not override the default implementation of the"
+                + "method get(java.lang.Class<T>, java.util.List<java.lang.reflect.Type>).  This implementation should be replaced by one "
+                + "that is obtained from one of the CodecRegistry builder methods in org.bson.codecs.configuration.CodecRegistries");
```
stIncMale, line 71: Given the [discussion above](https://github.com/mongodb/mongo-java-driver/pull/1007#discussion_r992378186), we don't need this method to have a default implementation, right?

stIncMale (changes requested):

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
jyemin: I think it's still a good idea, just in case there are applications that are (unnecessarily) creating their own CodecRegistry implementations only to have them treated as CodecProvider.  This will avoid unnecessary compilation errors for that scenario.

stIncMale: Are there situations when a user may see the error message from the default implementation? Unless I am confused, that can't happen. If that's the case, we can just use `Assertions.fail()` as the default implementation.

jyemin: Yeah, I agree. This message is wrong given what we know now.

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
jyemin: I guess I can update the Iterable and Map Codec to do the same.

jyemin: If I understand you correctly @stIncMale , we will need to add 

```
        if (types.isEmpty()) {
            return this;
        }
```

at the start of every implementation of `Parameterizable#parameterize`. 

But we should only be calling `Parameterizable#parameterize` as a result of a call to `CodecRegistry#get(java.lang.Class<T>, java.util.List<java.lang.reflect.Type>)`, but without the isEmpty check we end up calling it in as a result of a call to `CodecRegistry#get(java.lang.Class<T>)` as well.

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
jyemin: Done

bson/src/main/org/bson/codecs/configuration/CodecRegistry.java
jyemin: Let me know if it's ok to resolve this thread

stIncMale: I think this thread can be resolved.

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
stIncMale: We discussed with @jyemin and agreed that:
- `CodecRegistry.get(Class<T> clazz, List<Type> typeArguments)` should check `clazz.getTypeParameters().length == typeArguments.size()` and on top of that check `!typeArguments.isEmpty()`.
- `Parameterizable.parameterize` should check `!typeArguments.isEmpty()`.
- `ChildCodecRegistry.types`/`LazyCodec.types` should become `@Nullable List<Type>` and must not be empty. As a result, `ChildCodecRegistry.getTypes` should return `Optional<List<Type>>`.

jyemin: > CodecRegistry.get(Class<T> clazz, List<Type> typeArguments) should check clazz.getTypeParameters().length == typeArguments.size() and on top of that check !typeArguments.isEmpty().

Also check for null. Done

> Parameterizable.parameterize should check !typeArguments.isEmpty().

I just documented it.  All the implementations already seem to check the size so isEmpty check isn't necessary

> ChildCodecRegistry.types/LazyCodec.types should become @Nullable List<Type> and must not be empty. As a result, ChildCodecRegistry.getTypes should return Optional<List<Type>>.

Done

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
```diff
@@ -39,11 +41,16 @@ public ProvidersCodecRegistry(final List<? extends CodecProvider> codecProviders
...
     @Override
     public <T> Codec<T> get(final Class<T> clazz) {
-        return get(new ChildCodecRegistry<T>(this, clazz));
+        return get(new ChildCodecRegistry<T>(this, clazz, null));
     }
 
     @Override
     public <T> Codec<T> get(final Class<T> clazz, final List<Type> typeArguments) {
+        notNull("typeArguments", typeArguments);
+        isTrueArgument("typeArguments is not empty", typeArguments.size() >= 1);
```
stIncMale, line 50: `typeArguments.isEmpty()` is both more explicit than `typeArguments.size() >= 1` and for some implementations may be more performant (a silly example: if `LinkedList` were not to cache its size, then `LinkedList.size` would have had linear time complexity). While the latter is irrelevant here, readability is a good enough argument.

stIncMale (changes requested):

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
stIncMale: > All the implementations already seem to check the size so isEmpty check isn't necessary

`RecordCodec.parametrize` just passes `types` to the constructor of `RecordCodec`, which does the `types.size() != clazz.getTypeParameters().length` check. However, this check does not prevent `types` from being empty.

A different question: do you think annotating fields and constructor parameters with `@Nullable` (relevant for `ChildCodecRegistry`, `LazyCodec`, `CodecCacheKey`) is not useful?

jyemin: It's just that our `Nullable` annotation is in driver-core, so we don't really use `Nullable` in bson module at this point

stIncMale (approved):

bson/src/main/org/bson/internal/ChildCodecRegistry.java
```diff
@@ -20,36 +20,64 @@
...
+import static org.bson.assertions.Assertions.isTrueArgument;
+import static org.bson.assertions.Assertions.notNull;
+
 // An implementation of CodecRegistry that is used to detect cyclic dependencies between Codecs
 class ChildCodecRegistry<T> implements CodecRegistry {
 
     private final ChildCodecRegistry<?> parent;
     private final CycleDetectingCodecRegistry registry;
     private final Class<T> codecClass;
+    private final List<Type> types;
```
rozza, line 37: Suggestion: Annotate as nullable - so its clear in the code.

Also in LazyCodec and the cache key

rozza: Apologies - see this was merged

jyemin: No worries.  I mentioned in another thread that we don't have Nullable annotations available in the bson module.  I think we already have a ticket for that.

PR #1009, opened 2022-10-04
Update InitialDnsSeedlistDiscoveryTest

jyemin:
JAVA-4532

This is harder to read than I thought it would be, probably because I tried to do a few things at the same time:

1. Refactor the test class so that it's a single test instead of multiple tests.  This matches the spec test description more closely
2. Pull out a couple of methods that do the bulk of the assertions
3. Add `parsed_options` assertions
4. Support `ping` field in tests

If it's too hard to read I can split it up.  Let me know.

### DIFF
--- a/driver-core/src/test/resources/initial-dns-seedlist-discovery/README.rst
+++ b/driver-core/src/test/resources/initial-dns-seedlist-discovery/README.rst
@@ -14,8 +14,9 @@ replica set name ``repl0``.
 
 The tests in the ``load-balanced`` directory MUST be executed against a
 load-balanced sharded cluster with the mongos servers running on localhost ports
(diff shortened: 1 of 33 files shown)


stIncMale (commented):
Why do we have two effectively separate implementations of the test runner (`assertErrorCondition`, `assertNonErrorCondition`)? Intuitively, I'd expect that there is a single implementation that ignores some failures of some operations when `isError`, and fails otherwise.

jyemin:
It just works out that way from following the logic of the test runner specification, which says:

> If error is specified and true, drivers MUST verify that initializing the MongoClient throws an error. If error is not specified or is false, both initializing the MongoClient and running a ping operation must succeed without throwing any errors.

None of the assertions required for the non-error case are applicable for the error case, and vice versa.

stIncMale (approved):

PR #1010, opened 2022-10-05
Prioritize electionId over setVersion on 6.0+ servers

jyemin:
JAVA-4707

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
@@ -261,19 +261,16 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
             return false;
         }
 
-        if (nullSafeCompareTo(newDescription.getElectionId(), maxElectionId) > 0) {
-            maxElectionId = newDescription.getElectionId();
-        }
-        if (nullSafeCompareTo(newDescription.getSetVersion(), maxSetVersion) > 0) {
-            maxSetVersion = newDescription.getSetVersion();
-        }
+        maxElectionId = nullSafeMax(newDescription.getElectionId(), maxElectionId);
+        maxSetVersion = nullSafeMax(newDescription.getSetVersion(), maxSetVersion);
+
+        invalidateOldPrimaries(newDescription.getAddress());
 
         if (isNotAlreadyPrimary(newDescription.getAddress())) {
             LOGGER.info(format("Discovered replica set primary %s with max election id %s and max set version %d",
                     newDescription.getAddress(), newDescription.getElectionId(), newDescription.getSetVersion()));
         }
 
-        invalidateOldPrimaries(newDescription.getAddress());
         return true;
     }
 
@@ -313,6 +310,16 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
         return first.compareTo(second);
     }
 
+    private static <T extends Comparable<T>> T nullSafeMax(final T first, final T second) {
+        if (first == null) {
+            return second;
+        }
+        if (second == null) {
+            return first;
+        }
+        return first.compareTo(second) >= 0 ? first : second;
+    }
+
     private boolean isNotAlreadyPrimary(final ServerAddress address) {
         ServerTuple serverTuple = addressToServerTupleMap.get(address);
         return serverTuple == null || !serverTuple.description.isPrimary();
@@ -321,7 +328,7 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
     private boolean handleShardRouterChanged(final ServerDescription newDescription) {
         if (!newDescription.isShardRouter()) {
             LOGGER.error(format("Expecting a %s, but found a %s.  Removing %s from client view of cluster.",
-                                 SHARD_ROUTER, newDescription.getType(), newDescription.getAddress()));
+                    SHARD_ROUTER, newDescription.getType(), newDescription.getAddress()));
             removeServer(newDescription.getAddress());
         }
         return true;
--- a/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
@@ -257,10 +257,6 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
             Integer setVersion = newDescription.getSetVersion();
             if (newDescription.getMaxWireVersion() >= SIX_DOT_ZERO_WIRE_VERSION) {
                 if (isValidPrimary(electionId, setVersion)) {
-                    if (isNewPrimary(electionId, setVersion)) {
-                        LOGGER.info(format("Setting max election id to %s and max set version to %d from replica set primary %s",
-                                newDescription.getElectionId(), newDescription.getSetVersion(), newDescription.getAddress()));
-                    }
                     maxElectionId = electionId;
                     maxSetVersion = setVersion;
                 } else {
@@ -270,10 +266,6 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
             } else {
                 if (setVersion != null && electionId != null) {
                     if (isValidPrimaryPre60(electionId, setVersion)) {
-                        if (nullSafeCompareTo(electionId, maxElectionId) > 0) {
-                            LOGGER.info(format("Setting max election id to %s from replica set primary %s",
-                                    newDescription.getElectionId(), newDescription.getAddress()));
-                        }
                         maxElectionId = electionId;
                     } else {
                         invalidatePotentialPrimary(newDescription);
@@ -281,14 +273,13 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
                     }
                 }
                 if (nullSafeCompareTo(setVersion, maxSetVersion) > 0) {
-                    LOGGER.info(format("Setting max set version to %d from replica set primary %s",
-                            newDescription.getSetVersion(), newDescription.getAddress()));
                     maxSetVersion = setVersion;
                 }
             }
 
             if (isNotAlreadyPrimary(newDescription.getAddress())) {
-                LOGGER.info(format("Discovered replica set primary %s", newDescription.getAddress()));
+                LOGGER.info(format("Discovered replica set primary %s with max election id %s and max set version %d",
+                        newDescription.getAddress(), electionId, setVersion));
             }
             invalidateOldPrimaries(newDescription.getAddress());
         }
@@ -298,11 +289,6 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
     private boolean isValidPrimary(final ObjectId electionId, final Integer setVersion) {
         return nullSafeCompareTo(electionId, maxElectionId) > 0
                 || (nullSafeCompareTo(electionId, maxElectionId) == 0 && nullSafeCompareTo(setVersion, maxSetVersion) >= 0);
-    }
-
-    private boolean isNewPrimary(final ObjectId electionId, final Integer setVersion) {
-        return nullSafeCompareTo(electionId, maxElectionId) > 0
-                || (nullSafeCompareTo(electionId, maxElectionId) == 0 && nullSafeCompareTo(setVersion, maxSetVersion) > 0);
     }
 
     private boolean isValidPrimaryPre60(final ObjectId electionId, final Integer setVersion) {
--- a/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
@@ -278,8 +278,8 @@ public abstract class AbstractMultiServerCluster extends BaseCluster {
         ObjectId electionId = description.getElectionId();
         Integer setVersion = description.getSetVersion();
         if (description.getMaxWireVersion() >= SIX_DOT_ZERO_WIRE_VERSION) {
-            return nullSafeCompareTo(electionId, maxElectionId) <= 0
-                    && (nullSafeCompareTo(electionId, maxElectionId) != 0 || nullSafeCompareTo(setVersion, maxSetVersion) < 0);
+            return nullSafeCompareTo(electionId, maxElectionId) < 0
+                    || (nullSafeCompareTo(electionId, maxElectionId) == 0 && nullSafeCompareTo(setVersion, maxSetVersion) < 0);
         } else {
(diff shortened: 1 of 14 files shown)


driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -247,46 +246,43 @@ private boolean handleReplicaSetMemberChanged(final ServerDescription newDescrip
...
-                LOGGER.info(format("Canonical address %s does not match server address.  Removing %s from client view of cluster",
-                                   newDescription.getCanonicalAddress(), newDescription.getAddress()));
-            }
+            LOGGER.info(format("Canonical address %s does not match server address.  Removing %s from client view of cluster",
+                    newDescription.getCanonicalAddress(), newDescription.getAddress()));
             removeServer(newDescription.getAddress());
             return true;
         }
 
         if (newDescription.isPrimary()) {
```
jyemin, line 255: I re-wrote the existing logic for pre-6.0 servers to match the pseudocode in the specification.

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -247,46 +246,43 @@ private boolean handleReplicaSetMemberChanged(final ServerDescription newDescrip
...
                     return false;
                 }
-
-                if (!electionId.equals(maxElectionId)) {
-                    if (LOGGER.isInfoEnabled()) {
-                        LOGGER.info(format("Setting max election id to %s from replica set primary %s", electionId,
-                                newDescription.getAddress()));
+            } else {
+                if (setVersion != null && electionId != null) {
+                    if (nullSafeCompareTo(maxSetVersion, setVersion) <= 0
```
jyemin, line 271: I used Intellij refactoring to reverse the condition on this if statement from what it is in the specification.  I did it so that there is symmetry between this branch and the one for the 6.0+ case above.  I think it's justified for that reason.  

CC @nbbeeken consider doing the same in the spec?

jyemin, line 271: I removed the null checks for maxSetVersion and maxElectionId from the spec because `nullSafeCompareTo` makes them redundant

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -247,46 +246,43 @@ private boolean handleReplicaSetMemberChanged(final ServerDescription newDescrip
...
-                    maxElectionId = electionId;
                 }
-            }
-
-            if (setVersion != null
-                    && (maxSetVersion == null || setVersion.compareTo(maxSetVersion) > 0)) {
-                if (LOGGER.isInfoEnabled()) {
-                    LOGGER.info(format("Setting max set version to %d from replica set primary %s", setVersion,
-                            newDescription.getAddress()));
+                if (nullSafeCompareTo(setVersion, maxSetVersion) > 0) {
```
jyemin, line 281: Note that in the current version of the spec this conditional is nested in `if (setVersion != null && electionId != null)`.  This is an error in the spec that will be corrected soon.  This is the correct spot for it.

CC @nbbeeken

nbbeeken: Correct and we've merged the fix for this: https://github.com/mongodb/specifications/pull/1312

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -247,46 +246,43 @@ private boolean handleReplicaSetMemberChanged(final ServerDescription newDescrip
...
-                                maxSetVersion, maxElectionId));
-                    }
-                    addressToServerTupleMap.get(newDescription.getAddress()).server.resetToConnecting();
+            if (newDescription.getMaxWireVersion() >= SIX_DOT_ZERO_WIRE_VERSION) {
+                if (nullSafeCompareTo(electionId, maxElectionId) > 0
+                        || (nullSafeCompareTo(electionId, maxElectionId) == 0 && nullSafeCompareTo(setVersion, maxSetVersion) >= 0)) {
+                    LOGGER.info(format("Setting max election id to %s and max set version to %d from replica set primary %s",
+                            newDescription.getElectionId(), newDescription.getSetVersion(), newDescription.getAddress()));
+                    maxElectionId = newDescription.getElectionId();
+                    maxSetVersion = newDescription.getSetVersion();
```
nbbeeken, line 264: Should you instead be using the variables that were fetched from these getters on lines 256/257? Or vice versa, just noticed a mix of getting and variable usage in the pre-6 section

jyemin: Yeah, it was just left over this way from the original code.  Will fix.

jyemin: Done

nbbeeken (approved):

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
nbbeeken: We're making the code more readable https://github.com/mongodb/specifications/pull/1315 But I think keeping the buggy RSM logic (pre-6.0) untouched has value for referencing the existing behavior, implementers shouldn't have to touch their old code, but it would be interesting to find out if others run into the issue Java had vs python and node just working.

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -247,46 +246,43 @@ private boolean handleReplicaSetMemberChanged(final ServerDescription newDescrip
...
+                if (setVersion != null && electionId != null) {
+                    if (nullSafeCompareTo(maxSetVersion, setVersion) <= 0
+                            && (nullSafeCompareTo(maxSetVersion, setVersion) != 0 || nullSafeCompareTo(maxElectionId, electionId) <= 0)) {
+                        LOGGER.info(format("Setting max election id to %s from replica set primary %s",
+                                newDescription.getElectionId(), newDescription.getAddress()));
+                        maxElectionId = electionId;
+                    } else {
+                        invalidatePotentialPrimary(newDescription);
+                        return false;
                     }
```
stIncMale, line 279: It appears to me, that the following code fragment is easier to digest and to compare with the original code, than the one highlighted by this comment:

```java
// current
if (nullSafeCompareTo(maxSetVersion, setVersion) <= 0
        && (nullSafeCompareTo(maxSetVersion, setVersion) != 0 || nullSafeCompareTo(maxElectionId, electionId) <= 0)) {
    LOGGER.info(format("Setting max election id to %s from replica set primary %s",
            newDescription.getElectionId(), newDescription.getAddress()));
    maxElectionId = electionId;
} else {
    invalidatePotentialPrimary(newDescription);
    return false;
}
```
```java
// proposed equivalent
if (setVersion != null && electionId != null) {
    if (nullSafeCompareTo(maxSetVersion, setVersion) > 0
            || (nullSafeCompareTo(maxSetVersion, setVersion) == 0 && nullSafeCompareTo(maxElectionId, electionId) > 0)) {
        invalidatePotentialPrimary(newDescription);
        return false;
    }
    LOGGER.info(format("Setting max election id to %s from replica set primary %s",
            newDescription.getElectionId(), newDescription.getAddress()));
    maxElectionId = electionId;
}
```

The proposed fragment also makes it clear that the new code has a different logging behavior: previously `"Setting max election id to %s from replica set primary %s"` was only logged if `!electionId.equals(maxElectionId)`, but now it is logged regardless of this condition.

The proposed fragment is also much simpler to compare with the new logic used if `newDescription.getMaxWireVersion() >= SIX_DOT_ZERO_WIRE_VERSION`, and see that the tuple elements are compared in a different order.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
jyemin: See my comment [here](https://github.com/mongodb/mongo-java-driver/pull/1010#discussion_r988344581) for the justification of the current logic.  If it's not compelling, I'll switch it around again.

stIncMale: In my opinion the "unreversed" code is much simpler to read (I was unable to understand the condition until I reversed it manually), and is more alike with the 6.0+ case above. Maybe it's just me.

jyemin: Sorry, I missed your comment on the changed logging behavior.  That change is unintentional, on both conditional branches, and needs to be reverted.  I did that, and also extracted some methods that hopefully clarify the intention and also why I organized the code this way at a high level.

jyemin:
@stIncMale ready for re-review

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
jyemin: I still wasn't happy with it, so I tried again.  I think the new code flow is much simpler, though it diverges even more from the spec pseudocode.  CC @nbbeeken 

I also simplified the logging to just a single statement, instead of three.  In practice, you'll only get a new electionId or setVersion with a new primary, so a single statement is justified IMO (the logging is also not yet part of the spec, so we can do what we want here).

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -247,64 +246,78 @@ private boolean handleReplicaSetMemberChanged(final ServerDescription newDescrip
...
+        Integer setVersion = description.getSetVersion();
+        if (description.getMaxWireVersion() >= SIX_DOT_ZERO_WIRE_VERSION) {
+            return nullSafeCompareTo(electionId, maxElectionId) <= 0
+                    && (nullSafeCompareTo(electionId, maxElectionId) != 0 || nullSafeCompareTo(setVersion, maxSetVersion) < 0);
+        } else {
+            return setVersion != null && description.getElectionId() != null
+                    && (nullSafeCompareTo(setVersion, maxSetVersion) < 0
+                    || nullSafeCompareTo(setVersion, maxSetVersion) == 0
+                    && nullSafeCompareTo(description.getElectionId(), maxElectionId) < 0);
         }
```
stIncMale, line 288: While
```java
nullSafeCompareTo(electionId, maxElectionId) <= 0 && (nullSafeCompareTo(electionId, maxElectionId) != 0 || nullSafeCompareTo(setVersion, maxSetVersion) < 0)
```
is trivially a correct negation of
```java
nullSafeCompareTo(electionId, maxElectionId) > 0 || (nullSafeCompareTo(electionId, maxElectionId) == 0 && nullSafeCompareTo(setVersion, maxSetVersion) >= 0)
```
my brain is absolutely unable (and I tried hard, I even drew truth tables) to make any sense of the former, while the latter is perfectly clear.

Can we please rewrite this condition as follows?
```java
nullSafeCompareTo(electionId, maxElectionId) < 0 || (nullSafeCompareTo(electionId, maxElectionId) == 0 && nullSafeCompareTo(setVersion, maxSetVersion) < 0)
```

Not only this way it is clear what's going on, but also it matches the structure of the pre 6.0 condition.

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -247,64 +246,78 @@ private boolean handleReplicaSetMemberChanged(final ServerDescription newDescrip
...
+        ObjectId electionId = description.getElectionId();
+        Integer setVersion = description.getSetVersion();
+        if (description.getMaxWireVersion() >= SIX_DOT_ZERO_WIRE_VERSION) {
+            return nullSafeCompareTo(electionId, maxElectionId) <= 0
+                    && (nullSafeCompareTo(electionId, maxElectionId) != 0 || nullSafeCompareTo(setVersion, maxSetVersion) < 0);
+        } else {
+            return setVersion != null && description.getElectionId() != null
+                    && (nullSafeCompareTo(setVersion, maxSetVersion) < 0
+                    || nullSafeCompareTo(setVersion, maxSetVersion) == 0
+                    && nullSafeCompareTo(description.getElectionId(), maxElectionId) < 0);
```
stIncMale, line 287: While this condition is correct as a result of `&&` taking precedence over `||` as per https://docs.oracle.com/javase/tutorial/java/nutsandbolts/operators.html, it is not necessary obvious to a reader (I don't remember operator precedence, so I had to Google it and then analyze the expression). Can we please put some parentheses to make the expression more readable?

```java
...
&& (nullSafeCompareTo(setVersion, maxSetVersion) < 0
|| (nullSafeCompareTo(setVersion, maxSetVersion) == 0 // a new `(` in the beginning
&& nullSafeCompareTo(description.getElectionId(), maxElectionId) < 0)) // a new `)` at the end
```

stIncMale (changes requested):
The new code is so much better than the version I reviewed the first time! 👏  I still have a couple of readability concerns, which may be caused by my stupidity / lack of knowledge, but if the proposed code is as readable to you as the current code, it'd be great to apply the proposed changes, because to me it drastically affects the readability.

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
stIncMale: > I think the new code flow is much simpler, though it diverges even more from the spec pseudocode.

I think this simply means that it may be a good idea to refactor the the spec pseudocode.

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
jyemin: Done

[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
jyemin: Much better.  Done.

jyemin:
@stIncMale ready for review

stIncMale (approved):

PR #1011, opened 2022-10-06
J4745 experiment

jyemin:


bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
```diff
@@ -17,15 +17,18 @@
...
+import org.bson.internal.CodecCache.CodecCacheKey;
 
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.List;
 
 import static org.bson.assertions.Assertions.isTrueArgument;
 
-public final class ProvidersCodecRegistry implements CodecRegistry, CycleDetectingCodecRegistry {
+public final class ProvidersCodecRegistry implements CycleDetectingCodecRegistry {
```
jyemin, line 31: If I make this implement `ParameterizationAwareCodecRegistry`, all the `RecordCodecTest` tests pass.  If I leave it like this, they fail as indicated on the other PR

jyemin: My opinion continues to be that the approach in the other PR is less risky, since it results in an exception in all cases rather than an exception in some and type system violations in others.

jyemin: It's actually weirder than I thought.  I just noticed that when I run the record codec tests, the simpler parameterization tests fail, e.g. `testRecordWithNestedMapOfRecords`, which the more complicated ones, e.g. `testSelfReferentialRecords` actually succeed.  That's because in the more nested scenarios like the latter, the `CodecRegistry `instance that the `RecordCodec` gets passed to its constructor is actually an instance of `ChildCodecRegistry`, which _does_ implement `ParameterizationAwareCodecRegistry`, so all the logic that we want does get executed in that case.

I think this will be hard to explain/support.

jyemin: I have an idea, demonstrated in the last commit, that might get us out of this mess.  The key to it is that  `CodecRegistry` extends `CodecProvider`.  So we can always take a `CodecRegistry` and wrap it in another `CodecRegistry` using the `CodecRegistries.fromProviders` method.

So how about if MongoClient wraps the registry from MongoClientSettings using that `fromProviders` method, thus treating the application-provided registry as a `CodecProvider`?  In fact, we already use that trick to apply the `UuidRepresentation`.  See `CodecRegistries#withUuidRepresentation` and how it's used in MongoClient implementations.

jyemin: Demonstrated the idea [here](https://github.com/mongodb/mongo-java-driver/pull/1007/commits/126853ac7d79074be7de85fb2d9c91a73a1c40f1)

jyemin:
Further conversation can be had on the main PR for this issue.

PR #1012, opened 2022-10-06
Add javadoc to `AbstractMapCodec`

stIncMale:
`AbstractMapCodec` was added to the public API in https://github.com/mongodb/mongo-java-driver/commit/26b15f387e14a56afb2003970e9ee22d32b01861. It is not clear to me whether it was intentional, and the class is designed to be extended by users, or accidental. Without a documentation comment, `javadoc` spits out a warning during the build process.


jyemin:
I think we should make it package private instead.

jyemin (approved):

PR #1013, opened 2022-10-08
Propagate the original error for write errors labeled `NoWritesPerformed`

stIncMale:
TODO for @stIncMale: what's achieved by introducing `NoWritesPerformed` given that we already have `RetryableWriteError`? 

JAVA-4701

### DIFF
--- a/driver-core/src/test/functional/com/mongodb/internal/operation/MixedBulkWriteOperationSpecification.groovy
+++ b/driver-core/src/test/functional/com/mongodb/internal/operation/MixedBulkWriteOperationSpecification.groovy
@@ -846,14 +846,14 @@ class MixedBulkWriteOperationSpecification extends OperationFunctionalSpecificat
         given:
         getCollectionHelper().insertDocuments(getTestInserts())
         def operation = new MixedBulkWriteOperation(getNamespace(),
-                [new InsertRequest(new BsonDocument('_id', new BsonInt32(7))),
-                 new InsertRequest(new BsonDocument('_id', new BsonInt32(1)))   // duplicate key
-                ], false, ACKNOWLEDGED, true)
+                [new DeleteRequest(new BsonDocument('_id', new BsonInt32(2))), // existing key
+                 new InsertRequest(new BsonDocument('_id', new BsonInt32(1)))  // existing (duplicate) key
+                ], true, ACKNOWLEDGED, true)
 
         def failPoint = BsonDocument.parse('''{
             "configureFailPoint": "failCommand",
             "mode": {"times": 2 },
-            "data": { "failCommands": ["insert"],
+            "data": { "failCommands": ["delete"],
                       "writeConcernError": {"code": 91, "errmsg": "Replication is being shut down"}}}''')
         configureFailPoint(failPoint)
 
--- a/driver-sync/src/test/functional/com/mongodb/client/Fixture.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/Fixture.java
@@ -20,10 +20,13 @@ import com.mongodb.ConnectionString;
 import com.mongodb.MongoClientSettings;
 import com.mongodb.ServerAddress;
 import com.mongodb.client.internal.MongoClientImpl;
+import com.mongodb.connection.ClusterDescription;
 import com.mongodb.connection.ServerDescription;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 
 import static com.mongodb.ClusterFixture.getConnectionString;
 import static com.mongodb.ClusterFixture.getMultiMongosConnectionString;
@@ -110,11 +113,39 @@ public final class Fixture {
     }
 
     public static ServerAddress getPrimary() throws InterruptedException {
-        getMongoClient();
-        List<ServerDescription> serverDescriptions = getPrimaries(((MongoClientImpl) mongoClient).getCluster().getDescription());
+        return getPrimary(getMongoClient());
+    }
+
+    /**
+     * Beware of a potential race condition hiding here: the primary you discover may differ from the one used by the {@code client}
+     * when performing some operations, as the primary may change.
+     */
+    public static ServerAddress getPrimary(final MongoClient client)
+            throws InterruptedException {
+        final Supplier<ClusterDescription> clusterDescriptionSupplier;
+        if (client instanceof MongoClientImpl) {
+            clusterDescriptionSupplier = () -> ((MongoClientImpl) client).getClusterDescription();
+        } else {
+            // com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient
+            try {
+                Object wrappedReactiveClient = client.getClass().getMethod("getWrapped").invoke(client);
+                clusterDescriptionSupplier =
+                        () -> {
+                            try {
+                                return (ClusterDescription) wrappedReactiveClient.getClass().getMethod("getClusterDescription")
+                                        .invoke(wrappedReactiveClient);
+                            } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+                                throw new RuntimeException(e);
+                            }
+                        };
+            } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+                throw new RuntimeException(e);
+            }
+        }
+        List<ServerDescription> serverDescriptions = getPrimaries(clusterDescriptionSupplier.get());
         while (serverDescriptions.isEmpty()) {
             Thread.sleep(100);
-            serverDescriptions = getPrimaries(((MongoClientImpl) mongoClient).getCluster().getDescription());
+            serverDescriptions = getPrimaries(clusterDescriptionSupplier.get());
         }
         return serverDescriptions.get(0).getAddress();
     }
--- a/driver-sync/src/test/functional/com/mongodb/client/Fixture.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/Fixture.java
@@ -19,14 +19,10 @@ package com.mongodb.client;
 import com.mongodb.ConnectionString;
 import com.mongodb.MongoClientSettings;
 import com.mongodb.ServerAddress;
-import com.mongodb.client.internal.MongoClientImpl;
-import com.mongodb.connection.ClusterDescription;
 import com.mongodb.connection.ServerDescription;
 
-import java.lang.reflect.InvocationTargetException;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Supplier;
 
 import static com.mongodb.ClusterFixture.getConnectionString;
 import static com.mongodb.ClusterFixture.getMultiMongosConnectionString;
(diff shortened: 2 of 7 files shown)


driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -179,6 +189,64 @@ public static <R> void poolClearedExceptionMustBeRetryable(
...
+                                .append("failCommands", new BsonArray(singletonList(new BsonString("insert"))))
+                                .append("errorCode", new BsonInt32(errorCode))
+                                .append("errorLabels", new BsonArray(errorLabels.stream().map(BsonString::new).collect(Collectors.toList()))));
+        ServerAddress primaryServerAddress = Fixture.getPrimary();
+        CompletableFuture<FailPoint> futureFailPoint = new CompletableFuture<>();
+        CommandListener commandListener = new CommandListener() {
+            private final AtomicBoolean configureFailPoint = new AtomicBoolean(true);
+
+            @Override
+            public void commandFailed(final CommandFailedEvent event) {
```
stIncMale, line 218: The prose test [says to use `CommandSucceededEvent`](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.rst#prose-tests:~:text=Via%20the%20command%20monitoring%20CommandSucceededEvent), but the server responds with `ok: 0` to the first attempt in the test, which means, we need to use `CommandFailedEvent`. [People observer different server behavior when manually executing the test steps in `mongosh`](https://mongodb.slack.com/archives/C72LB5RPV/p1665185092241459?thread_ts=1665183651.355829&cid=C72LB5RPV), but I have not been able to observe the same.

This situation is weird and is likely caused by a difference in the `insert` command that is being sent, though I don't know what that difference is.

jyemin:
@stIncMale have a look at https://jira.mongodb.org/browse/SERVER-66116 for the original report that led to this.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
jyemin: Let's get to the bottom of this before merging this PR

jyemin (changes requested):
Commented

driver-core/src/test/functional/com/mongodb/internal/operation/MixedBulkWriteOperationSpecification.groovy
```diff
@@ -846,14 +846,14 @@ class MixedBulkWriteOperationSpecification extends OperationFunctionalSpecificat
...
+                 new InsertRequest(new BsonDocument('_id', new BsonInt32(1)))  // existing (duplicate) key
+                ], true, ACKNOWLEDGED, true)
 
         def failPoint = BsonDocument.parse('''{
             "configureFailPoint": "failCommand",
             "mode": {"times": 2 },
-            "data": { "failCommands": ["insert"],
+            "data": { "failCommands": ["delete"],
                       "writeConcernError": {"code": 91, "errmsg": "Replication is being shut down"}}}''')
         configureFailPoint(failPoint)
```
stIncMale, line 858: I don't have a complete understanding of these test, but here is what I think I understand:

- it needs to fail the first attempt with a write concern error, and then it needs to see that the it is retried;
- to achieve the above, it fails the second attempt with a different error (tries to insert a duplicate key), and expects to observe it.

The latest server labels the second error as `NoWritesPerformed`, because it fails to insert a duplicate key. As a result, the updated driver chooses the first error, thus not letting the test to observe the error from the second attempt, i.e., not letting the test to see that the command was retried. By deleting one of the existing keys, the test prevents the server from reporting `NoWritesPerformed`, which leads to the test being able to observe the error from the second attempt.

stIncMale:
It turned out that the fail points from the new nests did not leak into the failing tests, which means that the changes to `Fixture` and `FailPoint` are not necessary. However, these changes do prevent leaking fail points in some cases, which is why I think they should be committed. If you think they should be moved to a different PR, please let me know.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -179,6 +188,74 @@ public static <R> void poolClearedExceptionMustBeRetryable(
...
+                                        .map(BsonString::new).collect(Collectors.toList()))))
+                        .append("failCommands", new BsonArray(singletonList(new BsonString("insert")))));
+        try (MongoClient client = clientCreator.apply(getMongoClientSettingsBuilder()
+                .retryWrites(true)
+                .addCommandListener(commandListener)
+                .applyToServerSettings(builder ->
+                        // see `poolClearedExceptionMustBeRetryable` for the explanation
+                        builder.heartbeatFrequency(50, TimeUnit.MILLISECONDS))
+                .build());
+             FailPoint ignored = FailPoint.enable(failPointDocument, Fixture.getPrimary(client))) {
```
jyemin, line 243: I tried passing in `primaryServerAddress` instead of calling `Fixture.getPrimary()` again and the test still passed. Do you have a reason to think it's necessary to call it again with this MongoClient?  The primary should be the same regardless.

driver-sync/src/test/functional/com/mongodb/client/Fixture.java
```diff
@@ -110,11 +113,39 @@ public static MongoClientSettings.Builder getMongoClientSettings(final Connectio
...
+     * Beware of a potential race condition hiding here: the primary you discover may differ from the one used by the {@code client}
+     * when performing some operations, as the primary may change.
+     */
+    public static ServerAddress getPrimary(final MongoClient client)
+            throws InterruptedException {
+        final Supplier<ClusterDescription> clusterDescriptionSupplier;
+        if (client instanceof MongoClientImpl) {
+            clusterDescriptionSupplier = () -> ((MongoClientImpl) client).getClusterDescription();
+        } else {
+            // com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient
```
jyemin, line 129: I'm unclear why this is necessary, as `getClusterDescription` is defined on the `com.mongodb.client.MongoClient` interface not `MongoClientImpl` and is implement in `SyncMongoClient`.

Also see my comment on the new prose test, as it's also unclear to me why this method needs to be called again with its own MongoClient instance

jyemin (changes requested):

jyemin:
> It turned out that the fail points from the new nests did not leak into the failing tests, which means that the changes to Fixture and FailPoint are not necessary. However, these changes do prevent leaking fail points in some cases, which is why I think they should be committed. If you think they should be moved to a different PR, please let me know.

I only noticed this comment after posting mine.  I don't mind the change to Failpoint, but I don't see the need for the change to Fixture.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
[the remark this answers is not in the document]
stIncMale: It's actually not, as it's within `try (...here...)`.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -179,6 +188,74 @@ public static <R> void poolClearedExceptionMustBeRetryable(
...
 
+    /**
+     * Prose test #3.
+     */
+    @Test
+    public void originalErrorMustBePropagatedIfNoWritesPerformed() throws InterruptedException {
+        originalErrorMustBePropagatedIfNoWritesPerformed(MongoClients::create);
+    }
+
+    @SuppressWarnings("try")
```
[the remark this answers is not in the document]
stIncMale: It suppressed the following compiler warning when running `./gradlew clean check -x test -x integrationTest` with OpenJDK JDK 17:

```
> Task :driver-sync:compileTestGroovy
/Users/valentin.kovalenko/Documents/projects/mongo-java-driver/driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java:243: warning: [try] auto-closeable resource ignored is never referenced in body of corresponding try statement
             FailPoint ignored = FailPoint.enable(failPointDocument, Fixture.getPrimary(client))) {
                       ^
1 warning
```

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -179,6 +188,74 @@ public static <R> void poolClearedExceptionMustBeRetryable(
...
+                        // see `poolClearedExceptionMustBeRetryable` for the explanation
+                        builder.heartbeatFrequency(50, TimeUnit.MILLISECONDS))
+                .build());
+             FailPoint ignored = FailPoint.enable(failPointDocument, Fixture.getPrimary(client))) {
+            MongoCollection<Document> collection = client.getDatabase(getDefaultDatabaseName())
+                    .getCollection("originalErrorMustBePropagatedIfNoWritesPerformed");
+            collection.drop();
+            try {
+                collection.insertOne(new Document());
+            } catch (MongoWriteConcernWithResponseException e) {
```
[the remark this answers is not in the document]
stIncMale: If it is leaking, which I don't think is the case, then `MongoClientImpl` is also leaking in the same way: both classes are `public`, and user may use `catch`/`instanceof` to check for these types. I will replace it with `MongoException` to avoid depending on an unnecessary strict type, given that the type is not part of the public API.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -179,6 +188,74 @@ public static <R> void poolClearedExceptionMustBeRetryable(
...
+            final Function<MongoClientSettings, MongoClient> clientCreator) throws InterruptedException {
+        assumeTrue(serverVersionAtLeast(6, 0) && isDiscoverableReplicaSet());
+        ServerAddress primaryServerAddress = Fixture.getPrimary();
+        CompletableFuture<FailPoint> futureFailPointFromListener = new CompletableFuture<>();
+        CommandListener commandListener = new CommandListener() {
+            private final AtomicBoolean configureFailPoint = new AtomicBoolean(true);
+
+            @Override
+            public void commandSucceeded(final CommandSucceededEvent event) {
+                if (event.getCommandName().equals("insert") && configureFailPoint.compareAndSet(true, false)) {
```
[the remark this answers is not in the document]
stIncMale: Apparently, we don't. However, thank you for spotting this, maybe when the test execution goes sideways, not having a second fail point simplifies figuring out the cause of the failure. I'll add the condition.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: > I tried passing in primaryServerAddress instead of calling Fixture.getPrimary() again and the test still passed.

It's just to reduce the probability of encountering a race condition described on the new `Fixture.getPrimary` method. However, since I am neglecting the same when calling `FailPoint.enable` from a listener (just to avoid dealing with `InterruptedException` from `Fixture.getPrimary`  in `CommandListener.commandSucceeded`), there is no sense to start caring about it in a different place in the test. I'll change this.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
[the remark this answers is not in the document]
jyemin: That's not quite what I mean.  If I recall correctly, we never intended for _instances_ of this class to be thrown all the way to the application.  Rather, I think the intention was for it to be thrown, caught internally by the driver, and eventually re-thrown as a different exception, e.g. a `MongoBulkWriteException`.    I don't remember the details of why these convolutions were necessary, but it seems to have originated in the commit for JAVA-2847.

driver-sync/src/test/functional/com/mongodb/client/Fixture.java
stIncMale: > getClusterDescription is defined on the com.mongodb.client.MongoClient interface not MongoClientImpl and is implement in SyncMongoClient

Thank you! I was confused by the original code of the method casting to `MongoClientImpl`. The implementation is now much simpler.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: This discrepancy as caused by me creating an incorrect fail point document.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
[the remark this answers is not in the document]
stIncMale: Done.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: Done.

driver-sync/src/test/functional/com/mongodb/client/Fixture.java
stIncMale: Done.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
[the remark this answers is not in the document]
stIncMale: I created https://jira.mongodb.org/browse/JAVA-4775 to track this bug, which I introduced in 4.4.0 (more details are in the ticket).

jyemin (approved):
LGTM

PR #1017, opened 2022-10-17
Use AWS SDK v2/v1 if available for AWS credential fetching

jyemin:
To fetch AWS credentials from the environment:

  * Use AWS SDK v2 if available on the classpath
  * Otherwise, use AWS SDK v2 if available on the classpath
  * Otherwise, use the existing driver implementation

JAVA-4718

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -372,7 +372,7 @@ functions:
             MONGODB_URI="mongodb://$USER:$PASS@localhost"
           EOF
           JAVA_VERSION=${JAVA_VERSION} PROJECT_DIRECTORY=${PROJECT_DIRECTORY} \
-          USE_BUILT_IN_AWS_CREDENTIAL_PROVIDER=${USE_BUILT_IN_AWS_CREDENTIAL_PROVIDER} \
+          AWS_CREDENTIAL_PROVIDER=${AWS_CREDENTIAL_PROVIDER} \
           .evergreen/run-mongodb-aws-test.sh
 
   "run aws auth test with assume role credentials":
@@ -403,7 +403,7 @@ functions:
             MONGODB_URI="mongodb://$USER:$PASS@localhost"
           EOF
           JAVA_VERSION=${JAVA_VERSION} PROJECT_DIRECTORY=${PROJECT_DIRECTORY} DRIVERS_TOOLS=${DRIVERS_TOOLS} \
-          USE_BUILT_IN_AWS_CREDENTIAL_PROVIDER=${USE_BUILT_IN_AWS_CREDENTIAL_PROVIDER} \
+          AWS_CREDENTIAL_PROVIDER=${AWS_CREDENTIAL_PROVIDER} \
           .evergreen/run-mongodb-aws-test.sh
 
   "run aws auth test with aws EC2 credentials":
@@ -424,7 +424,7 @@ functions:
           ${PREPARE_SHELL}
           # Write an empty prepare_mongodb_aws so no auth environment variables are set.
           echo "" > "${PROJECT_DIRECTORY}/prepare_mongodb_aws.sh"
-          JAVA_VERSION=${JAVA_VERSION} .evergreen/run-mongodb-aws-test.sh
+          JAVA_VERSION=${JAVA_VERSION} AWS_CREDENTIAL_PROVIDER=${AWS_CREDENTIAL_PROVIDER} .evergreen/run-mongodb-aws-test.sh
 
   "run aws auth test with aws credentials as environment variables":
     - command: shell.exec
@@ -453,7 +453,7 @@ functions:
         working_dir: "src"
         script: |
           ${PREPARE_SHELL}
-          JAVA_VERSION=${JAVA_VERSION} .evergreen/run-mongodb-aws-test.sh
+          JAVA_VERSION=${JAVA_VERSION} AWS_CREDENTIAL_PROVIDER=${AWS_CREDENTIAL_PROVIDER} .evergreen/run-mongodb-aws-test.sh
 
   "run aws auth test with aws credentials and session token as environment variables":
     - command: shell.exec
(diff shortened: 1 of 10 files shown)


driver-core/src/main/com/mongodb/internal/authentication/AwsCredentialHelper.java
```diff
@@ -17,61 +17,49 @@
...
 
 /**
  * Utility class for working with AWS authentication.
  *
  * <p>This class should not be considered a part of the public API.</p>
  */
 public final class AwsCredentialHelper {
+    public static final Logger LOGGER = Loggers.getLogger("authenticator");
 
-    public static AwsCredential obtainFromEnvironment() {
```
jyemin, line 34: Refactored all this code into `BuiltInAwsCredentialProvider`

driver-core/src/main/com/mongodb/internal/authentication/AwsCredentialHelper.java
```diff
@@ -17,61 +17,49 @@
...
-                ec2OrEcsResponse.getString("Token").getValue());
+    static {
+        try {
+            Class.forName("software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider");
+            awsCredentialSupplier = new AwsSdkV2CredentialSupplier();
+            LOGGER.info("Using software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider from AWS SDK v2 to retrieve AWS "
+                    + "credentials");
+        } catch (ClassNotFoundException e) {
+            awsCredentialSupplier = new BuiltInAwsCredentialSupplier();
+            LOGGER.info("Using built-in driver implementation to retrieve AWS credentials. Consider adding a dependency to "
```
jyemin, line 46: Consider whether this should be logged as a warning.

rozza (changes requested):
One nit - then LGTM

driver-core/build.gradle
```diff
@@ -47,6 +47,9 @@ dependencies {
     api "io.netty:netty-buffer", optional
     api "io.netty:netty-transport", optional
     api "io.netty:netty-handler", optional
+
+    implementation 'software.amazon.awssdk:auth:2.17.261', optional
```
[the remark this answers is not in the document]
jyemin: Done

jyemin:
I added two more features:

1. Support for AWS SDK v1 (requested by a user of Amazon Glue, where only v1 is provided)
2. Support for testing all configurations in ECS (required a different way of specifying the three configurations

Boy, there are a ton of tasks now.  We should consider a different way of doing this in Evergreen.

blink1073 (approved):

driver-core/build.gradle
```diff
@@ -47,6 +47,12 @@ dependencies {
     api "io.netty:netty-buffer", optional
     api "io.netty:netty-transport", optional
     api "io.netty:netty-handler", optional
+
+    // Optionally depend on both AWS SDK v2 and v1.  The driver will use v2 is present, v1 if present, or built-in functionality if
+    // neither are present
+    implementation "software.amazon.awssdk:auth:$awsSdkV2Version", optional
+    implementation "com.amazonaws:aws-java-sdk-core:$awsSdkV1Version", optional
```
rozza, line 54: Would it be worth making v1 a compileOnly dependency and add it to testImplementation.  That way it wouldn't show up in the pom but still can be found if on the runtime path? It would be another hint to use the v2 version.

.evergreen/.evg.yml
```diff
@@ -1813,15 +1833,22 @@ buildvariants:
     - name: "plain-auth-test"
 
 - matrix_name: "aws-auth-test"
-  matrix_spec: { ssl: "nossl", jdk: ["jdk8", "jdk17"], version: ["4.4", "5.0", "6.0", "latest"], os: "ubuntu" }
-  display_name: "MONGODB-AWS Auth test ${version} ${jdk}"
+  matrix_spec: { ssl: "nossl", jdk: ["jdk8", "jdk17"], version: ["4.4", "5.0", "6.0", "latest"], os: "ubuntu",
+                 aws-credential-provider: "*" }
+  display_name: "MONGODB-AWS Auth test ${version} ${jdk} ${aws-credential-provider}"
```
rozza, line 1838: Do we need to run all tests on all variants? 

Do we need to setup any AWS specific environment variables to test the v2/v1 sdks pick them up?

rozza (commented):
Looking good - not sure how to reduce the evergreen load.

Have a couple of questions for consideration

driver-core/build.gradle
jyemin: I'm not sure... According to the [blog](https://blog.gradle.org/introducing-compile-only-dependencies) introducing compileOnly dependencies:

> Dependencies required at compile time but required at runtime only when using certain features, a.k.a. optional dependencies

It sounds applicable to _any_ optional dependency, which would include both v1 and v2 at this point, as well as a few others that we have.  Should they all be compileOnly? It's not clear to me when to use compileOnly and when to use optional (is optional provided natively, or via plugin?).

.evergreen/.evg.yml
jyemin: > Do we need to run all tests on all variants?

You could argue that as long as we can prove that _any_ credential fetching is working for the SDK case, then _all_ of them will work.  @blink1073 suggested that we test all cases, so I went with that just to be safe.  But it might be too safe.  

One possibility is that we test all the different credential-fetching scenarios in a single task (one task for each of the three fetching implementations).  That will reduce the number of tasks 

> Do we need to setup any AWS specific environment variables to test the v2/v1 sdks pick them up?

No, the SDK uses the same environment variables as we do.  Our built-in implementation is intended to emulate the SDK.

blink1073: We run the Python AWS auth tests against MongoDB 4.4, 5.0, 6.0, rapid, and latest.  We run against a single version of Python on macos and ubuntu, and all supported versions of Python on Windows.  The ECS and EC2 tests only run on Ubuntu.

driver-core/build.gradle
jyemin: Just saw this in the release notes for compileOnly:

> However, unlike Maven provided dependencies, compile only dependencies in Gradle are not included on the test classpath

But we need this in our test classpath at least for some tasks.

rozza: So m,y question was really about about signalling usage in the pom.xml.  If we add it as a `compileOnly` and `testImplementation` dependency, it won't be mentioned in the pom.xml but will still be testable.

So do we want to advertize it as an optional dependency as well as v2 or just make it an implicit but tested dependency.

.evergreen/.evg.yml
rozza: Thanks - resolving.

rozza (approved):
LGTM - happy either way for the v1 dependency to be in the pom

PR #1018, opened 2022-10-17
Simplify `CodecCache`

stIncMale:
JAVA-4751


bson/src/main/org/bson/internal/CodecCache.java
```diff
@@ -65,29 +64,18 @@ public String toString() {
...
-
-    @SuppressWarnings("unchecked")
-    public synchronized <T> Codec<T> putIfMissing(final CodecCacheKey codecCacheKey, final Codec<T> codec) {
-        Optional<Codec<?>> cachedCodec = codecCache.computeIfAbsent(codecCacheKey, clz -> Optional.of(codec));
-        if (cachedCodec.isPresent()) {
-            return (Codec<T>) cachedCodec.get();
-        }
-        codecCache.put(codecCacheKey, Optional.of(codec));
-        return codec;
-    }
```
stIncMale, line 86: Instead of improving this method as I originally described [here](https://github.com/mongodb/mongo-java-driver/pull/1007#discussion_r987300328), I did more changes, further simplifying the cache. Before this PR we were caching failures of finding a codec as `Optional.empty()` values just for the sake of throwing an exception faster if asked for the same codec again. I see no point in trying to optimize that path.

bson/src/test/unit/org/bson/internal/ProvidersCodecRegistrySpecification.groovy
```diff
@@ -115,22 +115,32 @@ class ProvidersCodecRegistrySpecification extends Specification {
...
+
+            @Override
+            Codec get(final Class clazz, final CodecRegistry registry) {
+                if (counter == 0) {
+                    counter++
+                    return codec
+                }
+                throw new AssertionError((Object)'Must not be called more than once.')
+            }
+        }
```
stIncMale, line 130: For some reason, I failed to use [Spock stubbing](https://spockframework.org/spock/docs/2.0/all_in_one.html#_returning_fixed_values) here (the stabbed `provider` works as expected if used directly from the test, but does not work when used via `registry`. So I had to stub the object manually.

jyemin (approved):
LGTM

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
```diff
@@ -67,19 +68,18 @@ public <T> Codec<T> get(final Class<T> clazz, final CodecRegistry registry) {
...
-        if (!codecCache.containsKey(codecCacheKey)) {
+        return codecCache.<T>get(codecCacheKey).orElseGet(() -> {
             for (CodecProvider provider : codecProviders) {
                 Codec<T> codec = provider.get(context.getCodecClass(), context);
                 if (codec != null) {
                     if (codec instanceof Parameterizable && context.getTypes().isPresent()) {
                         codec = (Codec<T>) ((Parameterizable) codec).parameterize(context, context.getTypes().get());
                     }
-                    return codecCache.putIfMissing(codecCacheKey, codec);
+                    return codecCache.putIfAbsent(codecCacheKey, codec);
```
stIncMale, line 78: I have just realized that this whole thing can and should be done as a single `compute`. Will do.

rozza (approved):
Very nice - LGTM

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
stIncMale: Turns out, `IllegalStateException: Recursive update` is a thing despite the map being concurrent :( Well, we can't have all the good things.

stIncMale:
> I think you're going to have to revert the last commit as many tests are failing with a "recursive update" exception thrown by ConcurrentHashMap.

Yes. While `ConcurrentMap.computeIfAbsent` does not have this restriction, `ConcurrentHashMap.computeIfAbsent` documents it.

jyemin (approved):

PR #1019, opened 2022-10-19
Support decoding to Collection and Map subclasses

jyemin:
JAVA-4742

The main complication to adding this feature is that the existing IterableCodec is not able to extend a class with the parameterized types `<T, C extends Collection<T>>`.  The same goes for MapCodec extending a class with the parameterized types `<T, M extends Map<String, T>>`.  The type system prevents this in the face of the default behavior of using ArrayList and HashMap, respectively, for the existing public constructors.  Also, since Iterable does not have an `add` method, it's problematic to add a decode method that supports parameterization with Iterable.

I tried a bunch of experiments to get these codecs to work with this new feature, failed, and went with the following approach:

- Revert the recent changes to IterableCodec and MapCodec that added support for parameterization.  Since they are part of the API they can't be removed, but they are deprecated.
- Add new CollectionCodec and MapCodecV2 codecs that don't suffer from the same problems
- Add CollectionCodecProvider and use it in all the default registries ahead of IterableCodecProvider (which has to remain to support case where it actually _is_ an Iterable that is not a Collection.  Without this we can't encode anything created by our own `com.mongodb.internal.Iterables` class, for example. )
- Create instances of MapCodecV2 instead of MapCodec in MapCodecProvider


bson/src/main/org/bson/codecs/IterableCodecProvider.java
```diff
@@ -27,7 +27,9 @@
  * A {@code CodecProvider} for classes than implement the {@code Iterable} interface.
  *
  * @since 3.3
+ * @deprecated Prefer {@link CollectionCodecProvider}
  */
+@Deprecated
```
rozza, line 32: As we need this for our own driver code - should it actually be deprecated?

rozza (changes requested):

bson/src/main/org/bson/codecs/IterableCodecProvider.java
jyemin: Maybe not.  I'll think about it more.

rozza:
I'm of the opinion that if we require the IterableCodec / IterableCodecProvider it should not be deprecated but the suggestion to use the CollectionCodecProvider instead works.  

If we really want to deprecate it, it might be better to do so it in a separate ticket and refactor our code that requires it and remove it from the default registry (and accept it might break some niche usecases).

bson/src/main/org/bson/codecs/MapCodecV2.java
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @param <M> the actual type of the Map, e.g. {@code NavigableMap<String, Object>}
+ */
+@SuppressWarnings("rawtypes")
+final class MapCodecV2<M extends Map<String, Object>> extends AbstractMapCodec<Object, M>
+        implements OverridableUuidRepresentationCodec<M>, Parameterizable {
+
+    private static final CodecRegistry DEFAULT_REGISTRY = fromProviders(asList(new ValueCodecProvider(), new BsonValueCodecProvider(),
+            new DocumentCodecProvider(), new CollectionCodecProvider(), new MapCodecProvider()));
+    private static final BsonTypeClassMap DEFAULT_BSON_TYPE_CLASS_MAP = new BsonTypeClassMap();
```
stIncMale, line 53: Both are unused.

bson/src/main/org/bson/codecs/MapCodecV2.java
```diff
@@ -0,0 +1,119 @@
...
+    private final BsonTypeCodecMap bsonTypeCodecMap;
+    private final CodecRegistry registry;
+    private final Transformer valueTransformer;
+    private final UuidRepresentation uuidRepresentation;
+
+    /**
+     /**
```
stIncMale, line 60: One of these lines is not needed.

bson/src/test/unit/org/bson/codecs/CollectionCodecSpecification.groovy
```diff
@@ -0,0 +1,235 @@
...
+import static org.bson.UuidRepresentation.C_SHARP_LEGACY
+import static org.bson.UuidRepresentation.JAVA_LEGACY
+import static org.bson.UuidRepresentation.PYTHON_LEGACY
+import static org.bson.UuidRepresentation.STANDARD
+import static org.bson.UuidRepresentation.UNSPECIFIED
+import static org.bson.codecs.configuration.CodecRegistries.fromCodecs
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders
+import static org.bson.codecs.configuration.CodecRegistries.fromRegistries
+
+class CollectionCodecSpecification extends Specification {
```
stIncMale, line 42: Currently these tests pass if the `else` branch in `AbstractCollectionCodec` is replaced with `throw Assertions.fail();`. It appears that we lack a test that decodes to a custom `Collection` that has a no-arg constructor. `CopyOnWriteArrayList` can be used as such a custom `Collection` in the `should decode to specified generic class` test.

Note that a similar thing is **not** applicable to `MapCodecV2Specification`, as it tests against `WeakHashMap`.

bson/src/test/unit/org/bson/codecs/MapCodecV2Specification.groovy
```diff
@@ -64,11 +64,11 @@ import static org.bson.codecs.configuration.CodecRegistries.fromCodecs
 import static org.bson.codecs.configuration.CodecRegistries.fromProviders
 import static org.bson.codecs.configuration.CodecRegistries.fromRegistries
 
-class MapCodecSpecification extends Specification {
```
stIncMale, line 67: We still have `IterableCodecSpecification`, but we are getting rid of `MapCodecSpecification`. Shouldn't the latter also stay? It seems reasonable to delete `IterableCodecSpecification`/`MapCodecSpecification` together with `IterableCodec`/`MapCodec`, but not before the latter are deleted.

stIncMale (changes requested):

bson/src/main/org/bson/codecs/IterableCodec.java
```diff
@@ -17,26 +17,27 @@
...
  * Encodes and decodes {@code Iterable} objects.
  *
  * @since 3.3
+ * @deprecated Prefer {@link CollectionCodecProvider}
  */
+@Deprecated
 @SuppressWarnings("rawtypes")
-public class IterableCodec extends AbstractIterableCodec<Object>
-        implements OverridableUuidRepresentationCodec<Iterable<Object>>, Parameterizable {
+public class IterableCodec implements Codec<Iterable>, OverridableUuidRepresentationCodec<Iterable> {
```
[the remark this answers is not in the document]
stIncMale: Nope, no changes are needed as we simply reverted to #864.

bson/src/main/org/bson/codecs/IterableCodecProvider.java
jyemin: Done: removed deprecation.

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -112,7 +114,7 @@ public final class MongoClientSettings {
      * <li>{@link com.mongodb.DBRefCodecProvider}</li>
      * <li>{@link com.mongodb.DBObjectCodecProvider}</li>
      * <li>{@link org.bson.codecs.DocumentCodecProvider}</li>
-     * <li>{@link org.bson.codecs.IterableCodecProvider}</li>
+     * <li>{@link org.bson.codecs.CollectionCodecProvider}</li>
```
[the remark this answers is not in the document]
jyemin: Done: added it back

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
```diff
@@ -78,33 +78,33 @@
...
 import static org.bson.codecs.pojo.Conventions.DEFAULT_CONVENTIONS;
 import static org.bson.codecs.pojo.Conventions.NO_CONVENTIONS;
 import static org.bson.codecs.pojo.Conventions.SET_PRIVATE_FIELDS_CONVENTION;
 import static org.bson.codecs.pojo.Conventions.USE_GETTERS_FOR_SETTERS;
-import static org.bson.codecs.pojo.Conventions.CLASS_AND_PROPERTY_CONVENTION;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
+@SuppressWarnings("deprecated")
```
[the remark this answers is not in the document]
jyemin: Done: fixed typo and removed redundant suppression on methods

bson/src/main/org/bson/codecs/CollectionCodecProvider.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = bsonTypeClassMap.hashCode();
+        result = 31 * result + (valueTransformer != null ? valueTransformer.hashCode() : 0);
+        return result;
```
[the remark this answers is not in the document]
jyemin: No objections.  This is what IntelliJ auto-generated, though I'm not sure why.  When I deleted it and re-generated, it used the preferred approach

Done

bson/src/main/org/bson/codecs/MapCodecV2.java
jyemin: Done: removed

bson/src/main/org/bson/codecs/MapCodecV2.java
jyemin: Done: removed

bson/src/test/unit/org/bson/codecs/MapCodecV2Specification.groovy
jyemin: Done: reverted delete

bson/src/test/unit/org/bson/codecs/CollectionCodecSpecification.groovy
jyemin: Done: added CopyOnWriteArrayList to the where clause of the test.  The only lack of coverage now is the exceptional conditions.  Not sure it's worth it to cover those though.

stIncMale: 🤷‍♂️ I am fine with not having those

stIncMale (approved):

PR #1020, opened 2022-10-20
Document that CodecRegistry instances configured by the application a…

jyemin:
…re treated as CodecProvider

JAVA-4777

Once we agree on the verbiage here, I will copy it to all the other places that the registry is configured (all the withCodecRegistry methods)

### DIFF
--- a/driver-core/src/main/com/mongodb/MongoClientSettings.java
+++ b/driver-core/src/main/com/mongodb/MongoClientSettings.java
@@ -400,9 +400,15 @@ public final class MongoClientSettings {
         /**
          * Sets the codec registry
          *
+         * <p>The {@link CodecRegistry} configured by this method is effectively treated by the driver as an instance of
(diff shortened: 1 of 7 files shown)


rozza (approved):

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -400,9 +400,15 @@ public Builder credential(final MongoCredential credential) {
         /**
          * Sets the codec registry
          *
+         * <p>The {@link CodecRegistry} configured by this method is effectively treated by the driver as an instance of
+         * {@link org.bson.codecs.configuration.CodecProvider}, which {@link CodecRegistry} extends. So there is no benefit to defining
+         * a class that implements {@link CodecRegistry}. Rather, an application should always create {@link CodecRegistry} instances
+         * using the factory methods in {@link org.bson.codecs.configuration.CodecRegistries}.</p>
```
stIncMale, line 406: 👏 this is a nice simplification for users.

stIncMale (approved):

PR #1021, opened 2022-10-21
Ensure reactor context is passed to any child Mono/Flux

rozza:
JAVA-4788

### DIFF
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorPublisher.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorPublisher.java
@@ -112,6 +112,7 @@ abstract class BatchCursorPublisher<T> implements Publisher<T> {
                                     sink.success(results.get(0));
                                 }
                             })
+                            .contextWrite(sink.contextView())
                             .subscribe();
                 }));
     }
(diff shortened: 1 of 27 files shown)


rozza:
I'm not too happy with the testing I just couldn't think of an alternative way to get such large coverage

jyemin (commented):
Seems like there are other places that the tests aren't catching, e.g. `BatchCursorPublisher#first`.  I'm not sure how to find them all except by examining every place where we subscribe to a Publisher or a Mono

jyemin:
Skipping the pass-through assertion for the failing tests.  They fail because they execute no operations and so there is no context pass-through.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorPublisher.java
```diff
@@ -112,6 +112,7 @@ public Publisher<T> first() {
                                     sink.success(results.get(0));
                                 }
                             })
+                            .contextWrite(sink.contextView())
```
rozza, line 117: This is an oddity because it just propagates the context view to the inner publisher. 

But the context didn't break the chain as the outer publisher (the batch cursor publisher) had already been fixed. So the tests don't actually fail if this change is removed.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncMongoIterable.java
```diff
@@ -50,6 +51,9 @@ public MongoCursor<T> cursor() {
 
     @Override
     public T first() {
+        if (wrapped instanceof BatchCursorPublisher<?>) {
```
jyemin, line 54: Is this ever false?  If not, can just cast with no check

jyemin (approved):
LGTM

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncMongoIterable.java
rozza: Ack

PR #1023, opened 2022-10-24
Draft: expressions checkpoint 1

katcharov:
This draft PR partially implements behaviour that would be completed in future PRs. Its purpose is to make the proposed API available for discussion/commentary.

The best place to start would likely be the top of `class ExpressionsFunctionalTest`.

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+public interface DocumentExpression extends Expression {
+
+    BooleanExpression getFieldBoolean(String field);
+
+    NumberExpression getFieldNumber(String field);
+
+    IntegerExpression getFieldInteger(String field);
+
+    StringExpression getFieldString(String field);
+
+    DateExpression getFieldDate(String field);
+
+    DocumentExpression getFieldDocument(String field);
+
+    <T extends DocumentExpression> ArrayExpression<T> getFieldArray(String field);
+
+    // TODO and getFieldIntegerArrayArrayArray?
+    ArrayExpression<DocumentExpression> getFieldDocumentArray(String field);
+
+    <R extends DocumentExpression> R setField(String path, Expression exp);
+
+    DocumentExpression removeField(String path);
+
+    // TODO all document/array methods (maybe all) need to return T extends Type to allow for "schema" classes
+}
--- a/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
@@ -1,0 +1,306 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+import org.bson.BsonArray;
+import org.bson.BsonDocument;
+import org.bson.BsonString;
+import org.bson.BsonValue;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+public class MqlExpression<T extends Expression>
+        implements Expression, BooleanExpression, IntegerExpression, NumberExpression,
+        StringExpression, DateExpression, DocumentExpression, ArrayExpression<T> {
+
+    private final Function<CodecRegistry, BsonValue> fn;
+
+    protected MqlExpression(final Function<CodecRegistry, BsonValue> fn) {
+        this.fn = fn;
+    }
+
+    /**
+     * Exposes the evaluated BsonValue so that expressions may be used in
+     * aggregations. Non-public, as it is intended to be used only by the
+     * {@link ExpressionCodec}.
+     */
+    BsonValue toBsonValue(final CodecRegistry codecRegistry) {
+        return fn.apply(codecRegistry);
+    }
+
+    private Function<CodecRegistry, BsonValue> astDoc(final String name, final BsonDocument value) {
+        return (cr) -> new BsonDocument(name, value);
+    }
+
+    private Function<CodecRegistry, BsonValue> ast(final String name) {
+        return (cr) -> new BsonDocument(name, this.toBsonValue(cr));
+    }
+
+    private Function<CodecRegistry, BsonValue> ast(final String name, final Expression param1) {
+        return (cr) -> {
+            BsonArray value = new BsonArray();
+            value.add(this.toBsonValue(cr));
+            value.add(extractBsonValue(cr, param1));
+            return new BsonDocument(name, value);
+        };
+    }
+
+    private Function<CodecRegistry, BsonValue> ast(final String name, final Expression param1, final Expression param2) {
+        return (cr) -> {
+            BsonArray value = new BsonArray();
+            value.add(this.toBsonValue(cr));
+            value.add(extractBsonValue(cr, param1));
+            value.add(extractBsonValue(cr, param2));
+            return new BsonDocument(name, value);
+        };
+    }
+
+    /**
+     * Takes an expression and converts it to a BsonValue. MqlExpression will be
+     * the only implementation of Expression and all subclasses, so this will
+     * not mis-cast an expression as anything else.
+     */
+    @SuppressWarnings("rawtypes")
+    protected BsonValue extractBsonValue(final CodecRegistry cr, final Expression expression) {
+        return ((MqlExpression) expression).toBsonValue(cr);
+    }
+
+    /**
+     * Converts an MqlExpression to any subtype of Expression. Users must not
+     * extend Expression or its subtypes, so MqlExpression will implement any R.
+     */
+    @SuppressWarnings("unchecked")
+    <R extends Expression> R assertImplementsAllExpressions() {
+        return (R) this;
+    }
+
+    protected <R extends Expression> R newMqlExpression(final Function<CodecRegistry, BsonValue> ast) {
+        return new MqlExpression<>(ast).assertImplementsAllExpressions();
+    }
+
+    private <R extends Expression> R variable(final String variable) {
+        return newMqlExpression((cr) -> new BsonString(variable));
+    }
+
+    /** @see BooleanExpression */
+
+    @Override
+    public BooleanExpression not() {
+        return new MqlExpression<>(ast("$not"));
+    }
+
+    @Override
+    public BooleanExpression or(final BooleanExpression or) {
+        return new MqlExpression<>(ast("$or", or));
+    }
+
+    @Override
+    public BooleanExpression and(final BooleanExpression and) {
+        return new MqlExpression<>(ast("$and", and));
+    }
+
+    @Override
+    public <Q extends Expression> Q cond(final Q left, final Q right) {
+        return newMqlExpression(ast("$cond", left, right));
+    }
+
+    /** @see DocumentExpression */
+
+    private Function<CodecRegistry, BsonValue> getFieldInternal(final String fieldBoolean) {
+        // TODO these "getFieldX" names are all very long, and longer with composite paths
+        return (cr) -> astDoc("$getField", new BsonDocument().append("input", this.fn.apply(cr)).append("field", new BsonString(fieldBoolean))).apply(cr);
+    }
+
+    @Override
+    public BooleanExpression getFieldBoolean(final String field) {
+        return new MqlExpression<>(getFieldInternal(field));
+    }
+
+    @Override
+    public NumberExpression getFieldNumber(final String field) {
+        return new MqlExpression<>(getFieldInternal(field));
+    }
+
+    @Override
+    public IntegerExpression getFieldInteger(final String field) {
+        return new MqlExpression<>(getFieldInternal(field));
+    }
+
+    @Override
+    public StringExpression getFieldString(final String field) {
+        return new MqlExpression<>(getFieldInternal(field));
+    }
+
+    @Override
+    public DateExpression getFieldDate(final String field) {
+        return new MqlExpression<>(getFieldInternal(field));
+    }
+
+    @Override
+    public DocumentExpression getFieldDocument(final String field) {
+        return new MqlExpression<>(getFieldInternal(field));
+    }
+
+    @Override
+    public <R extends DocumentExpression> ArrayExpression<R> getFieldArray(final String field) {
+        return new MqlExpression<>(getFieldInternal(field));
+    }
+
+    @Override
+    public ArrayExpression<DocumentExpression> getFieldDocumentArray(final String arrObj) {
+        return new MqlExpression<>(getFieldInternal(arrObj));
+    }
+
+    @Override
+    public <R extends DocumentExpression> R setField(final String path, final Expression exp) {
+        return newMqlExpression((cr) -> astDoc("$setField", new BsonDocument()
+                .append("field", new BsonString(path))
+                .append("input", this.toBsonValue(cr))
+                .append("value", extractBsonValue(cr, exp))
+                .toBsonDocument(BsonDocument.class, cr)).apply(cr));
+    }
+
+    @Override
+    public DocumentExpression removeField(final String path) {
+        // do not expose the $$REMOVE variable
+        return this.setField(path, new MqlExpression<>((cr) -> new BsonString("$$REMOVE")));
+    }
+
+    /** @see Expression */
+
+    @Override
+    public <Q extends Expression, R extends Expression> R dot(final Function<Q, R> f) {
+        return f.apply(this.assertImplementsAllExpressions());
+    }
+
+    @Override
+    public <R extends Expression> R ifNull(final R ifNull) {
+        return new MqlExpression<>(ast("$ifNull", ifNull, Expressions.ofNull()))
+                .assertImplementsAllExpressions();
+    }
+
+    @Override
+    public BooleanExpression eq(final Expression eq) {
+        return new MqlExpression<>(ast("$eq", eq));
+    }
+
+    @Override
+    public BooleanExpression gt(final Expression gt) {
+        return new MqlExpression<>(ast("$gt", gt));
+    }
+
+    @Override
+    public <T0 extends Expression, R0 extends Expression, TV1 extends Expression> T0 let(
+            final TV1 var1, final BiFunction<T0, TV1, R0> ex) {
+        return newMqlExpression((cr) -> astDoc("$let", new BsonDocument()
+                .append("vars", new BsonDocument()
+                        .append("var1", extractBsonValue(cr, var1)))
+                .append("in", extractBsonValue(cr, ex.apply(
+                        this.assertImplementsAllExpressions(),
+                        variable("$$var1"))))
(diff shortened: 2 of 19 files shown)


driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+        assertExpression(
+                BsonDocument.parse("{'a': 1, 'z': ''}"),
+                a1.setField("z", a1.getFieldString("missing").toLower()));
+        /*
+        The behaviour of missing fields appears to be as follows, and equivalent to $$REMOVE:
+        propagates -- getField, cond branches...
+        false -- not, or, cond check...
+        0     -- sum...
+        ""    -- toLower...
+        null  -- multiply, add, subtract, year, filter, reduce, map, result within map...
```
katcharov, line 228: I'm not sure what should be done with missing fields. They are like an optional (a document maps from strings to optionals). Imagine we have an array of products, some of which are missing the field, and the user writes the following (the `2` is some sort of surcharge):

```
products.map(p => p.getField("USD").sum(2)).reduce(sum)
```
Possible intended results:

1. An exception because any one product is missing the field. Alternatively, a null value (in place of an error).
2. For those products that have a USD, halve the value, and then sum the results. The field is optional, the mult [is mapped](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#map-java.util.function.Function-) on the optional, and the reduction pre-filters-out all empty optionals.

The current MQL/JSON API is not consistent in the approach it takes. Why does `toLower` return the empty string, allowing for a proper concat reduction, while array operations do not yield the empty array? If the same problem arises for all reductive operators (concat, union, or, and, add, mult, ...), why do we have sum instead of asking the user to place a filter prior to add? In the above code, if there are only 10 products, and all are missing the field, the result might not be 0 but 20.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+                (1 + 1) - (1 - 1), // = 2
+                one.let(
+                        one.dot(decrement),
+                        (IntegerExpression e, IntegerExpression zero) -> e.add(one).subtract(zero)),
+                "{'$let': {'vars': {'var1': {'$subtract': [1, 1]}}, " //  ^^^^^^^^^ moved "down/right"
+                        + "'in': {'$subtract': [{'$add': [1, 1]}, '$$var1']}}}");
+
+        // TODO naming: dot, apply, map, chain?
+        // TODO cf: setField("three", three.dot(triple))
+        // TODO let must have different variable names. Perhaps a post-processor.
```
katcharov, line 384: Are there strong arguments for calling `dot` something else?

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+
+        // comparison:
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/gt/
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/gte/
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/lt/
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/lte/
+
+
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/cmp/ (89)
+        // TODO implement cmp as a three-branch cond?
```
katcharov, line 465: Thoughts?

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/ifNull/
+        // This is similar to orElse, but treats nulls as empty Optionals
+        // TODO the last position is auto-filled with null
+        // TODO this diverges from the MQL
+        // TODO just as with 1.add(2).add(3), multi-param ifNull should be chained: a.ifNull(b).ifNull(c)
```
katcharov, line 471: We should fill the last "mandatory" position with null. This should also be the behaviour of the server ifNull.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/sum/
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/multiply/
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/subtract/
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/divide/
+
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/max/
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/min/ (63)
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/round/
+
+        // TODO helper for bound constrain via max/min?
```
katcharov, line 503: There are a number of places where we can introduce helper methods. IIRC we have done this for full text search, by adding something like `gtlt`. But if we add these to the API, they should really be added to the server - but that is much more difficult to do.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -0,0 +1,43 @@
...
+    StringExpression getFieldString(String field);
+
+    DateExpression getFieldDate(String field);
+
+    DocumentExpression getFieldDocument(String field);
+
+    <T extends DocumentExpression> ArrayExpression<T> getFieldArray(String field);
+
+    // TODO and getFieldIntegerArrayArrayArray?
+    ArrayExpression<DocumentExpression> getFieldDocumentArray(String field);
```
katcharov, line 36: How far do we want to go with these convenience methods? Is there a better way to express these? We can do something like `getFieldArray(f, DocumentExpression.class)`, but even that is limited.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -0,0 +1,43 @@
...
+
+    <R extends DocumentExpression> R setField(String path, Expression exp);
+
+    DocumentExpression removeField(String path);
+
+    // TODO all document/array methods (maybe all) need to return T extends Type to allow for "schema" classes
```
katcharov, line 42: Is there any reason not to have all interface methods that return, for example, `BooleanExpression` instead return `<R extends BooleanExpression> R`? For DocumentExpression, this allows generated schema objects.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+                CURRENT.getFieldBoolean("isNull").not(),
+                "{'$not': {'$getField': {'input': '$$CURRENT', 'field': 'isNull'}}}");
+        assertExpression(
+                true,
+                CURRENT.getFieldBoolean("isInt0").not(),
+                "{'$not': {'$getField': {'input': '$$CURRENT', 'field': 'isInt0'}}}");
+        assertExpression(
+                false,
+                CURRENT.getFieldBoolean("isInt1").not(),
+                "{'$not': {'$getField': {'input': '$$CURRENT', 'field': 'isInt1'}}}");
```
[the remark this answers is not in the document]
stIncMale: > null ... and undefined values

Huh, so SQL uses ternary logic, but even that was not enough for MQL, and it uses four-valued logic? Fortunately, `undefined` is deprecated, though the [docs](https://www.mongodb.com/docs/rapid/reference/mongodb-extended-json-v1/#undefined-type) explain nothing regarding the reasons behind introducing it and how it could have been useful to users.

> $not evaluates as false the following: null, ...

Given that in SQL `not null` is `null` because [`null` means "unknown"](https://www.postgresql.org/docs/current/functions-logical.html#:~:text=SQL%20uses%20a%20three%2Dvalued%20logic%20system%20with%20true%2C%20false%2C%20and%20null%2C%20which%20represents%20%E2%80%9Cunknown%E2%80%9D.), evaluating `{ $not: [ null ] }` as `true` goes against what people familiar with SQL (virtually everyone) expect.

I agree that the `$not` behavior seems as bizarre as that of the JavaScript's `==`, which is bad.

> I do not think that our API should define this behaviour

I don't see how we can forbid `CURRENT.getFieldBoolean("isInt1").not()` given that nothing here tells us they the `"isInt1"` field is not of the `Boolean` BSON type. As for not defining the behavior, I don't think we should specify the server behavior at all, because this requires us to duplicate the server docs. The only thing we should document is the correspondence between our expressions and MQL expressions with links to the corresponding server docs pages. Based on the history of discussing similar questions, I expect others to disagree with me. But please remember that each time we duplicate server docs in our docs, we introduce one more piece of docs that may eventually become outdated / inconsistent with other docs (has happened multiple times), thus potentially misleading users and undermining confidence in our docs.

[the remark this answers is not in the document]
katcharov: > I don't see how we can forbid `CURRENT.getFieldBoolean("isInt1").not()`

We do not forbid this, but we intentionally specify that the behaviour is undefined (which diverges from the server docs). This is intended to encourage users of our API to write better code, just as writers of modern JavaScript are told to use `===` instead of `==`.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+        // for clarity:
+        "Convert2MethodRef"})
+public class ExpressionsFunctionalTest extends OperationTest {
+
+    @Test
+    public void booleanExpressionsTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/#boolean-expression-operators
+        // (Complete as of 6.0)
+
+        // literals
```
jyemin, line 73: Consider pros and cons of BooleanExpression.of(boolean value)

katcharov: I think `Expressions.of` would be good.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+        assertExpression(
+                true && false ? "abc" : "xyz",
+                tru.and(fal).cond(abc, xyz),
+                "{'$cond': [{'$and': [true, false]}, 'abc', 'xyz']}");
+        assertExpression(
+                true || false ? "abc" : "xyz",
+                tru.or(fal).cond(abc, xyz),
+                "{'$cond': [{'$or': [true, false]}, 'abc', 'xyz']}");
+
+        // fields
```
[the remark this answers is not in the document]
katcharov: Notes:
```
{x:8, some_field: [ {x:1,y}, {x:2,y} ]}
CURRENT.getArrayField("some_field").map(d -> CURRENT.getField("x"))

[null, null]
[1,2]
```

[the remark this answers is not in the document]
katcharov: `current()`

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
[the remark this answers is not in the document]
katcharov: Considering resolved. We do not allow users to use types "as if" they were other types (or at least, we do not define behaviour in such cases, and try to prevent this).

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+                array123.map(v -> v.multiply(2)),
+                // MQL:
+                "{'$map': {'input': [1, 2, 3], 'in': {'$multiply': ['$$this', 2]}}}");
+
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/reduce/
+        assertExpression(
+                Stream.of(1, 2, 3)
+                        .reduce(1, (a, b) -> a * b),
+                array123.reduce(Expressions.ofInteger(1), (a, b) -> a.multiply(b)),
+                // TODO: implicit monoid?
```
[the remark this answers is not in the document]
katcharov: This approach was not taken. It is not feasible to encode this level of info in the return types, which are simple (for example) integers.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
katcharov: The user is able to use `getInteger(name, default)`. They can also filter out values. We may, in the future, want to consider providing `filterToInteger` methods on arrays, but this is uncertain.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
katcharov: For this, we use the isIntegerOr (etc.) methods. There is no ifNull (though we use it internally).

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
katcharov: This was generally not done. If it is important enough to do, we or customers will realize this in the future.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: These have generally been limited. The above was removed.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
katcharov: We used passTo, which was much better.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
katcharov: Out of scope.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: We did not do this, except for Document, for that same reason.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -0,0 +1,306 @@
...
+
+    @Override
+    public <Q extends Expression> Q cond(final Q left, final Q right) {
+        return newMqlExpression(ast("$cond", left, right));
+    }
+
+    /** @see DocumentExpression */
+
+    private Function<CodecRegistry, BsonValue> getFieldInternal(final String fieldBoolean) {
+        // TODO these "getFieldX" names are all very long, and longer with composite paths
```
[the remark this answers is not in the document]
katcharov: We used getBoolean etc.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
katcharov: We did this.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/and/
+        assertExpression(true && false, tru.and(fal), "{'$and': [true, false]}");
+        assertExpression(false && true, fal.and(tru), "{'$and': [false, true]}");
+
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/not/
+        assertExpression(!true, tru.not(), "{'$not': true}");
+        assertExpression(!false, fal.not(), "{'$not': false}");
+
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/cond/
```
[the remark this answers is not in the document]
katcharov: This was not static.

At some point, I considered chaining cond off of the preceding expression, exp.cond(condition, left, right). This would have allowed the use of lambdas in the left and right positions, which made use of whatever exp was (this might have even used $let in the background). But this was too cumbersome - if that was needed, users can go with switchOn.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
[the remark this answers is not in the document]
katcharov: We used `current()`.

Another alternative is to provide something like:

```
collection
  .agg()
  .stage(cur -> matchOrSomeOtherStage( ... cur ...))
```

Though this was not used.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+        assertExpression(
+                true || false ? "abc" : "xyz",
+                tru.or(fal).cond(abc, xyz),
+                "{'$cond': [{'$or': [true, false]}, 'abc', 'xyz']}");
+
+        // fields
+        getCollectionHelper().insertDocuments("[{ 'isTrue': true}]");
+        assertExpression(
+                true,
+                CURRENT.getFieldBoolean("isTrue"),
```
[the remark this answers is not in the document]
katcharov: Used `getBoolean`.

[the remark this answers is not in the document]
katcharov: I no longer understand the two above comments.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+                CURRENT.getFieldBoolean("isInt1").not(),
+                "{'$not': {'$getField': {'input': '$$CURRENT', 'field': 'isInt1'}}}");
+    }
+
+    @Test
+    public void arrayExpressionsTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/#array-expression-operators
+        // (Incomplete)
+
+        // literals
```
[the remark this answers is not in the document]
katcharov: We now have:

`public static <T extends Expression> ArrayExpression<T> ofArray(final T... array)`

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+                        + "'input': '$$CURRENT', 'field': 'arrObj'}}, "
+                        + "'in': {'$getField': {'input': '$$this', 'field': 'v'}}}}");
+    }
+
+    @Test
+    public void documentExpressionsTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/#object-expression-operators
+        // (Complete as of 6.0)
+
+        // literals
```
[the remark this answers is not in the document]
katcharov: We now have `public static DocumentExpression of(final Bson document) {`.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+        DocumentExpression a1 = Expressions.ofDocument(BsonDocument.parse("{a: 1}"));
```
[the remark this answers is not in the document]
katcharov: We now have `public static DocumentExpression of(final Bson document) {`

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+        d.setA(d.a.setY(d.a.y.multiply(2)));
+        // TODO: assertion
+
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/arrayToObject/ (48)
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/objectToArray/ (23)
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/mergeObjects/
+
+    }
+
+    @Test
```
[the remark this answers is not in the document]
katcharov: There should be no need to review further. I believe everything was moved up. Everything relevant to the initial release should have been moved forward to the various PRs.

PR #1024, opened 2022-10-25
JAVA-4789: Make bson.record.codec module optional for OSGi

Joscorbe:
Currently the Java Driver can't be used in OSGi containers using a Java version lower than 17, which according to the compatibility matrix should not happen, as it should be compatible from Java 8+.

The problem is that there is a dependency of bson.codecs.record, which requires Java17+ according to its MANIFEST.MF:

Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"

This change will make the Import-Package attribute for bson.codecs.record optional, meaning when trying to use it in a OSGi container it won't fail because the dependency can't be satisfied in lower versions than 17.

After building with this change applied, that line in the MANIFEST.MF file will look like this:

org.bson.codecs.record;resolution:=optional;version="[4.7,5)"

We can see above this line that the same approach is used for jdk.net.

This change will still require Java 17+ to build the driver, but allows its use in OSGi containers using lower versions.

[JAVA-4789](https://jira.mongodb.org/browse/JAVA-4789)

### DIFF
--- a/driver-core/build.gradle
+++ b/driver-core/build.gradle
@@ -87,6 +87,7 @@ afterEvaluate {
             'jnr.unixsocket.*;resolution:=optional',
             'com.mongodb.crypt.capi.*;resolution:=optional',
             'jdk.net.*;resolution:=optional', // Used by SocketStreamHelper & depends on JDK version
+            'org.bson.codecs.record.*;resolution:=optional', // Depends on JDK version
             '*' // import all that is not excluded or modified before
     ].join(',')
 }


Joscorbe:
Versions 4.6.x of the driver are also affected by this problem.

jyemin:
Backported to 4.7.x branch: https://github.com/mongodb/mongo-java-driver/commit/2f2e579e99a085d2e5a4f3812ab58fb303524406.

@Joscorbe our policy is to backport fixes to the latest release branch, so this fix will be included in 4.7.3 release (which will be released within a week) as well as the next minor release, 4.8.0.

PR #1025, opened 2022-10-25
Implement boolean expressions

katcharov:
JAVA-4779

This implements not/and/or/cond for boolean expressions.

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -16,6 +16,12 @@
 
 package com.mongodb.client.model.expressions;
 
+/**
+ * Expresses an array value. An array value is a finite, ordered collection of
+ * elements of a certain type.
+ *
+ * @param <T> the type of the elements in the array
+ */
 public interface ArrayExpression<T extends Expression> extends Expression {
 
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+public interface ArrayExpression<T extends Expression> extends Expression {
+
+}
--- a/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
@@ -19,39 +19,39 @@ package com.mongodb.client.model.expressions;
 public interface BooleanExpression extends Expression {
 
     /**
-     * Evaluates this expression and returns logical true if the result is
-     * logical false. Otherwise, false.
+     * Returns logical true if this expression evaluates to logical false.
+     * Returns logical false if this expression evaluates to logical true.
      *
      * @return True if false; false if true.
      */
     BooleanExpression not();
 
     /**
-     * Returns logical true if this or the other expression evaluate to logical
-     * true. Otherwise, false.
+     * Returns logical true if this or the other expression evaluates to logical
+     * true. Returns logical false if both evaluate to logical false.
      *
      * @param or the other boolean expression.
-     * @return True if either true, false otherwise.
+     * @return True if either true, false if both false.
      */
     BooleanExpression or(BooleanExpression or);
 
     /**
-     * Evaluates this and the other expression and returns logical true if both
-     * this and the other expression evaluate to logical true. Otherwise, false.
+     * Returns logical true if both this and the other expression evaluate to
+     * logical true. Returns logical false if either evaluate to logical false.
      *
      * @param and the other boolean expression.
-     * @return true if both true, false otherwise.
+     * @return true if both true, false if either false.
      */
     BooleanExpression and(BooleanExpression and);
 
     /**
-     * Evaluates this expression. If true, returns the result of the evaluated
-     * left branch expression. Otherwise, returns the result of the evaluated
-     * right branch expression.
+     * If this expression evaluates to logical true, returns the result of the
+     * evaluated left branch expression. If this evaluates to logical false,
+     * returns the result of the evaluated right branch expression.
      *
-     * @param left the left branch expression, evaluated if this is true.
-     * @param right the right branch expression, evaluated if this is false.
-     * @return Left if true, right if false.
+     * @param left the left branch expression
+     * @param right the right branch expression
+     * @return left if true, right if false.
      * @param <T> The type of the resulting expression.
      */
     <T extends Expression> T cond(T left, T right);
--- a/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
@@ -17,7 +17,7 @@
 package com.mongodb.client.model.expressions;
 
 /**
- * Expresses a logical boolean value.
+ * Expresses a boolean value.
  */
 public interface BooleanExpression extends Expression {
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+public interface BooleanExpression extends Expression {
+
+    /**
+     * Evaluates this expression and returns logical true if the result is
+     * logical false. Otherwise, false.
+     *
+     * <p>If this expression does not evaluate to logical true or false, then
+     * the result is undefined.
+     *
+     * @return True if false; false if true.
+     */
+    BooleanExpression not();
+
+    /**
+     * Returns logical true if this or the other expression evaluate to logical
+     * true. Otherwise, false.
+     *
+     * <p>If this expression does not evaluate to logical true or false, then
+     * the result is undefined. The order of expression evaluation, and whether
+     * both are evaluated, is undefined.
+     *
+     * @param or the other boolean expression.
+     * @return True if either true, false otherwise.
+     */
+    BooleanExpression or(BooleanExpression or);
+
+    /**
+     * Evaluates this and the other expression and returns logical true if both
+     * this and the other expression evaluate to logical true. Otherwise, false.
+     *
+     * <p>If this expression does not evaluate to logical true or false, then
+     * the result is undefined. The order of expression evaluation is undefined.
+     *
+     * @param and the other boolean expression.
+     * @return true if both true, false otherwise.
+     */
+    BooleanExpression and(BooleanExpression and);
+
+    /**
+     * Evaluates this expression. If true, returns the result of the evaluated
+     * left branch expression. Otherwise, returns the result of the evaluated
+     * right branch expression.
+     *
+     * <p>If this expression does not evaluate to logical true or false, then
+     * the result is undefined. The order of expression evaluation, and whether
+     * both are evaluated, is undefined.
+     *
+     * @param left the left branch expression, evaluated if this is true.
+     * @param right the right branch expression, evaluated if this is false.
+     * @return Left if true, right if false.
+     * @param <T> The type of the resulting expression.
+     */
+    <T extends Expression> T cond(T left, T right);
+}
--- a/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
@@ -16,6 +16,9 @@
 
 package com.mongodb.client.model.expressions;
 
+/**
+ * Expresses a logical boolean value.
+ */
 public interface BooleanExpression extends Expression {
 
     /**
--- a/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
@@ -22,9 +22,6 @@ public interface BooleanExpression extends Expression {
      * Evaluates this expression and returns logical true if the result is
      * logical false. Otherwise, false.
      *
-     * <p>If this expression does not evaluate to logical true or false, then
-     * the result is undefined.
-     *
      * @return True if false; false if true.
      */
     BooleanExpression not();
@@ -32,10 +29,6 @@ public interface BooleanExpression extends Expression {
     /**
      * Returns logical true if this or the other expression evaluate to logical
      * true. Otherwise, false.
-     *
-     * <p>If this expression does not evaluate to logical true or false, then
-     * the result is undefined. The order of expression evaluation, and whether
-     * both are evaluated, is undefined.
      *
      * @param or the other boolean expression.
      * @return True if either true, false otherwise.
@@ -46,9 +39,6 @@ public interface BooleanExpression extends Expression {
      * Evaluates this and the other expression and returns logical true if both
      * this and the other expression evaluate to logical true. Otherwise, false.
      *
-     * <p>If this expression does not evaluate to logical true or false, then
-     * the result is undefined. The order of expression evaluation is undefined.
-     *
      * @param and the other boolean expression.
      * @return true if both true, false otherwise.
      */
@@ -58,10 +48,6 @@ public interface BooleanExpression extends Expression {
      * Evaluates this expression. If true, returns the result of the evaluated
      * left branch expression. Otherwise, returns the result of the evaluated
      * right branch expression.
-     *
-     * <p>If this expression does not evaluate to logical true or false, then
-     * the result is undefined. The order of expression evaluation, and whether
-     * both are evaluated, is undefined.
      *
      * @param left the left branch expression, evaluated if this is true.
      * @param right the right branch expression, evaluated if this is false.
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+public interface DocumentExpression extends Expression {
+
+}
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -16,6 +16,11 @@
 
 package com.mongodb.client.model.expressions;
 
+/**
+ * Expresses a document value. A document is an ordered set of fields, which may
+ * be of an arbitrary string value, that each map to a value of any other
+ * expression type.
+ */
 public interface DocumentExpression extends Expression {
 
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -17,9 +17,8 @@
 package com.mongodb.client.model.expressions;
 
 /**
- * Expresses a document value. A document is an ordered set of fields, which may
- * be of an arbitrary string value, that each map to a value of any other
- * expression type.
+ * Expresses a document value. A document is an ordered set of fields, where the
+ * key is a string value, mapping to a value of any other expression type.
  */
 public interface DocumentExpression extends Expression {
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -16,10 +16,31 @@
 
 package com.mongodb.client.model.expressions;
 
-// TODO: notice to users that this should be treated as sealed.
+import com.mongodb.annotations.Evolving;
+
 /**
+ * Expressions express values that may be represented in, or computations that
+ * may be performed within, a MongoDB server. Each expression resolves to some
+ * value, much like any Java expression ultimately resolves to some value, and
+ * expressions may be thought of as boxed values, except that resolution might
+ * potentially happen on a server, rather than locally.
  *
+ * <p>Users should treat these interfaces as sealed, and must not implement any
+ * expression interfaces.
+ *
+ * <p>Expressions are typed. It is possible to execute expressions against data
+ * that is of the wrong type, such as by applying the "not" boolean expression
+ * to a document field that is an integer. This API does not define the output
+ * in such cases. Where data might have divergent types (for example, if there
+ * is some field that could be null, or missing, or a boolean, or an int) then
+ * users should make it explicit what should be done in each case. Likewise,
+ * unless otherwise specified, this API does not define the order of evaluation
+ * for all arguments, as well as whether all arguments to some expression shall
+ * be evaluated.
+ *
+ * @see Expressions
  */
+@Evolving
 public interface Expression {
 
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+// TODO: notice to users that this should be treated as sealed.
+/**
+ *
+ */
+public interface Expression {
+
+}
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -19,24 +19,25 @@ package com.mongodb.client.model.expressions;
 import com.mongodb.annotations.Evolving;
 
 /**
- * Expressions express values that may be represented in, or computations that
- * may be performed within, a MongoDB server. Each expression resolves to some
- * value, much like any Java expression ultimately resolves to some value, and
- * expressions may be thought of as boxed values, except that resolution might
- * potentially happen on a server, rather than locally.
+ * Expressions express values that may be represented in (or computations that
+ * may be performed within) a MongoDB server. Each expression evaluates to some
+ * value, much like any Java expression evaluates to some value. Expressions may
+ * be thought of as boxed values. Evaluation of an expression will usually occur
+ * on a MongoDB server.
  *
  * <p>Users should treat these interfaces as sealed, and must not implement any
  * expression interfaces.
  *
  * <p>Expressions are typed. It is possible to execute expressions against data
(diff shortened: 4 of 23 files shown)


driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+// TODO: notice to users that this should be treated as sealed.
+/**
+ *
+ */
```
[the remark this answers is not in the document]
jyemin: I think we use `@Evolving` annotation, plus some prose in the Javadoc.

driver-core/src/main/com/mongodb/client/model/expressions/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * API for MQL expressions.
+ *
+ * <p>All undefined behaviour may become defined in the future.
+ */
+@NonNullApi
+@Beta(Beta.Reason.CLIENT) // TODO beta until semi-functional
```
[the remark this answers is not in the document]
jyemin: If we mark the package as Beta, where would that show up in the Javadoc?

driver-core/src/test/functional/com/mongodb/client/model/expressions/ExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    private void assertEval(final Object expected, final Expression toEvaluate) {
+        Bson addFieldsStage = addFields(new Field<>("val", toEvaluate));
+        List<Bson> stages = new ArrayList<>();
+        stages.add(addFieldsStage);
+        List<BsonDocument> results;
+        if (getCollectionHelper().count() == 0) {
+            // TODO: replace documents stage for older versions of the server?
+            assumeTrue(serverVersionAtLeast(5, 1));
```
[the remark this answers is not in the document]
jyemin: We test on all supported server versions, but tests for particular expressions can be skipped/disabled/ignored based on detected server version.  Consider splitting this test into many smaller tests (one per expression).  This has the advantage that 

1. We can skip a test for any given expression based on server version or any other condition
2. It's easier to see exactly which expressions are failing if we regress.

driver-core/src/main/com/mongodb/client/model/expressions/ExpressionCodec.java
```diff
@@ -0,0 +1,70 @@
...
+import org.bson.codecs.BsonValueCodecProvider;
+import org.bson.codecs.Codec;
+import org.bson.codecs.DecoderContext;
+import org.bson.codecs.EncoderContext;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
+
+@Immutable
+public final class ExpressionCodec implements Codec<Expression> {
```
jyemin, line 32: No need for this to be public

jyemin, line 32: This should be:

     class ExpressionCodec<T extends Expression> implements Codec<T>

and the constructor should have a `Class<T> clazz` parameter, which can then be returned from `getEncoderClass`

jyemin: Or if we can only encode MqlExpression instances, then, it should be:

    class ExpressionCodec<T extends MqlExpression> implements Codec<T>

and we should have an `MqlExpresionCodecProvider`

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+public interface ArrayExpression<T extends Expression> extends Expression {
```
jyemin, line 19: Do these empty interfaces need to be part of this PR?

driver-core/src/main/com/mongodb/client/model/expressions/ExpressionCodecProvider.java
```diff
@@ -0,0 +1,35 @@
...
+import org.bson.codecs.configuration.CodecRegistry;
+
+@Immutable
+public class ExpressionCodecProvider implements CodecProvider {
+    @Override
+    @SuppressWarnings("unchecked")
+    public <T> Codec<T> get(final Class<T> clazz, final CodecRegistry registry) {
+        // TODO: should this be MqlExpression.class?
+        if (Expression.class.isAssignableFrom(clazz)) {
+            return (Codec<T>) new ExpressionCodec(registry);
```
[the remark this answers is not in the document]
jyemin: I commented on this in the ExpressionCodec.  I think this should be `MqlExpression` if that's what we're able to encode.  No point in saying we can encode `Expression` if it's just going to throw a `ClassCastException` when you try.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -0,0 +1,72 @@
...
+    private Expressions() {}
+
+    /**
+     * Returns an expression having the same logical boolean value as the
+     * provided boolean primitive.
+     *
+     * @param of the boolean primitive
+     * @return the boolean expression
+     */
+    public static BooleanExpression ofBoolean(final boolean of) {
```
jyemin, line 34: Could/should these be static methods on the `BooleanExpression` interface instead?  If so, a single

    static Boolean Expression of(boolean b) { ...}

would suffice.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Logical false.
+     *
+     * @return logical false.
+     */
+    public static BooleanExpression ofFalse() {
+        return ofBoolean(false);
+    }
+
+    public static IntegerExpression ofInteger(final int of) {
```
jyemin, line 57: Can we remove these two methods from this PR?

jyemin (changes requested):
Took a quick first pass.  Looking good.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: They are used for testing cond.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
katcharov: I want it to be clear that `MqlExpression` intends to implement all of these.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: If we use `of` for everything, there would be conflicts on static import, so I don't think we should.

If we put these on different interfaces, the user cannot import `Expressions.*`, or scan through all of these "literals" from one file. Putting `ofBoolean` on BoolEx might make it a bit easier to find. I think perhaps we should come back to this and see how it looks when more expressions are implemented.

driver-core/src/main/com/mongodb/client/model/expressions/ExpressionCodecProvider.java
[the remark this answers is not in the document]
katcharov: > I commented on this in the ExpressionCodec. I think this should be MqlExpression if that's what we're able to encode. No point in saying we can encode Expression if it's just going to throw a ClassCastException when you try @jyemin 

I partially implemented this and then balked. We have `class MqlExpression<T extends Expression>`, but the T is just a "fake" <Expression> for most Expressions (it's there for arrays, which are the only container expression type). I get "raw use" for `MqlExpressionCodec implements Codec<MqlExpression>`, but it seems odd that there should be a T in there for arbitrary MqlExpressions.

I'm not very sure, but I think perhaps we really are trying to encode Expressions into BSON. It is an implementation detail that the only class that implements Expression is MqlExpression. (And consequently, there would never be a ClassCastException). If we create other implementations, then it seems we should handle them as part of this codec.

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
```diff
@@ -0,0 +1,72 @@
...
+package com.mongodb.client.model.expressions;
+
+public interface BooleanExpression extends Expression {
+
+    /**
+     * Evaluates this expression and returns logical true if the result is
+     * logical false. Otherwise, false.
+     *
+     * <p>If this expression does not evaluate to logical true or false, then
+     * the result is undefined.
```
katcharov, line 26: not defined, not `undefined`.

driver-core/src/main/com/mongodb/client/model/expressions/ExpressionCodec.java
katcharov: How would `ExpressionCodec<T extends Expression>` come into play? That is, what would have to come after `BooleanExpression b = ???` for this to be necessary?

driver-core/src/main/com/mongodb/client/model/expressions/ExpressionCodecProvider.java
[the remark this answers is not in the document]
jyemin: > I'm not very sure, but I think perhaps we really are trying to encode Expressions into BSON

Perhaps, but it's not possible since an `Expression` isn't rich enough as a type to be able to build a generic Codec for.  

A CodecProvider should only provide a codec for a type that it is able to encode.  In this case, that type is MqlExpression.  So I think this should just be `MqlExpressionCodecProvider` and it should return instances of `MqlExpressionCodec, which should implement `Codec<MqlExpression>`.

If we end up adding more Expression implementations, we can add a new CodecProvider for those implementations.

driver-core/src/main/com/mongodb/client/model/expressions/ExpressionCodec.java
jyemin: I'm wrong.  This can just be:

    class MqlExpressionCodec implements Codec<MqlExpression>

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
jyemin: Discussed on the draft PR.  Resolving

driver-core/src/main/com/mongodb/client/model/expressions/package-info.java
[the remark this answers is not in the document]
katcharov: Removed Beta

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
katcharov: Moved all of these to a generic explanation at the top of Expression

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -0,0 +1,78 @@
...
+    private Expressions() {}
+
+    /**
+     * Returns an expression having the same logical boolean value as the
+     * provided boolean primitive.
+     *
+     * @param of the boolean primitive
+     * @return the boolean expression
+     */
+    public static BooleanExpression ofBoolean(final boolean of) {
```
jyemin, line 34: Did we decide in the draft PR to just call this `of`?  And if so, can we get rid of `ofTrue` and `ofFalse`?

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -0,0 +1,122 @@
...
+
+import org.bson.BsonArray;
+import org.bson.BsonDocument;
+import org.bson.BsonString;
+import org.bson.BsonValue;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import java.util.function.Function;
+
+public class MqlExpression<T extends Expression>
```
jyemin, line 27: Can this be package private and final?

driver-core/src/test/functional/com/mongodb/client/model/expressions/BooleanExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+package com.mongodb.client.model.expressions;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@SuppressWarnings({"PointlessBooleanExpression", "ConstantConditions", "ConstantConditionalExpression"})
+class BooleanExpressionsFunctionalTest extends AbstractExpressionsFunctionalTest {
```
jyemin, line 24: We should have unit tests that assert the generated MQL along with the integration tests that assert that the MQL does what we expect it to do.

jyemin (changes requested):

driver-core/src/test/functional/com/mongodb/client/model/expressions/BooleanExpressionsFunctionalTest.java
katcharov:
```
assertExpression(true || false, tru.or(fal), "{'$or': [true, false]}");
```

`assertExpression`  asserts that each expression evaluates to the provided result on the left, and that the expression generates the appropriate MQL on the right.

jyemin: Oh, I missed that.  Thanks.

driver-core/src/test/functional/com/mongodb/client/model/expressions/AbstractExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,107 @@
...
+import static com.mongodb.client.model.Aggregates.addFields;
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * This test file groups expressions of each type under one test method. Each of
+ * these methods begins by showing how to express literals of the relevant type.
+ * The ensuing assertions then show how to express the computation in Java (when
+ * reasonable), then using the API under test, and then in MQL.
+ */
```
stIncMale, line 49: Was this doc comment intended to a different file? `AbstractExpressionsFunctionalTest.java` does not have any test methods.

driver-core/src/test/functional/com/mongodb/client/model/expressions/AbstractExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @AfterEach
+    public void tearDown() {
+        getCollectionHelper().drop();
+    }
+
+    protected void assertExpression(final Object expectedResult, final Expression expression, final String expectedMql) {
+        assertEval(expectedResult, expression);
+
+        BsonValue expressionValue = ((MqlExpression) expression).toBsonValue(fromProviders(new BsonValueCodecProvider()));
```
stIncMale, line 65: [nit]

Using `MqlExpression<?>` instead of `MqlExpression` removes the "raw use" IDE warning.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
stIncMale, line 19: `./gradlew clean check jar testClasses docs -x test -x integrationTest` emits multiple warnings about doc comments missing on the public API program elements added in this PR (there are no such warnings for the existing code). We need to add doc comments to get rid of the warnings.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
jyemin: +1 and I don't know why Checkstyle doesn't catch a lot of these.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -0,0 +1,122 @@
...
+import org.bson.BsonValue;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import java.util.function.Function;
+
+final class MqlExpression<T extends Expression>
+        implements Expression, BooleanExpression, IntegerExpression, NumberExpression,
+        StringExpression, DateExpression, DocumentExpression, ArrayExpression<T> {
+
+    private final Function<CodecRegistry, BsonValue> fn;
```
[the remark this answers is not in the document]
katcharov: The `ast` methods are intended to have the shape of an AST, but in MqlExpression, the AST is immediately turned into this function, rather than a proper AST.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -0,0 +1,122 @@
...
+        return (R) this;
+    }
+
+    protected <R extends Expression> R newMqlExpression(final Function<CodecRegistry, BsonValue> ast) {
+        return new MqlExpression<>(ast).assertImplementsAllExpressions();
+    }
+
+    private <R extends Expression> R variable(final String variable) {
+        return newMqlExpression((cr) -> new BsonString(variable));
+    }
```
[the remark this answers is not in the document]
katcharov: I will leave both, they will be used in the next PR.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpressionCodecProvider.java
```diff
@@ -0,0 +1,34 @@
...
+
+package com.mongodb.client.model.expressions;
+
+import com.mongodb.annotations.Immutable;
+import org.bson.codecs.Codec;
+import org.bson.codecs.configuration.CodecProvider;
+import org.bson.codecs.configuration.CodecRegistry;
+
+@Immutable
+public class MqlExpressionCodecProvider implements CodecProvider {
```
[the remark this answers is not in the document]
katcharov: MqlExpressionCodec is in the same package as MqlExpression to ensure package-private on `toBsonValue`, which I think is more important to protect than MqlExpressionCodec. So I do not think this can be non-public or in another package.

driver-core/src/test/functional/com/mongodb/client/model/OperationTest.java
```diff
@@ -73,11 +76,32 @@ MongoNamespace getNamespace() {
...
 
     public static Document toDocument(final BsonDocument bsonDocument) {
         return getDefaultCodecRegistry().get(Document.class).decode(bsonDocument.asBsonReader(), DecoderContext.builder().build());
     }
+    public static BsonDocument toBsonDocument(final BsonDocument bsonDocument) {
+        return getDefaultCodecRegistry().get(BsonDocument.class).decode(bsonDocument.asBsonReader(), DecoderContext.builder().build());
+    }
+
+
+    protected List<Bson> assertPipeline(final String stageAsString, final Bson stage) {
```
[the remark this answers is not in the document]
katcharov: I suspect merge conflicts. Will leave for now.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -16,6 +16,12 @@
 
 package com.mongodb.client.model.expressions;
 
+/**
+ * Expresses an array value. An array value is a finite, ordered collection of
+ * elements of a certain type.
+ *
+ * @param <T> the type of the elements in the array
+ */
```
stIncMale, line 24: Since expressions are evaluated on the server, and the server can operate only with BSON types, it does not seem possible to abstract expression types of BSON types: if you do that, then a user can

- neither express a value stored in MongoDB via the expression API as it's unspecified what type should be used
- nor store an expression value to MongoDB as it is unclear how it will be represented there (for example, with the current lack of specification, you may be expressing all values as BSON `String` values that contain extended JSON).

So it appears to me that if an expression type has a BSON counterpart (which is true for everything but the `Map` type you proposed), it should be documented as such.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,6 +16,11 @@
 
 package com.mongodb.client.model.expressions;
 
+/**
+ * Expresses a document value. A document is an ordered set of fields, which may
+ * be of an arbitrary string value, that each map to a value of any other
```
stIncMale, line 21: > fields, which may be of an arbitrary string value

Field values are not necessary strings.

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
```diff
@@ -16,6 +16,9 @@
 
 package com.mongodb.client.model.expressions;
 
+/**
+ * Expresses a numeric integer value.
```
stIncMale, line 20: Isn't _"numeric integer"_, _"logical boolean"_ akin to electrical electronics?

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpressionCodecProvider.java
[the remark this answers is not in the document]
stIncMale: If we move `MqlExpression`, `MqlExpressionCodec `, `MqlExpressionCodecProvider` to `com.mongodb.internal.client.model.expressions`, then they will all be excluded from the public API by means of being in the `internal` package, and `toBsonValue` will stay package-access (though the latter will not be of the essence anymore).

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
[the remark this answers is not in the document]
stIncMale: I have two concerns:

- How reviewers are supposed to differentiate actually dead code from dead code that is expected to become live in conditionally on the future PR without wasting their time?
- What happens with this code if as a result of the preview process the code in the new PR gets changed, and this method does not become live? This code will not be seen as new in the next PR, so if the next PR stops using the method, it will be more likely that the dead method stays.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: I mean that the fields (names) are an arbitrary string value (though I am not sure about this, for example how is whitespace handled).

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpressionCodecProvider.java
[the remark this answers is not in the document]
katcharov: `Expressions` might end up using of package-private functionality on the aforementioned. Once this is merged, I will create a draft PR to merge `expressions` into `master`, and I think we should move comments like this there.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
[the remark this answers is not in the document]
stIncMale: Discussed in Zoom:

- If dead code is placed in a PR intentionally, knowing that the next PR will use it (for example, if there will be two next PRs crated concurrently, and both need the code), then the author leaves a comment about it to reviewers. If the code stays dead as a result of review changes in the next PRs, then the situation is indistinguishable from a situation when a PR changes existing code such that it makes some other existing code dead, and should be addressed in the same way.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpressionCodecProvider.java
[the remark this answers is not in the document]
stIncMale: Discussed in Zoom, and realized that `MqlExpressionCodecProvider` has to be part of public API, because if a user wants to specify a custom codec provider that does not have all the codecs from `MongoClient.getDefaultCodecRegistry`, but also wants to be able to use expressions via the existing methods that accept `Object`(a.k.a. `TExpression`), then a users needs access to `MqlExpressionCodecProvider`<sup>1</sup>.

---

<sup>1</sup> Technically, a user may use `MongoClient.getDefaultCodecRegistry` in its custom codec provider in such a way that it is only used for encoding expressions, but it's a very unfriendly approach. Moreover, if `MqlExpressionCodecProvider` is not exposed, a user don't even have a reliable way of knowing that expressions are encoded via the codecs functionality and that `MongoClient.getDefaultCodecRegistry` contains a non-public codec for them.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
stIncMale: Discussed in Zoom, decided that `MqlExpressionCodecProvider` is the right place to specify the relation between the abstract types whose values can be represented by the expression API and the types of a specific execution context, e.g., MongoDB. Fortunately, we also [decided](https://github.com/mongodb/mongo-java-driver/pull/1025#discussion_r1009936352) that `MqlExpressionCodecProvider` has to be public API.

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
stIncMale: Discussed in Zoom, decided to call them Platonic integer and Platonic boolean :trollface:

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpressionCodecProvider.java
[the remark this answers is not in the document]
stIncMale: Also, if we are making `MqlExpressionCodecProvider` part of the public API, we need to document that it is included in `MongoClientSettings.getDefaultCodecRegistry`.

stIncMale (approved):

driver-core/src/test/functional/com/mongodb/client/model/expressions/AbstractExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,101 @@
...
+    public void setUp() {
+        getCollectionHelper().drop();
+    }
+
+    @AfterEach
+    public void tearDown() {
+        getCollectionHelper().drop();
+    }
+
+    protected void assertExpression(final Object expectedResult, final Expression expression, final String expectedMql) {
```
rozza, line 56: Like it :)

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -0,0 +1,121 @@
...
+        };
+    }
+
+    /**
+     * Takes an expression and converts it to a BsonValue. MqlExpression will be
+     * the only implementation of Expression and all subclasses, so this will
+     * not mis-cast an expression as anything else.
+     */
+    private static BsonValue extractBsonValue(final CodecRegistry cr, final Expression expression) {
+        return ((MqlExpression<?>) expression).toBsonValue(cr);
```
rozza, line 79: I'm surprised this doesn't issue an unchecked warning, I don't get why it doesn't?!

rozza (approved):
LGTM - I have one question more for my learning than anything else

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
stIncMale: The [`unchecked`](https://docs.oracle.com/en/java/javase/19/docs/specs/man/javac.html#:~:text=Gives%20more%20detail%20for%20unchecked%20conversion%20warnings%20that%20are%20mandated%20by%20the%20Java%20Language%20Specification) `-Xlint` `javac` key deals only with the unchecked narrowing conversions. The JLS states that [_"A narrowing reference conversion from a type `S` to a parameterized class or interface type `T` is unchecked, unless at least one of the following is true: All of the type arguments of `T` are unbounded wildcards. ..."_](https://docs.oracle.com/javase/specs/jls/se19/html/jls-5.html#jls-5.1.6.2).

It's easy to see why the above is fine: a paremetrized type with all type arguments being unbounded wildcards is a type for which the program has no expectation on the type arguments, so there is nothing that `javac` was required to check but was unable to.

rozza: Thanks :)

stIncMale: While I was away (if you know what I [mean](https://www.google.com/search?q=toilet&oq=toilet&aqs=chrome..69i57.810j0j7&sourceid=chrome&ie=UTF-8)), I realized that the argument _"the program has no expectation on the type arguments"_ is applicable to raw types just as it is applicable to types parametrized with `?`. However, `javac` warns us about raw `MqlExpression` and does not warn when we cast to `MqlExpression<?>`.

While I can't give a formal explanation, I have a hand-weaving explanation via an example. However, I will have to use `List<String>` instead of `Expression` and `ArrayList<?>` instead of `MqlExpression<?>` because there is no practical problem with `MqlExpression` in the code we are discussing, but `javac` cannot know that, and warns us by following a simple algorithm.

The problem with rawtypes (`ArrayList`), which we don't have with `?`-parametrized types (`ArrayList<?>`) and therefore, there is no need for a compiler to warn us, is demonstrated with the following example:

```java
List<String> strings = new ArrayList<>();
ArrayList objects = (ArrayList) strings; // javac warning: [rawtypes] found raw type: ArrayList
// pollute the heap
objects.add(42); // javac warning: [unchecked] unchecked call to add(E) as a member of the raw type ArrayList
// enjoy the consequences
strings.get(0); // java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String
```

the same example won't compile if we use `ArrayList<?>`:

```java
List<String> strings = new ArrayList<>();
ArrayList<?> objects = (ArrayList<?>) strings;
// try to pollute the heap
objects.add(42); // javac error: incompatible types: int cannot be converted to CAP#1 ... where ... CAP#1 extends Object from capture of ?
// there are no consequences as the code does not compile
strings.get(0);
```

So not only a `?`-paremetrized type express no expectation on the type arguments, but also makes the values of the type kinda sorta "unmodifiable" (I have no idea how to formally express this property), which is how it is different from raw types.

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
```diff
@@ -0,0 +1,61 @@
...
+
+package com.mongodb.client.model.expressions;
+
+/**
+ * Expresses a boolean value.
+ */
+public interface BooleanExpression extends Expression {
+
+    /**
+     * Returns logical true if this expression evaluates to logical false.
```
jyemin, line 25: I think the Javadoc should be written in terms of expressions rather than values.  So in this case something like:

`Returns an expression representing the logical negation of this {@code BooleanExpression}.`

jyemin (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
katcharov: This is turning out a bit more difficult to formulate at this stage than I thought it would. I'd like to defer docs until a later PR.

jyemin: Acknowledged, but please open a JAVA ticket for it.

jyemin (approved):

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
katcharov: https://jira.mongodb.org/browse/JAVA-4799

PR #1026, opened 2022-10-27
Clean up internal packages

jyemin:
* Ensure that every public type has Javadoc indicating it's not part of the public API
* Remove boilerplate Javadoc
* Remove some unused methods

I don't expect anyone to look at all the diffs, but just try to get a flavor for what I removed and what I didn't


stIncMale (approved):

stIncMale:
It's especially cool that some code was also deleted :)

PR #1027, opened 2022-10-28
Add AsyncOperations to match SyncOperations

jyemin:
JAVA-4795

One more change to clean up use of internal `operation` package

The benefit is that, like with `SyncOperation`, `AsyncOperations` doesn't expose the concrete class of any operation, and just deals with the interfaces like `AsyncReadOperation` and `AsyncWriteOperation`.

This gets us closer to a state where we can make all concrete Operation classes package private.  The main thing stopping us from doing it now is that driver-legacy was never refactored to use SyncOperations and still constructs each operation directly.

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncOperations.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncOperations.java
@@ -1,0 +1,274 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.internal.operation;
+
+import com.mongodb.AutoEncryptionSettings;
+import com.mongodb.MongoNamespace;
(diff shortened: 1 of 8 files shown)


driver-core/src/main/com/mongodb/internal/operation/AsyncOperations.java
```diff
@@ -0,0 +1,274 @@
...
+import org.bson.BsonValue;
+import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.conversions.Bson;
+
+import java.util.List;
+
+/**
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public final class AsyncOperations<TDocument> {
```
jyemin, line 55: Copied from SyncOperations and just changed the interface types to be the Async equivalents

driver-core/src/main/com/mongodb/internal/operation/Operations.java
```diff
@@ -74,7 +74,7 @@
 import static java.util.Collections.singletonList;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
-public final class Operations<TDocument> {
+final class Operations<TDocument> {
```
jyemin, line 77: This can now be package private

driver-reactive-streams/src/test/unit/com/mongodb/reactivestreams/client/internal/TestHelper.java
```diff
@@ -171,8 +170,6 @@ private static Object checkValueTypes(final Object instance) {
         Object actual = instance instanceof Optional ? ((Optional<Object>) instance).orElse(instance) : instance;
         if (actual instanceof AsyncReadOperation || actual instanceof AsyncWriteOperation) {
             return getClassPrivateFieldValues(actual);
-        } else if (actual instanceof Operations) {
```
jyemin, line 174: I put a break point here and the branch was never entered, so I just removed it

stIncMale (approved):

PR #1030, opened 2022-10-31
Move legacy write operations into driver-legacy

jyemin:
JAVA-4795

* Deleted Insert/Update/Delete Operation since they served no useful function
* Moved BaseWriteOperation into driver-legacy and refactored it so it could server all three write types
* Moved and consolidated the functional tests


driver-legacy/src/main/com/mongodb/LegacyMixedBulkWriteOperation.java
```diff
@@ -204,26 +148,24 @@ private WriteConcernResult translateBulkWriteResult(final BulkWriteResult bulkWr
...
-        } else if (getType() == DELETE) {
+        } else if (type == DELETE) {
             count = bulkWriteResult.getDeletedCount();
         }
         return count;
     }
 
     private boolean getUpdatedExisting(final BulkWriteResult bulkWriteResult) {
-        if (getType() == UPDATE) {
+        if (type == UPDATE || type == REPLACE) {
```
jyemin, line 161: Tests uncovered this.  UpdateOperation claimed all updates where of type UPDATE, but in reality some are of type REPLACE.  Now that we're getting the type from the first write request, some updates are of type REPLACE and this condition was being incorrectly skipped.

driver-legacy/src/main/com/mongodb/LegacyMixedBulkWriteOperation.java
```diff
@@ -182,11 +126,11 @@ private BsonDocument manufactureGetLastErrorResponse(final MongoBulkWriteExcepti
 
     private void addBulkWriteResultToResponse(final BulkWriteResult bulkWriteResult, final BsonDocument response) {
         response.put("ok", new BsonInt32(1));
-        if (getType() == INSERT) {
+        if (type == INSERT) {
             response.put("n", new BsonInt32(0));
-        } else if (getType() == DELETE) {
+        } else if (type == DELETE) {
```
jyemin, line 131: There's unfortunately no existing (or new) test coverage of these branches.

driver-legacy/src/main/com/mongodb/LegacyMixedBulkWriteOperation.java
```diff
@@ -47,64 +40,42 @@
...
     private Boolean bypassDocumentValidation;
-    private BsonValue comment;
 
-    public BaseWriteOperation(final MongoNamespace namespace, final boolean ordered, final WriteConcern writeConcern,
-                              final boolean retryWrites) {
+    LegacyMixedBulkWriteOperation(final MongoNamespace namespace, final boolean ordered, final WriteConcern writeConcern,
+            final boolean retryWrites, final List<? extends WriteRequest> writeRequests) {
+        this.writeRequests = notNull("writeRequests", writeRequests);
+        isTrueArgument("writeRequests not empty", !writeRequests.isEmpty());
+        this.type = writeRequests.get(0).getType();
```
stIncMale, line 58: There appears to be nothing preventing `writeRequests` from containing requests of different types. We can change this by making the constructor `private` and exposing `static` methods that take lists of specific subtypes of `WriteRequest` that are not subtypes of each other.

stIncMale (changes requested):

driver-legacy/src/main/com/mongodb/LegacyMixedBulkWriteOperation.java
jyemin: Done, though since update and replace are both represented by UpdateRequest, I didn't bother to try to distinguish between them.  We could add an assertion to ensure that update and replace are not mixed, but seems like overkill.

driver-legacy/src/main/com/mongodb/LegacyMixedBulkWriteOperation.java
```diff
@@ -47,64 +40,42 @@
 
 
 /**
- * Abstract base class for write operations.
- *
- * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ * Abstract base class for legacy write operations.
  */
-public abstract class BaseWriteOperation implements AsyncWriteOperation<WriteConcernResult>, WriteOperation<WriteConcernResult> {
+class LegacyMixedBulkWriteOperation implements WriteOperation<WriteConcernResult> {
```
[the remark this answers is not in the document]
jyemin: Done

jyemin:
Changes seem uncontroversial, so merging.

PR #1032, opened 2022-11-02
Expressions

katcharov:
JAVA-3879

Tracking PR.

This PR may be used to track anything that needs to be done prior to this expressions branch being merged into master.


stIncMale (commented):
-

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -0,0 +1,47 @@
...
+ * this API must mitigate any risk of applying an expression to some type where
+ * resulting behaviour is not defined by this API (for example, by checking for
+ * null, by ensuring that field types are correctly specified). Likewise, unless
+ * otherwise specified, this API does not define the order of evaluation for all
+ * arguments, and whether all arguments to some expression will be evaluated.
+ *
+ * @see Expressions
+ */
+@Evolving
+public interface Expression {
```
stIncMale, line 45: All javadoc comments covered by [JAVA-4799](https://jira.mongodb.org/browse/JAVA-4799) must be re-reviewed before merging this PR.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -0,0 +1,125 @@
...
+
+
+    public static ArrayExpression<IntegerExpression> ofIntegerArray(final int... ofIntegerArray) {
+        List<BsonValue> array = Arrays.stream(ofIntegerArray)
+                .mapToObj(BsonInt32::new)
+                .collect(Collectors.toList());
+        return new MqlExpression<>((cr) -> new BsonArray(array));
+    }
+
+    public static DocumentExpression ofDocument(final Bson document) {
```
katcharov, line 113: The names of all expression methods and helper methods should be reviewed for consistency.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: Javadoc: mapping table from MQL to corresponding expressions.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -0,0 +1,64 @@
...
+     * Performs a reduction on the elements of this array, using the provided
+     * identity value and an associative reducing function, and returns
+     * the reduced value. The initial value must be the identity value for the
+     * reducing function.
+     *
+     * @param initialValue the identity for the reducing function
+     * @param in the associative reducing function
+     * @return the reduced value
+     */
+    T reduce(T initialValue, BinaryOperator<T> in);
```
katcharov, line 62: If we can avoid exposing this method, we should try to do so using convenience methods (though there are some unresolved difficulties in how we would accomplish this). See https://github.com/mongodb/mongo-java-driver/pull/1031#discussion_r1013433929

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: 1. The return types of methods should be reviewed. (consider `<T extends DocumentEx>`)
2. Review all comments in https://github.com/mongodb/mongo-java-driver/pull/1023

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -0,0 +1,125 @@
...
+     * Returns an expression having the same string value as the provided
+     * string.
+     *
+     * @param of the string
+     * @return the string expression
+     */
+    public static StringExpression of(final String of) {
+        Assertions.notNull("String", of);
+        return new MqlExpression<>((codecRegistry) -> new BsonString(of));
+    }
```
stIncMale, line 88: The use of `Assertions.notNull` to check arguments in public API methods is inconsistent. The check should either be done in all API methods in the PR, or in none.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -0,0 +1,220 @@
...
+import org.bson.BsonArray;
+import org.bson.BsonDocument;
+import org.bson.BsonString;
+import org.bson.BsonValue;
+import org.bson.codecs.configuration.CodecRegistry;
+
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+
+final class MqlExpression<T extends Expression>
```
katcharov, line 28: Check all access modifiers.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+/**
+ * Expresses a document value. A document is an ordered set of fields, where the
+ * key is a string value, mapping to a value of any other expression type.
+ */
+public interface DocumentExpression extends Expression {
```
katcharov, line 23: This needs to clarify that this is intended to represent a record (where any particular record (class, in Java) would have a fixed set of field keys). To represent non-fixed keys, use maps (presently called MapExpressions).

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: For 1 above, confirm whether "schema objects" would extend DocEx.

katcharov: Make set and unset match in their return: https://github.com/mongodb/mongo-java-driver/pull/1052#discussion_r1031945379

stIncMale:
We need to document the server version requirement for each methods in the new API. This requirement depends on the MQL used by the implementation of a method, and the implementation is allowed be changed in the future only in such ways that either don't increase the server version requirement, or increase it up to the least supported server version.

driver-core/src/main/com/mongodb/client/model/expressions/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * API for MQL expressions.
+ *
+ * @see com.mongodb.client.model.expressions.Expression
+ */
+@NonNullApi
+package com.mongodb.client.model.expressions;
```
stIncMale, line 23: We need to document the server version requirement for each methods in the new API. This requirement depends on the MQL used by the implementation of a method, and the implementation is allowed be changed in the future only in such ways that either don't increase the server version requirement, or increase it up to the least supported server version.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: Should probably be renamed to "Record".

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    IntegerExpression subtract(IntegerExpression i);
+
+    default IntegerExpression subtract(final int subtract) {
+        return this.subtract(Expressions.of(subtract));
+    }
+
+    IntegerExpression max(IntegerExpression i);
+    IntegerExpression min(IntegerExpression i);
```
katcharov, line 42: It seems that min/max should also be available on Expression (and all subclasses).

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -0,0 +1,121 @@
...
+        implements Expression, BooleanExpression, IntegerExpression, NumberExpression,
+        StringExpression, DateExpression, DocumentExpression, ArrayExpression<T> {
+
+    private final Function<CodecRegistry, BsonValue> fn;
+
+    MqlExpression(final Function<CodecRegistry, BsonValue> fn) {
+        this.fn = fn;
+    }
```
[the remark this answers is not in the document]
stIncMale: It is used at least to call `Bson.toBsonDocument` in `Expressions.of(Bson)`.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
stIncMale: Moved from [here](https://github.com/mongodb/mongo-java-driver/pull/1050#discussion_r1030858270): `MqlExpression.isBoolean` and similar methods should probably be `private`.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArrayExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    @Test
+    public void mapTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/map/
+        assertExpression(
+                Stream.of(true, true, false)
+                        .map(v -> !v).collect(Collectors.toList()),
+                arrayTTF.map(v -> v.not()),
+                // MQL:
```
stIncMale, line 115: We may want to re-review the comments left in the code (currently, they are mostly in the test code) to see if they are helpful.

Some examples of questionable comments:
- `(Incomplete)` at the top of `ArrayExpressionsFunctionalTest`.
- `// MQL:` when passing a value of the `expectedMql` parameter.

driver-core/src/main/com/mongodb/client/model/expressions/package-info.java
stIncMale, line 23: We need Scala API. Similarly to `com.mongodb.client.model.search` (Java) vs `org.mongodb.scala.model.search` (Scala), we probably only need to add Scala type aliases and counterparts to `static` Java methods.

driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
```diff
@@ -0,0 +1,47 @@
...
+
+    StringExpression toUpper();
+
+    StringExpression concat(StringExpression concat);
+
+    IntegerExpression strLen();
+
+    IntegerExpression strLenBytes();
+
+    StringExpression substr(IntegerExpression start, IntegerExpression length);
```
katcharov, line 36: Should also add `substr(IntegerExpression start)` - skip? Forcing user to put a negative length is bad.

driver-core/src/test/functional/com/mongodb/client/model/expressions/AbstractExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,124 @@
...
+    private static class BsonDocumentFragmentCodec extends BsonDocumentCodec {
+        public BsonValue readValue(final BsonReader reader, final DecoderContext decoderContext) {
+            reader.readBsonType();
+            return super.readValue(reader, decoderContext);
+        }
+    }
+
+
+    static <R extends Expression> R ofRem() {
+        // $$REMOVE is intentionally not exposed to users
```
katcharov, line 119: Is it possible to eliminate `missing` from the type system? That is, it would not even be mentioned in docs, but only used internally by some of our operations. It is an interesting mechanism, but other languages seem to do fine without it.

It arises only during `get`-like operations, and is usually (but not always) converted to null by ensuing expressions.

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
```diff
@@ -0,0 +1,61 @@
...
+     * If this expression evaluates to logical true, returns the result of the
+     * evaluated left branch expression. If this evaluates to logical false,
+     * returns the result of the evaluated right branch expression.
+     *
+     * @param left the left branch expression
+     * @param right the right branch expression
+     * @return left if true, right if false.
+     * @param <T> The type of the resulting expression.
+     */
+    <T extends Expression> T cond(T left, T right);
```
katcharov, line 60:
```
one.eq(of(2)).cond(of("yes"), of("no"));
```
The present variant of cond does not supply the original value to the left and right branches. Above, the value is `one`, but it could be a long composite expression. This is inconvenient. (It also fails to imply that the not-taken branch will not be evaluated, since there is no lambda; this problem is also present in `and` and `or`.) 

```
one.condNumber(v -> v.eq(of(2)), v -> of("yes"), v -> of("no"));
```

This alternative has two problems. First, it is always verbose. Second, because there is no self type, distinct variations must be provided for each type, which makes the name confusing (`ifNumber` sounds like a typecheck).

If the value must be supplied, switch can be used instead. The only reason to choose the second variant is to enforce evaluation semantics, in which case the parameters of `and` and `or` should also be lambdas.

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
```diff
@@ -0,0 +1,59 @@
...
+
+    NumberExpression subtract(NumberExpression n);
+
+    default NumberExpression subtract(final Number subtract) {
+        return this.subtract(Expressions.numberToExpression(subtract));
+    }
+
+    NumberExpression max(NumberExpression n);
+
+    NumberExpression min(NumberExpression n);
```
katcharov, line 50: Consider whether [reduce/min](https://github.com/mongodb/mongo-java-driver/pull/1053#discussion_r1037632332), getField(s), isArrayOr, and various other "non-secure" methods could all have a default value specified, or use Expression, for this initial release to have a "secured" API.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -0,0 +1,98 @@
...
+
+    NumberExpression getNumber(String fieldName);
+
+    NumberExpression getNumber(String fieldName, NumberExpression other);
+
+    default NumberExpression getNumber(final String fieldName, final double other) {
+        return getNumber(fieldName, of(other));
+    }
+
+    default NumberExpression getNumber(final String fieldName, final Decimal128 other) {
```
katcharov, line 54: @stIncMale / see [prior discussion](https://github.com/mongodb/mongo-java-driver/pull/1052#discussion_r1031944153)

I plan to revert these methods back to Number, because of the following difference between of and getT:

```
        assertExpression(99, ofDoc("{}").getInteger("a", 99));
        assertExpression(99, ofDoc("{}").getNumber("a", 99)); // fails round-trip
        assertExpression(99L, ofDoc("{}").getNumber("a", 99L)); // fails round-trip
        assertExpression(99.0, ofDoc("{}").getNumber("a", 99.0));
        assertExpression(99, of(99));
        assertExpression(99L, of(99L));
        assertExpression(99.0, of(99.0));
        assertExpression(99, ofDoc("{}").getNumber("a", of(99)));
        assertExpression(99L, ofDoc("{}").getNumber("a", of(99L)));
```
If we use Number on `of`, then we must have both `ofInteger` and `ofNumber`. If we provide both a `getInteger(int)` and a `getNumber(int)`, then our intention in providing both methods seems confusing to users (if we did not catch the above round-tripping failure, they will not think it obvious).

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
katcharov: done in PR https://github.com/mongodb/mongo-java-driver/pull/1053

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: For clarity, the "round-tripping" is perhaps better described as a general problem of user expectations about types. It is true that technically, the failing `.getNumber("a", 99)` accepts a double rather than an int, but that is irrelevant since the user sees what appears to be an int, and would expect it to be accepted as an int, and then come back as an int.

katcharov: Reverted in https://github.com/mongodb/mongo-java-driver/pull/1053#discussion_r1060769700

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.time.Instant;
+
+import static com.mongodb.client.model.expressions.Expressions.of;
+
+/**
+ * Expresses a document value. A document is an ordered set of fields, where the
+ * key is a string value, mapping to a value of any other expression type.
+ */
+public interface DocumentExpression extends Expression {
```
stIncMale, line 30: The BSON spec calls an ordered collection of key/value pairs (with no key duplicates) a "document", while the MongoDB documentation calls it an "object". The "expressions API" splits the document/object type into two: "MQL document" and "MQL map<T>" in such a way that:

1. Any "MQL document" and any "MQL map" is an "BSON document/object".
2. Not any "BSON document/object" is "MQL document".
3. Not any "BSON document/object" is "MQL map<T>", unless `T` is `Expression` (which is supposed to be named "MQL value").

Due to item 2, the meaning of the word "document" in "BSON document/object" is different from that in "MQL document", i.e., we have overloaded its meaning. We should use a different word instead of overloading the existing one. Even "MQL record" appears to be better than "MQL document", because despite being used in "Java records", it is not used within the BSON/MongoDB domain. It is much more likely (almost guaranteed) that "MQL document" will be confused with "BSON document/object", and much less likely that "MQL record" will be confused with "Java record".

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
stIncMale: Suggestion: we can use the term "unchecked" similarly to unchecked conversions<sup>1</sup> in Java, because this is exactly what we are doing here; there is no need to invent a new word (secure/safe) for the same thing.

---

<sup>1</sup> The correctness of unchecked conversions can't checked at runtime; i.e., they always succeed, but potentially lead to incorrect behavior later.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: The naming hinges on whether our field-bearing class extends what we presently call `MapExpression<Expression>`. If it does, then it seems reasonable to call it a document. If it does not, then it should probably be called a record (but this seems odd, because now nothing is called a document).

stIncMale: How does extending or not extending `MapExpression<Expression>` removes the overloaded nature of the word "document"? Whether `DocumentExpression` is `MapExpression<Expression>` or not, "document" in `DocumentExpression` (what I called "MQL document" above) has still a different meaning than "BSON document" because item 2 [here](https://github.com/mongodb/mongo-java-driver/pull/1032#discussion_r1062674087) still holds.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
stIncMale: - `MqlExpression.sort` should probably be package-access.
- `MqlExpression.reduce`/`isDate`/`isDocument` should probably be `private`.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
stIncMale, line 30: We should not forget to do `MqlDocument extends MqlMap<Expression>`.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -0,0 +1,217 @@
...
+            List<BsonValue> list = new ArrayList<>();
+            for (T v : array) {
+                Assertions.notNull("elements of array", v);
+                list.add(((MqlExpression<?>) v).toBsonValue(cr));
+            }
+            return new AstPlaceholder(new BsonArray(list));
+        });
+    }
+
+    public static DocumentExpression of(final Bson document) {
```
stIncMale, line 187: To differentiate between `of(Bson)` producing `DocumentExpression` and `of(Bson)` producing `MapExpression`, we call the latter `ofMap` and have:
```
DocumentExpression of(Bson);
MapExpression ofMap(Bson);
MapExpression ofMap();

```

Renaming `of(Bson)` to `ofDocument(Bson)` appears to make naming more consistent:

- if `of` can be distinguished based on the input type, we use the name "of";
- if `of` cannot be distinguished based on the input type, we include the result type in the name.

Currently all methods follow the above rules expect for `DocumentExpression of(Bson)`.

driver-core/src/main/com/mongodb/client/model/expressions/package-info.java
stIncMale, line 23: Check if there are any `TODO`s in the code of this package and in tests that slipped through and resolve them.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -0,0 +1,115 @@
...
+     * @param other
+     * @return
+     */
+    BooleanExpression isBooleanOr(BooleanExpression other);
+    NumberExpression isNumberOr(NumberExpression other);
+    IntegerExpression isIntegerOr(IntegerExpression other);
+    StringExpression isStringOr(StringExpression other);
+    DateExpression isDateOr(DateExpression other);
+    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<? extends T> other);
+    <T extends DocumentExpression> T isDocumentOr(T other);
```
stIncMale, line 112: [optional]

Let's consider whether we should have `DocumentExpression isDocumentOr(DocumentExpression other)` for now here. If we come up with a realization of how to support user subtypes of `DocumentExpression`, and if it will require having a method like the current `isDocumentOr`, then we can add something like `<T extends DocumentExpression> T isUserDocumentOr(T other)`. I am proposing to consider this because it is both possible that either we won't realize how to support user types, or, that it will not need `isDocumentOr` to return a user type because values of user types will be constructed somehow else.

driver-core/src/main/com/mongodb/client/model/expressions/package-info.java
stIncMale, line 23: Given that this package currently provides the very core of a new computational API that uses MQL behind the scenes, is the name `expressions` really a good fit for it? Maybe `mql` is a better package name, or something else?

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
stIncMale: Extending turned out to be impossible.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -0,0 +1,115 @@
...
+     * also checks for nulls
+     * @param other
+     * @return
+     */
+    BooleanExpression isBooleanOr(BooleanExpression other);
+    NumberExpression isNumberOr(NumberExpression other);
+    IntegerExpression isIntegerOr(IntegerExpression other);
+    StringExpression isStringOr(StringExpression other);
+    DateExpression isDateOr(DateExpression other);
+    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<? extends T> other);
```
stIncMale, line 111: Let's add something like `@MqlUnchecked` (the draft PR: https://github.com/mongodb/mongo-java-driver/pull/1068) to simplify documenting what is not checked by the API, and to make it more noticeable to users. Usage examples:

```java
ArrayExpression {
    @MqlUnchecked(NON_EMPTY)
    T first();
}
```

```java
Expression {
    <T extends Expression> ArrayExpression<@MqlUnchecked T> isArrayOr(ArrayExpression<? extends T> other);
}
```

```java
DocumentExpression {
    @MqlUnchecked(TYPE)
    NumberExpression getNumber(String fieldName);
}
```

driver-core/src/main/com/mongodb/client/model/expressions/package-info.java
katcharov: I was wondering how to rename it, and `mql` is a good idea.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: I think these are addressed as of of https://github.com/mongodb/mongo-java-driver/pull/1070

Except for the mapping table, but I believe this should/would be part of our non-javadocs.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: It seems that we could skip methods which would cause an NPE regardless?

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
katcharov: (Some of those are used by branches, so non-private)

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: I believe we discussed and resolved this, that it will remain called Document?

driver-core/src/main/com/mongodb/client/model/expressions/package-info.java
katcharov: Should be done as of https://github.com/mongodb/mongo-java-driver/pull/1070

driver-core/src/test/functional/com/mongodb/client/model/expressions/AbstractExpressionsFunctionalTest.java
katcharov: `has` was used for this purpose, both on map and doc.

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
katcharov: These should be documented, and annotated, as of https://github.com/mongodb/mongo-java-driver/pull/1070

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: Unsure. It seems that ofDocument is a "default" that would be used much more often than map, and that this is similar to how we handled current/currentAsMap?

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: It seems that the result might be constructed on the bases of some information in the implementation of the "other" parameter. I think we should leave it for consistency - I think having two methods would be confusing.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: included as of https://github.com/mongodb/mongo-java-driver/pull/1070

driver-core/src/main/com/mongodb/client/model/expressions/package-info.java
stIncMale, line 23: We also need `@since` on the package and the new public classes/interfaces to express the driver version in which they were added.

stIncMale: Moved to https://github.com/mongodb/mongo-java-driver/pull/1070.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
stIncMale: This suggestion is conceptually the same as the following more general approach: _"if an invalid method argument always causes a runtime error (given the current state of the code), there is no need to validate the argument"_, which is a bad idea.

Just in case, the proposed approach (they are generalizable and are equally applicable to the more general approach described above):

1. Explicit checks provide an error message that explicitly points out to a user mistake.
2. Without an explicit check the NPE may, generally speaking, happen somewhere deep in the call stack that originates from method where an explicit check should have been put.
3. Placing the checks selectively burdens readers with the task of repeating the manual code analysis to figure out if the code fails with an NPE despite not having an explicit check, or there is a bug in the code.
4. Even if one manages to analyze all the current code correctly (which is questionable), future changes may, generally speaking, result in the method stopping to produce an NPE and starting to behave incorrectly in some other way.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
stIncMale: Yes, there is nothing left to do or discuss here.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
stIncMale: This is indeed consistent with `current/currentAsMap`.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: Fair enough. I need to think of how to reliably test this.

katcharov:
Rebased, minor no-conflict re-order and squash of one commit, and reworded commit messages to include JIRA and PR numbers.

stIncMale (approved):

PR #1033, opened 2022-11-02
Implement eq, ne, gt, gte, lt, lte

katcharov:
JAVA-4784

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -112,12 +112,9 @@ public final class Expressions {
 
     public static DocumentExpression ofDocument(final Bson document) {
         Assertions.notNull("document", document);
-        // All documents are wrapped in a $literal. It is possible to check
-        // for empty docs (see https://jira.mongodb.org/browse/SERVER-46422)
-        // and wrap them, and also to traverse into each document to see if
-        // it needs to be conditionally wrapped in a $literal, but doing so
-        // would be brittle and unsafe - we might miss some future case and
-        // allow an expression where a literal document was expected.
+        // All documents are wrapped in a $literal. If we don't wrap, we need to
+        // check for empty documents and documents that are actually expressions
+        // (and need to be wrapped in $literal anyway). This would be brittle.
         return new MqlExpression<>((cr) -> new BsonDocument("$literal",
                 document.toBsonDocument(BsonDocument.class, cr)));
     }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -117,7 +117,6 @@ public final class Expressions {
         // and to traverse for $$, but doing so would be unsafe and brittle.
         return new MqlExpression<>((cr) -> document == null
                 ? new BsonNull()
-                // TODO what is this first .class parameter?
                 : new BsonDocument("$literal", document.toBsonDocument(BsonDocument.class, cr)));
     }
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -18,12 +18,22 @@ package com.mongodb.client.model.expressions;
 
 import org.bson.BsonArray;
 import org.bson.BsonBoolean;
+import org.bson.BsonDateTime;
+import org.bson.BsonDocument;
+import org.bson.BsonDouble;
 import org.bson.BsonInt32;
+import org.bson.BsonInt64;
+import org.bson.BsonNull;
 import org.bson.BsonString;
 import org.bson.BsonValue;
+import org.bson.conversions.Bson;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Date;
 import java.util.List;
+import java.util.stream.Collectors;
+
 
 /**
  * Convenience methods related to {@link Expression}.
@@ -54,6 +64,15 @@ public final class Expressions {
     public static IntegerExpression of(final int of) {
         return new MqlExpression<>((codecRegistry) -> new BsonInt32(of));
     }
+    public static IntegerExpression of(final long of) {
+        return new MqlExpression<>((codecRegistry) -> new BsonInt64(of));
+    }
+    public static NumberExpression of(final double of) {
+        return new MqlExpression<>((codecRegistry) -> new BsonDouble(of));
+    }
+    public static NumberExpression of(final Date of) {
+        return new MqlExpression<>((codecRegistry) -> new BsonDateTime(of.getTime()));
+    }
 
     /**
      * Returns an expression having the same string value as the provided
@@ -81,4 +100,28 @@ public final class Expressions {
         return new MqlExpression<>((cr) -> new BsonArray(result));
     }
 
+
+    public static ArrayExpression<IntegerExpression> ofIntegerArray(final int... ofIntegerArray) {
+        List<BsonValue> array = Arrays.stream(ofIntegerArray)
+                .mapToObj(BsonInt32::new)
+                .collect(Collectors.toList());
+        return new MqlExpression<>((cr) -> new BsonArray(array));
+    }
+
+    public static DocumentExpression ofDocument(final Bson document) {
+        if (document == null) {
+            throw new IllegalArgumentException("document cannot be null");
+        }
+        // All documents are wrapped in a $literal. It is possible to check
+        // for empty docs (see https://jira.mongodb.org/browse/SERVER-46422)
+        // and to traverse for $$, but doing so would be unsafe and brittle.
+        return new MqlExpression<>((cr) -> document == null
+                ? new BsonNull()
+                // TODO what is this first .class parameter?
+                : new BsonDocument("$literal", document.toBsonDocument(BsonDocument.class, cr)));
+    }
+
+    public static <R extends Expression> R ofNull() {
+        return new MqlExpression<>((cr) -> new BsonNull()).assertImplementsAllExpressions();
+    }
 }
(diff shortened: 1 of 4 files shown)


driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -54,6 +64,15 @@ public static BooleanExpression of(final boolean of) {
     public static IntegerExpression of(final int of) {
         return new MqlExpression<>((codecRegistry) -> new BsonInt32(of));
     }
+    public static IntegerExpression of(final long of) {
+        return new MqlExpression<>((codecRegistry) -> new BsonInt64(of));
+    }
+    public static NumberExpression of(final double of) {
+        return new MqlExpression<>((codecRegistry) -> new BsonDouble(of));
+    }
+    public static NumberExpression of(final Date of) {
```
jyemin, line 73: Should this return a `DateExpression`?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -81,4 +100,27 @@ public static ArrayExpression<BooleanExpression> ofBooleanArray(final boolean...
...
+    }
+
+    public static DocumentExpression ofDocument(final Bson document) {
+        if (document == null) {
+            throw new IllegalArgumentException("document cannot be null");
+        }
+        // All documents are wrapped in a $literal. It is possible to check
+        // for empty docs (see https://jira.mongodb.org/browse/SERVER-46422)
+        // and to traverse for $$, but doing so would be unsafe and brittle.
+        return new MqlExpression<>((cr) -> document == null
```
jyemin, line 118: Why are we checking for null here if we throw on null just above?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -81,4 +100,27 @@ public static ArrayExpression<BooleanExpression> ofBooleanArray(final boolean...
         return new MqlExpression<>((cr) -> new BsonArray(result));
     }
 
+
+    public static ArrayExpression<IntegerExpression> ofIntegerArray(final int... ofIntegerArray) {
```
jyemin, line 104: Should we have a separate PR for all the `of` methods instead of sprinkling them in like this?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -81,4 +100,27 @@ public static ArrayExpression<BooleanExpression> ofBooleanArray(final boolean...
...
+        return new MqlExpression<>((cr) -> new BsonArray(array));
+    }
+
+    public static DocumentExpression ofDocument(final Bson document) {
+        if (document == null) {
+            throw new IllegalArgumentException("document cannot be null");
+        }
+        // All documents are wrapped in a $literal. It is possible to check
+        // for empty docs (see https://jira.mongodb.org/browse/SERVER-46422)
+        // and to traverse for $$, but doing so would be unsafe and brittle.
```
stIncMale, line 117: Could you please help me to understand this comment? Given the https://jira.mongodb.org/browse/SERVER-46422 link, I am guessing that all values are wrapped in `$literal` because this was one may `$addFields` with an empty document. Is this correct?

Regardless of whether the above guess is correct, I also need help with the following questions:

- What does _"to check for empty docs and to traverse for `$$`"_ mean and how is it related to creating a document expression?
- What does _"unsafe"_ mean in this context?

driver-core/src/test/functional/com/mongodb/client/model/expressions/ComparisonExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,186 @@
...
+                of(0).lt(of(0)),
+                "{'$lt': [0, 0]}");
+
+        assertExpression(
+                true,
+                ofNull().lt(of(0)),
+                "{'$lt': [null, 0]}");
+
+        for (int i = 0; i < sampleValues.size() - 1; i++) {
+            int j = i + 1;
```
stIncMale, line 135: We may also do `for (int j = i + 1; j < sampleValues.size(); j++)` to cover more pairs.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ComparisonExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,186 @@
...
+    // Comparison expressions are part of the the generic Expression class.
+
+    static <R extends Expression> R ofRem() {
+        // $$REMOVE is intentionally not exposed to users
+        return new MqlExpression<>((cr) -> new BsonString("$$REMOVE")).assertImplementsAllExpressions();
+    }
+
+    // https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/#std-label-bson-types-comparison-order
+    private final List<Expression> sampleValues = Arrays.asList(
+            ofRem(),
```
stIncMale, line 50: Why do we have [`$$REMOVE`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/project/) here given that users are unable to express it via `Expression`?

driver-core/src/test/functional/com/mongodb/client/model/expressions/ComparisonExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,186 @@
...
+        for (int i = 0; i < sampleValues.size() - 1; i++) {
+            int j = i + 1;
+            Expression first = sampleValues.get(i);
+            Expression second = sampleValues.get(j);
+            BsonValue evaluate = evaluate(first.lt(second));
+            if (!evaluate.asBoolean().getValue()) {
+                BsonValue v1 = ((MqlExpression<?>) first).toBsonValue(fromProviders(new BsonValueCodecProvider()));
+                BsonValue v2 = ((MqlExpression<?>) second).toBsonValue(fromProviders(new BsonValueCodecProvider()));
+                fail(i + "," + j + " --" + v1 + " < " + v2 + " should be true");
+            }
```
stIncMale, line 143: Should the same be done in `eqTest` instead of reporting only the indices of failed items?

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: Looks like I fixed this in an ensuing PR: https://github.com/mongodb/mongo-java-driver/pull/1037/files#diff-26a114bb575c4831fa43704d692f44d6fcf087823101af8ab71722d5d9e32276L121

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: I introduce and test the literal `of` methods for each expression type in the corresponding PR; I don't think they should all be introduced separately. For example, the boolean `of` was part of the prior boolean PR. But I had to advance these into this PR to properly test eq/lt/etc.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -81,4 +100,27 @@ public static ArrayExpression<BooleanExpression> ofBooleanArray(final boolean...
...
+        List<BsonValue> array = Arrays.stream(ofIntegerArray)
+                .mapToObj(BsonInt32::new)
+                .collect(Collectors.toList());
+        return new MqlExpression<>((cr) -> new BsonArray(array));
+    }
+
+    public static DocumentExpression ofDocument(final Bson document) {
+        if (document == null) {
+            throw new IllegalArgumentException("document cannot be null");
+        }
```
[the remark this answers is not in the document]
katcharov: Ok. Note that this obscures test coverage of line 113 above.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ComparisonExpressionsFunctionalTest.java
katcharov: They can express it by calling getField when a field is not present. I think it will ultimately have to be represented in some way (I am not sure what the best way will be; same for null), and here it is represented literally.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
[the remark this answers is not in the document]
katcharov: `of` wrappers for T will be introduced in the PR that introduces T. Wrappers for arrays, when array T or T is introduced, whichever is later.

(I think helpers for primitives are fine, but I don't think the API needs to be aware of BSON.)

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: Updated the comment. I don't want to dynamically detect whether a document needs to be wrapped in a literal. Instead, we should wrap all docs.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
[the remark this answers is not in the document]
jyemin: It at least need something like `Expressions.of(new Decimal128(...))`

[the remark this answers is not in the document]
stIncMale: > I think helpers for primitives are fine, but I don't think the API needs to be aware of BSON

`Decimal128` is a new Java type (just like `BigInteger` or `Long`) that we introduced as a counterpart to BSON `Decimal128` type. If the API does not allow users to use it, then it just limits the set of types one may compute over in MongoDB.

> of wrappers for T will be introduced in the PR that introduces T.

Does this mean that https://github.com/mongodb/mongo-java-driver/pull/1037 is the right place for `of(Decimal128)`?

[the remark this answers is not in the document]
katcharov: Correct. I added it there, though using BigDecimal.

jyemin (approved):

stIncMale (approved):

rozza (approved):

PR #1034, opened 2022-11-03
[JAVA-4816] perf: improving performance of ObjectId parseHexString()

eonwhite:
JIRA Ticket: https://jira.mongodb.org/browse/JAVA-4816

The current implementation of `ObjectId` `parseHexString` is somewhat slow, because it makes a total of 12 calls to `Integer.parseInt`, which is not the fastest to begin with. This imposes some overhead to all `ObjectId(string)` calls, which can really add up at scale, for example when deserializing a large amount of ObjectId data via Jackson.

Replacing this implementation with a hand-rolled version which converts each part of the hex string to a byte. This in my rough testing is about 5x faster than the current implementation of `parseHexString`.

Tests (on my M1 MacBook Pro):

```
    public static void main(String[] args) {
        long startTime = System.currentTimeMillis();
        for (int i = 10000000; i < 100000000; i++) {
            new ObjectId("1234567890abcdef" + i);
        }
        long endTime = System.currentTimeMillis();

        System.out.println("Time elapsed: " + (endTime - startTime));
    }
```

On master:

```
Time elapsed: 19460
```

On branch:

```
Time elapsed: 3617
```

### DIFF
--- a/bson/src/main/org/bson/types/ObjectId.java
+++ b/bson/src/main/org/bson/types/ObjectId.java
@@ -423,12 +423,12 @@ public final class ObjectId implements Comparable<ObjectId>, Serializable {
         for (int i = 0; i < b.length; i++) {
             char c1 = s.charAt(i * 2);
             char c2 = s.charAt(i * 2 + 1);
-            b[i] = hexCharToByte(c1) * 16 + hexCharToByte(c2)
+            b[i] = (byte) ((hexCharToInt(c1) << 4) + hexCharToInt(c2));
         }
         return b;
     }
 
-    private static byte hexCharToByte(char c) {
+    private static int hexCharToInt(final char c) {
         if (c >= '0' && c <= '9') {
             return c - 48;
         } else if (c >= 'a' && c <= 'f') {
--- a/bson/src/main/org/bson/types/ObjectId.java
+++ b/bson/src/main/org/bson/types/ObjectId.java
@@ -420,12 +420,11 @@ public final class ObjectId implements Comparable<ObjectId>, Serializable {
         }
 
         byte[] b = new byte[OBJECT_ID_LENGTH];
-        int i = 0;
-        while (i < OBJECT_ID_LENGTH) {
-            char c1 = s.charAt(i);
-            char c2 = s.charAt(i + 1);
+        for (int i = 0; i < b.length; i++) {
+            int pos = i << 1;
+            char c1 = s.charAt(pos);
+            char c2 = s.charAt(pos + 1);
             b[i] = (byte) ((hexCharToInt(c1) << 4) + hexCharToInt(c2));
-            i += 2;
         }
         return b;
     }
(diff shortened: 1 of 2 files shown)


bson/src/main/org/bson/types/ObjectId.java
```diff
@@ -421,11 +421,24 @@ private static byte[] parseHexString(final String s) {
 
         byte[] b = new byte[OBJECT_ID_LENGTH];
         for (int i = 0; i < b.length; i++) {
-            b[i] = (byte) Integer.parseInt(s.substring(i * 2, i * 2 + 2), 16);
+            char c1 = s.charAt(i * 2);
+            char c2 = s.charAt(i * 2 + 1);
+            b[i] = hexCharToByte(c1) * 16 + hexCharToByte(c2)
```
jyemin, line 426: This doesn't compile due to a missing semicolon.

jyemin (changes requested):
Hi @eonwhite thanks for the contribution.  Can you do a couple of things to move this forward:

1. Open an issue in the JAVA project of MongoDB's Jira, and then reference the issue number in the PR description (see the CONTRIBUTING guide for details).  Please attach your performance tests as well.
2. Run `./gradlew bson:check` to make sure all static and runtime checks are passing

bson/src/main/org/bson/types/ObjectId.java
eonwhite: Sorry about that, was (sloppily) converting the code from Kotlin. Have resolved this and other style issues and verified passing `bson:check`

eonwhite:
> Hi @eonwhite thanks for the contribution. Can you do a couple of things to move this forward:
> 
> 1. Open an issue in the JAVA project of MongoDB's Jira, and then reference the issue number in the PR description (see the CONTRIBUTING guide for details).  Please attach your performance tests as well.
> 2. Run `./gradlew bson:check` to make sure all static and runtime checks are passing

Thanks @jyemin -- have resolved issues and verified that tests pass.

bson/src/main/org/bson/types/ObjectId.java
```diff
@@ -415,9 +415,8 @@ private Object readResolve() {
     }
 
     private static byte[] parseHexString(final String s) {
-        if (!isValid(s)) {
-            throw new IllegalArgumentException("invalid hexadecimal representation of an ObjectId: [" + s + "]");
-        }
+        notNull("hexString", s);
```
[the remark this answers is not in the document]
jyemin: > should we perhaps add a test case to verify the behavior of passing a 24-character string that does not resolve to a hexadecimal

I did add that,  on line 150 of the test file.

> it would be more helpful for the above to return invalid hexadecimal: [MY_BAD_STRING_THAT_IS_24] than invalid hexadecimal: M

I'm indifferent about this, since I don't think it's common and probably can be worked out from context.  But if you'd like to change it, I'll accept it.

jyemin:
Thanks again for the contribution @eonwhite

PR #1035, opened 2022-11-03
Support the Azure VM-assigned managed identity for automatic KMS credentials

jyemin:
JAVA-4706

* Add support for obtaining Azure credentials for KMS using local endpoint
* Refactored GCP-specific integration test to handle both GCP and Azure

After discussing with Kevin, determined that given the coverage provided by the integration tests, there is not enough additional value in the unit tests (which require starting a mock server) to get them running in CI.  So although they are implemented in the first commit, they are removed in the second one.


.evergreen/.evg.yml
```diff
@@ -1445,21 +1445,38 @@ tasks:
...
             working_dir: "src"
             shell: "bash"
             script: |
               ${PREPARE_SHELL}
-              MONGODB_URI='mongodb://localhost:27017' SUCCESS=false ./.evergreen/run-mongodb-fle-gcp-auto.sh
-
+              export AZUREKMS_RESOURCEGROUP=${testazurekms_resourcegroup}
+              export AZUREKMS_VMNAME=${AZUREKMS_VMNAME}
+              export AZUREKMS_PRIVATEKEYPATH=/tmp/testazurekms_privatekey
+              AZUREKMS_CMD="MONGODB_URI='mongodb://localhost:27017' PROVIDER="azure" ./.evergreen/run-fle-on-demand-credential-test.sh" $DRIVERS_TOOLS/.evergreen/csfle/azurekms/run-command.sh
```
DmitryLukyanov, line 1479: I wonder why did you miss `KEY_NAME='${testazurekms_keyname}' KEY_VAULT_ENDPOINT='${testazurekms_keyvaultendpoint}'`?

DmitryLukyanov (commented):
I don't see this part in this PR: https://github.com/mongodb/specifications/commit/d6b8cce6abb3b8e1a0b8f1dc7ee737e18322cfce#diff-de01d398114f63480af23798e09ffca599aa758efd04f15a9beec89b4566041bR344. Has it been implemented somewhere else?

.evergreen/.evg.yml
jyemin: Looks like those values are hardcoded in the integration test.  Is that not ok?

jyemin:
Agreed to defer credential caching to subsequent PR.  Is there anything else missing besides that?

.evergreen/.evg.yml
DmitryLukyanov: better to use env variables instead hardcoded values. That values are going to be changed in this ticket https://jira.mongodb.org/browse/BUILD-16154. Then, applying it in the driver will be easier with env variables instead of changing it in the code. NOTE: we will need to change the code in c# driver, my fault :(

jyemin: Done

jyemin:
@DmitryLukyanov ready for another look

DmitryLukyanov (approved):
LGTM  (assuming tests passed)

PR #1036, opened 2022-11-04
Implement string expressions

katcharov:
JAVA-4801 

Implements string expressions from the top 50, and a few others.

See: https://www.mongodb.com/docs/manual/reference/operator/aggregation/#string-expression-operators

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -32,7 +32,6 @@ import org.bson.conversions.Bson;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Date;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -84,9 +83,7 @@ public final class Expressions {
      * @return the string expression
      */
     public static StringExpression of(final String of) {
-        if (of == null) {
-            throw new IllegalArgumentException("string cannot be null");
-        }
+        Assertions.notNull("String", of);
         return new MqlExpression<>((codecRegistry) -> new BsonString(of));
     }
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -84,6 +84,9 @@ public final class Expressions {
      * @return the string expression
      */
     public static StringExpression of(final String of) {
+        if (of == null) {
+            throw new IllegalArgumentException("string cannot be null");
+        }
         return new MqlExpression<>((codecRegistry) -> new BsonString(of));
     }
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
@@ -25,8 +25,6 @@ import org.bson.codecs.configuration.CodecRegistry;
 import java.util.function.BinaryOperator;
 import java.util.function.Function;
 
-import static com.mongodb.client.model.expressions.Expressions.of;
-
 final class MqlExpression<T extends Expression>
         implements Expression, BooleanExpression, IntegerExpression, NumberExpression,
         StringExpression, DateExpression, DocumentExpression, ArrayExpression<T> {
@@ -206,17 +204,17 @@ final class MqlExpression<T extends Expression>
     }
 
     @Override
-    public Expression strLenBytes() {
+    public IntegerExpression strLenBytes() {
         return new MqlExpression<>(ast("$strLenBytes"));
     }
 
     @Override
-    public Expression substr(final int start, final int length) {
-        return new MqlExpression<>(ast("$substrCP", of(start), of(length)));
+    public StringExpression substr(final IntegerExpression start, final IntegerExpression length) {
+        return new MqlExpression<>(ast("$substrCP", start, length));
     }
 
     @Override
-    public Expression substrBytes(final int start, final int length) {
-        return new MqlExpression<>(ast("$substrBytes", of(start), of(length)));
+    public StringExpression substrBytes(final IntegerExpression start, final IntegerExpression length) {
+        return new MqlExpression<>(ast("$substrBytes", start, length));
     }
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
@@ -25,6 +25,8 @@ import org.bson.codecs.configuration.CodecRegistry;
 import java.util.function.BinaryOperator;
 import java.util.function.Function;
 
+import static com.mongodb.client.model.expressions.Expressions.of;
+
 final class MqlExpression<T extends Expression>
         implements Expression, BooleanExpression, IntegerExpression, NumberExpression,
         StringExpression, DateExpression, DocumentExpression, ArrayExpression<T> {
@@ -180,4 +182,41 @@ final class MqlExpression<T extends Expression>
                 .append("in", extractBsonValue(cr, in.apply(varThis, varValue)))).apply(cr));
     }
 
+
+    /** @see StringExpression */
+
+    @Override
+    public StringExpression toLower() {
+        return new MqlExpression<>(ast("$toLower"));
+    }
+
+    @Override
+    public StringExpression toUpper() {
+        return new MqlExpression<>(ast("$toUpper"));
+    }
+
+    @Override
+    public StringExpression concat(final StringExpression concat) {
+        return new MqlExpression<>(ast("$concat", concat));
+    }
+
+    @Override
+    public IntegerExpression strLen() {
+        return new MqlExpression<>(ast("$strLenCP"));
+    }
+
+    @Override
+    public Expression strLenBytes() {
+        return new MqlExpression<>(ast("$strLenBytes"));
+    }
+
+    @Override
+    public Expression substr(final int start, final int length) {
+        return new MqlExpression<>(ast("$substrCP", of(start), of(length)));
+    }
+
+    @Override
+    public Expression substrBytes(final int start, final int length) {
+        return new MqlExpression<>(ast("$substrBytes", of(start), of(length)));
+    }
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
@@ -21,4 +21,17 @@ package com.mongodb.client.model.expressions;
  */
 public interface StringExpression extends Expression {
 
+    StringExpression toLower();
+
+    StringExpression toUpper();
+
+    Expression concat(StringExpression concat);
+
+    Expression strLen();
+
+    Expression strLenBytes();
+
+    Expression substr(int start, int length);
+
+    Expression substrBytes(int start, int length);
 }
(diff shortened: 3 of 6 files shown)


driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
```diff
@@ -21,4 +21,17 @@
  */
 public interface StringExpression extends Expression {
 
+    StringExpression toLower();
+
+    StringExpression toUpper();
+
+    Expression concat(StringExpression concat);
```
jyemin, line 28: Quick question: why don't any of these return a narrower Expression type?

katcharov: Autogen error.

driver-core/src/test/functional/com/mongodb/client/model/expressions/StringExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,156 @@
...
+                "b",
+                of("abc").substrBytes(1, 1),
+                "{'$substrBytes': ['abc', 1, 1]}");
+
+        // unicode
+        byte[] bytes = Arrays.copyOfRange(sushi.getBytes(), 0, 3);
+        String expected = new String(bytes, StandardCharsets.UTF_8);
+        assertExpression(expected,
+                of(sushi).substrBytes(0, 3));
+        // "starting index is a UTF-8 continuation byte" if from 1 length 1
```
stIncMale, line 154: Can you please help me to understand this comment and its purpose?

I also don't think that either the starting or the ending index passed to `substrBytes(0, 3)` is a continuation byte for two reasons:

- The server would have produced an error.
- `寿司` is encoded as 6 UTF-8 bytes (2 code points, each is encoded as 3 UTF-8 8-bit code units, resulting in 6 bytes in total). So taking the first 3 bytes is taking `寿`, we are not breaking any character here.

driver-core/src/test/functional/com/mongodb/client/model/expressions/StringExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,156 @@
...
+                of("abc").strLen(),
+                "{'$strLenCP': 'abc'}");
+
+        // unicode
+        assertExpression(
+                jalapeno.length(),
+                of(jalapeno).strLen(),
+                "{'$strLenCP': '" + jalapeno + "'}");
+        assertExpression(
+                sushi.length(),
```
stIncMale, line 84: We should replace all `String.length` invocations in these tests with `String.codePointCount`. For example, `sushi.codePointCount(0, suchi.length())`.

The tests currently pass just by a chance, and you can see `assertExpression(sushi.length(), of(sushi).strLen(), "{'$strLenCP': '" + sushi + "'}")` failing if instead of `sushi = "\u5BFF\u53F8"` (`寿司`)  you use `sushi = "\u5BFF\u53F8\uD83E\uDD84"` (`寿司🦄`), where the third code point corresponding to the character `🦄` is encoded as two UTF-16 16-bit code units (`\uD83E` and `\uDD84`).

driver-core/src/test/functional/com/mongodb/client/model/expressions/StringExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,156 @@
...
+        assertExpression(
+                sushi.getBytes(StandardCharsets.UTF_8).length,
+                of(sushi).strLenBytes(),
+                "{'$strLenBytes': '" + sushi + "'}");
+
+        // comparison
+        assertExpression(2, of(sushi).strLen());
+        assertExpression(6, of(sushi).strLenBytes());
+        assertExpression(8, of(jalapeno).strLen());
+        assertExpression(9, of(jalapeno).strLenBytes());
```
stIncMale, line 111: The assertions with magic numbers (2, 6, ...) check nothing in addition to what the assertions above check, but are confusing to a reader. Let's remove them.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -183,4 +185,41 @@ public T reduce(final T initialValue, final BiFunction<T,  T,  T> in) {
...
+
+    @Override
+    public StringExpression toUpper() {
+        return new MqlExpression<>(ast("$toUpper"));
+    }
+
+    @Override
+    public StringExpression concat(final StringExpression concat) {
+        return new MqlExpression<>(ast("$concat", concat));
+    }
```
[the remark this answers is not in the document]
rozza: Can you provide an example of doing that?  It may be planned for the javadoc. 

I just want to make sure users coming from the [$concat docs](https://www.mongodb.com/docs/manual/reference/operator/aggregation/concat/) will easily be able to translate to this api.

driver-core/src/test/functional/com/mongodb/client/model/expressions/StringExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,156 @@
...
+        assertExpression(
+                "abc".toUpperCase(),
+                of("abc").toUpper(),
+                "{'$toUpper': 'abc'}");
+    }
+
+    @Test
+    public void strLenTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/strLenCP/ (?)
+        // TODO naming: just strLen, lengthCP, lengthCodePoints, stringLengthInCodePoints?...
```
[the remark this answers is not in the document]
stIncMale: > MQL substr method maps to substrBytes, but mine maps to substrCP

Yeah, I noticed that. I'd prefer `*CP` method names, so that we match MQL, and users will also be less confused. However, since one of the key ideas is to abstract the API from MQL, I can live with the current naming.

driver-core/src/test/functional/com/mongodb/client/model/expressions/StringExpressionsFunctionalTest.java
stIncMale: If a human can't write the expected value on his own, which is the case in these tests, then the expected value is just magic and the corresponding assertion is almost of no use. This is especially true when you have assertions that do the same but without using magic numbers.

> They demonstrate the actual expected values

What is the use of this information?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -84,9 +83,7 @@ public static DateExpression of(final Instant of) {
      * @return the string expression
      */
     public static StringExpression of(final String of) {
-        if (of == null) {
-            throw new IllegalArgumentException("string cannot be null");
-        }
+        Assertions.notNull("String", of);
```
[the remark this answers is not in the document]
katcharov: I used the type rather than the name, because `of` isn't very descriptive for the error message. I used `of` because it seems to prevent intellij label spam for a method that we are trying to make compact.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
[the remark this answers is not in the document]
katcharov: `abc = ofStringArray("a", "b", "c")`

Presently, they'd need to write `abc.reduce(of(""), (a, b) -> a.concat(b));`

It would be preferable to write `abc.concat()`, but that syntax is elusive, because we don't have extension methods as in C#. I discuss the problem and give an example here https://github.com/mongodb/mongo-java-driver/pull/1031#discussion_r1013433929 using "add" rather than "concat".

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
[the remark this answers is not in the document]
stIncMale: > of isn't very descriptive for the error message

We try to use this check consistently (this is also how it was before my time). The usage in this PR is not consistent with how we tend to use the check, which is a price we may consider to pay if there is enough value gained as a result. What value does using `String` bring?

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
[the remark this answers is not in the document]
stIncMale: Can we document this scenario (concatenating multiple strings) and the solution on the `concat` method?

driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
```diff
@@ -16,9 +16,32 @@
...
 /**
  * Expresses a string value.
  */
 public interface StringExpression extends Expression {
 
+    StringExpression toLower();
+
+    StringExpression toUpper();
+
+    StringExpression concat(StringExpression concat);
```
katcharov, line 30: This appends the string to the current string. Should we have a prepend method? If so, should this method be renamed?

stIncMale: When a question is posed this way, it forces others to think about it from scratch, then describe, pros/cons, potential usages, and suggest a decision. I think that all of this work should be done by the one who asks the question so that others could analyze the proposed options and express their opinion.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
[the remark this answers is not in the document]
katcharov: If documented on concat, it should be documented on sum, add, or (any), and (all), min, max, as well as all other reductive operations, of which there are many. It may be best to document this in some general location.

Let's defer this until writing docs. I have added a comment to the final expressions MR: https://github.com/mongodb/mongo-java-driver/pull/1032#discussion_r1021685532

driver-core/src/test/functional/com/mongodb/client/model/expressions/StringExpressionsFunctionalTest.java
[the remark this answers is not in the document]
katcharov: Alright, I'll leave this for now but I think perhaps we should give all the method names a final review for consistency. Worth noting that, as you pointed out, strLen does not match Java's length method. See: https://github.com/mongodb/mongo-java-driver/pull/1032#discussion_r1021694487

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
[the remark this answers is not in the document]
katcharov: "String cannot be null" is immediately clear, while "of cannot be null" sounds like a typo.

driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
katcharov: Here is an example where prepend (first line) is much easier to naturally write, compared to the equivalent append:
```
[a, b, c, d].map(v -> v.toString().trim().first(3).prepend(">"))
[a, b, c, d].map(v -> of(">").append(v.toString().trim().first(3)))
```
We discussed this and are leaning towards using prepend/append here, though the reductive operation would be concat (see https://github.com/mongodb/mongo-java-driver/pull/1032#discussion_r1021929731 ). Additional input welcome.

driver-core/src/test/functional/com/mongodb/client/model/expressions/StringExpressionsFunctionalTest.java
stIncMale: Discussed in Zoom.

stIncMale (approved):

PR #1037, opened 2022-11-04
Implement arithmetic expressions

katcharov:
JAVA-4803

Implements arithmetic expressions from the top 50, and a few others.

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -17,6 +17,7 @@
 package com.mongodb.client.model.expressions;
 
 import com.mongodb.assertions.Assertions;
+import com.mongodb.lang.NonNull;
 import org.bson.BsonArray;
 import org.bson.BsonBoolean;
 import org.bson.BsonDateTime;
@@ -35,6 +36,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import static com.mongodb.client.model.expressions.MqlExpression.AstPlaceholder;
 
 /**
  * Convenience methods related to {@link Expression}.
@@ -52,7 +54,7 @@ public final class Expressions {
      */
     public static BooleanExpression of(final boolean of) {
         // we intentionally disallow ofBoolean(null)
-        return new MqlExpression<>((codecRegistry) -> new BsonBoolean(of));
+        return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonBoolean(of)));
     }
 
     /**
@@ -63,16 +65,17 @@ public final class Expressions {
      * @return the integer expression
      */
     public static IntegerExpression of(final int of) {
-        return new MqlExpression<>((codecRegistry) -> new BsonInt32(of));
+        return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonInt32(of)));
     }
     public static IntegerExpression of(final long of) {
-        return new MqlExpression<>((codecRegistry) -> new BsonInt64(of));
+        return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonInt64(of)));
     }
     public static NumberExpression of(final double of) {
-        return new MqlExpression<>((codecRegistry) -> new BsonDouble(of));
+        return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonDouble(of)));
     }
-    public static DateExpression of(final Instant of) {
-        return new MqlExpression<>((codecRegistry) -> new BsonDateTime(of.toEpochMilli()));
+    public static DateExpression of(@NonNull final Instant of) {
+        Assertions.notNull("Instant", of);
+        return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonDateTime(of.toEpochMilli())));
     }
 
     /**
@@ -82,9 +85,9 @@ public final class Expressions {
      * @param of the string
      * @return the string expression
      */
-    public static StringExpression of(final String of) {
+    public static StringExpression of(@NonNull final String of) {
         Assertions.notNull("String", of);
-        return new MqlExpression<>((codecRegistry) -> new BsonString(of));
+        return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonString(of)));
     }
 
     /**
@@ -99,7 +102,7 @@ public final class Expressions {
         for (boolean b : array) {
             result.add(new BsonBoolean(b));
         }
-        return new MqlExpression<>((cr) -> new BsonArray(result));
+        return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonArray(result)));
     }
 
 
@@ -107,7 +110,7 @@ public final class Expressions {
         List<BsonValue> array = Arrays.stream(ofIntegerArray)
                 .mapToObj(BsonInt32::new)
                 .collect(Collectors.toList());
-        return new MqlExpression<>((cr) -> new BsonArray(array));
+        return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonArray(array)));
     }
 
     public static DocumentExpression ofDocument(final Bson document) {
@@ -115,11 +118,12 @@ public final class Expressions {
         // All documents are wrapped in a $literal. If we don't wrap, we need to
         // check for empty documents and documents that are actually expressions
         // (and need to be wrapped in $literal anyway). This would be brittle.
-        return new MqlExpression<>((cr) -> new BsonDocument("$literal",
-                document.toBsonDocument(BsonDocument.class, cr)));
+        return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonDocument("$literal",
+                document.toBsonDocument(BsonDocument.class, cr))));
     }
 
     public static <R extends Expression> R ofNull() {
-        return new MqlExpression<>((cr) -> new BsonNull()).assertImplementsAllExpressions();
+        return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonNull()))
+                .assertImplementsAllExpressions();
     }
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -136,4 +136,20 @@ public final class Expressions {
         return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonNull()))
                 .assertImplementsAllExpressions();
     }
+
+    static NumberExpression numberToExpression(final Number number) {
+        if (number instanceof Integer) {
+            return of((int) number);
+        } else if (number instanceof Long) {
+            return of((long) number);
+        } else if (number instanceof Double) {
+            return of((double) number);
+        } else if (number instanceof Decimal128) {
+            return of((Decimal128) number);
+        } else if (number instanceof BigDecimal) {
+            return of((BigDecimal) number);
+        } else {
+            throw new IllegalArgumentException("Number must be an Integer, Long, Double, Decimal128, or BigDecimal");
+        }
+    }
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -80,6 +80,10 @@ public final class Expressions {
         Assertions.notNull("BigDecimal", of);
         return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonDecimal128(new Decimal128(of))));
     }
+    public static NumberExpression of(final Decimal128 of) {
+        Assertions.notNull("Decimal128", of);
+        return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonDecimal128(of)));
+    }
     public static DateExpression of(@NonNull final Instant of) {
         Assertions.notNull("Instant", of);
         return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonDateTime(of.toEpochMilli())));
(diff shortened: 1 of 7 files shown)


driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
```diff
@@ -21,4 +21,39 @@
  */
 public interface NumberExpression extends Expression {
 
+    // TODO this must not return <T extends NumberExpression> T
+    // since it allows: IntegerEx result = oneNum.multiply(oneInt)
+    NumberExpression multiply(NumberExpression n);
```
katcharov, line 26: I previously suggested that these should return `<T extends NumberExpression> T`, but that allows us to write `IntegerEx result = oneNum.multiply(oneInt)`, which is incorrect.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArithmeticExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,210 @@
...
+                of(1).divide(of(2)),
+                "{'$divide': [1, 2]}");
+        // divide always returns a Number, so the method is not on IntegerExpression
+
+        // convenience
+        assertExpression(0.5, of(1.0).divide(2.0));
+        assertExpression(0.5, of(1).divide(2.0));
+        // TODO these convenience methods are all for doubles, so while they
+        // TODO allow longs and ints, they do not carry that through to the server:
+        assertExpression(0.5, of(1).divide(2L));
```
katcharov, line 94: I have created convenience methods only for doubles. If it is critical that the underlying structure, a long, is correctly propagated into the server, the user would need to use `of`.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -72,7 +76,12 @@ public static IntegerExpression of(final long of) {
     public static NumberExpression of(final double of) {
         return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonDouble(of)));
     }
+    public static NumberExpression of(final BigDecimal of) {
```
jyemin, line 79: BigDecimal is fine, but we need Decimal128 too, as not all Decimal128 values are representable as BigDecimal.

jyemin (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
stIncMale, line 79: Note that `BigDecimal` is not equivalent to `Decimal128` (the IEEE 754 quadruple-precision binary floating-point format). We need to support `Decimal128` because MongoDB can do computations with values of this type.

jyemin: +1

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArithmeticExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,233 @@
...
+                of(2.0).add(of(2)),
+                "{'$add': [2.0, 2]}");
+
+        // convenience
+        assertExpression(3.0, of(1.0).add(2.0));
+        assertExpression(3, of(1).add(2));
+
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/sum/
+        // sum's alternative behaviour exists for purposes of reduction, but is
+        // inconsistent with multiply, and potentially confusing. Unimplemented.
```
katcharov, line 139: Note above comment

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArithmeticExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,233 @@
...
+                of(3.231).divide(of(1.28)));
+        assertExpression(
+                Decimal128.parse("2.52421875"),
+                of(Decimal128.parse("3.231")).divide(of(Decimal128.parse("1.28"))));
+        assertExpression(
+                Decimal128.parse("2.52421875"),
+                of(Decimal128.parse("3.231")).divide(of(1.28)));
+        assertExpression(
+                Decimal128.parse("2.52421875"),
+                of(3.231).divide(of(Decimal128.parse("1.28"))));
```
stIncMale, line 110: This fails for me (MongoDB 6.0.2, Liberica JDK 17.0.3):
```
org.opentest4j.AssertionFailedError: 
Expected :BsonDecimal128{value=2.52421875}
Actual   :BsonDecimal128{value=2.524218750000}
```

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
```diff
@@ -20,5 +20,26 @@
  * Expresses an integer value.
  */
 public interface IntegerExpression extends NumberExpression {
```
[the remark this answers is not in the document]
katcharov: They are on integerExpression specifically because `intEx.add(1)` must not take a double, as `numEx` would allow. This explains why there needs to be at least one extra convenience method on IntEx.

If I understand it, the suggestion is effectively to add:

```
    default NumberExpression add(final int add) {
        return this.add(Expressions.of(add));
    }
```

to NumberExpression. But users can already pass in integers using `default NumberExpression add(final double add)`. Even putting that aside, why not for all the underlying numeric types? This is addressed by my https://github.com/mongodb/mongo-java-driver/pull/1037#discussion_r1017253457 ; to elaborate, the drawback would be that for every arithmetic method taking a parameter, we would need not 0-1 but 5 overloads (and docs).

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArithmeticExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,233 @@
...
+                of(2.5).round(),
+                "{'$round': 2.5} ");
+        assertExpression(
+                -6.0,
+                of(-5.5).round());
+        // to place
+        assertExpression(
+                555.55,
+                of(555.555).round(of(2)),
+                "{'$round': [555.555, 2]} ");
```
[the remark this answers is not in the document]
katcharov: No harm in filing a server bug.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArithmeticExpressionsFunctionalTest.java
katcharov: Failed for me as well. Updated.

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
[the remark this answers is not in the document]
stIncMale: I understand why we have `add(IntegerExpression)` on `NumberExpression` (so that we have `add` that returns `IntegerExpression`). As a result, it makes sense to introduce the `IntegerExpression.add(int)` convenience method. That's clear. I also understand that you don't want to have convenience methods for all types, and you picked a single most useful type `double`<sup>1</sup> and introduced a convenience method for this type.

> why not for all the underlying numeric types?

Because the convenience method `add(int)` was introduced on `IntegerExpression` anyway, and the same is not true for other types.

> users can already pass in integers using default NumberExpression add(final double add)

Yes, they can. For this fact to work as an argument against having `add(int)`, any `int` value must be representable perfectly as `double`, otherwise there would have been still room for users to use `NumberExpression.add(of(int))`. And indeed, any `int` value can be represented perfectly as `double`, which is something I did not initially realize.

---

<sup>1</sup> `double` can't perfectly represent all `long`/`Decimal128` values because it has only 52 bits for representing mantissa.

stIncMale (approved):

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
```diff
@@ -21,4 +21,37 @@
  */
 public interface NumberExpression extends Expression {
 
+    NumberExpression multiply(NumberExpression n);
+
+    default NumberExpression multiply(final double multiply) {
```
katcharov, line 26: @jyemin raised an issue with `double` causing unexpected promotion of underlying BSON type causing a ClassCast during:

```
        long d1 = (long) result.get("d");
        long d2 = result.getLong("d");
        long d3 = result.get("d", Long.class);
```

Fixed in:

https://github.com/mongodb/mongo-java-driver/pull/1052/files#diff-276193793ed4e0294e212d9927de707b2d000804cbf649f2a0d3888e2ec9adf6L26

katcharov: I pulled this in to this PR

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
```diff
@@ -21,4 +21,37 @@
...
+        return this.add(Expressions.numberToExpression(add));
+    }
+
+    NumberExpression subtract(NumberExpression n);
+
+    default NumberExpression subtract(final Number subtract) {
+        return this.subtract(Expressions.numberToExpression(subtract));
+    }
+
+    NumberExpression max(NumberExpression n);
```
jyemin, line 48: Why don't `min` and `max` have `Number` overloads?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
jyemin: I'm having second thoughts about the inclusion of this method.  Given that Decimal128 use is fairly specialized, forcing the application to call `Decimal128.of(BigDecimal)` doesn't seem like too much of an imposition, and makes it clear that we don't actually support all BigDecimal instances, but only the ones that can be represented as Decimal128.

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
katcharov: Uncertain. It seems that, if an overload is useful at all, it's useful on multiply. But then there's `or(BoolEx)` where an overload seems inappropriate. I'm not sure where to draw the line formally - to me it seemed odd to `max` on a known number.

When I thought about an overload on `max`, it seemed that `max` should be renamed to `lowerBound` (that is, if your n is under 5, `n.lowerBound(5)` would return 5, and n otherwise).

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: Agreed

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -136,4 +136,20 @@ public static <R extends Expression> R ofNull() {
         return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonNull()))
                 .assertImplementsAllExpressions();
     }
+
+    static NumberExpression numberToExpression(final Number number) {
```
stIncMale, line 135: I don't think something prevents us from naming this method `of`. Am I missing something?

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -136,4 +136,20 @@ public static <R extends Expression> R ofNull() {
...
+            return of((long) number);
+        } else if (number instanceof Double) {
+            return of((double) number);
+        } else if (number instanceof Decimal128) {
+            return of((Decimal128) number);
+        } else if (number instanceof BigDecimal) {
+            return of((BigDecimal) number);
+        } else {
+            throw new IllegalArgumentException("Number must be an Integer, Long, Double, Decimal128, or BigDecimal");
+        }
```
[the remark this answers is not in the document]
katcharov: I disagree that the API is intended to use codecs. There is a codec that is supposed to convert expressions to BSON, so the codecs depend on this exp API existing, but the exp API does not depend on any codecs.

We are also leaving open the possibility of moving to an AST, the point of which is to decouple expressions from the BSON/MQL, and this would seem to go in the opposite direction.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: I just thought the name was more descriptive, since there was no constraint (as with `of`) of being as short as possible.

stIncMale: I have realized that this method is not part of public API. If I noticed that right away, I would not have asked about renaming.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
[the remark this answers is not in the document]
stIncMale: It's not a public method, so the thing with using codecs is as irrelevant as the thing with renaming. However, it is still useful to talk about the usage of `CodecRegistry` in `MqlExpression`.

> There is a codec that is supposed to convert expressions to BSON, so the codecs depend on this exp API existing, but the exp API does not depend on any codecs.

Currently, `MqlExpression.toBsonValue` (which is how `MqlExpressionCodec` encodes `MqlExpression` to BSON) passes `CodecRegistry` to `MqlExpression.fn`, yet as far as I understand, there are no places where this `CodecRegistry` is used. What is the planned use for the `CodecRegistry` that is passed to `MqlExpression.fn`?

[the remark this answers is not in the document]
stIncMale: I added a comment there some time ago: https://github.com/mongodb/mongo-java-driver/pull/1032#discussion_r1013283977. Thanks for the `DocumentExpression`!

stIncMale (approved):

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
jyemin: ok, we can always add more overloads in the future.

jyemin (approved):

PR #1039, opened 2022-11-07
Update encryption terminology

jyemin:
JAVA-4749

I didn't find any other obvious places to make the requested change.


rozza (commented):
Do a search on `Client side` and it turns up a few results.  Looks like `package.scala` needs updating and possibly the `@mongodb.url.manual` urls and descriptions as well.

rozza (approved):

PR #1040, opened 2022-11-07
Update MongoQueryException to extend MongoCommandException

jyemin:
The MongoQueryException class currently extends MongoServerException,
the reason being that prior to MongoDB 3.2 the find helper was implemented
using the OP_QUERY wire protocol message, which is not a command.  But now
that 3.6 is the minimum required server version, the find helper is always
implemented via the find command.  So MongoQueryException can now extend
MongoCommandException and include the full command response document.

JAVA-4676

### DIFF
--- a/driver-core/src/main/com/mongodb/MongoCommandException.java
+++ b/driver-core/src/main/com/mongodb/MongoCommandException.java
@@ -83,9 +83,10 @@ public class MongoCommandException extends MongoServerException {
     }
 
     /**
-     * For internal use only.
+     * Gets the full server response document describing the error.
      *
      * @return the full response to the command failure.
+     * @since 4.8
      */
     public BsonDocument getResponse() {
         return response;
--- a/driver-core/src/main/com/mongodb/MongoCursorNotFoundException.java
+++ b/driver-core/src/main/com/mongodb/MongoCursorNotFoundException.java
@@ -16,6 +16,8 @@
 
 package com.mongodb;
 
+import org.bson.BsonDocument;
+
 /**
  * Subclass of {@link MongoException} representing a cursor-not-found exception.
  *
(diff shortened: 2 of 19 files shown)


driver-core/src/main/com/mongodb/MongoCommandException.java
```diff
@@ -83,9 +83,10 @@ public String getErrorMessage() {
     }
 
     /**
-     * For internal use only.
+     * Gets the full server response document describing the error.
      *
      * @return the full response to the command failure.
+     * @since 4.8
```
jyemin, line 89: Treating this as an API addition based on the javadoc

driver-core/src/main/com/mongodb/MongoCursorNotFoundException.java
```diff
@@ -49,13 +64,4 @@ public MongoCursorNotFoundException(final long cursorId, final ServerAddress ser
     public long getCursorId() {
         return cursorId;
     }
-
-    /**
-     * The server address where the cursor is.
-     *
-     * @return the ServerAddress representing the server the cursor was on.
-     */
-    public ServerAddress getServerAddress() {
```
jyemin, line 58: Not binary breaking since the method is available on the superclass

driver-core/src/main/com/mongodb/MongoQueryException.java
```diff
@@ -51,43 +62,33 @@ public MongoQueryException(final ServerAddress address, final int errorCode, fin
...
-        addLabels(commandException.getErrorLabels());
+        this(commandException.getResponse(), commandException.getServerAddress());
     }
 
-    /**
-     * Gets the error code for this query failure.
-     *
-     * @return the error code
-     */
-    public int getErrorCode() {
```
jyemin, line 81: Not binary breaking since the methods are available on the superclass

driver-core/src/main/com/mongodb/MongoQueryException.java
```diff
@@ -51,43 +62,33 @@ public MongoQueryException(final ServerAddress address, final int errorCode, fin
...
-    }
-
-    /**
-     * Gets the error message for this query failure.
-     *
-     * @return the error message
-     */
-    public String getErrorMessage() {
-        return errorMessage;
+    private static BsonDocument manufactureResponse(final int errorCode, @Nullable final String errorCodeName, final String errorMessage) {
```
jyemin, line 85: Necessary to manufacture the response document to support the deprecated constructors.  In practice this will go unused since the driver doesn't use the deprecated constructors, but application tests may do so.

rozza (changes requested):
One nit but after LGTM

PR #1042, opened 2022-11-09
Expressions type safety experiment

jyemin:
In this Draft PR I attempted to come up with a way to ensure that the expression instance returned from any of the methods can not be improperly casted by the application to some other expression type, as it can now do the fact that MqlExpression implements all expression subtypes.

The results of the experiment were not particularly satisfying.  Both `cond` and `reduce`, in particular, caused problems, because in both, the type of the returned expression is based on the parameterized type of one or more of the parameters.  So the only solution I could come up with was to do `instanceof` checks on the parameters to determine what wrapping class to return an instance of.

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
@@ -29,6 +29,84 @@ final class MqlExpression<T extends Expression>
         implements Expression, BooleanExpression, IntegerExpression, NumberExpression,
         StringExpression, DateExpression, DocumentExpression, ArrayExpression<T> {
 
+    static class MqlWrappingExpression<T extends Expression> implements Expression {
+        private final MqlExpression<T> wrapped;
+
+        private MqlWrappingExpression(final MqlExpression<T> wrapped) {
+            this.wrapped = wrapped;
+        }
+
+        public MqlExpression<T> getWrapped() {
+            return wrapped;
+        }
+    }
+
+    static class MqlBooleanWrappingExpression<T extends Expression> extends MqlWrappingExpression<T> implements BooleanExpression {
+        MqlBooleanWrappingExpression(final MqlExpression<T> wrapped) {
+            super(wrapped);
+        }
+
+        @Override
+        public BooleanExpression not() {
+            return getWrapped().not();
+        }
+
+        @Override
+        public BooleanExpression or(final BooleanExpression or) {
+            return getWrapped().or(or);
+        }
+
+        @Override
+        public BooleanExpression and(final BooleanExpression and) {
+            return getWrapped().and(and);
+        }
+
+        @Override
+        public <R extends Expression> R cond(final R left, final R right) {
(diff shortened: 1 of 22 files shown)


driver-core/src/test/functional/com/mongodb/client/model/expressions/ArrayExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,91 @@
...
+    @Test
+    public void mapTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/map/
+        assertExpression(
+                Stream.of(true, true, false)
+                        .map(v -> !v).collect(Collectors.toList()),
+                arrayTTF.map(v -> v.not()),
+                // MQL:
+                "{'$map': {'input': [true, true, false], 'in': {'$not': '$$this'}}}");
+    }
```
katcharov, line 60: I added the following test, which should fail with a ClassCast:

```
        assertExpression(
                Arrays.asList(),
                ofBooleanArray().map((BooleanExpression q) -> {
                    IntegerExpression i = (IntegerExpression) q; // <-- passes
                    return q.not();
                }),
                "{'$map': {'input': [], 'in': {'$not': '$$this'}}}");
```

I don't know how this would be implemented for the general case. It seems that to know the "inner" type for `q` (so that it can be wrapped, to ensure a ClassCast), we need a value, but the array is empty, so there's no value to use. In this example we could infer the value because the constructor is ofBooleanArray. But the general `ofArray<T>` seems like it would need a `BooleanExpression.class` parameter that we could store and switch on. But then I don't know where we could put `BoolEx.class` for something like `ArrayEx<ArrayEx<BoolEx>>`.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -0,0 +1,284 @@
...
+import org.bson.codecs.configuration.CodecRegistry;
+
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+
+final class MqlExpression<T extends Expression>
+        implements Expression, BooleanExpression, IntegerExpression, NumberExpression,
+        StringExpression, DateExpression, DocumentExpression, ArrayExpression<T> {
+
+    static class MqlWrappingExpression<T extends Expression> implements Expression {
```
katcharov, line 32: I believe we could have `final class MqlExpression<T extends Expression> implements Expression {`, and move all the implements to respective subclasses, like `MqlBooleanExpression`. This would eliminate redundant code and also help break up the MqlExpression monolith. (Not that I think this needs to be done for the sake of this POC.)

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArrayExpressionsFunctionalTest.java
jyemin: I'm not sure.  Would have to think more.  But it reminds me of some of the recent work on parameterization of records.

PR #1045, opened 2022-11-14
Implement date expressions

katcharov:
JAVA-4804

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
@@ -31,7 +31,7 @@ public interface DateExpression extends Expression {
     IntegerExpression week(StringExpression timezone);
     IntegerExpression millisecond(StringExpression timezone);
 
-    StringExpression dateToString(StringExpression timezone);
+    StringExpression dateToString();
     StringExpression dateToString(StringExpression timezone, StringExpression format);
 
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
@@ -20,18 +20,18 @@ package com.mongodb.client.model.expressions;
  * Expresses a date value.
  */
 public interface DateExpression extends Expression {
-    IntegerExpression year();
-    IntegerExpression month();
-    IntegerExpression dayOfMonth();
-    IntegerExpression dayOfWeek();
-    IntegerExpression dayOfYear();
-    IntegerExpression hour();
-    IntegerExpression minute();
-    IntegerExpression second();
-    IntegerExpression week();
-    IntegerExpression millisecond();
+    IntegerExpression year(StringExpression timezone);
+    IntegerExpression month(StringExpression timezone);
+    IntegerExpression dayOfMonth(StringExpression timezone);
+    IntegerExpression dayOfWeek(StringExpression timezone);
+    IntegerExpression dayOfYear(StringExpression timezone);
+    IntegerExpression hour(StringExpression timezone);
+    IntegerExpression minute(StringExpression timezone);
+    IntegerExpression second(StringExpression timezone);
+    IntegerExpression week(StringExpression timezone);
+    IntegerExpression millisecond(StringExpression timezone);
 
-    StringExpression dateToString();
-    StringExpression dateToString(StringExpression format, StringExpression timezone);
+    StringExpression dateToString(StringExpression timezone);
+    StringExpression dateToString(StringExpression timezone, StringExpression format);
 
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
@@ -20,5 +20,18 @@ package com.mongodb.client.model.expressions;
  * Expresses a date value.
  */
 public interface DateExpression extends Expression {
+    IntegerExpression year();
+    IntegerExpression month();
+    IntegerExpression dayOfMonth();
+    IntegerExpression dayOfWeek();
+    IntegerExpression dayOfYear();
+    IntegerExpression hour();
+    IntegerExpression minute();
+    IntegerExpression second();
+    IntegerExpression week();
+    IntegerExpression millisecond();
+
+    StringExpression dateToString();
+    StringExpression dateToString(StringExpression format, StringExpression timezone);
 
 }
--- a/driver-core/src/test/functional/com/mongodb/client/model/expressions/DateExpressionsFunctionalTest.java
+++ b/driver-core/src/test/functional/com/mongodb/client/model/expressions/DateExpressionsFunctionalTest.java
@@ -31,12 +31,13 @@ class DateExpressionsFunctionalTest extends AbstractExpressionsFunctionalTest {
     // https://www.mongodb.com/docs/manual/reference/operator/aggregation/#date-expression-operators
 
     private final ZonedDateTime utcDateTime = ZonedDateTime.ofInstant(Instant.parse("2007-12-03T10:15:30.005Z"), ZoneId.of(ZoneOffset.UTC.getId()));
+    private final StringExpression utc = of("UTC");
+    private final DateExpression utcDateEx = of(utcDateTime.toInstant());
 
     @Test
     public void literalsTest() {
         assertExpression(
-                utcDateTime.toInstant(),
-                of(utcDateTime.toInstant()),
+                utcDateTime.toInstant(), utcDateEx,
                 "{'$date': '2007-12-03T10:15:30.005Z'}");
     }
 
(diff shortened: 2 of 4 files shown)


rozza (approved):

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
```diff
@@ -20,5 +20,18 @@
...
+    IntegerExpression dayOfWeek();
+    IntegerExpression dayOfYear();
+    IntegerExpression hour();
+    IntegerExpression minute();
+    IntegerExpression second();
+    IntegerExpression week();
+    IntegerExpression millisecond();
+
+    StringExpression dateToString();
+    StringExpression dateToString(StringExpression format, StringExpression timezone);
```
jyemin, line 35: I don't recall if we've had a discussion about how to handle expressions with multiple optional parameters.  In this case both format and timezone are independently optional and it seems necessary to support that in the API.

jyemin, line 35: Thoughts on the `onNull` expression supported by $dateToString?  Are you thinking to require `ifNull`?

driver-core/src/test/functional/com/mongodb/client/model/expressions/DateExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+package com.mongodb.client.model.expressions;
+
+import org.junit.jupiter.api.Test;
+
+import java.time.Instant;
+
+import static com.mongodb.client.model.expressions.Expressions.of;
+
+class DateExpressionsFunctionalTest extends AbstractExpressionsFunctionalTest {
```
stIncMale, line 25: We can use the Java SE API to compute expected values: https://github.com/stIncMale/mongo-java-driver/commit/e087e7fb9a334cb176024e9abcb0044b23d6fb74.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DateExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Test
+    public void weekTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/week/
+        assertExpression(
+                48,
+                of(instant).week(),
+                "{'$week': {'$date': '2007-12-03T10:15:30.005Z'}}");
+    }
```
stIncMale, line 117: This is the only test where the expected value can't be trivially computed with the Java SE API.

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
```diff
@@ -20,5 +20,18 @@
  * Expresses a date value.
  */
 public interface DateExpression extends Expression {
```
stIncMale, line 22: Discussed in Zoom that all `DateExpression` have to accept a timezone, similarly to the corresponding MQL expressions. This is a consequence of BSON not having a type that represents a pair of an instant and a timezone, so the timezone has to come from somewhere when computing `$hour` and the rest of date expressions.

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
katcharov: That server option presupposes that a DateExpression can have a value of null, which our API forbids. A user who has an Expression of an unknown type (could be null, could be date) would write:
```
expression.isDateOr(of(defaultInstance)).dateToString(...)
```
The `exp.isDateOr(deflt)` (and `doc.getDate("fieldName", deflt)`) methods are intended to be guards against nulls, and will arrive in the ensuing two PRs. I am unsure whether `ifNull` will even be necessary/coherent in our API.

(Incidentally, this is similar to the approach with limit on filter.)

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
katcharov: Alternatives discussed included:

1. the use of a non-expression type, `date.withZone(zone).year()`, where year uses the zone contained in what withZone returns. The problem is that whatever withZone returns would not be an expression, which violates a design constraint of the API (expressions always chain into other expressions), for almost no benefit.
2. `withZone` could return a `ZonedDateExpression`, which would be backed by a document `{date: bsonDate, zone: bsonString}`, but then this would need to be stored on the server, which is too unexpected.
3. when aggregating on a collection, a default timezone could be specified, which would make `.year()` return that timezone rather than always UTC, but we would need the overloads that take timezones regardless, since the user will not always use only the default.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DateExpressionsFunctionalTest.java
katcharov: cherrypicked

driver-core/src/test/functional/com/mongodb/client/model/expressions/DateExpressionsFunctionalTest.java
```diff
@@ -31,12 +31,13 @@ class DateExpressionsFunctionalTest extends AbstractExpressionsFunctionalTest {
     // https://www.mongodb.com/docs/manual/reference/operator/aggregation/#date-expression-operators
 
     private final ZonedDateTime utcDateTime = ZonedDateTime.ofInstant(Instant.parse("2007-12-03T10:15:30.005Z"), ZoneId.of(ZoneOffset.UTC.getId()));
+    private final StringExpression utc = of("UTC");
+    private final DateExpression utcDateEx = of(utcDateTime.toInstant());
```
stIncMale, line 35: We shouldn't call it `utcDateEx` because (unlike Java's `utcDateTime`) it's time zone–agnostic (it's semantically an instant). Let's call it `dateEx`?

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
stIncMale: Just so that we don't forget: `dateToString(StringExpression timezone)` propagates the incorrect server's default format `%Y-%m-%dT%H:%M:%S.%LZ` to our API. We discussed, and since the server can't be fixed without breaking compatibility, we should avoid exposing its default format either by using a different default format (e.g., `%Y-%m-%dT%H:%M:%S.%L%z`, note `%z` instead of `Z`), or by forcing a user to always specify the format. The former may be surprising as a user likely expects that the default used by the driver is the same as that used by the server.

stIncMale (approved):

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
jyemin: Ping on this.

katcharov: We appear to be spared in this particular case, because omitting one of the options would be confused or at least confusing, see: https://github.com/mongodb/mongo-java-driver/pull/1045#discussion_r1028655799 . That is, they should be mandatory, even if the server docs disagree.

My original thought (before @stIncMale pointed out the above) was to err on the side of being more explicit, especially in cases where the method is infrequent, and where implicit values might sometimes lead to confusion. Of course, this might not always work or make sense, but might help in other cases as well. It is also possible to sometimes get rid of optionals by moving them out (as was the case with limit on filter). So the general approach is to try to get rid of them in some way or another. This seems to lead to a cleaner API.

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
```diff
@@ -20,5 +20,18 @@
...
+    IntegerExpression year(StringExpression timezone);
```
jyemin, line 23: Especially given the `dateToString()` overload below, what's the rationale for not having similar no-arg overloads for all of these?  I can think of two:

1. reduced API surface
2. a conclusion that the default of "UTC" is not useful in practice

Or perhaps it's an oversight

stIncMale: `dateToString()` gives a string representation of the date in UTC, which is also a string representation of an instant, because BSON date is actually an instant contrary to its name. I.e., `dateToString()` is equivalent to Java's `Instant.toString`, which is useful and can (and should) be thought of without thinking about timezones.

`year()` on the other hand, makes sense only within a specific timezone, but that may not be obvious to a user. To make it obvious, this method always requires a timezone.

jyemin: That's all true, but does it mean that UTC is not a credible default?  We should assume that query team had a good reason for the default unless we hear otherwise.

stIncMale: Given the default server format, which is incorrect in all situations but when the timezone is UTC, and that the Java SE API does not have a similar "UTC is the default timezone when someone wants to extract `year`" (instead, the Java SE API forces one to convert `Instant` to `ZonedDateTime` via `ZonedDateTime.ofInstant(Instant instant, ZoneId zone)`, and then allows one to extract `ChronoUnit.YEARS`), it is not obvious that the query team had this API thought through.

katcharov: This also makes it clear that we are not storing a date-with-timezone. I think forcing the user to be explicit about this will prevent mistakes.

Also, overloads would be easy to add later if we're making a mistake, but the alternative would be difficult to remove.

jyemin (approved):

PR #1051, opened 2022-11-18
Fix `MixedBulkWriteOperation` such that it does not leak `MongoWriteConcernWithResponseException` to users

stIncMale:
JAVA-4775

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
@@ -180,32 +180,22 @@ public class MixedBulkWriteOperation implements AsyncWriteOperation<BulkWriteRes
             logRetryExecute(retryState);
             return withSourceAndConnection(binding::getWriteConnectionSource, true, (source, connection) -> {
                 ConnectionDescription connectionDescription = connection.getDescription();
-                int maxWireVersion = connectionDescription.getMaxWireVersion();
                 // attach `maxWireVersion` ASAP because it is used to check whether we can retry
-                retryState.attach(AttachmentKeys.maxWireVersion(), maxWireVersion, true);
-                BulkWriteTracker bulkWriteTracker = retryState.attachment(AttachmentKeys.bulkWriteTracker())
-                        .orElseThrow(Assertions::fail);
+                retryState.attach(AttachmentKeys.maxWireVersion(), connectionDescription.getMaxWireVersion(), true);
                 SessionContext sessionContext = binding.getSessionContext();
                 WriteConcern writeConcern = getAppliedWriteConcern(sessionContext);
-                if (!retryState.isFirstAttempt() && !isRetryableWrite(retryWrites, writeConcern, source.getServerDescription(),
-                        connectionDescription, sessionContext)) {
-                    RuntimeException prospectiveFailedResult = (RuntimeException) retryState.exception().orElse(null);
-                    retryState.breakAndThrowIfRetryAnd(() -> !(prospectiveFailedResult instanceof MongoWriteConcernWithResponseException));
-                    bulkWriteTracker.batch().ifPresent(bulkWriteBatch -> {
-                        assertTrue(prospectiveFailedResult instanceof MongoWriteConcernWithResponseException);
-                        bulkWriteBatch.addResult((BsonDocument) ((MongoWriteConcernWithResponseException) prospectiveFailedResult)
-                                .getResponse());
-                        BulkWriteTracker.attachNext(retryState, bulkWriteBatch);
-                    });
+                if (!isRetryableWrite(retryWrites, getAppliedWriteConcern(sessionContext),
+                        source.getServerDescription(), connectionDescription, sessionContext)) {
+                    handleMongoWriteConcernWithResponseException(retryState, true);
                 }
                 validateWriteRequests(connectionDescription, bypassDocumentValidation, writeRequests, writeConcern);
-                if (!bulkWriteTracker.batch().isPresent()) {
+                if (!retryState.attachment(AttachmentKeys.bulkWriteTracker()).orElseThrow(Assertions::fail).batch().isPresent()) {
                     BulkWriteTracker.attachNew(retryState, BulkWriteBatch.createBulkWriteBatch(namespace,
                             source.getServerDescription(), connectionDescription, ordered, writeConcern,
                             bypassDocumentValidation, retryWrites, writeRequests, sessionContext, comment, variables));
                 }
                 logRetryExecute(retryState);
-                return executeBulkWriteBatch(retryState, binding, connection, maxWireVersion);
+                return executeBulkWriteBatch(retryState, binding, connection);
             });
         });
         try {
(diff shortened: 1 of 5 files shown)


driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -233,7 +234,9 @@ public void commandSucceeded(final CommandSucceededEvent event) {
                         .append("writeConcernError", new BsonDocument()
                                 .append("code", new BsonInt32(91))
                                 .append("errorLabels", new BsonArray(Stream.of("RetryableWriteError")
-                                        .map(BsonString::new).collect(Collectors.toList()))))
+                                        .map(BsonString::new).collect(Collectors.toList())))
+                                .append("errmsg", new BsonString(""))
```
stIncMale, line 238: The driver needs this field when it creates `MongoWriteConcernException`.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -246,13 +249,14 @@ public void commandSucceeded(final CommandSucceededEvent event) {
...
-            }
-            fail("must not reach");
+            assertThrows(MongoWriteConcernException.class, () -> {
+                try {
+                    collection.insertOne(new Document());
+                } catch (MongoWriteConcernException e) {
+                    assertEquals(91, e.getCode());
+                    throw e;
+                }
+            });
```
stIncMale, line 259: Now that the leak is fixed, we can expect `MongoWriteConcernException` here.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -180,32 +180,22 @@ public BulkWriteResult execute(final WriteBinding binding) {
...
-                if (!retryState.isFirstAttempt() && !isRetryableWrite(retryWrites, writeConcern, source.getServerDescription(),
-                        connectionDescription, sessionContext)) {
-                    RuntimeException prospectiveFailedResult = (RuntimeException) retryState.exception().orElse(null);
-                    retryState.breakAndThrowIfRetryAnd(() -> !(prospectiveFailedResult instanceof MongoWriteConcernWithResponseException));
-                    bulkWriteTracker.batch().ifPresent(bulkWriteBatch -> {
-                        assertTrue(prospectiveFailedResult instanceof MongoWriteConcernWithResponseException);
-                        bulkWriteBatch.addResult((BsonDocument) ((MongoWriteConcernWithResponseException) prospectiveFailedResult)
-                                .getResponse());
-                        BulkWriteTracker.attachNext(retryState, bulkWriteBatch);
-                    });
```
stIncMale, line 199: This `if` (except for the `!isRetryableWrite` condition) was extracted to `handleMongoWriteConcernWithResponseException`, and then the new method was used in one more place, such that now it's called both before and after an attempt, as opposed to be called only before. That's what the fix is about.

So now `handleMongoWriteConcernWithResponseException` is called both before and after an `executeBulkWriteBatch` attempt (this is also how it was before the retrying mechanism was refactored). Maybe we can refactor `MixedBulkWriteOperation` such that there is only one call to `handleMongoWriteConcernWithResponseException`, but I did not see how.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -292,9 +271,10 @@ private BulkWriteResult executeBulkWriteBatch(final RetryState retryState, final
                 currentBulkWriteTracker = BulkWriteTracker.attachNext(retryState, currentBatch);
                 currentBatch = currentBulkWriteTracker.batch().orElseThrow(Assertions::fail);
             } catch (MongoException exception) {
-                if (!(retryState.isFirstAttempt() || (exception instanceof MongoWriteConcernWithResponseException))) {
+                if (!retryState.isFirstAttempt() && !(exception instanceof MongoWriteConcernWithResponseException)) {
```
stIncMale, line 274: Changed this because I found the new form easier to understand (I was trying to understand what's going on, not very successfully, though).

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -262,17 +241,17 @@ public void executeAsync(final AsyncWriteBinding binding, final SingleResultCall
...
         retryingBulkWrite.get(exceptionTransformingCallback(errorHandlingCallback(callback, LOGGER)));
     }
 
-    private BulkWriteResult executeBulkWriteBatch(final RetryState retryState, final WriteBinding binding, final Connection connection,
-            final int maxWireVersion) {
+    private BulkWriteResult executeBulkWriteBatch(final RetryState retryState, final WriteBinding binding, final Connection connection) {
         BulkWriteTracker currentBulkWriteTracker = retryState.attachment(AttachmentKeys.bulkWriteTracker())
                 .orElseThrow(Assertions::fail);
         BulkWriteBatch currentBatch = currentBulkWriteTracker.batch().orElseThrow(Assertions::fail);
+        int maxWireVersion = connection.getDescription().getMaxWireVersion();
```
stIncMale, line 254: Removed the redundancy: there is no need to pass `maxWireVersion` when we can get it from `connection`.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
rozza: FYI this can also be written as:

```
MongoWriteConcernException e = assertThrows(MongoWriteConcernException.class, () ->  collection.insertOne(new Document()));
assertEquals(91, e.getCode());
```

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
rozza: Should there be a separate catch for `MongoWriteConcernWithResponseException` to make this easier to understand?

driver-sync/src/test/functional/com/mongodb/client/MongoWriteConcernWithResponseExceptionTest.java
```diff
@@ -0,0 +1,144 @@
...
+import static com.mongodb.client.Fixture.getDefaultDatabaseName;
+import static com.mongodb.client.Fixture.getMongoClientSettingsBuilder;
+import static java.util.Collections.singletonList;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assume.assumeTrue;
+
+/**
+ * Tests in this class check that the internal {@link MongoWriteConcernWithResponseException} does not leak from our API.
+ */
+public final class MongoWriteConcernWithResponseExceptionTest {
```
rozza, line 53: Should this follow the pattern that the other sync tests work by being overrideable / extendable and having a `createMongoClient` method?  

Would save adding another way of doing similar things.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
rozza: Ack

rozza (approved):
Looks good - added a couple of comments for consideration only

jyemin (changes requested):

driver-sync/src/test/functional/com/mongodb/client/MongoWriteConcernWithResponseExceptionTest.java
```diff
@@ -0,0 +1,144 @@
...
+        try (MongoClient client = clientCreator.apply(getMongoClientSettingsBuilder()
+                .retryWrites(true)
+                .addCommandListener(commandListener)
+                .applyToServerSettings(builder -> builder.heartbeatFrequency(50, TimeUnit.MILLISECONDS))
+                .build());
+             FailPoint ignored = FailPoint.enable(firstAttemptFpDoc, primaryServerAddress)) {
+            MongoCollection<Document> collection = client.getDatabase(getDefaultDatabaseName())
+                    .getCollection("originalErrorMustBePropagatedIfNoWritesPerformed");
+            collection.drop();
+            assertThrows(MongoWriteConcernException.class, () -> {
```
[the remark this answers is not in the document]
stIncMale: `assertThrows(MongoWriteConcernException.class, ...)` verifies that `MongoWriteConcernException` is thrown. What if a different exception is thrown? - it simply reports that. I want to have the `The internal exception leaked` message when the different exception is `MongoWriteConcernWithResponseException`, and I don't see another way of achieving this. Of course, if this happens, both assertions will fail with the inner assertion being the cause of the outer one.

driver-sync/src/test/functional/com/mongodb/client/MongoWriteConcernWithResponseExceptionTest.java
stIncMale: I followed the approach used in `RetryableWritesProseTest`. All we need here is to reuse code. Class inheritance allows doing so, but, much more importantly, it also allows to  have subtype polymorphism. To me, using inheritance solely for the purpose of code reuse (as is the case with tests) is weird, and I tend to avoid doing so.

The approach with inheritance was also not possible with `RetryableWritesProseTest` without refactoring the existing code structure: versions of `RetryableWritesProseTest` in sync/reactive modules already extend different versions of `DatabaseTestCase`.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: Thank you, I have not realized `assertThrows` returns the thrown exception :) Done.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
stIncMale: Currently we have
```java
} catch (MongoException exception) {
    if (!retryState.isFirstAttempt() && !(exception instanceof MongoWriteConcernWithResponseException)) {
        addRetryableWriteErrorLabel(exception, maxWireVersion);
    }
    handleMongoWriteConcernWithResponseException(retryState, false);
    throw exception;
}
```
with the change you propose we will have
```java
} catch (MongoWriteConcernWithResponseException exception) {
    handleMongoWriteConcernWithResponseException(retryState, false);
    throw exception;
} catch (MongoException exception) {
    if (!retryState.isFirstAttempt()) {
        addRetryableWriteErrorLabel(exception, maxWireVersion);
    }
    handleMongoWriteConcernWithResponseException(retryState, false);
    throw exception;
}
```

The current code is shorter and does not have code duplication, the proposed code does not have the `instanceof` check. To me the current approach seems the lesser of two evils.

driver-sync/src/test/functional/com/mongodb/client/MongoWriteConcernWithResponseExceptionTest.java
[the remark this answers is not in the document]
jyemin: It seems overly complicated structure just to get a more helpful failure message, but I can live with it.

jyemin (approved):

PR #1052, opened 2022-11-18
Implement document expressions

katcharov:
JAVA-4782

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -28,54 +28,65 @@ import static com.mongodb.client.model.expressions.Expressions.of;
  */
 public interface DocumentExpression extends Expression {
 
-    <R extends DocumentExpression> R setField(String path, Expression exp);
+    DocumentExpression setField(String path, Expression exp);
 
     DocumentExpression unsetField(String path);
 
 
     BooleanExpression getBoolean(String field);
 
-    BooleanExpression getBoolean(String field, BooleanExpression orElse);
+    BooleanExpression getBoolean(String field, BooleanExpression other);
 
-    default BooleanExpression getBoolean(final String field, final boolean orElse) {
-        return getBoolean(field, of(orElse));
+    default BooleanExpression getBoolean(final String field, final boolean other) {
+        return getBoolean(field, of(other));
     }
 
     NumberExpression getNumber(String field);
 
-    NumberExpression getNumber(String field, NumberExpression orElse);
+    NumberExpression getNumber(String field, NumberExpression other);
 
-    default NumberExpression getNumber(final String field, final Number orElse) {
-        return getNumber(field, Expressions.numberToExpression(orElse));
+    default NumberExpression getNumber(final String field, final Number other) {
+        return getNumber(field, Expressions.numberToExpression(other));
     }
 
     IntegerExpression getInteger(String field);
 
+    IntegerExpression getInteger(String field, IntegerExpression other);
+
+    default NumberExpression getInteger(final String field, final int other) {
+        return getInteger(field, (IntegerExpression) Expressions.numberToExpression(other));
+    }
+
+    default NumberExpression getInteger(final String field, final long other) {
+        return getInteger(field, (IntegerExpression) Expressions.numberToExpression(other));
+    }
+
+
     StringExpression getString(String field);
 
-    StringExpression getString(String field, StringExpression orElse);
+    StringExpression getString(String field, StringExpression other);
 
-    default StringExpression getString(final String field, final String orElse) {
-        return getString(field, of(orElse));
+    default StringExpression getString(final String field, final String other) {
+        return getString(field, of(other));
     }
 
     DateExpression getDate(String field);
-    DateExpression getDate(String field, DateExpression orElse);
+    DateExpression getDate(String field, DateExpression other);
 
-    default DateExpression getDate(final String field, final Instant orElse) {
-        return getDate(field, of(orElse));
+    default DateExpression getDate(final String field, final Instant other) {
+        return getDate(field, of(other));
     }
 
     DocumentExpression getDocument(String field);
-    DocumentExpression getDocument(String field, DocumentExpression orElse);
+    DocumentExpression getDocument(String field, DocumentExpression other);
 
-    default DocumentExpression getDocument(final String field, final Bson orElse) {
-        return getDocument(field, of(orElse));
+    default DocumentExpression getDocument(final String field, final Bson other) {
+        return getDocument(field, of(other));
     }
 
     <T extends Expression> ArrayExpression<T> getArray(String field);
 
-    <T extends Expression> ArrayExpression<T> getArray(String field, ArrayExpression<T> orElse);
+    <T extends Expression> ArrayExpression<T> getArray(String field, ArrayExpression<T> other);
 
-    DocumentExpression merge(DocumentExpression ofDoc);
+    DocumentExpression merge(DocumentExpression other);
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -28,65 +28,65 @@ import static com.mongodb.client.model.expressions.Expressions.of;
  */
 public interface DocumentExpression extends Expression {
 
-    DocumentExpression setField(String path, Expression exp);
+    DocumentExpression setField(String fieldName, Expression exp);
 
-    DocumentExpression unsetField(String path);
+    DocumentExpression unsetField(String fieldName);
 
 
-    BooleanExpression getBoolean(String field);
+    BooleanExpression getBoolean(String fieldName);
 
-    BooleanExpression getBoolean(String field, BooleanExpression other);
+    BooleanExpression getBoolean(String fieldName, BooleanExpression other);
 
-    default BooleanExpression getBoolean(final String field, final boolean other) {
-        return getBoolean(field, of(other));
+    default BooleanExpression getBoolean(final String fieldName, final boolean other) {
+        return getBoolean(fieldName, of(other));
     }
 
-    NumberExpression getNumber(String field);
+    NumberExpression getNumber(String fieldName);
 
-    NumberExpression getNumber(String field, NumberExpression other);
+    NumberExpression getNumber(String fieldName, NumberExpression other);
 
-    default NumberExpression getNumber(final String field, final Number other) {
-        return getNumber(field, Expressions.numberToExpression(other));
+    default NumberExpression getNumber(final String fieldName, final Number other) {
+        return getNumber(fieldName, Expressions.numberToExpression(other));
     }
 
-    IntegerExpression getInteger(String field);
+    IntegerExpression getInteger(String fieldName);
 
-    IntegerExpression getInteger(String field, IntegerExpression other);
+    IntegerExpression getInteger(String fieldName, IntegerExpression other);
 
-    default NumberExpression getInteger(final String field, final int other) {
-        return getInteger(field, (IntegerExpression) Expressions.numberToExpression(other));
+    default IntegerExpression getInteger(final String fieldName, final int other) {
+        return getInteger(fieldName, of(other));
     }
 
-    default NumberExpression getInteger(final String field, final long other) {
-        return getInteger(field, (IntegerExpression) Expressions.numberToExpression(other));
+    default IntegerExpression getInteger(final String fieldName, final long other) {
+        return getInteger(fieldName, of(other));
     }
 
 
-    StringExpression getString(String field);
+    StringExpression getString(String fieldName);
 
-    StringExpression getString(String field, StringExpression other);
+    StringExpression getString(String fieldName, StringExpression other);
 
-    default StringExpression getString(final String field, final String other) {
-        return getString(field, of(other));
+    default StringExpression getString(final String fieldName, final String other) {
+        return getString(fieldName, of(other));
     }
 
-    DateExpression getDate(String field);
-    DateExpression getDate(String field, DateExpression other);
+    DateExpression getDate(String fieldName);
+    DateExpression getDate(String fieldName, DateExpression other);
 
-    default DateExpression getDate(final String field, final Instant other) {
-        return getDate(field, of(other));
+    default DateExpression getDate(final String fieldName, final Instant other) {
+        return getDate(fieldName, of(other));
     }
 
-    DocumentExpression getDocument(String field);
-    DocumentExpression getDocument(String field, DocumentExpression other);
+    DocumentExpression getDocument(String fieldName);
+    DocumentExpression getDocument(String fieldName, DocumentExpression other);
 
-    default DocumentExpression getDocument(final String field, final Bson other) {
-        return getDocument(field, of(other));
+    default DocumentExpression getDocument(final String fieldName, final Bson other) {
+        return getDocument(fieldName, of(other));
     }
 
-    <T extends Expression> ArrayExpression<T> getArray(String field);
+    <T extends Expression> ArrayExpression<T> getArray(String fieldName);
 
-    <T extends Expression> ArrayExpression<T> getArray(String field, ArrayExpression<T> other);
+    <T extends Expression> ArrayExpression<T> getArray(String fieldName, ArrayExpression<? extends T> other);
 
     DocumentExpression merge(DocumentExpression other);
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -32,6 +32,7 @@ public interface DocumentExpression extends Expression {
 
     DocumentExpression unsetField(String fieldName);
 
+    Expression getField(String fieldName);
 
     BooleanExpression getBoolean(String fieldName);
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -17,6 +17,7 @@
 package com.mongodb.client.model.expressions;
 
 import org.bson.conversions.Bson;
+import org.bson.types.Decimal128;
 
 import java.time.Instant;
 
@@ -46,7 +47,11 @@ public interface DocumentExpression extends Expression {
 
     NumberExpression getNumber(String fieldName, NumberExpression other);
 
-    default NumberExpression getNumber(final String fieldName, final Number other) {
+    default NumberExpression getNumber(final String fieldName, final double other) {
+        return getNumber(fieldName, Expressions.numberToExpression(other));
+    }
+
+    default NumberExpression getNumber(final String fieldName, final Decimal128 other) {
         return getNumber(fieldName, Expressions.numberToExpression(other));
     }
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -48,11 +48,11 @@ public interface DocumentExpression extends Expression {
     NumberExpression getNumber(String fieldName, NumberExpression other);
 
     default NumberExpression getNumber(final String fieldName, final double other) {
-        return getNumber(fieldName, Expressions.numberToExpression(other));
+        return getNumber(fieldName, of(other));
     }
 
     default NumberExpression getNumber(final String fieldName, final Decimal128 other) {
-        return getNumber(fieldName, Expressions.numberToExpression(other));
+        return getNumber(fieldName, of(other));
     }
 
     IntegerExpression getInteger(String fieldName);
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -16,10 +16,66 @@
 
 package com.mongodb.client.model.expressions;
 
+import org.bson.conversions.Bson;
+
+import java.time.Instant;
+
+import static com.mongodb.client.model.expressions.Expressions.of;
+
 /**
  * Expresses a document value. A document is an ordered set of fields, where the
  * key is a string value, mapping to a value of any other expression type.
  */
 public interface DocumentExpression extends Expression {
 
+    <R extends DocumentExpression> R setField(String path, Expression exp);
+
+    DocumentExpression unsetField(String path);
+
+
+    BooleanExpression getBoolean(String field);
+
+    BooleanExpression getBoolean(String field, BooleanExpression orElse);
+
+    default BooleanExpression getBoolean(final String field, final boolean orElse) {
+        return getBoolean(field, of(orElse));
+    }
+
+    NumberExpression getNumber(String field);
+
+    NumberExpression getNumber(String field, NumberExpression orElse);
+
+    default NumberExpression getNumber(final String field, final Number orElse) {
+        return getNumber(field, Expressions.numberToExpression(orElse));
+    }
+
+    IntegerExpression getInteger(String field);
+
+    StringExpression getString(String field);
+
+    StringExpression getString(String field, StringExpression orElse);
+
+    default StringExpression getString(final String field, final String orElse) {
+        return getString(field, of(orElse));
+    }
+
+    DateExpression getDate(String field);
+    DateExpression getDate(String field, DateExpression orElse);
+
+    default DateExpression getDate(final String field, final Instant orElse) {
+        return getDate(field, of(orElse));
+    }
+
+    DocumentExpression getDocument(String field);
+    DocumentExpression getDocument(String field, DocumentExpression orElse);
+
+    default DocumentExpression getDocument(final String field, final Bson orElse) {
+        return getDocument(field, of(orElse));
+    }
+
+    <T extends Expression> ArrayExpression<T> getArray(String field);
+
+    <T extends Expression> ArrayExpression<T> getArray(String field, ArrayExpression<T> orElse);
+
+    DocumentExpression merge(DocumentExpression ofDoc);
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -107,7 +107,7 @@ public interface Expression {
     NumberExpression isNumberOr(NumberExpression or);
     StringExpression isStringOr(StringExpression or);
     DateExpression isDateOr(DateExpression or);
-    ArrayExpression<Expression> isArrayOr(ArrayExpression<? extends Expression> or);
+    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<T> or);
     <T extends DocumentExpression> T isDocumentOr(T or);
     StringExpression asString();
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -103,11 +103,13 @@ public interface Expression {
      * @param or
      * @return
      */
-    BooleanExpression isBooleanOr(BooleanExpression or);
-    NumberExpression isNumberOr(NumberExpression or);
-    StringExpression isStringOr(StringExpression or);
-    DateExpression isDateOr(DateExpression or);
-    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<T> or);
-    <T extends DocumentExpression> T isDocumentOr(T or);
+    BooleanExpression isBooleanOr(BooleanExpression other);
+    NumberExpression isNumberOr(NumberExpression other);
+    IntegerExpression isIntegerOr(IntegerExpression other);
+    StringExpression isStringOr(StringExpression other);
+    DateExpression isDateOr(DateExpression other);
+    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<T> other);
+    <T extends DocumentExpression> T isDocumentOr(T other);
+
     StringExpression asString();
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -108,7 +108,7 @@ public interface Expression {
     IntegerExpression isIntegerOr(IntegerExpression other);
     StringExpression isStringOr(StringExpression other);
     DateExpression isDateOr(DateExpression other);
-    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<T> other);
+    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<? extends T> other);
     <T extends DocumentExpression> T isDocumentOr(T other);
 
     StringExpression asString();
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -100,7 +100,7 @@ public interface Expression {
 
     /**
      * also checks for nulls
-     * @param or
+     * @param other
      * @return
      */
     BooleanExpression isBooleanOr(BooleanExpression other);
--- a/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
@@ -146,6 +146,11 @@ final class MqlExpression<T extends Expression>
     }
 
     @Override
+    public Expression getField(final String fieldName) {
+        return new MqlExpression<>(getFieldInternal(fieldName));
+    }
+
+    @Override
     public BooleanExpression getBoolean(final String fieldName) {
         return new MqlExpression<>(getFieldInternal(fieldName));
     }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
@@ -139,10 +139,15 @@ final class MqlExpression<T extends Expression>
 
     /** @see DocumentExpression */
 
-    private Function<CodecRegistry, AstPlaceholder> getFieldInternal(final String fieldBoolean) {
-        return (cr) -> astDoc("$getField", new BsonDocument()
-                .append("input", this.fn.apply(cr).bsonValue)
-                .append("field", new BsonString(fieldBoolean)));
+    private Function<CodecRegistry, AstPlaceholder> getFieldInternal(final String fieldName) {
+        return (cr) -> {
+            BsonValue value = fieldName.startsWith("$")
+                    ? new BsonDocument("$literal", new BsonString(fieldName))
+                    : new BsonString(fieldName);
+            return astDoc("$getField", new BsonDocument()
+                    .append("input", this.fn.apply(cr).bsonValue)
+                    .append("field", value));
+        };
     }
 
     @Override
@@ -216,7 +221,7 @@ final class MqlExpression<T extends Expression>
     }
 
     @Override
-    public <T1 extends Expression> ArrayExpression<T1> getArray(final String fieldName, final ArrayExpression<? extends T1> other) {
+    public <R extends Expression> ArrayExpression<R> getArray(final String fieldName, final ArrayExpression<? extends R> other) {
         return getArray(fieldName).isArrayOr(other);
     }
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
@@ -169,6 +169,12 @@ final class MqlExpression<T extends Expression>
     public IntegerExpression getInteger(final String field) {
         return new MqlExpression<>(getFieldInternal(field));
     }
+
+    @Override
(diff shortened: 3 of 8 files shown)


driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -137,6 +137,99 @@ public <R extends Expression> R cond(final R left, final R right) {
...
+
+    private Function<CodecRegistry, AstPlaceholder> getFieldInternal(final String fieldBoolean) {
+        // TODO these "getFieldX" names are all very long, and longer with composite paths
+        return (cr) -> astDoc("$getField", new BsonDocument()
+                .append("input", this.fn.apply(cr).bsonValue)
+                .append("field", new BsonString(fieldBoolean)));
+    }
+
+    @Override
+    public BooleanExpression getBoolean(final String field) {
```
katcharov, line 150: I have renamed `getFieldBoolean` to `getBoolean` (same for other types). This seems nicer, and incidentally similar to jdbc resultset.

We could also rename setField->set and unsetField->unset, but need to see how this turns out wrt other method names.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,235 @@
...
+                // MQL:
+                "{'$unsetField': {'field': 'a', 'input': {'$literal': {'a': 1}}}}");
+    }
+
+    @Test
+    public void mergeTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/mergeObjects/
+        assertExpression(
+                BsonDocument.parse("{a: 1, b: 2}"),
+                ofDoc("{a: 1}").merge(ofDoc("{b: 2}")),
```
katcharov, line 224: Unsure whether this method should be on a record, since it would just take the right item. Makes more sense for when "objects" (documents) are being treated as maps (future PR).

jyemin (commented):
This PR isn't in a reviewable state (only the last commit seems relevant).

katcharov:
@jyemin When I rebase one of the branches, all of the ensuing branches need to be rebased (with conflicts resolved along the way). Sometimes I'm caught in the middle of this process (this happened to @stIncMale a couple of times), and the diff is unreadable. This has been resolved in array, date, and conv, and soon in this branch.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,235 @@
...
+        assertExpression(8, ofIntegerArray(9, 8, 7).elementAt(ofDoc("{a: 1.0}").getInteger("a")));
+        assertExpression("a", ofDoc("{a: 'A'}").getString("a").toLower());
+        assertExpression(12, ofDoc("{a: {'$date': '2007-12-03T10:15:30.005Z'}}").getDate("a").month());
+        assertExpression(3, ofDoc("{a: [3, 2]}").getArray("a").first());
+        assertExpression(2, ofDoc("{a: {b: 2}}").getDocument("a").getInteger("b"));
+    }
+
+    @Test
+    public void getFieldOrTest() {
+        // getInteger-Or lacks isInteger; isInteger lacks a way to allow doubles that are integers
```
katcharov, line 100: There is no `isInteger`. `1.0` is a valid integer (because Integers are a subset of Numbers on the server and therefore in our type system), but there seems to be no clean way to check this. Is there opposition to `i eq i.floor ? true : false` as an `isInteger` check, or perhaps some better alternative?

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -210,11 +301,17 @@ public BooleanExpression isArray() {
...
+     *
+     * @param or
+     * @return
+     * @param <R>
+     */
     @Override
-    public ArrayExpression<Expression> isArrayOr(final ArrayExpression<? extends Expression> or) {
-        // TODO it seems that ArrEx<T> does not make sense here
-        return (ArrayExpression<Expression>) this.isArray().cond(this.assertImplementsAllExpressions(), or);
+    public <R extends Expression> ArrayExpression<R> isArrayOr(final ArrayExpression<R> or) {
```
katcharov, line 313: Note the signature change. Previously, this was safer (by returning Array<Ex>), since it did not rely on the user asserting the type. But when I needed to use it in `getArray`, it was not useful enough. This might have been resolved with a cast, but I assumed that users would run into the same issue.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,238 @@
...
+        ""    -- toLower...
+        null  -- multiply, add, subtract, year, filter, reduce, map, result within map...
+        */
+    }
+
+    @Test
+    public void setFieldTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/setField/
+        // "setField" is "withField": nothing is modified, and a new copy is produced.
+        // Placing a field based on a literal:
```
stIncMale, line 166: Comments like this one, or the one above appear to be a replacement for the API docs that are not being written at the moment. I commented on similar comments in https://github.com/mongodb/mongo-java-driver/pull/1050. What will happen to them eventually? I am not sure they should be left in the test code.

stIncMale, line 166: > "setField" is "withField": nothing is modified, and a new copy is produced

The server docs state that `$setField` _"Adds, updates, or removes a specified field in a document"_. However, in all examples the documentation uses `$setField` within the `$replaceWith` pipeline stage. I.e., at least from the examples it does not look like `$setFields` modifies documents. Is the server documentation just incorrect, or are there situations when `$setField` actually modifies a document? I am guessing there are non, but maybe you know better.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
stIncMale: I don't understand this comment. Could you please elaborate?

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+    private static DocumentExpression ofDoc(final String ofDoc) {
+        return of(BsonDocument.parse(ofDoc));
+    }
+
+    private final DocumentExpression a1 = ofDoc("{a: 1}");
+    private final DocumentExpression ax1ay2 = ofDoc("{a: {x: 1, y: 2}}");
+
+    @Test
+    public void literalsTest() {
```
stIncMale, line 45: Currently, `DocumentExpression` can only be obtained from a literal or as a value of a field of a different `DocumentExpression`. This means that all `DocumentExpression`s can only come from a literal (not necessary directly) with no way to have a `DocumentExpression` that represents data from a database. My understanding is that MQL "applies" `$setField` to documents that potentially come from a database via the [system variables](https://www.mongodb.com/docs/manual/reference/aggregation-variables/#system-variables) `$ROOT` and `$CURRENT`, but there appears to be no way of creating a `DocumentExpression` for either. Am I misunderstanding something, or is this work that is planned for an upcoming PR?

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
stIncMale: We discussed this. It's sad that there is no way to make the API fully type-safe without significantly affecting the performance. Here it's because we are making the type of array elements part of the array type, which is the case neither in the BSON type system and in the MQL type system.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,66 @@
...
+
 /**
  * Expresses a document value. A document is an ordered set of fields, where the
  * key is a string value, mapping to a value of any other expression type.
  */
 public interface DocumentExpression extends Expression {
 
+    <R extends DocumentExpression> R setField(String path, Expression exp);
+
+    DocumentExpression unsetField(String path);
```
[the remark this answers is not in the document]
katcharov: It seems these should both return DocEx (since they would be ineligible for overriding anyway).

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,66 @@
...
+    DocumentExpression getDocument(String field);
+    DocumentExpression getDocument(String field, DocumentExpression orElse);
+
+    default DocumentExpression getDocument(final String field, final Bson orElse) {
+        return getDocument(field, of(orElse));
+    }
+
+    <T extends Expression> ArrayExpression<T> getArray(String field);
+
+    <T extends Expression> ArrayExpression<T> getArray(String field, ArrayExpression<T> orElse);
```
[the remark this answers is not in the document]
katcharov:
```
getArray("abc", ofIntegerArray(1, 2, 3))
```
seems best.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
katcharov: Will delete; needs to be on the Map type.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
katcharov: There will be a `current()` shim method that I will add when I write the integration-with-existing tests.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
stIncMale: Since the expressions API makes a distinction between a map (a variable set of field keys mapped to field values which are all of the same type) and a record (a arbitrary but fixed set of field keys mapped to field values of not necessary the same type), which are mixed within the same type "object"/"document" in MQL expressions, maybe we should consider not using the word "document" when referring to records, and just instead use the word "record"? I foresee that referring to records as documents may cause confusion (I was confused until we discussed this).

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,66 @@
...
+
+    default BooleanExpression getBoolean(final String field, final boolean orElse) {
+        return getBoolean(field, of(orElse));
+    }
+
+    NumberExpression getNumber(String field);
+
+    NumberExpression getNumber(String field, NumberExpression orElse);
+
+    default NumberExpression getNumber(final String field, final Number orElse) {
```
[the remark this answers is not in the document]
katcharov: The `of` methods return their type. But these have a type after `get`, and need to return that type. This is confusing to users if we don't use Number, unlike the case with `of`. If you still perceive an inconsistency, let's add a note to the final doc to consider switching `of` to use Number.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,66 @@
...
+        return getNumber(field, Expressions.numberToExpression(orElse));
+    }
+
+    IntegerExpression getInteger(String field);
```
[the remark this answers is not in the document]
katcharov: Added.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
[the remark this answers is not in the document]
katcharov: (and sufficient, therefore leaving as-is)

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,66 @@
...
+
+    default DocumentExpression getDocument(final String field, final Bson orElse) {
+        return getDocument(field, of(orElse));
+    }
+
+    <T extends Expression> ArrayExpression<T> getArray(String field);
+
+    <T extends Expression> ArrayExpression<T> getArray(String field, ArrayExpression<T> orElse);
+
+    DocumentExpression merge(DocumentExpression ofDoc);
```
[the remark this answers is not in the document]
katcharov: I am equally confused.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,66 @@
...
+
+    IntegerExpression getInteger(String field);
+
+    StringExpression getString(String field);
+
+    StringExpression getString(String field, StringExpression orElse);
+
+    default StringExpression getString(final String field, final String orElse) {
+        return getString(field, of(orElse));
+    }
```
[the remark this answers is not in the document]
katcharov: Discussed, I recall the main issue was that we would need to always specify the type as a type parameter (whenever we wanted to use any methods apart from eq/lt/etc.), and this seemed less troublesome for users.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
katcharov: 165 is a clarification that belongs on the javadocs, but also doesn't hurt here in case the test starts failing. 166 is just a normal doc, which in this method I have placed above each test because they get fairly long and confusing.

katcharov: See also https://www.mongodb.com/docs/manual/reference/operator/aggregation/unset/

When it says "adds", it means "creates a copy and adds" or "returns the same document but with that field added".

It never modifies things, for the `setField` that is on the agg pipeline itself (iirc, but now I am second-guessing).

This is, of course, confusing. Especially on the aggregation pipeline itself.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArrayExpressionsFunctionalTest.java
```diff
@@ -17,6 +17,7 @@
 package com.mongodb.client.model.expressions;
 
 import com.mongodb.MongoCommandException;
+import org.bson.Document;
```
stIncMale, line 20: [I cannot leave the comment on the relevant line as the line came from a different PR]

Now that we have `isIntegerOr`, let's replace `array123.elementAt((IntegerExpression) of(0.0))` in `ArrayExpressionsFunctionalTest.elementAtTest` with `array123.elementAt(of(0.0).isIntegerOr(of(-1)))`. This comment is from a [previous PR](https://github.com/mongodb/mongo-java-driver/pull/1050#discussion_r1034061470).

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -103,12 +103,13 @@ public interface Expression {
...
-    StringExpression isStringOr(StringExpression or);
-    DateExpression isDateOr(DateExpression or);
-    ArrayExpression<Expression> isArrayOr(ArrayExpression<? extends Expression> or);
-    <T extends DocumentExpression> T isDocumentOr(T or);
+    BooleanExpression isBooleanOr(BooleanExpression other);
+    NumberExpression isNumberOr(NumberExpression other);
+    IntegerExpression isIntegerOr(IntegerExpression other);
+    StringExpression isStringOr(StringExpression other);
+    DateExpression isDateOr(DateExpression other);
+    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<T> other);
```
stIncMale, line 111: The parameter type should be `ArrayExpression<? extends T>` for the same reason we agreed on `ArrayExpression<? extends Expression>` in a [previous PR](https://github.com/mongodb/mongo-java-driver/pull/1050/files?show-viewed-files=true&file-filters%5B%5D=#r1030839953), and for the same reason I [requested a change](https://github.com/mongodb/mongo-java-driver/pull/1053#discussion_r1033053835) in a [subsequent PR](https://github.com/mongodb/mongo-java-driver/pull/1053#discussion_r1033053835). The reason is: we want `ArrayExpression<IntegerExpression>` to be a subtype of `ArrayExpression<NumberExpression>` (it's false in Java, but the API may make you feel like it's true).

The code example that does not compile now, but will after the change:
```java
ArrayExpression<NumberExpression> a = ofBooleanArray().isArrayOr(ofIntegerArray());
```

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
katcharov: I agree. But the renaming should be done at the end because of rebasing. https://github.com/mongodb/mongo-java-driver/pull/1032/files#r1032699544

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -28,54 +28,65 @@
...
 
+    IntegerExpression getInteger(String field, IntegerExpression other);
+
+    default NumberExpression getInteger(final String field, final int other) {
+        return getInteger(field, (IntegerExpression) Expressions.numberToExpression(other));
+    }
+
+    default NumberExpression getInteger(final String field, final long other) {
+        return getInteger(field, (IntegerExpression) Expressions.numberToExpression(other));
+    }
```
stIncMale, line 62: Two notes here:

1. These methods should return `IntegerExpression`.
2. Why is casting `(IntegerExpression)` used instead of implementing the methods as `return getInteger(field, Expressions.of(other));`?

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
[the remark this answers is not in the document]
stIncMale: > The of methods return their type. But these have a type after get, and need to return that type. This is confusing to users if we don't use Number, unlike the case with of.

I think, this argument confuses the MQL number type (expressed in Java as `NumberExpression` and mentioned in the `getNumber` method name) with the Java `Number` (specified as the type of the second parameter) - they are different types. Also, if this argument were applied consistently, then neither `DocumentExpression.getInteger(String, long)` nor `Expressions.ofNumberArray(double...)`/`(Decimal128...)` should not have been introduced as they violate the approach implied by the argument.

> If you still perceive an inconsistency, let's add a note to the final doc to consider switching of to use Number.

The API inconsistency was introduced in this PR, and deciding what to do with it does not require postponing as far as I can see (if there were specific future decisions you foresee on which this decision has to rely on, I would expect those to have been already mentioned).

---

The current integer-specific API is consistent between `Expressions` and `DocumentExpression`:

```java
IntegerExpression Expressions.of(int)
IntegerExpression Expressions.of(long)

IntegerExpression DocumentExpression.getInteger(String, int)
IntegerExpression DocumentExpression.getInteger(String, long)
```

the API that is not integer-specific is not consistent between `Expressions` and `DocumentExpression`:

```java 
NumberExpression Expressions.of(double)
NumberExpression Expressions.of(Decimal128)

NumberExpression DocumentExpression.getNumber(String, Number)
```

To achieve consistency we should

1. either replace `getNumber(String, Number)` with `getNumber(String, double)` and `getNumber(String, Decimal128)`;
2. or replace `of(double)` and `of(Decimal128)` with `of(Number)` (in this case we also need to replace `Expressions.ofNumberArray(double...)`/`(Decimal128...)` with `Expressions.ofNumberArray(Number...)`).

Second option reduces the number of overloads at the cost of shifting some type-checks from compile-time to run-time. I will be fine with either option.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -28,54 +28,65 @@
...
-    default DocumentExpression getDocument(final String field, final Bson orElse) {
-        return getDocument(field, of(orElse));
+    default DocumentExpression getDocument(final String field, final Bson other) {
+        return getDocument(field, of(other));
     }
 
     <T extends Expression> ArrayExpression<T> getArray(String field);
 
-    <T extends Expression> ArrayExpression<T> getArray(String field, ArrayExpression<T> orElse);
+    <T extends Expression> ArrayExpression<T> getArray(String field, ArrayExpression<T> other);
```
stIncMale, line 89: Similarly to `Expression.isArrayOr` we need `ArrayExpression<? extends T>` as the parameter type here.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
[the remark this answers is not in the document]
stIncMale: > the main issue was that we would need to always specify the type as a type parameter

That's only if we were to replace `get[Type](String field)` methods that don't have the `other` parameter, but the suggestion was to remove the methods that have the `other` parameter with a single `get(String field)` method and allow users to explicitly combine that with the `Expression.is[Type]Or` method when needed (probably in most situations). The issue was that this does not seem very discoverable and is more verbose.

I thought about this more and regardless of whether we live the shortcut methods that spare a user from calling `Expression.is[Type]Or`, we seen to need to add `Expression DocumentExpression.getField(String)` to cover a situation when the field type is unknown. A simple example is when a field may contain either a date, or a string representation of a date. As far as I can see, we don't currently provide a way to get a date value from such a field, but with `getField` a user can do `getField("field").switch(...)`.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,77 @@
...
  * key is a string value, mapping to a value of any other expression type.
  */
 public interface DocumentExpression extends Expression {
 
+    DocumentExpression setField(String path, Expression exp);
+
+    DocumentExpression unsetField(String path);
+
+
+    BooleanExpression getBoolean(String field);
```
jyemin, line 36: Let's also support `Expression get(String field)` for cases where you don't know the type and either 

1. don't care what it is, you just want to include it in a projection or do only Expression-y things with it.
2. want to do a type test on it, e.g. `doc.get("unknown").isBooleanOr(of(true))`

jyemin, line 36: What is the behavior of `doc.getBoolean("a.b")`.  Does it 

1. return the value of the "a.b" field in this document
2. return the value of the "b" field inside the "a" field (assuming the value of the "a" field is a document)

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,77 @@
...
+
+import static com.mongodb.client.model.expressions.Expressions.of;
+
 /**
  * Expresses a document value. A document is an ordered set of fields, where the
  * key is a string value, mapping to a value of any other expression type.
  */
 public interface DocumentExpression extends Expression {
 
+    DocumentExpression setField(String path, Expression exp);
```
jyemin, line 31: According to the setField documentation:

> [$setField](https://www.mongodb.com/docs/manual/reference/operator/aggregation/setField/#mongodb-expression-exp.-setField) doesn't implicitly traverse objects or arrays. For example, [$setField](https://www.mongodb.com/docs/manual/reference/operator/aggregation/setField/#mongodb-expression-exp.-setField) evaluates a field value of "a.b.c" as a top-level field "a.b.c" instead of as a nested field, { "a": { "b": { "c": } } }.

So it's not correct to call the first parameter `path`.  It's really the field name.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,77 @@
...
+
+    DocumentExpression unsetField(String path);
```
jyemin, line 33: Same here

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
jyemin: If 1., the parameter should be named `field`.  If 2. is there any way to use dotted path notation in this API, and if not, should there be?

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
jyemin: Let's have a separate ticket and PR for `current()`. It's important enough to deserve top billing.

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,243 @@
...
+                "{'$literal': {}}");
+        // ensure is literal
+        assertExpression(BsonDocument.parse(
+                "{'lit': {'$not': true}}"),
+                of(BsonDocument.parse("{lit: {'$not': true} }")),
+                "{'$literal': {'lit': {'$not': true}}}");
+    }
+
+    @Test
+    public void getFieldTest() {
```
jyemin, line 69: Please add tests for field names with embedded `.` characters so that we have regression tests on the behavior

stIncMale (changes requested):
Sorry, I have to press "submit" even though I have not finished re-reviewing because GitHub UI messed up the display of pending comments, and I am afraid I may lose them.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
jyemin: If we're not supporting dotted path notation, I worry that because their use is so prevalent in the wild, applications will make the mistake of using them and being really confused by the results they get back

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -16,10 +16,77 @@
...
+
+import java.time.Instant;
+
+import static com.mongodb.client.model.expressions.Expressions.of;
+
 /**
  * Expresses a document value. A document is an ordered set of fields, where the
  * key is a string value, mapping to a value of any other expression type.
  */
 public interface DocumentExpression extends Expression {
```
jyemin, line 29: Both setField and getField allow the field name itself to be an expression.  That API should support that as well.

jyemin (changes requested):

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
katcharov: Also need to check if this is broken in the current implementation - on the first reading of the [getField](https://www.mongodb.com/docs/manual/reference/operator/aggregation/getField/) docs, it seemed it would be automatically treated as a literal, but I think it needs to be explicitly wrapped.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: The intent was to return (1) the value of the "a.b" field, not the path.

The potential error you mention seems serious to me, but I am unsure of how to address it.

- It seems that there needs to be some way to use a literal field notation, so whatever we do we will need one `Expression get()` or many getters that take a proper field.
- We could try `getBooleanField` and `getBooleanPath`, but this seems like overkill, and perhaps would not address the possible user error.
- There is a `get(key)` on MapExpression that does not and should not take a path. MqlExpression implements both interfaces.
- If we can implement what I have been calling "schema objects" (perhaps "object bindings"?) then one can write: `doc.a.b`. I think this would disincline users from any of the getField methods, and avoid the issue, but only to the extent that object bindings are possible and used by users.

I'll keep thinking. Suggestions welcome.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -169,6 +169,12 @@ public NumberExpression getNumber(final String field, final NumberExpression orE
     public IntegerExpression getInteger(final String field) {
         return new MqlExpression<>(getFieldInternal(field));
     }
+
+    @Override
+    public IntegerExpression getInteger(String field, IntegerExpression orElse) {
```
stIncMale: orElse -> other?

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: This is supported for Map<T> (arbitrary number of string keys to values all of the same type), but is not supported for Documents, which represent records (fixed number of string-named fields to values of different types). Neither one extends the other, because:

1. A record cannot have a `set(T)` method since this would allow fields to be set via a broader type than they actually are.
2. A `map<StringEx>` cannot have a method `getBoolean(field)` because this would allow string values to be pulled out as booleans.

Any situation that requires this is almost certainly more correctly handled via a `MapExpression<Expression>`, rather than a document (record).

jyemin: What is a `MapExpression`?  I don't see that anywhere.

stIncMale (changes requested):

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
katcharov: https://github.com/mongodb/mongo-java-driver/pull/1057

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -28,54 +28,65 @@
...
 
-    <R extends DocumentExpression> R setField(String path, Expression exp);
+    DocumentExpression setField(String path, Expression exp);
 
     DocumentExpression unsetField(String path);
```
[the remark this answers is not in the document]
katcharov: I renamed these to `fieldName`, though note the discussion below https://github.com/mongodb/mongo-java-driver/pull/1052#discussion_r1036405352 regarding whether they should be paths.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: Updated to `fieldName`. Perhaps ensuring consistency with setField is one reason to use fieldNames rather than paths in getField variations [discussed above](https://github.com/mongodb/mongo-java-driver/pull/1052#discussion_r1036405352).

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: It's in https://github.com/mongodb/mongo-java-driver/pull/1054

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
katcharov: I would generally read comments in failing tests.

Certain comments explain why certain things are or are not tested, or why tests are written or behave in a certain way. I am not sure what criteria to use for including or excluding such comments from tests, so I cannot hope to reliably exclude them in the future, though I don't have strong objections to removing them if some objection is raised in a PR.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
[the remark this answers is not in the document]
katcharov: added `getField`. See also get in map and getValue in entity ([other PR](https://github.com/mongodb/mongo-java-driver/pull/1054/files)).

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: Added `getField`.

jyemin: I think the more common case will be people wanting to treat the field names as paths that traverse subdocuments.  It's only in the last couple of years that the server or drivers even allowed field names with dots in them.  So that suggests that the default behavior should be to treat dotted field names as paths, and have an explicit way to treat them as field names.

At the same time, it's an attractive option to use this opportunity to "fix" the behavior, and provide explicit syntax for dotted paths.  So something like:

```
   Expression getField(String fieldName);    // treat "a.b" as a field name
   Expression getField(List<String> path);   // treat each string in the list as a field name
```

stIncMale: > At the same time, it's an attractive option to use this opportunity to "fix" the behavior, and provide explicit syntax for dotted paths. So something like:
> getField(List<String> path)

If we use any type but `String` to represent field names / paths, I think we should introduce a special type instead of using `List<String>`.

I would also prefer to just introduce a `Path` type with two ways of creating its instances, something like `Path.name` and `Path.path` (in the search API we have `SearchPath.fieldPath` and `SearchPath.wildcardPath`, and we never accept plain `String` as a path in the rest of the API). We may even introduce two subtypes if there are situations in MQL when a path is not allowed and only a field name may be used (I don't know if there are such situations). I discussed this with Maxim, and he does not like the idea due to its verbosity, others may think the same. But I think that introducing a type that covers both field names (just a degenerate case of a field path) and field paths is the right way to approach the task: a plain `String` can technically be used to represent a full name of a human, text of a book, a path in a file system, a field path in MongoDB, but it does not mean that using a plain String for any of that is necessarily a good idea. Having a special type allows one to both express the meaning of a value and validate it at creation.

katcharov: Discussed, all prefer the `getInteger(String fieldName)`. It would be beneficial to have a way to query by paths, but that can be considered later.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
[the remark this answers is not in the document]
katcharov: I think what you are saying is correct. I will consider the two options.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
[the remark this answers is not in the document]
katcharov: I chose the first one, because it is more consistent with the nearby `getInteger` methods.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: Added tests. "fieldName" is a fairly long parameter hint in IntelliJ, but let's see how that goes.

stIncMale (changes requested):

driver-core/src/test/functional/com/mongodb/client/model/expressions/DocumentExpressionsFunctionalTest.java
jyemin: Is this done?

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -46,7 +47,11 @@ default BooleanExpression getBoolean(final String fieldName, final boolean other
...
     NumberExpression getNumber(String fieldName, NumberExpression other);
 
-    default NumberExpression getNumber(final String fieldName, final Number other) {
+    default NumberExpression getNumber(final String fieldName, final double other) {
+        return getNumber(fieldName, Expressions.numberToExpression(other));
+    }
+
+    default NumberExpression getNumber(final String fieldName, final Decimal128 other) {
         return getNumber(fieldName, Expressions.numberToExpression(other));
     }
```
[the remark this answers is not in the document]
jyemin: Is this done?

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
jyemin: Can you show me what the code would look like?  I don't understand how it works in practice, e.g. say I have a document like:

```
{
   s1 : "str1",
   s2: "str2",
   s3 : "str3",
   i: 1
}
```

How would I do something like set one of the `s` fields where the suffix of the field name is the value of the `i` field?

jyemin (commented):
Updated some of the comment threads that I'm on.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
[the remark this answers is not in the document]
katcharov: [Yes](https://github.com/mongodb/mongo-java-driver/pull/1052/commits/c02249c38890d7fa99602848d080b2c5a2bf519d)

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
jyemin: Deferring at least to map expression PR

jyemin (approved):

PR #1053, opened 2022-11-22
Re-implement reduce

katcharov:
JAVA-4814

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -50,25 +50,30 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      */
     <R extends Expression> ArrayExpression<R> map(Function<? super T, ? extends R> in);
 
+    ArrayExpression<T> sort();
+
     IntegerExpression size();
 
-    BooleanExpression any(Function<T, BooleanExpression> map);
+    BooleanExpression any(Function<T, BooleanExpression> mapper);
 
-    BooleanExpression all(Function<T, BooleanExpression> map);
+    BooleanExpression all(Function<T, BooleanExpression> mapper);
 
-    NumberExpression sum(Function<T, NumberExpression> map);
+    NumberExpression sum(Function<T, NumberExpression> mapper);
 
-    NumberExpression multiply(Function<T, NumberExpression> map);
+    NumberExpression multiply(Function<T, NumberExpression> mapper);
 
-    NumberExpression max(Function<T, NumberExpression> map, NumberExpression orElse);
+    <R extends Expression> R max(Function<T, R> mapper, R orElse);
 
-    NumberExpression min(Function<T, NumberExpression> map, NumberExpression orElse);
+    <R extends Expression> R min(Function<T, R> mapper, R orElse);
 
-    StringExpression join(Function<T, StringExpression> map);
+    <R extends Expression> ArrayExpression<R> maxN(IntegerExpression n, Function<? super T, ? extends R> mapper);
+    <R extends Expression> ArrayExpression<R> minN(IntegerExpression n, Function<? super T, ? extends R> mapper);
 
-    <R extends Expression> ArrayExpression<R> concat(Function<T, ArrayExpression<R>> map);
+    StringExpression join(Function<T, StringExpression> mapper);
 
-    <R extends Expression> ArrayExpression<R> union(Function<T, ArrayExpression<R>> map);
+    <R extends Expression> ArrayExpression<R> concat(Function<? super T, ? extends ArrayExpression<? extends R>> mapper);
+
+    <R extends Expression> ArrayExpression<R> union(Function<? super T, ? extends ArrayExpression<? extends R>> mapper);
 
     /**
      * user asserts that i is in bounds for the array
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -50,26 +50,24 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      */
     <R extends Expression> ArrayExpression<R> map(Function<? super T, ? extends R> in);
 
-    ArrayExpression<T> sort();
-
     IntegerExpression size();
 
-    BooleanExpression any(Function<T, BooleanExpression> predicate);
+    BooleanExpression any(Function<? super T, BooleanExpression> predicate);
 
-    BooleanExpression all(Function<T, BooleanExpression> predicate);
+    BooleanExpression all(Function<? super T, BooleanExpression> predicate);
 
-    NumberExpression sum(Function<T, NumberExpression> mapper);
+    NumberExpression sum(Function<? super T, ? extends NumberExpression> mapper);
 
-    NumberExpression multiply(Function<T, NumberExpression> mapper);
+    NumberExpression multiply(Function<? super T, ? extends NumberExpression> mapper);
 
-    <R extends Expression> R max(R other, Function<? super T, ? extends R> mapper);
+    T max(T other);
 
-    <R extends Expression> R min(R other, Function<? super T, ? extends R> mapper);
+    T min(T other);
 
-    <R extends Expression> ArrayExpression<R> maxN(IntegerExpression n, Function<? super T, ? extends R> mapper);
-    <R extends Expression> ArrayExpression<R> minN(IntegerExpression n, Function<? super T, ? extends R> mapper);
+    ArrayExpression<T> maxN(IntegerExpression n);
+    ArrayExpression<T> minN(IntegerExpression n);
 
-    StringExpression join(Function<T, StringExpression> mapper);
+    StringExpression join(Function<? super T, StringExpression> mapper);
 
     <R extends Expression> ArrayExpression<R> concat(Function<? super T, ? extends ArrayExpression<? extends R>> mapper);
 
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -16,7 +16,6 @@
 
 package com.mongodb.client.model.expressions;
 
-import java.util.function.BinaryOperator;
 import java.util.function.Function;
 
 import static com.mongodb.client.model.expressions.Expressions.of;
@@ -51,28 +50,48 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      */
     <R extends Expression> ArrayExpression<R> map(Function<? super T, ? extends R> in);
 
-    /**
-     * Performs a reduction on the elements of this array, using the provided
-     * identity value and an associative reducing function, and returns
-     * the reduced value. The initial value must be the identity value for the
-     * reducing function.
-     *
-     * @param initialValue the identity for the reducing function
-     * @param in the associative reducing function
-     * @return the reduced value
-     */
-    T reduce(T initialValue, BinaryOperator<T> in);
-
     IntegerExpression size();
 
+    BooleanExpression any(Function<T, BooleanExpression> map);
+
+    BooleanExpression all(Function<T, BooleanExpression> map);
+
+    NumberExpression sum(Function<T, NumberExpression> map);
+
+    NumberExpression multiply(Function<T, NumberExpression> map);
+
+    NumberExpression max(Function<T, NumberExpression> map, NumberExpression orElse);
+
+    NumberExpression min(Function<T, NumberExpression> map, NumberExpression orElse);
+
+    StringExpression join(Function<T, StringExpression> map);
+
+    <R extends Expression> ArrayExpression<R> concat(Function<T, ArrayExpression<R>> map);
+
+    <R extends Expression> ArrayExpression<R> union(Function<T, ArrayExpression<R>> map);
+
+    /**
+     * user asserts that i is in bounds for the array
+     *
+     * @param i
+     * @return
+     */
     T elementAt(IntegerExpression i);
 
     default T elementAt(final int i) {
         return this.elementAt(of(i));
     }
 
+    /**
+     * user asserts that array is not empty
+     * @return
+     */
     T first();
 
+    /**
+     * user asserts that array is not empty
+     * @return
+     */
     T last();
 
     BooleanExpression contains(T contains);
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -54,17 +54,17 @@ public interface ArrayExpression<T extends Expression> extends Expression {
 
     IntegerExpression size();
 
-    BooleanExpression any(Function<T, BooleanExpression> mapper);
+    BooleanExpression any(Function<T, BooleanExpression> predicate);
 
-    BooleanExpression all(Function<T, BooleanExpression> mapper);
+    BooleanExpression all(Function<T, BooleanExpression> predicate);
 
     NumberExpression sum(Function<T, NumberExpression> mapper);
 
     NumberExpression multiply(Function<T, NumberExpression> mapper);
 
-    <R extends Expression> R max(Function<T, R> mapper, R orElse);
+    <R extends Expression> R max(R other, Function<? super T, ? extends R> mapper);
 
-    <R extends Expression> R min(Function<T, R> mapper, R orElse);
+    <R extends Expression> R min(R other, Function<? super T, ? extends R> mapper);
 
     <R extends Expression> ArrayExpression<R> maxN(IntegerExpression n, Function<? super T, ? extends R> mapper);
     <R extends Expression> ArrayExpression<R> minN(IntegerExpression n, Function<? super T, ? extends R> mapper);
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
@@ -17,7 +17,6 @@
 package com.mongodb.client.model.expressions;
 
 import org.bson.conversions.Bson;
-import org.bson.types.Decimal128;
 
 import java.time.Instant;
 
@@ -47,12 +46,8 @@ public interface DocumentExpression extends Expression {
 
     NumberExpression getNumber(String fieldName, NumberExpression other);
 
-    default NumberExpression getNumber(final String fieldName, final double other) {
-        return getNumber(fieldName, of(other));
-    }
-
-    default NumberExpression getNumber(final String fieldName, final Decimal128 other) {
-        return getNumber(fieldName, of(other));
+    default NumberExpression getNumber(final String fieldName, final Number other) {
+        return getNumber(fieldName, Expressions.numberToExpression(other));
     }
 
     IntegerExpression getInteger(String fieldName);
--- a/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
@@ -411,13 +411,13 @@ final class MqlExpression<T extends Expression>
     }
 
     @Override
-    public BooleanExpression any(final Function<T, BooleanExpression> mapper) {
-        return reduceMap(mapper, of(false), (a, b) -> a.or(b));
+    public BooleanExpression any(final Function<T, BooleanExpression> predicate) {
+        return reduceMap(predicate, of(false), (a, b) -> a.or(b));
     }
 
     @Override
-    public BooleanExpression all(final Function<T, BooleanExpression> mapper) {
-        return reduceMap(mapper, of(true), (a, b) -> a.and(b));
+    public BooleanExpression all(final Function<T, BooleanExpression> predicate) {
+        return reduceMap(predicate, of(true), (a, b) -> a.and(b));
     }
 
     @Override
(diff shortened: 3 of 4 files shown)


driver-core/src/test/functional/com/mongodb/client/model/expressions/ArrayExpressionsFunctionalTest.java
```diff
@@ -127,41 +127,128 @@ public void mapTest() {
...
     }
 
+    // https://www.mongodb.com/docs/manual/reference/operator/aggregation/reduce/
+    // reduce is implemented as each individual type of reduction (monoid)
+    // this prevents issues related to incorrect specification of identity values
+
     @Test
-    public void reduceTest() {
-        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/reduce/
+    public void reduceAnyTest() {
```
stIncMale, line 135: I propose calling all the tests correspondingly to the tested method names, i.e., without the "reduce" part.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -51,28 +50,48 @@
...
 
+    IntegerExpression size();
+
+    BooleanExpression any(Function<T, BooleanExpression> map);
+
+    BooleanExpression all(Function<T, BooleanExpression> map);
+
+    NumberExpression sum(Function<T, NumberExpression> map);
+
+    NumberExpression multiply(Function<T, NumberExpression> map);
```
stIncMale, line 61: Using `Function<T, NumberExpression> map` as a parameter of `multiply` (and other similar methods) is a clever way of working around the fact that in Java, unlike Rust (and, I am sure, other languages that unlike Java implement generics by duplicating code), we can't express that a method should be associated with a type only if the parametrization of the corresponding class meets certain criteria (in this particular case, the criterium is that only type argument is `NumberExpression`).

An alternative approach would have been to make `multiply` a `static` method that accepts `ArrayExpression<? extends NumberExpression>`. This way we would have directly associate the method with a type based on the criteria we need, at the cost of making the method `static`.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -51,28 +50,48 @@
      */
     <R extends Expression> ArrayExpression<R> map(Function<? super T, ? extends R> in);
 
+    IntegerExpression size();
+
+    BooleanExpression any(Function<T, BooleanExpression> map);
+
+    BooleanExpression all(Function<T, BooleanExpression> map);
```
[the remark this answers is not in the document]
katcharov: Those are not bad, but then it seems like we should rename filter to match the match aggregation stage. I wonder if they are Java-specific. Javascript seems to use `some` and `every` ([every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)). I am unsure.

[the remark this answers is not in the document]
katcharov: I think probably not, because here they are actually used as a map. If you write `.any(v -> v. many().methods().here())`, everything after the `v -> v` can be moved into a preceding map stage.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -51,28 +50,48 @@
...
+
+    NumberExpression sum(Function<T, NumberExpression> map);
+
+    NumberExpression multiply(Function<T, NumberExpression> map);
+
+    NumberExpression max(Function<T, NumberExpression> map, NumberExpression orElse);
+
+    NumberExpression min(Function<T, NumberExpression> map, NumberExpression orElse);
```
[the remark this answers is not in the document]
katcharov: I think we can still use `$min` for min, but then use a T for the return value. I like this.

I am not sure about implementing other operators beyond the "top-50", since otherwise there is no stopping.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArrayExpressionsFunctionalTest.java
katcharov: There is both concat(v -> v) and concat(otherArray).

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
katcharov: But then multiply would not be chained, yes? I think even if it fits into the apply method, this would be odd/inconvenient.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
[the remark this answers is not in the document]
katcharov: You were correct that implementing this using minN was best; I have added minN and maxN. I have named them `max` and `min` - perhaps the N should be included.

For changing min/max to be available everywhere: https://github.com/mongodb/mongo-java-driver/pull/1032/files#r1035214309

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -51,28 +50,48 @@
...
+
+    StringExpression join(Function<T, StringExpression> map);
+
+    <R extends Expression> ArrayExpression<R> concat(Function<T, ArrayExpression<R>> map);
+
+    <R extends Expression> ArrayExpression<R> setUnion(Function<T, ArrayExpression<R>> map);
```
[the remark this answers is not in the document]
katcharov: Implemented with adjustments.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -51,28 +50,53 @@
...
-     * Performs a reduction on the elements of this array, using the provided
-     * identity value and an associative reducing function, and returns
-     * the reduced value. The initial value must be the identity value for the
-     * reducing function.
-     *
-     * @param initialValue the identity for the reducing function
-     * @param in the associative reducing function
-     * @return the reduced value
-     */
-    T reduce(T initialValue, BinaryOperator<T> in);
```
jyemin, line 64: Can you explain how the below methods cover all the use cases of reduce?  It's not at all clear to me.

katcharov: The returned value of the binary operator `in` needs to come from some combination of its parameters `(a, b)`, and the only way to "combine" them is to use some binary method expressible in our language. I went through the list of [all methods we are presently implementing](https://docs.google.com/spreadsheets/d/1mqkG2XJ4Telnhmi1V19TpwWCjuy4jelNAfuIm9HqPcA/edit?pli=1#gid=1430738632) and considered whether each one was a reductive operation (column D). If it was, I implemented a method for it below.

Therefore every usage that involves reduction either can be modelled using one of the below, or if it cannot be modelled with one of the below, it cannot be modelled using this removed `reduce` operation regardless.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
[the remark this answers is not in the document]
stIncMale: > they are actually used as a map

It does not matter how the method is implemented. Names are supposed to convey behavior/meaning, not implementation details. To me, as a user, seeing the condition which any/all array element should satisfy referred to as "map" is not helpful.

If we were creating API whose purpose is building MQL, then naming method parameters in a way that corresponds to their names in MQL would have been reasonable. But in this case we are not creating such an API, and the MQL parameter name is not `map`, it's `in`.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -50,25 +50,30 @@
...
+    NumberExpression sum(Function<T, NumberExpression> mapper);
 
-    NumberExpression sum(Function<T, NumberExpression> map);
+    NumberExpression multiply(Function<T, NumberExpression> mapper);
 
-    NumberExpression multiply(Function<T, NumberExpression> map);
+    <R extends Expression> R max(Function<T, R> mapper, R orElse);
 
-    NumberExpression max(Function<T, NumberExpression> map, NumberExpression orElse);
+    <R extends Expression> R min(Function<T, R> mapper, R orElse);
```
stIncMale, line 67: Similar usages of `orElse` were changed to `other` in https://github.com/mongodb/mongo-java-driver/pull/1052. We should use `other` in `min`/`max`.

stIncMale, line 67: Let's switch the parameters places similarly to `minN`/`maxN`, as writing a lambda expression when specifying the value of a method parameter is more convenient when the parameter is at the last place.

stIncMale, line 67: Similarly to `minN`/`maxN`, the type of the mapper should be `Function<? super T, ? extends R>`

stIncMale, line 67: `ArrayExpression.first`/`last` don't take a fallback value, but `min`/`max` do. Since there is no conceptual difference between these two sets of methods, we should use the same approach in them.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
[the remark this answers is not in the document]
stIncMale: Either I am being blind (happens sometimes, I apologize if that's the case), or you decided not to introduce adjustments. Could you please point out what was adjusted in the method declaration compared to the proposed one?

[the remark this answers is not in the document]
katcharov: I was confused, the "adjustments" were all in the implementation, so no adjustments.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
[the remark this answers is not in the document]
katcharov: Oh I see what you mean. I have changed it to `predicate`, which I updated (I think) filter to (in a forthcoming javadoc PR).

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
katcharov: I am second-guessing that change, but I carried it through to this.

katcharov: I like it.

katcharov: There is also `elementAt`, `getBoolean(String)` (and all the rest), and `get` on MapExpression. I think my plan is to make the API type-safe wherever I can get away with it, and I think having a default is passable here but more problematic elsewhere.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
jyemin: I have a counterexample:

```
> db.test1.find()
[
  {
    a: [ { x: 3, y: 4 }, { x: 5, y: 11 } ]
  },
  {
    a: []
  }
]

> db.test1.aggregate([{$project : {
   _id : 0, 
   aReduced: {$reduce: {
       input: "$a", 
       initialValue: {x : 0, y: 1 }, 
       in: {x : {$sum: ["$$this.x", "$$value.x"] }, y: {$multiply: ["$$this.y", "$$value.y"] }   }    } } } 
   }
])
[ 
  { aReduced: { x: 8, y: 44 } }, 
  { aReduced: { x: 0, y: 1 } } 
]
```

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -50,25 +50,30 @@
      */
     <R extends Expression> ArrayExpression<R> map(Function<? super T, ? extends R> in);
 
+    ArrayExpression<T> sort();
```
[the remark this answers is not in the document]
katcharov: I think I will remove sort from the public API, but use it for tests.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -51,28 +50,53 @@
...
 
+    BooleanExpression any(Function<T, BooleanExpression> predicate);
+
+    BooleanExpression all(Function<T, BooleanExpression> predicate);
+
+    NumberExpression sum(Function<T, NumberExpression> mapper);
+
+    NumberExpression multiply(Function<T, NumberExpression> mapper);
+
+    <R extends Expression> R max(R other, Function<? super T, ? extends R> mapper);
```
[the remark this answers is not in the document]
katcharov: You are right

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
katcharov: This is a fair example. It is not a matter of expressiveness or functionality, since a user wishing to obtain the same result is able to write the following reduction (3 lines below), which produces the above results:

```
// setup; array contains the documents
ArrayExpression<DocumentExpression> array = ofArray(
          of(Document.parse("{ a: [ { x: 3, y: 4 }, { x: 5, y: 11 } ] }")),
          of(Document.parse("{ a: [ ] }")));
DocumentExpression empty = of(Document.parse("{}"));

// reduction
Function<ArrayExpression<DocumentExpression>, DocumentExpression> multiReduce = (a) -> empty
          .setField("x", a.sum(e -> e.getInteger("x")))
          .setField("y", a.multiply(e -> e.getInteger("y")));

// assertion and get/set
assertExpression(
          Arrays.asList(
               Document.parse("{ aReduced: { x: 8, y: 44 } }"),
               Document.parse("{ aReduced: { x: 0, y: 1 } }")),
          array.map(d -> empty.setField("aReduced", multiReduce.apply(d.getArray("a")))));
```
It is, however, a performance issue - we want to perform those two reductions at the same time. For the sake of performance, I agree that we should offer a variant that allows multiple reductions to occur in tandem, which is what [$group](https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/) does. That is, once we figure out the API for group, the problematic example you propose will be addressed, without the need to expose `reduce`.

1. We can always add reduce later but cannot remove it
2. Still possible to perform all operations using a less-error-prone syntax, though not as performantly
3. I believe we need to think more carefully about how to model the more unusual usages of `$reduce` in light of `$group` (note that `$reduce` is 37th by operation count)

jyemin: OK, I'm going to resolve this thread.  Please make sure that this is captured somewhere so it can be used to create documentation and generate subsequent Jira tickets.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
stIncMale: Currently, the `ArrayExpression.min` implementation is

```java
this.size().eq(of(0)).cond(
    other,
    results.minN(of(1), v -> v).first())
```

What stops us from doing a similar thing in `first`?

```java
private T unsafeFist() {
    return new MqlExpression<>(astWrapped("$first")).assertImplementsAllExpressions();
}

@Override
public T first(final T other) {
    return this.size().eq(of(0)).cond(
            other,
            unsafeFist());
}
```

and then, of course, `ArrayExpression.min` becomes

```java
@Override
public <R extends Expression> R min(final R other, final Function<? super T, ? extends R> mapper) {
    MqlExpression<R> results = (MqlExpression<R>) this.map(mapper);
    return this.size().eq(of(0)).cond(other, ((MqlExpression<R>)results.minN(of(1), v -> v)).unsafeFist());
}
```

I tried it and it even works. It expectedly changes the MQL behind `ArrayExpression.first`, but the MQL for `ArrayExpression.min` stays unchanged.

jyemin (approved):

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -51,28 +50,48 @@
      */
     <R extends Expression> ArrayExpression<R> map(Function<? super T, ? extends R> in);
 
+    IntegerExpression size();
+
+    BooleanExpression any(Function<T, BooleanExpression> map);
+
+    BooleanExpression all(Function<T, BooleanExpression> map);
+
+    NumberExpression sum(Function<T, NumberExpression> map);
```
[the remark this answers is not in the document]
katcharov: Applied to multiply, sum, all, any, and join; min and max already had this.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
katcharov: I think we discussed this ("securing" the API), but agreed that it was impractical to have everywhere due to performance, and therefore not worthwhile here (due to the additional inconvenience of needing to specify the other value). [Comment here](https://github.com/mongodb/mongo-java-driver/pull/1032/files#r1048611098).

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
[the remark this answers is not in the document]
katcharov: Removed

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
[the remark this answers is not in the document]
katcharov: Updated: max, min, maxN, and minN

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
stIncMale: If the conditional logic is deemed to introduce too much overhead in `ArrayExpression.first`/`last`, then why is it not deemed to have the same affect in `ArrayExpression.min`/`max`? I see no relevant difference between these sets of methods, which is why I think that the same approach should be used in both.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -47,12 +46,8 @@ default BooleanExpression getBoolean(final String fieldName, final boolean other
...
 
-    default NumberExpression getNumber(final String fieldName, final double other) {
-        return getNumber(fieldName, of(other));
-    }
-
-    default NumberExpression getNumber(final String fieldName, final Decimal128 other) {
-        return getNumber(fieldName, of(other));
+    default NumberExpression getNumber(final String fieldName, final Number other) {
+        return getNumber(fieldName, Expressions.numberToExpression(other));
     }
```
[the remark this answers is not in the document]
katcharov: Yes https://github.com/mongodb/mongo-java-driver/pull/1032#discussion_r1048853914 . The issue was that the one approach failed to round-trip values. But the other seemingly-consistent approach led to us having both `getInteger(int)` and `getNumber(int)` - if I as a user saw both methods, I would not see the point in having both, since we seemingly only ever need `getInteger`. So there is actually a difference (which does not apply in the case of `of`), and I have reverted to the way that seemed inconsistent.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
katcharov: First and last are derivations of elementAt, which is the array version of `get` for maps/records. These are all conceptually a single method. It seems unavoidable that users would need to assert for variations of `get` (getInteger, getArray...), since (as we have discussed) forcing a default value there seems like it would greatly slow down array traversal, which is a very common usage.

But this consideration does not seem to exist for min/max, because the check is applied only once for an array rather than once per element. So we are not compelled to offer a second method for this category of methods, especially a second method that is unsafe.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ArrayExpressionsFunctionalTest.java
```diff
@@ -131,14 +131,20 @@ public void mapTest() {
...
                 Stream.of(3, 1, 2)
-                        .sorted().collect(Collectors.toList()),
-                ofIntegerArray(3, 1, 2).sort(),
+                        .sorted().collect(Collectors.toList()), sort(integerExpressionArrayExpression),
                 // MQL:
                 "{'$sortArray': {'input': [3, 1, 2], 'sortBy': 1}}");
     }
 
+    @SuppressWarnings("unchecked")
+    private ArrayExpression<? extends NumberExpression> sort(final ArrayExpression<? extends NumberExpression> array) {
```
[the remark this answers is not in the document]
katcharov: Done

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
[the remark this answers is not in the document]
stIncMale: Discussed this over Zoom. I understand the situation and the decision now.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
stIncMale: Discussed this over Zoom, and while doing so, the following realization came to mind: as long as we tend to value performance over safety (which so far was the case for this API), `min`/`max` methods are different from `first`/`last` in that they are much more computationally intensive (when called on an unsorted array, they have linear time complexity, and if the array is sorted, then hopefully the server optimizes them, but I don't know if that's what happens). Adding a trivial conditional logic to a relatively heavy `min`/`max` is fine, as it does not significantly affect their performance. But adding the same check to an extremely lightweight constant-time methods `first`/`last` may roughy double the amount of work these methods do.

stIncMale (approved):

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -395,6 +401,77 @@ public T reduce(final T initialValue, final BinaryOperator<T> in) {
                 .append("in", extractBsonValue(cr, in.apply(varValue, varThis)))));
     }
 
+    @Override
+    public BooleanExpression any(final Function<? super T, BooleanExpression> predicate) {
+        MqlExpression<BooleanExpression> map = (MqlExpression<BooleanExpression>) this.map(predicate);
```
stIncMale, line 406: Let's rename the variable `map` here and in 6 more places below to something that is less confusing because `this.map` produces an array and not a map.

stIncMale (changes requested):

stIncMale (approved):

PR #1055, opened 2022-11-23
Implement switch expression

katcharov:
JAVA-4813


driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -118,4 +120,17 @@ public interface Expression {
...
+    /**
+     * Applies the given function to this argument. Note that "apply" usually
+     * applies functions to arguments; here, the parameters are reversed.
+     *
+     * @param f
+     * @return
+     * @param <T>
+     * @param <R>
+     */
+    <T extends Expression, R extends Expression> R apply(Function<T, R> f);
```
stIncMale, line 133: [Noticed this function in https://github.com/mongodb/mongo-java-driver/pull/1059 and decided to leave a comment here]

The problem with this function is that when called on `IntegerExpression` it passes `Expression` into `f`, i.e., it erases the information about the type. Let's introduce `IntegerExpression.applyToInteger`/`applyToBoolean`/etc. methods?

katcharov: Discussed. These will have to be renamed. We considered various alternative names for `apply`: `numberApply` and `numberPassTo` are unclear, while `applyToNumber` sounds backwards. We came to `numberInto`, which has a good form, and finally agreed on `passNumberTo`, which puts the operation "pass" at the start. The length of the method is not a problem because it should not be used often, and should be noticable when it is used. Incidentally, this resolves my prior issue with using the term "apply" (the opposite of "apply" is "pass"). For switch, we'll follow the same pattern `switchNumber` and so on.

To re-iterate, the problem is that the type `T` above is not `TSelf`, and has nothing to do with the actual type that the `apply` method is invoked upon. So in the example above, an empty string is effectively cast into an integer, and then used in division. A similar problem arises with `Array<T>`, but there we at least have `T`, so we can use the `v -> v` trick for reductive operations. The approach of copying a method down is similar to what was done for add/multiply/etc. for numbers and integers, except that in this case, the method names cannot be the same due to type erasure.

driver-core/src/main/com/mongodb/client/model/expressions/Branches.java
```diff
@@ -0,0 +1,92 @@
...
+        return (MqlExpression<?>) value;
+    }
+
+    // is fn
+
+    public <T extends Expression, R extends Expression> BranchesIntermediary<T, R> is(final Function<T, BooleanExpression> o, final Function<T, R> r) {
+        return this.with(value -> new SwitchCase<>(o.apply(value), r.apply(value)));
+    }
+
+    // eq lt lte
```
jyemin, line 45: Am I correct in concluding that all of the below methods are expressible using the above `is` method?

driver-core/src/test/functional/com/mongodb/client/model/expressions/ControlExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Test
+    public void switchTypesTest() {
+        Function<IntegerExpression, StringExpression> label = expr -> expr.switchMap(on -> on
+                .isBoolean(v -> v.asString().concat(of(" - bool")))
+                .isNumber(v -> v.asString().concat(of(" - number")))
+                .isString(v -> v.asString().concat(of(" - string")))
+                .isDate(v -> v.asString().concat(of(" - date")))
+                .isArray((ArrayExpression<IntegerExpression> v) -> v.sum(a -> a).asString().concat(of(" - array")))
```
jyemin, line 79: Why does this require specifying the type of `v`?

driver-core/src/test/functional/com/mongodb/client/model/expressions/ControlExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,205 @@
...
+        assertExpression("a", of(0).apply(s));
+        assertExpression("b", of(9).apply(s));
+        assertExpression("b", of(-9).apply(s));
+        assertExpression("c", of(11).apply(s));
+        assertExpression("d", of(10).apply(s));
+    }
+
+    @Test
+    public void switchTypesTest() {
+        Function<IntegerExpression, StringExpression> label = expr -> expr.switchMap(on -> on
```
jyemin, line 74: Why is the first type parameter an `IntegerExpression`?  I don't see anything integer-y here.

driver-core/src/main/com/mongodb/client/model/expressions/Branches.java
katcharov: Yes. These, and any others that might be added to these classes, are all convenience methods.

However, the isBoolean (and other type) methods perform a cast that is not externally permitted by our API, so that the right lambda receives the value as a boolean (this is like instanceof pattern matching in Java). It would be verbose for the user otherwise.

jyemin: Got it.  I was playing around with this and not sure how to express something like this (pseudocode):

```
switch
   case isBoolean b and b is true ...
   case isBoolean b and b is false ...
   case isString s and s is null or empty ...
   case isString s ...
```

I can't quite see how it works given the type system enforced by `BranchIntermediary`. Can you clarify?

driver-core/src/test/functional/com/mongodb/client/model/expressions/ComparisonExpressionsFunctionalTest.java
```diff
@@ -60,6 +60,17 @@ class ComparisonExpressionsFunctionalTest extends AbstractExpressionsFunctionalT
             of(Instant.now())
     );
 
+    @Test
+    public void literalsTest() {
```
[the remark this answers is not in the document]
katcharov: Switch handles nulls, so I had a closer look at ofNull. I saw that there was no null literal check, and I added it here.

I generally add a missing tests for something I was using in a PR in that same PR, rather than a new PR. Let me know if there's a better way to handle this.

driver-core/src/main/com/mongodb/client/model/expressions/Branches.java
katcharov: I'm unsure how to best express this without knowing the `...` (are they more checks, or the then branch, and if so, about how many lines are ensuing) - I've assumed they are short return values. In Java, this would be:

```
if (value instanceof Boolean b) {
    return b ? "one" : "two";
} else if (value == null) {
    return "three";
} else if (value instanceof String s) {
    return "four";
}
```
So in the API:
```
expr.switchMap(on -> on
        .isBoolean(b -> b.cond(of("one"), of("two")))
        .isNull(v -> of("three"))
        .isString(s -> of("four")))
```

As with Java, `null` is not considered a String value, and the same for `missing`, so the third case in the original example is never taken. (Note that the API is non-null, and attempts to forbid even the assignment of the null value to a string.)

It seems that complex cases can always be expressed via nesting (comparable to nested if statements), and common code can be extracted into a lambda or method:

```
expr.switchMap(on -> on
    .isString(s -> s.switchMap(on2 -> on2
        .eq(v1, lambda1)
        .eq(v2, lambda2)
    .isBoolean(b -> ...
```

Since everything can be expressed, the remaining question is whether we can offer a better second way of expressing the above. When I've fleshed out possible examples, I've found they don't make sense (often because Java has an infinite number of types that can be instance-of'd, but we have < 10, so the reasonable usages are different), or the examples are at least equally-well expressed using what we have in the API.

driver-core/src/test/functional/com/mongodb/client/model/expressions/ControlExpressionsFunctionalTest.java
katcharov: Alternatively:
`.<IntegerExpression>isArray(v -> ...`

`isArray` checks that the value is an array, but cannot check the type of the array (I see no reasonable way to do this). So the method is responsible for checking whether it's an array, but the user is responsible for "asserting" the type of the array (much like they assert the type during doc.getInteger).

katcharov:
[56e8756](https://github.com/mongodb/mongo-java-driver/pull/1055/commits/56e8756c41ace4918119de190cf2ca96d37998d2) is new, first commit is rebased.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -114,4 +116,9 @@ public interface Expression {
     <T extends Expression> MapExpression<T> isMapOr(MapExpression<T> other);
 
     StringExpression asString();
+
+    <R extends Expression> R passTo(Function<? super Expression, R> f);
+
+    <R extends Expression> R switchOn(Function<Branches, ? extends BranchesTerminal<? super Expression, R>> on);
```
stIncMale, line 122: I think it should be `? extends R` here and in the rest of `switch` methods.

stIncMale (changes requested):
`MqlExpression` and tests are yet to be reviewed.

stIncMale (changes requested):
The `switch[Type]On` + `Branches*` is the most complex generic structure I've ever encountered, and probably for the first time in a long time a structure that I absolutely fail to reason about. GJ raising the complexity to another level 💪

driver-core/src/test/functional/com/mongodb/client/model/expressions/ControlExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,244 @@
...
+    public void switchTest() {
+        // https://www.mongodb.com/docs/manual/reference/operator/aggregation/switch/
+        assertExpression("a", of(0).switchOn(on -> on.is(v -> v.eq(of(0)), v -> of("a"))));
+        assertExpression("a", of(0).switchOn(on -> on.isNumber(v -> of("a"))));
+        assertExpression("a", of(0).switchOn(on -> on.lte(of(9), v -> of("a"))));
+
+        // test branches
+        Function<IntegerExpression, BooleanExpression> isOver10 = v -> v.subtract(10).gt(of(0));
+        Function<IntegerExpression, StringExpression> s = e -> e
+                .switchIntegerOn((Branches on) -> on
```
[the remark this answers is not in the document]
katcharov: I think this is the only place where this was done.

driver-core/src/main/com/mongodb/client/model/expressions/Branches.java
```diff
@@ -0,0 +1,92 @@
...
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+
+public final class Branches {
```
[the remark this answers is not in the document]
katcharov: Right. I think there is only one.

driver-core/src/main/com/mongodb/client/model/expressions/Branches.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public <T extends Expression, R extends Expression> BranchesIntermediary<T, R> lte(final T v, final Function<T, R> r) {
+        return this.with(value -> new SwitchCase<>(value.lte(v), r.apply(value)));
+    }
+
+    // is type
+
+    public <T extends Expression, R extends Expression> BranchesIntermediary<T, R> isBoolean(final Function<BooleanExpression, R> r) {
+        return this.with(value -> new SwitchCase<>(mqlEx(value).isBoolean(), r.apply((BooleanExpression) value)));
+    }
```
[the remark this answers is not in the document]
katcharov: I have used something like:

`return is(v -> mqlEx(v).isMap(), v -> r.apply((MapExpression<Q>) v));`

In I hope all the relevant places.

driver-core/src/main/com/mongodb/client/model/expressions/Branches.java
```diff
@@ -0,0 +1,92 @@
...
+        return new BranchesIntermediary<>(v);
+    }
+
+    private static <T extends Expression> MqlExpression<?> mqlEx(final T value) {
+        return (MqlExpression<?>) value;
+    }
+
+    // is fn
+
+    public <T extends Expression, R extends Expression> BranchesIntermediary<T, R> is(final Function<T, BooleanExpression> o, final Function<T, R> r) {
```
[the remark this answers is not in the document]
katcharov: I believe I have made all of these changes in the relevant places (though the params varied).

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -114,4 +116,9 @@ public interface Expression {
     <T extends Expression> MapExpression<T> isMapOr(MapExpression<T> other);
 
     StringExpression asString();
+
+    <R extends Expression> R passTo(Function<? super Expression, ? extends R> f);
+
+    <R extends Expression> R switchOn(Function<Branches, ? extends BranchesTerminal<? super Expression, ? extends R>> on);
```
stIncMale, line 122: [I don't know what to do with this, probably - nothing]

I noticed that the following code compiles:

```java
of(1).switchOn(on -> on
        .eq(of(0), v -> of("true")))
```

but stops compiling if I add the default branch:

```java
of(1).switchOn(on -> on
        .eq(of(0), v -> of("true"))
        .defaults(v -> of("false")))
```

This is very confusing even to me, despite me having reviewed the code. The technical reason behind the error is that since `of(0)` produced `IntegerExpression`, `eq` returns `BranchesTerminal<IntegerExpression, ...>`, which makes `default` to expect `IntegerExpression` as its input. On the other hand, `switchOn` accepts `Function<Branches, ? extends BranchesTerminal<? super Expression, R>>`, i.e., `default` is guaranteed to be provided with an `Expression`, but not necessary `IntegerExpression`, which is what the IntelliJ's compiler unsuccessfully trying to tell us by saying _"no instance(s) of type variable(s) exist so that `Expression` conforms to `IntegerExpression`"_.

Now one may say _"but why doesn't the compiler think that `eq` produces `BranchesTerminal<IntegerExpression, ...>` in the first example that compiles?"_ To this my answer is - I have no idea, the answer is likely in the guts of the JLS type inference rules, and I know nothing about them.

The code can be fixed by either of the two following ways:

```java
of(1).switchOn(on -> on
        .eq((Expression) of(0), v -> of("true"))
        .defaults(v -> of("false")));
```

```java
of(1).switchIntegerOn(on -> on
        .eq(of(0), v -> of("true"))
        .defaults(v -> of("false")));
```

driver-core/src/main/com/mongodb/client/model/expressions/BranchesIntermediary.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public <Q extends Expression> BranchesIntermediary<T, R> isMap(final Function<MapExpression<Q>, R> r) {
+        return this.with(value -> new SwitchCase<>(mqlEx(value).isDocument(), r.apply((MapExpression<Q>) value)));
+    }
+
+    public BranchesIntermediary<T, R> isNull(final Function<Expression, R> r) {
+        return this.with(value -> new SwitchCase<>(mqlEx(value).isNull(), r.apply(value)));
+    }
+
+    public BranchesTerminal<T, R> defaults(final Function<T, R> r) {
```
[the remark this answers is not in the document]
stIncMale: It is still `Function<T, ...>`, not `Function<? super T, ...>`.

driver-core/src/main/com/mongodb/client/model/expressions/Branches.java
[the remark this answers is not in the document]
stIncMale: The following places still need a change:

- `Branches`/`BranchesIntermediary.is[Type]` need `Function<? super [Type], ...>` instead of `Function<[Type], ...>`.
- `BranchesIntermediary.is` needs `Function<? super T, BooleanExpression>` instead of `Function<T, BooleanExpression>`.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: This was not noticeable in the original implementation because it allowed usages that were too broad (not typesafe). Now, in the following, switchOn expects the "input" type (T) to be an Expression, but the `eq` forces an inference of IntegerEx. 

```
of(1).switchOn(on -> on
        .eq(of(0), v -> of("true"))
        .defaults(v -> of("false")))
```

But we should already know the type of the input, since we now have a variety of methods to infer the type from. So I have added a T to Branches, which I should have done when I added those various switchOn methods.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
```diff
@@ -442,12 +442,13 @@ public NumberExpression isNumberOr(final NumberExpression other) {
...
 
     public BooleanExpression isInteger() {
-        return this.isNumber().cond(this.eq(this.round()), of(false));
+        return this.isNumber().and(this.eq(this.round()));
     }
 
     @Override
     public IntegerExpression isIntegerOr(final IntegerExpression other) {
-        return this.isInteger().cond(this, other);
+        return this.isNumber().and(this.eq(this.round()))
```
stIncMale, line 450: It appears that this method still can be implemented as `return isInteger().cond(this, other);`, I don't think a change is needed here.

driver-core/src/test/functional/com/mongodb/client/model/expressions/TypeExpressionsFunctionalTest.java
```diff
@@ -65,6 +65,27 @@ public void isNumberOrTest() {
...
+        assertExpression(Decimal128.parse("1"), of(Decimal128.parse("1")).isIntegerOr(of(99)));
+        // non-numeric:
+        assertExpression(99, ofIntegerArray(1).isIntegerOr(of(99)));
+        assertExpression(99, ofNull().isIntegerOr(of(99)));
+
+        assertExpression(99,
+                of(BsonDocument.parse("{'a': 'str'}")).getField("a").isIntegerOr(of(99)));
+        // the following leads to "$round only supports numeric types, not string"
+        // because the literal type is known.
+        // assertExpression(99, of("str").isIntegerOr(of(99)));
```
stIncMale, line 86: So `$and` there is no short-circuiting in MongoDB that would made this work. This behavior violates the documentation for `Expression.isIntegerOr` added in https://github.com/mongodb/mongo-java-driver/pull/1059. We can't even say "the user asserts that the value is number" here because it is an unexpected requirement given methods like `isBooleanOr`,`isDateOr`, etc.

Currently I see the following ways we could attempt to solve the problem:

1. Attempt to somehow restructure the produced MQL such that MongoDB does not try to `$round` value that is not a number (this is assuming it errors not as a result of pure validating of the MQL and the input data).
2. Attempt to use `$convert` to convert before rounding, and then `$eq` the original and the converted&rounded value.

Since `Expression.isIntegerOr` was not added in this PR, I can move this comment to `Expressions`. Either way, this seems like an important thing to address with regard to `isIntegerOr`.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
```diff
@@ -46,5 +46,5 @@ default IntegerExpression subtract(final int subtract) {
     IntegerExpression abs();
 
     <R extends Expression> R passIntegerTo(Function<? super IntegerExpression, ? extends R> f);
-    <R extends Expression> R switchIntegerOn(Function<Branches, ? extends BranchesTerminal<? super IntegerExpression, ? extends R>> on);
+    <R extends Expression> R switchIntegerOn(Function<Branches<IntegerExpression>, ? extends BranchesTerminal<? super IntegerExpression, ? extends R>> on);
```
stIncMale, line 49: `switchIntegerOn` always passes `Branches<IntegerExpression>` as the input for the `on` function. Therefore, it appears that we'd like to accept `on` functions that take either `Branches<IntegerExpression>` or a supertype of this type, i.e., `Function<? super Branches<IntegerExpression>, ...>`.  This indeed, would have given our users more flexibility:


```java
// if a user has a function like the following
Function<Branches<? extends NumberExpression>, BranchesTerminal<NumberExpression, StringExpression>> branches = null;
// then the user could use it in both of the following pieces of code
of(0).switchNumberOn(f);
of(0).switchIntegerOn(f);
// btw, this code also requires `? super IntegerExpression` in `? extends BranchesTerminal<? super IntegerExpression, ...>`, which we currently have
```

However, similarly to (@stIncMale TODO past the link), it's impossible to obtain `BranchesTerminal<NumberExpression, ...>>` from `Branches<? extends NumberExpression>`, i.e., users can never implement the `branches` function in the above code. Therefore, it appears to be fine to accept just `Function<Branches<IntegerExpression>, ...>` in `switchNumberOn`.

stIncMale (changes requested):

driver-core/src/test/functional/com/mongodb/client/model/expressions/TypeExpressionsFunctionalTest.java
katcharov: `and`, `or`, `cond` (and I think others) avoid evaluating the right branch, unless necessary. However, this does not apply if you provide certain literal values. This is bad, because it should know that the branches are not evaluated. But it is good, because the user should probably not be sending certain non-evaluated expressions.

So this only applies if the user is creating expressions that they (or we) should probably be evaluating locally. And any existing code will already work this way. I do not think we should attempt to convert all values, since this is not a problem for all values.

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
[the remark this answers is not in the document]
katcharov: Removing `? super`

driver-core/src/test/functional/com/mongodb/client/model/expressions/TypeExpressionsFunctionalTest.java
katcharov: Used switch, since unlike cond it avoids this problem.

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
katcharov: Used switch.

driver-core/src/test/functional/com/mongodb/client/model/expressions/TypeExpressionsFunctionalTest.java
stIncMale: Shouldn't we use a similar approach in `isInteger`:

```java
public BooleanExpression isInteger() {
    return switchOn(on -> on
            .isNumber(v -> v.round().eq(v))
            .defaults(v -> of(false)));
}
```

to avoid similar problems:

```java
Expression yearManufacturedOrGarbage = of("");
yearManufacturedOrGarbage.switchOn(on -> on
        .isDate(v -> v.year(of("UTC")))
        // this causes an error at the moment, but works if `isNumber` is changed
        .isInteger(v -> v)
        .defaults(v -> of(0))));
```

driver-core/src/main/com/mongodb/client/model/expressions/MqlExpression.java
stIncMale: I meant that `isIntegerOr` could still be implemented via `isInteger`. But that no longer applies.

stIncMale (changes requested):
<a useless comment forced by GitHub>

driver-core/src/test/functional/com/mongodb/client/model/expressions/TypeExpressionsFunctionalTest.java
katcharov: Yes - fixed

stIncMale: We also need a test for `isInteger` covering this. One was added for `isIntegerOr` in https://github.com/mongodb/mongo-java-driver/pull/1055/commits/8e0ad99c443c48a73691cfc3206561f45c92e54c for a similar change.

katcharov: Updated. The prior test checked for this.

stIncMale (approved):

jyemin (approved):

PR #1057, opened 2022-11-29
Test expressions in context

katcharov:
JAVA-4820

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -172,6 +172,11 @@ public final class Expressions {
         return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonArray(result)));
     }
 
+    public static <R extends DocumentExpression> R current() {
+        return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonString("$$CURRENT")))
+                .assertImplementsAllExpressions();
+    }
+
     @SafeVarargs  // nothing is stored in the array
     public static <T extends Expression> ArrayExpression<T> ofArray(final T... array) {
         Assertions.notNull("array", array);
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -172,7 +172,7 @@ public final class Expressions {
         return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonArray(result)));
     }
 
-    public static <R extends DocumentExpression> R current() {
+    public static DocumentExpression current() {
         return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonString("$$CURRENT")))
                 .assertImplementsAllExpressions();
     }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -177,14 +177,6 @@ public final class Expressions {
                 .assertImplementsAllExpressions();
     }
 
-    /**
-     * Returns the "current" value as a {@linkplain MapExpression map} value.
-     * The "current" value is the top-level document currently being processed
-     * in the aggregation pipeline stage.
-     *
-     * @return the resulting value
-     * @param <R> the type of the resulting value
-     */
     public static <R extends Expression> MapExpression<R> currentAsMap() {
         return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonString("$$CURRENT")))
                 .assertImplementsAllExpressions();
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
@@ -177,6 +177,19 @@ public final class Expressions {
                 .assertImplementsAllExpressions();
     }
 
+    /**
+     * Returns the "current" value as a {@linkplain MapExpression map} value.
+     * The "current" value is the top-level document currently being processed
+     * in the aggregation pipeline stage.
+     *
+     * @return the resulting value
+     * @param <R> the type of the resulting value
+     */
+    public static <R extends Expression> MapExpression<R> currentAsMap() {
+        return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonString("$$CURRENT")))
+                .assertImplementsAllExpressions();
+    }
+
     @SafeVarargs  // nothing is stored in the array
     public static <T extends Expression> ArrayExpression<T> ofArray(final T... array) {
         Assertions.notNull("array", array);
--- a/driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.client.model.expressions;
+
+import com.mongodb.MongoClientSettings;
+import com.mongodb.client.AggregateIterable;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.model.Aggregates;
+import com.mongodb.client.model.Filters;
+import org.bson.Document;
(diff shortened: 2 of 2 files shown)


driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Test
+    public void matchTest() {
+        col.insertMany(Arrays.asList(
+                Document.parse("{_id: 1, x: 0, y: 2}"),
+                Document.parse("{_id: 2, x: 0, y: 3}"),
+                Document.parse("{_id: 3, x: 1, y: 3}")));
+
+        List<Document> results = aggregate(
+                match(Filters.expr(current().getInteger("x").eq(of(1)))));
```
katcharov, line 85: Was reading docs `match(expr(new Document("$and"`, seems that we might add a version of expr such that the above would become:

`Filters.expr(d -> d.getInteger("x").eq(of(1))`

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -172,6 +172,11 @@ public static ArrayExpression<StringExpression> ofStringArray(final String... ar
         return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonArray(result)));
     }
 
+    public static <R extends DocumentExpression> R current() {
```
jyemin, line 175: Why do we need a generic parameter here?  Do we expect `DocumentExpression` to be sub-typed?

jyemin, line 175: There are times that users would need access to the current document as a `MapExpression`.  How would that be done?

(feel free to wait on this until after the map expression PR is merged)

katcharov: Right. The sub-type would allow a helper class (aka schema class, binding glass) to condense:
`doc.getDocument("c").<IntegerExpression>getArray("a").first()`
to
`doc.c.a.first()`

katcharov: I think this is now possible via `current().isMapOr(ofMap(parse("{}")))`. This isn't great, but (while certainly possible) it seems very unusual for documents to be maps.

jyemin: It's not unusual for a document to be a `MapExpression<Expression>`. It always is.

katcharov: This is true in a certain sense, but I do not think that sense is relevant. For example, technically a Java object "is" (is representable as) a `Map<String, Object>`. But to use it as such, users must use the inconvenient reflection API.

A usage seems improbable if all of us have trouble imagining semi-realistic cases where users would have to store things in collections that need to be used as if they were maps. This would mean that they are storing things that have an actual identity (expressed in the `_id` field), and then some completely arbitrary fields (that is, the programmer cannot hard-code at least some of the other fields, and none of those fields were instead represented as a map on the thing). And this is happening often enough that the isMapOr conversion is burdensome.

driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,160 @@
...
+        col = client.getDatabase("testdb").getCollection("testcol");
+        col.drop();
+    }
+
+    @AfterEach
+    public void tearDown() {
+        client.close();
+    }
+
+    private String bsonToString(final Bson project) {
```
stIncMale, line 66: Can be `static`.

driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,160 @@
...
+                of(0), of(1), of(true), of(false)
+        ))));
+        assertEquals(
+                Arrays.asList(Document.parse("{ nestedArray: [ 4, 1, 0, 1, true, false ] }")),
+                aggregate(projectNestedArray));
+
+        // new, document
+        Bson projectDocument = project(fields(computed("nested",
+                //BsonDocument.parse("{ x: {$max : ['$x', 4] }}")
+                of(new Document()).setField("x", current().getInteger("x").max(of(4)))
```
stIncMale, line 135: I was initially confused by `BsonDocument.parse` and thought this is a commented out code that should not be here, but then I realized that it's a relaxed JSON view of what we pass to `computed`. Could we remove the `BsonDocument.parse` part and leave only JSON?

driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
stIncMale: I am imagining that this could be something for the future.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: We discussed this and decided to go with inheritance (which would have been done in #1054), but then went back to non-inheritance due to an implementation problem that is ultimately caused by type erasure.

Added currentAsMap.

driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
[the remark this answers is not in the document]
katcharov: I'm not sure what to think of empty documents - it seems that we could leave them out for now. I don't think the example in the test needs to be particularly realistic, but I updated it to be a non-empty document. (I did not use ofEmptyMap because I don't expect that it applies anymore, due to non-inheritance).

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -177,6 +177,19 @@ public static <R extends DocumentExpression> R current() {
...
     }
 
+    /**
+     * Returns the "current" value as a {@linkplain MapExpression map} value.
+     * The "current" value is the top-level document currently being processed
+     * in the aggregation pipeline stage.
+     *
+     * @return the resulting value
+     * @param <R> the type of the resulting value
+     */
```
stIncMale, line 187: [optional]

This text is different from that in https://github.com/mongodb/mongo-java-driver/pull/1059.

driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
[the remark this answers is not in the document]
stIncMale: Yes, if we are not doing `of(new Document())`, it appears to be fine to not introduce `DocumentExpression of()`.

driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
```diff
@@ -0,0 +1,195 @@
...
+                aggregate(projectNestedArray));
+
+        // new, document
+        Bson projectDocument = project(fields(computed("nested",
+                // the below is roughly: "{ x: {$max : ['$x', 4] }}"
+                of(Document.parse("{x: 9}")).setField("x", current().getInteger("x").max(of(4)))
+        )));
+        assertEquals(
+                Arrays.asList(Document.parse("{_id: 0, nested: { x: 4 } }")),
+                aggregate(projectDocument));
```
stIncMale, line 174: Shouldn't `nested.x` be set to the max of 9 and 4 here? Based on the assertion it is set to 4 for some reason.

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/client/model/expressions/InContextExpressionsFunctionalTest.java
katcharov: `of(Document.parse("{x: 9}"))` is not the current document, it is a literal. It is having its `x` value correctly set ("overwritten") to the maximum of 1 and 4.

stIncMale: Thank you, I was reading this absolutely incorrectly.

stIncMale (approved):

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -172,6 +172,16 @@ public static ArrayExpression<StringExpression> ofStringArray(final String... ar
         return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonArray(result)));
     }
 
+    public static <R extends DocumentExpression> R current() {
```
stIncMale, line 175: While we don't know how will `Expression.isDocumentOr` return values of user-supplied subtypes of `DocumentExpression`, or whether it will even be possible to implement the idea of user subtypes, it at least it does not immediately seem absolutely impossible.

Returning a value of a user-defined type from a `static` method `Expressions.current()` is, on the other hand, absolutely impossible, provided that we exclude any kind of global state meddling (which we should). Thus, `Expressions.current` should return `DocumentExpression`.

stIncMale (changes requested):

stIncMale (approved):

katcharov:
Rebased, no conflicts

jyemin (approved):
LGTM

PR #1058, opened 2022-11-30
Move check for previous value of "closed" outside of lambda

jyemin:
JAVA-4821


lem21h:
Wouldn't it be simpler to use AtomicBolean?

rozza (approved):
LGTM

PR #1059, opened 2022-12-01
Javadoc for expressions

katcharov:
Partial work for JAVA-4799

Covers boolean, date, number, integer, and expression, but not others, since PRs are still in progress.

There are naming changes, to parameters especially.

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -21,35 +21,39 @@ import java.util.function.Function;
 import static com.mongodb.client.model.expressions.Expressions.of;
 
 /**
- * Expresses an array value. An array value is a finite, ordered collection of
- * elements of a certain type.
+ * An array {@link Expression value} in the context of the MongoDB Query
+ * Language (MQL). An array is a finite, ordered collection of elements of a
+ * certain type. It is also known as a finite mathematical sequence.
  *
- * @param <T> the type of the elements in the array
+ * @param <T> the type of the elements
  */
 public interface ArrayExpression<T extends Expression> extends Expression {
 
     /**
-     * Returns an array consisting of those elements in this array that match
-     * the given predicate condition. Evaluates each expression in this array
-     * according to the cond function. If cond evaluates to logical true, then
-     * the element is preserved; if cond evaluates to logical false, the element
-     * is omitted.
+     * An array consisting of only those elements in {@code this} array that
+     * match the provided predicate.
      *
-     * @param cond the function to apply to each element
-     * @return the new array
+     * @param predicate the predicate to apply to each element to determine if
+     *                  it should be included.
+     * @return the resulting array.
      */
-    ArrayExpression<T> filter(Function<? super T, ? extends BooleanExpression> cond);
+    ArrayExpression<T> filter(Function<? super T, ? extends BooleanExpression> predicate);
 
     /**
-     * Returns an array consisting of the results of applying the given function
-     * to the elements of this array.
+     * An array consisting of the results of applying the provided function to
+     * the elements of {@code this} array.
      *
-     * @param in the function to apply to each element
-     * @return the new array
-     * @param <R> the type contained in the resulting array
+     * @param in the function to apply to each element.
+     * @return the resulting array.
+     * @param <R> the type of the elements of the resulting array.
      */
     <R extends Expression> ArrayExpression<R> map(Function<? super T, ? extends R> in);
 
+    /**
+     * The size of {@code this} array.
+     *
+     * @return the size.
+     */
     IntegerExpression size();
 
     BooleanExpression any(Function<? super T, BooleanExpression> predicate);
@@ -113,7 +117,25 @@ public interface ArrayExpression<T extends Expression> extends Expression {
 
     ArrayExpression<T> distinct();
 
+    /**
+     * The result of passing {@code this} value to the provided function.
+     * Equivalent to {@code f.apply(this)}, and allows lambdas and static,
+     * user-defined functions to use the chaining syntax.
+     *
+     * @see Expression#passTo
+     * @param f the function to apply.
+     * @return the resulting value.
+     * @param <R> the type of the resulting value.
+     */
     <R extends Expression> R passArrayTo(Function<? super ArrayExpression<T>, ? extends R> f);
 
-    <R extends Expression> R switchArrayOn(Function<Branches<ArrayExpression<T>>, ? extends BranchesTerminal<ArrayExpression<T>, ? extends R>> on);
+    /**
+     * The result of applying the provided switch mapping to {@code this} value.
+     *
+     * @see Expression#switchOn
+     * @param mapping the switch mapping.
+     * @return the resulting value.
+     * @param <R> the type of the resulting value.
+     */
+    <R extends Expression> R switchArrayOn(Function<Branches<ArrayExpression<T>>, ? extends BranchesTerminal<ArrayExpression<T>, ? extends R>> mapping);
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -19,6 +19,8 @@ package com.mongodb.client.model.expressions;
 import java.util.function.Function;
 
 import static com.mongodb.client.model.expressions.Expressions.of;
+import static com.mongodb.client.model.expressions.MqlUnchecked.Unchecked.PRESENT;
+import static com.mongodb.client.model.expressions.MqlUnchecked.Unchecked.NON_EMPTY;
 
 /**
  * An array {@link Expression value} in the context of the MongoDB Query
@@ -85,6 +87,7 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      * @param i
      * @return
      */
+    @MqlUnchecked(PRESENT)
     T elementAt(IntegerExpression i);
 
     default T elementAt(final int i) {
@@ -95,6 +98,7 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      * user asserts that array is not empty
      * @return
      */
+    @MqlUnchecked(NON_EMPTY)
     T first();
 
     /**
--- a/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
@@ -19,44 +19,41 @@ package com.mongodb.client.model.expressions;
 import java.util.function.Function;
 
 /**
- * Expresses a boolean value.
+ * A logical boolean value, either true or false.
  */
 public interface BooleanExpression extends Expression {
 
     /**
-     * Returns logical true if this expression evaluates to logical false.
-     * Returns logical false if this expression evaluates to logical true.
+     * The logical negation of {@code this} value.
      *
-     * @return True if false; false if true.
+     * @return the resulting value.
      */
     BooleanExpression not();
 
     /**
-     * Returns logical true if this or the other expression evaluates to logical
-     * true. Returns logical false if both evaluate to logical false.
+     * The logical conjunction of {@code this} and the {@code other} value.
      *
-     * @param or the other boolean expression.
-     * @return True if either true, false if both false.
+     * @param other the other boolean value.
+     * @return the resulting value.
      */
-    BooleanExpression or(BooleanExpression or);
+    BooleanExpression or(BooleanExpression other);
 
     /**
-     * Returns logical true if both this and the other expression evaluate to
-     * logical true. Returns logical false if either evaluate to logical false.
+     * The logical disjunction of {@code this} and the {@code other} value.
      *
-     * @param and the other boolean expression.
-     * @return true if both true, false if either false.
+     * @param other the other boolean value.
+     * @return the resulting value.
      */
-    BooleanExpression and(BooleanExpression and);
+    BooleanExpression and(BooleanExpression other);
+    // TODO-END check the evaluation semantics of and/or
 
     /**
-     * If this expression evaluates to logical true, returns the result of the
-     * evaluated left branch expression. If this evaluates to logical false,
-     * returns the result of the evaluated right branch expression.
+     * The {@code left} branch when {@code this} is true,
+     * and the {@code right} branch otherwise.
      *
-     * @param left the left branch expression
-     * @param right the right branch expression
-     * @return left if true, right if false.
+     * @param left the left branch.
+     * @param right the right branch.
+     * @return the resulting value.
      * @param <T> The type of the resulting expression.
      */
     <T extends Expression> T cond(T left, T right);
--- a/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
@@ -19,7 +19,8 @@ package com.mongodb.client.model.expressions;
 import java.util.function.Function;
 
 /**
- * A logical boolean value, either true or false.
+ * A boolean {@linkplain Expression value} in the context of the
+ * MongoDB Query Language (MQL).
  */
 public interface BooleanExpression extends Expression {
 
@@ -45,20 +46,37 @@ public interface BooleanExpression extends Expression {
      * @return the resulting value.
      */
     BooleanExpression and(BooleanExpression other);
-    // TODO-END check the evaluation semantics of and/or
 
     /**
-     * The {@code left} branch when {@code this} is true,
-     * and the {@code right} branch otherwise.
+     * The {@code ifTrue} value when {@code this} is true,
+     * and the {@code ifFalse} value otherwise.
      *
-     * @param left the left branch.
-     * @param right the right branch.
+     * @param ifTrue the ifTrue value.
+     * @param ifFalse the ifFalse value.
      * @return the resulting value.
      * @param <T> The type of the resulting expression.
      */
-    <T extends Expression> T cond(T left, T right);
+    <T extends Expression> T cond(T ifTrue, T ifFalse);
 
+    /**
+     * The result of passing {@code this} value to the provided function.
+     * Equivalent to {@code f.apply(this)}, and allows lambdas and static,
+     * user-defined functions to use the chaining syntax.
+     *
+     * @see Expression#passTo
+     * @param f the function to apply.
+     * @return the resulting value.
+     * @param <R> the type of the resulting value.
+     */
     <R extends Expression> R passBooleanTo(Function<? super BooleanExpression, ? extends R> f);
 
-    <R extends Expression> R switchBooleanOn(Function<Branches<BooleanExpression>, ? extends BranchesTerminal<BooleanExpression, ? extends R>> on);
+    /**
+     * The result of applying the provided switch mapping to {@code this} value.
+     *
+     * @see Expression#switchOn
+     * @param mapping the switch mapping.
+     * @return the resulting value.
+     * @param <R> the type of the resulting value.
+     */
+    <R extends Expression> R switchBooleanOn(Function<Branches<BooleanExpression>, ? extends BranchesTerminal<BooleanExpression, ? extends R>> mapping);
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
@@ -19,8 +19,11 @@ package com.mongodb.client.model.expressions;
 import java.util.function.Function;
 
 /**
- * An instantaneous date and time. Tracks a UTC datetime, the number of
- * milliseconds since the Unix epoch. Does not track the timezone.
+ * A UTC date-time {@linkplain Expression value} in the context
+ * of the MongoDB Query Language (MQL). Tracks the number of
+ * milliseconds since the Unix epoch, and does not track the timezone.
+ *
+ * @mongodb.driver.manual reference/operator/aggregation/dateToString/ Format Specifiers, UTC Offset, and Olson Timezone Identifier
  */
 public interface DateExpression extends Expression {
 
@@ -123,11 +126,30 @@ public interface DateExpression extends Expression {
      * provided {@code timezone}, and formatted according to the {@code format}.
      *
      * @param timezone the UTC Offset or Olson Timezone Identifier.
-     * @param format the format specifier. TODO-END what standard is this?
+     * @param format the format specifier.
      * @return the resulting value.
      */
     StringExpression asString(StringExpression timezone, StringExpression format);
 
+    /**
+     * The result of passing {@code this} value to the provided function.
+     * Equivalent to {@code f.apply(this)}, and allows lambdas and static,
+     * user-defined functions to use the chaining syntax.
+     *
+     * @see Expression#passTo
+     * @param f the function to apply.
+     * @return the resulting value.
+     * @param <R> the type of the resulting value.
+     */
     <R extends Expression> R passDateTo(Function<? super DateExpression, ? extends R> f);
-    <R extends Expression> R switchDateOn(Function<Branches<DateExpression>, ? extends BranchesTerminal<DateExpression, ? extends R>> on);
+
+    /**
+     * The result of applying the provided switch mapping to {@code this} value.
+     *
+     * @see Expression#switchOn
+     * @param mapping the switch mapping.
+     * @return the resulting value.
+     * @param <R> the type of the resulting value.
+     */
+    <R extends Expression> R switchDateOn(Function<Branches<DateExpression>, ? extends BranchesTerminal<DateExpression, ? extends R>> mapping);
 }
--- a/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
@@ -19,20 +19,113 @@ package com.mongodb.client.model.expressions;
 import java.util.function.Function;
 
 /**
- * Expresses a date value.
+ * An instantaneous date and time. Tracks a UTC datetime, the number of
+ * milliseconds since the Unix epoch. Does not track the timezone.
  */
 public interface DateExpression extends Expression {
+
+    /**
+     * The year of {@code this} date as determined by the provided
+     * {@code timezone}.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression year(StringExpression timezone);
+
+    /**
+     * The month of {@code this} date as determined by the provided
+     * {@code timezone}, as an integer between 1 and 12.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression month(StringExpression timezone);
+
+    /**
+     * The day of the month of {@code this} date as determined by the provided
+     * {@code timezone}, as an integer between 1 and 31.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression dayOfMonth(StringExpression timezone);
+
+    /**
+     * The day of the week of {@code this} date as determined by the provided
+     * {@code timezone}, as an integer between 1 (Sunday) and 7 (Saturday).
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression dayOfWeek(StringExpression timezone);
+
+    /**
+     * The day of the year of {@code this} date as determined by the provided
+     * {@code timezone}, as an integer between 1 and 366.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression dayOfYear(StringExpression timezone);
+
+    /**
+     * The hour of {@code this} date as determined by the provided
+     * {@code timezone}, as an integer between 0 and 23.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression hour(StringExpression timezone);
+
+    /**
+     * The minute of {@code this} date as determined by the provided
+     * {@code timezone}, as an integer between 0 and 59.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression minute(StringExpression timezone);
+
+    /**
+     * The second of {@code this} date as determined by the provided
+     * {@code timezone}, as an integer between 0 and 59, and 60 in the case
+     * of a leap second.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression second(StringExpression timezone);
+
+    /**
+     * The week of the year of {@code this} date as determined by the provided
+     * {@code timezone}, as an integer between 0 and 53.
+     *
+     * <p>Weeks begin on Sundays, and week 1 begins with the first Sunday of the
+     * year. Days preceding the first Sunday of the year are in week 0.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression week(StringExpression timezone);
+
+    /**
+     * The millisecond part of {@code this} date as determined by the provided
+     * {@code timezone}, as an integer between 0 and 999.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     IntegerExpression millisecond(StringExpression timezone);
 
+    /**
+     * The string representation of {@code this} date as determined by the
+     * provided {@code timezone}, and formatted according to the {@code format}.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @param format the format specifier. TODO-END what standard is this?
+     * @return the resulting value.
+     */
     StringExpression asString(StringExpression timezone, StringExpression format);
 
     <R extends Expression> R passDateTo(Function<? super DateExpression, ? extends R> f);
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -20,6 +20,8 @@ import com.mongodb.annotations.Evolving;
 
 import java.util.function.Function;
 
+import static com.mongodb.client.model.expressions.MqlUnchecked.Unchecked.TYPE_ARGUMENT;
+
 /**
  * A value in the context of the MongoDB Query Language (MQL).
  *
@@ -215,7 +217,7 @@ public interface Expression {
      * @return the resulting value.
      * @param <T> the type of the elements of the resulting array.
      */
-    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<? extends T> other);
+    <T extends Expression> ArrayExpression<@MqlUnchecked(TYPE_ARGUMENT) T> isArrayOr(ArrayExpression<? extends T> other);
 
     /**
      * {@code this} value as a {@linkplain DocumentExpression document} if
--- a/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
@@ -21,11 +21,67 @@ import com.mongodb.annotations.Evolving;
 import java.util.function.Function;
 
 /**
- * <p>Users should treat these interfaces as sealed, and must not implement any
- * sub-interfaces.
+ * A value in the context of the MongoDB Query Language (MQL).
  *
- * TODO-END: 'cause an error', 'execution context', wrong types/unsafe operations
- * TODO-END: types and how missing/null are not part of any type.
+ * <p>The API provided by this base type and its subtypes is the Java-native
+ * variant of MQL. It is used to query the MongoDB server, to perform remote
+ * computations, to store and retrieve data, or to otherwise work with data on
+ * a MongoDB server or compatible execution context. Though the methods exposed
+ * through this API generally correspond to MQL operations, this correspondence
+ * is not exact.
+ *
+ * <p>The following is an example of usage within an aggregation pipeline. Here,
+ * the current document value is obtained and its "numberArray" field is
+ * filtered and summed, in a style similar to that of the Java Stream API:
+ *
+ * <pre>{@code
+ * import static com.mongodb.client.model.expressions.Expressions.current;
+ * MongoCollection<Document> col = ...;
+ * AggregateIterable<Document> result = col.aggregate(Arrays.asList(
+ *     addFields(new Field<>("result", current()
+ *         .<MqlNumber>getArray("numberArray")
+ *         .filter(v -> v.gt(of(0)))
+ *         .sum(v -> v)))));
+ * }</pre>
+ *
+ * <p>Values are typically initially obtained via the current document and its
+ * fields, or specified via statically-imported methods on the
+ * {@link Expressions} class.
+ *
+ * <p>As with the Java Stream API's terminal operations, corresponding Java
+ * values are not directly available, but must be obtained indirectly via
+ * {@code MongoCollection.aggregate} or {@code MongoCollection.find}.
+ * Certain methods may cause an error, which will be produced
+ * through these "terminal operations".
+ *
+ * <p>The null value is not part of, and cannot be used as if it were part
+ * of, any explicit type (except the root type {@link Expression} itself).
+ * See {@link Expressions#ofNull} for more details.
+ *
+ * <p>This API specifies no "missing" or "undefined" value. Users may use
+ * {@link MapExpression#has} to check whether a value is present.
+ *
+ * <p>This type hierarchy differs from the {@linkplain org.bson} types in that
+ * they provide computational operations, the numeric types are less granular,
+ * and it offers multiple abstractions of certain types (document, map, entry).
+ * It differs from the corresponding Java types (such as {@code int},
+ * {@link String}, {@link java.util.Map}) in that the operations
+ * available differ, and in that an implementation of this API may be used to
+ * produce MQL in the form of BSON. (This API makes no guarantee regarding the
+ * BSON output produced by its implementation, which in any case may vary due
+ * to optimization or other factors.)
+ *
+ * <p>Some methods within the API constitute an assertion by the user that the
+ * data is of a certain type. For example, {@link DocumentExpression#getArray}}
+ * requires that the underlying field is both an array, and an array of some
+ * certain type. If the field is not an array in the underlying data, behaviour
+ * is undefined by this API (though behaviours may be defined by the execution
+ * context, users are strongly discouraged from relying on behaviour that is not
+ * part of this API).
+ *
+ *
+ * <p>Users should treat these interfaces as sealed, and should not create
+ * implementations.
  *
  * @see Expressions
  */
@@ -33,7 +89,16 @@ import java.util.function.Function;
 public interface Expression {
 
     /**
+     * The method {@link Expression#eq} should be used to compare values for
+     * equality. This method checks reference equality.
+     */
+    @Override
+    boolean equals(Object other);
+
+    /**
      * Whether {@code this} value is equal to the {@code other} value.
+     *
+     * <p>The result does not correlate with {@link Expression#equals(Object)}.
      *
      * @param other the other value.
      * @return the resulting value.
@@ -42,6 +107,8 @@ public interface Expression {
 
     /**
      * Whether {@code this} value is not equal to the {@code other} value.
+     *
+     * <p>The result does not correlate with {@link Expression#equals(Object)}.
      *
      * @param other the other value.
      * @return the resulting value.
@@ -83,7 +150,7 @@ public interface Expression {
     BooleanExpression lte(Expression other);
 
     /**
-     * {@code this} value as a {@link BooleanExpression boolean} if
+     * {@code this} value as a {@linkplain BooleanExpression boolean} if
      * {@code this} is a boolean, or the {@code other} boolean value if
      * {@code this} is null, or is missing, or is of any other non-boolean type.
      *
@@ -93,12 +160,9 @@ public interface Expression {
     BooleanExpression isBooleanOr(BooleanExpression other);
 
     /**
-     * {@code this} value as a {@link NumberExpression number} if
+     * {@code this} value as a {@linkplain NumberExpression number} if
      * {@code this} is a number, or the {@code other} number value if
      * {@code this} is null, or is missing, or is of any other non-number type.
-     *
-     * <p>Since integers are a subset of numbers, if a value is an integer,
-     * this does not imply that it is not a number, and vice-versa.
      *
      * @param other the other value.
      * @return the resulting value.
@@ -106,12 +170,9 @@ public interface Expression {
     NumberExpression isNumberOr(NumberExpression other);
 
     /**
-     * {@code this} value as an {@link IntegerExpression integer} if
+     * {@code this} value as an {@linkplain IntegerExpression integer} if
      * {@code this} is an integer, or the {@code other} integer value if
      * {@code this} is null, or is missing, or is of any other non-integer type.
-     *
-     * <p>Since integers are a subset of numbers, if a value is an integer,
-     * this does not imply that it is not a number, and vice-versa.
      *
      * @param other the other value.
      * @return the resulting value.
@@ -119,9 +180,9 @@ public interface Expression {
     IntegerExpression isIntegerOr(IntegerExpression other);
 
     /**
-     * {@code this} value as a {@link BooleanExpression boolean} if
-     * {@code this} is a boolean, or the {@code other} boolean value if
-     * {@code this} is null, or is missing, or is of any other non-boolean type.
+     * {@code this} value as a {@linkplain StringExpression string} if
+     * {@code this} is a string, or the {@code other} string value if
+     * {@code this} is null, or is missing, or is of any other non-string type.
      *
      * @param other the other value.
      * @return the resulting value.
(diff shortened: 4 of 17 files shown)


driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
+     *
+     * @param other the other value.
+     * @return the resulting value.
+     */
     StringExpression isStringOr(StringExpression other);
+
+    /**
+     * {@code this} value as a {@link StringExpression string} if
+     * {@code this} is a string, or the {@code other} string value if
+     * {@code this} is null, or is missing, or is of any other non-string type.
```
stIncMale, line 134: string -> date

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
+     *
+     * @param other the other value.
+     * @return the resulting value.
+     */
     IntegerExpression isIntegerOr(IntegerExpression other);
+
+    /**
+     * {@code this} value as a {@link BooleanExpression boolean} if
+     * {@code this} is a boolean, or the {@code other} boolean value if
+     * {@code this} is null, or is missing, or is of any other non-boolean type.
```
stIncMale, line 124: boolean -> string

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
+     * {@code this} is an array, or the {@code other} array value if
+     * {@code this} is null, or is missing, or is of any other non-array type.
+     *
+     * <p>Warning: this operation does not guarantee type safety. While this
+     * operation is guaranteed to produce an array, the type of the elements of
+     * that array are not guaranteed by the API. The specification of a type by
+     * the user is an unchecked assertion that all elements are of that type,
+     * and that no element is null, is missing, or is of some other type. If the
+     * user cannot make such an assertion, some appropriate super-type should be
+     * chosen, and elements should be individually type-checked.
```
stIncMale, line 153: _`some appropriate super-type should be chosen, and elements should be individually type-checked`_ - shouldn't `and` be `and/or`? For example, if the chosen super-type is `Expression`, then no checks are needed.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
+     */
     <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<T> other);
+
+    // TODO-END doc after Map merged, "record" and "schema objects" are decided
     <T extends DocumentExpression> T isDocumentOr(T other);
 
 
     /**
-     * server error if type cannot be converted to string (arrays, objects)
+     * The {@link StringExpression string} value corresponding to this value.
```
stIncMale, line 166: corresponding -> representing?

That's what `Object.toString` uses, and it indeed looks like the right word. An arbitrary mapping from `Expression` to `StringExpression` establishes correspondence, but that does not necessary mean that the strings corresponding to values in this mapping represent the values in a way that `asString` is supposed to: even if there is no `parse` that will take the result of `asString` and produce the represented value, a human may look at the result of `asString` and understand what value it represents.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
+     * @see StringExpression#parseInteger()
+     * TODO-END all the others? implement?
+     * @return the resulting value.
      */
     StringExpression asString();
 
     /**
-     * Applies the given function to this argument. Note that "apply" usually
-     * applies functions to arguments; here, the parameters are reversed.
+     * Applies the provided function to {@code this}.
```
stIncMale, line 184: If we are making it vague whether `isBooleanOr` (and similar methods) return `this`/`other` references or not, then we probably should use that approach here and say `Applies the provided function to {@code this} value.`

Note that the documentation of `Expression.switchMap` says _`applying the provided switch mapping to {@code this} value`_, not `... to {@code this}`, which is consistent with that I am proposing here.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
      *
-     * @param f
-     * @return
-     * @param <T>
-     * @param <R>
+     * <p>Equivalent to {@code f.apply(this)}, and allows lambdas and static,
+     * user-defined functions to use the chaining syntax. For example:
+     * {@code myInteger.apply(isEven)} (here, {@code isEven} is a function
+     * taking an {@link IntegerExpression integer} and yielding a
+     * {@link BooleanExpression boolean}).
```
stIncMale, line 190: It seems simpler to say `{@code isEven} is {@code Function<IntegerExpression, BooleanExpression>}` then to expand this with words. But if we are saying "for example", let's provide an actual complete code example where `isEven` is implemented and `myInteger.apply` is assigned to a variable of type `IntegerExpression`.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
+     *
+     * @param f the function to apply.
+     * @return the resulting value.
+     * @param <T> the type of {$code this} value.
+     * @param <R> the type of the resulting value.
      */
     <T extends Expression, R extends Expression> R apply(Function<T, R> f);
```
stIncMale, line 197: The type of `Function<T, R>` is too strict. I keep pointing out problems like this, but they keep coming. This is not productive, and I would hope that the problem stops constantly reappearing in the new code.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
+     * <p>Equivalent to {@code f.apply(this)}, and allows lambdas and static,
+     * user-defined functions to use the chaining syntax. For example:
+     * {@code myInteger.apply(isEven)} (here, {@code isEven} is a function
+     * taking an {@link IntegerExpression integer} and yielding a
+     * {@link BooleanExpression boolean}).
+     *
+     * @param f the function to apply.
+     * @return the resulting value.
+     * @param <T> the type of {$code this} value.
+     * @param <R> the type of the resulting value.
```
stIncMale, line 195: Here and in `switchMap` we can say `the type of {@code this}` / `the type of the result`, "value" appears to serve no purpose. Am I missing its purpose here?

stIncMale (changes requested):
So far I reviewed only `Expression`.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
      */
-    BooleanExpression lte(Expression lte);
+    BooleanExpression lte(Expression other);
 
     /**
-     * also checks for nulls
-     * @param other
-     * @return
+     * {@code this} value as a {@link BooleanExpression boolean} if
+     * {@code this} is a boolean, or the {@code other} boolean value if
```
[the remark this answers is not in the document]
katcharov: Right

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
+     *
+     * @param other the other value.
+     * @return the resulting value.
      */
     BooleanExpression isBooleanOr(BooleanExpression other);
+
+    /**
+     * {@code this} value as a {@link NumberExpression number} if
+     * {@code this} is a number, or the {@code other} number value if
+     * {@code this} is null, or is missing, or is of any other non-number type.
```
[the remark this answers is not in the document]
katcharov: I want to emphasize this, because this specific method (isNumberOr) is (in great part) intended to deal with those unusual values.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
+     */
     NumberExpression isNumberOr(NumberExpression other);
+
+    /**
+     * {@code this} value as an {@link IntegerExpression integer} if
+     * {@code this} is an integer, or the {@code other} integer value if
+     * {@code this} is null, or is missing, or is of any other non-integer type.
+     *
+     * <p>Since integers are a subset of numbers, if a value is an integer,
+     * this does not imply that it is not a number, and vice-versa.
```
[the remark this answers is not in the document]
katcharov: This seems right, I will remove.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: Let's go with "and if necessary"

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: Let's go with `The {@link StringExpression string} representation of {@code this} value`

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: I agree.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: I agree

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: Agreed to leave, since this would make this semantically inconsistent with the rest of docs.

jyemin (commented):
I find

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,202 @@
...
 @Evolving
 public interface Expression {
 
     /**
-     * Returns logical true if the value of this expression is equal to the
-     * value of the other expression. Otherwise, false.
+     * The method {@link Expression#eq} should be used to compare values for
+     * equality.
+     */
+    @Deprecated // TODO-END (?) <p>Marked deprecated to prevent erroneous use.
```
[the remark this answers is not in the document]
katcharov: I've seen abuses, though not the deprecation of equals in particular. I don't plan to argue for it or keep it, but I was worried that users will mistake it for `eq`, and was hoping for some creative way to address this.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,202 @@
 import java.util.function.Function;
 
 /**
```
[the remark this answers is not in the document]
katcharov: I think:

> A value that may be stored in or otherwise operated upon by MongoDB.

Followed by some less-dense variation on:

> This API is a language binding: it is a way to use a remote system as if
> it were a native Java API. It is a way to represent values and computations
> in some "execution context". This API is also a Domain Specific Language: it
> is the Java-native variant of MQL, which is used to query MongoDB or (more
> broadly) to work with data on MongoDB. 

It would be followed by a few paragraphs covering important points and gotchas.

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
```diff
@@ -17,45 +17,43 @@
 package com.mongodb.client.model.expressions;
 
 /**
- * Expresses a boolean value.
+ * A logical boolean value, either true or false.
```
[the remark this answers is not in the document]
stIncMale: Slack discussion: https://mongodb.slack.com/archives/GCKKT44RM/p1671122088820819.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,116 +21,200 @@
...
-     * value of the other expression. Otherwise, false.
+     * Whether {@code this} value is not equal to the {@code other} value.
      *
-     * @param ne the other expression
-     * @return true if equal, false otherwise
+     * @param other the other value.
+     * @return the resulting value.
      */
-    BooleanExpression ne(Expression ne);
+    BooleanExpression ne(Expression other);
```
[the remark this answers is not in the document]
stIncMale: It's now on `eq`, but it is equally applicable to `ne`.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
stIncMale: Resolving because I left the same [comment](https://github.com/mongodb/mongo-java-driver/pull/1055/files#r1064907718) in the PR that introduced the method.

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
```diff
@@ -19,46 +19,128 @@
...
+     */
+    NumberExpression max(NumberExpression other);
+
+    /**
+     * The smaller value of {@code this} and the {@code other} value.
+     *
+     * @param other the other value.
+     * @return the resulting value.
+     */
+    NumberExpression min(NumberExpression other);
```
stIncMale, line 118: Should "larger"/"smaller" be hyperlinks to "gt"/"lt"? The same is applicable to `IntegerExpression`.

```java
{@linkplain #gt(Expression) larger}
{@linkplain #lt(Expression) smaller}
```

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
```diff
@@ -19,46 +19,128 @@
 import java.util.function.Function;
 
 /**
- * Expresses a numeric value.
+ * A number value. {@link IntegerExpression Integers} are a subset of numbers,
+ * and so, for example, the integer 0 and the number 0 are the same value,
+ * and are equal.
  */
 public interface NumberExpression extends Expression {
```
stIncMale, line 26: `IntegerExpression` explains the "integers are a subset of numbers" and appears to be the right place for such an explanation. Is it really useful to have that duplicated here?

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
```diff
@@ -19,46 +19,128 @@
...
+
+    /**
+     * The integer result of rounding {@code this} to the nearest even value.
+     *
+     * @return the resulting value.
+     */
     IntegerExpression round();
 
+    /**
+     * The result of rounding {@code this} to the nearest even {@code place}.
```
stIncMale, line 128: This reads like we are rounding to "the nearest even place", while actually we are rounding to the specified "decimal place" using the "round half to even" approach.

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
```diff
@@ -19,22 +19,117 @@
 import java.util.function.Function;
 
 /**
- * Expresses a date value.
+ * An instantaneous date and time. Tracks a UTC datetime, the number of
+ * milliseconds since the Unix epoch. Does not track the timezone.
  */
 public interface DateExpression extends Expression {
```
stIncMale, line 25: I don't understand this sentence and it is not clear whether it adds any information to what follows after. What does it add?

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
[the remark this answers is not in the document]
katcharov: (I removed this)

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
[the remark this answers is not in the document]
katcharov: I added a long description (similar to [Stream](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html)). Unsure which parts of it, if any, might be moved to the package docs (none are obviously unrelated to the type hierarchy of which this is the root).

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
```diff
@@ -19,49 +19,47 @@
 import java.util.function.Function;
 
 /**
- * Expresses a boolean value.
+ * A logical boolean value, either true or false.
```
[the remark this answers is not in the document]
katcharov: The redundancy is there because I seem to be taking the opportunity to remind users that "null" is not a valid value. But I have removed this.

I thought that capitalizing would indicate that it is a Java Boolean, while wrapping in `code` would indicate it is the primitive, but leaving it uncapitalized would be in line with speaking abstractly, which is what I think we are trying to do here? And if we are, I don't think it is usually capitalized, or perhaps that is that not the case?

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
```diff
@@ -19,49 +19,47 @@
...
-     * returns the result of the evaluated right branch expression.
+     * The {@code left} branch when {@code this} is true,
+     * and the {@code right} branch otherwise.
      *
-     * @param left the left branch expression
-     * @param right the right branch expression
-     * @return left if true, right if false.
+     * @param left the left branch.
+     * @param right the right branch.
+     * @return the resulting value.
```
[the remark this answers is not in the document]
katcharov: Right, this focused too much on evaluation semantics.

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
```diff
@@ -19,32 +19,98 @@
...
-        return this.subtract(Expressions.of(subtract));
+    /**
+     * The result of subtracting the {@code other} value from {@code this}.
+     *
+     * @param other the other value.
+     * @return the resulting value.
+     */
+    default IntegerExpression subtract(final int other) {
+        return this.subtract(Expressions.of(other));
     }
```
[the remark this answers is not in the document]
katcharov: For subtract, I considered and avoided "difference" because to me that word might (especially in the absence of the word "subtract") imply "how much of a difference is there between 1 and 2?" -> "1" (so the absolute difference). I think I avoided "quotient" because "the quotient of dividing" did not sound grammatical to me. What do you think?

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
katcharov: Which specific sentence of the 3 are you referring to? (There is some redundancy there, intended to stress and make certain implications clear that might not be immediately obvious to everyone.)

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
katcharov: I thought it was not harmful to duplicate information (as a general convention in our docs), and helpful to note it here for those who might not read the Integer docs, since it is a notable deviation from how things work in Java.

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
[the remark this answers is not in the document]
stIncMale: Abstract Boolean type and logic are capitalized similarly to Newtonian mechanics.

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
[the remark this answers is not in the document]
stIncMale: It still uses "branch" in the description of the method.

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
```diff
@@ -51,8 +52,8 @@ public interface BooleanExpression extends Expression {
      * The {@code left} branch when {@code this} is true,
      * and the {@code right} branch otherwise.
      *
-     * @param left the left branch.
-     * @param right the right branch.
+     * @param left the left value.
+     * @param right the right value.
      * @return the resulting value.
      * @param <T> The type of the resulting expression.
      */
```
stIncMale, line 59: What if we rename `left` value to `ifTrue` value and `right` to `ifFalse`/`otherwise` value? I'd say that makes using the method more straightforward.

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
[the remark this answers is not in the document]
stIncMale: > avoided "difference" because to me that word might (especially in the absence of the word "subtract") imply "how much of a difference is there between 1 and 2?" -> "1"

It would be "difference of subtracting" similarly to "product of multiplying".

> "the quotient of dividing" did not sound grammatical to me

But it is grammatical, it is no different from "sum of adding". The reason it feels weird is because people usually don't use this word in conversations.

> What do you think?

I think that we should be consistent: either we call results of these operations by their names, or we refer to  all of them using a generic term.

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
stIncMale: Terrible, sorry) I was talking about _"An instantaneous date and time"_. It's true that people say something like "instantaneous coordinate" to emphasize that they are talking about the coordinate value at a specific time (though to me even that seems tautologous). However, when we talk about time, I can't understand understand what "instantaneous date/time" is - "the time value at a specific time" does not make sense.

Maybe we could just remove the first sentence?

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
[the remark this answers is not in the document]
katcharov: I should have caught that.

By the way, do you see another way to implement cond? It has been mentioned a few times in the design document, that users would find it strange to chain `cond` off of a boolean in this way. The alternative I can imagine is to have a "nested functional" cond static method that takes 3 params, but this seems inconsistent with the design approach. It is also possible to have `bigExpressionHere.cond(e -> e.eq(1), e -> leftExpression, e -> rightExpression)` (which is a bit more similar to our `switch`), but it seemed very verbose (and I vaguely remember there was a more technical problem with it, which I cannot now recall).

driver-core/src/main/com/mongodb/client/model/expressions/IntegerExpression.java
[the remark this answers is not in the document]
katcharov: Alright, fair enough. But I think there are places elsewhere where there is not a more-specific term than "result".

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -228,6 +286,18 @@ public static DocumentExpression of(final Bson document) {
                 document.toBsonDocument(BsonDocument.class, cr))));
     }
 
+    /**
+     * The null value in the context of the MongoDB Query Language (MQL).
+     *
+     * <p>The null value is not part of, and cannot be used as if it were part of,
+     * any other type. It has no explicit type of its own. Instead of checking for
```
stIncMale, line 308: "cannot be used as if it were part of any other type" -> "cannot be used as if it were part of any other **explicit** type"?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -228,6 +286,18 @@ public static DocumentExpression of(final Bson document) {
...
+     * null, users should check for their expected type, via methods such as
+     * {@link Expression#isNumberOr(NumberExpression)}. Where the null value must
+     * be checked explicitly, users may use {@link Branches#isNull} within
+     * {@link Expression#switchOn}.
```
stIncMale, line 312: _"Instead of checking for null, users should check for their expected type"_ - as a user I don't understand why I should check for other types, which null is not of, when I need to check for null (as someone who knows how `isNumberOr` works, I understand what you wanted to convey, but I don't think it was successfully conveyed)

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -229,8 +242,10 @@ public static <T extends Expression> ArrayExpression<T> ofArray(final T... array
...
 
     /**
-     * Returns an {@linkplain EntryExpression entry} value. An entry is a key
-     * and a value, and is used with {@linkplain MapExpression maps}. An entry
+     * Returns an {@linkplain EntryExpression entry} value. An entry is a
+     * {@linkplain StringExpression string} key and some
+     * {@linkplain Expression value}. Entries are used with
+     * {@linkplain MapExpression maps}. An entry
      * value may be equal to a document value with a field "k" equal to the
      * entry's key and a field "v" equal to the entry's value.
```
stIncMale, line 250: What does "may be" mean here, is it not always equal?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -272,9 +287,9 @@ public static <T extends Expression> MapExpression<T> ofMap(final Bson map) {
 
     /**
      * Returns a {@linkplain DocumentExpression document} value corresponding to the
-     * provided BSON document.
+     * provided {@link Bson Bson document}.
      *
-     * @param document the document as BSON.
+     * @param document the {@link Bson Bson document}.
      * @return the resulting value.
      */
```
stIncMale, line 294: `{@link Bson Bson document}` -> `{@linkplain Bson BSON document}`?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -207,11 +207,24 @@ public static <R extends DocumentExpression> R current() {
...
+     * in the aggregation pipeline stage.
+     *
+     * @return the resulting value
+     * @param <R> the type of the resulting value
+     */
+    public static <R extends DocumentExpression> MapExpression<R> currentAsMap() {
+        return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonString("$$CURRENT")))
+                .assertImplementsAllExpressions();
+    }
+
```
stIncMale, line 222: I think this method should be generic over `<T extends Expression>` and return `MapExpression<T>`, where `T` is the type of map values (this last part affects the documentation).

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: It is vague because pair arrays might, in the future, need to be entries, and we might figure out a way to make it work.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,12 +21,68 @@
...
+ * requires that the underlying field is both an array, and an array of some
+ * certain type. If the field is not an array in the underlying data, behaviour
+ * is undefined by this API (though behaviours may be defined by the execution
+ * context, users are strongly discouraged from relying on behaviour that is not
+ * part of this API).
+ *
+ * <p>As with the Java Stream API's terminal operations, corresponding Java
+ * values are not directly available, but must be obtained indirectly via the
+ * aggregation pipeline, or via find. Certain methods may cause an error, which
+ * will be produced through these "terminal operations".
```
stIncMale, line 82: - This section should probably be moved up directly before _"The following is an example of usage within an aggregation pipeline ..."_
- _"via the aggregation pipeline, or via find"_ - we should link to the relevant Java APIs.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: I'll think about this, but I am unsure what else to say.

The idea, just to make it more clear, is that there is no null check, because there is no null type to convert to and then use "as a null". So the only time you should need to null-check is in the context of a number of switch branches. This is not usually the case in Java, because an `Integer` variable can equal null.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,12 +21,68 @@
 import java.util.function.Function;
 
 /**
```
stIncMale, line 23: [optional]

I propose us to separate the documentation of the whole API (the place for it is the package docs) and the documentation of the `Expression` class. But I am neither insisting on it nor volunteering to do the work of separating it.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -172,11 +194,27 @@ public static ArrayExpression<StringExpression> ofStringArray(final String... ar
...
 
+    /**
+     * Returns the current or root {@linkplain DocumentExpression document} value,
+     * the top-level document, currently being processed in the aggregation
+     * pipeline stage.
+     *
+     * @return the resulting value
+     * @param <R> the type of the resulting value
+     */
     public static <R extends DocumentExpression> R current() {
```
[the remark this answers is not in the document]
katcharov: Oh I see. Yes. If it were a lambda we would not have this issue. I wonder if users would be confused by the idea of a "virtual" value?

What if we remove "returns...", perhaps replacing it with "refers to the current"? Alternatively, some wording about "variable", which would be a novel concept that might make things compatible, but also confusing? And maybe some wording, "in the context of an aggregation pipeline"? How does this work with "find"...

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
katcharov: I am not sure what the wording should be here, and in the overload. Java uses the wording:

> Rounding mode to round towards the "nearest neighbor" unless both neighbors are equidistant, in which case, round towards the even neighbor. 

So perhaps

```
The result of rounding {@code this} using the "half even" rounding mode, 
which rounds towards the "nearest neighbor" unless both neighbors are 
equidistant, in which case, it rounds towards the even neighbor.
```

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -21,35 +21,39 @@
...
+     * An array consisting of only those elements in {@code this} array that
+     * match the provided predicate.
      *
-     * @param cond the function to apply to each element
-     * @return the new array
+     * @param predicate the predicate to apply to each element to determine if
+     *                  it should be included.
+     * @return the resulting array.
      */
     ArrayExpression<T> filter(Function<? super T, ? extends BooleanExpression> cond);
```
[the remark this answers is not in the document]
katcharov: Right, I suspect this was lost in a rebase.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -190,6 +228,17 @@ public static <T extends Expression> ArrayExpression<T> ofArray(final T... array
...
+     * and a value, and is used with {@linkplain MapExpression maps}. An entry
+     * value may be equal to a document value with a field "k" equal to the
+     * entry's key and a field "v" equal to the entry's value.
+     *
+     * @param k the key.
+     * @param v the value.
+     * @return the resulting value.
+     * @param <T> the type of the key.
+     */
     public static <T extends Expression> EntryExpression<T> ofEntry(final String k, final T v) {
```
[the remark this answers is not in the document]
katcharov: Right - I'll move it there, though otherwise I'll stick to boolean, date, number, integer, and expression (and expressions).

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -172,11 +194,27 @@ public static ArrayExpression<StringExpression> ofStringArray(final String... ar
         return new MqlExpression<>((cr) -> new AstPlaceholder(new BsonArray(result)));
     }
 
+    /**
+     * Returns the current or root {@linkplain DocumentExpression document} value,
+     * the top-level document, currently being processed in the aggregation
+     * pipeline stage.
```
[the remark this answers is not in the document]
katcharov: Added with some variation.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: What non-explicit other (not-null) type can it be used as part of?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: Right, this is an error.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,12 +21,68 @@
...
+ * computations, to store and retrieve data, or to otherwise work with data
+ * on a MongoDB server or compatible execution context.
+ *
+ * <p>The following is an example of usage within an aggregation pipeline. Here,
+ * the current document value is obtained and its "numberArray" field is
+ * filtered and summed, in a style similar to that of the Java Stream API:
+ *
+ * <pre>{@code
+ * AggregateIterable<Document> result = col.aggregate(Arrays.asList(
+ *     addFields(new Field<>("result", current()
```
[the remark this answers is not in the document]
katcharov: I have added `import static com.mongodb.client.model.expressions.Expressions.current;` - I worry that people will think it is a convention to write `MqlValue.current()`, rather than just something we do to make the imports clear.

For your second point, I think because this method is on the "border" of Java and the existing agg, rather than strictly part of the API, that it should be "contained" into this class, which deals with this "border".

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,12 +21,68 @@
...
+ *
+ * <p>Values are typically obtained via the current document and its fields, or
+ * specified via statically-imported methods on the {@link Expressions} class.
+ *
+ * <p>The null value is not part of, and cannot be used as if it were part of,
+ * any other type. It has no explicit type of its own. Instead of checking for
+ * null, users should check for their expected type, via methods such as
+ * {@link Expression#isNumberOr(NumberExpression)}. Where the null value must
+ * be checked explicitly, users may use {@link Branches#isNull} within
+ * {@link Expression#switchOn}.
```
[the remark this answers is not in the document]
katcharov: I am not sure it discourages, but in any case, I kept the first sentence, and then the "see" you suggested.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,12 +21,68 @@
...
+ *
+ * <p>There is no explicit "missing" or "undefined" value. Users may use
+ * {@link DocumentExpression#has}.
```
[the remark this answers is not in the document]
katcharov: Indeed we do not. But now I am not sure, it seems that it should not exist on document, but should exist on map. I cannot immediately think of a reason why it needs to be on document.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: I cannot clearly identify docs presently on the Expression class that are not fairly related to it, and so do not belong there, but instead belong to the package. At the same time, there should probably be some package docs (or at least a link to the Expression class and the Expressions class), but I do not have some specific text to put there.

jyemin: The thing about package docs is that they are harder (impossible?) to see directly in the IDE, which is where I suspect most people actually look at the javadoc.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: How could I link those APIs from here?

katcharov: I moved it after the section discussing "initial" operations/expressions.

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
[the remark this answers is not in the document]
stIncMale: > cond static method that takes 3 params, but this seems inconsistent with the design approach

Yes, everything else is chainable, and a static method is not.

> bigExpressionHere.cond(e -> e.eq(1), e -> leftExpression, e -> rightExpression)

This seems fine at first glance, but is indeed similar to
```java
of(0).switchOn(on -> on.eq(of(0), v -> of("A")).defaults(v -> of("B")))
```
which makes this form almost useless. I think it's fine for us to have `cond` as it is now, and if users don't like it, maybe they will let us know.

P.S. For some reason the code I wrote doesn't compile, but I don't immediately see what the problem is. I will leave a comment in https://github.com/mongodb/mongo-java-driver/pull/1055, because this seems worth looking at.

driver-core/src/main/com/mongodb/client/model/expressions/BooleanExpression.java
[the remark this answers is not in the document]
stIncMale: Resolving to avoid wasting time discussing things that are not obvious bugs.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
stIncMale: [marked as optional]

It's own type that has a single null value. That's what Java does for `null`, and that's the type one may infer (an implicit type) for null values for MQL similarly how this API inferred numbers and integers.

This API does not allow using the null value as if it were of any ("part of") type explicitly defined by the API, or were convertible to any explicit type. But like any other value, the null value has a type (if not explicit, then implicit).

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
stIncMale: I see. Then the question is: how may this information be useful to a user, i.e., what code will a user be able to write given this information that he would not be able to write if not having it, without relying on unspecified API behavior?

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -208,14 +209,15 @@ public static <R extends DocumentExpression> R current() {
...
      * The "current" value is the top-level document currently being processed
      * in the aggregation pipeline stage.
      *
-     * @return the resulting value
-     * @param <R> the type of the resulting value
+     * @return a reference to the current value as a map.
+     * @param <R> the type of the map's values.
      */
-    public static <R extends DocumentExpression> MapExpression<R> currentAsMap() {
+    public static <R extends Expression> MapExpression<R> currentAsMap() {
```
stIncMale, line 220: We should have a test for this method, like we do for `current`.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
stIncMale: [marked as optional]

> How could I link those APIs from here?

Then code dependencies don't allow me to use `@link`, I use `{@code MongoCollection.aggregate}`/`{@code MongoCollection.find}`.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
stIncMale: In IntelliJ one may view package docs in exactly the same way one views other docs:

- either hover the mouse for some time over the last name section of the package name you are interested in anywhere in the code where a package name is used;
- or place the caret on the same thing and press you hotkey combination for viewing docs.

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
stIncMale: [marked optional]

_"The result of rounding `{@code this}` to `{@code place}` decimal places using the "half to even" approach."_ addresses the issue of _"nearest even"_ appearing to be applied to _"`{@code place}`"_ in _"nearest even `{@code place}`"_, which I talked about in the first comment.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: added in #1057 
https://github.com/mongodb/mongo-java-driver/pull/1057/files#diff-72f687508b977da46d04d15699a94df0a1ec21300101ddca46b5a514e12177d9R110

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: `some_doc_in_their_database.eq( some_entry )`. If we make the above (hypothetical) changes, this will start to fail, unless we change the contract specified by the docs.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: I tried to clarify a bit.

driver-core/src/main/com/mongodb/client/model/expressions/NumberExpression.java
katcharov: Used this for `round(IntegerExpression place)`, but not for `round()` (if that should be updated too, let me know).

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
stIncMale: [optional]

So I understand that the goal is to tell users that they must not rely on the representation of `EntryExpression` that they may infer via, for example, comparing it with a `Document` via `eq`. If that's correct, wouldn't it be more clear to say so directly instead of telling users the current representation (_"a document value with a field "k" equal to the entry's key and a field "v" equal to the entry's value"_) and implying that they must not rely on it by using the wording _"an entry value **may** be equal to"_? Formally, this wording achieves the goal, which is why I am marking this discussion as optional.

stIncMale (changes requested):
<a useless comment forced by GitHub>

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
katcharov: I'll remove this sentence.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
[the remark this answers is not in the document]
katcharov: Added in #1054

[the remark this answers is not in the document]
stIncMale: We now have `MapExpression.has` and need to remove `{@link DocumentExpression#has}` from the docs.

katcharov:
Rebased. Conflicts on switch/pass (none on MqlEx, which should confirm that the correct signature was used).

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
```diff
@@ -19,22 +19,135 @@
...
+ * of the MongoDB Query Language (MQL). Tracks the number of
+ * milliseconds since the Unix epoch, and does not track the timezone.
  */
 public interface DateExpression extends Expression {
+
+    /**
+     * The year of {@code this} date as determined by the provided
+     * {@code timezone}.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
```
stIncMale, line 32: Let's either give examples of both the offset and the identifier values, or link to the relevant part of the MongoDB documentation. I am fine with either duplicating this information everywhere, or specifying it in the doc comment for `DateExpression`.

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
```diff
@@ -19,22 +19,135 @@
...
+     * @return the resulting value.
+     */
     IntegerExpression millisecond(StringExpression timezone);
 
+    /**
+     * The string representation of {@code this} date as determined by the
+     * provided {@code timezone}, and formatted according to the {@code format}.
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @param format the format specifier.
```
stIncMale, line 127: Let's either specify the format and give examples, or link to the relevant part of the MongoDB documentation.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,104 +21,310 @@
...
+ * {@code MongoCollection.aggregate} or {@code MongoCollection.find}.
+ * Certain methods may cause an error, which will be produced
+ * through these "terminal operations".
+ *
+ * <p>The null value is not part of, and cannot be used as if it were part
+ * of, any explicit type (except the root type {@link Expression} itself).
+ * See {@link Expressions#ofNull} for more details.
+ *
+ * <p>There is no explicit "missing" or "undefined" value. Users may use
+ * {@link MapExpression#has}.
```
stIncMale, line 62: We discussed that the intent here was to express that the only special value that is not of any explicit type other than `Expression` is the null value, which can be expressed via `Expressions.ofNull`.

The current wording _"no explicit "missing" or "undefined" value"_ implies (at least that's my interpretation) that there are such implicit values (can't be expressed, unlike null, but are expected to be encountered). Maybe instead of the two highlighted sections, we could just say something like

> `{@linkplain Expressions#ofNull Null}` is a special value that is of the `{@link Expression}` type and is not of any other explicit type. The API does not specify other special values.

If not, let's still somehow make it clear that neither "missing", nor "undefined", nor "any other special value" is specified.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
katcharov: Do you know what this format is called? https://www.mongodb.com/docs/manual/reference/operator/aggregation/dateToString/

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,104 +21,310 @@
...
+ * part of this API).
+ *
+ *
+ * <p>Users should treat these interfaces as sealed, and should not create
+ * implementations.
  *
  * @see Expressions
  */
 @Evolving
 public interface Expression {
```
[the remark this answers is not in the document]
katcharov: I will apply this in the final doc PR which covers docs etc. (along with, for example, server version annotations).

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -211,17 +210,48 @@ public interface Expression {
...
      * @param other the other value.
      * @return the resulting value.
      * @param <T> the type of the elements of the resulting array.
      */
     <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<? extends T> other);
 
-    // TODO-END doc after Map merged, "record" and "schema objects" are decided
+    /**
+     * {@code this} value as a {@linkplain DocumentExpression document} if
+     * {@code this} is a document or document-like value (such as a
```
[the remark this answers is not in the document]
katcharov: I have updated the wording.

I think it is better to avoid "users must rely" phrasing since the implication/obligation seems much broader than simply specifying method behaviour.

When it comes to explaining document/map/entry, it seems better to plainly state that we offer "multiple abstractions of certain types (document, map, entry)".

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: I think the "cannot be used as if it were" wording/emphasis is important for Java developers. I think calling it "special" might be confusing. I think we still need to call out missing/undefined specifically, since they are mentioned in the server docs. I have updated the wording.

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
stIncMale: I don't. Note that since the MongoDB documentation doesn't refer to any existing format, one can't assume that it uses one: even if you find a commonly-known format that at the moment is identical to the one described in the docs, that does not mean MongoDB can't diverge from it in the future, thus making one's assumption incorrect.

Putting the above consideration aside, it looks like you would have been fine with referring to a format if it were defined somewhere else. What is the reason for trying to avoid referring to the format specification in the MongoDB docs?

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -21,104 +21,310 @@
...
+     * the user is an unchecked assertion that all values are of that type,
+     * and that no value is null of some other type. If the
+     * user cannot make such an assertion, some appropriate super-type should be
+     * chosen, and if necessary the values should be individually type-checked.
+     *
+     * @param other the other value.
+     * @return the resulting value.
+     * @param <T> the type of the values of the resulting map.
+     */
     <T extends Expression> MapExpression<T> isMapOr(MapExpression<? extends T> other);
```
stIncMale, line 254: > all values are of that type, and that no value is null of some other type

Just saying _"all values are of that type"_ appears to be enough. Explicitly excluding the null value is incorrect because `MapExpression<Expression>` may have null values: null is of the `Expression` type.

katcharov: Right.

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
katcharov: Added a link

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
katcharov: If there was such a format, we could confirm it, and then refer to it.

I added what you suggested to the top of dateEx.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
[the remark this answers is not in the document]
stIncMale: [optional]

> all {@linkplain EntryExpression entries} created using the present API

If we introduce ways to create non-document-like entries in the future, the phrase "present API" will automatically start wrongly encompassing them. If we instead list all the ways that produce document-like entires, then we save ourselves from such a problem.

Zooming out, I don't see a reason you want users to consider entries document-like, given that the structure of such a document is not specified.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
stIncMale: In `Expression.isMapOr`:
> the values of that **array**

Should be map.

The change regarding null values is needed also in `Expression.isArrayOr`.

driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -211,18 +266,27 @@ public static <T extends Expression> MapExpression<T> ofMap() {
     }
 
     /**
-     * user asserts type of values is T
+     * Returns a {@linkplain MapExpression map} value corresponding to the
+     * provided {@link Bson Bson document}. The user asserts that all values
```
stIncMale, line 270: The prose style and text formatting is different in `Expression.isArrayOr`/`isMapOr`. Shouldn't we use the same everywhere? I am not saying what it should be.

driver-core/src/main/com/mongodb/client/model/expressions/DateExpression.java
stIncMale: The added link is about the format specifier, not the offset.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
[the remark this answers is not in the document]
stIncMale: Right, the annotation may be added later. This comment is about documenting all unchecked methods with text (unless you decide not to do so and use only the annotation instead, which is what I would have preferred). The reason I used the annotation in the table is to conveniently express what a user must assert.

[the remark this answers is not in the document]
katcharov: Please leave unresolved, for easier reference.

stIncMale (approved):

driver-core/src/main/com/mongodb/client/model/expressions/MqlUnchecked.java
```diff
@@ -0,0 +1,75 @@
...
+ * Documents places where the API relies on a user asserting
+ * something that is not checked at run-time.
+ * If the assertion turns out to be false, the API behavior is unspecified.
+ *
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+@Documented
+@Retention(RetentionPolicy.SOURCE)
+@Target({ElementType.METHOD, ElementType.TYPE_USE})
+public @interface MqlUnchecked {
```
stIncMale, line 34: Note how this annotation is documented as not being a part of the public API despite residing in a non-`internal` package and being `public`. This is because we want to generate the API docs for this annotation, but we don't want users to use it either to annotate their code, or to analyze the code at compile time.

PR #1060, opened 2022-12-02
Improve logging in server monitor

jyemin:
Log as error any exception except MongoInterruptedException in server monitor before exiting monitor thread

JAVA-4822

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
@@ -16,6 +16,7 @@
 
 package com.mongodb.internal.connection;
 
+import com.mongodb.MongoInterruptedException;
 import com.mongodb.MongoNamespace;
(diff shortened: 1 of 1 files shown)


driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -178,6 +179,10 @@ public void run() {
                     }
                     waitForNext();
                 }
+            } catch (MongoInterruptedException e) {
+                // ignore
+            } catch (RuntimeException e) {
+                LOGGER.error(format("Server monitor for %s exiting with exception", serverId), e);
```
jyemin, line 185: In addition to catching and ignoring `MongoInterruptedException`, it seems better to use our own logger to log any other exceptions rather than rely on the default behavior of threads when their Runnable exits exceptionally.

rozza (approved):
LGTM

PR #1062, opened 2022-12-08
Add lookup overloads for $documents

katcharov:
[JAVA-4674](https://jira.mongodb.org/browse/JAVA-4674)

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/Aggregates.java
+++ b/driver-core/src/main/com/mongodb/client/model/Aggregates.java
@@ -313,7 +313,6 @@ public final class Aggregates {
      * @mongodb.driver.manual reference/operator/aggregation/lookup/ $lookup
      * @mongodb.server.release 3.6
      * @since 3.7
-     *
      */
     public static Bson lookup(@Nullable final String from, final List<? extends Bson> pipeline, final String as) {
         return lookup(from, null, pipeline, as);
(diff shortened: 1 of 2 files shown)


driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -334,14 +333,50 @@ public static Bson lookup(@Nullable final String from, final List<? extends Bson
      * @param as            the name of the new array field to add to the input documents.
      * @return the $lookup pipeline stage
      * @mongodb.driver.manual reference/operator/aggregation/lookup/ $lookup
-     * @mongodb.server.release 3.6
```
rozza, line 337: This API signature hasn't changed? So the server release and since should stay the same

rozza (commented):
I'm not sure this as clear as I'd imagined and think it may be too brittle / easy for users to mess up.

So I'd be happy if we didn't add these overrides.

jyemin (commented):
As is, it doesn't seem to add much value since we are not using the type system to enforce that the first stage is $documents.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
katcharov: Good eye, I think I meant to paste this into the other lookup.

katcharov:
Closing this and the JIRA, per above comments.

PR #1067, opened 2023-01-12
Add structured logging support for command logging

jyemin:
* Align command logging messages to specification
* Add command logging unified tests

One leftover that I opened a separate ticket for:

* Support for max document length: this involves plumbing the value all the way down from MongoClientSettings

JAVA-4752
JAVA-4753

### DIFF
--- a/config/checkstyle/suppressions.xml
+++ b/config/checkstyle/suppressions.xml
@@ -86,6 +86,7 @@
     <suppress checks="ParameterNumber" files="DefaultClusterableServerFactory"/>
     <suppress checks="ParameterNumber" files="Operations"/>
     <suppress checks="ParameterNumber" files="ChangeStreamDocument"/>
+    <suppress checks="ParameterNumber" files="StructuredLogger"/>
 
     <!--Legacy code that has not yet been cleaned-->
     <suppress checks="FinalClass" files="AggregationOptions"/>
--- a/config/codenarc/codenarc.xml
+++ b/config/codenarc/codenarc.xml
@@ -143,6 +143,7 @@
         <exclude name="UnnecessaryObjectReferences"/>
         <exclude name="UnnecessaryBigDecimalInstantiation"/>
         <exclude name="UnnecessarySetter"/>
+        <exclude name="UnnecessaryGString"/>
     </ruleset-ref>
     <ruleset-ref path='rulesets/unused.xml'>
         <rule-config name='UnusedObject'>
--- a/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
@@ -851,7 +851,7 @@ public class InternalStreamConnection implements InternalConnection {
 
     private CommandEventSender createCommandEventSender(final CommandMessage message, final ByteBufferBsonOutput bsonOutput,
             final RequestContext requestContext) {
-        if (!isMonitoringConnection && opened() && (commandListener != null || COMMAND_PROTOCOL_LOGGER.isDebugEnabled(getClusterId()))) {
+        if (!isMonitoringConnection && opened() && (commandListener != null || COMMAND_PROTOCOL_LOGGER.isDebugRequired(getClusterId()))) {
             return new LoggingCommandEventSender(SECURITY_SENSITIVE_COMMANDS, SECURITY_SENSITIVE_HELLO_COMMANDS, description,
                     commandListener, requestContext, message, bsonOutput, COMMAND_PROTOCOL_LOGGER);
         } else {
--- a/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
@@ -82,6 +82,7 @@ import static com.mongodb.internal.connection.ProtocolHelper.getOperationTime;
 import static com.mongodb.internal.connection.ProtocolHelper.getRecoveryToken;
 import static com.mongodb.internal.connection.ProtocolHelper.getSnapshotTimestamp;
 import static com.mongodb.internal.connection.ProtocolHelper.isCommandOk;
+import static com.mongodb.internal.logging.StructuredLogMessage.Level.DEBUG;
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
 
@@ -851,7 +852,7 @@ public class InternalStreamConnection implements InternalConnection {
 
     private CommandEventSender createCommandEventSender(final CommandMessage message, final ByteBufferBsonOutput bsonOutput,
             final RequestContext requestContext) {
-        if (!isMonitoringConnection && opened() && (commandListener != null || COMMAND_PROTOCOL_LOGGER.isDebugRequired(getClusterId()))) {
+        if (!isMonitoringConnection && opened() && (commandListener != null || COMMAND_PROTOCOL_LOGGER.isRequired(DEBUG, getClusterId()))) {
             return new LoggingCommandEventSender(SECURITY_SENSITIVE_COMMANDS, SECURITY_SENSITIVE_HELLO_COMMANDS, description,
                     commandListener, requestContext, message, bsonOutput, COMMAND_PROTOCOL_LOGGER);
         } else {
--- a/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
@@ -30,6 +30,7 @@ import com.mongodb.ServerAddress;
 import com.mongodb.annotations.NotThreadSafe;
 import com.mongodb.connection.AsyncCompletionHandler;
 import com.mongodb.connection.ClusterConnectionMode;
+import com.mongodb.connection.ClusterId;
 import com.mongodb.connection.ConnectionDescription;
 import com.mongodb.connection.ConnectionId;
 import com.mongodb.connection.ServerConnectionState;
@@ -42,6 +43,7 @@ import com.mongodb.event.CommandListener;
 import com.mongodb.internal.async.SingleResultCallback;
 import com.mongodb.internal.diagnostics.logging.Logger;
 import com.mongodb.internal.diagnostics.logging.Loggers;
+import com.mongodb.internal.logging.StructuredLogger;
 import com.mongodb.internal.session.SessionContext;
 import com.mongodb.lang.Nullable;
 import org.bson.BsonBinaryReader;
@@ -845,15 +847,19 @@ public class InternalStreamConnection implements InternalConnection {
         }
     }
 
-    private static final Logger COMMAND_PROTOCOL_LOGGER = Loggers.getLogger("protocol.command");
+    private static final StructuredLogger COMMAND_PROTOCOL_LOGGER = new StructuredLogger("protocol.command");
 
     private CommandEventSender createCommandEventSender(final CommandMessage message, final ByteBufferBsonOutput bsonOutput,
             final RequestContext requestContext) {
-        if (opened() && (commandListener != null || COMMAND_PROTOCOL_LOGGER.isDebugEnabled())) {
+        if (!isMonitoringConnection && opened() && (commandListener != null || COMMAND_PROTOCOL_LOGGER.isDebugEnabled(getClusterId()))) {
             return new LoggingCommandEventSender(SECURITY_SENSITIVE_COMMANDS, SECURITY_SENSITIVE_HELLO_COMMANDS, description,
                     commandListener, requestContext, message, bsonOutput, COMMAND_PROTOCOL_LOGGER);
         } else {
             return new NoOpCommandEventSender();
         }
+    }
+
+    private ClusterId getClusterId() {
+        return description.getConnectionId().getServerId().getClusterId();
     }
 }
--- a/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
@@ -108,6 +108,7 @@ public class InternalStreamConnection implements InternalConnection {
     private static final Logger LOGGER = Loggers.getLogger("connection");
 
     private final ClusterConnectionMode clusterConnectionMode;
+    private final boolean isMonitoringConnection;
     private final ServerId serverId;
     private final ConnectionGenerationSupplier connectionGenerationSupplier;
     private final StreamFactory streamFactory;
@@ -139,10 +140,20 @@ public class InternalStreamConnection implements InternalConnection {
     }
 
     public InternalStreamConnection(final ClusterConnectionMode clusterConnectionMode, final ServerId serverId,
+            final ConnectionGenerationSupplier connectionGenerationSupplier,
+            final StreamFactory streamFactory, final List<MongoCompressor> compressorList,
+            final CommandListener commandListener, final InternalConnectionInitializer connectionInitializer) {
+        this(clusterConnectionMode, false, serverId, connectionGenerationSupplier, streamFactory, compressorList, commandListener,
+                connectionInitializer);
+    }
+
+    public InternalStreamConnection(final ClusterConnectionMode clusterConnectionMode, final boolean isMonitoringConnection,
+                                    final ServerId serverId,
                                     final ConnectionGenerationSupplier connectionGenerationSupplier,
                                     final StreamFactory streamFactory, final List<MongoCompressor> compressorList,
                                     final CommandListener commandListener, final InternalConnectionInitializer connectionInitializer) {
         this.clusterConnectionMode = clusterConnectionMode;
+        this.isMonitoringConnection = isMonitoringConnection;
         this.serverId = notNull("serverId", serverId);
         this.connectionGenerationSupplier = notNull("connectionGeneration", connectionGenerationSupplier);
         this.streamFactory = notNull("streamFactory", streamFactory);
--- a/driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
@@ -21,6 +21,7 @@ import com.mongodb.RequestContext;
 import com.mongodb.connection.ClusterId;
 import com.mongodb.connection.ConnectionDescription;
 import com.mongodb.event.CommandListener;
+import com.mongodb.internal.logging.StructuredLogMessage.Entry;
 import com.mongodb.internal.logging.StructuredLogger;
 import com.mongodb.lang.Nullable;
 import org.bson.BsonDocument;
@@ -30,8 +31,11 @@ import org.bson.codecs.RawBsonDocumentCodec;
 import org.bson.json.JsonMode;
 import org.bson.json.JsonWriter;
 import org.bson.json.JsonWriterSettings;
+import org.bson.types.ObjectId;
 
 import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
 
 import static com.mongodb.assertions.Assertions.assertNotNull;
@@ -72,33 +76,17 @@ class LoggingCommandEventSender implements CommandEventSender {
     @Override
     public void sendStartedEvent() {
         if (loggingRequired()) {
-            String commandString = redactionRequired ? "{}" : getTruncatedJsonCommand(commandDocument);
-            if (description.getServiceId() == null) {
-                logger.debug("Command started", getClusterId(),
-                        "Command \"%s\" started on database %s using a connection with driver-generated ID %d and server-generated ID %d "
-                                + "to %s:%s. The request ID is %s. Command: %s",
-                        "commandName", commandName,
-                        "databaseName", message.getNamespace().getDatabaseName(),
-                        "driverConnectionId", description.getConnectionId().getLocalValue(),
-                        "serverConnectionId", description.getConnectionId().getServerValue(),
-                        "serverHost", description.getServerAddress().getHost(),
-                        "serverPort", description.getServerAddress().getPort(),
-                        "requestId", message.getId(),
-                        "command", commandString);
-            } else {
-                logger.debug("Command started", getClusterId(),
-                        "Command \"%s\" started on database %s using a connection with driver-generated ID %d and server-generated ID %d "
-                                + "to %s:%s with service ID %s. The request ID is %s. Command: %s",
-                        "commandName", commandName,
-                        "databaseName", message.getNamespace().getDatabaseName(),
-                        "driverConnectionId", description.getConnectionId().getLocalValue(),
-                        "serverConnectionId", description.getConnectionId().getServerValue(),
-                        "serverHost", description.getServerAddress().getHost(),
-                        "serverPort", description.getServerAddress().getPort(),
-                        "serviceId", description.getServiceId(),
-                        "requestId", message.getId(),
-                        "command", commandString);
-            }
+            List<Entry> entries = new ArrayList<>();
+            StringBuilder builder = new StringBuilder("Command \"%s\" started on database %s");
+            entries.add(new Entry("commandName", commandName));
+            entries.add(new Entry("databaseName", message.getNamespace().getDatabaseName()));
+
+            appendCommonLogFragment(entries, builder);
+
+            builder.append(" Command: %s");
+            entries.add(new Entry("command", redactionRequired ? "{}" : getTruncatedJsonCommand(commandDocument)));
+
+            logger.debug("Command started", getClusterId(), builder.toString(), entries.toArray(new Entry[0]));
         }
 
         if (eventRequired()) {
@@ -113,6 +101,7 @@ class LoggingCommandEventSender implements CommandEventSender {
         commandDocument = null;
     }
 
+
     @Override
     public void sendFailedEvent(final Throwable t) {
         Throwable commandEventException = t;
@@ -124,30 +113,14 @@ class LoggingCommandEventSender implements CommandEventSender {
         long elapsedTimeNanos = System.nanoTime() - startTimeNanos;
 
         if (loggingRequired()) {
-            if (description.getServiceId() == null) {
-                logger.debug("Command failed", getClusterId(), commandEventException,
-                        "Command \"%s\" failed in %.2f ms using a connection with driver-generated ID %d and server-generated ID "
-                                + "%d to %s:%s. The request ID is %d.",
-                        "commandName", commandName,
-                        "durationMS", elapsedTimeNanos / NANOS_PER_MILLI,
-                        "driverConnectionId", description.getConnectionId().getLocalValue(),
-                        "serverConnectionId", description.getConnectionId().getServerValue(),
-                        "serverHost", description.getServerAddress().getHost(),
-                        "serverPort", description.getServerAddress().getPort(),
-                        "requestId", message.getId());
-            } else {
-                logger.debug("Command failed", getClusterId(), commandEventException,
-                        "Command \"%s\" failed in %.2f ms using a connection with driver-generated ID %d and server-generated ID "
-                                + "%d to %s:%s with service ID %s. The request ID is %d.",
-                        "commandName", commandName,
-                        "durationMS", elapsedTimeNanos / NANOS_PER_MILLI,
-                        "driverConnectionId", description.getConnectionId().getLocalValue(),
-                        "serverConnectionId", description.getConnectionId().getServerValue(),
-                        "serverHost", description.getServerAddress().getHost(),
-                        "serverPort", description.getServerAddress().getPort(),
-                        "serviceId", description.getServiceId(),
-                        "requestId", message.getId());
-            }
+            List<Entry> entries = new ArrayList<>();
+            StringBuilder builder = new StringBuilder("Command \"%s\" failed in %.2f ms");
+            entries.add(new Entry("commandName", commandName));
+            entries.add(new Entry("durationMS", elapsedTimeNanos / NANOS_PER_MILLI));
+
+            appendCommonLogFragment(entries, builder);
+
+            logger.debug("Command failed", getClusterId(), commandEventException, builder.toString(), entries.toArray(new Entry[0]));
         }
 
         if (eventRequired()) {
@@ -170,34 +143,19 @@ class LoggingCommandEventSender implements CommandEventSender {
         long elapsedTimeNanos = System.nanoTime() - startTimeNanos;
 
         if (loggingRequired()) {
+            List<Entry> entries = new ArrayList<>();
+            StringBuilder builder = new StringBuilder("Command \"%s\" succeeded in %.2f ms");
+            entries.add(new Entry("commandName", commandName));
+            entries.add(new Entry("durationMS", elapsedTimeNanos / NANOS_PER_MILLI));
+
+            appendCommonLogFragment(entries, builder);
+
+            builder.append(" Command reply: %s");
             BsonDocument responseDocumentForEvent = redactionRequired ? new BsonDocument() : reply;
             String replyString = redactionRequired ? "{}" : getTruncatedJsonCommand(responseDocumentForEvent);
-            if (description.getServiceId() == null) {
-                logger.debug("Command succeeded", getClusterId(),
-                        "Command \"%s\" succeeded in %.2f ms using a connection with driver-generated ID %d and server-generated ID %d to "
-                                + "%s:%s. The request ID is %d. Command reply: %s",
-                        "commandName", commandName,
-                        "durationMS", elapsedTimeNanos / NANOS_PER_MILLI,
-                        "driverConnectionId", description.getConnectionId().getLocalValue(),
-                        "serverConnectionId", description.getConnectionId().getServerValue(),
-                        "serverHost", description.getServerAddress().getHost(),
-                        "serverPort", description.getServerAddress().getPort(),
-                        "requestId", message.getId(),
-                        "reply", replyString);
-            } else {
-                logger.debug("Command succeeded", getClusterId(),
-                        "Command \"%s\" succeeded in %.2f ms using a connection with driver-generated ID %d and server-generated ID %d to "
-                                + " %s:%s with service ID %s. The request ID is %d. Command reply: %s",
-                        "commandName", commandName,
-                        "durationMS", elapsedTimeNanos / NANOS_PER_MILLI,
-                        "driverConnectionId", description.getConnectionId().getLocalValue(),
-                        "serverConnectionId", description.getConnectionId().getServerValue(),
-                        "serverHost", description.getServerAddress().getHost(),
-                        "serverPort", description.getServerAddress().getPort(),
-                        "serviceId", description.getServiceId(),
-                        "requestId", message.getId(),
-                        "reply", replyString);
-            }
+            entries.add(new Entry("reply", replyString));
+
+            logger.debug("Command succeeded", getClusterId(), builder.toString(), entries.toArray(new Entry[0]));
         }
 
         if (eventRequired()) {
@@ -218,6 +176,30 @@ class LoggingCommandEventSender implements CommandEventSender {
 
     private boolean eventRequired() {
         return commandListener != null;
+    }
+
+    private void appendCommonLogFragment(final List<Entry> entries, final StringBuilder builder) {
+        builder.append(" using a connection with driver-generated ID %d");
+        entries.add(new Entry("driverConnectionId", description.getConnectionId().getLocalValue()));
+
+        Integer connectionServerValue = description.getConnectionId().getServerValue();
+        if (connectionServerValue != null) {
+            builder.append(" and server-generated ID %d");
+            entries.add(new Entry("serverConnectionId", connectionServerValue));
+        }
+
+        builder.append(" to %s:%s");
+        entries.add(new Entry("serverHost", description.getServerAddress().getHost()));
+        entries.add(new Entry("serverPort", description.getServerAddress().getPort()));
+
+        ObjectId descriptionServiceId = description.getServiceId();
+        if (descriptionServiceId != null) {
+            builder.append(" with service ID %s");
+            entries.add(new Entry("serviceId", descriptionServiceId));
+        }
+
+        builder.append(". The request ID is %s.");
+        entries.add(new Entry("requestId", message.getId()));
     }
 
     private static String getTruncatedJsonCommand(final BsonDocument commandDocument) {
--- a/driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
@@ -208,7 +208,7 @@ class LoggingCommandEventSender implements CommandEventSender {
     }
 
     private boolean loggingRequired() {
-        return logger.isDebugEnabled(getClusterId());
+        return logger.isDebugRequired(getClusterId());
     }
 
 
--- a/driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
@@ -21,6 +21,7 @@ import com.mongodb.RequestContext;
 import com.mongodb.connection.ClusterId;
 import com.mongodb.connection.ConnectionDescription;
 import com.mongodb.event.CommandListener;
+import com.mongodb.internal.logging.StructuredLogMessage;
 import com.mongodb.internal.logging.StructuredLogMessage.Entry;
 import com.mongodb.internal.logging.StructuredLogger;
 import com.mongodb.lang.Nullable;
@@ -42,6 +43,8 @@ import static com.mongodb.assertions.Assertions.assertNotNull;
 import static com.mongodb.internal.connection.ProtocolHelper.sendCommandFailedEvent;
 import static com.mongodb.internal.connection.ProtocolHelper.sendCommandStartedEvent;
 import static com.mongodb.internal.connection.ProtocolHelper.sendCommandSucceededEvent;
+import static com.mongodb.internal.logging.StructuredLogMessage.Component.COMMAND;
+import static com.mongodb.internal.logging.StructuredLogMessage.Level.DEBUG;
 
 class LoggingCommandEventSender implements CommandEventSender {
     private static final int MAX_COMMAND_DOCUMENT_LENGTH_TO_LOG = 1000;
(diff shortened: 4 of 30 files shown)


config/codenarc/codenarc.xml
```diff
@@ -143,6 +143,7 @@
         <exclude name="UnnecessaryObjectReferences"/>
         <exclude name="UnnecessaryBigDecimalInstantiation"/>
         <exclude name="UnnecessarySetter"/>
+        <exclude name="UnnecessaryGString"/>
```
jyemin, line 146: I just got so sick of this.  As Java programmers writing occasional tests in Groovy, it seems unnecessary to enforce this constraint on us. Just allow the double quotes for simple strings and move on.

config/checkstyle/suppressions.xml
```diff
@@ -86,6 +86,7 @@
     <suppress checks="ParameterNumber" files="DefaultClusterableServerFactory"/>
     <suppress checks="ParameterNumber" files="Operations"/>
     <suppress checks="ParameterNumber" files="ChangeStreamDocument"/>
+    <suppress checks="ParameterNumber" files="StructuredLogger"/>
```
jyemin, line 89: Needed because of the the StructuredLogger varargs APIs

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -108,6 +111,7 @@ public class InternalStreamConnection implements InternalConnection {
     private static final Logger LOGGER = Loggers.getLogger("connection");
 
     private final ClusterConnectionMode clusterConnectionMode;
+    private final boolean isMonitoringConnection;
```
jyemin, line 114: So that we can not log commands sent on monitoring connections, as per spec

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
```diff
@@ -37,15 +38,15 @@
...
 
 class LoggingCommandEventSender implements CommandEventSender {
     private static final int MAX_COMMAND_DOCUMENT_LENGTH_TO_LOG = 1000;
+    private static final double NANOS_PER_MILLI = 1_000_000.0d;
 
     private final ConnectionDescription description;
     @Nullable private final CommandListener commandListener;
     private final RequestContext requestContext;
-    private final Logger logger;
+    private final StructuredLogger logger;
```
jyemin, line 49: Replaced Logger with StructuredLogger to allow implementation of the spec tests, which require the individual dynamic fields to be available for assertions.

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
```diff
@@ -71,12 +72,33 @@ class LoggingCommandEventSender implements CommandEventSender {
...
     public void sendStartedEvent() {
         if (loggingRequired()) {
-            String commandString = redactionRequired ? format("{\"%s\": ...", commandName) : getTruncatedJsonCommand();
-            // TODO: log RequestContext?
-            logger.debug(
-                    format("Sending command '%s' with request id %d to database %s on connection [%s] to server %s",
-                            commandString, message.getId(),
-                            message.getNamespace().getDatabaseName(), description.getConnectionId(), description.getServerAddress()));
+            String commandString = redactionRequired ? "{}" : getTruncatedJsonCommand(commandDocument);
+            if (description.getServiceId() == null) {
```
jyemin, line 76: These conditionals are unfortunate but necessary given [this requirement](https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst#id27).

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
```diff
@@ -71,12 +72,33 @@ class LoggingCommandEventSender implements CommandEventSender {
...
+                logger.debug("Command started", getClusterId(),
+                        "Command \"%s\" started on database %s using a connection with driver-generated ID %d and server-generated ID %d "
```
jyemin, line 78: The message formats were changed to match the specification.

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
```diff
@@ -91,38 +113,41 @@ public void sendStartedEvent() {
...
-    }
-
     @Override
     public void sendFailedEvent(final Throwable t) {
         Throwable commandEventException = t;
         if (t instanceof MongoCommandException && redactionRequired) {
-            commandEventException = new MongoCommandException(new BsonDocument(), description.getServerAddress());
+            MongoCommandException originalCommandException = (MongoCommandException) t;
+            commandEventException = new MongoCommandException(new BsonDocument(), originalCommandException.getServerAddress());
+            commandEventException.setStackTrace(t.getStackTrace());
```
jyemin, line 122: This seems the right thing to do.

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
```diff
@@ -133,53 +158,82 @@ public void sendFailedEvent(final Throwable t) {
...
-        }
+        sendSucceededEvent(responseBuffers.getResponseDocument(message.getId(), new RawBsonDocumentCodec()));
     }
 
     @Override
     public void sendSucceededEventForOneWayCommand() {
+        sendSucceededEvent(new BsonDocument("ok", new BsonInt32(1)));
+    }
+
+    private void sendSucceededEvent(final BsonDocument reply) {
```
jyemin, line 169: DRY-ed it up given that the spec doesn't differentiate the log messages for one-way command (which are used for unacknowledged writes)

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -0,0 +1,208 @@
...
+/**
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public class StructuredLogger {
+
+    private final Logger logger;
+
+    private static final Map<String, StructuredLoggingInterceptor> INTERCEPTORS = new HashMap<>();
+
+    public static void addInterceptor(final String applicationName, final StructuredLoggingInterceptor interceptor) {
```
jyemin, line 40: This is here so that the tests can intercept the log messages and make assertions.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -0,0 +1,208 @@
...
+    @VisibleForTesting(otherwise = PRIVATE)
+    public StructuredLogger(final Logger logger) {
+        this.logger = logger;
+    }
+
+    public boolean isDebugEnabled(final ClusterId clusterId) {
+        return logger.isDebugEnabled() || hasInterceptor(clusterId);
+    }
+
+    public void debug(final String message, final ClusterId clusterId, final String format,
```
jyemin, line 69: Many more of these messages will be added as more specs add logging.  These are just the ones need for command logging.

driver-core/src/test/unit/com/mongodb/internal/connection/LoggingCommandEventSenderSpecification.groovy
```diff
@@ -109,25 +109,29 @@ class LoggingCommandEventSenderSpecification extends Specification {
 
         then:
         1 * logger.debug {
-            it == "Sending command \'{\"ping\": 1, \"\$db\": \"test\"}\' with request id ${message.getId()} to database test " +
-                    "on connection [connectionId{localValue:${connectionDescription.connectionId.localValue}}] " +
-                    'to server 127.0.0.1:27017'
+            it == "Command \"ping\" started on database test using a connection with driver-generated ID " +
```
jyemin, line 112: This is what pushed me over the edge on disabling the unnecessary gstring codenarc rule

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/unified/CommandLoggingTest.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.junit.Assume.assumeFalse;
+
+public class CommandLoggingTest extends UnifiedReactiveStreamsTest {
+    public CommandLoggingTest(@SuppressWarnings("unused") final String fileDescription,
+                              @SuppressWarnings("unused") final String testDescription,
+                              final String schemaVersion, @Nullable final BsonArray runOnRequirements, final BsonArray entities,
+                              final BsonArray initialData, final BsonDocument definition) {
+        super(schemaVersion, runOnRequirements, entities, initialData, definition);
+        // The driver does not currently support operation identifiers
+        assumeFalse(fileDescription.equals("operation-id"));
```
jyemin, line 37: The same skipping that is necessary in CommandMonitoringTest

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
```diff
@@ -373,6 +379,30 @@ public String toString() {
...
+                final List<StructuredLogMessage> actualMessages) {
+            super();
+            this.client = client;
+            this.expectedMessages = expectedMessages;
+            this.actualMessages = actualMessages;
+        }
+
+        @Override
+        public String toString() {
+            // TODO: fix this up to be like the others
```
jyemin, line 398: Oops missed this one.  Will take a look

driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
```diff
@@ -74,6 +74,10 @@ private void assertValuesMatch(final BsonValue initialExpected, @Nullable final
                     case "$$matchesHexBytes":
                         expected = expectedDocument.getString("$$matchesHexBytes");
                         break;
+                    case "$$matchAsRoot":
+                        // TODO: figure out why no logic was needed here to get tests to pass
```
jyemin, line 78: Will investigate.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
rozza: Wish there was a neater / cleaner way to for this pattern.

Would `debug(final Supplier<String> logMessage, final Supplier<StructuredLogMessage> structuredLogMessage)` be cleaner?

Or even `debug(final Supplier<StructuredLogMessage> structuredLogMessage)` and have the structuredLogMessage actually handle calling logger.debug?

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -0,0 +1,208 @@
...
+        this(Loggers.getLogger(suffix));
+    }
+
+    @VisibleForTesting(otherwise = PRIVATE)
+    public StructuredLogger(final Logger logger) {
+        this.logger = logger;
+    }
+
+    public boolean isDebugEnabled(final ClusterId clusterId) {
+        return logger.isDebugEnabled() || hasInterceptor(clusterId);
```
rozza, line 66: This seems off, either the name of the method or clarifying that its a debug interceptor.  Recommend changing the method name.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    public void debug(final String message, final ClusterId clusterId, final String format,
+            final String k1, final Object v1,
+            final String k2, final Object v2,
+            final String k3, final Object v3,
+            final String k4, final Object v4,
+            final String k5, final Object v5,
+            final String k6, final Object v6) {
+        if (hasInterceptor(clusterId)) {
```
rozza, line 76: How expensive is this check?

rozza (commented):
Couple of initial comments.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: Looking at HashMap source code, looks like it amounts to a null check on `HashMap.table`

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: I opted not to do this because there is no reason to create `StructuredLogMessage` instances at all except during testing.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
stIncMale: I have used `@VisibleForTesting(otherwise = PRIVATE)` in such cases previously, and I think that's perfectly fine. An example: `Timeout.startNanos()`.

stIncMale: At first glance, it may apprear that one could introduce a "virtual" `AccessModifier.NONE` to convey that a program element would not even exist if it were not for tests, but I don't think this adds any useful information: if we have a `@VisibleForTesting(otherwise = PRIVATE)` element and observe that it is used neither in tests nor in the declaring class (I am obviously ignoring outer classes here, but the idea is the same), then we know it can be deleted. What could `AccessModifier.NONE` inform us about on top of that? It could only "prevent" us from using the annotated program element within the declaring class. However, such prevention does not make sense: the code in a class can do whatever it wants, the `@VisibleForTesting` is not there to restrict that; this annotation is supposed restrict the usage of exposed program elements *outside* of the class.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: I'm not sure what to name it.  I don't want clients to have to think about interceptors directly since it's just for testing, but I needed to expose it somehow because of the ClusterId.  Maybe `isDebugRequired`?

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -0,0 +1,210 @@
...
+            getInterceptor(clusterId).intercept(new StructuredLogMessage(logger.getName(), "debug", message, clusterId,
+                    new Entry(k1, v1), new Entry(k2, v2), new Entry(k3, v3), new Entry(k4, v4), new Entry(k5, v5), new Entry(k6, v6)));
+        }
+        if (logger.isDebugEnabled()) {
+            logger.debug(format(format, v1, v2, v3, v4, v5, v6));
+        }
+    }
+
+    public void debug(final String message, final ClusterId clusterId, final String format,
```
rozza, line 87: Some of these methods aren't used, so should be removed?

An alternative API might be

```
debug(final String message, final ClusterId clusterId, final Throwable exception, final String format, final Object... arguments)
```

Follows the slf4j convention and then you can process the arguments as kv pairs.

But I suppose that adds overhead.

rozza (commented):
Couple of TODOs left.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
rozza: Its only use is in:

```
    private boolean loggingRequired() {
        return logger.isDebugEnabled(getClusterId());
    }
```

So either `loggingRequired` or `shouldLog` seems more appropriate.

driver-core/src/test/unit/com/mongodb/internal/connection/LoggingCommandEventSenderSpecification.groovy
rozza: Huzzah! I hated that rule back when all tests were groovy tests - it always caught me out.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: Removed unused methods.

Yes, it would add overhead and I also appreciate the type safety that you get with the current approach

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: Right, but in the context of StructuredLogger you have to include the word "debug" in the method signature.  It's only "loggingRequired" in `LoggingEventCommandSender` because that class knows that it's only using debug logging.

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
rozza: ping

driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
rozza: ping

rozza (changes requested):
Looking good - just your todos left.

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
jyemin: Done

jyemin:
Fixed the first TODO.  Removed the second after confirming that the assertions are properly executed.  I'm not sure anymore what I was concerned about.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: Changed it to `isDebugRequired` based on above.  Let me know what you think.

jyemin:
I'm not sure what I was thinking was so hard about `failureIsRedacted`, but when I went back to look at it again it was all clear and simple so I just pushed the fix onto this PR.

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
kmahar: does this logic preserve the error code, code name, and error labels? as discussed in the spec [here](https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.rst#security)

kmahar: whoops, I meant to comment that one line up.

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
```diff
@@ -71,12 +72,33 @@ class LoggingCommandEventSender implements CommandEventSender {
...
+            if (description.getServiceId() == null) {
+                logger.debug("Command started", getClusterId(),
+                        "Command \"%s\" started on database %s using a connection with driver-generated ID %d and server-generated ID %d "
+                                + "to %s:%s. The request ID is %s. Command: %s",
+                        "commandName", commandName,
+                        "databaseName", message.getNamespace().getDatabaseName(),
+                        "driverConnectionId", description.getConnectionId().getLocalValue(),
+                        "serverConnectionId", description.getConnectionId().getServerValue(),
+                        "serverHost", description.getServerAddress().getHost(),
+                        "serverPort", description.getServerAddress().getPort(),
```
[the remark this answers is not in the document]
jyemin: There is always a port.  The driver does support unix domain sockets, in which the port 27017 will be reported here (which is incorrect, but you can't tell that from here anyway without some sort of instance-of check for `UnixServerAddress`.  I don't think it's worth it.

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
jyemin: No, it doesn't look like it.  Is there any way to add (unified) tests for that behavior?

driver-sync/src/test/functional/com/mongodb/client/unified/CommandLoggingTest.java
```diff
@@ -0,0 +1,52 @@
...
+public class CommandLoggingTest extends UnifiedSyncTest {
+
+
+    public CommandLoggingTest(@SuppressWarnings("unused") final String fileDescription,
+                              @SuppressWarnings("unused") final String testDescription,
+                              final String schemaVersion,
+                              @Nullable final BsonArray runOnRequirements, final BsonArray entities, final BsonArray initialData,
+                              final BsonDocument definition) {
+        super(schemaVersion, runOnRequirements, entities, initialData, definition);
+        assumeFalse(isServerlessTest());
```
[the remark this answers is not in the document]
jyemin: No, but I didn't try.  I just copied this line from the command monitoring tests, but maybe it shouldn't be there either.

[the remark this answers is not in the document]
kmahar: we run the command monitoring tests against serverless for Rust with no issue

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
kmahar: I don't think so, since we aren't requiring a specific error representation in the log messages and so drivers might just be asserting on a stringified error in the tests (this is what Rust is doing, since `tracing` doesn't allow us to attach non-primitive types to structured log messages). 

I tried to cover this in the test runner spec by saying 

> When failureIsRedacted is present and its value is true, the test runner MUST assert that a failure is present and that the failure has been redacted according to the rules defined for error redaction in the [command logging and monitoring specification](https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.rst#security).

and then in the definition of redaction, the preservation of code/codename/labels is discussed. 

I wrote unit tests for the redaction functionality in Rust to confirm this. we could consider adding prose unit tests I suppose

jyemin: I'd like to defer this work to a separate PR, just to keep this one from getting even more complicated.

jyemin: Added JAVA-4843

kmahar: SGTM

driver-core/src/main/com/mongodb/internal/logging/StructuredLogMessage.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.mongodb.connection.ClusterId;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public class StructuredLogMessage {
```
stIncMale, line 27: It is helpful to have `@Nullable` annotations on fields, methods and formal parameters where applicable, instead of analyzing the code each time we try to use internal API.

The same is applicable to `StructuredLogger`.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -0,0 +1,146 @@
...
+            getInterceptor(clusterId).intercept(new StructuredLogMessage(logger.getName(), "debug", message, clusterId,
+                    new Entry(k1, v1), new Entry(k2, v2), new Entry(k3, v3), new Entry(k4, v4), new Entry(k5, v5), new Entry(k6, v6),
+                    new Entry(k7, v7), new Entry(k8, v8)));
+        }
+        if (logger.isDebugEnabled()) {
+            logger.debug(format(format, v1, v2, v3, v4, v5, v6, v7, v8));
+        }
+    }
+
+    public void debug(final String message, final ClusterId clusterId, final String format,
```
stIncMale, line 90: What's called `message` in `StructuredLogger` is called `messageId` in `StructuredLogMessage`. Shouldn't we consistently use the same term `messageId` everywhere?

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
stIncMale: Currently, all the calls to `StructuredLogger.debug` are already guarded by `StructuredLogger.isDebugRequired`, which is why I don't understand attempts to reduce the overhead of creating objects when calling `StructuredLogger.debug`: `debug` can accept `StructuredLogMessage` as its input and it will have no performance effect as far as I can see.

Are there concerns that in the future it will become too inconvenient to do the `isDebugRequired` checks before calling `StructuredLogger.debug`? If not, then accepting `StructuredLogMessage` seems to be a better approach, and we will only need two overloads of `debug` (one that takes an exception and one that does not).

driver-core/src/main/com/mongodb/internal/logging/StructuredLoggingInterceptor.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.internal.logging;
+
+/**
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public interface StructuredLoggingInterceptor {
```
stIncMale, line 22: Shouldn't this interface be `@ThreadSafe`, which would also mean that `TestLoggingInterceptor` needs to be implemented in a thread-safe manner?

stIncMale (changes requested):
Within the 2 hours I was suggested to dedicate to reviewing the PR, I reviewed only the following:

- `StructuredLogger`
- `StructuredLogMessage`
- `StructuredLoggingInterceptor`

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
stIncMale: Ah, it's because we are trying to avoid creating `StructuredLogMessage`s when debug logging is enabled but there are no interceptors (the name of `StructuredLogger.isDebugRequired` is confusing).

stIncMale: The only alternative that does not have performance penalty that comes to my mind is accepting `Object... entries` (and I am not sure it's better): it can be passed to `String.format` as is, thus resulting in no overhead when debug logging is enabled but the interceptor is not present. And when the interceptor is present, we don't care about the overhead of converting it to `Entry... entries` (which is insignificant either way).

Pros: we need only two overloads of the `debug` method.
Cons: we move the checks that each name have a value and that each name is a string from compile-time to run-time.

```java
Entry[] constructEntries(final Object... deconstructedEntries) {
    assertTrue((deconstructedEntries.length & 1) == 0);
    Entry[] entries = new Entry[deconstructedEntries.length / 2];
    for (int i = 0; i < entries.length - 1; i += 2) {
        entries[i] = new Entry((String) deconstructedEntries[i], deconstructedEntries[i + 1]);
    }
    return entries;
}
```

```java
public void debug(final String message, final ClusterId clusterId, final Throwable exception, final String format,
        final Object... entries) {
    if (hasInterceptor(clusterId)) {
        getInterceptor(clusterId).intercept(new StructuredLogMessage(logger.getName(), "debug", message, clusterId, exception,
                constructEntries(entries)));
    }
    if (logger.isDebugEnabled()) {
        logger.debug(format(format, entries), exception);
    }
}
```

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -0,0 +1,146 @@
...
+import static java.lang.String.format;
+
+/**
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public class StructuredLogger {
+
+    private final Logger logger;
+
+    private static final Map<String, StructuredLoggingInterceptor> INTERCEPTORS = new HashMap<>();
```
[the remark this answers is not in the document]
jyemin: Are there any concerns about the behavior in production scenarios? (I don't think so given that `INTERCEPTORS` is a final variable.)  If that's the case, how worried should we be about the test scenario?  Are we willing to give up some performance here in production to make test scenarios safe?  I'm not so sure.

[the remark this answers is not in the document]
stIncMale: > Are there any concerns about the behavior in production scenarios?

If we are choosing to write code (even test code) that we can't seem to show to be correct for the sake of not harming performance of production code, I think we ought to at the very least (necessary but not sufficient) demonstrate that the alternative code that we can show to be correct, measurably hurts production performance.

> Are we willing to give up some performance here in production to make test scenarios safe?

In a situation when there are no writes at all, which is what we have in production, OpenJDK JDK `CHM.get` is different from `HM.get` in the two following ways:

1.  it computes the key has eagerly, which is weird because algorithmically this is not needed, i.e., it's likely an oversight, but it is also likely that JIT optimizes that computation away and that's why the authors don't care;
2. it does a single volatile read of a shared variable instead of a plain read, which, when there are no writes to the variable, barely have overhead over plain reads; but even if there were writes, we just don't care about the cost of a volatile read at the very least because `ch.qos.logback.classic.Logger.isDebugEnabled` does at least one such read as a result of calling `CopyOnWriteArrayList.size()`.

Hence the question: what performance penalty are we talking about?

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
```diff
@@ -91,38 +113,41 @@ public void sendStartedEvent() {
...
-                                    + "to server %s",
-                            message.getId(), getElapsedTimeFormattedInMilliseconds(elapsedTimeNanos), description.getConnectionId(),
-                            description.getServerAddress()),
-                    commandEventException);
+            if (description.getServiceId() == null) {
+                logger.debug("Command failed", getClusterId(), commandEventException,
+                        "Command \"%s\" failed in %.2f ms using a connection with driver-generated ID %d and server-generated ID "
+                                + "%d to %s:%s. The request ID is %d.",
+                        "commandName", commandName,
+                        "durationMS", elapsedTimeNanos / NANOS_PER_MILLI,
```
stIncMale, line 132: I propose to use `NANOSECONDS.toMillis(elapsedTimeNanos)` instead of reinventing it.

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
```diff
@@ -71,12 +72,33 @@ class LoggingCommandEventSender implements CommandEventSender {
...
+                        "commandName", commandName,
+                        "databaseName", message.getNamespace().getDatabaseName(),
+                        "driverConnectionId", description.getConnectionId().getLocalValue(),
+                        "serverConnectionId", description.getConnectionId().getServerValue(),
+                        "serverHost", description.getServerAddress().getHost(),
+                        "serverPort", description.getServerAddress().getPort(),
+                        "serviceId", description.getServiceId(),
+                        "requestId", message.getId(),
+                        "command", commandString);
+            }
```
stIncMale, line 101: 1. This unfortunate significant code duplication is caused by the [spec requirement](https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst#omitting-null-values-from-log-messages). If we had two logged values that may independently be `null`, would we quadruplicate the logging code?
2. The question is especially relevant because we log `description.getConnectionId().getServerValue()`, and `ConnectionId.getServerValue` is documented as `@Nullable`. Do we know that in all situations when we currently call `StructuredLogger.debug`, the server value is not `null`? If yes, could we add an assertion for it?

jyemin: 1.  Yes, I think we would
2. I'd be uncomfortable adding this assertion, as I think it would be the first of its kind for this value and may expose differences in implementations of various server implementation (mongocryptd, Data Lake, etc).  If the value is actually null, I don't think it actually causes a problem in production, but only if there was an assertion in the spec tests (which there isn't)

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
jyemin: It's done this way to avoid rounding, since partial milliseconds can be significant in high speed networks. Note that `NANOS_PER_MILLI` is a double not an int.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogMessage.java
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
stIncMale: 1. Do we want to somehow rise a concern about the spec requirement based on such consequences?
1.1. [regardless of what the answer to the previous question is] Do you actually think this is fine?
2. If we are not adding such an assertion because it may fail, then shouldn't we add two more logging branches to cover the nullability of `serverConnectionId` in addition to the nullability of `serviceId`?

jyemin: I think we're out of this mess at the cost of some extra allocations of Entry instances.  PTAL

driver-core/src/main/com/mongodb/internal/logging/StructuredLoggingInterceptor.java
jyemin: There is no good way to use the test interceptor in a scenario where there are multiple threads, since it relies on the order of the log messages.  If we ever do get into those types of scenarios, we can address it then.  

I suspect that most of our test infrastructure code is not actually thread safe for this or similar reasons.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: I decided to go all in on `StructuredLogMessage` in the latest commit, so now there is  just a single `log` method in `StructuredLogger`.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -0,0 +1,146 @@
...
+            final String k1, final Object v1,
+            final String k2, final Object v2,
+            final String k3, final Object v3,
+            final String k4, final Object v4,
+            final String k5, final Object v5,
+            final String k6, final Object v6,
+            final String k7, final Object v7,
+            final String k8, final Object v8) {
+        if (hasInterceptor(clusterId)) {
+            getInterceptor(clusterId).intercept(new StructuredLogMessage(logger.getName(), "debug", message, clusterId,
```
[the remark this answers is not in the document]
jyemin: Addressed this in the latest updates.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogMessage.java
jyemin: I made the package a NonNullApi and added Nullable as needed.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -834,15 +848,19 @@ public void onResult(@Nullable final ByteBuf result, @Nullable final Throwable t
...
-        if (opened() && (commandListener != null || COMMAND_PROTOCOL_LOGGER.isDebugEnabled())) {
+        if (!isMonitoringConnection && opened() && (commandListener != null || COMMAND_PROTOCOL_LOGGER.isRequired(DEBUG, getClusterId()))) {
             return new LoggingCommandEventSender(SECURITY_SENSITIVE_COMMANDS, SECURITY_SENSITIVE_HELLO_COMMANDS, description,
                     commandListener, requestContext, message, bsonOutput, COMMAND_PROTOCOL_LOGGER);
         } else {
             return new NoOpCommandEventSender();
         }
     }
+
+    private ClusterId getClusterId() {
```
rozza, line 863: Any chance of a NPE here?

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -0,0 +1,103 @@
...
+        //noinspection SwitchStatementWithTooFewBranches
+        switch (level) {
+            case DEBUG:
+                return logger.isDebugEnabled();
+            default:
+                throw new UnsupportedOperationException();
+        }
+   }
+
+    public void log(final StructuredLogMessage message, final String format) {
```
rozza, line 82: This is looking cleaner ;)

I think we can we negate the cost of always creating a StructuredLogMessage - by just providing a supplier. That way its only created in the testing environments where there is an interceptor.

Would require decoupling the logging message from the StructuredLogMessage.

Something like:

```
public void log(final Consumer<Logger> loggingConsumer, final Supplier<StructuredLogMessage> structuredLogMessageSupplier)
```

This way the cost only has to be applied if either logger.isDebugEnabled and/or there is an inteceptor.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
jyemin: No there is always a ClusterId, just not always a description (which is null checked)

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: Given that almost everywhere I look in the code we already guard with `is<Level>Enabled`, I'm inclined to just leave it like this.  But I'm not sure what the intent of the `loggingConsumer` is in your example.  Can you explain how that would be used?

rozza: My thoughts were instead of `LoggingCommandEventSender`having to be aware of the `StructuredLogger` requirements with the `loggingRequired` checks you could move all that inside.

Eg: 

```
// Change the method name to include the level
public void debug(final ClusterId clusterId, 
                  final Consumer<Logger> loggingConsumer, 
                  final Supplier<StructuredLogMessage> structuredLogMessageSupplier)
```

The usage becomes:

```
logger.debug(getClusterId(), 
             l -> { 
               l.debug(format("Sending command '%s' with request id %d to database %s on connection [%s] to server %s", 
               commandString, message.getId(), message.getNamespace().getDatabaseName(), 
               description.getConnectionId(), description.getServerAddress()));},
            () -> {
                // Create Entries
                List<Entry> ....
                // return the Structured Log Message
                return new StructuredLogMessage(COMMAND, DEBUG, "Command started", getClusterId(), entries);
            }  
);
```

That way all the guards can occur in the StructuredLogger and the StructuredLogMessage is only computed if its actually required.

kmahar (approved):
lgtm from a spec compliance perspective.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
jyemin: My first attempt at this did actually include a method per level but I didn't like how it required repeating the logging level in the `StructureLogMessage` parameter list, and then having to check for that to make sure they actually match.  That seemed odd to me so I changed it to a single `log` method

The other tricky thing is that sometimes (as in command logging) there is an expensive log entry value to compute, and if you have to do it twice, once in each lambda, there is no way to share the value between the two.  So we'd end up needing a conditional anyway to check whether logging is required, in order to create any expensive value.  Also, in general I think having to repeat the entry values is going to be cumbersome, especially in command logging where there are entries that are conditionally included based on whether they are null.

Another possibility to consider here is to emulate the [SLF4J 2.0 API](https://www.slf4j.org/api/org/slf4j/Logger.html), particularly `LoggingEventBuilder`.  Obviously that has been vetted pretty carefully.

rozza: `LoggingEventBuilder` looks good. Perhaps in 5.0 we could remove our own logging wrapper and just use SLF4J?

rozza: Definitely there are costs with either approach and both are cumbersome - lets move on and readdress in the future if required.

rozza (approved):
LGTM

driver-core/src/main/com/mongodb/internal/connection/LoggingCommandEventSender.java
```diff
@@ -71,12 +79,17 @@ class LoggingCommandEventSender implements CommandEventSender {
...
+            StringBuilder builder = new StringBuilder("Command \"%s\" started on database %s");
+            entries.add(new Entry("commandName", commandName));
+            entries.add(new Entry("databaseName", message.getNamespace().getDatabaseName()));
+
+            appendCommonLogFragment(entries, builder);
+
+            builder.append(" Command: %s");
+            entries.add(new Entry("command", redactionRequired ? "{}" : getTruncatedJsonCommand(commandDocument)));
+
+            logger.log(new StructuredLogMessage(COMMAND, DEBUG, "Command started", getClusterId(), entries), builder.toString());
```
stIncMale, line 92: [optional]

This is much better that what was previously. However, this code screams "use a specialized builder to make sure that when the builder for the format of an unstructured message is updated, the entries for the structured log message are also updated". So I envision something like

```java
static final class StructuredLogMessageBuilder {
    private final StringBuilder unstructuredMessageFormatBuilder;
    private final List<Entry> entries;
    StructuredLogMessageBuilder() {
        unstructuredMessageFormatBuilder = new StringBuilder();
        entries = new ArrayList<>();
    }
    StructuredLogMessageBuilder append(final String format, final Entry entry) {
        unstructuredMessageFormatBuilder.append(format);
        entries.add(entry);
        return this;
    }
    // These params can be moved to the constructor of `StructuredLogMessageBuilder`
    // and stored as fields there, but for the sake of imaginary efficiency,
    // I did not do that.
    StructuredLogMessage build(
            final StructuredLogMessage.Component component,
            final StructuredLogMessage.Level level,
            final String messageId,
            final ClusterId clusterId,
            @Nullable final Throwable exception) {
        return new StructuredLogMessage(component, level, messageId, clusterId, excep
    }
    String unstructuredMessageFormat() {
        return unstructuredMessageFormatBuilder.toString();
    }
}
```
which results in

```java
StructuredLogMessageBuilder logMessageBuilder = new StructuredLogMessageBuilder()
        .append("Command \"%s\" failed", new Entry("commandName", commandName))
        .append(" in %.2f ms", new Entry("durationMS", elapsedTimeNanos / NANOS_PER_MILLI));
appendCommonLogFragment(logMessageBuilder);
logger.log(
        logMessageBuilder.build(COMMAND, DEBUG, "Command failed", getClusterId(), commandEventException),
        logMessageBuilder.unstructuredMessageFormat());
```

stIncMale: Just for information (this thread was and continues to be optional):

1. To achieve _"could easily append a format string fragment that doesn't include exactly one placeholder"_ we may have `append(String format, Entry... entries)` instead of `append(String format, Entry entry)`.
2. We also don't have to use `StructuredLogMessageBuilder` when it's simpler to construct `StructuredLogMessage` without it.

driver-core/src/main/com/mongodb/internal/logging/StructuredLoggingInterceptor.java
jyemin: Done (and also annotated with `FunctionalInterface`.  I consider just replacing with `Consumer<StructuredLogMessage` but left it alone.

stIncMale (approved):

jyemin:
Merging on command line since trying to preserve two commits

PR #1068, opened 2023-01-18
Add `@MqlUnchecked`

stIncMale:


driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -108,7 +109,7 @@ public interface Expression {
     IntegerExpression isIntegerOr(IntegerExpression other);
     StringExpression isStringOr(StringExpression other);
     DateExpression isDateOr(DateExpression other);
-    <T extends Expression> ArrayExpression<T> isArrayOr(ArrayExpression<? extends T> other);
+    <T extends Expression> ArrayExpression<@MqlUnchecked T> isArrayOr(ArrayExpression<? extends T> other);
```
katcharov, line 112: I like this.

I think we should have no default, and specify the issue each time. For example, here, it should perhaps be `MqlUnchecked(GENERIC)`? This way, the user might click on the term to get guidance.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -43,6 +45,7 @@ default BooleanExpression getBoolean(final String fieldName, final boolean other
         return getBoolean(fieldName, of(other));
     }
 
+    @MqlUnchecked(TYPE)
     NumberExpression getNumber(String fieldName);
```
katcharov, line 49: I wonder if it should be called MqlAsserted(TYPE) instead? It seems to read a bit more easily, and suggest something on the user's part, rather than something perhaps unavoidable in the API? Unsure.

stIncMale: While the Java SE API does not provide conceptually similar methods, the Java language has a concept of _unchecked_ conversions. The Rust standard library provides functions where a user is supposed to guarantee something that is not checked by the function; it uses the [_"unchecked"_ suffix in the names of those functions](https://doc.rust-lang.org/std/?search=_unchecked), and even has an explicit [corresponding naming convention](https://rust-lang.github.io/api-guidelines/dependability.html?highlight=unchecked#dynamic-enforcement:~:text=The%20convention%20is%20to%20mark%20these%20opt%2Dout%20functions%20with%20a%20suffix%20like%20_unchecked%20or%20by%20placing%20them%20in%20a%20raw%20submodule.).

That's why I think that "unchecked" is a better word - we are following existing conventions.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
stIncMale: `MqlUnchecked(TYPE_ARGUMENT)` then I'd say, as the inferred/user-specified [type argument](https://docs.oracle.com/javase/specs/jls/se17/html/jls-4.html#jls-4.5.1) of the parameterized type returned by the method is exactly where the API gives up on checking.

PR #1070, opened 2023-01-20
Javadoc for expressions, part 2

katcharov:
The remaining docs for [JAVA-4799](https://jira.mongodb.org/browse/JAVA-4799), continuing #1059.

Also adds `has` to document, and a fix to multiply (found while writing docs).

### DIFF
--- a/driver-core/src/main/com/mongodb/annotations/Sealed.java
+++ b/driver-core/src/main/com/mongodb/annotations/Sealed.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ * Copyright 2010 The Guava Authors
+ * Copyright 2011 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Signifies that the annotated class or interface should be treated as sealed:
+ * it must not be extended or implemented.
+ *
+ * <p>Using such classes and interfaces is no different from using ordinary
+ * unannotated classes and interfaces.
+ *
+ * <p>This annotation does not imply that the API is experimental or
+ * {@link Beta}, or that the quality or performance of the API is inferior.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.TYPE)
+@Documented
+@Sealed
+public @interface Sealed {
+}
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -105,6 +105,7 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      * The {@linkplain #gt(Expression) largest} value all the values of
      * {@code this} array, or the {@code other} value if this array is empty.
      *
+     * @mongodb.server.release 5.2
      * @param other the other value.
      * @return the resulting value.
      */
@@ -114,6 +115,7 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      * The {@linkplain #lt(Expression) smallest} value all the values of
      * {@code this} array, or the {@code other} value if this array is empty.
      *
+     * @mongodb.server.release 5.2
      * @param other the other value.
      * @return the resulting value.
      */
@@ -124,6 +126,7 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      * {@code this} array, or all elements if the array contains fewer than
      * {@code n} elements.
      *
+     * @mongodb.server.release 5.2
      * @param n the number of elements.
      * @return the resulting value.
      */
@@ -134,6 +137,7 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      * {@code this} array, or all elements if the array contains fewer than
      * {@code n} elements.
      *
+     * @mongodb.server.release 5.2
      * @param n the number of elements.
      * @return the resulting value.
      */
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -247,6 +247,7 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      * the array is not empty.
      * If the array is empty then the behaviour of the API is not defined.
      *
+     * @mongodb.server.release 4.4
      * @return the resulting value.
      */
     @MqlUnchecked(PRESENT)
@@ -259,6 +260,7 @@ public interface ArrayExpression<T extends Expression> extends Expression {
      * the array is not empty.
      * If the array is empty then the behaviour of the API is not defined.
      *
+     * @mongodb.server.release 4.4
      * @return the resulting value.
      */
     T last();
--- a/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
+++ b/driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
@@ -20,7 +20,6 @@ import java.util.function.Function;
 
 import static com.mongodb.client.model.expressions.Expressions.of;
 import static com.mongodb.client.model.expressions.MqlUnchecked.Unchecked.PRESENT;
-import static com.mongodb.client.model.expressions.MqlUnchecked.Unchecked.NON_EMPTY;
 
 /**
  * An array {@link Expression value} in the context of the MongoDB Query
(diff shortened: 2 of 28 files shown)


driver-core/src/main/com/mongodb/client/model/expressions/Expressions.java
```diff
@@ -139,6 +171,14 @@ public static NumberExpression of(final Decimal128 of) {
         return new MqlExpression<>((codecRegistry) -> new AstPlaceholder(new BsonDecimal128(of)));
     }
 
+    /**
+     * Returns an {@linkplain ArrayExpression array} of
+     * {@linkplain NumberExpression numbers} corresponding to
+     * the provided {@link Decimal128 Decimal128s}.
```
stIncMale, line 177: [optional]

`{@link Decimal128 Decimal128s}` -> `{@link Decimal128}s` (the same with strings and instants below)?

1. It never makes sense to duplicate the linked element name, unless you link to a fully-qualified name but want to display only the simple name.
2. "s" is not part of the type name.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -58,67 +57,267 @@
...
+     *
+     * <p>The mapper may be used to transform the values of {@code this} array
+     * into {@linkplain NumberExpression numbers}. If no transformation is
+     * necessary, then the identify function {@code array.sum(v -> v)} should
+     * be used.
+     *
+     * @param mapper the mapper function.
+     * @return the resulting value.
+     */
     NumberExpression sum(Function<? super T, ? extends NumberExpression> mapper);
```
stIncMale, line 88: [here and in other places]

1. identify -> identity
2. [optional] `the identify function {@code array.sum(v -> v)} should be used` -> `the {@linkplain Function#identity() identity function} should be used`?

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -58,67 +57,267 @@
...
+     * necessary, then the identify function {@code array.join(v -> v)} should
+     * be used.
+     *
+     * @param mapper the mapper function.
+     * @return the resulting value.
+     */
     StringExpression join(Function<? super T, StringExpression> mapper);
 
+    /**
+     * The array-concatenation of all the array values of {@code this} array,
```
stIncMale, line 162: `array-concatenation` -> `{@linkplain #concat(ArrayExpression) array-concatenation}`?

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -58,67 +57,267 @@
...
+     * be used.
+     *
+     * @param mapper the mapper function.
+     * @return the resulting value.
+     * @param <R> the type of the elements of the array.
+     */
     <R extends Expression> ArrayExpression<R> concat(Function<? super T, ? extends ArrayExpression<? extends R>> mapper);
 
+    /**
+     * The set union of all the array values of {@code this} array,
```
stIncMale, line 178: `set union` -> `{@linkplain #union(ArrayExpression) set-union}`?

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -58,67 +57,267 @@
...
+     * @return the resulting value.
+     * @param <R> the type of the elements of the array.
+     */
     <R extends Expression> ArrayExpression<R> union(Function<? super T, ? extends ArrayExpression<? extends R>> mapper);
 
+    /**
+     * The {@linkplain MapExpression map} value corresponding to the
+     * {@linkplain EntryExpression entry} values of {@code this} array,
+     * via the provided {@code mapper}. Returns the empty array if the array
+     * is empty.
```
stIncMale, line 197: `Returns the empty array` -> `Returns the empty map`

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
```diff
@@ -58,67 +57,267 @@
...
-     * user asserts that i is in bounds for the array
+     * Returns the element at the provided index {@code i} for
+     * {@code this} array.
      *
-     * @param i
-     * @return
+     * <p>Warning: The use of this method is an assertion that
+     * the index {@code i} is in bounds for the array.
+     * If the index is out of bounds for this array, then
+     * the behaviour of the API is not defined.
```
stIncMale, line 218: Let's replace "defined" with "specified" here and in other places to be consistent with `MqlUnchecked` and to avoid allusions to ["undefined behavor"](https://en.cppreference.com/w/c/language/behavior) in C, that puts no restrictions are put on the behavior of the program.

driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
```diff
@@ -25,34 +25,142 @@
...
+    /**
+     * Converts {@code this} string to a {@linkplain DateExpression date},
+     * using the specified {@code timezone} and {@code format}.
+     *
+     * @mongodb.driver.manual reference/operator/aggregation/dateToString/#std-label-format-specifiers Format Specifiers
+     * @param format the format.
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
     DateExpression parseDate(StringExpression timezone, StringExpression format);
```
stIncMale, line 166: Let's link to `DateExpression#asString` from all `parseDate` methods.

driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
```diff
@@ -25,34 +25,142 @@
...
+    /**
+     * Converts {@code this} string to a {@linkplain DateExpression date}.
+     *
+     * <p>This will cause an error if this string does not represent a valid
+     * date string (such as "2018-03-20", "2018-03-20T12:00:00Z", or
+     * "2018-03-20T12:00:00+0500").
+     *
+     * @return the resulting value.
+     */
     DateExpression parseDate();
```
stIncMale, line 144: I seems that we should specify that this method behaves just like `parseDate(StringExpression format)` with the default format, which we should also specify.

driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
```diff
@@ -25,34 +25,142 @@
...
 
+    /**
+     * Converts {@code this} string to a {@linkplain DateExpression date},
+     * using the specified {@code format}.
+     *
+     * @mongodb.driver.manual reference/operator/aggregation/dateToString/#std-label-format-specifiers Format Specifiers
+     * @param format the format.
+     * @return the resulting value.
+     */
     DateExpression parseDate(StringExpression format);
```
stIncMale, line 155: We should document how this method behaves when the parsed data (and, correspondingly, the format) does not contain timezone information.

driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
stIncMale, line 166: We should document that this method errors if the parsed data (and, correspondingly, the format) contains timezone information.

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
```diff
@@ -21,45 +21,162 @@
 import static com.mongodb.client.model.expressions.Expressions.of;
 import static com.mongodb.client.model.expressions.MqlUnchecked.Unchecked.PRESENT;
 
+/**
+ * A map {@link Expression value} in the context of the MongoDB Query
+ * Language (MQL). An map is a finite, unordered a set of
```
stIncMale, line 26: 1. "unordered" - I don't think we should say that it's unordered. Having the entries ordered in a map is more important than having the fields ordered in a document (and the document type is documented as being ordered): there a user has no way of observing the order until a document is converted to a map (the user still may influence the order via the document API).
2. "unordered a set" -> "a set"? (though, given #1 this is irrelevant).

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -25,14 +25,14 @@
 import static com.mongodb.client.model.expressions.Expressions.ofMap;
 import static com.mongodb.client.model.expressions.MqlUnchecked.Unchecked.PRESENT;
 import static com.mongodb.client.model.expressions.MqlUnchecked.Unchecked.TYPE;
+import static com.mongodb.client.model.expressions.MqlUnchecked.Unchecked.TYPE_ARGUMENT;
 
 /**
  * Expresses a document value. A document is an ordered set of fields, where the
  * key is a string value, mapping to a value of any other expression type.
  */
```
stIncMale, line 33: In all previous PRs this doc was told to be considered a draft and not reviewed. Here it should be reviewed, but there is no way for reviewers to know that they should review it. I wonder what other docs like this one we have.

stIncMale, line 33: 1. `key is a string value` - we don't have keys in documents, we have fields (a pair of a name and a value). We could just "field name is a string" without saying "string value" to preserve the word "value" fully for when we talk about the field value.
2. Similarly to the documentation of a map, we should say that "no field name is repeated".

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
```diff
@@ -21,45 +21,162 @@
...
+ * {@link EntryExpression entries} of a certain type, such that no entry key
+ * is repeated. It is a mapping from keys to values.
+ *
+ * @param <T> the type of the elements
```
stIncMale, line 30: The map documentation does not define an "element". Should this be "entry value"?

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
```diff
@@ -21,45 +21,162 @@
...
 public interface MapExpression<T extends Expression> extends Expression {
 
+    /**
+     * True if {@code this} map has a value (including null) for
+     * the provided key.
+     *
+     * @param key the key.
+     * @return the resulting value.
+     */
     BooleanExpression has(StringExpression key);
```
stIncMale, line 41: This method is documented differently from `DocumentExpression.has`: here it is explicitly said that keys mapped to the null values are treated as being present, but the same is neither said nor is obvious in `DocumentExpression.has`. We should further clarify the `DocumentExpression.has` method.

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
```diff
@@ -21,45 +21,162 @@
...
+     * The value corresponding to the provided {@code key}, or the
+     * {@code other} value if an entry for the key is not present.
+     *
+     * @param key the key.
+     * @param other the other value.
+     * @return the resulting value.
+     */
     default T get(final String key, final T other) {
         return get(of(key), other);
     }
```
stIncMale, line 101: `not present` -> `not {@linkplain #has(StringExpression) present}`?

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
```diff
@@ -21,45 +21,162 @@
...
+     * the specified {@code key} absent.
+     *
+     * <p>This does not affect the original map.
+     *
+     * @param key the key.
+     * @return the resulting value.
+     */
     default MapExpression<T> unset(final String key) {
         return unset(of(key));
     }
```
stIncMale, line 151: 1. `absent` -> `{@linkplain #has(StringExpression) absent}`?
2. [optional] It appears that a single term should be chosen and used: either "has"/"does not have", or "present"/"not present", or "present"/"absent". But currently the documentation seemingly uses a random mix. Given that the method name is `has`, it would appear that "has"/"does not have" should be the terms. But maybe you will decide to rename the method `has`.

stIncMale (changes requested):
Reviewed the first 2 commits but the `DocumentExpression`.

Some comments on array are also applicable to maps/documents and vice-versa, comments on `BranchesIntermediary` are also sometimes applicable to `Branches`.

stIncMale (changes requested):
Reviewed `DocumentExpression`.

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
katcharov: Good catch. (For 2, I will omit - I think the code example is more helpful, and the v->v format more legible)

driver-core/src/main/com/mongodb/client/model/expressions/ArrayExpression.java
katcharov: 4 instances in directory "is not defined" replaced.

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
katcharov: Used "A map is a finite set of..."

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: I checked, and it is this one and String. We documented all the classes, and only a few methods of the original classes (boolean, and some others, which should have been fixed).

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -25,86 +25,487 @@
...
+     *
+     * <p>Note: Any field considered to be a document by this API
+     * will also be considered a map, and vice-versa.
+     *
+     * @mongodb.server.release 5.0
+     * @param fieldName the name of the field.
+     * @param other the other value.
+     * @return the resulting value.
+     */
     DocumentExpression getDocument(String fieldName, DocumentExpression other);
```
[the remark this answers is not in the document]
katcharov: Removed any "Note: Any field...", and replaced with alternative wording.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -25,86 +25,487 @@
...
+     * enforced by the API. The use of this method is an
+     * unchecked assertion that the field is present and
+     * the field value is of the specified type.
+     *
+     * @mongodb.server.release 5.0
+     * @param fieldName the name of the field.
+     * @return the resulting value.
+     * @param <T> the type.
+     */
+    @MqlUnchecked({PRESENT, TYPE, TYPE_ARGUMENT})
```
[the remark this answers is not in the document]
katcharov: It seems useful to remind users to carefully look at the type argument as well, as we do elsewhere.

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
```diff
@@ -21,45 +21,162 @@
...
 
+    /**
+     * The value corresponding to the provided {@code key}, or the
+     * {@code other} value if an entry for the key is not present.
+     *
+     * @param key the key.
+     * @param other the other value.
+     * @return the resulting value.
+     */
     default T get(final String key, final T other) {
```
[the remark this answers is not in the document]
katcharov: Good catch.

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -166,6 +165,7 @@ public interface Expression {
      * {@code this} is a number, or the {@code other} number value if
      * {@code this} is null, or is missing, or is of any other non-number type.
      *
+     * @mongodb.server.release 4.4
```
stIncMale, line 168: Moved this from https://github.com/mongodb/mongo-java-driver/pull/1032: We also need @since on the package and the new public classes/interfaces to express the driver version in which they were added.

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
```diff
@@ -23,11 +23,11 @@
 
 /**
  * A map {@link Expression value} in the context of the MongoDB Query
- * Language (MQL). An map is a finite, unordered a set of
- * {@link EntryExpression entries} of a certain type, such that no entry key
- * is repeated. It is a mapping from keys to values.
+ * Language (MQL). A map is a finite set of
```
stIncMale, line 26: "finite" is stated here but is not in the `DocumentExpression`. As a reader of the documentation I am now confused whether this is because a document may be an infinite set of fields?

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
stIncMale: My understanding is that you still want to not express that map entries are ordered. The order of map entries is an important property that is already there and we need good reasons to deprive users of it. What are those reasons?

driver-core/src/main/com/mongodb/client/model/expressions/BranchesIntermediary.java
```diff
@@ -37,66 +43,197 @@ private static <T extends Expression> MqlExpression<?> mqlEx(final T value) {
...
     @SuppressWarnings("unchecked")
-    public <Q extends Expression> BranchesIntermediary<T, R> isArray(final Function<? super ArrayExpression<Q>, ? extends R> r) {
-        return is(v -> mqlEx(v).isArray(), v -> r.apply((ArrayExpression<Q>) v));
-    }
-
-    public BranchesIntermediary<T, R> isDocument(final Function<? super DocumentExpression, ? extends R> r) {
-        return is(v -> mqlEx(v).isDocumentOrMap(), v -> r.apply((DocumentExpression) v));
-    }
-
+    public <Q extends Expression> BranchesIntermediary<T, R> isArray(final Function<? super ArrayExpression<Q>, ? extends R> mapping) {
```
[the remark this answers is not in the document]
stIncMale: The documentation of `Expression.isArrayOr` talks about "the type of elements of the ... array", while this documentation talks about the type argument of the parameterized type. Either approach works, but it would have been better if the documentation were consistent.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -25,86 +25,483 @@
...
+     * unchecked assertion that the type argument is correct.
+     *
+     * @mongodb.server.release 5.0
+     * @param fieldName the name of the field.
+     * @param other the other value.
+     * @return the resulting value.
+     * @param <T> the type.
+     */
+    @MqlUnchecked({TYPE_ARGUMENT})
     <T extends Expression> MapExpression<T> getMap(String fieldName, MapExpression<? extends T> other);
```
stIncMale, line 430: The placement of `MqlUnchecked` with `TYPE_ARGUMENT` is inconsistent within `DocumentExpression`.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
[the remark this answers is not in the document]
stIncMale: If it deemed important to talk about the raw type and the parameterization instead of simply talking about the type, then shouldn't the same be done in the javadoc comments? The current wording is _"The use of this method is an unchecked assertion that ... the field value is of the specified type"_ (this example is from `DocumentExpression.getMap(String)`).

driver-core/src/main/com/mongodb/client/model/expressions/MapExpression.java
katcharov: We discussed this. The proposal was to effectively use the wording at the class level:

> Despite a document in MongoDB being an ordered set of fields, users must not rely on this guarantee (or property) when using DocumentExpression/MapExpression."

I put a shorter variant of this on `entrySet`, in line with the docs on `distinct`. I could see no way for users to rely on this property via our API using Docs, and only entrySet can rely on it for Maps. (The behaviour of storing/fetching is outside of the scope of our API, though we try to maintain/round-trip properties.) Prominently mentioning it seems to give the misleading impression that this is a bigger concern that it is in practice.

driver-core/src/main/com/mongodb/client/model/expressions/StringExpression.java
```diff
@@ -21,38 +21,158 @@
...
 
+    /**
+     * Converts {@code this} string to a {@linkplain DateExpression date},
+     * using the specified {@code timezone} and {@code format}.
+     *
+     * @see DateExpression#asString()
+     * @see DateExpression#asString(StringExpression, StringExpression)
+     * @mongodb.driver.manual reference/operator/aggregation/dateToString/#std-label-format-specifiers Format Specifiers
```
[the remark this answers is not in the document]
katcharov: I said I would remember and then forgot.

driver-core/src/main/com/mongodb/annotations/Sealed.java
```diff
@@ -0,0 +1,42 @@
...
+ * <p>Using such classes and interfaces is no different from using ordinary
+ * unannotated classes and interfaces.
+ *
+ * <p>This annotation does not imply that the API is experimental or
+ * {@link Beta}, or that the quality or performance of the API is inferior.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.TYPE)
+@Documented
+@Sealed
```
stIncMale, line 40: What does `@Sealed` mean when placed on a declaration of an annotation (here and in `@MqlUnchecked`)? As far as I know, the Java language does not allow implementing or extending an annotation.

driver-core/src/main/com/mongodb/annotations/Evolving.java
```diff
@@ -29,11 +29,13 @@
...
- * Note that the presence of this annotation implies nothing about the quality or performance of the API in question.
+ *
+ * <p>This annotation does not imply that the API is experimental or
+ * {@link Beta}, or that the quality or performance of the API is inferior.
  */
 @Retention(RetentionPolicy.CLASS)
 @Target(ElementType.TYPE)
 @Documented
-@Evolving
+@Sealed
```
stIncMale, line 39: The approach we have in the team is to try and avoid doing drive-by changes. I suggest creating a separate PR with the `master` branch being its target as these changes do not belong with the work on the expressions API.

P.S. Just for information, in case you decide to propose the changes in a different PR (as this PR is not a place for them), the changes seem quite questionable to me:

1. I don't understand what `@Sealed` means here, and I doubt it is the right thing to do (I left a comment on `Sealed.java` about this).
2. I don't understand removing `@Evolving`. An explanation for it may be a good start.
3. As far as I can see, the change of the wording brings nothing.

driver-core/src/main/com/mongodb/client/model/expressions/EntryExpression.java
```diff
@@ -25,12 +28,15 @@
  * {@linkplain Expression value}. Entries are used with
  * {@linkplain MapExpression maps}.
  *
- * Entries are {@linkplain Expression#isDocumentOr document-like} and
+ * <p>Entries are {@linkplain Expression#isDocumentOr document-like} and
  * {@linkplain Expression#isMapOr map-like}, unless the method returning the
  * entry specifies otherwise.
```
stIncMale, line 33: [optional]

The closing tag is missing. You will find many places in the existing code where this is the case, but at some point we had a discussion (in a PR) were it was decided that we should specify closing tags. See `Locks` for an example of a new code that does this.

I only now noticed this, and after checking, saw that this is how all docs in the expressions API are written. Marking this as optional to avoid blocking the progress.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
```diff
@@ -249,19 +254,20 @@ public interface Expression {
      * @return the resulting value.
      * @param <T> the type of the values of the resulting map.
      */
-    <T extends Expression> MapExpression<T> isMapOr(MapExpression<? extends T> other);
+    <T extends Expression> MapExpression<T> isMapOr(MapExpression<@MqlUnchecked(TYPE_ARGUMENT) ? extends T> other);
```
stIncMale, line 257: Here and in `DocumentExpression.getMap(String, MapExpression)` we should annotate the type argument in the type that we return, not in the type that we accept (like it's in `isArrayOr`).

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
```diff
@@ -25,86 +28,500 @@
...
+     *
+     * <p>Warning: The type argument of the resulting value is not
+     * enforced by the API. The use of this method is an
+     * unchecked assertion that the type argument is correct.
+     *
+     * @return the resulting value.
+     * @param <T> the type.
+     */
+
+    <T extends Expression> MapExpression<T> asMap();
```
stIncMale, line 524: In https://github.com/mongodb/mongo-java-driver/pull/1059 this method returns `MapExpression<Expression>`, which is why the table does not list it. But now that it has changed, it also needs to be annotated.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/expressions/Expression.java
katcharov: Moved to next PR due to sensitive rebase.

driver-core/src/main/com/mongodb/client/model/expressions/DocumentExpression.java
katcharov: Moved to next PR due to sensitive rebase.

stIncMale (approved):

PR #1071, opened 2023-01-27
PojoCodec BsonId error message

rozza:
Help users when using a BsonId in a constructor and there is no known Id property in the model.

JAVA-4838

### DIFF
--- a/bson/src/test/unit/org/bson/codecs/pojo/entities/conventions/CreatorConstructorNoKnownIdModel.java
+++ b/bson/src/test/unit/org/bson/codecs/pojo/entities/conventions/CreatorConstructorNoKnownIdModel.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


jyemin (commented):
This doesn't match the description for JAVA-4838.  Is what's outlined there not possible or is there another PR coming?  If the former, let's change the ticket description to match this more limited scope of work.

rozza:
Ah I thought I had added a strike through of the original idea.

I have clarified the ticket and the reason not to support it for decoding.

bson/src/test/unit/org/bson/codecs/pojo/entities/conventions/CreatorConstructorNoKnownIdModel.java
```diff
@@ -0,0 +1,69 @@
...
+import org.bson.codecs.pojo.annotations.BsonProperty;
+
+import java.util.Objects;
+
+public class CreatorConstructorNoKnownIdModel {
+    private final String stringField;
+    private final long longField;
+
+    @BsonCreator
+    public CreatorConstructorNoKnownIdModel(
```
jyemin, line 30: Is the difference between this and the similar

```
  @BsonCreator
  public CreatorConstructorIdModel(@BsonId final String id, 
                                   @BsonProperty("integersField") final List<Integer> integerField,
                                   @BsonProperty("longField") final long longField)
````

just that in this existing one the property is named `id` and the BsonId annotation is ignored?  Is that why this one doesn't throw an exception?

I wonder why we allowed `ElementType.PARAMETER` as a target for `BsonId` in the first place...  This is the only place I see it used as a parameter.

rozza: `CreatorConstructorIdModel` has a public `getId()` method. The PojoCodec considers either `id` or `_id` properties as the `id` property. So in that case it can map to `id`.

I believe parameter support was added via  a PR in JAVA-2599. 

To be fair this is an edge case scenario and prevents a possible NPE.

jyemin (approved):
ok, i get it now.  LGTM

PR #1072, opened 2023-01-27
Support configuration of max document length for command logging

jyemin:
JAVA-4841

### DIFF
--- a/driver-core/src/main/com/mongodb/LoggerSettings.java
+++ b/driver-core/src/main/com/mongodb/LoggerSettings.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb;
+
+import com.mongodb.annotations.Immutable;
+
+import java.util.Objects;
+
+import static com.mongodb.assertions.Assertions.notNull;
+
+/**
+ * An immutable class representing settings for logging.
+ *
+ * <p>
+ * The driver logs using SLF4J 1.0 API with a root logger of {@code org.mongodb.driver}.
+ * </p>
+ *
+ * @since 4.9
+ */
+@Immutable
+public final class LoggerSettings {
+    private final int maxCommandLoggingDocumentLength;
+    /**
+     * Gets a builder for an instance of {@code LoggerSettings}.
+     * @return the builder
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Creates a builder instance.
+     *
+     * @param loggerSettings existing LoggerSettings to default the builder settings on.
+     * @return a builder
+     */
+    public static Builder builder(final LoggerSettings loggerSettings) {
+        return builder().applySettings(loggerSettings);
+    }
+
+    /**
+     * A builder for an instance of {@code LoggerSettings}.
+     */
+    public static final class Builder {
+        private int maxCommandLoggingDocumentLength = 1000;
+        private Builder() {
+        }
+
+        /**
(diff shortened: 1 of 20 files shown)


driver-core/src/main/com/mongodb/LoggerSettings.java
```diff
@@ -0,0 +1,143 @@
...
+     * them to the maximum length defined by this setting.
+     * </p>
+     *
+     * <p>
+     * Defaults to 1000 characters.
+     * </p>
+     *
+     * @return the max command logging document length
+     */
+    public int getMaxCommandLoggingDocumentLength() {
```
jyemin, line 111: This could be named just `maxDocumentLength` to generalize it, though not sure where else we'd ever apply this.

kmahar: I don't have any other cases for logging a document in mind right now that aren't a command or reply. but we did make the language in the spec vague as to uses, and had it apply to all extJSON:

> Drivers MUST support configuring the maximum logged length for extended JSON documents in log messages.

so I guess in theory it could come up in the future. maybe if decided to log something like a transaction recovery token or a change stream resume token? I did make the name more general in Rust just in case: https://github.com/mongodb/mongo-rust-driver/blob/c641a38ff02cb384a6ceb7d1d39a97bc270cb015/src/client/options/mod.rs#L530-L542

driver-core/src/main/com/mongodb/LoggerSettings.java
```diff
@@ -0,0 +1,143 @@
...
+ * An immutable class representing settings for logging.
+ *
+ * <p>
+ * The driver logs using SLF4J 1.0 API with a root logger of {@code org.mongodb.driver}.
+ * </p>
+ *
+ * @since 4.9
+ */
+@Immutable
+public final class LoggerSettings {
```
stIncMale, line 35: [optional]
This documentation seems to be a good place to leave a link to https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/logging.

driver-core/src/main/com/mongodb/LoggerSettings.java
stIncMale: Duplicating the "logging" part in the method name seems redundant: what else can this setting be for if not for logging given that the method is on the `LoggerSettings` class? Given that these are not internal methods, I think we should name them `getMaxDocumentLength`/`maxDocumentLength`.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/LoggerSettings.java
jyemin: Done

driver-core/src/main/com/mongodb/LoggerSettings.java
jyemin: Done

driver-core/src/main/com/mongodb/LoggerSettings.java
jyemin: Note: I couldn't use the existing javadoc taglets due to the common prefix that they use.  So just did it the old-fashioned way (which I like better anyway because it will show up in IntelliJ more clearly)

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnectionFactory.java
```diff
@@ -48,20 +50,22 @@ class InternalStreamConnectionFactory implements InternalConnectionFactory {
...
-                                    @Nullable final MongoCredentialWithCache credential,
-                                    @Nullable final String applicationName, final MongoDriverInformation mongoDriverInformation,
-                                    final List<MongoCompressor> compressorList,
-                                    @Nullable final CommandListener commandListener, @Nullable final ServerApi serverApi) {
+            final StreamFactory streamFactory,
+            @Nullable final MongoCredentialWithCache credential,
+            @Nullable final String applicationName, final MongoDriverInformation mongoDriverInformation,
+            final List<MongoCompressor> compressorList,
+            final LoggerSettings loggerSettings, @Nullable final CommandListener commandListener, @Nullable final ServerApi serverApi) {
         this.clusterConnectionMode = clusterConnectionMode;
```
katcharov, line 64: [optional] When a method header must be broken due to length (exceeds 120), could we as a rule place each parameter (including the first) on a new line, and order the assignments in parameter order? (For the same reasons that we do not put all fields and assignments on the same line.)

This also makes it easier to scan and correlate: for example, there is a null check on compressorList, but loggerSettings is not marked `@Nullable` and has no such check (and the same is the case for clusterConnectionMode).

stIncMale (approved):

katcharov (approved):
LGTM

PR #1073, opened 2023-01-27
Expressions final

katcharov:
Outstanding issues from [JAVA-3879](https://jira.mongodb.org/browse/JAVA-3879), primarily renames.

Largest renames were Expression->Mql, but this also includes (in separated commits) renames of methods that seemed inconsistent with similar methods, or which were otherwise problematic (for example, `strLen` became `length`, `concat(fn)` was renamed to `concatArrays` to distinguish it from array-array `concat`, and `has` was renamed to `hasField` for consistency with the other document accessors, and so on).

A dubious convenience method was removed, and missed null-checks and annotations were added (noticed in a prior PR, but too late to change there due to the scope of these renames).

Renames that were performed exclusively using automated tooling are in separate commits marked "automated".


stIncMale (approved):

katcharov:
Manually squashed some commits.

PR #1075, opened 2023-02-01
Ensure BatchCursorFlux does not drop an error

rozza:
The BatchCursor class can be closed by the BatchCursorFlux class but there is no signalling for the BatchCursor onNext publisher.

This leads to a next() called after the cursor was closed error and that breaks the reactive streams spec as the error ends up being dropped.

JAVA-4849

### DIFF
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
@@ -21,21 +21,29 @@ import reactor.core.publisher.Mono;
 
 import java.util.List;
 
-import static com.mongodb.reactivestreams.client.internal.MongoOperationPublisher.sinkToCallback;
-
 /**
  * <p>This class is not part of the public API and may be removed or changed at any time</p>
  */
 public class BatchCursor<T> implements AutoCloseable {
 
     private final AsyncBatchCursor<T> wrapped;
+    private volatile boolean cursorClosed = false;
 
     public BatchCursor(final AsyncBatchCursor<T> wrapped) {
         this.wrapped = wrapped;
     }
 
     public Publisher<List<T>> next() {
-        return Mono.create(sink -> wrapped.next(sinkToCallback(sink)));
+        return Mono.create(sink -> wrapped.next(
+                (result, t) -> {
+                    if (t != null && !cursorClosed) {
+                        sink.error(t);
+                    } else if (result == null) {
+                        sink.success();
+                    } else {
+                        sink.success(result);
+                    }
+                }));
     }
 
     public void setBatchSize(final int batchSize) {
@@ -51,6 +59,7 @@ public class BatchCursor<T> implements AutoCloseable {
     }
 
     public void close() {
+        cursorClosed = true;
         wrapped.close();
     }
 
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
@@ -17,13 +17,11 @@ package com.mongodb.reactivestreams.client.internal;
 
 import com.mongodb.internal.async.AsyncBatchCursor;
 import org.reactivestreams.Publisher;
-import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
 
 import java.util.List;
 import java.util.function.Supplier;
 
-import static com.mongodb.reactivestreams.client.internal.MongoOperationPublisher.sinkToCallback;
 
 /**
  * <p>This class is not part of the public API and may be removed or changed at any time</p>
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
@@ -21,6 +21,7 @@ import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
 
 import java.util.List;
+import java.util.function.Supplier;
 
 import static com.mongodb.reactivestreams.client.internal.MongoOperationPublisher.sinkToCallback;
 
@@ -36,13 +37,13 @@ public class BatchCursor<T> implements AutoCloseable {
     }
 
     public Publisher<List<T>> next() {
-        return Mono.create(sink -> wrapped.next(sinkToCallback(sink)));
+        return next(() -> false);
     }
 
-    public Publisher<List<T>> nextWithSink(final FluxSink<T> parentSink) {
+    public Publisher<List<T>> next(final Supplier<Boolean> hasBeenCancelled) {
         return Mono.create(sink -> wrapped.next(
                 (result, t) -> {
-                    if (!parentSink.isCancelled()) {
+                    if (!hasBeenCancelled.get()) {
                         if (t != null) {
                             sink.error(t);
                         } else if (result == null) {
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
@@ -21,29 +21,21 @@ import reactor.core.publisher.Mono;
 
 import java.util.List;
 
+import static com.mongodb.reactivestreams.client.internal.MongoOperationPublisher.sinkToCallback;
+
 /**
  * <p>This class is not part of the public API and may be removed or changed at any time</p>
  */
 public class BatchCursor<T> implements AutoCloseable {
 
     private final AsyncBatchCursor<T> wrapped;
-    private volatile boolean cursorClosed = false;
 
     public BatchCursor(final AsyncBatchCursor<T> wrapped) {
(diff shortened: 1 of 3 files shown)


driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -85,12 +87,14 @@ private void recurseCursor(){
             } else {
                 batchCursor.setBatchSize(calculateBatchSize(sink.requestedFromDownstream()));
                 Mono.from(batchCursor.next())
-                        .doOnCancel(this::closeCursor)
```
rozza, line 88: There are no hooks to actually cancel this Mono.

jyemin (commented):
Does the test reproduce the error reported in JAVA-4849?  It looks rather different, and I'm having trouble understanding exactly how this fix relates to the original report.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
```diff
@@ -299,6 +302,30 @@ void changeStreamPublisherCompletesAfterDroppingCollection() {
         subscriber.assertNoErrors();
     }
 
+    @Test
+    @DisplayName("Ensure BatchCursor does not drop an error")
+    public void testBatchCursorDoesNotDropAnError() {
+        try {
+            AtomicBoolean errorDropped = new AtomicBoolean();
+            Hooks.onErrorDropped(t -> errorDropped.set(true));
```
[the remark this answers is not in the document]
rozza: > Dropping hooks are invoked when the source of an operator does not comply with the Reactive Streams specification. These kind of errors are outside of the normal execution path (that is, they cannot be propagated through onError).

https://projectreactor.io/docs/core/release/reference/#hooks

rozza:
To add some more context - the "error" reported is a logging message similar to:

> 14:34:25.594 [Thread-17] ERROR reactor.core.publisher.Operators - Operator called default onErrorDropped
reactor.core.Exceptions$ErrorCallbackNotImplemented: com.mongodb.MongoException: next() called after the cursor was closed.
Caused by: com.mongodb.MongoException: next() called after the cursor was closed.

So the program works as expected it just logs an error as the default `onErrorDropped` is called. 

The test ensures that after the fix no `onErrorDropped` hook is called.

https://projectreactor.io/docs/core/release/reference/#hooks-dropping

rozza:
Updated to check the parent sink to see if its been cancelled. This should be a better flag to use here.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
```diff
@@ -38,6 +39,21 @@ public Publisher<List<T>> next() {
         return Mono.create(sink -> wrapped.next(sinkToCallback(sink)));
     }
 
+    public Publisher<List<T>> nextWithSink(final FluxSink<T> parentSink) {
+        return Mono.create(sink -> wrapped.next(
+                (result, t) -> {
+                    if (!parentSink.isCancelled()) {
+                        if (t != null) {
```
jyemin, line 46: Any way to test the first two conditions?

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
```diff
@@ -299,6 +302,30 @@ void changeStreamPublisherCompletesAfterDroppingCollection() {
...
+            Mono.from(collection.insertMany(createDocs(200))).block();
+
+            Flux.fromStream(IntStream.range(1, 200).boxed())
```
jyemin, line 313: I guess this test is timing-dependent?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
```diff
@@ -38,6 +39,21 @@ public Publisher<List<T>> next() {
         return Mono.create(sink -> wrapped.next(sinkToCallback(sink)));
     }
 
+    public Publisher<List<T>> nextWithSink(final FluxSink<T> parentSink) {
```
jyemin, line 42: No way to just have a single method here?  It could cause future bugs if callers are confused about which one to use.

jyemin: One thought I had is to add the parent FluxSink as a field, but I'm not sure how we'd get it through to the constructor.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
rozza: Well the test case ensures that sink.error is not called so does test it.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
rozza: I'll investigate - I think cursor.first() doesn't use a parent sink to handle the flattening of the batches so doesn't follow the same pattern - but will check.

rozza: Turns out Mono Sink and Flux Sink aren't the same - so couldnt remove both methods. I did neaten it up though with a Supplier instead.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/internal/BatchCursorFluxTest.java
rozza: Yes and resource dependent.  

Essentially, this test forces 400 queries to the collection but abandons 50% of the cursors as one of the queries returns a result quicker than the other. This ends up triggering the race consistently.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
jyemin: It tests the false branch but not the true branch.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -84,7 +84,7 @@ private void recurseCursor(){
                 sink.complete();
             } else {
                 batchCursor.setBatchSize(calculateBatchSize(sink.requestedFromDownstream()));
-                Mono.from(batchCursor.nextWithSink(sink))
+                Mono.from(batchCursor.next(() -> sink.isCancelled()))
```
jyemin, line 87: With the most recent commit, I now wonder whether we should instead leave BatchCursor alone, and add a conditional here that tests for `sink.isCancelled()` (though I'm not totally sure what would go in the true branch).

What do you think?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
rozza: Normal usage tests that the parent branch is not cancelled. The test will test both, cancelled and not cancelled.  As its a race results may return before the cancellation.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
rozza: Its tricky as we have a guard against [!sink.isCancelled()](https://github.com/mongodb/mongo-java-driver/pull/1075/files/c23be9fd486b9eb9295a4d7d4d1dacb9c4d8f155..df97b0e55e27bd9fa67c597588619add657fc39b#diff-82d8a56e77330e2af60a4b410b30501aa7374a90a90a7e8079952e09b9f2758bR82), we also check if the `batchCursor.isClosed()` before trying to create the Mono.

So the race is from the start of `recurseCursor` to it being actioned by the `subscribe()` call.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
rozza: Not sure how we can easily test `!parentSink.isCancelled() && t != null`.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
jyemin: OK, I'm not thinking clearly.  We have to defer the check for isCancelled until later.  It doesn't matter what the value is right now (hence using a Supplier).

jyemin (approved):
LGTM

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursor.java
rozza: Added a test - I killed the cursor between next calls :)

PR #1076, opened 2023-02-01
Update the legacy uber packages pom.xml

rozza:
Update the name and description to indicate these are legacy and not the most recent release.

JAVA-4854

### DIFF
--- a/driver-async/build.gradle
+++ b/driver-async/build.gradle
@@ -15,10 +15,11 @@
  */
 
 archivesBaseName = 'mongodb-driver-async'
-description = 'The MongoDB Asynchronous Driver.'
-
+description = '''This artifact is not the most recent release of the MongoDB Java Driver.
+                |To use the most recent release of the MongoDB Java Driver, please use mongodb-driver-reactivestreams instead.'''.stripMargin()
 ext {
-    pomName = 'The MongoDB Asynchronous Driver'
+    pomName = 'Legacy MongoDB Asynchronous Driver'
+    pomUrl = 'https://www.mongodb.com/docs/drivers/reactive-streams/'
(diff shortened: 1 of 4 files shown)


jyemin (commented):
Should we do the same for async (and suggest reactive streams)?

rozza:
@jyemin good idea - also update embedded description as well.

jyemin (commented):
I have some concern that this will cause a different confusion for users.  If they are using mongo-java-driver or mongodb-driver with the legacy API (com.mongodb.MongoClient), and they just switch to mongodb-driver-sync, they will be confused because what they really need is mongodb-driver-legacy.  What do you think about mentioning that artifact as well?

CC @ashni-mongodb

rozza:
Its a tricky one - saying this is a legacy driver and then suggesting another legacy driver might also be confusing :(

jyemin:
Maybe we shouldn't call the 3.x driver legacy?

ashni-mongodb:
Hmm agreed that there might be confusion marking this as legacy. What if we call it unsupported or unmaintained so that we can retain the "legacy" language for the "mongodb-driver-legacy" artifact?

Something like this, maybe?

"This driver is no longer maintained / updated by MongoDB. For the most up-to-date version of the Java Driver, please see mongodb-driver-sync. To use the legacy API, please refer to mongodb-driver-legacy."

jyemin:
Maybe "this artifact" instead of "this driver".  And use the work artifact as a suffix for all the names, so it's clear what we're talking about.

jyemin (approved):
That looks good.  Ship it!

katcharov (approved):

PR #1079, opened 2023-02-04
Add the `ClientEncryption.createEncryptedCollection` helper method

stIncMale:
Links to the relevant spec changes with their status are gathered and being kept updated in this [comment](https://jira.mongodb.org/browse/JAVA-4679?focusedCommentId=5155565&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5155565).

JAVA-4679


driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
```diff
@@ -360,4 +361,19 @@ public String toString() {
...
+     * @since 4.9
+     */
+    @Override
+    public CreateCollectionOptions clone() {
+        try {
+            return (CreateCollectionOptions) super.clone();
+        } catch (CloneNotSupportedException e) {
+            throw fail(e.toString());
+        }
+    }
```
stIncMale, line 378: 1. Implementing `Cloneable` on `CreateCollectionOptions` is backward-compatible. Also, this implementation does not screw over classes that extend `CreateCollectionOptions`: they can override `clone` by calling `super.clone`, cast to their type and it will work.
2. I hesitated between creating a copy constructor and adding `clone`. While I prefer copy constructors, `clone` guarantees that we won't forget to copy new fields in the copy constructor when we add more fields.

I am not full of confidence here and is open to hearing opinions if there are any.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
```diff
@@ -183,6 +198,85 @@ public Publisher<RewrapManyDataKeyResult> rewrapManyDataKey(final Bson filter, f
                         }));
     }
 
+    @Override
+    public Publisher<BsonDocument> createEncryptedCollection(final MongoDatabase database, final String collectionName,
```
stIncMale, line 202: @rozza, your knowledge of the Project Reactor and the reactive streams programming is being needed when reviewing this method.

The intended behavior of the implementation of this method can be peeked either in the synchronous implementation, or in the spec: [Create Encrypted Collection Helper](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#create-encrypted-collection-helper).

The thing that made this implementation really difficult for me (beyond the lack of knowledge are):

- if either `createDataKey` or `createCollection` fails, the method must attach to the error signal the (possibly partially) updated `encryptedFields` that it would have produced in case of a success.

I achieved the needed behavior by capturing `dataKeyMayBeCreated` and `actualEncryptedFields` in the returned `Publisher`. As far as I understand, the `Publisher` is constructed such that there are no data races when updating the `actualEncryptedFields` document (updates are done to the unrelated `field` documents), and all reads of the updated `actualEncryptedFields` are ordered after (in the HB order) the updates because of the combination of `then` and `defer` methods.

Even if this implementation is correct (at the very least I doubt it is idiomatic), it can't work if there are more than one `Subscriber` (regardless of whether they exist concurrently or not), so had make the result a one-shot publisher.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
```diff
@@ -183,6 +198,85 @@ public Publisher<RewrapManyDataKeyResult> rewrapManyDataKey(final Bson filter, f
...
+                        .filter(field -> Objects.equals(field.get(keyIdBsonKey), BsonNull.VALUE))
+                        // here we rely on the `createDataKey` publisher being cold, i.e., doing nothing until it is subscribed to
+                        .map(field -> Mono.fromDirect(createDataKey(kmsProvider, dataKeyOptions))
+                                // This is the closest we can do with reactive streams to setting the `dataKeyMayBeCreated` flag
+                                // immediately before calling `createDataKey`.
+                                .doOnSubscribe(subscription -> dataKeyMayBeCreated.set(true))
+                                .map(dataKey -> new SimpleImmutableEntry<>(field, dataKey))
+                        )
+                        .iterator();
+                Flux<SimpleImmutableEntry<BsonDocument, BsonBinary>> publisherOfFieldsAndDataKeyIds = Flux.concat(publishersOfFieldsAndDataKeyIds)
```
stIncMale, line 247: As a result of `Flux.concat` all data keys are created sequentially. This way we stop as soon as there is an error, instead possibly continuing to invoke `createDataKey` despite there already being an error.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
```diff
@@ -183,6 +198,85 @@ public Publisher<RewrapManyDataKeyResult> rewrapManyDataKey(final Bson filter, f
...
+                        // Therefore, all write actions in `doOnNext` above happen-before the invocation of `createCollection`,
+                        // which means `createCollection` is guaranteed to observe all those write actions, i.e.,
+                        // it is guaranteed to observe the updated document via the `actualEncryptedFields` reference.
+                        // Similarly, the updated document is observed in all other places below via the `actualEncryptedFields` reference.
+                        .thenEmpty(Mono.defer(() -> Mono.fromDirect(database.createCollection(collectionName,
+                                createCollectionOptions.clone().encryptedFields(actualEncryptedFields))))
+                        )
+                        .doOnError(RuntimeException.class, e -> {
+                            if (dataKeyMayBeCreated.get()) {
+                                throw new MongoUpdatedEncryptedFieldsException(actualEncryptedFields, format("Failed to create %s.", namespace), e);
```
stIncMale, line 267: Exceptions from `doOnError`/`defer` are handled by the Project Reactor such that the reactive streams protocol is not violated. I am not sure what is a better / more idiomatic way of transforming the error.

P.S. While writing this comment, I tried to find a better way again and found `Flux.handle`. Is this what is supposed to be used?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
```diff
@@ -183,6 +198,85 @@ public Publisher<RewrapManyDataKeyResult> rewrapManyDataKey(final Bson filter, f
...
+                            } else {
+                                throw e;
+                            }
+                        });
```
stIncMale, line 271: Exceptions from `doOnError`/`defer` are handled by the Project Reactor such that the reactive streams protocol is not violated. I am not sure what is a better / more idiomatic way of transforming the error.

P.S. While writing this comment, I tried to find a better way again and found `Flux.handle`. Is this what is supposed to be used?

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
```diff
@@ -171,6 +183,65 @@ public RewrapManyDataKeyResult rewrapManyDataKey(final Bson filter, final Rewrap
...
+        }
+        CodecRegistry codecRegistry = options.getKeyVaultMongoClientSettings() == null
+                ? MongoClientSettings.getDefaultCodecRegistry()
+                : options.getKeyVaultMongoClientSettings().getCodecRegistry();
+        BsonDocument actualEncryptedFields = rawEncryptedFields.toBsonDocument(BsonDocument.class, codecRegistry).clone();
+        BsonValue fields = actualEncryptedFields.get("fields");
+        if (fields != null) {
+            if (!fields.isArray()) {
+                throw new MongoConfigurationException(format("`encryptedFields` is incorrectly configured for the collection %s."
+                        + " `encryptedFields.fields` must be an array, but is %s type.", namespace, fields.getBsonType()));
```
stIncMale, line 206: This exception is not according to the current spec, and comes from the following:

- https://mongodb.slack.com/archives/C0406ECL478/p1675269046945409
- https://jira.mongodb.org/browse/DRIVERS-2540 (does not have a completed spec change), which proposes to throw in the `createCollection` methods. And I [think](https://jira.mongodb.org/browse/DRIVERS-2540?focusedCommentId=5159364&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5159364) that if `createCollection` methods need to throw, so should the `createEncryptedCollection` method.

If there is a concern that we should not throw this exception until (if ever) it is approved as a spec change, I will undo.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
```diff
@@ -183,6 +198,85 @@ public Publisher<RewrapManyDataKeyResult> rewrapManyDataKey(final Bson filter, f
...
+            final CreateCollectionOptions createCollectionOptions, final CreateEncryptedCollectionParams createEncryptedCollectionParams) {
+        notNull("collectionName", collectionName);
+        notNull("createCollectionOptions", createCollectionOptions);
+        notNull("createEncryptedCollectionParams", createEncryptedCollectionParams);
+        // this publisher captures and mutates state, which is why it has to be one-shot
+        return OneShotPublisher.from(Mono.defer(() -> {
+            MongoNamespace namespace = new MongoNamespace(database.getName(), collectionName);
+            Bson rawEncryptedFields = createCollectionOptions.getEncryptedFields();
+            if (rawEncryptedFields == null) {
+                throw new MongoConfigurationException(format("`encryptedFields` is not configured for the collection %s.", namespace));
```
stIncMale, line 212: Instead of throwing here, it is likely more idiomatic to return `Mono.error`, given that we are in the `Supplier` that is passed to `Mono.defer`. However, if we to follow this approach consistently, then the whole `Supplier` passed to `Mono.defer` should be wrapped in a `try` statement that catches all non-fatal exceptions and returns `Mono.error` instead of throwing them.

But I see no reason to do that:

- it is one more thing we should always remember to do;
- it arguably makes the code less readable;
- why doing that when the Project Reactor already [wraps](https://github.com/reactor/reactor-core/blob/v3.5.0/reactor-core/src/main/java/reactor/core/publisher/MonoDefer.java#L43-L50) the `Supplier` passed to `Mono.defer` in the `try` statement and signals `onError` for us?

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
```diff
@@ -171,6 +183,65 @@ public RewrapManyDataKeyResult rewrapManyDataKey(final Bson filter, final Rewrap
...
+        notNull("createEncryptedCollectionParams", createEncryptedCollectionParams);
+        MongoNamespace namespace = new MongoNamespace(database.getName(), collectionName);
+        Bson rawEncryptedFields = createCollectionOptions.getEncryptedFields();
+        if (rawEncryptedFields == null) {
+            throw new MongoConfigurationException(format("`encryptedFields` is not configured for the collection %s.", namespace));
+        }
+        CodecRegistry codecRegistry = options.getKeyVaultMongoClientSettings() == null
+                ? MongoClientSettings.getDefaultCodecRegistry()
+                : options.getKeyVaultMongoClientSettings().getCodecRegistry();
+        BsonDocument actualEncryptedFields = rawEncryptedFields.toBsonDocument(BsonDocument.class, codecRegistry).clone();
```
rozza, line 201: If a user used a `RawBsonDocument` that might trip us up here.

I'm not sure if we need to guard against that, its just its immutable and will error later.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
rozza: I would do this work outside of the construction of the publisher. That way it follows the general validation eg `notNull` checks and reduces the code in the callable.

rozza (commented):
Added a few comments.

I think there may be a more "functional" way with the reactive streams approach.  But I liked the procedural nature of it and how it follows the sync version :+1:

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
stIncMale: I initially had it like that, but then noticed that the other `MongoConfigurationException` we throw if `fields` is not an array, should be signalled via the publisher: we need to encode the `encryptedFields` config to `BsonDocument`, and this does not seem anymore like just an argument check.

Reporting `MongoConfigurationException` from the same method in two different ways (`throw` or signal `onError)` seems weird, so I decided to also signal the first `MongoConfigurationException` that you commented on. Given this consideration, does the code seem better now?

driver-sync/src/main/com/mongodb/client/vault/ClientEncryption.java
```diff
@@ -190,6 +195,32 @@ public interface ClientEncryption extends Closeable {
...
+     * @throws MongoUpdatedEncryptedFieldsException If an exception happens after creating at least one data key.
+     * This exception makes the updated {@code encryptedFields}
+     * {@linkplain MongoUpdatedEncryptedFieldsException#getEncryptedFields() available} to the caller.
+     *
+     * @since 4.9
+     * @mongodb.server.release 6.0
+     * @mongodb.driver.manual reference/command/create Create Command
+     */
+    @Beta(Beta.Reason.SERVER)
+    BsonDocument createEncryptedCollection(MongoDatabase database, String collectionName, CreateCollectionOptions createCollectionOptions,
```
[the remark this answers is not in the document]
stIncMale: `MongoUpdatedEncryptedFieldsException` is unchecked because `MongoException` is unchecked, and `MongoUpdatedEncryptedFieldsException` is of the `MongoException` type.

If you are talking about `throws`, then the JLS says: [_"It is permitted but not required to mention unchecked exception classes ([§11.1.1](https://docs.oracle.com/javase/specs/jls/se19/html/jls-11.html#jls-11.1.1)) in a throws clause."_](https://docs.oracle.com/javase/specs/jls/se19/html/jls-8.html#jls-8.4.6:~:text=It%20is%20permitted%20but%20not%20required%20to%20mention%20unchecked%20exception%20classes%20(%C2%A711.1.1)%20in%20a%20throws%20clause.) It appears that we use it in the public API extremely rarely, but we do: `DnsClient.getResourceRecordData`.

Since this appear to cause confusion, I'll remove the `throws` clause.

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
```diff
@@ -171,6 +183,65 @@ public RewrapManyDataKeyResult rewrapManyDataKey(final Bson filter, final Rewrap
...
+            }
+            String kmsProvider = createEncryptedCollectionParams.getKmsProvider();
+            DataKeyOptions dataKeyOptions = new DataKeyOptions();
+            BsonDocument masterKey = createEncryptedCollectionParams.getMasterKey();
+            if (masterKey != null) {
+                dataKeyOptions.masterKey(masterKey);
+            }
+            String keyIdBsonKey = "keyId";
+            // any mutable non-thread-safe Boolean should do
+            AtomicBoolean dataKeyMayBeCreated = new AtomicBoolean();
```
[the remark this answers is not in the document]
stIncMale: This method is required to ignore non-documents: [_"If `F` is not a document element, skip it."_](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#create-encrypted-collection-helper:~:text=If%20F%20is%20not%20a%20document%20element%2C%20skip%20it.).

This follow the general approach: avoid validating in the driver what can be validated by the server (if a type system prevents specifying an invalid value via the driver API - that's fine).

The reason `fields.isArray()` is checked instead of relaying the check to the server is described in DRIVERS-2540 (I mentioned this in https://github.com/mongodb/mongo-java-driver/pull/1079#discussion_r1096460072).

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
stIncMale: Thanks! Given that `RawBsonDocument` is immutable, the best solution here appears to fix the `RawBsonDocument.clone` method. Currently, despite the return type of the method being `BsonDocument`, the method returns `RawBsonDocument`. Moreover, it creates the new `RawBsonDocument` by cloning `RawBsonDocument.bytes`, which, as far as I can see, has no effect beyond wasting resources.

We can change the implementation of `RawBsonDocument.clone` such that it returns `toBaseBsonDocument()` - a normal mutable `BsonDocument`. Given that currently `RawBsonDocument.clone` is not documented beyond what `Object.equals` says, this is not a breaking change. What do you think?

config/spotbugs/exclude.xml
```diff
@@ -240,4 +240,18 @@
         <Method name="acquirePermitOrGetAvailableOpenedConnection"/>
         <Bug pattern="NS_NON_SHORT_CIRCUIT"/>
     </Match>
+
+    <!-- Can actually be null, but is not annotated as `@Nullable`. Annotating it as such causes warnings
+         in other places where `null` is not handled, see https://jira.mongodb.org/browse/JAVA-4861.
+         When the aforementioned ticket is done, it will be clear what to do with the warnings suppressed here. -->
```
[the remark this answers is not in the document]
stIncMale: Done (updated the ticket description).

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
```diff
@@ -171,6 +183,65 @@ public RewrapManyDataKeyResult rewrapManyDataKey(final Bson filter, final Rewrap
...
+                fields.asArray()
+                        .stream()
+                        .filter(BsonValue::isDocument)
+                        .map(BsonValue::asDocument)
+                        .filter(field -> field.containsKey(keyIdBsonKey))
+                        .filter(field -> Objects.equals(field.get(keyIdBsonKey), BsonNull.VALUE))
+                        .forEachOrdered(field -> {
+                            // It is crucial to set the `dataKeyMayBeCreated` flag either immediately before calling `createDataKey`,
+                            // or after that in a `finally` block.
+                            dataKeyMayBeCreated.set(true);
```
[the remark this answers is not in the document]
stIncMale: Done.

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
```diff
@@ -171,6 +183,65 @@ public RewrapManyDataKeyResult rewrapManyDataKey(final Bson filter, final Rewrap
...
+                        + " `encryptedFields.fields` must be an array, but is %s type.", namespace, fields.getBsonType()));
+            }
+            String kmsProvider = createEncryptedCollectionParams.getKmsProvider();
+            DataKeyOptions dataKeyOptions = new DataKeyOptions();
+            BsonDocument masterKey = createEncryptedCollectionParams.getMasterKey();
+            if (masterKey != null) {
+                dataKeyOptions.masterKey(masterKey);
+            }
+            String keyIdBsonKey = "keyId";
+            // any mutable non-thread-safe Boolean should do
```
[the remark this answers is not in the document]
stIncMale: Updated. Do you find the updated text clearer?

driver-core/src/main/com/mongodb/client/model/CreateEncryptedCollectionParams.java
```diff
@@ -0,0 +1,85 @@
...
+ * @since 4.9
+ */
+@Beta(Beta.Reason.SERVER)
+public final class CreateEncryptedCollectionParams {
+    private final String kmsProvider;
+    @Nullable
+    private BsonDocument masterKey;
+
+    /**
+     * A constructor.
```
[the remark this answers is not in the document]
stIncMale: For some reason, I was under the impression that we have static checks enforcing something to be said. But apparently, we don't 👏 Done.

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
```diff
@@ -171,6 +183,65 @@ public RewrapManyDataKeyResult rewrapManyDataKey(final Bson filter, final Rewrap
...
+                            dataKeyMayBeCreated.set(true);
+                            BsonBinary dataKeyId = createDataKey(kmsProvider, dataKeyOptions);
+                            field.put(keyIdBsonKey, dataKeyId);
+                        });
+                database.createCollection(collectionName, createCollectionOptions.clone().encryptedFields(actualEncryptedFields));
+            } catch (RuntimeException e) {
+                if (dataKeyMayBeCreated.get()) {
+                    throw new MongoUpdatedEncryptedFieldsException(actualEncryptedFields, format("Failed to create %s.", namespace), e);
+                } else {
+                    throw e;
```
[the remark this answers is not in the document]
stIncMale: If it happens to be a checked exception, then it must have been thrown sneakily, and I am not sure we should wrap it in a `MongoException`. `RetryingSyncSupplier.get`, updated in [Replace RuntimeException in catch with Exception to support Kotlin #971](https://github.com/mongodb/mongo-java-driver/pull/971) (Maxim linked in a different comment) wraps checked (sneakily thrown) exceptions in `RuntimeException` instead of wrapping them in `MongoException`. I did the same here.

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
```diff
@@ -171,6 +183,65 @@ public RewrapManyDataKeyResult rewrapManyDataKey(final Bson filter, final Rewrap
...
+                        .filter(field -> Objects.equals(field.get(keyIdBsonKey), BsonNull.VALUE))
+                        .forEachOrdered(field -> {
+                            // It is crucial to set the `dataKeyMayBeCreated` flag either immediately before calling `createDataKey`,
+                            // or after that in a `finally` block.
+                            dataKeyMayBeCreated.set(true);
+                            BsonBinary dataKeyId = createDataKey(kmsProvider, dataKeyOptions);
+                            field.put(keyIdBsonKey, dataKeyId);
+                        });
+                database.createCollection(collectionName, createCollectionOptions.clone().encryptedFields(actualEncryptedFields));
+            } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
stIncMale: Done.

stIncMale:
@rozza

> I think there may be a more "functional" way with the reactive streams approach.

What is that approach? I am most of all interested in how to attach the data key IDs to `MongoUpdatedEncryptedFieldsException` with a functional approach?

driver-core/src/main/com/mongodb/client/model/CreateCollectionOptions.java
stIncMale: Done.

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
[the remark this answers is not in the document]
rozza: Thanks :+1:

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
rozza: Lets do this work in a separate ticket.

katcharov (approved):
lgtm

rozza:
>@rozza
> > I think there may be a more "functional" way with the reactive streams approach.
>
>What is that approach? I am most of all interested in how to attach the data key IDs to MongoUpdatedEncryptedFieldsException with a functional approach? 

Me and my big mouth - turned out to be a fun exercise! Here is my attempt of a more functional approach:

```
 @Override
    public Publisher<BsonDocument> createEncryptedCollection(final MongoDatabase database, final String collectionName,
            final CreateCollectionOptions createCollectionOptions, final CreateEncryptedCollectionParams createEncryptedCollectionParams) {
        notNull("collectionName", collectionName);
        notNull("createCollectionOptions", createCollectionOptions);
        notNull("createEncryptedCollectionParams", createEncryptedCollectionParams);

        MongoNamespace namespace = new MongoNamespace(database.getName(), collectionName);
        Bson rawEncryptedFields = createCollectionOptions.getEncryptedFields();
        if (rawEncryptedFields == null) {
            throw new MongoConfigurationException(format("`encryptedFields` is not configured for the collection %s.", namespace));
        }
        CodecRegistry codecRegistry = options.getKeyVaultMongoClientSettings() == null
                ? MongoClientSettings.getDefaultCodecRegistry()
                : options.getKeyVaultMongoClientSettings().getCodecRegistry();

        BsonDocument encryptedFields = rawEncryptedFields.toBsonDocument(BsonDocument.class, codecRegistry);
        BsonValue fields = encryptedFields.get("fields");
        if (fields != null) {
            if (!fields.isArray()) {
                throw new MongoConfigurationException(format("`encryptedFields` is incorrectly configured for the collection %s."
                        + " `encryptedFields.fields` must be an array, but is %s type.", namespace, fields.getBsonType()));
            }

            String kmsProvider = createEncryptedCollectionParams.getKmsProvider();
            DataKeyOptions dataKeyOptions = new DataKeyOptions();
            BsonDocument masterKey = createEncryptedCollectionParams.getMasterKey();
            if (masterKey != null) {
                dataKeyOptions.masterKey(masterKey);
            }
            String keyIdBsonKey = "keyId";

            return Flux.concat(fields.asArray()
                            .stream()
                            .map(field -> {
                                Mono<BsonValue> newField = Mono.fromSupplier(() -> field);
                                if (field.isDocument()) {
                                    BsonDocument fieldDocument = field.asDocument();
                                    if (fieldDocument.get(keyIdBsonKey, BsonBoolean.FALSE).equals(BsonNull.VALUE)) {
                                        newField = Mono.fromDirect(createDataKey(kmsProvider, dataKeyOptions))
                                                .map(dataKeyId -> (BsonValue) fieldDocument.append(keyIdBsonKey, dataKeyId))
                                                .contextWrite(ctx -> ctx.put("createdDataKey", "true"));
                                    }
                                }
                                return newField;
                            })
                            .collect(Collectors.toList()))
                    .collectList()
                    .map(fieldsArray -> encryptedFields.clone().append("fields", new BsonArray(fieldsArray)))
                    .flatMap(actualEncryptedFields ->
                            Mono.deferContextual(ctx ->
                                    Mono.from(database.createCollection(collectionName,
                                                    createCollectionOptions.clone().encryptedFields(actualEncryptedFields)))
                                            .doOnError(t -> ctx.hasKey("createdDataKey"), e -> {
                                                throw new MongoUpdatedEncryptedFieldsException(actualEncryptedFields,
                                                        format("Failed to create %s.", namespace), e);
                                            })
                                            .thenReturn(actualEncryptedFields))
                    );
        }
        return Mono.from(database.createCollection(collectionName, createCollectionOptions))
                .thenReturn(encryptedFields);
    }
```

* User input validation is done outside of the Publisher.
* All side effects are kept within the bounds of the Publishers and propagated downstream for consumption by later Publishers.
* I had to use the [Context-Propagation](https://projectreactor.io/docs/core/release/reference/#context.propagation) to capture if I had called `createDataKey` for "Case 3: Invalid keyId". Not used that before but it is handy.
* This should remove the need for using the `oneShotPublisher` - although like 99% of our Publishers its non nonsensical to subscribe multiple times.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionAutoDataKeysTest.java
```diff
@@ -0,0 +1,248 @@
...
+     * See
+     * <a href="https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#case-3-invalid-keyid">
+     * Case 3: Invalid keyId</a>.
+     */
+    @ParameterizedTest(name = DISPLAY_NAME_PLACEHOLDER + " {0}")
+    @MethodSource("arguments")
+    void invalidKeyId(final KmsProvider kmsProvider) {
+        CreateCollectionOptions createCollectionOptions = new CreateCollectionOptions().encryptedFields(Document.parse(
+                "{"
+                + "  fields: [{"
```
rozza, line 152: Should there be a test for `{fields: []}` ? Is that valid or invalid and does the code create the collection?

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
stIncMale: Submitted https://github.com/mongodb/mongo-java-driver/pull/1081 for review.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionAutoDataKeysTest.java
stIncMale: > Should there be a test for {fields: []} ?

Neither the current spec nor the [upcoming test changes](https://jira.mongodb.org/browse/DRIVERS-2538) have such a test. There are a few existing CSFLE tests that use `fields: []`, for example https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/legacy/fle2-NoEncryption.json#L16-L18.

I don't know if we need a test for this value. Do you want me to add one?

> Is that valid or invalid and does the code create the collection?

Whether an empty `fields` array is a valid or not configuration is not specified anywhere:

- https://www.mongodb.com/docs/manual/reference/command/create/
- https://www.mongodb.com/docs/manual/core/csfle/quick-start/
- https://mongodb.github.io/mongo-java-driver/4.8/apidocs/mongodb-driver-core/com/mongodb/client/model/CreateCollectionOptions.html#encryptedFields(org.bson.conversions.Bson)
- https://mongodb.github.io/mongo-java-driver/4.8/apidocs/mongodb-driver-core/com/mongodb/AutoEncryptionSettings.html#getEncryptedFieldsMap()

An experiment shows that the server creates a collection when `fields` is an empty array.

rozza: Thanks for looking into it - as CSFLE tests cover it, resolving.

stIncMale:
@rozza Thanks! I had to read-up about `Context`.

> All side effects are kept within the bounds of the Publishers and propagated downstream for consumption by later Publishers.
> I had to use the Context-Propagation to capture if I had called createDataKey

`Context` can't be used to propagate information downstream because it only propagates information upstream (from a subscriber through the chain of operators to the source of the data). I read about this in the section [Adding a Context to a Reactive Sequence](https://projectreactor.io/docs/core/release/reference/index.html#context).

I also experimented just in case, and as expected, the context entry with the `createdDataKey` key is not propagated downstream.

----

However, I realized that the publisher indeed does not have to be one-shot because the mutable state it captures is created once per `Subscriber` by the `Supplier` passed to `Mono.defer`.

I took your idea to use `doOnError(Predicate, Consumer)`.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
stIncMale: Note that `doOnError` suppresses the original `e` when we throw `MongoUpdatedEncryptedFieldsException`, instead of treating the thrown exception as a replacement for the original one. As a result, we have `e` as both the cause of `MongoUpdatedEncryptedFieldsException` and an element in the suppressed set, which causes ugly stack-traces and implies that we are abusing `doOnError`. However, it may be that the Reactor simply does not support what we need.

driver-sync/src/main/com/mongodb/client/internal/ClientEncryptionImpl.java
stIncMale: I undone this part because I [realized](https://jira.mongodb.org/browse/DRIVERS-2540?focusedCommentId=5180748&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5180748) that the proposal in DRIVERS-2540 is not fully-baked.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
stIncMale: Based on your experimental code [here](https://github.com/mongodb/mongo-java-driver/pull/1079#issuecomment-1421137054), I infer that you still think that failures to render `Bson` should be treated as fatal and not signalled via the returned publisher. So I do this work outside of the construction of the publisher as you originally requested.

rozza (approved):
LGTM!

I had a couple of thoughts regarding your comments - but they don't require any actioning:

> Context can't be used to propagate information downstream because it only propagates information upstream 

According to the [context.write](https://projectreactor.io/docs/core/release/reference/index.html#context.write) docs:
> The Context is propagated from the bottom of the chain towards the top.

So I read that as in order of the chain (operations), the bottom being the `publishersOfUpdatedFields` in your code and the top being the final Mono. 

This makes sense to me as, I used the context in order to get the `invalidKeyId`test to pass which was failing when not using the context.


> My understanding is that the function provided to map is called only if collectList() emits a list, which happens only if there were no errors. On top of that, collectList is documented as "discards the elements in the List upon cancellation or error triggered by a data signal."

One to check, if the fields list is empty, then `collectList` will either emit an empty list or just complete. The `map` stage becomes a noop either way and will complete without any changes.

The impact of an error or cancellation signals would be the same for either approach. An error in any stage will bubble up to the top and be caught by any error handlers.  I think the same for cancellation signals as well.

stIncMale:
@rozza, for the sake of knowledge sharing, I have to disagree:

> The Context is propagated from the bottom of the chain towards the top.
> So I read that as in order of the chain (operations), the bottom being the publishersOfUpdatedFields in your code and the top being the final Mono.

The following example from the [docs](https://projectreactor.io/docs/core/release/reference/index.html#_simple_context_examples):

```java
String key = "message";
Mono<String> r = Mono.just("Hello")
    .contextWrite(ctx -> ctx.put(key, "World")) 
    .flatMap( s -> Mono.deferContextual(ctx ->
        Mono.just(s + " " + ctx.getOrDefault(key, "Stranger")))); 

StepVerifier.create(r)
            .expectNext("Hello Stranger") 
            .verifyComplete();
```

demonstrates what _"from the bottom of the chain towards the top"_ mean. If it worked according to your interpretation, the emitted string would have been "Hello World", but it is "Hello Stranger".

----

> I used the context in order to get the invalidKeyIdtest to pass which was failing when not using the context.

The `invalidKeyId` test neither triggers nor checks for `MongoUpdatedEncryptedFieldsException`, so even if we remove the `doOnError` call from either your or my implementation, it passes.

----

> One to check, if the fields list is empty, then collectList will either emit an empty list or just complete. The map stage becomes a noop either way and will complete without any changes.
>
> The impact of an error or cancellation signals would be the same for either approach. An error in any stage will bubble up to the top and be caught by any error handlers. I think the same for cancellation signals as well.

As I mentioned in the PR description, not all needed tests are done in the spec, the spec change in DRIVERS-2538 is still not done. However, we can experiment by temporarily changing the code outside of the `createEncryptedCollection` method.

1. If we modify `ClientEncryptionImpl.createDataKey` such that it throws an exception, `createEncryptedCollection` must emit `MongoUpdatedEncryptedFieldsException`. The implementation proposed in https://github.com/mongodb/mongo-java-driver/pull/1079#issuecomment-1421137054 does not do that, while the implementation in the PR does.
2. If we modify `ClientEncryptionImpl.createDataKey` such that it throws an exception the second time it's called, and modify the `insertEncryptedValue` such that it has two elements in `fields` with `keyId: null`, then not only `createEncryptedCollection` must emit `MongoUpdatedEncryptedFieldsException`, but also `MongoUpdatedEncryptedFieldsException.getEncryptedFields` must report the key ID of the key that was created successfully. Again, the proposed implementation does not do that. Moreover, the function passed into `flatMap` is never called in this scenario, i.e., `ctx` never comes into play.

   I made these modifications in the [JAVA-4679_experiments](https://github.com/stIncMale/mongo-java-driver/tree/JAVA-4679_experiments) branch (commit https://github.com/stIncMale/mongo-java-driver/commit/8ba7938ed18826056b39fafaba210967e4aa5c30) in case you want to try yourself.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
stIncMale: @rozza Thank you for pointing out the `onErrorMap` method, the emitted exceptions are nice now. There is no excuse for how I failed to find it myself.

PR #1081, opened 2023-02-07
Introduce `BsonUtil.mutableDeepCopy`

stIncMale:
Currently we have at least two places (that I know of) where we rely on `BsonDocument.clone` returning not only a deep copy, but mutable deep copy:

1. `AbstractConstructibleBson.newMerged`
2. The new method `ClientEncryption.createEncryptedCollection` introduced in https://github.com/mongodb/mongo-java-driver/pull/1079.

@rozza [pointed out](https://github.com/mongodb/mongo-java-driver/pull/1079#discussion_r1097664225) that `RawBsonDocument` breaks the "mutable" part and can come from users: the `cloneIsDeepCopyAndMutable` test added in this PR demonstrates this.

`RawBsonDocument.clone` returns `BsonDocument` (not `RawBsonDocument`), and documents nothing about its behavior (neither does `BsonDocument.clone`). So while there may be different approaches of solving the problem, these facts allow for the straightforward approach to fix the problem proposed in this PR.

JAVA-4874

### DIFF
--- a/bson/src/main/org/bson/internal/BsonUtil.java
+++ b/bson/src/main/org/bson/internal/BsonUtil.java
@@ -45,27 +45,19 @@ public final class BsonUtil {
         return new BsonJavaScriptWithScope(original.getCode(), mutableDeepCopy(original.getScope()));
     }
 
-    private static <T extends BsonValue> T mutableDeepCopy(final T original) {
-        BsonValue copy;
+    private static BsonValue mutableDeepCopy(final BsonValue original) {
         switch (original.getBsonType()) {
             case DOCUMENT:
-                copy = mutableDeepCopy(original.asDocument());
-                break;
+                return mutableDeepCopy(original.asDocument());
             case ARRAY:
-                copy = mutableDeepCopy(original.asArray());
-                break;
+                return mutableDeepCopy(original.asArray());
             case BINARY:
-                copy = mutableDeepCopy(original.asBinary());
-                break;
+                return mutableDeepCopy(original.asBinary());
             case JAVASCRIPT_WITH_SCOPE:
-                copy = mutableDeepCopy(original.asJavaScriptWithScope());
-                break;
+                return mutableDeepCopy(original.asJavaScriptWithScope());
             default:
-                copy = original;
+                return original;
         }
-        @SuppressWarnings("unchecked")
-        T result = (T) copy;
-        return result;
     }
 
     private BsonUtil() {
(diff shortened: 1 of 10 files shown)


bson/src/test/unit/org/bson/BsonDocumentTest.java
```diff
@@ -23,13 +23,16 @@
 import org.bson.json.JsonReader;
 import org.bson.json.JsonWriter;
 import org.bson.json.JsonWriterSettings;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
```
rozza, line 26: :)

rozza (approved):
LGTM - please also add a Jira ticket to associate with this change.

jyemin:
I'm not so sure about this, and would like to discuss it in the context of a Jira ticket.

stIncMale:
@jyemin, @rozza I created a corresponding Jira ticket: JAVA-4874.

stIncMale:
Instead of modifying `RawBsonDocument.clone`, I introduced `Util.mutableDeepCopy`.

rozza (commented):
Not sure I like the lack of description in name of the `Util` class - suggest `BsonDocumentHelper` or `BsonDocumentUtil`.  Will make it more discoverable in the future.

I like the tests use a mix of BsonValue (BsonDocumentWrapper, RawBsonDocument etc) types.

Other than the name - LGTM

stIncMale:
@rozza Renamed to `BsonUtil`.

bson/src/main/org/bson/internal/BsonUtil.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private static BsonBinary mutableDeepCopy(final BsonBinary original) {
+        return new BsonBinary(original.getType(), original.getData().clone());
+    }
+
+    private static BsonJavaScriptWithScope mutableDeepCopy(final BsonJavaScriptWithScope original) {
+        return new BsonJavaScriptWithScope(original.getCode(), mutableDeepCopy(original.getScope()));
+    }
+
+    private static <T extends BsonValue> T mutableDeepCopy(final T original) {
```
jyemin, line 48: Since we have to suppress an unchecked warning anyway, is it any better/safer to use `T extends BsonValue ` here rather than just `BsonValue`?  I guess it serves as documentation of the intention even if it's not enforced by the compiler?

stIncMale: This is bad. I wrote it this way to avoid casting everywhere I call it. But now that you asked, I checked, and everywhere the method is called (just two places), `BsonValue` is expected, so no, there is no reason for this method to be generic or to have any explicit conversions.

jyemin (approved):
LGTM

PR #1082, opened 2023-02-09
Fix the "database name is too long" error that happens when running against Serverless

stIncMale:
JAVA-4679


rozza (approved):

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionAutoDataKeysTest.java
```diff
@@ -88,7 +88,7 @@ public void setUp() {
                         provider -> provider.name, provider -> provider.propertiesSupplier)))
                 .build());
         client.getDatabase(KEY_VAULT_NAMESPACE.getDatabaseName()).drop();
-        db = client.getDatabase(AbstractClientSideEncryptionAutoDataKeysTest.class.getSimpleName());
+        db = client.getDatabase("test");
```
katcharov, line 91: Perhaps this should substring the first n characters of the test name instead? (I assume we use the original convention to match up databases with tests.)

katcharov (changes requested):

katcharov (approved):
lgtm

PR #1087, opened 2023-02-15
Fix documentation of ClientEncryptionSettings#keyVaultMongoClientSett…

jyemin:
…ings property to indicate it can't be null

* Update Javadoc
* Add notNull assertion in ClientEncryptionSettings constructor
* Remove unnecessary null checks and corresponding SpotBugs exclusions for them

JAVA-4861

### DIFF
--- a/config/spotbugs/exclude.xml
+++ b/config/spotbugs/exclude.xml
@@ -240,18 +240,4 @@
         <Method name="acquirePermitOrGetAvailableOpenedConnection"/>
         <Bug pattern="NS_NON_SHORT_CIRCUIT"/>
     </Match>
-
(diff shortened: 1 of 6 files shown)


jyemin:
Tests were failing.  One fix was obvious but the one to the unified tests was less so.  And I think the issue there was why the settings were made pseudo-nullable before this patch.  It's because the unified tests require that a MongoClient entity be used as the key vault client, which the Java driver does not allow.  So a private API was added to allow it and the settings were not required to be non-null to allow the tests to work.  I addressed that by just setting the property in the unified tests even though its value is ignored by the private API.

katcharov (approved):
lgtm

PR #1090, opened 2023-02-17
Initial driver-kotlin-sync implementation

rozza:
Wraps the driver-sync to provide a kotlin sync API

In general a faithful wrapping of the sync java driver. Uses reified overloads to provide syntactic sugar instead of passing resultClass.

MongoIterable deviates from the sync API so to not implement Iterable in order to protect the user from leaving open cursors. In Kotlin there are two Iterable style implementations: Iterable and Sequence neither are auto closable. Sequence is more appropriate here as it lazily consumes the iterator, but it does not auto close. See: https://youtrack.jetbrains.com/issue/KT-34719 So MongoIterable provides a loan method that treats the cursor as a Sequence and automatically closes after use.

JAVA-4869

Todo
----

 - [x] Confirm the build / test
 - [x] Publishing
 - [x] Add extensions to any Option classes that to default timeunit `maxTime(maxTime: Long, timeUnit: TimeUnit = TimeUnit.MILLISECONDS)`
 - [ ] Is `MongoIterable` the correct name?
 - [ ] Confirm package namespace - currently `com.mongodb.kotlin.client`


driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/AggregateIterableTest.kt
```diff
@@ -0,0 +1,105 @@
...
+        whenever(wrapped.explain(Document::class.java, verbosity)).doReturn(mock())
+        whenever(wrapped.explain(BsonDocument::class.java, verbosity)).doReturn(mock())
+
+        iterable.allowDiskUse(true)
+        iterable.batchSize(batchSize)
+        iterable.bypassDocumentValidation(true)
+        iterable.collation(collation)
+        iterable.comment(bsonComment)
+        iterable.comment(comment)
+        iterable.explain()
```
rozza, line 69: Test all the overloaded variants - default, passing the class and declaring the type

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoIterable.kt
```diff
@@ -0,0 +1,87 @@
...
+     * This allows the [MongoIterable] to be safely treated as a lazily evaluated sequence.
+     *
+     * Note: Sequence filters and aggregations have a performance cost, it is best to use server side filters and
+     * aggregations where available.
+     *
+     * @param R the result type
+     * @param consumer the sequence consumer
+     * @return the result of the consumer
+     */
+    public fun <R> use(consumer: (Sequence<T>) -> R): R = cursor().use { consumer.invoke(it.asSequence()) }
```
jyemin, line 65: I'm unclear why this is needed if you can call `cursor()` and then `use` the cursor.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoIterable.kt
```diff
@@ -0,0 +1,87 @@
...
+    public fun forEach(action: (T) -> Unit): Unit = use { it.forEach(action) }
+
+    /**
+     * Appends all elements to the given [destination] collection.
+     *
+     * @param C the type of the collection
+     * @param destination the destination collection
+     * @return the collection
+     */
+    public fun <C : MutableCollection<in T>> toCollection(destination: C): C = use { it.toCollection(destination) }
```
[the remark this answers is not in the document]
jyemin: Hmm, it's a bit confusing given `AggregateIterable.toCollection()`. Do you think that will flummox users?

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoClient.kt
```diff
@@ -0,0 +1,283 @@
...
+ *
+ * @see MongoClient.create
+ */
+public class MongoClient(@PublishedApi internal val wrapped: JMongoClient) : Closeable {
+
+    /**
+     * A factory for [MongoClient] instances.
+     *
+     * @see MongoClient
+     * @since 4.10
```
jyemin, line 48: Note: this is the only place I noticed this javadoc annotation

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoChangeStreamCursor.kt
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mongodb.kotlin.client
+
+import com.mongodb.client.MongoChangeStreamCursor as JMongoChangeStreamCursor
+import org.bson.BsonDocument
+
+/**
+ * The Mongo Cursor interface implementing the iterator protocol.
```
jyemin, line 22: Copy pasta?

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Explain the execution plan for this operation with the given verbosity level
+     *
+     * @param R the type of the document class
+     * @param verbosity the verbosity of the explanation
+     * @return the execution plan
+     * @see [Explain command](https://www.mongodb.com/docs/manual/reference/command/explain/)
+     */
+    public inline fun <reified R : Any> explain(verbosity: ExplainVerbosity? = null): R =
```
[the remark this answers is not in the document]
rozza: Due to: https://youtrack.jetbrains.com/issue/KT-11111

config/detekt/detekt.yml
```diff
@@ -0,0 +1,712 @@
+build:
```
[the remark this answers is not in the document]
rozza: Yeap - its what we used in Kbson

config/mongodb.license
```diff
@@ -0,0 +1,15 @@
+/*
```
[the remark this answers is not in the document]
rozza: Used as part of our kotlin spotless configuration - ensures the license is present.

driver-kotlin-sync/src/integration/kotlin/com/mongodb/kotlin/client/CrudTest.kt
```diff
@@ -0,0 +1,49 @@
...
+
+import com.mongodb.client.AbstractCrudTest
+import com.mongodb.client.Fixture
+import com.mongodb.client.MongoDatabase
+import com.mongodb.event.CommandListener
+import com.mongodb.kotlin.client.syncadapter.SyncMongoClient
+import org.bson.BsonArray
+import org.bson.BsonDocument
+
+data class CrudTest(
```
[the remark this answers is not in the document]
rozza: Will change - its not going to need the copy or equals methods.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
```diff
@@ -0,0 +1,192 @@
...
+import org.bson.Document
+import org.bson.conversions.Bson
+
+/**
+ * Iterable like implementation for aggregate operations.
+ *
+ * @param T The type of the result.
+ * @see [Aggregation command](https://www.mongodb.com/docs/manual/reference/command/aggregate)
+ */
+public class AggregateIterable<T : Any>(@PublishedApi internal val wrapped: JAggregateIterable<T>) :
```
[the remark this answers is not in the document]
rozza: Its just to reduce the amount of code surface and maintenance burden - same as in Scala.
I haven't seen any evidence of users mixing Java / Scala so far it has been a non issue. 

We could make the constructor internal eg:

```
public class AggregateIterable<T : Any> 
@PublishedApi internal constructor(@PublishedApi internal val wrapped: JAggregateIterable<T>) :
    MongoIterable<T>(wrapped) {
```

driver-kotlin-sync/src/integration/kotlin/com/mongodb/kotlin/client/syncadapter/SyncAggregateIterable.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.mongodb.ExplainVerbosity
+import com.mongodb.client.AggregateIterable as JAggregateIterable
+import com.mongodb.client.model.Collation
+import com.mongodb.kotlin.client.AggregateIterable
+import java.util.concurrent.TimeUnit
+import org.bson.BsonValue
+import org.bson.Document
+import org.bson.conversions.Bson
+
+data class SyncAggregateIterable<T : Any>(val wrapped: AggregateIterable<T>) :
```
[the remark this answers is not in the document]
rozza: Updated to internal classes

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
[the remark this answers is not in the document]
rozza: Or we could add interfaces and implementations

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
```diff
@@ -0,0 +1,192 @@
...
+    MongoIterable<T>(wrapped) {
+    /**
+     * Sets the number of documents to return per batch.
+     *
+     * @param batchSize the batch size
+     * @return this
+     * @see [Batch Size](https://www.mongodb.com/docs/manual/reference/method/cursor.batchSize/#cursor.batchSize)
+     */
+    public override fun batchSize(batchSize: Int): AggregateIterable<T> = apply { wrapped.batchSize(batchSize) }
```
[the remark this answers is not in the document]
rozza: Neat syntactic sugar: https://kotlinlang.org/docs/scope-functions.html#apply

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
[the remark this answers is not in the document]
rozza: Some more Kotlin features: 

> The IDE treats it as internal, but you can indirectly expose the code via inline functions.
> 
> The primary purpose is to allow public inline functions to call it without exposing it to being called directly from outside the module. I

See: https://kotlinlang.org/docs/inline-functions.html#restrictions-for-public-api-inline-functions

That said the wrapped value can actually be private in most instances and aren't used inline. So have updated.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/ClientSession.kt
```diff
@@ -0,0 +1,230 @@
...
+     *
+     * @return the transaction options
+     */
+    public fun getTransactionOptions(): TransactionOptions = wrapped.transactionOptions
+
+    /**
+     * Start a transaction in the context of this session with default transaction options. A transaction can not be
+     * started if there is already an active transaction on this session.
+     */
+    public fun startTransaction(): Unit = wrapped.startTransaction()
```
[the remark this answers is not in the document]
rozza: Because `com.mongodb.session.ClientSession` doesn't implement `startTransaction`

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/ClientSession.kt
```diff
@@ -0,0 +1,230 @@
...
+import com.mongodb.TransactionOptions
+import com.mongodb.client.ClientSession as clientClientSession
+import com.mongodb.session.ClientSession as jClientSession
+import com.mongodb.session.ServerSession
+import java.util.concurrent.TimeUnit
+import org.bson.BsonDocument
+import org.bson.BsonTimestamp
+
+/** A client session that supports transactions. */
+public class ClientSession(public val wrapped: clientClientSession) : jClientSession {
```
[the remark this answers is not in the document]
rozza: Yeap no need. Will update and can remove all the internal API methods.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoIterable.kt
rozza: Its a helper method that makes it easy for a user to consume the cursor as a Sequence and provides access to all the sequence functions.


```
iterable.use { it.take(2) } // Valid code

iterable.cursor().use { it.take(2) } // invalid doesn't compile

iterable.cursor().use { it.asSequence().take(2) } // Valid code but verbose / non obvious


```

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoClient.kt
rozza: Thanks - missed a few

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoChangeStreamCursor.kt
rozza: Yeap :)

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/ClientSession.kt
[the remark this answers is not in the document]
rozza: No longer implement an interface for this class.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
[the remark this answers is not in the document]
rozza: All now private or internal.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoIterable.kt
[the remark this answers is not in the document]
rozza: I think its fine thanks to the difference in type signatures.

.evergreen/.evg.yml
```diff
@@ -2052,6 +2069,13 @@ buildvariants:
   tasks:
     - name: "scala-tests"
 
+- matrix_name: "kotlin-tests"
+  matrix_spec: { auth: "noauth", ssl: "nossl", jdk: ["jdk8", "jdk17"], version: ["4.4", "5.0", "6.0", "latest"], topology: "replicaset", os: "ubuntu" }
```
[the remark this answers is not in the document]
rozza: Done

.evergreen/.evg.yml
```diff
@@ -2052,6 +2069,13 @@ buildvariants:
...
+  display_name: "${jdk} ${version} ${topology} ${os}"
```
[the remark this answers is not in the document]
rozza: Done

.evergreen/.evg.yml
```diff
@@ -2052,6 +2069,13 @@ buildvariants:
   tasks:
     - name: "scala-tests"
 
+- matrix_name: "kotlin-tests"
```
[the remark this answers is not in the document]
rozza: I think this will come out in JAVA-4895

driver-kotlin-sync/src/integration/kotlin/com/mongodb/kotlin/client/CrudTest.kt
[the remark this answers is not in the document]
rozza: Done - well removed in favor of UnifiedCrudTest

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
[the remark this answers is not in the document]
rozza: Haven't actioned this.

We could:

 A: Keep as is
 B: Private Constructors - may have issues with sync wrappers
 C: Split to interfaces and implementations

driver-kotlin-sync/src/integration/kotlin/com/mongodb/kotlin/client/UnifiedCrudTest.kt
```diff
@@ -0,0 +1,52 @@
...
+package com.mongodb.kotlin.client
+
+import java.io.IOException
+import java.net.URISyntaxException
+import org.bson.BsonArray
+import org.bson.BsonDocument
+import org.junit.Assume.assumeFalse
+import org.junit.runners.Parameterized
+
+internal class UnifiedCrudTest(
```
jyemin, line 25: I didn't mean that `CrudTest` should be deleted.  Just that we should also add this.  Can you re-add `CrudTest`?

jyemin (changes requested):

driver-kotlin-sync/src/integration/kotlin/com/mongodb/kotlin/client/UnifiedCrudTest.kt
rozza: :facepalm: whoops - its back :)

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoClientTest.kt
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Test
+    fun shouldCallTheUnderlyingClusterDescription() {
+        val mongoClient = MongoClient(wrapped)
+        whenever(wrapped.clusterDescription).doReturn(mock())
+
+        mongoClient.getClusterDescription()
+
+        verify(wrapped).clusterDescription
```
katcharov, line 64: These wrap tests take a different approach (one per method, individual mocks, etc.) compared to some of the others (where many methods are tested in a row, even lower down in this class). Why the difference?

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoClientTest.kt
```diff
@@ -0,0 +1,165 @@
...
+
+class MongoClientTest {
+
+    @Mock val wrapped: JMongoClient = mock()
+    @Mock val clientSession: ClientSession = ClientSession(mock())
+
+    @Test
+    fun shouldHaveTheSameMethods() {
+        val jMongoClientFunctions = JMongoClient::class.declaredFunctions.map { it.name }.toSet()
+        val kMongoClientFunctions = MongoClient::class.declaredFunctions.map { it.name }.toSet()
```
katcharov, line 43: Why not do the same `.filterNot { it.name == "wrapped" }` etc. check here (and elsewhere), as I've seen in some other `shouldHaveTheSameMethods` methods?

katcharov (commented):
I looked over the wrappers, looked good, and nice straightforward handling of the default parameters. A few minor comments related to the tests.

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/ListDatabasesIterableTest.kt
```diff
@@ -0,0 +1,69 @@
...
+        iterable.maxTime(1, TimeUnit.SECONDS)
+        iterable.nameOnly(true)
+
+        verify(wrapped).authorizedDatabasesOnly(true)
+        verify(wrapped).batchSize(batchSize)
+        verify(wrapped).comment(bsonComment)
+        verify(wrapped).comment(comment)
+        verify(wrapped).filter(filter)
+        verify(wrapped).maxTime(1, TimeUnit.MILLISECONDS)
+        verify(wrapped).maxTime(1, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
rozza: Interesting, I suppose it comes down to readability / maintainability. Code coverage ensures and the API check combined are used to ensure coverage. The approach of mocking is very explicit and follows similar patterns used in the Scala driver.

For wrappers some API methods should differ, where it is idiomatic to do so. For example supporting default values and if possible not requiring `resultClass` (as it is in Scala but not Kotlin).

We still need a mock as we don't have a real wrapped entity to check against and sometimes that requires extra logic in terms of returned values. I don't know how well reflection and mocking would work together and how debuggable the combination would be.

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoChangeStreamCursorTest.kt
```diff
@@ -0,0 +1,72 @@
...
+    @Test
+    fun shouldHaveTheSameMethods() {
+        val jMongoChangeStreamCursorFunctions =
+            JMongoChangeStreamCursor::class.declaredFunctions.map { it.name }.toSet()
+        val kMongoChangeStreamCursorFunctions =
+            MongoChangeStreamCursor::class.declaredFunctions.map { it.name }.toSet() +
+                MongoChangeStreamCursor::class
+                    .declaredMemberProperties
+                    .filterNot { it.name == "wrapped" }
+                    .map { "get${it.name.replaceFirstChar{c -> c.uppercaseChar() }}" }
```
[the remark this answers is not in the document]
rozza: Good question. In Kotlin classes can have both properties and functions.

In certain places I used properties for any getters that are immutable and don't change state.

Officially the [coding convention](https://kotlinlang.org/docs/coding-conventions.html#functions-vs-properties) states:

> **Functions vs properties**﻿
> In some cases functions with no arguments might be interchangeable with read-only properties. Although the semantics are similar, there are some stylistic conventions on when to prefer one to another.
>
> Prefer a property over a function when the underlying algorithm:
> 
> - does not throw
> - is cheap to calculate (or cached on the first run)
> - returns the same result over invocations if the object state hasn't changed

So for `MongoChangeStreamCursor` I used a property for `ServerAddress` but functions for `getResumeToken` as that changes state.

I could use properties for more functions that meet the above criteria but I think functions signalling mutable property values and properties signalling immutable values is the right decision.  As this is a wrapper it makes no sense to add a setter for a property (which would declare it mutable) as that is all internal to the wrapper.

@cmelchior - as my goto Kotlin expert - is my interpretation of the use of properties idiomatic? Or Should I be using properties more?

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoClientTest.kt
rozza: Due to the lack of properties in the kotlin class.

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoClientTest.kt
rozza: So we seem to have a couple of conventions:

1) A test for each method. Which applies to the MongoClient, Database and Collection. Where there are lots of methods and overloads for each method.
2) Usage based tests case are more brief and tied to usage. Which includes chainable API's like MongoIterable and then an outlier with MongoCursor.

This follows on from the Scala convention, which probably followed on from the convention established when the reactive stream driver wrapped the async driver. The original reasons for such an approach is lost.

driver-kotlin-sync/build.gradle.kts
```diff
@@ -0,0 +1,185 @@
...
+    // Align versions of all Kotlin components
+    implementation(platform("org.jetbrains.kotlin:kotlin-bom"))
+    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
+
+    api(project(path = ":bson", configuration = "default"))
+    api(project(path = ":driver-sync", configuration = "default"))
+
+    testImplementation("org.jetbrains.kotlin:kotlin-reflect")
+    testImplementation("org.jetbrains.kotlin:kotlin-test-junit")
+    testImplementation("org.mockito.kotlin:mockito-kotlin:4.1.0")
```
zigzago, line 67: usually mockk is a better lib for kotlin than mockito-kotlin

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoClient.kt
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * Gets the list of databases
+     *
+     * @param T the type of the class to use
+     * @param clientSession the client session with which to associate this operation
+     * @param resultClass the target document type of the iterable.
+     * @return the list databases iterable interface
+     */
+    public fun <T : Any> listDatabases(clientSession: ClientSession, resultClass: Class<T>): ListDatabasesIterable<T> =
```
zigzago, line 187: you can use the inline reified trick here (with @JvmName)

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoClient.kt
```diff
@@ -0,0 +1,282 @@
...
+        /**
+         * Create a new client with the given connection string.
+         *
+         * @param connectionString the connection string, defaults to `mongodb://localhost`.
+         * @param mongoDriverInformation any driver information to associate with the MongoClient
+         * @return the client
+         */
+        public fun create(
+            connectionString: ConnectionString = ConnectionString("mongodb://localhost"),
+            @Nullable mongoDriverInformation: MongoDriverInformation? = null
```
zigzago, line 67: `@Nullable` cloud be removed

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoIterable.kt
```diff
@@ -0,0 +1,87 @@
...
```
zigzago, line 86: It would be great to add a `<T> toList(): List<T> method` - this is what devs use day to day

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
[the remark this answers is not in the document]
zigzago: mixin java/kotlin code project is a common use case (progressive migration). You can use internal constructors and add a factory method to use the same connection pool

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
```diff
@@ -0,0 +1,191 @@
...
+    public fun toCollection(): Unit = wrapped.toCollection()
+
+    /**
+     * Enables writing to temporary files. A null value indicates that it's unspecified.
+     *
+     * @param allowDiskUse true if writing to temporary files is enabled
+     * @return this
+     * @see [Aggregation command](https://www.mongodb.com/docs/manual/reference/command/aggregate/)
+     */
+    public fun allowDiskUse(allowDiskUse: Boolean?): AggregateIterable<T> = apply { wrapped.allowDiskUse(allowDiskUse) }
```
zigzago, line 58: why using a nullable Boolean ? (same question for the other methods)

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoCollection.kt
```diff
@@ -0,0 +1,1496 @@
...
+     * @param mapFunction A JavaScript function that associates or "maps" a value with a key and emits the key and value
+     *   pair.
+     * @param reduceFunction A JavaScript function that "reduces" to a single object all the values associated with a
+     *   particular key.
+     * @return an iterable containing the result of the map-reduce operation
+     * @see [map-reduce](https://www.mongodb.com/docs/manual/reference/command/mapReduce/)
+     */
+    @Suppress("DEPRECATION")
+    @Deprecated("Map Reduce has been deprecated. Use Aggregation instead", replaceWith = ReplaceWith(""))
+    public inline fun <reified R : Any> mapReduce(mapFunction: String, reduceFunction: String): MapReduceIterable<R> =
```
zigzago, line 592: Do you really need to add theses deprecated methods ? :)

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
```diff
@@ -0,0 +1,191 @@
...
+ */
+public class AggregateIterable<T : Any>(private val wrapped: JAggregateIterable<T>) : MongoIterable<T>(wrapped) {
+    /**
+     * Sets the number of documents to return per batch.
+     *
+     * @param batchSize the batch size
+     * @return this
+     * @see [Batch Size](https://www.mongodb.com/docs/manual/reference/method/cursor.batchSize/#cursor.batchSize)
+     */
+    public override fun batchSize(batchSize: Int): AggregateIterable<T> = apply { wrapped.batchSize(batchSize) }
```
zigzago, line 40: You imply that wrapped.batchSize() returns wrapped. As JAggregateIterable is an interface, this is not guaranteed. (except if the javadoc mention is a contract)

driver-kotlin-sync/build.gradle.kts
rozza: It does seem nicer but we use Mockito else where, so preferred one mocking library.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoClient.kt
rozza: Good catch

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoIterable.kt
rozza: Added

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
rozza: This is so users can unset previously set values.  Not all values have defaults and this really means use the servers default for the setting.  As these can change overtime its best not to set a default in the driver as it makes the setting sticky and hard to change.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoCollection.kt
rozza: I'd happily not add them :) @jyemin any reasons to include them?

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
rozza: This is the contract - implementations have to return themselves (this) for these methods.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoCollection.kt
jyemin: I think we can kill them.

rozza: Done

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoClient.kt
rozza: Ack - will try to work around it. The main issue is inlined methods cannot be called from Java code.  However, we should still be able test the resulting MongoIterables just we'd have to construct them manually.

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoClientTest.kt
katcharov: Alright

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/ListDatabasesIterableTest.kt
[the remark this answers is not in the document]
katcharov: Fair enough

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoChangeStreamCursorTest.kt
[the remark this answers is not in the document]
katcharov: (Consider resolved from my side, pending @cmelchior response.)

katcharov (approved):
LGTM

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
zigzago: ok

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
zigzago: ok

zigzago:
For some reason I can not resolve my own comments, but the PR is ok for me

zigzago (approved):

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoChangeStreamCursorTest.kt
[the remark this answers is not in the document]
cmelchior: I would partly agree and partly disagree. We are using the same reference in their coding convention but probably lean slightly more towards exposing them as properties since that is easier to consume when writing the code and relying on autocomplete.

Their heuristics are also a bit elastic, i.e. what is "cheap" and what determines "object state".

I think it is a mistake to treat properties as constants though, both `val` and `var` properties are allowed to change.

In this particular case, I would probably lean towards turning `getResumeToken` into a `resumeToken` property.

Some examples from the Realm Kotlin SDK:

- [`User`](https://github.com/realm/realm-kotlin/blob/main/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt#L38) .. most things are properties, even things that "change".
- [`BaseRealm`](https://github.com/realm/realm-kotlin/blob/main/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/BaseRealm.kt#L33) ... we had a lengthy discussion about `schema()` but it ended up a function because it is an expensive data structure to build and return.

API design is tricky :D

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoClient.kt
rozza: @jyemin - turns out removing the resultClass methods has a number of consequences.

- The wrapped MongoClient / MongoDatabase / MongoCollections then have to become publicly visible. (There is no package private in Kotlin) and reified functions require public access. (https://kotlinlang.org/docs/inline-functions.html#restrictions-for-public-api-inline-functions)
- The syncadapter cannot call the reified methods - so essentially you have to manually construct the MongoIterables by calling the wrapped methods directly. So doesn't actually  test the full API.
- The API becomes hard to wrap by 3rd parties. Reified types in methods are inlined at compile time, so like the sync adapter users with a Class<T> cannot access the reified method. Having an overload is the way around that.

Only suggestion I have it we could half the overloaded API size by making clientSession optional for all the methods that take a resultClass. eg: 

```
public fun <T : Any> listDatabases(clientSession: ClientSession?, resultClass: Class<T>): ListDatabasesIterable<T>
```

cmelchior: Note, you can use `@PublishedApi` to allow reified methods calling internal methods: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-published-api/

driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoChangeStreamCursorTest.kt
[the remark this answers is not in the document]
rozza: Sounds good - updated a few functions to properties

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/AggregateIterable.kt
```diff
@@ -0,0 +1,191 @@
...
+import org.bson.Document
+import org.bson.conversions.Bson
+
+/**
+ * Iterable like implementation for aggregate operations.
+ *
+ * @param T The type of the result.
+ * @see [Aggregation command](https://www.mongodb.com/docs/manual/reference/command/aggregate)
+ */
+public class AggregateIterable<T : Any>(private val wrapped: JAggregateIterable<T>) : MongoIterable<T>(wrapped) {
```
jyemin, line 32: Should we add `@since` annotations to all public types?  Is that a thing in Kotlin APIs?

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoCursor.kt
```diff
@@ -0,0 +1,73 @@
...
+ *      while (c.hasNext()) {
+ *          println(c.next())
+ *      }
+ *  }
+ * ```
+ *
+ * @param T The type of documents the cursor contains
+ * @property wrapped the underlying sync cursor
+ */
+public open class MongoCursor<T : Any>(private val wrapped: JMongoCursor<T>) : Iterator<T>, Closeable {
```
jyemin, line 38: Why is this class open for extension? Is it because of `MongoChangeStreamCursor`?  If so, should we make this class `sealed` instead?

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoClient.kt
jyemin: Can we add a resolution to this thread, please?

jyemin (changes requested):

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoClient.kt
rozza: The initial comment is resolved (using inline trick)

As @cmelchior mentions you can use `@PublishedApi` annotation when using reified methods and internal values. Which can be used for my first point (As long as the syncadapter is moved into the same package namespace).

However, the removal of resultClass has wider implications as mentioned in my thread - so should remain - so that users can call them if they ever need to at runtime.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoCursor.kt
rozza: sealed classes cannot be initiated but will add a base cursor type.

rozza: Done!

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/MongoCursor.kt
```diff
@@ -61,13 +55,81 @@ public open class MongoCursor<T : Any>(private val wrapped: JMongoCursor<T>) : I
...
+ *   collection.watch().cursor().use { c ->
+ *      while (c.hasNext()) {
+ *          println(c.next())
+ *      }
+ *  }
+ * ```
+ *
+ * @param T The type of documents the cursor contains
+ */
+public interface MongoChangeStreamCursor<T : Any> : MongoCursor<T> {
```
jyemin, line 81: Should this also be sealed?

rozza: It extends a sealed interface - so implicitly is

rozza: Have made it explicit just incase.

rozza:
Will manually rebase and squash

PR #1094, opened 2023-03-05
Work around create failure in unified tests with retry

jyemin:
On 4.0 sharded clusters, the create collection command often fails (on the same test) with error code 11601 (Interrupted).  Work around it by retrying the command.

JAVA-4883

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/AbstractUnifiedTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/AbstractUnifiedTest.java
@@ -71,6 +71,7 @@ import static com.mongodb.ClusterFixture.getMultiMongosConnectionString;
 import static com.mongodb.ClusterFixture.isDataLakeTest;
 import static com.mongodb.ClusterFixture.isSharded;
 import static com.mongodb.ClusterFixture.serverVersionAtLeast;
+import static com.mongodb.ClusterFixture.serverVersionLessThan;
 import static com.mongodb.ClusterFixture.setDirectConnection;
 import static com.mongodb.client.CommandMonitoringTestHelper.assertEventsEquality;
 import static com.mongodb.client.CommandMonitoringTestHelper.getExpectedEvents;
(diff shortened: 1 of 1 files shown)


driver-sync/src/test/functional/com/mongodb/client/AbstractUnifiedTest.java
```diff
@@ -168,7 +169,16 @@ public void setUp() {
         collectionHelper.killAllSessions();
 
         if (!isDataLakeTest()) {
-            collectionHelper.create(collectionName, new CreateCollectionOptions(), WriteConcern.MAJORITY);
+            try {
+                collectionHelper.create(collectionName, new CreateCollectionOptions(), WriteConcern.MAJORITY);
+            } catch (MongoCommandException e) {
+                // Older sharded clusters sometimes reply with this error.  Work around it by retrying once.
+                if (e.getErrorCode() == 11601 && isSharded() && serverVersionLessThan(4, 2)) {
```
jyemin, line 176: I decided to just do the retry in the unified runner because it's a unified test that is failing.  But could also push the retry down into the helper method, which is used by other runners.

driver-sync/src/test/functional/com/mongodb/client/AbstractUnifiedTest.java
```diff
@@ -168,7 +169,16 @@ public void setUp() {
         collectionHelper.killAllSessions();
 
         if (!isDataLakeTest()) {
-            collectionHelper.create(collectionName, new CreateCollectionOptions(), WriteConcern.MAJORITY);
+            try {
+                collectionHelper.create(collectionName, new CreateCollectionOptions(), WriteConcern.MAJORITY);
+            } catch (MongoCommandException e) {
+                // Older sharded clusters sometimes reply with this error.  Work around it by retrying once.
```
jyemin, line 175: I decided not to report the bug because 4.0 is nearly out of support.

driver-sync/src/test/functional/com/mongodb/client/AbstractUnifiedTest.java
katcharov: If the bug is making other runners flaky, it seems the fix should be applied widely if possible.

I did not understand how to interpret the frequency of this bug in the chart - is retrying once enough to never practically see this bug again, or should this be a for loop?

katcharov: If the bug is making other runners flaky, it seems the fix should be applied widely if possible.

I did not understand how to interpret the frequency of this bug in the chart - is retrying once enough to never practically see this bug again, or should this be a for loop?

katcharov (commented):
Just the one comment

driver-sync/src/test/functional/com/mongodb/client/AbstractUnifiedTest.java
jyemin: For reasons I do not entirely understand, it's one specific unified test that is failing, and retrying once seems to do the trick in the multiple runs I've executed.  I don't want to retry until success as that could potentially cause the whole test suite to time out.

katcharov: I see, that makes sense

katcharov (approved):
LGTM!

PR #1096, opened 2023-03-10
fix: ensure no NPE is thrown with SerializedFluxSink

imclem:
Hi there 👋,

We use Mongodb reactive stream driver with akka-stream and had an issue when trying to upgrade our mongodb driver from 4.1.x to 4.9.0. The same code which was working with 4.1.x does not work anymore with 4.9.0. 
We tracked downed the issue and it's happening since the addition of reactor to this project. 

The issue is that when performing a find/distinct which returns `null` values, when the results are used in a `SerializedFluxSink` a NullPointerException is thrown and everything stops.
This is because `SerializedFluxSink.next` performs a call to `Object.requireNonNull()`.

To fix this NPE we filter out null values from the results.

I'm not sure this is the appropriate way to test/fix, suggestions are welcome.
In the meantime we workaround this on our side by filtering out null values in the query.

Thanks

### DIFF
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
@@ -24,6 +24,7 @@ import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
 import reactor.util.context.Context;
 
+import java.util.Objects;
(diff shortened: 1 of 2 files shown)


rozza:
Hi @imclem,

Apologies for the delay in response - I have opened a ticket for this and its scheduled for review: JAVA-4908

Ross

imclem:
Thanks

rozza (approved):
LGTM

PR #1097, opened 2023-03-16
Add OperationContext to sync path

jyemin:
This is really phase one: adding OperationContext and pushing it down to all the places it will ultimately be needed.

Phase two will be actually adding it to existing events and log messages according to the specifications.

JAVA-4907

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/binding/StaticBindingContext.java
+++ b/driver-core/src/main/com/mongodb/internal/binding/StaticBindingContext.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.internal.binding;
+
+import com.mongodb.RequestContext;
+import com.mongodb.ServerApi;
+import com.mongodb.internal.connection.OperationContext;
+import com.mongodb.internal.session.SessionContext;
+import com.mongodb.lang.Nullable;
+
+/**
+ *
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public class StaticBindingContext implements BindingContext {
+    private final SessionContext sessionContext;
+    private final ServerApi serverApi;
+    private final RequestContext requestContext;
+    private final OperationContext operationContext;
+
+    public StaticBindingContext(final SessionContext sessionContext, @Nullable final ServerApi serverApi,
+            final RequestContext requestContext, final OperationContext operationContext) {
+        this.sessionContext = sessionContext;
+        this.serverApi = serverApi;
+        this.requestContext = requestContext;
+        this.operationContext = operationContext;
+    }
+
+    @Override
+    public SessionContext getSessionContext() {
+        return sessionContext;
+    }
+
+    @Nullable
+    @Override
+    public ServerApi getServerApi() {
+        return serverApi;
+    }
+
+    @Override
+    public RequestContext getRequestContext() {
+        return requestContext;
+    }
+
+    @Override
+    public OperationContext getOperationContext() {
+        return operationContext;
+    }
+}
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -13,25 +13,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+package com.mongodb.internal.connection;
 
-package com.mongodb.internal.binding;
-
-import com.mongodb.internal.async.SingleResultCallback;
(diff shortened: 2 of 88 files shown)


driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mongodb.internal.connection;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+public class OperationContext {
```
jyemin, line 20: It was simplest just to make this a class with the id generator built in.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -0,0 +1,31 @@
...
+    private static final AtomicLong NEXT_ID = new AtomicLong(0);
+    private final long id;
+
+    public OperationContext() {
+        id = NEXT_ID.incrementAndGet();
+    }
+
+    public long getId() {
+        return id;
```
katcharov, line 29: Are these ever created but not used (for example, optimistically), and if so, is it worth making this lazy to give the user cleaner numbering?

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class OperationContext {
+    private static final AtomicLong NEXT_ID = new AtomicLong(0);
+    private final long id;
+
+    public OperationContext() {
+        id = NEXT_ID.incrementAndGet();
+    }
+
+    public long getId() {
```
katcharov, line 28: We never call this ourselves?

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package com.mongodb.internal.connection;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+public class OperationContext {
+    private static final AtomicLong NEXT_ID = new AtomicLong(0);
+    private final long id;
+
+    public OperationContext() {
```
katcharov, line 24: It feels a bit unsettling just how many classes this gets threaded into, though I have no other suggestion.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
katcharov: From where would the id information be used or logged by a user?

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
jyemin: I can't think of a case where that would happen.  At a minimum we're going to do server selection and log a message containing this id.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
jyemin: Not yet.  That's coming next.  It will eventually be used in most log statements and events

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
jyemin: My first attempt at this was to make this a super-interface of SessionContext, but the refactoring required to do that got a bit out of control, so I backed off that.  But the motivation for that was exactly as you say, since SessionContext is already threaded into many of the places this is needed.

katcharov: For the SessionContext idea, the plan would have been to add `getOperationContext` to it, and store it in various implementers? And the issue was too many changes to implementers of SessionContext?

jyemin: I was going to make SessionContext extend OperationContext but there wouldn't have been any implementers of OperationContext only.  But the current split between ClusterBinding and ClientSessionBinding made that difficult.  I went down the road of combining those two classes but ran into some complications that I didn't feel were justified for this effort.

katcharov (approved):
LGTM to proceed with the approach

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -258,7 +259,8 @@ static <T> T executeCommand(final WriteBinding binding, final String database, f
         notNull("binding", binding);
         return transformer.apply(assertNotNull(
                 connection.command(database, command, new NoOpFieldNameValidator(), primary(), new BsonDocumentCodec(),
-                        binding.getSessionContext(), binding.getServerApi(), binding.getRequestContext())), connection);
+                        binding.getSessionContext(), binding.getServerApi(), binding.getRequestContext(), binding.getOperationContext())),
```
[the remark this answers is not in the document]
jyemin: I will give it a shot and see what it looks like.

driver-core/src/main/com/mongodb/internal/binding/StaticBindingContext.java
```diff
@@ -0,0 +1,63 @@
...
+import com.mongodb.ServerApi;
+import com.mongodb.internal.connection.OperationContext;
+import com.mongodb.internal.session.SessionContext;
+import com.mongodb.lang.Nullable;
+
+/**
+ *
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public class StaticBindingContext implements BindingContext {
```
jyemin, line 29: I'm open to a better name for this class, but it's internal so I didn't think about it too much

rozza (approved):
LGTM

PR #1098, opened 2023-03-20
Proposal for per-MongoClient DNS lookup configuration

jyemin:
* Add InetAddressResolver and InetAddressResolverProvider interfaces
* Add configuration of InetAddressResolver and DnsClient in MongoClientSettings and ConnectionString

### DIFF
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -23,6 +23,7 @@ import com.mongodb.internal.diagnostics.logging.Logger;
 import com.mongodb.internal.diagnostics.logging.Loggers;
 import com.mongodb.internal.dns.DefaultDnsResolver;
 import com.mongodb.lang.Nullable;
+import com.mongodb.spi.dns.DnsClient;
 import org.bson.UuidRepresentation;
 
 import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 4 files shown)


driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -296,6 +297,21 @@ public class ConnectionString {
      * @since 3.0
      */
     public ConnectionString(final String connectionString) {
+        this(connectionString, null);
+    }
+
+    /**
```
jyemin, line 303: This is unfortunate but I don't see a way around it.

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolverProvider.java
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.spi.dns;
+
+/**
```
jyemin, line 19: This won't be used but makes it consistent with DnsClient

PR #1101, opened 2023-03-28
Add operationId to command events

jyemin:
JAVA-4907


driver-core/src/main/com/mongodb/internal/connection/SendMessageCallback.java
jyemin, line 1: Removed this unused file rather than fix it to use the new API.

driver-core/src/main/com/mongodb/event/CommandEvent.java
```diff
@@ -39,13 +59,12 @@ public abstract class CommandEvent {
      * @param connectionDescription the connection description
      * @param commandName the command name
      * @since 4.4
+     * @deprecated Prefer {@link CommandEvent#CommandEvent(RequestContext, long, int, ConnectionDescription, String)}
      */
+    @Deprecated
```
jyemin, line 64: Kind of annoying that we have to preserve BC on constructors like this but it's what we do.

driver-core/src/main/com/mongodb/event/CommandSucceededEvent.java
```diff
@@ -36,31 +36,54 @@ public final class CommandSucceededEvent extends CommandEvent {
...
+     * @param requestId the request id
+     * @param connectionDescription the connection description
+     * @param commandName the command name
+     * @param response the command response
+     * @param elapsedTimeNanos the non-negative elapsed time in nanoseconds for the operation to complete
+     * @since 4.4
+     * @deprecated Prefer
+     * {@link CommandSucceededEvent#CommandSucceededEvent(RequestContext, long, int, ConnectionDescription, String, BsonDocument, long)}
+     */
+    @Deprecated
```
katcharov, line 70: Why deprecate these, rather than just leave them as helpers? In what way is the `requestContext` optional in a way that `operationId` is not?

driver-core/src/main/com/mongodb/event/CommandSucceededEvent.java
```diff
@@ -36,31 +36,54 @@ public final class CommandSucceededEvent extends CommandEvent {
...
      * @param requestId the request id
      * @param connectionDescription the connection description
      * @param commandName the command name
      * @param response the command response
      * @param elapsedTimeNanos the non-negative elapsed time in nanoseconds for the operation to complete
-     * @since 4.4
+     * @since 4.10
      */
-    public CommandSucceededEvent(@Nullable final RequestContext requestContext, final int requestId,
+    public CommandSucceededEvent(@Nullable final RequestContext requestContext, final long operationId, final int requestId,
```
katcharov, line 49: In what circumstances would end-users need to use these constructors?

katcharov (approved):
LGTM. Just a couple of questions.

driver-core/src/main/com/mongodb/event/CommandSucceededEvent.java
jyemin: Generally I would only expect an end user to use these only for testing.  But we haven't distinguished them as _not_ a part of the API, so that means they are part of the API as public members of public classes

driver-core/src/main/com/mongodb/event/CommandSucceededEvent.java
jyemin: So we can minimize API surface over time by removing these in the next major release.

The `requestContext` is nullable because it is optionally configured via `MongoClientSettings#getContextProvider()`.  The `operationId` is really required now, so we don't want to keep around constructors that don't provide a value for it.

PR #1102, opened 2023-04-03
Refactor tests, credential cache

katcharov:
Refactoring for:
JAVA-4757

### DIFF
--- a/bson/src/test/unit/util/ThreadTestHelpers.java
+++ b/bson/src/test/unit/util/ThreadTestHelpers.java
@@ -33,7 +33,7 @@ public final class ThreadTestHelpers {
     public static void executeAll(final int nThreads, final Runnable c) {
         ExecutorService service = null;
         try {
-            service = Executors.newFixedThreadPool(10);
+            service = Executors.newFixedThreadPool(nThreads);
             CountDownLatch latch = new CountDownLatch(nThreads);
             List<Throwable> failures = Collections.synchronizedList(new ArrayList<>());
             for (int i = 0; i < nThreads; i++) {
@@ -50,6 +50,7 @@ public final class ThreadTestHelpers {
             try {
                 latch.await();
             } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
                 throw new RuntimeException(e);
             }
             if (!failures.isEmpty()) {
--- a/bson/src/test/unit/util/ThreadTestHelpers.java
+++ b/bson/src/test/unit/util/ThreadTestHelpers.java
@@ -19,6 +19,7 @@ package util;
 import org.opentest4j.MultipleFailuresError;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
@@ -34,7 +35,7 @@ public final class ThreadTestHelpers {
         try {
             service = Executors.newFixedThreadPool(10);
             CountDownLatch latch = new CountDownLatch(nThreads);
-            List<Throwable> failures = new ArrayList<>();
+            List<Throwable> failures = Collections.synchronizedList(new ArrayList<>());
             for (int i = 0; i < nThreads; i++) {
                 service.submit(() -> {
                     try {
@@ -62,5 +63,4 @@ public final class ThreadTestHelpers {
             }
         }
     }
-
 }
--- a/bson/src/test/unit/util/ThreadTestHelpers.java
+++ b/bson/src/test/unit/util/ThreadTestHelpers.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package util;
+
+import org.opentest4j.MultipleFailuresError;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+public final class ThreadTestHelpers {
+
+    private ThreadTestHelpers() {
+    }
+
+    public static void executeAll(final int nThreads, final Runnable c) {
+        ExecutorService service = null;
+        try {
+            service = Executors.newFixedThreadPool(10);
+            CountDownLatch latch = new CountDownLatch(nThreads);
+            List<Throwable> failures = new ArrayList<>();
+            for (int i = 0; i < nThreads; i++) {
+                service.submit(() -> {
+                    try {
+                        c.run();
+                    } catch (Throwable e) {
+                        failures.add(e);
+                    } finally {
+                        latch.countDown();
+                    }
+                });
+            }
+            try {
+                latch.await();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+            if (!failures.isEmpty()) {
+                MultipleFailuresError multipleFailuresError = new MultipleFailuresError("Failed to execute all", failures);
+                failures.forEach(multipleFailuresError::addSuppressed);
+                throw multipleFailuresError;
+            }
+        } finally {
+            if (service != null) {
+                service.shutdown();
+            }
+        }
+    }
+
+}
--- a/driver-core/src/main/com/mongodb/MongoCredential.java
+++ b/driver-core/src/main/com/mongodb/MongoCredential.java
@@ -362,14 +362,22 @@ public final class MongoCredential {
      * @param source    the source of the user name, typically a database name
      * @param password  the password
      */
-    MongoCredential(@Nullable final AuthenticationMechanism mechanism, @Nullable final String userName, final String source,
-                    @Nullable final char[] password) {
+    MongoCredential(
+            @Nullable final AuthenticationMechanism mechanism,
+            @Nullable final String userName,
+            final String source,
+            @Nullable final char[] password) {
         this(mechanism, userName, source, password, Collections.emptyMap());
     }
 
-    MongoCredential(@Nullable final AuthenticationMechanism mechanism, @Nullable final String userName, final String source,
-                    @Nullable final char[] password, final Map<String, Object> mechanismProperties) {
-        if (mechanism != MONGODB_X509 && mechanism != MONGODB_AWS && userName == null) {
+    MongoCredential(
+            @Nullable final AuthenticationMechanism mechanism,
+            @Nullable final String userName,
+            final String source,
+            @Nullable final char[] password,
+            final Map<String, Object> mechanismProperties) {
+
+        if (userName == null && !Arrays.asList(MONGODB_X509, MONGODB_AWS).contains(mechanism)) {
             throw new IllegalArgumentException("username can not be null");
         }
 
@@ -399,7 +407,6 @@ public final class MongoCredential {
 
     private boolean mechanismRequiresPassword(@Nullable final AuthenticationMechanism mechanism) {
         return mechanism == PLAIN || mechanism == SCRAM_SHA_1 || mechanism == SCRAM_SHA_256;
-
     }
 
     /**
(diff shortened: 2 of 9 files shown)


jyemin:
Adding @DmitryLukyanov to take first crack at this.

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
```diff
@@ -104,9 +99,21 @@ private void testValidUris() {
...
+        } else {
+            if (!getExpectedValue("credential").equals(BsonNull.VALUE)) {
+                fail(String.format("Connection string '%s' should not produce credentials", input));
+            }
         }
     }
 
+    @Nullable
+    private MongoCredential getMongoCredential() {
+        ConnectionString connectionString;
```
jyemin, line 111: `return new ConnectionString(input).getCredential()` is simpler

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
```diff
@@ -104,9 +99,21 @@ private void testValidUris() {
...
             assertMechanism("credential.mechanism", credential.getMechanism());
+        } else {
+            if (!getExpectedValue("credential").equals(BsonNull.VALUE)) {
+                fail(String.format("Connection string '%s' should not produce credentials", input));
+            }
         }
     }
 
+    @Nullable
+    private MongoCredential getMongoCredential() {
```
jyemin, line 110: Not sure this adds any value given that 1 of 2 uses ignore the return value.

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
```diff
@@ -104,9 +99,21 @@ private void testValidUris() {
             }
 
             assertMechanism("credential.mechanism", credential.getMechanism());
+        } else {
+            if (!getExpectedValue("credential").equals(BsonNull.VALUE)) {
+                fail(String.format("Connection string '%s' should not produce credentials", input));
```
jyemin, line 104: should not -> should

I think?

driver-core/src/main/com/mongodb/MongoCredential.java
```diff
@@ -362,14 +362,22 @@ public MongoCredential withMechanism(final AuthenticationMechanism mechanism) {
      * @param source    the source of the user name, typically a database name
      * @param password  the password
      */
-    MongoCredential(@Nullable final AuthenticationMechanism mechanism, @Nullable final String userName, final String source,
-                    @Nullable final char[] password) {
+    MongoCredential(
+            @Nullable final AuthenticationMechanism mechanism,
```
jyemin, line 366: I don't think this is our coding standard now (one parameter per line).

driver-core/src/main/com/mongodb/MongoCredential.java
```diff
@@ -362,14 +362,22 @@ public MongoCredential withMechanism(final AuthenticationMechanism mechanism) {
...
-                    @Nullable final char[] password, final Map<String, Object> mechanismProperties) {
-        if (mechanism != MONGODB_X509 && mechanism != MONGODB_AWS && userName == null) {
+    MongoCredential(
+            @Nullable final AuthenticationMechanism mechanism,
+            @Nullable final String userName,
+            final String source,
+            @Nullable final char[] password,
+            final Map<String, Object> mechanismProperties) {
+
+        if (userName == null && !Arrays.asList(MONGODB_X509, MONGODB_AWS).contains(mechanism)) {
```
jyemin, line 380: Is OIDC going to be added to this list?  Is that what motivated this change?

driver-core/src/main/com/mongodb/MongoCredential.java
```diff
@@ -411,14 +418,16 @@ private boolean mechanismRequiresPassword(@Nullable final AuthenticationMechanis
...
     <T> MongoCredential(final MongoCredential from, final String mechanismPropertyKey, final T mechanismPropertyValue) {
-        notNull("mechanismPropertyKey", mechanismPropertyKey);
-
-        this.mechanism = from.mechanism;
-        this.userName = from.userName;
-        this.source = from.source;
-        this.password = from.password;
-        this.mechanismProperties = new HashMap<>(from.mechanismProperties);
-        this.mechanismProperties.put(mechanismPropertyKey.toLowerCase(), mechanismPropertyValue);
+        this(from.mechanism, from.userName, from.source, from.password, mapWith(
```
jyemin, line 421: Are there existing regression tests for this constructor? It doesn't look the semantics have changed, but would be good to have a test that asserts it.

driver-core/src/main/com/mongodb/internal/connection/MongoCredentialWithCache.java
```diff
@@ -63,28 +63,23 @@ public void putInCache(final Object key, final Object value) {
...
-        return cache.lock;
+    public <V> V withLock(final Supplier<V> k) {
+        return Locks.withLock(cache.lock, k);
     }
 
     static class Cache {
         private final ReentrantLock lock = new ReentrantLock();
-        private Object cacheKey;
-        private Object cacheValue;
+        private final ConcurrentHashMap<Object, Object> cache = new ConcurrentHashMap<>();
```
jyemin, line 72: Does it need to be concurrent since access is protected by a lock anyway?

jyemin, line 72: Is a requirement for support of multiple cache keys coming?  Is that the motivation for this change?

jyemin: Alternatively consider removing the lock.

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
```diff
@@ -133,27 +140,32 @@ private void assertMechanism(final String key, final String actual) {
...
-                    String expectedValue = document.getString(key).getValue();
-
-                    // If the mechanism is "GSSAPI", the default SERVICE_NAME, which is stated as "mongodb" in the specification,
-                    // is set to null in the driver.
-                    if (credential.getMechanism().equals("GSSAPI") && key.equals("SERVICE_NAME") && expectedValue.equals("mongodb")) {
-                        assertNull(credential.getMechanismProperty(key, null));
-                    } else {
-                        assertEquals(expectedValue, credential.getMechanismProperty(key, null));
-                    }
+        if (expected.isNull()) {
```
DmitryLukyanov, line 143: just a question, do you validate this option https://github.com/mongodb/specifications/blob/master/source/auth/tests/legacy/connection-string.json#L487?

bson/src/test/unit/util/ThreadTestHelpers.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.opentest4j.MultipleFailuresError;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+public final class ThreadTestHelpers {
```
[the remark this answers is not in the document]
katcharov: OIDC cache tests require that two connections are opened concurrently.

driver-core/src/main/com/mongodb/MongoCredential.java
katcharov: Could we consider making this a coding standard? That is, if there are enough parameters that the line is broken, it should be broken per parameter.

The current form obscures the number of parameters, their details, their correspondence to fields, and how signatures compare to their overloads. It seems equivalent in legibility to putting field declarations on the same line.

driver-core/src/main/com/mongodb/MongoCredential.java
katcharov: Right.

(There was a greater upheaval of the validation logic here before, but I had to move the checks further down in the process.)

driver-core/src/main/com/mongodb/MongoCredential.java
katcharov: Yes, I believe so. This constructor is used only in:
```
    public <T> MongoCredential withMechanismProperty(final String key, final T value) {
        return new MongoCredential(this, key, value);
    }
```
which has 43 usages, over half of which are directly in tests, including `MongoCredentialSpecification`. I broke it by commending out `result.put(key, value);`, which caused existing tests to fail.

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
katcharov: I believe I do read that field, but it is not in this PR since it is just OIDC-independent refactoring that I extracted from my current work.

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
katcharov: Looks like an extraction artifact. I have:
```
        if (credential != null) {
            BsonArray callbacks = (BsonArray) getExpectedValue("callback");
        ...
```
immediately after the `credential =` line. I could condense, but then I would have to expand it later. Do you have a preference?

driver-core/src/main/com/mongodb/internal/connection/MongoCredentialWithCache.java
katcharov: I added an `OidcCacheEntry`, to be stored under its own key.

It seemed that if we support multiple cache keys at all (the implication is that we do, because we have a "key" field), then they should be mapped separately (as if under different namespaces), and it was an oversight to do otherwise. This is the safest way to avoid interfering with (and needing to think about) other entries.

Though, looking at the usages, it appears that we would not share MongoCredentialWithCache across authenticators. Also, the usage in `ScramShaAuthenticator` potentially depends on having one "key" at a time, but that logic seems specific to that authenticator and shouldn't be part of the cache.

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
katcharov: ([See above comment](https://github.com/mongodb/mongo-java-driver/pull/1102#discussion_r1156328535))

driver-core/src/main/com/mongodb/internal/connection/MongoCredentialWithCache.java
katcharov: Reverted to old pair and documented.

katcharov: Reverted and documented.

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
jyemin: Small enough that it's fine to just expand it later.

jyemin:
I can't remember whether we intend to merge this or just close it.

driver-core/src/main/com/mongodb/MongoCredential.java
jyemin: Sure, we can consider it.  But it's not in scope.

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
jyemin: Did you end up simplifying this as I requested?

katcharov: Done now

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
katcharov: Done now

bson/src/test/unit/util/ThreadTestHelpers.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private ThreadTestHelpers() {
+    }
+
+    public static void executeAll(final int nThreads, final Runnable c) {
+        ExecutorService service = null;
+        try {
+            service = Executors.newFixedThreadPool(10);
```
stIncMale, line 36: 10 -> `nThreads`?

bson/src/test/unit/util/ThreadTestHelpers.java
```diff
@@ -0,0 +1,66 @@
...
+                        failures.add(e);
+                    } finally {
+                        latch.countDown();
+                    }
+                });
+            }
+            try {
+                latch.await();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
```
stIncMale, line 53: We should do `Thread.currentThread().interrupt()` before throwing.

driver-core/src/test/unit/com/mongodb/AuthConnectionStringTest.java
```diff
@@ -133,27 +126,32 @@ private void assertMechanism(final String key, final String actual) {
...
+                boolean expectedValue = document.getBoolean(key).getValue();
+                assertNotNull(actualMechanismProperty);
+                assertEquals(expectedValue, actualMechanismProperty);
+            } else {
+                fail("unsupported property type");
             }
         }
     }
 
+    @Nullable
```
stIncMale, line 154: `getExpectedValue` was annotated with `@Nullable`, yet it is always used as if it were not able to return `null`.

stIncMale (changes requested):

stIncMale (approved):
Approved, but before merging, we need to make sure that the static check and tests pass.

stIncMale (approved):
Approved, but before merging, we need to make sure that the static check and tests pass.

Update: what an interesting duplication ☝️

PR #1104, opened 2023-04-07
Support per-MongoClient DNS lookup configuration

jyemin:
* Add InetAddressResolver interface
* Add configuration of InetAddressResolver and DnsClient in MongoClientSettings and ConnectionString
* Use configured InetAddressResolver and DnsClient to resolve all SRV, TXT, A, and AAAA records

JAVA-4911

### DIFF
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -23,6 +23,7 @@ import com.mongodb.internal.diagnostics.logging.Logger;
 import com.mongodb.internal.diagnostics.logging.Loggers;
 import com.mongodb.internal.dns.DefaultDnsResolver;
 import com.mongodb.lang.Nullable;
+import com.mongodb.spi.dns.DnsClient;
 import org.bson.UuidRepresentation;
 
 import java.io.UnsupportedEncodingException;
@@ -296,6 +297,21 @@ public class ConnectionString {
      * @since 3.0
      */
     public ConnectionString(final String connectionString) {
+        this(connectionString, null);
+    }
+
+    /**
+     * Creates a ConnectionString from the given string with the given {@link DnsClient}.
+     *
+     * <p>If setting {@link MongoClientSettings#getDnsClient()} explicitly, care should be taken to call this constructor with the same
+     * {@link DnsClient}.
+     *
+     * @param connectionString the connection string
+     * @param dnsClient        the DNS client with which to resolve TXT record for the mongodb+srv protocol
+     * @since 4.10
+     * @see MongoClientSettings#getDnsClient()
+     */
+    public ConnectionString(final String connectionString, @Nullable final DnsClient dnsClient) {
         this.connectionString = connectionString;
         boolean isMongoDBProtocol = connectionString.startsWith(MONGODB_PREFIX);
         isSrvProtocol = connectionString.startsWith(MONGODB_SRV_PREFIX);
@@ -394,7 +410,7 @@ public class ConnectionString {
         }
 
         String txtRecordsQueryParameters = isSrvProtocol
-                ? new DefaultDnsResolver().resolveAdditionalQueryParametersFromTxtRecords(unresolvedHosts.get(0)) : "";
+                ? new DefaultDnsResolver(dnsClient).resolveAdditionalQueryParametersFromTxtRecords(unresolvedHosts.get(0)) : "";
         String connectionStringQueryParameters = unprocessedConnectionString;
 
         Map<String, List<String>> connectionStringOptionsMap = parseOptions(connectionStringQueryParameters);
--- a/driver-core/src/main/com/mongodb/MongoClientSettings.java
+++ b/driver-core/src/main/com/mongodb/MongoClientSettings.java
@@ -630,8 +630,8 @@ public final class MongoClientSettings {
          * and used to create an instance of {@link DnsClient}. If no implementation is discovered, then
          * {@code com.sun.jndi.dns.DnsContextFactory} will be used to resolve these records.
          *
-         * <p>If {@link #applyConnectionString(ConnectionString) applying a connection string to these settings}, care must be taken to
-         * also pass the same {@link DnsClient} as an argument to the {@link ConnectionString} constructor.
+         * <p>If {@linkplain #applyConnectionString(ConnectionString) applying a connection string to these settings}, care must be
+         * taken to also pass the same {@link DnsClient} as an argument to the {@link ConnectionString} constructor.
          *
          * @param dnsClient the DNS client
          * @return the DNS client
--- a/driver-core/src/main/com/mongodb/MongoClientSettings.java
+++ b/driver-core/src/main/com/mongodb/MongoClientSettings.java
@@ -29,6 +29,8 @@ import com.mongodb.connection.SslSettings;
 import com.mongodb.connection.StreamFactoryFactory;
 import com.mongodb.event.CommandListener;
 import com.mongodb.lang.Nullable;
+import com.mongodb.spi.dns.DnsClient;
+import com.mongodb.spi.dns.InetAddressResolver;
 import org.bson.UuidRepresentation;
 import org.bson.codecs.BsonCodecProvider;
 import org.bson.codecs.BsonValueCodecProvider;
@@ -107,6 +109,8 @@ public final class MongoClientSettings {
     private final boolean heartbeatConnectTimeoutSetExplicitly;
 
     private final ContextProvider contextProvider;
+    private final DnsClient dnsClient;
+    private final InetAddressResolver inetAddressResolver;
 
     /**
      * Gets the default codec registry.  It includes the following providers:
@@ -161,6 +165,39 @@ public final class MongoClientSettings {
     }
 
     /**
+     * Gets the {@link DnsClient} to use for resolving DNS queries.
+     *
+     * <p>If set, it will be used to resolve SRV and TXT records for mongodb+srv connections. Otherwise,
+     * implementations of {@link com.mongodb.spi.dns.DnsClientProvider} will be discovered via {@link java.util.ServiceLoader}.
+     * If no implementations are discovered, then {@code com.sun.jndi.dns.DnsContextFactory} will be used to resolve these records.
+     *
+     * <p>If applying a connection string to these settings, care must be taken to also pass the same {@link DnsClient} as an argument to
+     * the {@link ConnectionString} constructor.
+     *
+     * @return the DNS client
+     * @since 4.10
+     * @see ConnectionString#ConnectionString(String, DnsClient)
+     */
+    @Nullable
+    public DnsClient getDnsClient() {
+        return dnsClient;
+    }
+
+    /**
+     * Gets the {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
+     *
+     * <p>If set, it will be used to look up the {@link java.net.InetAddress} for each host, via
+     * {@link InetAddressResolver#lookupByName(String)}. Otherwise, {@link java.net.InetAddress#getAllByName(String)} will be used.
+     *
+     * @return the {@link java.net.InetAddress} resolver
+     * @since 4.10
+     */
+    @Nullable
+    public InetAddressResolver getInetAddressResolver() {
+        return inetAddressResolver;
+    }
+
+    /**
      * A builder for {@code MongoClientSettings} so that {@code MongoClientSettings} can be immutable, and to support easier construction
      * through chaining.
      */
@@ -193,6 +230,8 @@ public final class MongoClientSettings {
         private int heartbeatSocketTimeoutMS;
 
         private ContextProvider contextProvider;
+        private DnsClient dnsClient;
+        private InetAddressResolver inetAddressResolver;
 
         private Builder() {
         }
@@ -211,6 +250,7 @@ public final class MongoClientSettings {
             credential = settings.getCredential();
             uuidRepresentation = settings.getUuidRepresentation();
             serverApi = settings.getServerApi();
+            dnsClient = settings.getDnsClient();
             streamFactoryFactory = settings.getStreamFactoryFactory();
             autoEncryptionSettings = settings.getAutoEncryptionSettings();
             contextProvider = settings.getContextProvider();
@@ -220,6 +260,7 @@ public final class MongoClientSettings {
             socketSettingsBuilder.applySettings(settings.getSocketSettings());
             connectionPoolSettingsBuilder.applySettings(settings.getConnectionPoolSettings());
             sslSettingsBuilder.applySettings(settings.getSslSettings());
+
             if (settings.heartbeatConnectTimeoutSetExplicitly) {
                 heartbeatConnectTimeoutMS = settings.heartbeatSocketSettings.getConnectTimeout(MILLISECONDS);
             }
@@ -273,7 +314,7 @@ public final class MongoClientSettings {
         /**
          * Applies the {@link LoggerSettings.Builder} block and then sets the loggerSettings.
          *
-         * @param block the block to apply to the LoggerSettins.
+         * @param block the block to apply to the LoggerSettings.
          * @return this
          * @see MongoClientSettings#getLoggerSettings()
          * @since 4.9
@@ -580,6 +621,45 @@ public final class MongoClientSettings {
             return this;
         }
 
+        /**
+         * Sets the {@link DnsClient} to use for resolving DNS queries.
+         *
+         * <p> If set, it will be used to resolve SRV and TXT records for mongodb+srv connections. Otherwise,
+         * implementation of {@link com.mongodb.spi.dns.DnsClientProvider} will be discovered via {@link java.util.ServiceLoader}
+         * and used to create an instance of {@link DnsClient}. If no implementation is discovered, then
+         * {@code com.sun.jndi.dns.DnsContextFactory} will be used to resolve these records.
+         *
+         * <p>If applying a connection string to these settings, care must be taken to also pass the same {@link DnsClient} as an
+         * argument to the {@link ConnectionString} constructor.
+         *
+         * @param dnsClient the DNS client
+         * @return the DNS client
+         * @since 4.10
+         * @see ConnectionString#ConnectionString(String, DnsClient)
+         */
+        public Builder dnsClient(@Nullable final DnsClient dnsClient) {
+            this.dnsClient = dnsClient;
+            return this;
+        }
+
+        /**
+         * Sets the {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
+         *
+         * <p>If set, it will be used to look up the {@link java.net.InetAddress} for each host, via
+         * {@link InetAddressResolver#lookupByName(String)}. Otherwise,
+         * an implementation of {@link com.mongodb.spi.dns.InetAddressResolverProvider} will be discovered via
+         * {@link java.util.ServiceLoader} and used to create an instance of {@link InetAddressResolver}.  If no implementation is
+         * discovered, {@link java.net.InetAddress#getAllByName(String)} will be used to lookup the {@link java.net.InetAddress}
+         * instances for a host.
+         *
+         * @param inetAddressResolver the InetAddress provider
+         * @return the {@link java.net.InetAddress} resolver
+         * @since 4.10
+         */
+        public Builder inetAddressResolver(@Nullable final InetAddressResolver inetAddressResolver) {
+            this.inetAddressResolver = inetAddressResolver;
+            return this;
+        }
 
         // Package-private to provide interop with MongoClientOptions
         Builder heartbeatConnectTimeoutMS(final int heartbeatConnectTimeoutMS) {
@@ -964,6 +1044,8 @@ public final class MongoClientSettings {
         compressorList = builder.compressorList;
         uuidRepresentation = builder.uuidRepresentation;
         serverApi = builder.serverApi;
+        dnsClient = builder.dnsClient;
+        inetAddressResolver = builder.inetAddressResolver;
         autoEncryptionSettings = builder.autoEncryptionSettings;
         heartbeatSocketSettings = SocketSettings.builder()
                 .readTimeout(builder.heartbeatSocketTimeoutMS == 0
--- a/driver-core/src/main/com/mongodb/MongoClientSettings.java
+++ b/driver-core/src/main/com/mongodb/MongoClientSettings.java
@@ -251,6 +251,7 @@ public final class MongoClientSettings {
             uuidRepresentation = settings.getUuidRepresentation();
             serverApi = settings.getServerApi();
             dnsClient = settings.getDnsClient();
+            inetAddressResolver = settings.getInetAddressResolver();
             streamFactoryFactory = settings.getStreamFactoryFactory();
             autoEncryptionSettings = settings.getAutoEncryptionSettings();
             contextProvider = settings.getContextProvider();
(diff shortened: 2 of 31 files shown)


driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -296,6 +297,21 @@ public class ConnectionString {
      * @since 3.0
      */
     public ConnectionString(final String connectionString) {
+        this(connectionString, null);
+    }
+
+    /**
+     * Creates a ConnectionString from the given string with the given {@link DnsClient}.
+     *
+     * <p>If setting {@link MongoClientSettings#getDnsClient()} explicitly, care should be taken to call this constructor with the same
```
jyemin, line 306: This is unfortunate but I don't see a way around it, as ConnectionString is responsible for TXT record resolution.

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
```diff
@@ -0,0 +1,83 @@
...
+import com.mongodb.spi.dns.InetAddressResolverProvider;
+
+import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+final class ServerAddressWithResolver extends ServerAddress {
```
jyemin, line 32: The important bit.  Why do we extend ServerAddress?  It's because the `StreamFactory` interface is part of the API, and defines the method `Stream create(ServerAddress serverAddress)`, and third-party implementations can assume they are allowed to call `ServerAddress#getSocketAddress` to resolve the IP address.  By extending ServerAddress, we can inject the InetAddressResolver behavior into all Stream implementations (including our own).

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final long serialVersionUID = -1;  // this internal class will never be serialized
+
+    private static final InetAddressResolver DEFAULT_INET_ADDRESS_PROVIDER;
+
+    static {
+        DEFAULT_INET_ADDRESS_PROVIDER = StreamSupport.stream(ServiceLoader.load(InetAddressResolverProvider.class).spliterator(), false)
```
jyemin, line 38: As with DnsClient, this is not integration tested.

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
```diff
@@ -35,21 +37,23 @@
  */
 public final class DefaultDnsResolver implements DnsResolver {
 
+    private static final DnsClient DEFAULT_DNS_CLIENT;
+
+    static {
+        DEFAULT_DNS_CLIENT = StreamSupport.stream(ServiceLoader.load(DnsClientProvider.class).spliterator(), false)
```
jyemin, line 43: Simplify the code to match what's now done in `ServerAddressWithResolver`

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public boolean equals(final Object o) {
+        return super.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
```
stIncMale, line 82: 1. Implementing overrides like this is equivalent to not overriding at all. The latter approach is taken for the `toString` method.
2. With the way `ServerAddress.equals` is implemented, no `ServerAddressWithResolver` is ever equal to an object whose runtime type is `ServerAddress` (and wise versa, because the current `equals` implementation is symmetric, as it must be). Is this actually OK, or do we have sets / maps / something else where we rely on equality and objects of these two types may be compared with each other?

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolver.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * <p>Implementations are encouraged to implement their own caching policies, as there is
+     * no guarantee that the caller will implement a cache.
+     *
+     * <p> Inspired by the interface of the same name that was added in Java 18, but duplicated
+     * here due to the minimum supported Java version of Java 8 at the time of creation.
+     *
+     * @param host the specified hostname
+     * @return a list of IP addresses for the requested host
+     * @throws NullPointerException if either parameter is {@code null}
```
stIncMale, line 52: `if either parameter` - there is only one parameter.

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolverProvider.java
```diff
@@ -0,0 +1,45 @@
...
+ * Service-provider class for {@link InetAddressResolver}.
+ *
+ * <p> A resolver provider is a factory for custom implementations of {@linkplain
+ * InetAddressResolver InetAddress resolvers}. A resolver defines operations for
+ * looking up (resolving) host names.
+ *
+ * <p>A resolver provider is a concrete implementation of this interface that has a
+ * zero-argument constructor and implements the abstract methods specified below.
+ *
+ * <p>The driver discovers implementations of this interface via {@link java.util.ServiceLoader}.
```
stIncMale, line 29: The first of the two paragraphs is both excessive and incomplete (it misses the ["The type must be public and must not be an inner class"](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/ServiceLoader.html) requirement). The same appears to be the case for the original line in the documentation of `java.net.spi.InetAddressResolverProvider`. I propose to leave only the second paragraph.

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolver.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public interface InetAddressResolver {
+    /**
+     * Given the name of a host, returns a list of IP addresses of the requested
+     * address family associated with a provided hostname.
+     *
+     * <p> {@code host} should be a machine name, such as "{@code www.example.com}",
+     * not a textual representation of its IP address. No validation is performed on
+     * the given {@code host} name: if a textual representation is supplied, the name
+     * resolution is likely to fail and {@link UnknownHostException} may be thrown.
```
stIncMale, line 42: [optional]
Am I the only one to whom this text adopted from the Java SE docs appears to be weird? This interface is supposed to document requirements for a user who implements the interface. Whether to perform a validation and whether to fail if the argument is a string representation of an IP address, is up to the user. The documentation may say that the implementation does not have to support string representations of IP addresses and does not have to do any validation. But instead, the documentation is formulated as if it's a documentation of a provided implementation, which it is not.

stIncMale (changes requested):
I have not yet reviewed the whole PR.

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+final class ServerAddressWithResolver extends ServerAddress {
+    private static final long serialVersionUID = -1;  // this internal class will never be serialized
```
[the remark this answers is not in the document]
jyemin: If we want to be thorough we could define readResolve and serialize this as a ServerAddress.  

In practice our exceptions are not really serializable due to other classes that are included as properties.

At the least I can remove the comment

driver-sync/src/test/functional/com/mongodb/client/AbstractDnsConfigurationTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Test
+    public void testDnsClientConfiguration() throws InterruptedException, ExecutionException, TimeoutException {
+        DnsException exception = new DnsException("", new Exception());
+        DnsClient dnsClient = (name, type) -> {
+            throw exception;
+        };
+
+        CompletableFuture<Throwable> exceptionReceived = new CompletableFuture<>();
+        CountDownLatch latch = new CountDownLatch(1);
```
stIncMale, line 78: `latch` is not really used and can be removed.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
stIncMale: The changes made in this PR require `DnsClient` to be thread-safe, which is a breaking change. Previosuly, `DnsClient` was not marked thread-safe and it was not required to be threadsafe:
- A new instance of `DnsClient` was always created in the constructor of `DefaultDnsResolver` (`DnsClientProvider.create` is documented as constructing a new instance).
- A new instance of `DefaultDnsResolver` was created and used in the constructor of `ConnectionString`.
- A new instance of `DefaultDnsResolver` was created per `DefaultDnsSrvRecordMonitor`, where it was used by the single corresponding monitoring thread.

driver-core/src/main/com/mongodb/ConnectionString.java
jyemin: I just can't think of a concrete plan for it, even if we allow breaking changes.  What happens if you call any of the `applyConnectioString` methods on the setting classes, and then pass a MongoClientSettings in to `MongoClient#create`.  If you then try to apply the TXT record to the `MongoClientSettings` you have the problem that the settings are supposed to trump the TXT record and so you have to remember in the settings the difference between a value that is the default and the same value that was specified explicitly.  Tricky.

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -211,6 +250,7 @@ private Builder(final MongoClientSettings settings) {
             credential = settings.getCredential();
             uuidRepresentation = settings.getUuidRepresentation();
             serverApi = settings.getServerApi();
+            dnsClient = settings.getDnsClient();
```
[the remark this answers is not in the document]
jyemin: Huh, I expected the assertion on line 217 of MongoClientSettingsSpecification to catch this bug...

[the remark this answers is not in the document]
jyemin: Ok, I see it.  It's because I didn't add the two new properties to equals/hashCode

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolverProvider.java
jyemin: I'm not sure which paragraph you want to delete.  You say the first of the two paragraphs, but seems like you're referring to the second paragraph (not including the one-line summary).  Are you asking to delete the paragraph starting with "A resolver provider is a concrete implementation"?  If so, why not fix the paragraph to make it complete?  What's excessive about it?

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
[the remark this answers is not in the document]
jyemin: Removed the comment.  Done?

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolver.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * This interface defines operations for looking up host names.
+ *
+ * <p> The <i>system-wide resolver</i> can be customized by
+ * <a href="InetAddressResolverProvider.html#system-wide-resolver">
+ * deploying an implementation</a> of {@link InetAddressResolverProvider}.
+ * <p>
+ * @since 4.10
+ */
+public interface InetAddressResolver {
```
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
jyemin: Done

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
stIncMale: I see you have already asked the Quarkus team in https://jira.mongodb.org/browse/JAVA-4512. I agree that we can just make the change if their implementation is thread-safe.

driver-core/src/main/com/mongodb/ConnectionString.java
stIncMale: # On getters

A `ConnectionString` that does not do what it's not supposed to do should verify the format of the string, may destructure it on components, but should not provide any getters to users (even if currently implementing most of them don't require any IO, who knows what may happen in the future). Therefore, a `MongoClientSettings`/`MongoClientSettings.Builder` that was fed a `ConnectionString` should not provide any getters to users either. I don't think these getters are actually required by users: even if in some bizarre cases a user want to know what's inside a `MongoClientSettings`/`Builder`, he may pack that data in any form needed together with `MongoClientSettings`/`Builder` in another structure (a weird requirement needs some additional work on the user side, and is not supported out of the box, that's fine).

If for some reason we end up strongly wanting to provide getters (I hope we don't), we may have two sets of `MongoClientSettings`/`Builder` types: as long as a `ConnectionString` (or any other future mechanism that requires IO or customizable logic similarly to the `+srv` connection string modifier) is not used, we use the types with getters, otherwise we return a type without getters (see the next section for more details).

# On the path forward

I see two paths here (only the first one seems practical, I don't even know why I describe the second one):

1. We allow applying a connection string to settings only when a new settings object is created. For example, this means that `Builder.applyConnectionString` should create a new `Builder`, instead of modifying `this`. If we have two sets of `MongoClientSettings`/`Builder` types (see the previous section for more details), then `Builder.applyConnectionString` returns a `Builder` without getters.
2. We maintain a sequence of updates in `MongoClientSettings`/`Builder`:
  - `applyConnectionString` introduces a new element to the sequence;
  - any other setter is merged together with the immediate previous update as long as that update is not `applyConnectionString`, otherwise it also introduces a new element to the sequence.
At the point when we are ready to resolve `ConnectionString` (and any other settings that require IO or customizable logic), we resolve and merge the updates in the sequence.

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
stIncMale: We have `ConcurrentMap<ServerAddress, ServerTuple> AbstractMultiServerCluster.addressToServerTupleMap` and use `Set<ServerAddress>` as the type of a local variable / return value in multiple places. Are we sure that instances of `ServerAddress` and instances of `ServerAddressWithResolver` are never mixed together? Even if don't mix them as the contents of the aforementioned collections, it's enough to read from them (`get` from map, or `contains` on a set) using an instance of the type different from the one stored inside, to get ourselves into trouble.

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -580,6 +621,45 @@ public Builder contextProvider(@Nullable final ContextProvider contextProvider)
...
 
+        /**
+         * Sets the {@link DnsClient} to use for resolving DNS queries.
+         *
+         * <p> If set, it will be used to resolve SRV and TXT records for mongodb+srv connections. Otherwise,
+         * implementation of {@link com.mongodb.spi.dns.DnsClientProvider} will be discovered via {@link java.util.ServiceLoader}
+         * and used to create an instance of {@link DnsClient}. If no implementation is discovered, then
+         * {@code com.sun.jndi.dns.DnsContextFactory} will be used to resolve these records.
+         *
+         * <p>If applying a connection string to these settings, care must be taken to also pass the same {@link DnsClient} as an
```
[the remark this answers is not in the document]
stIncMale: With `@link` the whole _"applying a connection string to these settings"_ is rendered with monospace font. Let's use `@linkplain`.

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolverProvider.java
stIncMale: > A resolver provider is a concrete implementation of this interface that has a zero-argument constructor and implements the abstract methods specified below.

Is an inclompete duplicate of

> The driver discovers implementations of this interface via {@link java.util.ServiceLoader}.

The documentation of `ServiceLoader` specifies all the requirements, and we, in my opinion, should not duplicate them.

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -160,6 +164,39 @@ public static Builder builder(final MongoClientSettings settings) {
...
+    @Nullable
+    public DnsClient getDnsClient() {
+        return dnsClient;
+    }
+
+    /**
+     * Gets the {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
+     *
+     * <p>If set, it will be used to look up the {@link java.net.InetAddress} for each host, via
+     * {@link InetAddressResolver#lookupByName(String)}. Otherwise, {@link java.net.InetAddress#getAllByName(String)} will be used.
```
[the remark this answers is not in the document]
stIncMale: Created https://jira.mongodb.org/browse/JAVA-4937.

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolver.java
stIncMale: It's better to make the requirements of our `InetAddressResolver` as similar as possible to those of the Java SE API `InetAddressResolver`. I.e., we also should not require `InetAddressResolver` to resolve textual representations of IP addresses.

Also, I think we should not copy the text from the Java SE API for this interface: instead of expressing it as a requirement for the implementor (our user), it is expressed as it if it written for the consumer (the driver maintainers), which is wrong.

stIncMale (changes requested):
.

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
jyemin: It looks like it is.

jyemin: Added ThreadSafe annotation to both interfaces

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
jyemin: Is your suggestion to remove equal/hashCode here?  If so, I think it's fine.  Either way, I don't anticipate that these methods will ever be called by the driver or Stream implementation of an instance of this class.

driver-core/src/main/com/mongodb/MongoSocketException.java
```diff
@@ -33,7 +33,7 @@ public class MongoSocketException extends MongoException {
      * @param msg the message
      * @param e the cause
      */
-    MongoSocketException(final String msg, final ServerAddress serverAddress, final Throwable e) {
+    public MongoSocketException(final String msg, final ServerAddress serverAddress, final Throwable e) {
```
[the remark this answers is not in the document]
jyemin: We generally need them to be public for the reason mentioned.  The `VisibleForTesting` annotation is not really accurate, but I suppose we could use it for constructors that we don't otherwise want to be public.

driver-core/src/main/com/mongodb/MongoClientSettings.java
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
jyemin: Note that this class only exists because the StreamFactory interface is public.  It it wasn't we would have much more flexibility in the design.

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolverProvider.java
jyemin: Done, and also done in `DnsClientProvider`

driver-core/src/main/com/mongodb/spi/dns/InetAddressResolver.java
jyemin: Done.  I added a check in `ServerAddressWithResolver` for IP literals.

driver-core/src/main/com/mongodb/internal/connection/ServerAddressWithResolver.java
stIncMale: > Is your suggestion to remove equal/hashCode here?

I did not have a suggestion, I was just interested in whether we are sure that instances of `ServerAddress` and `ServerAddressWithResolver` are never used together in a context where `hashCode`/`equals` are used.

driver-core/src/main/com/mongodb/MongoSocketException.java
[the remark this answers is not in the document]
stIncMale: > We generally need them to be public for the reason mentioned.

Being `public` and being part of the public API is not the same. In this case, the constructor has to be `public`, but, as far as I can judge from this conversation, does not have to be part of the public API.

> The `VisibleForTesting` annotation is not really accurate

I agree, it should not be used here. But we can document the constructor as not being part of the public API, similarly to how the constructor of `MongoConnectionPoolClearedException` is documented. By doing so, we reduce the API surface. Do you see reasons why this is a bad idea?

driver-core/src/main/com/mongodb/ConnectionString.java
stIncMale: Does the above seem like something we could explore further when designing the 5.0 API improvements? If yes, I'll create a ticket. If not, or if we don't really want to really have any significant API changes in 5.0, then I won't.

I went to such length in this discussion because the requirement `If setting {@link MongoClientSettings#getDnsClient()} explicitly, care should be taken to call this constructor with the same {@link DnsClient}` screams at me that we failed at the API design here.

stIncMale (approved):
Approved, with two optional things:

- https://github.com/mongodb/mongo-java-driver/pull/1104#discussion_r1169344096
- https://github.com/mongodb/mongo-java-driver/pull/1104#discussion_r1169341571

driver-core/src/main/com/mongodb/ConnectionString.java
stIncMale: I found something related in the SDAM spec https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-summary.rst:

_"A multi-threaded or asynchronous client's constructor MUST NOT do any I/O. ... The justification is that if clients can be constructed when the deployment is in some states but not in other states, it leads to an unfortunate scenario: When the deployment is passing through a strange state, long-running applications may keep working, but any applications restarted during this period fail."_

Thile the SDAM spec focuses on specifically client constructor and the state of the MongoDB deployment, the principle of not doing IO in constructors should always be followed for the same reason pointed out in the spec: object creation should not fail because of the temporary unfortunate external state. The constructor of `ConnectionString` may fail due not network issues, which may result in the whole application failing to start. Despite the constructor of a `ConnectionString` formally not being part of the client constructor in the Java driver, I think, in essence the our driver violates the idea that the SDAM spec expresses.

driver-core/src/main/com/mongodb/MongoSocketException.java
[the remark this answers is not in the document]
jyemin: There's just a lack of consistency in that the large majority of our exception class constructors are already part of the API and there is not a very clear path to changing that (would we deprecate them all?).  So making this one not part of the API doesn't help us all that much.  I wouldn't be against a ticket that considers options for shrinking out total constructor API.

driver-core/src/main/com/mongodb/ConnectionString.java
jyemin: Maybe what was bad was supporting TXT records. :).  I can only say that we knew at the time we were violating the SDAM spec recommendation but didn't see a lot of good options given the API we had to support at the time.

stIncMale: Created https://jira.mongodb.org/browse/JAVA-4944.

driver-core/src/main/com/mongodb/MongoSocketException.java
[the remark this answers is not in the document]
stIncMale: Created https://jira.mongodb.org/browse/JAVA-4942 and https://jira.mongodb.org/browse/JAVA-4943.

PR #1108, opened 2023-04-17
Fix duplicate retry log messages

stIncMale:
The code that logs retries is now added by the `decorate*WithRetries` methods. Sometimes, this may result in us not logging even the short description of the command that is being retried, because currently we know the command (we create it) only after checking out a connection and checking if the server supports retries. So if the first attempt fails at a point before we create the command (e.g., because of the `MongoConnectionPoolClearedException`), we don't do `retryState.attach(AttachmentKeys.commandDescriptionSupplier(), command::getFirstKey, false)`, and the retrying attempt does not have the information about the command that it retries. This situation will be improved when [Log operation ID when retrying](https://jira.mongodb.org/browse/JAVA-4939) is done.

# Log messages examples
## Before the PR
For each retry attempt we emitted two corresponding log messages. It looked either like
```
> Retrying the operation aggregate ... attempt #2
> Retrying the operation aggregate ... attempt #2
```
or like
```
> Retrying an operation ... attempt #2
> Retrying the operation aggregate ... attempt #2
```

Note how in the second example the first message does not provide a description (`aggregate`) of operation that was being retried. This is because when a retry attempt starts, it may have no information on what command is bing retried, if the first attempt failed too soon to create the command and leave its description for the retry attempt to use.

The first of the two messages in each example is from the start of a retry attempt, the second message is from a point at which the retry attempt is guaranteed to know what is being retried (this point is somewhere in the middle of a retry attempt).

## After the PR
For each retry attempt we emit one corresponding log message. It looks either like

```
> Retrying the operation 'aggregate' ... attempt #2
```

or, if the information on the retrying command is not yet available

```
> Retrying the operation ... attempt #2
```

The message comes from the start of a retry attempt.

So now we don't duplicate retry messages, and don't log them from the middle of a retry attempt, but as a result, we may sometimes not log the description of the retried command. Logging operation ID will allow users to correlate this messages with previous log messages to still see what is being retried, which is why I filed [Log operation ID when retrying](https://jira.mongodb.org/browse/JAVA-4939). Alternatively, we may refactor our code to propagate the description of the retried command from the higher API level (the one that is closer to the user), but I don't think we should do that.

JAVA-4910

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
@@ -185,13 +185,19 @@ final class CommandOperationHelper {
 
     static <R> Supplier<R> decorateReadWithRetries(final RetryState retryState, final Supplier<R> readFunction) {
         return new RetryingSyncSupplier<>(retryState, CommandOperationHelper::chooseRetryableReadException,
-                CommandOperationHelper::shouldAttemptToRetryRead, readFunction);
+                CommandOperationHelper::shouldAttemptToRetryRead, () -> {
+            logRetryExecute(retryState);
+            return readFunction.get();
+        });
     }
 
     static <R> AsyncCallbackSupplier<R> decorateReadWithRetries(final RetryState retryState,
             final AsyncCallbackSupplier<R> asyncReadFunction) {
         return new RetryingAsyncCallbackSupplier<>(retryState, CommandOperationHelper::chooseRetryableReadException,
-                CommandOperationHelper::shouldAttemptToRetryRead, asyncReadFunction);
+                CommandOperationHelper::shouldAttemptToRetryRead, callback -> {
+            logRetryExecute(retryState);
+            asyncReadFunction.get(callback);
+        });
     }
 
     static <D, T> T executeRetryableRead(
@@ -213,13 +219,12 @@ final class CommandOperationHelper {
             final CommandReadTransformer<D, T> transformer,
             final boolean retryReads) {
         RetryState retryState = initialRetryState(retryReads);
-        Supplier<T> read = decorateReadWithRetries(retryState, () -> {
-            logRetryExecute(retryState);
-            return withSourceAndConnection(readConnectionSourceSupplier, false, (source, connection) -> {
+        Supplier<T> read = decorateReadWithRetries(retryState, () ->
+            withSourceAndConnection(readConnectionSourceSupplier, false, (source, connection) -> {
                 retryState.breakAndThrowIfRetryAnd(() -> !canRetryRead(source.getServerDescription(), binding.getSessionContext()));
                 return createReadCommandAndExecute(retryState, binding, source, database, commandCreator, decoder, transformer, connection);
-            });
-        });
+            })
+        );
         return read.get();
     }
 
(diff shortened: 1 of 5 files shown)


driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -603,8 +607,8 @@ static void logRetryExecute(final RetryState retryState) {
             Throwable exception = retryState.exception().orElseThrow(Assertions::fail);
             int oneBasedAttempt = retryState.attempt() + 1;
             LOGGER.debug(commandDescription == null
-                    ? format("Retrying an operation due to the error \"%s\"; attempt #%d", exception, oneBasedAttempt)
-                    : format("Retrying the operation %s due to the error \"%s\"; attempt #%d",
+                    ? format("Retrying the operation due to the error \"%s\"; attempt #%d", exception, oneBasedAttempt)
+                    : format("Retrying the operation '%s' due to the error \"%s\"; attempt #%d",
```
stIncMale, line 611: I doubt _"Retrying **the** operation ..."_ is grammatically correct in a situation when we don't specify what operation is being retried, but having the same longer prefix for two messages seems better.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -412,9 +419,8 @@ static <T, R> R executeRetryableWrite(
                     // attach `maxWireVersion`, `retryableCommandFlag` ASAP because they are used to check whether we should retry
                     retryState.attach(AttachmentKeys.maxWireVersion(), maxWireVersion, true)
                             .attach(AttachmentKeys.retryableCommandFlag(), isRetryWritesEnabled(command), true)
-                            .attach(AttachmentKeys.commandDescriptionSupplier(), command::getFirstKey, true)
+                            .attach(AttachmentKeys.commandDescriptionSupplier(), command::getFirstKey, false)
                             .attach(AttachmentKeys.command(), command, false);
```
stIncMale, line 423: I don't know why the command description for writes was auto-cleared, but given that the command was not auto-cleared, auto-clearing the description did not make sense.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -471,9 +476,8 @@ static <T, R> void executeRetryableWriteAsync(
...
```
stIncMale, line 480: This is similar to https://github.com/mongodb/mongo-java-driver/pull/1108/files?show-viewed-files=true&file-filters%5B%5D=#r1169033282.

jyemin:
> The code that logs retries is now added by the decorate*WithRetries methods. Sometimes, this may result in us not logging even the short description of the command that is being retried, because currently we know the command (we create it) only after checking out a connection and checking if the server supports retries. 

Are you saying that this is the result of the refactoring, or this it's the result of the required retry algorithm, regardless of the implementation strategy.  I think you mean the former, but want to confirm before proceeding.

stIncMale:
> Are you saying that this is the result of the refactoring, or this it's the result of the required retry algorithm, regardless of the implementation strategy. I think you mean the former, but want to confirm before proceeding.

@jyemin, I added examples and more explanations to the description of the PR.

stIncMale:
> It would be more useful if the driver emits the retry log message with information about the command.

There are two issues with such an approach in addition to returning the sprinkled logging calls all over the place, which is why I did not choose this path:

1. We log a message stating that we are retrying well after we have started the attempt, which results in emitting some messages about activities that are part of retrying before the message stating that we are retrying. This makes our logs confusing.
2. If we log in the middle of an attempt, we will not log that we have attempted to retry in some cases.

The second item is on its own worse than what we have with the current PR, but combined with the first issue, it's way worse. If we want our cake and eat it too, we should get back to the situation when we emit two log messages per a retry attempt, but we change the wording in the messages to something like:

- First message: _"Begin retrying an operation <[sometimes missing] command description> ... attempt #2"_
- Second message: _"Created the <command description> command to retry it ... attempt #2"_

jyemin (approved):
LGTM

PR #1109, opened 2023-04-19
Fix UnixServerAddress interaction with InetAddressResolver

jyemin:
JAVA-4911


driver-sync/src/test/functional/com/mongodb/client/AbstractDnsConfigurationTest.java
```diff
@@ -69,6 +73,36 @@ public void clusterDescriptionChanged(final ClusterDescriptionChangedEvent event
...
+        CompletableFuture<UnknownHostException> exceptionReceived = new CompletableFuture<>();
+        MongoClientSettings settings = MongoClientSettings.builder()
+                .applyToClusterSettings(builder ->
+                        builder.hosts(Collections.singletonList(new ServerAddress("127.0.01")))
+                                .addClusterListener(new ClusterListener() {
+                                    @Override
+                                    public void clusterDescriptionChanged(final ClusterDescriptionChangedEvent event) {
+                                        ServerDescription serverDescription = event.getNewDescription().getServerDescriptions().get(0);
+                                        if (serverDescription.getState() == ServerConnectionState.CONNECTED) {
+                                            exceptionReceived.complete(exception);
```
stIncMale, line 95: `exception` is thrown by `resolver`, which must not happen as the `resolver` must not be used by this test. But then `exceptionReceived` is also completed with `exception` on something that looks like the "happy path". I am not sure this is correct. But even if it is, then it is really difficult (I can't) to see why this is correct.

I also realized, that I don't understand what's going on with `exceptionReceived` in `testInetAddressResolverConfiguration`. Am I lacking some knowledge, or is the test indeed confusing?

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/client/AbstractDnsConfigurationTest.java
```diff
@@ -69,6 +73,36 @@ public void clusterDescriptionChanged(final ClusterDescriptionChangedEvent event
         }
     }
 
+    @ParameterizedTest(name = "InetAddressResolver should not be used to resolve IP literal {0}")
+    @ValueSource(strings = {"127.0.0.1", "::1", "[0:0:0:0:0:0:0:1]"})
+    public void testInetAddressResolverDoesNotResultIpLiteral() throws InterruptedException, ExecutionException, TimeoutException {
```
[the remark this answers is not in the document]
jyemin: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractDnsConfigurationTest.java
jyemin: Copy pasta.  Renamed the variable to `serverConnectedFuture`.

jyemin: > Am I lacking some knowledge, or is the test indeed confusing?

I'm not sure :). Happy to talk it over.

driver-sync/src/test/functional/com/mongodb/client/AbstractDnsConfigurationTest.java
```diff
@@ -57,15 +63,49 @@ public void testInetAddressResolverConfiguration() throws InterruptedException,
...
                                         }
                                     }
                                 }))
                 .inetAddressResolver(resolver)
                 .build();
 
         try (MongoClient ignored = createMongoClient(settings)) {
-            assertEquals(exception, exceptionReceived.get(10, SECONDS));
+            Throwable exceptionReceived = exceptionReceivedFuture.get(1, SECONDS);
+            assertEquals(MongoSocketException.class, exceptionReceived.getClass());
```
jyemin, line 75: This is overkill but added it for clarity

driver-sync/src/test/functional/com/mongodb/client/AbstractDnsConfigurationTest.java
stIncMale: The following in `testInetAddressResolverDoesNotResultIpLiteral` seems less confusing:

- `CompletableFuture<Boolean> serverConnectedFuture = new CompletableFuture<>()`
- `serverConnectedFuture.complete(true)`
- `assertTrue(serverConnectedFuture.get(10, SECONDS))`

And if done, one does not need `exception` anymore, and can do the following:

```java
InetAddressResolver resolver = host -> {
    fail();
};
```

jyemin: Done.  Except that `fail()` in the resolver won't do what you want it to do, since it's not invoked on the main thread.  But I did inline the exception

stIncMale (approved):

driver-sync/src/test/functional/com/mongodb/client/AbstractDnsConfigurationTest.java
stIncMale: > fail() in the resolver won't do what you want it to do

I think it would result in the test failure, even if indirectly. But what exception the `resolver` throws really does not matter here.

PR #1110, opened 2023-04-24
Stop creating the ECC collection in v2 of queryable encryption

stIncMale:
- Reviewing commit by commit allows skipping the commit that synchronizes the tests with the specs repo.
- Creating a collection with `encryptedFields` specified now requires MongoDB 7.0+, which is why I specified `@mongodb.server.release 7.0` everywhere `encryptedFields` is used (the driver does not do the check when dropping a collection, but it still does not seem meaningful to document `DropCollectionOptions.encryptedFields` any different than other places).
- `CreateCollectionOptions.encryptedFields`/`DropCollectionOptions.encryptedFields` were not documented as `@Beta`, which appears to be a mistake, given that the whole queryable encryption feature is in preview. I documented them as such.

JAVA-4858

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
@@ -16,6 +16,8 @@
 
 package com.mongodb.internal.operation;
 
+import com.mongodb.MongoClientException;
+import com.mongodb.MongoException;
 import com.mongodb.WriteConcern;
 import com.mongodb.client.model.ChangeStreamPreAndPostImagesOptions;
 import com.mongodb.client.model.Collation;
@@ -23,6 +25,7 @@ import com.mongodb.client.model.TimeSeriesGranularity;
 import com.mongodb.client.model.TimeSeriesOptions;
 import com.mongodb.client.model.ValidationAction;
 import com.mongodb.client.model.ValidationLevel;
+import com.mongodb.connection.ConnectionDescription;
 import com.mongodb.internal.async.SingleResultCallback;
 import com.mongodb.internal.binding.AsyncWriteBinding;
 import com.mongodb.internal.binding.WriteBinding;
@@ -51,6 +54,7 @@ import static com.mongodb.internal.operation.OperationHelper.LOGGER;
 import static com.mongodb.internal.operation.OperationHelper.releasingCallback;
 import static com.mongodb.internal.operation.OperationHelper.withAsyncConnection;
 import static com.mongodb.internal.operation.OperationHelper.withConnection;
+import static com.mongodb.internal.operation.ServerVersionHelper.serverIsLessThanVersionSevenDotZero;
 import static com.mongodb.internal.operation.WriteConcernHelper.appendWriteConcernToCommand;
 import static java.util.Arrays.asList;
 import static java.util.Collections.singletonList;
@@ -232,6 +236,7 @@ public class CreateCollectionOperation implements AsyncWriteOperation<Void>, Wri
     @Override
     public Void execute(final WriteBinding binding) {
         return withConnection(binding, connection -> {
+            checkEncryptedFieldsSupported(connection.getDescription());
             getCommandFunctions().forEach(commandCreator ->
                 executeCommand(binding, databaseName, commandCreator.get(), connection,
                         writeConcernErrorTransformer())
@@ -276,14 +281,11 @@ public class CreateCollectionOperation implements AsyncWriteOperation<Void>, Wri
      * <li>Create the collection with name encryptedFields["escCollection"] using default options.
      *   If encryptedFields["escCollection"] is not set, use the collection name enxcol_.<collectionName>.esc.
      *   Creating this collection MUST NOT check if the collection namespace is in the AutoEncryptionOpts.encryptedFieldsMap.
-     * <li>Create the collection with name encryptedFields["eccCollection"] using default options.
-     *   If encryptedFields["eccCollection"] is not set, use the collection name enxcol_.<collectionName>.ecc.
-     *   Creating this collection MUST NOT check if the collection namespace is in the AutoEncryptionOpts.encryptedFieldsMap.
      * <li>Create the collection with name encryptedFields["ecocCollection"] using default options.
      *   If encryptedFields["ecocCollection"] is not set, use the collection name enxcol_.<collectionName>.ecoc.
      *   Creating this collection MUST NOT check if the collection namespace is in the AutoEncryptionOpts.encryptedFieldsMap.
      * <li>Create the collection collectionName with collectionOptions and the option encryptedFields set to the encryptedFields.
-     * <li>Create the the index {"__safeContent__": 1} on collection collectionName.
+     * <li>Create the index {"__safeContent__": 1} on collection collectionName.
      *  </ol>
      * </p>
      * @return the list of commands to run to create the collection
@@ -294,7 +296,6 @@ public class CreateCollectionOperation implements AsyncWriteOperation<Void>, Wri
         }
         return asList(
                 () -> getCreateEncryptedFieldsCollectionCommand("esc"),
-                () -> getCreateEncryptedFieldsCollectionCommand("ecc"),
                 () -> getCreateEncryptedFieldsCollectionCommand("ecoc"),
                 this::getCreateCollectionCommand,
                 () -> new BsonDocument("createIndexes", new BsonString(collectionName))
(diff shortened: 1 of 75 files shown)


stIncMale:
@jyemin I added you because I'd like to have you reviewed specific commit that updates the API docs: https://github.com/mongodb/mongo-java-driver/pull/1110/commits/34f38d0e834399e9549c329251618d6611933eb2.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -361,6 +362,27 @@ private BsonDocument getCreateCollectionCommand() {
...
+            throw new MongoClientException("Driver support of Queryable Encryption is incompatible with server."
+                    + " Upgrade server to use Queryable Encryption.");
+        }
+    }
+
+    /**
+     * @return {@code true} iff the {@linkplain #checkEncryptedFieldsSupported(ConnectionDescription) check} was successful,
+     * and the {@code callback} was not completed with a failed result.
+     */
+    private boolean checkEncryptedFieldsSupported(final AsyncConnection connection, final SingleResultCallback<Void> callback) {
```
jyemin, line 376: Consider changing the type of the first parameter to `ConnectionDescription` so it's clear there is no I/O being done.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -361,6 +362,27 @@ private BsonDocument getCreateCollectionCommand() {
...
 
+    private void checkEncryptedFieldsSupported(final ConnectionDescription connectionDescription) throws MongoException {
+        if (encryptedFields != null && serverIsLessThanVersionSevenDotZero(connectionDescription)) {
+            throw new MongoClientException("Driver support of Queryable Encryption is incompatible with server."
+                    + " Upgrade server to use Queryable Encryption.");
+        }
+    }
+
+    /**
+     * @return {@code true} iff the {@linkplain #checkEncryptedFieldsSupported(ConnectionDescription) check} was successful,
```
jyemin, line 373: Is this a new async design pattern you're introducing?  It doesn't look familiar.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -384,6 +406,9 @@ public void onResult(@Nullable final Void result, @Nullable final Throwable t) {
                 finalCallback.onResult(null, t);
                 return;
             }
+            if (!checkEncryptedFieldsSupported(connection, finalCallback)) {
```
jyemin, line 409: The intent would be clearer if this check were moved to line 255.  That would match the behavior of the sync version.

jyemin (changes requested):
LGTM on the Javadoc.  A few implementation questions/comments.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
stIncMale: The same approach is used by `LoopState.breakAndCompleteIf`/`RetryState.breakAndCompleteIfRetryAnd` to make async callback-based code to look similar to synchronous code.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
stIncMale: I am surprised I didn't do this originally.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
stIncMale: I thought moving it closer to the invocation of `getCommandFunctions` (though, not inside the constructor of `ProcessCommandsCallback`) is better. Does not matter much. Done.

jyemin: Notice that `this` is passed as an argument to `executeCommandAsync` at the end of this function, so the check will happen three times if it were left here.  It will give the same answer all three times, but the intent is clearer if it only happens once.

jyemin (approved):

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
stIncMale: Thanks!

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -361,6 +366,27 @@ private BsonDocument getCreateCollectionCommand() {
...
+
+    /**
+     * @return {@code true} iff the {@linkplain #checkEncryptedFieldsSupported(ConnectionDescription) check} was successful.
+     * The {@code callback} is completed (with a failed result) iff the check was not successful.
+     */
+    private boolean checkEncryptedFieldsSupported(final ConnectionDescription connectionDescription, final SingleResultCallback<Void> callback) {
+        try {
+            checkEncryptedFieldsSupported(connectionDescription);
+            return true;
+        } catch (Exception e) {
```
frest0512, line 384: [Optional] 

I noticed that checkEncryptedFieldsSupported method is declared to throw MongoException. However, the method catches Exception, which covers all types of exceptions including both checked and runtime exceptions. This could potentially catch more than necessary. Considering narrowing it down to RuntimeException since if a checked exception is thrown down the road later, the compiler will bring it to attention to treat it as special case. Otherwise, it might be caught unexpectedly.

[RSPEC-2221](https://rules.sonarsource.com/java/RSPEC-2221)

frest0512: Since it is designed for a callback to handle all exceptions and log errors for the user to see, therefore, disregard comment above.

driver-core/src/main/com/mongodb/internal/operation/ServerVersionHelper.java
```diff
@@ -31,6 +31,7 @@ public final class ServerVersionHelper {
     public static final int FOUR_DOT_FOUR_WIRE_VERSION = 9;
     public static final int FIVE_DOT_ZERO_WIRE_VERSION = 12;
     public static final int SIX_DOT_ZERO_WIRE_VERSION = 17;
+    private static final int SEVEN_DOT_ZERO_WIRE_VERSION = 21;
```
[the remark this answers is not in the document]
stIncMale: Other fields are used outside of the `com.mongodb.internal.operation` package, the new field is used only within the `ServerVersionHelper`, so `private` is enough.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
```diff
@@ -361,6 +366,27 @@ private BsonDocument getCreateCollectionCommand() {
...
+    private void checkEncryptedFieldsSupported(final ConnectionDescription connectionDescription) throws MongoException {
+        if (encryptedFields != null && serverIsLessThanVersionSevenDotZero(connectionDescription)) {
+            throw new MongoClientException("Driver support of Queryable Encryption is incompatible with server."
+                    + " Upgrade server to use Queryable Encryption.");
+        }
+    }
+
+    /**
+     * @return {@code true} iff the {@linkplain #checkEncryptedFieldsSupported(ConnectionDescription) check} was successful.
+     * The {@code callback} is completed (with a failed result) iff the check was not successful.
```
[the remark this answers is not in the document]
stIncMale: It's ["if and only if"](https://en.wikipedia.org/wiki/If_and_only_if). The statement

> The callback is completed if the check was not successful.

leaves unspecified what happens if the check was not successful. I.e., to be precise, one has to write something like

> The callback is completed if the check was not successful, otherwise the callback is not touched.

Or, one may use "iff" to write

> The callback is completed iff the check was not successful.

which is as short as the first variant, and as precise as the second one.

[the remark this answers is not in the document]
stIncMale: Note also that the statement

> The callback is completed only if the check was not successful.

does not allow deriving whether the check was successful in a situation when the callback is not completed, while "iff" does.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
stIncMale: > Considering narrowing it down to RuntimeException since if a checked exception is thrown down the road later, the compiler will bring it to attention to treat it as special case.

I would have loved if we could do this, and actually, this is what I was doing previously. However, then https://jira.mongodb.org/browse/JAVA-4575 happened (which is really, Kotlin's fault). To avoid thinking too much each time whether a method may sneakily throw a checked exception due to executing external code, we decided to catch `Exception` even though it's weird if one thinks only about Java.

driver-core/src/main/com/mongodb/internal/operation/CreateCollectionOperation.java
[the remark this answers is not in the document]
vbabanin: Understood, thanks for explanation!

vbabanin (approved):

PR #1113, opened 2023-05-04
Implement prose tests checking that neither `mongocryptd` nor its client are created when they are not supposed to be

stIncMale:
Note that while JAVA-4794 adds one test, the other tests in the same section [8. Bypass Spawning mongocryptd](https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#bypass-spawning-mongocryptd) appear to be missing in our codebase, so I also implemented them.

The "Commits" WUI view does something weird to the commit messages. The actual commit messages are something like the following, and point to the ID of the group in the spec:

- Add tests [`#8`](https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#bypass-spawning-mongocryptd) JAVA-4794
- Add test [`#20`](https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#20-bypass-creating-mongocryptd-client-when-shared-library-is-loaded) JAVA-4741

There are small deviations in some tests comparing to their prose description, made to make our life better in a situation when the tests fail. I created DRIVERS-2624 with a PR suggesting to include the changes in the spec.

JAVA-4741
JAVA-4794


.evergreen/run-csfle-tests-with-mongocryptd.sh
```diff
@@ -64,12 +64,10 @@ echo "Running tests with Java ${JAVA_VERSION}"
...
-          --tests com.mongodb.client.ClientSideEncryptionTest \
-          --tests com.mongodb.client.unified.ClientSideEncryptionTest \
+          --tests "*.ClientEncryption*" --tests "*.ClientSideEncryption*" \
       driver-reactive-streams:test \
-          --tests com.mongodb.reactivestreams.client.ClientSideEncryptionTest \
-          --tests com.mongodb.reactivestreams.client.unified.ClientSideEncryptionTest \
+          --tests "*.ClientEncryption*" --tests "*.ClientSideEncryption*" \
       driver-scala:integrationTest \
-          --tests org.mongodb.scala.ClientSideEncryptionTest
+          --tests "*.ClientEncryption*" --tests "*.ClientSideEncryption*"
```
stIncMale, line 73: This change together with the `assumeFalse/assumeTrue(CRYPT_SHARED_LIB_PATH_SYS_PROP_VALUE != null)` checks in `AbstractClientSideEncryptionNotSpawnMongocryptdTest` solves the issue discussed in https://mongodb.slack.com/archives/CFPHLTRMK/p1683160528003019: some of the tests added in this PR not only do not require `crypt_shared`, but require the library not to be loaded, which is why they are run separately (in different process) from those that load `crypt_shared` (once it is loaded, it can't be unloaded).

driver-core/src/main/com/mongodb/internal/NoCheckedAutoCloseable.java
```diff
@@ -0,0 +1,47 @@
...
+ *          }
+ *      }
+ *  }
+ * }</pre>
+ * One may avoid the warning by using {@link NoCheckedAutoCloseable} as the type of the {@code shutdown1}, {@code shutdown2} variables
+ * and removing the {@code throws Exception} declaration. Alternatively, one may annotate the {@code MyAutoCloseable} class
+ * with {@code @SuppressWarnings("try")}, in which case annotating the {@code MyAutoCloseable.close} method
+ * with {@code @SuppressWarnings ("try")} is no longer needed, but such an approach is too coarse-grained.
+ */
+public interface NoCheckedAutoCloseable extends AutoCloseable {
```
stIncMale, line 44: The documentation of this interface explains why it was added.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionTest.java
```diff
@@ -195,10 +196,7 @@ public void setUp() {
         }
 
         Map<String, Object> extraOptions = new HashMap<>();
-        String cryptSharedLibPath = System.getProperty("org.mongodb.test.crypt.shared.lib.path", "");
-        if (!cryptSharedLibPath.isEmpty()) {
-            extraOptions.put("cryptSharedLibPath", cryptSharedLibPath);
-        }
```
stIncMale, line 201: I also needed to use this Java system property, so I refactored this code a bit.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionNotCreateMongocryptdClientTest.java
```diff
@@ -0,0 +1,224 @@
...
+        assumeTrue(serverVersionAtLeast(4, 2));
+        assumeTrue(CRYPT_SHARED_LIB_PATH_SYS_PROP_VALUE != null);
+        mongocryptdConnectionTracker = ConnectionTracker.start();
+        client = createMongoClient(MongoClientSettings.builder(getMongoClientSettings())
+                .autoEncryptionSettings(AutoEncryptionSettings.builder()
+                        .kmsProviders(singletonMap(LOCAL_KMS_PROVIDER_ID, singletonMap("key", localKmsProviderKey())))
+                        .keyVaultNamespace(KEY_VAULT_NAMESPACE.getFullName())
+                        .extraOptions(Stream.of(
+                                new SimpleImmutableEntry<>("cryptSharedLibPath", CRYPT_SHARED_LIB_PATH_SYS_PROP_VALUE),
+                                new SimpleImmutableEntry<>("mongocryptdURI", format("mongodb://%s:%d/db?serverSelectionTimeoutMS=%d",
```
stIncMale, line 88: The prose description does not specify `serverSelectionTimeoutMS`, however it speeds up the test in a situation when the test fails.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionNotCreateMongocryptdClientTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+    protected abstract MongoClient createMongoClient(MongoClientSettings settings);
+
+    @Test
+    @SuppressWarnings("try")
+    void whenCryptSharedLoaded() throws Exception {
+        //noinspection unused
+        try (AutoCloseable assertNoConnectionsOnAutoCloseToPreserveAssertionFailure = mongocryptdConnectionTracker) {
+            collection.insertOne(Document.parse("{unencrypted: 'test'}"));
+        }
```
stIncMale, line 119: If we don't assert when automatically closing, then we may not see the assertion error if the driver fails behave correctly: if the driver attempts to spawn `mongocryptd`, it will fail as the socket is bound (by the test), which will lead to `insertOne` failing, which will result is skipping the assertion.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionNotCreateMongocryptdClientTest.java
```diff
@@ -0,0 +1,224 @@
...
+            int foundPort = serverSocket.getLocalPort();
+            if (foundPort != DEFAULT_MONGOCRYPTD_PORT) {
+                return foundPort;
+            } else {
+                return findAvailableMongocryptdLoopbackPort();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to find an available port", e);
+        }
+    }
```
stIncMale, line 134: We could have just used `serverSocket.bind(new InetSocketAddress(InetAddress.getLoopbackAddress(), 0))` right away in `ConnectionTracker.start` instead of introducing the `findAvailableMongocryptdLoopbackPort` method. However, the same approach is not helpful for the tests in `AbstractClientSideEncryptionNotSpawnMongocryptdTest`. Additionally, making sure that we never use the default `mongocryptd` port in `AbstractClientSideEncryptionNotSpawnMongocryptdTest` also improve the isolation between different tests.

To avoid using different approaches, I simply use `findAvailableMongocryptdLoopbackPort` in both `AbstractClientSideEncryptionNotCreateMongocryptdClientTest` and `AbstractClientSideEncryptionNotSpawnMongocryptdTest.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionNotSpawnMongocryptdTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    private void setUpKeyVaultNamespace() {
+        try (MongoClient client = createMongoClient(MongoClientSettings.builder().build())) {
+            MongoDatabase db = client.getDatabase(KEY_VAULT_NAMESPACE.getDatabaseName());
+            db.drop();
+            db.getCollection(KEY_VAULT_NAMESPACE.getCollectionName(), BsonDocument.class)
+                    .withWriteConcern(WriteConcern.MAJORITY)
+                    .insertOne(EXTERNAL_KEY);
+        }
+    }
```
stIncMale, line 195: The prose tests do not require having any keys in the `KEY_VAULT_NAMESPACE`: only `viaMongocryptdBypassSpawn` attempts to insert an encrypted field, and that insertion is supposed to fail. However, when the driver misbehaves or tests fail (in my case, `viaMongocryptdBypassSpawn` was failing if `viaLoadingSharedLibrary` runs before it in the same process due to `viaLoadingSharedLibrary` loading `crypt_shared`), things may get too difficult to understand if the encryption is not properly set up.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionNotSpawnMongocryptdTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    private static Map<String, Object> commonExtraAutoEncryptionOptions(final InetSocketAddress mongocryptdSocketAddress) {
+        return singletonMap("mongocryptdSpawnArgs", asList(
+                // We pick a random available `mongocryptd` port and also include it in the PID file path
+                // to reduce the chances of different test runs interfering with each other. The interference
+                // may come from the fact that once spawned, `mongocryptd` stays up and running for some time,
+                // which may cause failures in other runs if they use the same `mongocryptd` port / PID file.
+                format("--pidfilepath=bypass-spawning-mongocryptd-%d.pid", mongocryptdSocketAddress.getPort()),
+                format("--port=%d", mongocryptdSocketAddress.getPort())));
```
stIncMale, line 204: The comment in the code explains why this is done the way it is, even though the prose tests do not require anything like that.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionNotCreateMongocryptdClientTest.java
vbabanin: Thanks for the explanation. One suggestion I have is to move the assertNoConnections() method to the finally block. I think it could make it easier to follow the context of the test (like in style of the given, when, then). If we move it to the finally block, it will be easier to locate the assertion quickly. What do you think?

```
        try (AutoCloseable assertNoConnectionsOnAutoCloseToPreserveAssertionFailure = mongocryptdConnectionTracker) {
                collection.insertOne(Document.parse("{unencrypted: 'test'}"));
        }finally {
            mongocryptdConnectionTracker.assertNoConnections();
        }
```

stIncMale: Added a comment, as discussed.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionNotSpawnMongocryptdTest.java
```diff
@@ -0,0 +1,243 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @SafeVarargs
+    @SuppressWarnings("varargs")
+    private static <K, V> Map<K, V> merge(final Map<K, V> map, final Entry<K, V>... entries) {
+        HashMap<K, V> result = new HashMap<>(map);
+        result.putAll(Stream.of(entries).collect(Collectors.toMap(Entry::getKey, Entry::getValue)));
```
vbabanin, line 240: nit: extra Map created in collect. Might be
`for (Map.Entry<K, V> entry : entries) {
    result.put(entry.getKey(), entry.getValue());
}`

vbabanin (approved):

vbabanin (approved):

stIncMale:
@jyemin Are you OK with me merging this tests-only PR based on the approval of one reviewer, or do you want to review it?

jyemin:
I'd like to take a look so please hold off on merging

driver-core/src/main/com/mongodb/internal/NoCheckedAutoCloseable.java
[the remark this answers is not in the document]
stIncMale: It's no longer in the PR: you said what suppressing `try` warnings on the class level is fine despite being coarse-grained, and I did just that, which eliminates the need for this interface.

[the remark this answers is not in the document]
jyemin: Sorry, this was an old comment that I neglected to delete before publishing.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionNotCreateMongocryptdClientTest.java
```diff
@@ -0,0 +1,223 @@
...
+            collection.insertOne(Document.parse("{unencrypted: 'test'}"));
+            // ConnectionTracker.assertNoConnections is called on auto-close
+        }
+    }
+
+    static int findAvailableMongocryptdLoopbackPort() {
+        try (ServerSocket serverSocket = new ServerSocket()) {
+            serverSocket.bind(new InetSocketAddress(InetAddress.getLoopbackAddress(), 0));
+            int foundPort = serverSocket.getLocalPort();
+            if (foundPort != DEFAULT_MONGOCRYPTD_PORT) {
```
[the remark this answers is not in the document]
stIncMale: We recurse into `findAvailableMongocryptdLoopbackPort` before we close `serverSocket`, which means that the arbitrary available port picked by the `bind` method is guaranteed to be different from the one that caused us to recurse (as the one caused us to recurse is still bound). I.e., in the worst case we recurse once.

.evergreen/run-csfle-tests-with-mongocryptd.sh
```diff
@@ -64,12 +64,10 @@ echo "Running tests with Java ${JAVA_VERSION}"
       ${GRADLE_EXTRA_VARS} \
       --stacktrace --info --continue \
       driver-legacy:test \
-          --tests com.mongodb.ClientSideEncryptionLegacyTest \
+          --tests "*.ClientEncryption*" --tests "*.ClientSideEncryption*" \
```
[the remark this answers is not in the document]
stIncMale: It appears more appropriate to rename all the existing CSFLE tests as part of a separate PR, if we do that at all. I used the `*.Client*Encryption*` pattern.

jyemin (approved):
LGTM

PR #1115, opened 2023-05-12
Deprecate `Parameterizable`, introduce default `CodecProvider.get(Class<T>, List<Type>, CodecRegistry)` instead

stIncMale:
Turns out, we don't even need to introduce `ParameterizedCodecProvider` to replace `Parameterizable`. All we need is a new default method on `CodecProvider`. And the cost of still supporting `Parameterizable` is really small: a single `if` block in `ProvidersCodecRegistry.getFromCodecProvider`.

I also refactored `RecordCodec`, `CollectionCodec`, `MapCodecV2`, `org.bson.codecs.kotlin.DataClassCodec` such that they no longer implement `Parameterizable`.

### How is this related to JAVA-4954?

JAVA-4954 is about trying to improve `PojoCodecImpl` to be more like `RecordCodec`. Given that we discovered a better alternative to `Parameterizable`, which can be implemented without breaking public API, it seems worth making the improvement before trying to refactor `PojoCodecImpl`. Otherwise, we will likely have to refactor `PojoCodecImpl` again in the future.

JAVA-4967

### DIFF
--- a/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodecProvider.kt
+++ b/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodecProvider.kt
@@ -18,9 +18,13 @@ package org.bson.codecs.kotlin
 import org.bson.codecs.Codec
 import org.bson.codecs.configuration.CodecProvider
 import org.bson.codecs.configuration.CodecRegistry
+import java.lang.reflect.Type
 
 /** A Kotlin reflection based Codec Provider for data classes */
 public class DataClassCodecProvider : CodecProvider {
     override fun <T : Any> get(clazz: Class<T>, registry: CodecRegistry): Codec<T>? =
-        DataClassCodec.create(clazz.kotlin, registry)
+        get(clazz, emptyList(), registry)
+
+    override fun <T : Any> get(clazz: Class<T>, typeArguments: List<Type>, registry: CodecRegistry): Codec<T>? =
+        DataClassCodec.create(clazz.kotlin, registry, typeArguments)
 }
--- a/bson/src/main/org/bson/codecs/CollectionCodec.java
+++ b/bson/src/main/org/bson/codecs/CollectionCodec.java
@@ -20,18 +20,15 @@ import org.bson.BsonReader;
 import org.bson.BsonWriter;
 import org.bson.Transformer;
 import org.bson.UuidRepresentation;
-import org.bson.codecs.configuration.CodecConfigurationException;
 import org.bson.codecs.configuration.CodecRegistry;
 
-import java.lang.reflect.Type;
 import java.util.Collection;
 import java.util.List;
 
 import static org.bson.assertions.Assertions.notNull;
-import static org.bson.codecs.ContainerCodecHelper.getCodec;
 
 /**
- * A parameterized Codec for {@code Collection<Object>}.
+ * A codec for {@code Collection<Object>}.
  *
  * <p>Supports {@link Collection}, {@link List}, {@link java.util.AbstractCollection}, {@link java.util.AbstractList},
  * {@link java.util.Set}, {@link java.util.NavigableSet}, {@link java.util.SortedSet}, {@link java.util.AbstractSet} or any
@@ -47,7 +44,7 @@ import static org.bson.codecs.ContainerCodecHelper.getCodec;
  */
 @SuppressWarnings("rawtypes")
 final class CollectionCodec<C extends Collection<Object>> extends AbstractCollectionCodec<Object, C>
-        implements OverridableUuidRepresentationCodec<C>, Parameterizable {
+        implements OverridableUuidRepresentationCodec<C> {
 
     private final CodecRegistry registry;
     private final BsonTypeCodecMap bsonTypeCodecMap;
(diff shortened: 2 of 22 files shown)


bson/src/main/org/bson/codecs/configuration/CodecProvider.java
```diff
@@ -34,10 +38,34 @@ public interface CodecProvider {
...
      * Get a {@code Codec} using the given context, which includes, most importantly, the Class for which a {@code Codec} is required.
      *
+     * <p>This method is called only if {@link #get(Class, List, CodecRegistry)} is not properly overridden.</p>
+     *
      * @param clazz the Class for which to get a Codec
      * @param registry the registry to use for resolving dependent Codec instances
      * @param <T> the type of the class for which a Codec is required
      * @return the Codec instance, which may be null, if this source is unable to provide one for the requested Class
      */
     <T> Codec<T> get(Class<T> clazz, CodecRegistry registry);
```
stIncMale, line 48: The driver no longer calls the "old" `CodecProvider.get`, i.e., we may even deprecate this method in the future and remove in 5.0, thus going back to having only one method `CodecProvider`.

bson/src/main/org/bson/codecs/configuration/CodecProvider.java
```diff
@@ -34,10 +38,34 @@ public interface CodecProvider {
...
+     * or because another {@link CodecProvider} did not propagate {@code clazz}'s type arguments via the {@code registry},
+     * which may if that {@link CodecProvider} does not properly override {@link #get(Class, List, CodecRegistry)}.
+     * @param registry the registry to use for resolving dependent Codec instances
+     * @return the Codec instance, which may be null, if this source is unable to provide one for the requested Class
+     * @param <T> the type of the class for which a Codec is required
+     * @since 4.10
+     */
+    default <T> Codec<T> get(Class<T> clazz, List<Type> typeArguments, CodecRegistry registry) {
+        return get(clazz, registry);
+    }
```
stIncMale, line 70: We don't have `Beta` in `bson`, but maybe we should. This will allow us to mark the new method as `@Beta(Beta.Reason.CLIENT)` and give ourselves more time to realize whether the signature of the method should be different. @jyemin What do you think?

bson/src/main/org/bson/codecs/CollectionCodec.java
```diff
@@ -47,7 +44,7 @@
  */
 @SuppressWarnings("rawtypes")
 final class CollectionCodec<C extends Collection<Object>> extends AbstractCollectionCodec<Object, C>
-        implements OverridableUuidRepresentationCodec<C>, Parameterizable {
+        implements OverridableUuidRepresentationCodec<C> {
```
stIncMale, line 47: I noticed that while `CollectionCodec` implements `OverridableUuidRepresentationCodec`, `ParameterizedCollectionCodec` does not. The same goes about `MapCodecV2` and `ParameterizedMapCodec`. Is this intentional / OK?

bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodecProvider.kt
```diff
@@ -18,9 +18,13 @@ package org.bson.codecs.kotlin
...
+import java.lang.reflect.Type
 
 /** A Kotlin reflection based Codec Provider for data classes */
 public class DataClassCodecProvider : CodecProvider {
     override fun <T : Any> get(clazz: Class<T>, registry: CodecRegistry): Codec<T>? =
-        DataClassCodec.create(clazz.kotlin, registry)
+        get(clazz, emptyList(), registry)
+
+    override fun <T : Any> get(clazz: Class<T>, typeArguments: List<Type>, registry: CodecRegistry): Codec<T>? =
+        DataClassCodec.create(clazz.kotlin, registry, typeArguments)
```
stIncMale, line 29: If this provider is rolled back to the previous implementation, some tests in `DataClassCodecTest` (for example, `testDataClassWithParameterizedDataClass`), which this PR did not modify, start failing with `CodecConfigurationException: Could not find codec for number with type N`. I.e., they do test parameterization and it appear to work when the provider properly overrides the new `get` method.

bson/src/main/org/bson/internal/ProvidersCodecRegistry.java
```diff
@@ -51,40 +53,55 @@ public <T> Codec<T> get(final Class<T> clazz) {
     @Override
     public <T> Codec<T> get(final Class<T> clazz, final List<Type> typeArguments) {
         notNull("typeArguments", typeArguments);
-        isTrueArgument("typeArguments is not empty", !typeArguments.isEmpty());
```
stIncMale, line 54: This assertion and the similar assertion in `ChildCodecRegistry.get` contradicted the documentation of `ChildCodecRegistry.get`, which explicitly allows empty `typeArguments` (this PR has not changed that documentation). While all our tests happened to pass despite the assertions existing, users can pass empty `typeArguments` and get an unjustified error.

jyemin (changes requested):
Very nice improvement.

bson/src/main/org/bson/codecs/configuration/CodecProvider.java
stIncMale: We were mistakenly happy previously more than once about things in this domain. Let's hope, this is the last time)

bson/src/main/org/bson/codecs/configuration/CodecProvider.java
```diff
@@ -34,10 +38,34 @@ public interface CodecProvider {
     /**
      * Get a {@code Codec} using the given context, which includes, most importantly, the Class for which a {@code Codec} is required.
      *
+     * <p>This method is called only if {@link #get(Class, List, CodecRegistry)} is not properly overridden.</p>
```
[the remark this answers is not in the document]
jyemin: Yep.

jyemin (approved):

PR #1119, opened 2023-05-17
ByteBufBsonDocument remains a ByteBuf

jyemin:
This one got a bit complicated.  The root cause of the bug is that `ByteBufBsonDocument#entrySet` hydrates its byte buffer into the base class's hash map instead of implementing the entry set via iterating over the raw BSON.  To avoid this hydration completely, including for nested documents and arrays, requires quite a few code changes, including the introduction of a new class, `ByteBufBsonArray`, that does the same job for nested arrays.

JAVA-4917

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonArray.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonArray.java
@@ -1,0 +1,300 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonArray.java
```diff
@@ -0,0 +1,300 @@
...
+    public List<BsonValue> getValues() {
+        List<BsonValue> values = new ArrayList<>();
+        for (BsonValue cur: this) {
+            //noinspection UseBulkOperation
+            values.add(cur);
+        }
+        return values;
+    }
+
+    private static final String READ_ONLY_MESSAGE = "This BsonArray instance is read-only";
```
vbabanin, line 58: nit: static final can be moved to the top of the class

driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonArray.java
```diff
@@ -0,0 +1,300 @@
...
+
+    @Override
+    public int size() {
+        int size = 0;
```
vbabanin, line 62: Since the `ByteBufBsonArray` is read-only and its size doesn't change after initialization, we can consider caching the `size` value to avoid additional computations during each `size()` call.

jyemin:
> Could / should RawBsonDocument now delegate to ByteBufBsonDocument internally 

Added JAVA-5023 to consider this

PR #1122, opened 2023-05-22
JAVA-4966 : Fix Macro for sealed trait

kistria:
https://jira.mongodb.org/browse/JAVA-4966

If the sealed trait and the Macro are not defined is the same module, we have the following error `No known subclasses of the sealed trait`
Add a `val _` to make sure that `isCaseClass` returns the correct value.  This "hack" is described in https://github.com/scala/bug/issues/7755

It's complicated to test that, because it requires having multi modules. But this change solved the issue in my project.


rozza (approved):
Thanks for the fix @kistria

PR #1125, opened 2023-05-23
Cleanup documentation referring to removed property

kdombeck:
This should have probably been removed in this [commit](https://github.com/mongodb/mongo-java-driver/commit/dfcf57f3dfea23609bfe911b788cc347173d4d76).

Maybe the modified sentence in this PR should just be completely removed so there is not duplication of documentation and actual properties in the class. If you would like that let me know and I can remove the entire sentence.


jyemin:
Good catch.  Thanks!

PR #1130, opened 2023-05-30
Add builders for percentile and median accumulators/window functions

vbabanin:
This pull request (PR) introduces the capability to calculate the median and percentile of numeric values in the MongoDB aggregation pipeline for $group and $setWindowFields stages

JAVA-3860


driver-core/src/main/com/mongodb/client/model/Accumulators.java
```diff
@@ -67,6 +67,42 @@ public static <TExpression> BsonField avg(final String fieldName, final TExpress
...
+     * @param inExpression   The input expression.
+     * @param pExpression    The expression representing a percentiles of interest.
+     * @param method         The method to be used for computing the percentiles.
+     * @param <InExpression> The type of the input expression.
+     * @param <PExpression>  The type of the percentile expression.
+     * @return The requested {@link BsonField}.
+     * @mongodb.driver.manual reference/operator/aggregation/percentile/ $percentile
+     * @mongodb.server.release 7.0
+     */
+    public static <InExpression, PExpression> BsonField percentile(final String fieldName, final InExpression inExpression,
```
vbabanin, line 85: In the design doc, you may notice that if no method value is specified, the default value is 'discrete', suggesting the possibility of creating an overloaded method without the `method` argument. However, at present, the decision regarding the default choice for the `method` has been deferred. Therefore, all methods currently require the method argument.

driver-core/src/main/com/mongodb/client/model/Accumulators.java
```diff
@@ -67,6 +67,42 @@ public static <TExpression> BsonField avg(final String fieldName, final TExpress
...
+     * of percentiles of interest within a group, where each element is a numeric value between 0.0 and 1.0 (inclusive).
+     *
+     * @param fieldName      The field computed by the accumulator.
+     * @param inExpression   The input expression.
+     * @param pExpression    The expression representing a percentiles of interest.
+     * @param method         The method to be used for computing the percentiles.
+     * @param <InExpression> The type of the input expression.
+     * @param <PExpression>  The type of the percentile expression.
+     * @return The requested {@link BsonField}.
+     * @mongodb.driver.manual reference/operator/aggregation/percentile/ $percentile
```
jyemin, line 82: Add `@since 4.10` to both methods

PR #1132, opened 2023-06-01
Remove "Public Technical Preview" from Queryable Encryption Equality …

vbabanin:
This PR removes @Beta annotations from Queryable Encryption API.

JAVA-4958

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ ext {
     zstdVersion = '1.5.5-2'
     awsSdkV2Version = '2.18.9'
     awsSdkV1Version = '1.12.337'
-    mongoCryptVersion = '1.8.0'
+    mongoCryptVersion = '1.8.0-alpha1'
     projectReactorVersion = '2022.0.0'
     junitBomVersion = '5.8.2'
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -55,7 +55,7 @@ ext {
     zstdVersion = '1.5.5-2'
     awsSdkV2Version = '2.18.9'
     awsSdkV1Version = '1.12.337'
-    mongoCryptVersion = '1.8.0-alpha1'
+    mongoCryptVersion = '1.8.0'
```
jyemin, line 58: This hasn't been released yet (https://repo1.maven.org/maven2/org/mongodb/mongodb-crypt/).

jyemin: I assume that's why all the checks are failing.

Let's revert this change and do it separately, after we release 1.8.0 Java bindings.

jyemin (changes requested):

build.gradle
vbabanin: Reverted back to version 1.9.0-alpha1

jyemin (approved):

PR #1135, opened 2023-06-05
Kotlin coroutine update.

rozza:
The `Flow` interface is not stable for inheritance in 3rd party libraries. Updated code to use the `AbstractFlow` as recommended instead via the sealed `MongoAbstractFlow` class.

JAVA-4950

### DIFF
--- a/config/detekt/baseline.xml
+++ b/config/detekt/baseline.xml
@@ -10,7 +10,7 @@
     <ID>MaxLineLength:MapReduceIterable.kt$MapReduceIterable$*</ID>
     <ID>SwallowedException:MockitoHelper.kt$MockitoHelper.DeepReflectionEqMatcher$e: Throwable</ID>
     <ID>TooManyFunctions:ClientSession.kt$ClientSession : jClientSession</ID>
-    <ID>TooManyFunctions:FindFlow.kt$FindFlow&lt;T : Any> : Flow</ID>
+    <ID>TooManyFunctions:FindFlow.kt$FindFlow&lt;T : Any> : MongoAbstractFlow</ID>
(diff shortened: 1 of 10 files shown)


stIncMale (commented):
I looked at the changes, and I don't have objections, which is expected, given that I don't know Kotlin. However, I also can't reasonably approve the PR, so I will stay a bystander.

jyemin (approved):

vsct-jburet:
> Hi @zigzago - just wanted to double check you are happy with this PR

If you look at the implementation of MongoAbstractFlow, it adds some behaviour

```kotlin
public final override suspend fun collect(collector: FlowCollector<T>) {
        val safeCollector = SafeCollector(collector, coroutineContext)
        try {
            collectSafely(safeCollector)
        } finally {
            safeCollector.releaseIntercepted()
        }
    }
```
Also in the doc of AbstractFlow (https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-abstract-flow/):

"Base class for stateful implementations of Flow". I'm not sure that mongo flows are stateful.

I don't say your implementation will not work, but it would be cleaner for my point of view to use the kotlin delegate pattern:

```kotlin
class AggregateFlow<T : Any>(private val wrapped: AggregatePublisher<T>) : Flow<T> by wrapped.asFlow()
```

Then you rely completely on Publisher.asFlow() kotlin implementation (which is safe).

HTH

Julien

rozza:
Hi @vsct-jburet,

Good points, they are stateful in that they contain the cursor state, but that is all managed by the `Publisher` and not the `Flow`.  Will update to `MongoFlow` and delegate instead.

Ross

vbabanin (approved):

PR #1137, opened 2023-06-06
Improve Time-Series Bucketing Scalability

vbabanin:
Introduce customizable bucketMaxSpanSeconds and bucketRoundingSeconds options for Time-Series collections, providing users with more control over bucketing behaviors. These options allow for setting the maximum time difference between timestamps within a bucket and adjusting the rounding precision. Enhancing flexibility in managing Time-Series data.

JAVA-4888

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
+++ b/driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
@@ -18,6 +18,9 @@ package com.mongodb.client.model;
 
 import com.mongodb.lang.Nullable;
 
+import java.util.concurrent.TimeUnit;
+
+import static com.mongodb.assertions.Assertions.isTrueArgument;
 import static com.mongodb.assertions.Assertions.notNull;
 
 /**
@@ -31,6 +34,8 @@ public final class TimeSeriesOptions {
     private final String timeField;
     private String metaField;
     private TimeSeriesGranularity granularity;
+    private Long bucketMaxSpanSeconds;
+    private Long bucketRoundingSeconds;
 
     /**
      * Construct a new instance.
@@ -104,12 +109,102 @@ public final class TimeSeriesOptions {
         return this;
     }
 
+    /**
+     * Returns the maximum time span between measurements in a bucket.
+     *
+     * @param timeUnit the time unit.
+     * @return time span between measurements.
+     * @mongodb.server.release 6.3
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     * @see #bucketMaxSpan(Long, TimeUnit)
+     * @since 4.10
+     */
+    @Nullable
+    public Long getBucketMaxSpan(final TimeUnit timeUnit) {
+        if (bucketMaxSpanSeconds == null) {
+            return null;
+        }
+        return timeUnit.convert(bucketMaxSpanSeconds, TimeUnit.SECONDS);
+    }
+
+    /**
+     * Sets the maximum time span between measurements in a bucket.
+     * <p>
+     * The value of {@code bucketMaxSpanSeconds} must be the same as {@code bucketRoundingSeconds}.
+     * If you set the bucketMaxSpanSeconds, parameter, you can't set the granularity parameter
+     * </p>
+     *
+     * @param bucketMaxSpan - time span between measurements. After conversion to seconds using
+     *                      {@link TimeUnit#convert(long, java.util.concurrent.TimeUnit)}, the value must be &gt;= 1.
+     * @param timeUnit      - the time unit.
+     * @return this
+     * @mongodb.server.release 6.3
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     * @see #getBucketMaxSpan(TimeUnit)
+     * @since 4.10
+     */
+    public TimeSeriesOptions bucketMaxSpan(@Nullable final Long bucketMaxSpan, final TimeUnit timeUnit) {
+        if (bucketMaxSpan == null) {
+            this.bucketMaxSpanSeconds = null;
+        } else {
+            this.bucketMaxSpanSeconds = TimeUnit.SECONDS.convert(bucketMaxSpan, timeUnit);
+            isTrueArgument("bucketMaxSpan, after conversion to seconds, must be >= 1", bucketMaxSpanSeconds > 0);
+        }
+        return this;
+    }
+
+    /**
+     * Returns the time interval that determines the starting timestamp for a new bucket.
+     *
+     * @param timeUnit the time unit.
+     * @return the time interval.
+     * @mongodb.server.release 6.3
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     * @see #bucketRounding(Long, TimeUnit)
+     * @since 4.10
+     */
+    @Nullable
+    public Long getBucketRounding(final TimeUnit timeUnit) {
+        if (bucketRoundingSeconds == null) {
+            return null;
+        }
+        return timeUnit.convert(bucketRoundingSeconds, TimeUnit.SECONDS);
+    }
+
+    /**
+     * Specifies the time interval that determines the starting timestamp for a new bucket.
+     * <p>
+     * The value of {@code bucketRounding} must be the same as {@code bucketMaxSpan}. If you set the {@code bucketRounding}, parameter,
+     * you can't set the granularity parameter.
+     * </p>
+     *
+     * @param bucketRounding - time interval. After conversion to seconds using
+     *                       {@link TimeUnit#convert(long, java.util.concurrent.TimeUnit)}, the value must be &gt;= 1.
+     * @param timeUnit       - the time unit.
+     * @return this
+     * @mongodb.server.release 6.3
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     * @see #getBucketRounding(TimeUnit)
+     * @since 4.10
+     */
+    public TimeSeriesOptions bucketRounding(@Nullable final Long bucketRounding, final TimeUnit timeUnit) {
+        if (bucketRounding == null) {
+            this.bucketRoundingSeconds = null;
+        } else {
+            this.bucketRoundingSeconds = TimeUnit.SECONDS.convert(bucketRounding, timeUnit);
+            isTrueArgument("bucketRounding, after conversion to seconds, must be >= 1", bucketMaxSpanSeconds > 0);
+        }
+        return this;
+    }
+
     @Override
     public String toString() {
         return "TimeSeriesOptions{"
                 + "timeField='" + timeField + '\''
                 + ", metaField='" + metaField + '\''
                 + ", granularity=" + granularity
+                + ", bucketMaxSpanSeconds=" + bucketMaxSpanSeconds
+                + ", bucketRoundingSeconds=" + bucketRoundingSeconds
                 + '}';
     }
 }
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
@@ -1202,6 +1202,12 @@ final class UnifiedCrudHelper {
                 case "metaField":
                     options.metaField(cur.getValue().asString().getValue());
                     break;
+                case "bucketMaxSpanSeconds":
+                    options.bucketMaxSpan(cur.getValue().asInt32().longValue(), TimeUnit.SECONDS);
+                    break;
+                case "bucketRoundingSeconds":
+                    options.bucketRounding(cur.getValue().asInt32().longValue(), TimeUnit.SECONDS);
+                    break;
                 case "granularity":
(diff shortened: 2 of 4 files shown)


driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
```diff
@@ -104,12 +109,102 @@ public TimeSeriesOptions granularity(@Nullable final TimeSeriesGranularity granu
...
+     * @param bucketMaxSpan time span between measurements. After conversion to seconds using {@link TimeUnit#convert(long, java.util.concurrent.TimeUnit)},
+     * the value must be &gt;= 1.
+     * @param timeUnit the time unit.
+     * @return this
+     * @since 4.10
+     * @mongodb.server.release 6.3
+     * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
+     * @see #getBucketMaxSpan(TimeUnit)
+     */
+    public TimeSeriesOptions bucketMaxSpan(@Nullable final Long bucketMaxSpan, final TimeUnit timeUnit) {
```
vbabanin, line 146: As for now, TimeUnit is used in order to maintain consistency within the library since this approach ensures a unified way of handling time intervals throughout the codebase. 

There is a task  [JAVA-4191](https://jira.mongodb.org/browse/JAVA-4191) for introducing Duration for the API, which will enhance the functionality further

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
```diff
@@ -104,12 +109,102 @@ public TimeSeriesOptions granularity(@Nullable final TimeSeriesGranularity granu
...
+
+    /**
+     * Sets the maximum time span between measurements in a bucket.
+     * <p>
+     * The value of {@code bucketMaxSpanSeconds} must be the same as {@code bucketRoundingSeconds}.
+     * If you set the {@code bucketMaxSpanSeconds}, parameter, you can't set the granularity parameter.
+     * </p>
+     *
+     * @param bucketMaxSpan time span between measurements. After conversion to seconds using {@link TimeUnit#convert(long, java.util.concurrent.TimeUnit)},
+     * the value must be &gt;= 1.
```
rozza, line 138: Nit: Also accepts `null` so add something to say a `null` value will unset any previously set value.

rozza (changes requested):
Looks good only a couple of nits - that can be applied to both settings

rozza (approved):
LGTM!

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -1202,6 +1202,12 @@ private TimeSeriesOptions createTimeSeriesOptions(final BsonDocument timeSeriesD
                 case "metaField":
                     options.metaField(cur.getValue().asString().getValue());
                     break;
+                case "bucketMaxSpanSeconds":
+                    options.bucketMaxSpan(cur.getValue().asInt32().longValue(), TimeUnit.SECONDS);
+                    break;
+                case "bucketRoundingSeconds":
+                    options.bucketRounding(cur.getValue().asInt32().longValue(), TimeUnit.SECONDS);
```
stIncMale, line 1209: Why do we do `asInt32` here and above, given that the driver uses `new BsonInt64` in `CreateCollectionOperation.getCreateCollectionCommand`?

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
```diff
@@ -104,12 +109,102 @@ public TimeSeriesOptions granularity(@Nullable final TimeSeriesGranularity granu
         return this;
     }
 
+    /**
+     * Returns the maximum time span between measurements in a bucket.
+     *
+     * @param timeUnit the time unit.
+     * @return time span between measurements, or {@code null} if not set.
+     * @since 4.10
+     * @mongodb.server.release 6.3
```
stIncMale, line 118: Are we sure these options are since 6.3? The unified test specifies `"minServerVersion": "7.0"`.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
stIncMale, line 1: Given that `granularity` is now mutually exclusive with the new properties, let's also modify the documentation of `TimeSeriesGranularity.SECONDS` from

> If granularity of a time-series collection is unspecified, this is the default value.

to simple

> This is the default value.

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
```diff
@@ -104,12 +109,102 @@ public TimeSeriesOptions granularity(@Nullable final TimeSeriesGranularity granu
...
+    /**
+     * Sets the maximum time span between measurements in a bucket.
+     * <p>
+     * The value of {@code bucketMaxSpanSeconds} must be the same as {@code bucketRoundingSeconds}.
+     * If you set the {@code bucketMaxSpanSeconds}, parameter, you can't set the granularity parameter.
+     * </p>
+     *
+     * @param bucketMaxSpan time span between measurements. After conversion to seconds using {@link TimeUnit#convert(long, java.util.concurrent.TimeUnit)},
+     * the value must be &gt;= 1. {@code null} can be provided to unset any previously set value.
+     * @param timeUnit the time unit.
```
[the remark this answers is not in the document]
vbabanin: I want to make sure I understand your request correctly. I believe, the statement `{@code null} can be provided to unset any previously set value` implies that if the argument for `bucketMaxSpan`or `bucketRounding` is set to null, then any previously set value will be unset (meaning that it returns to previous state when it is disregarded). It is implied that all unset values are ignored, it might be worth mentioning at the class level documentation. 

I guess for better clarify we could mention that "By default all unset values are ignored" and "{@code null} can be provided to unset any previously set value."

Let me know if you have any further suggestions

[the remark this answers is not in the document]
stIncMale: I was talking about the value of the `timeUnit` parameter. This argument is ignored if `bucketMaxSpan` is `null`, i.e., when unsetting.

However, maybe it shouldn't be completely ignored, maybe the `null`-check still must be done, given that the parameter is not annotated with `@Nullable`? If we don't have the check, it is quite likely that the methods will be called as `bucketRounding(null, null)`.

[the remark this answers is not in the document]
vbabanin: Oh, i misunderstood initially. Yes, i agree,  adding a null check for the timeUnit parameter is a good idea. Thank you for bringing it up

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
```diff
@@ -104,12 +109,102 @@ public TimeSeriesOptions granularity(@Nullable final TimeSeriesGranularity granu
...
+        if (bucketMaxSpanSeconds == null) {
+            return null;
+        }
+        return timeUnit.convert(bucketMaxSpanSeconds, TimeUnit.SECONDS);
+    }
+
+    /**
+     * Sets the maximum time span between measurements in a bucket.
+     * <p>
+     * The value of {@code bucketMaxSpanSeconds} must be the same as {@code bucketRoundingSeconds}.
```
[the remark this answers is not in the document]
vbabanin: I agree with your viewpoint. However, as mentioned in the [design doc](https://docs.google.com/document/d/1OFq2jyyMLcOkNcnDXaYKXNfdTzDsHufHBAwSp_HDjIo/edit) we have the requirement:

> If a user sets one value, they must set the other. If set directly, we will enforce that the value of bucketRoundingSeconds is equal to the value of bucketMaxSpanSeconds. The latter restriction could be relaxed in the future.

So keeping both options separate allows for flexibility to relax the requirement.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
vbabanin: Since we follow the JSON standard for our spec tests, it's not possible to explicitly indicate whether a numeric value should be interpreted as a `long` or an `int`. In our JSON parser, the value provided in the spec tests is treated as an `UNQUOTED_STRING`, and then the scanned digit is processed as follows:

```
long value = Long.parseLong(lexeme);
if (value < Integer.MIN_VALUE || value > Integer.MAX_VALUE) {
    return new JsonToken(JsonTokenType.INT64, value);
} else {
    return new JsonToken(JsonTokenType.INT32, (int) value);
}
```

You can refer to the code snippet in [JsonScanner.java#L449](https://github.com/vbabanin/mongo-java-driver/blob/d9aa6044e1a6b440bcb013c330497f2813484050/bson/src/main/org/bson/json/JsonScanner.java#L449) for more details.

The type determination is based on the actual value of the digit in the JSON file, which can be either an `int` or a `long`. In the case of the `bucketMaxSpan` and `bucketRounding` spec tests, `int `values are used so far.

If we were to always retrieve the value as an Int64, even when it is an Int32 in the JSON, we would encounter an error: org.bson.BsonInvalidOperationException: Value expected to be of type INT64 is of unexpected type INT32.

Similar cases can be found for maxTimeMs. You can refer to [UnifiedCrudHelper.java#L345](https://github.com/vbabanin/mongo-java-driver/blob/c633293f1dab303425969269bc93b9cc52bef96c/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java#L345) and [FindOperation.java](https://github.com/vbabanin/mongo-java-driver/blob/0efc78c41c63e91da23532695fc31ee1dbd157d0/driver-core/src/main/com/mongodb/internal/operation/FindOperation.java#L414) for additional examples.

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
```diff
@@ -104,12 +109,102 @@ public TimeSeriesOptions granularity(@Nullable final TimeSeriesGranularity granu
...
+     * If you set the {@code bucketMaxSpanSeconds}, parameter, you can't set the granularity parameter.
```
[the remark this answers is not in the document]
vbabanin: Regarding 1, I agree. I will add it accordingly

Regarding 2, I understand convenience for having the `granularity` option erase `bucketMaxSpan` and `bucketRounding` to avoid potential confusion. However, I'm concerned that this approach might lead to unintended side effects for users. There is a possibility that users might accidentally specify `bucketMaxSpan` and then `granularity `when they only intended to set `bucketMaxSpan`. In such cases, if `granularity` silently erases `bucketMaxSpan`, the user's request would be successful on the server side, and they might spend a significant amount of time debugging or remain unaware of the misconfiguration/bug. To prevent this, i would prefer to follow your suggestion of throwing an IllegalStateException when a conflict is detected. This approach would provide immediate feedback to users, clearly indicating the misconfiguration and conflict.

Let me know if you have any further suggestions

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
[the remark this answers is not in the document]
stIncMale: Oh, if this is the requirement at the moment, let's make it more explicit in the docs, because it appears very unintuitive to me. Adding something like _"..., which also means that the options must either be both set or both unset"_ may help.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
stIncMale: Got it, thanks!

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
vbabanin: Thank you for bringing up the concern. According to the documentation at [timeseries-procedures](https://www.mongodb.com/docs/v7.0/core/timeseries/timeseries-procedures/) and the server ticket [SERVER-67598](https://jira.mongodb.org/browse/SERVER-67598), this feature has been available since version 6.3. I have created a corresponding pull request to update the spec tests. You can find it at [Pull Request #1433](https://github.com/mongodb/specifications/pull/1433).

driver-core/src/main/com/mongodb/client/model/TimeSeriesOptions.java
```diff
@@ -32,6 +32,7 @@
  * @mongodb.driver.manual core/timeseries-collections/ Time-series collections
  */
 public final class TimeSeriesOptions {
+    private static final String PARAMETER_TIME_UNIT = "timeUnit";
```
stIncMale, line 35: The names of the `bucketMaxSpan`, `bucketRounding` parameters are duplicated in string literals in argument checks, but the name of the `timeUnit` parameter is static field to avoid duplication. This is not consistent.

P.S. I think, duplicating a parameter name in error messages across methods is always fine.

stIncMale (changes requested):

stIncMale (approved):

PR #1138, opened 2023-06-06
Add second case for the RewrapManyDataKeyOpts prose test.

vbabanin:
This PR adds one missing prose test from the spec.

[JAVA-4766](https://jira.mongodb.org/browse/JAVA-4766)

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/AbstractClientEncryptionRewrapManyDataKeyProseTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/AbstractClientEncryptionRewrapManyDataKeyProseTest.java
@@ -17,6 +17,7 @@
 package com.mongodb.client;
 
 import com.mongodb.ClientEncryptionSettings;
+import com.mongodb.MongoClientException;
 import com.mongodb.MongoClientSettings;
 import com.mongodb.client.model.vault.DataKeyOptions;
 import com.mongodb.client.model.vault.EncryptOptions;
@@ -26,7 +27,10 @@ import com.mongodb.client.vault.ClientEncryption;
 import org.bson.BsonBinary;
 import org.bson.BsonDocument;
 import org.bson.BsonString;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.function.Executable;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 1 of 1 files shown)


driver-sync/src/test/functional/com/mongodb/client/AbstractClientEncryptionRewrapManyDataKeyProseTest.java
```diff
@@ -147,4 +159,26 @@ public void rewrapWithSeparateClientEncryption(final String srcProvider, final S
...
+                .build());
+
+        RewrapManyDataKeyOptions rewrapManyDataKeyOptions = new RewrapManyDataKeyOptions().masterKey(dstKey);
+
+        //when
+        Executable executable = () -> clientEncryption.rewrapManyDataKey(new BsonDocument(), rewrapManyDataKeyOptions);
+
+        //then
+        MongoClientException mongoClientException = assertThrows(MongoClientException.class, executable);
+        assertEquals(expectedErrorMessage, mongoClientException.getMessage());
```
katcharov, line 182: From the usages I see, the assertion message is inlined. I think this is preferable because it limits the scope of the variable (to 0), and moves it closer to the usage.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientEncryptionRewrapManyDataKeyProseTest.java
```diff
@@ -147,4 +159,26 @@ public void rewrapWithSeparateClientEncryption(final String srcProvider, final S
         assertEquals(testString, clientEncryption1.decrypt(ciphertext));
         assertEquals(testString, clientEncryption2.decrypt(ciphertext));
     }
+
+    @Test
+    public void shouldThrowClientErrorWhenProviderIsNotSpecified() {
+        //given
+        String expectedErrorMessage = "Missing the provider but supplied a master key in the RewrapManyDataKeyOptions";
+        BsonDocument dstKey = MASTER_KEYS_BY_PROVIDER.get("aws");
```
katcharov, line 167: The [prose test](https://github.com/mongodb/specifications/commit/4703757119d96fa1438158ece0a30c2186f47800) says "masterKey: {}", but here it seems to be getting the aws key - is this intentional?

katcharov (changes requested):

driver-sync/src/test/functional/com/mongodb/client/AbstractClientEncryptionRewrapManyDataKeyProseTest.java
vbabanin: This test checks the expected behavior when any key is present, so I used a more descriptive dataset that was already available in the class. However, to match the spec more accurately, I updated it to 'masterKey: {}'. Good catch!

katcharov (approved):
LGTM!

PR #1142, opened 2023-06-09
Fix MongoClient leak in auto-encryption

jyemin:
When AutoEncryptionSettings#keyVaultMongoClientSettings is non-null and AutoEncryptionSettings#isBypassAutoEncryption is false, the second internal MongoClient is now closed when the containing MongoClient is closed.

The bug is fixed in both the synchronous and reactive streams drivers.

JAVA-4993

### DIFF
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/Crypt.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/Crypt.java
@@ -64,7 +64,9 @@ public class Crypt implements Closeable {
     private final KeyManagementService keyManagementService;
     private final boolean bypassAutoEncryption;
     @Nullable
-    private final MongoClient internalClient;
+    private final MongoClient collectionInfoRetrieverClient;
+    @Nullable
+    private final MongoClient keyVaultClient;
 
     /**
      * Create an instance to use for explicit encryption and decryption, and data key creation.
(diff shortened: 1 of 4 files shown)


driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/Crypt.java
```diff
@@ -227,8 +232,9 @@ public void close() {
         //noinspection EmptyTryBlock
         try (MongoCrypt ignored = this.mongoCrypt;
              CommandMarker ignored1 = this.commandMarker;
-             MongoClient ignored2 = this.internalClient;
-             KeyManagementService ignored3 = this.keyManagementService
+             MongoClient ignored2 = this.collectionInfoRetrieverClient;
+             MongoClient ignored3 = this.keyVaultClient;
+             KeyManagementService ignored4 = this.keyManagementService
```
stIncMale, line 237: This is not related to the PR changes, but it's not obvious why the async `KeyManagementService` uses `TlsChannelStreamFactoryFactory` (and needs to be closed), while its sync counterpart does not. Do you happen to know anything about it?

stIncMale (approved):

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/Crypt.java
jyemin: I think it's because with async I/O you need a thread pool to handle the I/O completion, and for the library that we're using for async TLS there is no "system" (i.e. global) thread pool that is used (unlike with the JDK's `AsynchrounousSocketChannel`)

stIncMale: Thanks!

PR #1143, opened 2023-06-16
Kotlinx serialization decoding optional ObjectId / BsonValues fails to hydrate properly

rozza:
Found when writing the documentation:

The DefaultBsonDecoder required extra handling to ensure it did not try to double decode optional values and overwrite them with a null value.

Tracking currentIndex and processed status allows the default decoder to correctly track if all elements have been processed and report the correct value when calling decodeElementIndex.

JAVA-5031

### DIFF
--- a/bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecTest.kt
+++ b/bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecTest.kt
@@ -73,6 +73,7 @@ import org.bson.codecs.kotlinx.samples.DataClassWithNestedParameterized
 import org.bson.codecs.kotlinx.samples.DataClassWithNestedParameterizedDataClass
 import org.bson.codecs.kotlinx.samples.DataClassWithNulls
 import org.bson.codecs.kotlinx.samples.DataClassWithObjectIdAndBsonDocument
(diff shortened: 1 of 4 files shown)


bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecTest.kt
```diff
@@ -472,6 +473,52 @@ class KotlinSerializerCodecTest {
         assertRoundTrips(expected, dataClass)
     }
 
+    @Test
+    fun testDataClassWithOptionalObjectIdAndBsonDocument() {
+        val subDocument =
+            """{
+    | "_id": 1,
```
jyemin, line 480: It's not clear to me why it's necessary to test with such a complex subdocument.  What's being tested by this?  The fact that I don't know makes me think I don't understand the issue.

rozza: The subdocument is just copy-pasta.

I'll update the tests and move to a const.

rozza:
Updated the test cases and simplified the decoder.

`decodeElementIndex` now correctly returns all indexes for fields even if they aren't present in the data.

jyemin (approved):

PR #1147, opened 2023-06-22
Kotlinx codec provider limit to data, value or sealed classes.

rozza:
JAVA-5035

### DIFF
--- a/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodec.kt
+++ b/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodec.kt
@@ -85,10 +85,14 @@ private constructor(
             serializersModule: SerializersModule = defaultSerializersModule,
             bsonConfiguration: BsonConfiguration = BsonConfiguration()
         ): Codec<T>? {
-            return try {
-                create(kClass, kClass.serializer(), serializersModule, bsonConfiguration)
-            } catch (exception: SerializationException) {
+            return if (!(kClass.isData || kClass.isValue || kClass.isSealed)) {
                 null
+            } else {
+                try {
+                    create(kClass, kClass.serializer(), serializersModule, bsonConfiguration)
+                } catch (exception: SerializationException) {
+                    null
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodec.kt
```diff
@@ -85,10 +85,14 @@ private constructor(
             serializersModule: SerializersModule = defaultSerializersModule,
             bsonConfiguration: BsonConfiguration = BsonConfiguration()
         ): Codec<T>? {
-            return try {
-                create(kClass, kClass.serializer(), serializersModule, bsonConfiguration)
-            } catch (exception: SerializationException) {
+            return if (!(kClass.isData || kClass.isValue || kClass.isSealed)) {
```
jyemin, line 88: Which of the assertions below test which of the tests here?  It seems that at least one test is missing since there are three conditions here and two assertions.

rozza: Ah that test ensures the two cases (double & CharSequence) that we knew were being picked up now return null.

All three types are already tested by the other tests.

jyemin: It seems like before this change this method was returning a codec for classes that were not data or value or sealed classes, and now it returns null for those.  So I'm not clear how an assertion that it now returns null for all of these could already be in the code.

rozza: Ah - `KotlinSerializerCodecTest` has existing tests for data, value and sealed classes which we want to continue to support.

The `shouldReturnNullForSerializableButNotValueClassOrSealedOrDataClassTypes` adds a test to ensure other serializable types (that we know about) fail.

jyemin: Oh, I'm dumb.  Since it's a ! for all three then this is fine.

jyemin (approved):

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodec.kt
rozza: Merged as is in 4.10.x branch

But added 28b9727ee7d38001c9cec2ce0970765750f12e33 to the master branch to improve the readability of this check and the same for the DataclassCodec.

PR #1148, opened 2023-06-22
Build: Make bson-kotlin and bson-kotlinx optional

rozza:
Reduce the dependency tree for driver-core by making the kotlin dependencies optional.

Kotlin-coroutine and Kotlin-sync now take a hard dependency on bson-kotlin. So they will work out the box. Users will now have to opt into bson-kotlinx support.

JAVA-5036

### DIFF
--- a/driver-core/build.gradle
+++ b/driver-core/build.gradle
@@ -37,8 +37,8 @@ def classifiers = ["linux-x86_64", "linux-aarch_64", "osx-x86_64", "osx-aarch_64
 dependencies {
     api project(path: ':bson', configuration: 'default')
     implementation project(path: ':bson-record-codec', configuration: 'default')
-    implementation project(path: ':bson-kotlin', configuration: 'default')
-    implementation project(path: ':bson-kotlinx', configuration: 'default')
+    implementation project(path: ':bson-kotlin', configuration: 'default'), optional
+    implementation project(path: ':bson-kotlinx', configuration: 'default'), optional
 
     implementation "com.github.jnr:jnr-unixsocket:$jnrUnixsocketVersion", optional
     api platform("io.netty:netty-bom:$nettyVersion")
--- a/driver-kotlin-coroutine/build.gradle.kts
+++ b/driver-kotlin-coroutine/build.gradle.kts
@@ -65,6 +65,7 @@ dependencies {
 
     api(project(path = ":bson", configuration = "default"))
     api(project(path = ":driver-reactive-streams", configuration = "default"))
+    implementation(project(path = ":bson-kotlin", configuration = "default"))
 
     testImplementation("org.jetbrains.kotlin:kotlin-reflect")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit")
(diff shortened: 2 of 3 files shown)


driver-kotlin-coroutine/build.gradle.kts
```diff
@@ -65,6 +65,7 @@ dependencies {
 
     api(project(path = ":bson", configuration = "default"))
     api(project(path = ":driver-reactive-streams", configuration = "default"))
+    implementation(project(path = ":bson-kotlin", configuration = "default"))
```
rozza, line 68: Adds a direct dependency - should we also add kotlinx here as well?

driver-kotlin-sync/build.gradle.kts
```diff
@@ -19,7 +19,6 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 plugins {
     id("org.jetbrains.kotlin.jvm")
     `java-library`
-    kotlin("plugin.serialization")
```
rozza, line 22: This wasn't used or needed - same goes for bson-kotlinx in the test.

driver-core/build.gradle
```diff
@@ -37,8 +37,8 @@ def classifiers = ["linux-x86_64", "linux-aarch_64", "osx-x86_64", "osx-aarch_64
 dependencies {
     api project(path: ':bson', configuration: 'default')
     implementation project(path: ':bson-record-codec', configuration: 'default')
-    implementation project(path: ':bson-kotlin', configuration: 'default')
-    implementation project(path: ':bson-kotlinx', configuration: 'default')
+    implementation project(path: ':bson-kotlin', configuration: 'default'), optional
```
jyemin, line 40: If we just removed these what would happen? Would the build fail?

rozza: The driver-core `KotlinCodecProvider` is included as part of the default registry.

driver-kotlin-coroutine/build.gradle.kts
rozza: With just bson-kotlin:

Pros: Minimizes the dependencies - doesn't drag in kotlinx serialization.
Cons: bson-kotlinx becomes an extra opt in requirement.

```
implementation("org.mongodb:mongodb-driver-kotlin-coroutine:4.10.0")
implementation("org.mongodb:bson-kotlinx:4.10.0")
```

I think this PR is the best course of action - the automatic extra dependency on bson-kotlin is small. So don't think we should add kotlinx automagically.

driver-core/build.gradle
jyemin: Right: that class does canary tests for `KotlinSerializerCodecProvider` and `DataClassCodecProvider`.  So optional seems appropriate.

driver-kotlin-coroutine/build.gradle.kts
jyemin: Yeah, I'm just not sure how widely kotlinx serialization will be used.  I guess we can always make it non-optional in the future.  We may need a documentation update for this.

jyemin (approved):
We should document this is a breaking change, even though it's doubtful anyone will be affected by it.

PR #1152, opened 2023-06-23
Kotlin: Improve KotlinSerializerCodec

rozza:
Only support classes annotated as serializable.

JAVA-5035


rozza:
This way we still support any explicitly declared classes that are Serializable.

The previous check for `kClass.isSealed` also picked up Java Sealed interfaces and would error.

Note: Java records didn't return true from the previous check (kClass.isData || kClass.isValue)

rozza:
Other [approaches ](https://github.com/search?q=+isKotlinClass+&type=code)check for the presence of the kotlin `Metadata` annotation. However, we only want `@Serializable` kotlin classes.

jyemin (approved):

PR #1154, opened 2023-06-30
gRPC POC phase 1

stIncMale:
`Java5018.java` contains a smoke test and instructions on how to set up and start atlasproxy. The core code where we integrate with Java gRPC is in the following classes:

- `SharingGrpcStreamFactoryFactory.Channels`
- `GrpcStream`

**Limitations**

1. Only synchronous API is implemented.
2. No mechanism is implemented to detect in the following places whether gRPC is being used. Consequently, some tests fail, and this codebase should not be used without gRPC if performance matters.
2.1. `QueryBatchCursor` and `AsyncQueryBatchCursor` always pin the connection that was used to create the corresponding server cursor.
2.2. `TransactionContext.isConnectionPinningRequired` always returns `true`.
2.3. `com.mongodb.client.internal`/`com.mongodb.reactivestreams.client.internal.ClientSessionBinding.isConnectionSourcePinningRequired` always returns `true`.
3. While an attempt to support TLS was made, there is no way to test it because atlasproxy does not currently support gRPC with TLS.
4. Authentication over gRPC is not implemented because atlasproxy does not support it.
5. Closing a `MongoClient` does not release Java gRPC channels it created. Consequently, this codebase should not be used if many `MongoClient` instances are created and closed.

JAVA-5018

### DIFF
--- a/driver-core/build.gradle
+++ b/driver-core/build.gradle
@@ -46,6 +46,10 @@ dependencies {
     api "io.netty:netty-transport", optional
     api "io.netty:netty-handler", optional
 
+    implementation 'io.grpc:grpc-netty:1.56.0', optional
+    // required by Java gRPC for Java SE 9+
+    compileOnly 'org.apache.tomcat:annotations-api:6.0.53', optional
+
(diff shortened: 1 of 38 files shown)


jyemin:
Might as well close this...

PR #1157, opened 2023-07-05
Do various doc and other minor improvements

stIncMale:
Various things I noticed while working on JAVA-5018.


driver-core/src/main/com/mongodb/internal/connection/MessageSettings.java
```diff
@@ -27,8 +27,20 @@
...
+    /**
+     * <a href="https://www.mongodb.com/docs/current/reference/command/hello/#mongodb-data-hello.maxMessageSizeBytes">
+     * {@code maxMessageSizeBytes}</a>.
+     */
     private static final int DEFAULT_MAX_MESSAGE_SIZE = 0x2000000;   // 32MB
+    /**
+     * <a href="https://www.mongodb.com/docs/current/reference/command/hello/#mongodb-data-hello.maxWriteBatchSize">
+     * {@code maxWriteBatchSize}</a>.
+     */
     private static final int DEFAULT_MAX_BATCH_COUNT = 1000;
```
stIncMale, line 44: Is my understanding (links to the docs) correct here? I am specifically not sure about `DEFAULT_MAX_BATCH_COUNT`: we use 1000, while the MongoDB docs for [`maxWriteBatchSize`](https://www.mongodb.com/docs/current/reference/command/hello/#mongodb-data-hello.maxWriteBatchSize) says `100_000`.

rozza: MongoDB 3.4 and previous had 1,000. In 3.6 it went up to 100,000: https://www.mongodb.com/docs/v4.0/reference/command/isMaster/#isMaster.maxWriteBatchSize

stIncMale: 👍

jyemin: Right, this is just the driver default, if `hello` doesn't provide one.  In practice it always does now.

driver-core/src/main/com/mongodb/connection/Stream.java
```diff
@@ -46,7 +46,7 @@ public interface Stream extends BufferProvider{
     /**
      * Write each buffer in the list to the stream in order, blocking until all are completely written.
      *
-     * @param buffers the buffers to write
+     * @param buffers the buffers to write. This method must {@linkplain ByteBuf#release() release} each buffer when it no longer needs it.
```
jyemin, line 49: Is this true?  `SocketStream` does not seem to do it.

driver-core/src/main/com/mongodb/connection/Stream.java
```diff
@@ -100,7 +100,7 @@ default ByteBuf read(int numBytes, int additionalTimeout) throws IOException {
      * Write each buffer in the list to the stream in order, asynchronously.  This method should return immediately, and invoke the given
      * callback on completion.
      *
-     * @param buffers the buffers to write
+     * @param buffers the buffers to write. This method must {@linkplain ByteBuf#release() release} each buffer when it no longer needs it.
```
jyemin, line 103: Same as above.  `AsynchronousChannelStream` doesn't seem to do it.

driver-core/src/main/com/mongodb/connection/Stream.java
jyemin: It looks like `ByteBufferBsonOutput#close`, called from `InternalStreamConnection#sendCommandMessage`, is responsible for the release.

jyemin (changes requested):

driver-core/src/main/com/mongodb/connection/Stream.java
stIncMale: This statement is true, and the apparent deviation of this requirement that you see in `SocketStream`/`AsynchronousChannelStream.write` is a result of two peculiarities (I am being generous here):

1. Neither `SocketStream.write` nor `AsynchronousChannelStream.write` use the buffers they were given after completing (either returning in the synchronous case, or completing the handler in the asynchronous case). So even if the buffers stay unreleased by `write`, the method does not leave a reference to them in the heap memory. This, in turn, means that there may neither be negative effects of aliasing, nor memory leaks caused by `write` not releasing the buffers.
2. `SocketStream`/`AsynchronousChannelStream.write` receive `PooledByteBufNIO` of `ByteBuf`, backed by `ByteBufNIO`, or (I am not sure) `CompositeByteBuf`. All of these `ByteBuf` implementations implement the `duplicate` method by creating a new wrapping `ByteBuf` whose reference count is unrelated to the wrapped buffer: `duplicate()` does not retain the wrapper, and `duplicate().release()` does not release it. Such implementations are, of course, a recipe for disaster, because it allows the wrapped buffer to be released despite the wrapping buffer still having positive reference count, which may result in the wrapped buffer being aliased.

TL;DR: `SocketStream`/`AsynchronousChannelStream.write` do not release buffers, but releasing the buffers they are given is completely pointless anyway due to how they are implemented. Our code here is brittle and seemingly works only due to a lucky coincidence.

The situation with the `NettyByteBuf` implementation of `ByteBuf` and `NettyStream` is different: `NettyByteBuf.duplicate` does retain the wrapped buffer and the wrapper delegates its `retain`/`release`/`getReferenceCount` to the wrapped buffer. These two things (retaining when wrapping + delegation) guarantees that the wrapper's lifetime is not larger than the wrapped's lifetime. And `NettyStream.write` actually releases the buffers: [_"outbound messages are created by your application, and it is the responsibility of Netty to release these after writing them out to the wire"_](https://netty.io/wiki/reference-counted-objects.html#wiki-h3-9).

So, the documentation change for the `Stream.write` message is correct. `SocketStream`/`AsynchronousChannelStream` violate it and get away with that only because `ByteBufNIO` and `CompositeByteBuf` are also broken. However, the behavior of `ByteBufNIO`/`CompositeByteBuf.duplicate` means that I need to fix the doc change for the `ByteBufferBsonOutput.getByteBuffers` method (see the new commit).

rozza (approved):
LGTM

driver-core/src/main/com/mongodb/connection/Stream.java
jyemin: I'm still not so sure.  Generally when a reference counted object is passed as a parameter, my assumption is that it's the caller's responsibility to release its own reference to to it when the call completes.  The receiving object should only release if it also retains, and it doesn't need to retain unless it keeps it for longer than the completion of the call.

For "async" calls it's a bit different in that the above applies not until the async call returns, but until the callback is invoked.

So to my eyes the bug is that `InternalStreamConnection` calls `ByteBufferBsonOutput#getByteBuffers` and receives a list of `List<ByteBuf>` that it, in come cases, does not release.

jyemin (changes requested):
Discussed offline.

driver-core/src/main/com/mongodb/connection/Stream.java
stIncMale: I rolled back changes in `Stream` and created https://jira.mongodb.org/browse/JAVA-5084.

bson/src/main/org/bson/ByteBuf.java
```diff
@@ -434,7 +434,7 @@ public interface ByteBuf  {
     ByteBuffer asNIO();
 
     /**
-     * Gets the current reference count, which starts at 0.
+     * Gets the current reference count, which starts at 1.
      *
      * @return the current count, which must be greater than or equal to 0
```
katcharov, line 439: The "@return ...gte 0" contradicts "starts at 1"?

stIncMale: It's not: it's 0 for a released `ByteBuf`.

stIncMale: Changed the wording to
> Gets the current reference count, which is 1 for a new {@link ByteBuf}.

katcharov: LGTM (did not review otherwise, just noticed this while using git; so will remove self as reviewer).

jyemin (approved):

PR #1159, opened 2023-07-13
Support $changeStreamSplitLargeEvent

katcharov:
JAVA-4955

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
+++ b/driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
@@ -59,7 +59,7 @@ public final class ChangeStreamDocument<TDocument> {
     private final BsonInt64 txnNumber;
     private final BsonDocument lsid;
     private final BsonDateTime wallTime;
-    private final BsonDocument splitEvent;
+    private final SplitEvent splitEvent;
     @BsonExtraElements
     private final BsonDocument extraElements;
 
@@ -96,7 +96,7 @@ public final class ChangeStreamDocument<TDocument> {
             @Nullable @BsonProperty("txnNumber") final BsonInt64 txnNumber,
             @Nullable @BsonProperty("lsid") final BsonDocument lsid,
             @Nullable @BsonProperty("wallTime") final BsonDateTime wallTime,
-            @Nullable @BsonProperty("splitEvent") final BsonDocument splitEvent,
+            @Nullable @BsonProperty("splitEvent") final SplitEvent splitEvent,
             @Nullable @BsonProperty final BsonDocument extraElements) {
         this.resumeToken = resumeToken;
         this.namespaceDocument = namespaceDocument;
@@ -471,18 +471,15 @@ public final class ChangeStreamDocument<TDocument> {
         return wallTime;
     }
 
-
     /**
-     * When the change stream's backing aggregation pipeline contains the
-     * <code>$changeStreamSplitLargeEvent</code> stage, events larger than 16MB
-     * will be split into multiple events. This split event information takes
-     * the form: <code>{ fragment: &lt;current>, of: &lt;total> }</code>.
+     * The split event.
      *
      * @return the split event
      * @since 4.11
      * @mongodb.server.release 7.0
      */
-    public BsonDocument getSplitEvent() {
+    @Nullable
+    public SplitEvent getSplitEvent() {
         return splitEvent;
     }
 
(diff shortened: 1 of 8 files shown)


driver-sync/src/test/functional/com/mongodb/client/ChangeStreamProseTest.java
```diff
@@ -401,6 +405,40 @@ public void testGetResumeTokenReturnsIdOfPreviousDocument() {
...
+                        new ChangeStreamPreAndPostImagesOptions(true)));
+
+        // #. Insert into _C_ a document at least 10mb in size ...
+        int mb10 = 10 * 1024 * 1024;
+        collection.insertOne(new Document("value", new BsonString(repeat(mb10, "q"))));
+
+        // #. Create a change stream _S_ by calling ``watch`` on _C_ with
+        //    pipeline ``[{ "$changeStreamSplitLargeEvent": {} }]`` and ``fullDocumentBeforeChange=required``.
+        ChangeStreamIterable<Document> changeStream = collection
+                .watch(asList(Document.parse("{ $changeStreamSplitLargeEvent: {} }")))
```
katcharov, line 424: I did not create a stage builder, since we do not seem to offer stages for other stages like this (for example, $currentOp).

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -430,6 +471,21 @@ public BsonDateTime getWallTime() {
...
+     * <code>$changeStreamSplitLargeEvent</code> stage, events larger than 16MB
+     * will be split into multiple events. This split event information takes
+     * the form: <code>{ fragment: &lt;current>, of: &lt;total> }</code>.
+     *
+     * @return the split event
+     * @since 4.11
+     * @mongodb.server.release 7.0
+     */
+    public BsonDocument getSplitEvent() {
+        return splitEvent;
```
katcharov, line 486: This could return a proper class, similar to `UpdateDescription`, but it produces a `BsonDocument`, like `lsid`. I do not see a clear way to decide this. I chose this because it seemed uncommon.

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -108,9 +111,47 @@ public ChangeStreamDocument(@BsonProperty("operationType") final String operatio
...
+     * @param updateDescription the update description
+     * @param txnNumber the transaction number
+     * @param lsid the identifier for the session associated with the transaction
+     * @param wallTime the wall time of the server at the moment the change occurred
+     * @param extraElements any extra elements that are part of the change stream document but not otherwise mapped to fields
+     *
+     * @since 4.7
+     */
+    @Deprecated
+    public ChangeStreamDocument(@BsonProperty("operationType") final String operationTypeString,
```
katcharov, line 138: This is an unfortunate proliferation of constructors, but it seems in line with what was done previously in this class.

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
stIncMale: `lsid` is not supposed to be deconstructed by a user, while deconstructing `splitEvent` is pretty much a requirement if a user wants to meaningfully use change events that are split into fragments (though, I have no idea if the MongoDB Kafka Connector will have any automatic handling for those in the future, and I understand even less about the MongoDB Spark Connector). It seems to me that giving users a convenient type to work with makes all the sense here.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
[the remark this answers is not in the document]
stIncMale: Based on https://mongodb.slack.com/archives/C72LB5RPV/p1689609952334209, it looks like the work required to start supporting `$changeStreamSplitLargeEvent` in drivers was not done, even though it was supposed to be.

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -430,6 +471,21 @@ public BsonDateTime getWallTime() {
         return wallTime;
     }
 
+
+    /**
+     * When the change stream's backing aggregation pipeline contains the
+     * <code>$changeStreamSplitLargeEvent</code> stage, events larger than 16MB
+     * will be split into multiple events. This split event information takes
+     * the form: <code>{ fragment: &lt;current>, of: &lt;total> }</code>.
```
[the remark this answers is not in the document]
katcharov: Added to SplitEvent class docs

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -103,14 +107,52 @@ public ChangeStreamDocument(@BsonProperty("operationType") final String operatio
         this.fullDocument = fullDocument;
         this.clusterTime = clusterTime;
         this.operationTypeString = operationTypeString;
-        this.operationType = OperationType.fromString(operationTypeString);
+        this.operationType = operationTypeString == null ? null : OperationType.fromString(operationTypeString);
```
katcharov, line 110: fromString does not take null; note that it is used in the OperationTypeCodec

stIncMale: I am glad we are not passing `null` to `OperationType.fromString`, even though it (arguably, incorrectly) has the code to handle `null`s.

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -380,6 +422,7 @@ public String getOperationTypeString() {
      *
      * @return the operationType
      */
+    @Nullable
```
stIncMale, line 425: 1. We need `@Nullable` not only on `getOperationType`, but also on `getOperationTypeString`.
2. Let's document on both methods that they may return `null` only if `$changeStreamSplitLargeEvent` is used.

driver-core/src/main/com/mongodb/client/model/changestream/SplitEvent.java
stIncMale, line 1: Let's modify `ChangeStreamDocumentCodecSpecification.'should round trip ChangeStreamDocument successfully'` with at least one case for `SplitEvent`.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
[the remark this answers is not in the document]
stIncMale: Should I resolve this, or is there something you are still waiting for?

stIncMale (approved):

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
vbabanin: Should we include a "@deprecated" tag to specify the reason for deprecation?

driver-core/src/main/com/mongodb/client/model/changestream/SplitEvent.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Objects;
+
+/**
+ * The current fragment, out of the total number of fragments.
+ * When the change stream's backing aggregation pipeline contains the
+ * <code>$changeStreamSplitLargeEvent</code> stage, events larger than 16MB
+ * will be split into multiple events.
+ *
+ * @since 4.11
+ * @mongodb.server.release 7.0
```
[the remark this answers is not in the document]
katcharov: It was apparently backported to 6.0.9, but the fix version is 7 https://jira.mongodb.org/browse/SERVER-55062 . I will leave 7, since it was on the Jira for this ticket, but we can check with the team.

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
katcharov: I added the following just now,
```
     * @deprecated Prefer a non-deprecated constructor
```
but when I read it, it seemed too obvious, like "@deprecated use something that isn't deprecated". I think one convention (apart from not including the tag) is to link to the preferred replacement, but here, I do not think we should do this since there have already been N replacements, so it seemed like a link would become outdated.

If you still think this is worth including, I don't feel too strongly, but let me know the wording.

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -430,6 +476,18 @@ public BsonDateTime getWallTime() {
         return wallTime;
     }
 
+    /**
+     * The split event.
+     *
+     * @return the split event
+     * @since 4.11
+     * @mongodb.server.release 7.0
```
vbabanin, line 484: Since SplitEvent works with 6.0.9 and the documentation has been updated for it, let's change it to 6.0.9 as well

vbabanin (changes requested):

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
katcharov: Good catch

vbabanin (approved):

PR #1164, opened 2023-08-01
Add `TimePoint` and use it in `Timeout`

stIncMale:
This PR introduces internal API needed for JAVA-5076 (see https://github.com/mongodb/mongo-java-driver/pull/1166). I separated the change in this PR to make https://github.com/mongodb/mongo-java-driver/pull/1166 more focused on business logic.

The internal API introduced in the PR is intended to solve the following problems:

1. I need to measure elapsed time. `System.nanoTime` is such a tool, but it would have been better to have a tool that exposes types less generic than `long`, and values that are meaningful on their own (a value returned by `System.nanoTime` is meaningless on its own).
2. I need to measure elapsed time from the same instant that a `Timeout` was started at.


driver-core/src/main/com/mongodb/internal/time/Timeout.java
```diff
@@ -13,164 +13,119 @@
...
-     * otherwise an object that represents the specified {@code durationNanos}.
+     * otherwise a timeout of {@code durationNanos}.
      * <p>
      * Note that the contract of this method is also used in some places to specify the behavior of methods that accept
      * {@code (long timeout, TimeUnit unit)}, e.g., {@link com.mongodb.internal.connection.ConcurrentPool#get(long, TimeUnit)},
-     * so it cannot be changed without updating those methods.
+     * so it cannot be changed without updating those methods.</p>
      */
-    public static Timeout startNow(final long durationNanos) {
+    public static Timeout started(final long durationNanos, final Timer timer) {
```
stIncMale, line 72: Accepting `Timer` allows

- having multiple timeouts that start at the same instant
- [more importantly for `JAVA-5076`] having a timer and a timeout started at the same instant.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
```diff
@@ -13,164 +13,119 @@
...
-     *             in which case the timeout has expired.</li>
-     *         </ol>
-     *         Since it is impossible to differentiate the former from the latter, and the former is much more likely to happen in practice,
-     *         this class interprets a positive value as the exact number of elapsed nanoseconds.
-     *     </li>
-     * </ul>
-     * <hr>
-     * <sup>(*)</sup> n is positive and odd.
-     */
-    private long elapsedNanos(final long currentNanos) {
```
stIncMale, line 135: This method went to `Timer`.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
katcharov: I understand that Timeout is a class that can be "started" only at the time that `startNow` is called, and it times out ("expires") at a certain future time. This appears to be adequate for the above requirements, so I am unsure why Timer is needed. The name "Timer" is also confusing... it does not seem to "time" anything, but is a wrapper around a start time, making it seem more like an Instant. It is good that this PR split from the other PR, but since it is split, I may be misunderstanding the requirements:

* Timeout itself should be a value class, so I think this means that you need multiple timeouts with the same startNanos, but different durationNanos. It seems this can be better accomplished by introducing a `withDuration` method, that returns a class with the same startNanos, but a different duration. The implementation of INFINITE and IMMEDIATE should be changed: I do not think that they are correct as they are. It seems they should still record the correct startTime. For example, the isImmediate method should return true when any timer is started with a duration of 0, but it presently checks for equality against IMMEDIATE. I think perhaps 0 and MAX_LONG should be used to represent immediate and infinite, and that there should be no Timeout constants.
* It appears from the other PR that the only reason to have a timer and a timeout started at the same time is to [get the elapsed time.](https://github.com/mongodb/mongo-java-driver/pull/1166/files#diff-7f8dc15248ba3fe10795342e55c0525fee969cc73d918141a5da85dbded487fbR274) I think this can be handled by getting the elapsed time from a Timeout directly. If there is some other requirement, could you please expand on this?

driver-core/src/main/com/mongodb/internal/time/Timer.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * A <a href="https://docs.oracle.com/javase/8/docs/api/java/lang/doc-files/ValueBased.html">value-based</a> class
+ * for tracking elapsed time. The maximum duration this class can measure reliably is {@link Long#MAX_VALUE} nanoseconds,
+ * which is more than 292 years. The class must not be used to measure longer durations.
+ * <p>
+ * This class is not part of the public API and may be removed or changed at any time.</p>
+ */
+@Immutable
+public final class Timer {
+    private static final Timer USELESS = new Timer(0);
```
katcharov, line 36: This naming seems unclear.

katcharov (changes requested):
I am unsure about the approach - I did not review in detail, pending the main comment above.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
stIncMale: There are many things packed in the above comment, and I am not sure most of them leads to something. I tried replying to all of them and grouping replies in numbered sections.

## 1. I am unsure why `Timer` is needed

It is needed for "for tracking elapsed time", as documented. Since it is now needed not only for implementing `Timeout`, but also on its own, the functionality was extracted from internals of `Timeout` into a separate class.

> "Timer" ... is a wrapper around a start time

It is important to point out that the "start time" value is practically meaningless, and only the difference between two such values is meaningful. Both the Go and Rust standard libraries provide tools for measuring elapsed time that hide such meaningless "start time" values. Unfortunately, the Java SE API  does not do that and instead exposes `System.nanoTime`. This is most likely because Java did not have a way to make such an abstraction zero-cost at the time the API was designed (nor does it have such a way now, but it will when value objects and null-restricted types are implemented). The Java SE API does have to be concerned with use cases like microbenchmarks, but we do not, and we can afford to introduce the abstraction.

> ... need multiple timeouts with the same startNanos, but different durationNanos. It seems this can be better accomplished by introducing a `withDuration` method, ...

This is an alternative approach to the current API (it also eliminates the need to have the `duration` parameter in the `Timeout.started` methods), but I don't think it is better.

> ... get the elapsed time. I think this can be handled by getting the elapsed time from a Timeout directly.

It is true that technically the `Timeout` class could provide API for measuring elapsed time, it could similarly provide any other API that is not needed to achieve the purpose of `Timeout`. But I think that it is better to have a separate tool for this job.

Just in case you are surprised that given the above, there are `Timeout.remaining`/`remainingOrInfinite` methods - they are an unfortunate byproduct of us having to integrate with external APIs that accept a timeout in a form different from our `Timeout`. While an approach that does not expose `Timeout.remaining`/`remainingOrInfinite` methods is possible, it requires separate `Timeout` methods for integrating with each external API, and results in code that does not look like usual code that uses those external APIs.

## 2. The name "Timer" is also confusing... it does not seem to "time" anything

I used this name because this is how you called a [similar class](https://github.com/mongodb/mongo-kafka/commit/c25e0526bfd43c619b2e06a2739b7c525a7af9ec#diff-136a2047e289c24f10ac62956d07711230e8e438a2680d9be51fbfc4770b58f4R19) in MongoDB Kafka connector, so I am surprised there is now a concern about the name. However, **I agree that it is not a "timer" but a stopwatch, and I am happy to rename it this way if you agree.**

## 3. `Timer` seem more like an `Instant`

Seeing `Timer` not as a stopwatch, but as a point on a timeline with the origin somewhere within the time interval in which the `Timer.start` method was running, is a valid way of interpreting it. In fact, both the Go and Rust standard libraries do exactly this: in Go the type that allows measuring elapsed time is called `Time`<sup>1</sup>, and in Rust the corresponding struct is called `Instant` (despite having the same name, it is not semantically similar to `Instant` from the Java SE API). **I will be happy to use this interpretation, if this is what you propose** (this approach in conflict with item 2 above: only one may be chosen). **If so, do you have a name for the class better than `Time`?**

## The implementation of INFINITE and IMMEDIATE should be changed: I do not think that they are correct

> For example, the isImmediate method should return true when any timer is started with a duration of 0

What you described, is how `isImmediate` behaves. See `assertEquals(Timeout.immediate(), Timeout.started(0, timer)` in `TimeoutTest.started`.

## 4. 0 and MAX_LONG should be used to represent immediate and infinite

0 does represent an immediate timeout, and `Long.MAX_VALUE` does represent an infinite one.

## 5. there should be no Timeout constants

This PR does not introduce the `Timeout.immediate`/`infinite` method, but I know why having the them is convenient: they are useful in situations when an immediate/infinite timeout is statically known to be needed; for `Timeout.immediate()` we already have such a place in the codebase. I don't know what substantiates the opinion that `Timeout.immediate`/`infinite` methods should not exist.

---

<sup>1</sup> `Time` in the Go standard library combines the functionality of a monotonic clock with the functionality of a wall clock (a.k.a. system clock). I neither understand the reasoning behind such an API design, nor how the API is implemented.

driver-core/src/main/com/mongodb/internal/time/Timer.java
stIncMale: `USELESS` is what will likely be called a default value when [`null`-restricted types](https://openjdk.org/jeps/401) are implemented. Values of these types may have a heap layout that is even more optimal than that of [value objects](https://openjdk.org/jeps/8277163) (essentially, on par with the layout in C/C++/Rust/etc.).

However, since `USELESS` is confusing, I will use `null` as `Timeout.timer` in situations when the value of this field is not needed. I am postponing the change until the concerns in https://github.com/mongodb/mongo-java-driver/pull/1164#discussion_r1289051261 are resolved.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
stIncMale: I have realized that the interpretation I described in section 3

> ... a point on a timeline with the origin somewhere within the time interval in which the `Timer.start` method was running, ...

was weaker than what we have in reality, as it did not allow comparing different points, while we can, in fact, compare them. I updated it by making the description stronger, such that it allows for point comparison.

stIncMale: After thinking more about the interpretation in section 2 vs. interpretation in section 3, I find the latter more useful and convenient, so I will just make changes accordingly.

stIncMale: Of all the standard libraries for languages for which there are official MongoDB drivers, only the following provide an abstraction (either the one described in section 2, or in section 3 above):

| Language | Relevant type name and link | Abstraction, section 2 or 3 |
| --- | --- | --- |
| C# | [`Stopwatch`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch?view=net-7.0) | section 2 |
| C++ | [`time_point`](https://en.cppreference.com/w/cpp/chrono/steady_clock) | section 3 |
| Kotlin | [`ValueTimeMark`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/-time-source/-monotonic/-value-time-mark/) | section 3 |
| Swift | [`Instant`](https://developer.apple.com/documentation/swift/continuousclock#structures) | section 3 |
| Go | [`Time`](https://pkg.go.dev/time#Time) | section 3 |
| Rust | [`Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html) | section 3 |

The approach described in section 3 is much more popular, with `Instant` being the most popular name. Since it is already taken by a class in the Java SE API, I picked `TimePoint` from C++. `Time` from Go does not feel specific enough, and `ValueTimeMark` form Kotlin feels like a fruit of a diseased imagination.

driver-core/src/main/com/mongodb/internal/time/Timer.java
stIncMale: I refactored `Timer` into `TimePoint`, and started using `null` instead of `Timer.useless`.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
katcharov: This PR sets out how we should time things in the driver, and will affect more than the linked PR. We seem to have the following requirements:

* Determine how much time has `elapsed` since a certain past time
* Determine whether an originally-future point in time has passed (timed out, `expired`)...
* ...and also how much time is `remaining` before that point (it seems unimportant whether arises from an internal or external need, it's still something we must support and represent)

Currently, `Timeout` covers `expired`/`remaining`. This PR currently proposes `TimePoint`, to cover `elapsed`. This introduces a new class, and an interaction with the existing class (Timeouts must be started using TimePoints). If we take this approach, I think that we should preserve the existing `startNow` methods. They are at least fine convenience methods, and in any case there seems to be no need to couple the classes together in this way.

A Timeout appears to be just a TimePoint that is originally intended to be in the future, though it might eventually (or immediately) be in the past. It seems unnecessary to implement this same concept using two different classes. This seems to be the approach taken by Kotlin's ValueTimeMark. If we have a strong justification for needing to differentiate between a TimePoint that tracks elapsed time, and a TimePoint that tracks remaining time, we could use interfaces to hide the appropriate methods.

stIncMale: I think that the idea to replace timeouts with deadlines, which can be represented as `TimePoint`s is really great!👌

- We will need `TimePoint.infinity()` for the deadline corresponding to `Timeout.infinite()`, and it can be encoded with `TimePoint.nanos` being `null`.
- Despite there being no special value left (`null` is the only special value we have in this approach) for encoding a ingle deadline for `Timeout.immediate()`, every result of `TimePoint.now()` is such a deadline.

I'll implement this approach here.

stIncMale: > I'll implement this approach here.

I jumped the gun there. After discussing with Jeff whether I should do such refactoring now, or let the team triage the idea, I created https://jira.mongodb.org/browse/JAVA-5112. So if we start using deadlines instead of timeouts internally, that will be a PR for the created task.

stIncMale: > we should preserve the existing `startNow` methods

Done.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
```diff
@@ -29,148 +31,120 @@
...
-     * Must not be called on {@linkplain #isInfinite() infinite} timeouts.
+     *
+     * @throws AssertionError If the timeout is {@linkplain #isInfinite() infinite} or {@linkplain #isImmediate() immediate}.
      */
     @VisibleForTesting(otherwise = PRIVATE)
-    long remainingNanos(final long currentNanos) {
-        assertFalse(isInfinite() || isImmediate());
-        long elapsedNanos = elapsedNanos(currentNanos);
-        return elapsedNanos < 0 ? 0 : Math.max(0, durationNanos - elapsedNanos);
+    long saturatingRemainingNanos(final TimePoint now) {
```
katcharov, line 124: [optional] The term "saturating" was confusing for me. I had to look it up, and I think it refers to [this](https://en.wikipedia.org/wiki/Saturation_arithmetic). I think the implication is that it is "clamped" to 0, but I think this is already implied by the term "remaining".

(But, if a term should be used, I suspect that it should also be used for the `remaining` method below?)

driver-core/src/main/com/mongodb/internal/time/TimePoint.java
```diff
@@ -0,0 +1,111 @@
...
+    static TimePoint at(final long nanos) {
+        return new TimePoint(nanos);
+    }
+
+    /**
+     * The {@link Duration} between this {@link TimePoint} and {@code t}.
+     * A {@linkplain Duration#isNegative() negative} {@link Duration} means that
+     * this {@link TimePoint} is {@linkplain #compareTo(TimePoint) before} {@code t}.
+     */
+    public Duration durationSince(final TimePoint t) {
```
katcharov, line 64: [optional] It seems that the only public methods needed to fulfill the requirements are `now()` and `elapsed()`.

katcharov (approved):
Just a couple of optional comments.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
stIncMale: > this is already implied by the term "remaining"

Thanks, that was tautology. I removed "saturating".

driver-core/src/main/com/mongodb/internal/time/TimePoint.java
stIncMale: `t.elapsed()` is `TimePoint.now().durationSince(t)`. It can be added for convenience, but it is not enough to implement `Timeout.remainingNanos(TimePoint)`, which is used in `TimeoutTest` to test computation of the remaining time when we have control over what we consider "now".

I think, this could be worked around, and one can get by with `elapsed()`, but I don't see a compelling reason. For a type like `TimePoint` being able to compute the distance from another point (`distanceFrom`) seems like a natural functionality. Given that for time, there are more specific words (distance -> duration, from -> since), `distanceFrom` becomes `durationSince`. That is, the `durationSince` method is not an out-of-place functionality.

stIncMale: I added `elapsed`.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
```diff
@@ -29,148 +31,120 @@
...
+     *
+     * @throws AssertionError If the timeout is {@linkplain #isInfinite() infinite} or {@linkplain #isImmediate() immediate}.
      */
     @VisibleForTesting(otherwise = PRIVATE)
-    long remainingNanos(final long currentNanos) {
-        assertFalse(isInfinite() || isImmediate());
-        long elapsedNanos = elapsedNanos(currentNanos);
-        return elapsedNanos < 0 ? 0 : Math.max(0, durationNanos - elapsedNanos);
+    long remainingNanos(final TimePoint now) {
+        return Math.max(0, durationNanos - now.durationSince(assertNotNull(start)).toNanos());
```
vbabanin, line 125: [Optional] According to the javadoc in `TimePoint`, we can replace `TimePoint.now().durationSince(this)` with the method `elapsed()` to avoid passing a parameter down function.

```
    @VisibleForTesting(otherwise = PRIVATE)
    long remainingNanos() {
        return Math.max(0, durationNanos - assertNotNull(start).elapsed().toNanos());
    }
```

vbabanin (approved):
LGTM,  just one optional comment

driver-core/src/main/com/mongodb/internal/time/Timeout.java
stIncMale: The whole point of exposing `remainingNanos` for tests is so that tests could pass an arbitrary "now". If this were not needed, `remainingNanos` would not have been extracted from `remaining`. With `elapsed`, using an arbitrary "now" is impossible.

PR #1172, opened 2023-08-09
Evergreen: Updated static checks

rozza:
Ensures that Kotlin and Scala checks are also run.

* Renamed compile.sh to static-checks.sh so its discoverable
* Added compileIntegrationTest to Kotlin/Scala checks
* Updated the evergreen static-checks.sh to run the Kotlin & Scala checks

JAVA-5095

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -901,7 +901,7 @@ tasks:
       commands:
         - func: "exec script"
           vars:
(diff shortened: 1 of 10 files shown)


rozza:
Only needs a single LGTM - so first reviewer wins.

katcharov (approved):
LGTM. Looks like the new scalaCheck basically copies the kotlinCheck.

PR #1173, opened 2023-08-10
CSOT: Added timeout to settings & connection string

rozza:
JAVA-4064


rozza:
Note: I tried to remain consistent with the rest of the code.  So added the `MS` suffix where appropriate. Also in the legacy driver the API for timeouts returns milliseconds and doesn't take a `TimeUnit`.

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -280,6 +282,7 @@ public class ConnectionString {
     private Integer maxConnectionLifeTime;
     private Integer maxConnecting;
     private Integer connectTimeout;
+    private Long timeout;
```
katcharov, line 285: should this have the unit, `timeoutMS`?

katcharov (approved):
LGTM, and a question

driver-core/src/main/com/mongodb/ConnectionString.java
rozza: Added to the javadocs its milliseconds. Its annoying that its inconsistent but I opted to keep it consistently inconsistent as not all timeouts will be deprecated.

PR #1178, opened 2023-08-14
Get rid of all `synchronized` blocks and methods in production code

stIncMale:
The philosophy behind this PR:

1. the changes must not break what was guaranteed to work;
2. but if existing code is not correct (some of the touched code looks dubious to me), it's not the goal to deeply analyze it and fix as part of this PR, though if something is obvious - it's fine to fix it.

I recommend enabling the "Hide whitespace" option when viewing changes.

I would appreciate reviewers being especially vigilant reviewing changes in this PR: if you can't convince yourself that the change does not break what was guaranteed to work, please bring your concern up, with as details if possible, as we should then think about it more / discuss it. For example, if you can point out a specific guarantee the new code does not have compared to the original code, and how that specific guarantee is important; or if you can see an undesirable execution that was prohibited by the original code, and is allowed now, then please bring it up.

JAVA-5105

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
@@ -69,16 +69,20 @@ public abstract class AsynchronousChannelStream implements Stream {
         return bufferProvider;
     }
 
-    public synchronized ExtendedAsynchronousByteChannel getChannel() {
+    public ExtendedAsynchronousByteChannel getChannel() {
         return channel;
     }
 
-    protected synchronized void setChannel(final ExtendedAsynchronousByteChannel channel) {
+    protected void setChannel(final ExtendedAsynchronousByteChannel channel) {
         isTrue("current channel is null", this.channel == null);
         if (isClosed) {
             closeChannel(channel);
         } else {
             this.channel = channel;
+            if (isClosed) {
+                this.channel = null;
+                closeChannel(channel);
+            }
         }
     }
 
@@ -158,16 +162,14 @@ public abstract class AsynchronousChannelStream implements Stream {
     }
 
     @Override
-    public synchronized void close() {
+    public void close() {
         isClosed = true;
-        try {
-            closeChannel(channel);
-        } finally {
-            channel = null;
-        }
+        ExtendedAsynchronousByteChannel channel = this.channel;
+        this.channel = null;
+        closeChannel(channel);
     }
 
-    private void closeChannel(final ExtendedAsynchronousByteChannel channel) {
+    private void closeChannel(@Nullable final ExtendedAsynchronousByteChannel channel) {
         try {
             if (channel != null) {
                 channel.close();
--- a/driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
@@ -37,7 +37,7 @@ import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicReference;
 
-import static com.mongodb.assertions.Assertions.isTrue;
+import static com.mongodb.assertions.Assertions.assertTrue;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 /**
@@ -47,7 +47,7 @@ public abstract class AsynchronousChannelStream implements Stream {
     private final ServerAddress serverAddress;
     private final SocketSettings settings;
     private final PowerOfTwoBufferPool bufferProvider;
-    private volatile ExtendedAsynchronousByteChannel channel;
+    private final AtomicReference<ExtendedAsynchronousByteChannel> channel;
     private volatile boolean isClosed;
 
     public AsynchronousChannelStream(final ServerAddress serverAddress, final SocketSettings settings,
@@ -55,6 +55,7 @@ public abstract class AsynchronousChannelStream implements Stream {
         this.serverAddress = serverAddress;
         this.settings = settings;
         this.bufferProvider = bufferProvider;
+        channel = new AtomicReference<>();
     }
 
     public ServerAddress getServerAddress() {
@@ -70,18 +71,16 @@ public abstract class AsynchronousChannelStream implements Stream {
     }
 
     public ExtendedAsynchronousByteChannel getChannel() {
-        return channel;
+        return channel.get();
     }
 
     protected void setChannel(final ExtendedAsynchronousByteChannel channel) {
-        isTrue("current channel is null", this.channel == null);
         if (isClosed) {
             closeChannel(channel);
         } else {
-            this.channel = channel;
+            assertTrue(this.channel.compareAndSet(null, channel));
             if (isClosed) {
-                this.channel = null;
-                closeChannel(channel);
+                closeChannel(this.channel.getAndSet(null));
             }
         }
     }
@@ -120,7 +119,7 @@ public abstract class AsynchronousChannelStream implements Stream {
             timeout += additionalTimeout;
         }
 
-        channel.read(buffer.asNIO(), timeout, MILLISECONDS, null, new BasicCompletionHandler(buffer, handler));
+        getChannel().read(buffer.asNIO(), timeout, MILLISECONDS, null, new BasicCompletionHandler(buffer, handler));
     }
 
     @Override
@@ -164,9 +163,7 @@ public abstract class AsynchronousChannelStream implements Stream {
     @Override
     public void close() {
         isClosed = true;
-        ExtendedAsynchronousByteChannel channel = this.channel;
-        this.channel = null;
-        closeChannel(channel);
+        closeChannel(this.channel.getAndSet(null));
     }
 
     private void closeChannel(@Nullable final ExtendedAsynchronousByteChannel channel) {
@@ -210,7 +207,7 @@ public abstract class AsynchronousChannelStream implements Stream {
 
     private class AsyncWritableByteChannelAdapter {
         void write(final ByteBuffer src, final AsyncCompletionHandler<Void> handler) {
-            channel.write(src, null, new AsyncWritableByteChannelAdapter.WriteCompletionHandler(handler));
+            getChannel().write(src, null, new AsyncWritableByteChannelAdapter.WriteCompletionHandler(handler));
         }
 
         private class WriteCompletionHandler extends BaseCompletionHandler<Void, Integer, Object> {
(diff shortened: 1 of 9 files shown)


driver-legacy/src/main/com/mongodb/DBCollection.java
```diff
@@ -137,9 +139,10 @@ public class DBCollection {
     private volatile ReadPreference readPreference;
     private volatile WriteConcern writeConcern;
     private volatile ReadConcern readConcern;
+    private final Lock factoryAndCodecLock = new ReentrantLock();
     private DBEncoderFactory encoderFactory;
     private DBDecoderFactory decoderFactory;
-    private DBCollectionObjectFactory objectFactory;
+    private volatile DBCollectionObjectFactory objectFactory;
```
stIncMale, line 145: An obvious fix here, but it means nothing with regard to whether the rest of the code is correctly synchronized.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/ResizingByteBufferFlux.java
```diff
@@ -51,19 +52,16 @@ public void subscribe(final CoreSubscriber<? super ByteBuffer> actual) {
...
-                            requested.addAndGet(l);
-                            if (!startedProcessing) {
-                                startedProcessing = true;
-                                upstream().request(1);
-                            }
+                        requested.addAndGet(l);
+                        if (startedProcessing.compareAndSet(false, true)) {
+                            upstream().request(1);
                         }
                     });
```
stIncMale, line 66: I am not sure synchronization is needed here at all, because of the following rules:

- [`Subscription` #1](https://github.com/reactive-streams/reactive-streams-jvm#user-content-3.1): _`Subscription.request` and `Subscription.cancel` MUST only be called inside of its `Subscriber` context._
- [`Subscriber` #7](https://github.com/reactive-streams/reactive-streams-jvm#user-content-2.7): _A `Subscriber` MUST ensure that all calls on its `Subscription`'s request and cancel methods are performed serially._

But, because I am not sure it is not needed (I may be missing or misunderstanding something), I preserved the synchronization.

rozza: Its not related to `cancel`. 
Its to ensure that multiple calls to `subscription.request(n);` don't both trigger an upstream request.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -244,11 +244,8 @@ private ServerDescription lookupServerDescription(final ServerDescription curren
                 }
             } catch (Throwable t) {
                 averageRoundTripTime.reset();
-                InternalConnection localConnection;
-                synchronized (this) {
```
rozza, line 248: Any ideas why this was synchronized in the first place?  Looks like it was added to be overly defensive.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -310,14 +307,9 @@ private long waitForSignalOrTimeout() throws InterruptedException {
         }
 
         public void cancelCurrentCheck() {
```
rozza, line 309: Again this looks like it was purposely defensive - any ideas why? 

Could `cancelCurrentCheck` and other usages of `connection` be interweaved - hence requiring the synchronization?

rozza (commented):
Had a couple of questions :)

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/ResizingByteBufferFlux.java
stIncMale: > Its not related to `cancel`.

Right, I simply quoted the two relevant rules in full.

> Its to ensure that multiple calls to subscription.request(n); don't both trigger an upstream request.

But according to the quoted rules, `request` cannot run concurrently with itself (nor can it run concurrently with `cancel`). This is what makes original `synchronization` and the `AtomicBoolean` replacing it, weird. This also makes using `AtomicLong` for `requested` weird: "at worst", it could be just `volatile`, and "at best" it may also need no synchronization at all, but I did analyze this part because it is irrelevant to the PR.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
stIncMale: > Again this looks like it was purposely defensive - any ideas why?

I can't reconstruct the ideas of why `synchronized` was added because, as far as I can see, it achieves nothing.

> Could cancelCurrentCheck and other usages of connection be interweaved - hence requiring the synchronization?

The only thing that matters for this PR is whether `synchronized` as it is in the code, gives us guarantees that we don't have without it, or not. As far as I can see, it does not. If there is any code that must not be run concurrently with the code inside the `synchronized` block, and that code is not itself guarded by the `synchronized` block using the same `this`, then it is irrelevant whether this `synchronized` block exists or not.

Thus, we only need to consider whether it is fine for the code in the block to run concurrently with itself and with the code in the other `synchronized` block. I convinced myself that it is, and asked reviewers to do the analysis on their own, and bring up specific issues if they see them.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
stIncMale: The reply https://github.com/mongodb/mongo-java-driver/pull/1178#discussion_r1300315024 is equally applicable here.

driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
```diff
@@ -69,16 +69,20 @@ public PowerOfTwoBufferPool getBufferProvider() {
...
         return channel;
     }
 
-    protected synchronized void setChannel(final ExtendedAsynchronousByteChannel channel) {
+    protected void setChannel(final ExtendedAsynchronousByteChannel channel) {
         isTrue("current channel is null", this.channel == null);
         if (isClosed) {
             closeChannel(channel);
         } else {
             this.channel = channel;
```
vbabanin, line 81: It seems there's a possible synchronization order when `getChannel` might see an intermediate value:

[Set] setChannel method is called to set a new channel (channel 1).  `line: 81`
[Close] close method is called, setting isClosed to true. `line: 166`
[Get] getChannel method is called to retrieve the channel. It returns the value (channel 1) because isClosed is still false. `line: 73`
[Close] The channel (channel 1) is closed.  `line: 175`
[Close] The channel value is set to null again.  `line: 168`
[Set] setChannel method sets the channel back to null because it encounters isClosed as true. `line: 83`

Depending on how the class is used, there might be a brief window during which the `getChannel()` method could return an intermediate value due to the simultaneous execution of` setChannel()` and `close()`.

stIncMale: In your execution, `getChannel` runs concurrently with `setChannel` and `close`. And what we get as a result is:

1. `getChannel` observes `channel1` that was passed to `setChannel`.
2. `getChannel` returns `channel1`, and the channel is not closed.
3. `close` closes `channel1`.

Having all these methods `synchronized` as they previously were, does not prevent the above from happening, if the methods run sequentially (they can't run concurrently if they are `synchronized`) in the following order: `setChannel`, `getChannel`, `close`. Do you agree that this is equivalent to your execution, or am I missing something?


> getChannel() method could return an intermediate value

What do you mean by `channel1` being an "intermediate" value? The only two possible values are `null` and `channel1`.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/ResizingByteBufferFlux.java
rozza: Good spot, so this may never have been an issue or is no longer be an issue since we bumped the reactor version (2022.0.0  had some changes that fix compliance with the Reactive streams spec.)

Added JAVA-5114 to track

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
rozza: Yeah, I agree 👍

stIncMale: Currently, `ServerMonitor.cancelCurrentCheck` is not called concurrently with itself, i.e., regardless of whether it's `synchronized` or not, it does not run concurrently with itself. This, in turn, means that `ServerMonitor.cancelCurrentCheck` alone can't result in `ServerMonitorRunnable.connection.close` being called concurrently with itself.

However, if `ServerMonitor.cancelCurrentCheck` is called concurrently with `ServerMonitorRunnable.run`, which is totally possible, then not having the two `synchronized` blocks that were removed, allows concurrently running `ServerMonitorRunnable.run` and `ServerMonitor.cancelCurrentCheck` to call `ServerMonitorRunnable.connection.close` concurrently.

But, there are other reasons in this class why `ServerMonitorRunnable.connection.close` may be called concurrently with itself: `ServerMonitor.close` may run concurrently with `ServerMonitorRunnable.run`, and it also may result in `ServerMonitorRunnable.connection.close` being called concurrently (the same reasoning applies to `RoundTripTimeRunnable.connection`). However, nothing was done in `DefaultServerMonitor` to prevent this, which may suggest that there was not intent to prevent `ServerMonitorRunnable.connection.close` from being called concurrently.

Overall, it is extremely difficult at times to reason about concurrency in the driver, because in many cases the thread-safety and related requirements/guarantees are not documented, especially for internal code. One of the outstanding culprits is `close` method on any interface/class. It is virtually never possible to say correctly or with confidence whether concurrent calls are allowed or not. The code we are discussing right now is a good example: one would expect that `InternalConnection` must never be used concurrently, yet, as I pointed out above, it seems like it may be, regardless of whether `synchronized` blocks are there or not.

TL;RD Thank you for pointing out the problem, the change may indeed introduce issues on top of those that `DefaultServerMonitor` seems to have anyway. I will replace `synchronized` critical sections with critical sections implemented using `DefaultServerMonitor.lock`.

driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
stIncMale: After the discussion https://github.com/mongodb/mongo-java-driver/pull/1178#discussion_r1301481129, I checked and discovered that neither `ExtendedAsynchronousByteChannel`, nor `AsynchronousByteChannel` allow or imply in any way that `close` may be called concurrently with itself. Consequently, I changed the code in `AsynchronousChannelStream` to guarantee that we don't call `ExtendedAsynchronousByteChannel.close` concurrently with itself.

rozza (approved):
LGTM

vbabanin (approved):

PR #1179, opened 2023-08-15
Make the ClientSession close method thread-safe

jyemin:
Replace the volatile boolean with an AtomicBoolean to ensure that the side effects of the close method only execute once.

JAVA-5107

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
+++ b/driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
@@ -26,6 +26,8 @@ import com.mongodb.session.ServerSession;
 import org.bson.BsonDocument;
 import org.bson.BsonTimestamp;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import static com.mongodb.assertions.Assertions.assertTrue;
 import static com.mongodb.assertions.Assertions.isTrue;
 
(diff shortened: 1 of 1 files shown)


stIncMale:
I have two concerns:

1. If we do the change, it may be worth explaining "why" in the code. Seeing code that deals with thread-safety in a class which must not be used concurrently is surprising.
2. This change patches one hole, but there are more. Both `ClientSessionImpl` and `BaseClientSessionImpl` have mutable plain fields, and do plain reads from / writes to them. If application code shares a session with multiple threads, and starts using it concurrently from those threads, that may lead to the session becoming corrupted (data races in Java are not UB, but are still bad, and race conditions are not a good thing either). Are we pathing this specific hole because the corruption it may cause spreads outside of a specific session?

jyemin:
> Are we patching this specific hole because the corruption it may cause spreads outside of a specific session?

Yes, as it causes persistent corruption of the session pool.

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
```diff
@@ -182,8 +183,7 @@ private BsonTimestamp greaterOf(@Nullable final BsonTimestamp newOperationTime)
 
     @Override
     public void close() {
-        if (!closed) {
-            closed = true;
+        if (closed.compareAndSet(false, true)) {
```
stIncMale, line 186: [optional]
It will be helpful to have a comment here explaining why synchronization exists in a class, instances of which must not be used concurrently. Such a comment may address

- why concurrency is relevant for instances of this class at all
- what specific problem is solved here
- why problems that concurrency may cause in other places of this class are disregarded

stIncMale (approved):
Approving because the only suggestion I left is optional.

PR #1182, opened 2023-08-21
Proposal: added generic param to CodecProvider.java

bozzelliandrea:
as the title pertains this is just a proposal and requires multiple code changes to complete, the reason for this pull request is to ask, why not use the generic parameter at the interface declaration level? Just to avoid unnecessary casts during implementation or multiple warnings.

### DIFF
--- a/bson/src/main/org/bson/codecs/configuration/CodecProvider.java
+++ b/bson/src/main/org/bson/codecs/configuration/CodecProvider.java
@@ -33,7 +33,7 @@ import java.util.List;
  *
  * @since 3.0
  */
-public interface CodecProvider {
+public interface CodecProvider<T> {
 
     /**
      * Get a {@code Codec} using the given context, which includes, most importantly, the Class for which a {@code Codec} is required.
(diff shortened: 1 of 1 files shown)


jyemin:
Hi @bozzelliandrea I don't see how this would work in practice.  The `CodecProvider` interface is designed so that a single instance can provide Codecs for multiple classes (see for example `org.bson.codecs.record.RecordCodecProvider`).  

Agreed that the warning suppression that this requires is annoying but I don't see a way around it.

jyemin:
Hi @bozzelliandrea I don't see how this would work in practice.  The `CodecProvider` interface is designed so that a single instance can provide Codecs for multiple classes (see for example `org.bson.codecs.record.RecordCodecProvider`).  

Agreed that the warning suppression that this requires is annoying but I don't see a way around it.

jyemin:
If you'd like to pursue this idea further, please open an issue in our Jira, as per our [contributing guidelines](https://github.com/mongodb/mongo-java-driver/blob/master/CONTRIBUTING.md#bug-fixes-and-new-features).

PR #1183, opened 2023-08-21
Add FaaS (AWS Lambda) test app

katcharov:
JAVA-4758

Modified evergreen project variables, as per above Jira.


.evergreen/run-deployed-lambda-aws-tests.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+RELATIVE_DIR_PATH="$(dirname "${BASH_SOURCE[0]:-$0}")"
+. "${RELATIVE_DIR_PATH}/javaConfig.bash"
+
+# compiled outside of lambda workflow. Note "SkipBuild: True" in template.yaml
+./gradlew -version
+./gradlew -PxmlReports.enabled=true --info -x test -x integrationTest -x spotlessApply jar shadowJar
```
katcharov, line 11: I ran into issues getting things working when building via template.yaml; this was a suggested alternative.

driver-lambda/src/main/com/mongodb/lambdatest/LambdaTestApp.java
```diff
@@ -0,0 +1,143 @@
...
+                    .getCollection("test");
+            BsonValue id = collection.insertOne(new Document("n", 1)).getInsertedId();
+            collection.deleteOne(new Document("_id", id));
+
+            BsonDocument responseBody = new BsonDocument()
+                    .append("totalCommandDuration", new BsonDouble(totalCommandDuration))
+                    .append("totalCommandCount", new BsonDouble(totalCommandCount))
+                    .append("totalHeartbeatDuration", new BsonDouble(totalHeartbeatDuration))
+                    .append("totalHeartbeatCount", new BsonDouble(totalHeartbeatCount))
+                    .append("openConnections", new BsonDouble(openConnections));
```
katcharov, line 118: Note possible differences in naming vs [other drivers](https://github.com/mongodb/node-mongodb-native/commit/ce8e69b6db0eb4c0b9271ce43e334d70b967d227#diff-d63b102d9c9e785a60bfe991630209e6f855b532610cf4aae1f9c2fe71b98164R66). Happy to change to match (though perhaps in that case the [spec](https://github.com/mongodb/specifications/blob/master/source/faas-automated-testing/faas-automated-testing.rst#implementing-the-function) could specify these exactly.)

durran: Don't think the exact naming is important as we aren't asserting anything.

durran:
@katcharov Do you have a patch build showing this running in evergreen? I don't see the variant in the evergreen PR run. (But the evergreen changes appear like they should be working)

katcharov:
@durran I think this should be it: https://spruce.mongodb.com/task/mongo_java_driver_rhel8_test_atlas_test_aws_lambda_deployed_patch_18cb29924b0434efbfcb288e304a8059072d3684_64e3995e32f417c8e241be77_23_08_21_17_05_35/logs?execution=0
```
[2023/08/21 17:24:29.075] {"statusCode":200,"headers":{"X-Custom-Header":"application/json","Content-Type":"application/json"},"body":"{\"totalCommandDuration\": 214.0, \"totalCommandCount\": 4.0, \"totalHeartbeatDuration\": 184068.0, \"totalHeartbeatCount\": 18.0, \"openConnections\": 1.0}"}START RequestId: 79bb53a8-b024-4f02-83db-66272fea225d Version: $LATEST
[2023/08/21 17:24:29.075] END RequestId: 79bb53a8-b024-4f02-83db-66272fea225d
```
3 instances of stats being emitted like this. (If the missing variant should be there, please let me know, and I will try to make sure it's added. I believe `test_atlas_task_group` is new as of this PR.)

stIncMale:
> I think this should be it

@katcharov Just reconfigure the patch submitted as a result of pushing to the commit branch, such that it runs the tasks that were introduced in this PR (you can access the patch through the Evergreen "Details" link in the list of GitHub checks). That's where reviewers are naturally expecting to see the results of the tasks relevant to the PR.

Note that with each new commit pushed you'll need to reconfigure that patch. It is still better than submitting a separate patch and then linking it again each time you add a new commit to the PR.

durran (approved):
Configuration side looks good to me. I won't comment on Java specific items.

.evergreen/.evg.yml
```diff
@@ -2023,6 +2066,12 @@ buildvariants:
   tasks:
      - name: "perf"
 
+- name: rhel8-test-atlas
+  display_name: Atlas Cluster Tests
+  run_on: rhel80-large
+  tasks:
+    - test_atlas_task_group
```
katcharov, line 2073: @rozza Should this run when a commit is merged into master, and if so, where could I set that up? (I did not see anything definite in the project settings.)

rozza: Great question!

> Evergreen monitors a set of github repositories, waiting for new commits. When a new commit comes in or enough time has passed, Evergreen schedules builds for different variants (different OSes, compile flags, etc). <sup>[\[1\]](https://docs.devprod.prod.corp.mongodb.com/evergreen/Home/#how-it-works)</sup> 

So as the evergreen project monitors the master branch, all build variants will run by default. 

See [limiting when a task will run](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#limiting-when-a-task-will-run) to learn about opting out. One such example is the publish release task which opts out by setting `git_tag_only: true`.

Our evg.yml dates back from early evergreen days - I find it hard to grok and I wonder if there are newer features that we don't use. (a thought for another day).

rozza (commented):
Please confirm a patch has successfully run against the latest changes, then I'll LGTM.

driver-lambda/src/main/com/mongodb/lambdatest/LambdaTestApp.java
```diff
@@ -0,0 +1,143 @@
...
+    public APIGatewayProxyResponseEvent handleRequest(final APIGatewayProxyRequestEvent input, final Context context) {
+        try {
+            MongoCollection<Document> collection = mongoClient
+                    .getDatabase("lambdaTest")
+                    .getCollection("test");
+            BsonValue id = collection.insertOne(new Document("n", 1)).getInsertedId();
+            collection.deleteOne(new Document("_id", id));
+
+            BsonDocument responseBody = new BsonDocument()
+                    .append("totalCommandDuration", new BsonDouble(totalCommandDuration))
```
vbabanin, line 114: [nit] Naming this field as `totalCommandDurationMS` might help log readability, as it provides an indication of the measurement.

driver-lambda/template.yaml
```diff
@@ -0,0 +1,69 @@
...
+    Type: String
+    Description: The MongoDB connection string.
+
+Globals:
+  Function:
+    Timeout: 30
+    MemorySize: 128
+    Tracing: Active
+  Api:
+    TracingEnabled: true
```
vbabanin, line 16: Do we need tracing enabled for this test?

driver-lambda/src/main/com/mongodb/lambdatest/LambdaTestApp.java
```diff
@@ -0,0 +1,143 @@
...
+                    .append("totalCommandCount", new BsonDouble(totalCommandCount))
```
vbabanin, line 115: I might be missing something, however, as I see spec requires us only to count the durations of the commands but not the total count of commands.

> Drivers MUST record the durations and counts of the heartbeats, the durations of the commands, as well as keep track of the number of open connections, and report this information in the function response as JSON.

driver-lambda/src/main/com/mongodb/lambdatest/LambdaTestApp.java
```diff
@@ -0,0 +1,143 @@
...
+                    .append("totalHeartbeatDuration", new BsonDouble(totalHeartbeatDuration))
+                    .append("totalHeartbeatCount", new BsonDouble(totalHeartbeatCount))
+                    .append("openConnections", new BsonDouble(openConnections));
+
+            return templateResponse()
```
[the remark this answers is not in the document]
katcharov: I do not believe capturing a running total violates the quoted spec.

We took a similar "running total" approach in the Kafka JMX. I believe that the user is able to derive any timings, by taking a sample S when a reset would otherwise happen, and then subtracting S from the usual sample E, to get the change during the interval. The benefit is that the user has precise control over when the "reset" happens (that is, the start point), and queries do not change the application state. There are, of course, circumstances where a "hardcoded" start and end must be used (for example, the execution of a particular process must be timed very rigorously), but I do not think there is such a requirement here.

(I hope I am not misreading what was meant by separate tracking of commands/heartbeats - there is both `totalCommandDuration`, and `totalHeartbeatDuration`.)

driver-lambda/template.yaml
katcharov: I believe this is from [the spec](https://github.com/mongodb/specifications/blob/master/source/faas-automated-testing/faas-automated-testing.rst#project-initialization) saying:
> Drivers MAY choose to also enable X-Ray tracing and CloudWatch Application Insights during these next steps.
My thought was that this was suggested because it might be needed in the future or come in handy.

driver-lambda/src/main/com/mongodb/lambdatest/LambdaTestApp.java
katcharov: I agree, but I do not think it will be harmful to provide the count as well for consistency; also, Node supplies this as well.

driver-lambda/build.gradle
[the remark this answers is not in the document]
katcharov: I agree (perhaps `driver-benchmark-lambda`?), but it seems a mild improvement; however, I am apprehensive about changing the naming / configuration (which was difficult to get working). If you feel your suggestion is more of a "should" than a "might", please let me know.

driver-lambda/src/main/com/mongodb/lambdatest/LambdaTestApp.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.Map;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+/**
+ * Test App for AWS lambda functions
+ */
+public class LambdaTestApp implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
+    private final MongoClient mongoClient;
+    private double openConnections = 0;
```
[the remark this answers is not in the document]
katcharov: I vaguely remember having some reason for these, which is why I left them when I submitted the PR, but I cannot remember it now. Updated.

driver-lambda/build.gradle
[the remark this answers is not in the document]
vbabanin: Using `driver-benchmark-lambda` sounds reasonable. Given the potential complexity of altering the naming, let's keep it as it is.

driver-lambda/src/main/com/mongodb/lambdatest/LambdaTestApp.java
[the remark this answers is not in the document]
vbabanin: Thank you for providing this explanation. As I understand it, the testing aims to highlight the unique differences between lambda invocations. Maintaining separate results for each invocation ensures clear and distinct data in the logs without the need for additional calculations. For those interested in cumulative totals, they can perform the necessary calculations using the recorded durations.

However, I don't have a strong preference here, so feel free to resolve this thread

driver-lambda/template.yaml
vbabanin: Thanks for mentioning the spec's suggestion about enabling AWS X-Ray tracing. Since we don't seem to use distributed tracing for Lambda invocation time testing, I would suggest not using it until required. AWS X-Ray might introduce extra charges

vbabanin (approved):

driver-lambda/template.yaml
katcharov: I have disabled this

driver-lambda/src/main/com/mongodb/lambdatest/LambdaTestApp.java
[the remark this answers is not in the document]
katcharov: I checked with `@durran` and the resetting behaviour was indeed preferred, so I have changed it (with the exception of not resetting the open connection count, which should not be reset).

katcharov:
@rozza [test-aws-lambda-deployed passed](https://spruce.mongodb.com/task/mongo_java_driver_rhel8_test_atlas_test_aws_lambda_deployed_patch_80a9cb4f961f727f4a5ac85882e658723e7f04d9_65009934e3c33107e320adb8_23_09_12_17_00_37/logs?execution=0). The remaining failures, I believe, would be fixed in master.

rozza (approved):
LGTM

PR #1185, opened 2023-08-28
Proposal for deprecating Stream-related API

jyemin:
Adds:

* TransportSettings abstract class
* NettySettings subclass of TransportSettings
* A new method on MongoClientSettings to configure TransportSettings

Deprecates:

* MongoClientSettings#streamFactoryFactory
* NettyStreamFactory
* AsynchronousSocketChannelStreamFactory
* AsynchronousSocketChannelStreamFactoryFactory
* BufferProvider
* SocketStreamFactory
* Stream
* StreamFactory
* StreamFactoryFactory
* TlsChannelStreamFactoryFactory

JAVA-5051

### DIFF
--- a/bson/src/main/org/bson/ByteBufNIO.java
+++ b/bson/src/main/org/bson/ByteBufNIO.java
@@ -26,6 +26,10 @@ import java.util.concurrent.atomic.AtomicInteger;
  *
  * @since 3.0
  */
+@Deprecated
+// TODO: Can we?  Can we go even further and deprecate ByteBuf? The main public usage in in RawBsonArray and RawBsonDocument
+// and in org.bson.io package. It would be nice because ByteBuf is not a good abstraction currently and removing it from the API would
+// give us more flexibility to refactor it.
 public class ByteBufNIO implements ByteBuf {
     private ByteBuffer buf;
     private final AtomicInteger referenceCount = new AtomicInteger(1);
--- a/driver-core/src/main/com/mongodb/MongoClientSettings.java
+++ b/driver-core/src/main/com/mongodb/MongoClientSettings.java
@@ -27,6 +27,7 @@ import com.mongodb.connection.ServerSettings;
 import com.mongodb.connection.SocketSettings;
 import com.mongodb.connection.SslSettings;
 import com.mongodb.connection.StreamFactoryFactory;
+import com.mongodb.connection.TransportSettings;
 import com.mongodb.event.CommandListener;
 import com.mongodb.lang.Nullable;
 import com.mongodb.spi.dns.DnsClient;
@@ -490,9 +491,25 @@ public final class MongoClientSettings {
          * @param streamFactoryFactory the stream factory factory
          * @return this
          * @see #getStreamFactoryFactory()
+         * @see #transportSettings(TransportSettings)
          */
+        @Deprecated
         public Builder streamFactoryFactory(final StreamFactoryFactory streamFactoryFactory) {
             this.streamFactoryFactory = notNull("streamFactoryFactory", streamFactoryFactory);
+            return this;
+        }
+
+        /**
+         * Enables Netty with the given settings.
+         *
+         * <p>
+         * If Netty is enabled with this method, it is an error to also use {@link #streamFactoryFactory}.
+         * </p>
+         *
+         * @param nettySettings the settings for Netty
+         * @return this
+         */
+        public Builder transportSettings(final TransportSettings nettySettings) {
             return this;
         }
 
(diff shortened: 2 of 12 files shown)


driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -771,12 +788,26 @@ public CodecRegistry getCodecRegistry() {
...
 
+    /**
+     * Gets the settings for the underlying transport API
+     *
+     * @return the settings for the underlying transport API
+     *
+     * @since 4.11
+     */
+    @Nullable
+    public TransportSettings getTransportSettings() {
```
jyemin, line 807: Questions:

1. Is "transport" the right word for this?
2. Should this be part of SocketSettings instead?  (But then it's weird that NettySettings has an SslContext property that is more closely associated with SslSettings)
3. Is it weird that TransportSettings has no methods?

driver-core/src/main/com/mongodb/connection/netty/NettySettings.java
```diff
@@ -0,0 +1,173 @@
...
+     * @return the builder
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * A builder for an instance of {@code NettyStreamFactoryFactory}.
+     */
+    public static final class Builder {
```
jyemin, line 61: Should this similarly extend a `TransportSettings.Builder` class that has no methods?  It doesn't seem useful.

driver-core/src/main/com/mongodb/connection/TransportSettings.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @since 4.11
+ */
+public abstract class TransportSettings {
+    /**
+     * A builder for NettySettings.
+     *
+     * @return a builder for NettySettings
+     */
+    public static NettySettings.Builder nettyBuilder() {
```
jyemin, line 32: I added this for discoverability, but it duplicates `NettySettings.builder()`.  Should we have both public, only this one, or only `NettySettings.builder()?

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -490,12 +491,28 @@ public Builder codecRegistry(final CodecRegistry codecRegistry) {
          * @param streamFactoryFactory the stream factory factory
          * @return this
          * @see #getStreamFactoryFactory()
+         * @see #transportSettings(TransportSettings)
          */
+        @Deprecated
         public Builder streamFactoryFactory(final StreamFactoryFactory streamFactoryFactory) {
```
stIncMale, line 497: Let's use `@deprecated` instead of `@see` to specify the replacing API, etc.

driver-core/src/main/com/mongodb/connection/TransportSettings.java
```diff
@@ -0,0 +1,35 @@
...
+package com.mongodb.connection;
+
+import com.mongodb.connection.netty.NettySettings;
+
+/**
+ * Transport settings for the driver.
+ *
+ * @since 4.11
+ */
+public abstract class TransportSettings {
```
stIncMale, line 26: I envision `GrpcSettings extends TransportSettings` in the future. Even if `GrpcSettings` does not bear any settings in the beginning or ever, specifying `GrpcSettings` will be the way to enable gRPC via `MongoClientSettings`. How does this seem to you?

driver-core/src/main/com/mongodb/connection/TransportSettings.java
stIncMale: I prefer having these only on `TransportSettings`. If we ever need them on subtypes, which I doubt, we will be able to add the methods.

driver-core/src/main/com/mongodb/connection/netty/NettySettings.java
stIncMale: If we have such a hierarchy, `NettySettings.Builder.setCommonSomething` must return `NettySettings.Builder`, despite `TransportSettings.Builder.setCommonSomething` returning `TransportSettings.Builder`. It seems to me that we can introduce such a hierarchy later, when/if we actually need to have a `setCommonSomething` method. Unless you see a technical reason why postponing may be a problem, I prefer to postpone introducing the hierarchy now.

driver-core/src/main/com/mongodb/connection/TransportSettings.java
jyemin: I anticipated the same.  The only weird part perhaps is that, since the Java gRPC client uses Netty as its I/O substrate, we may want to embed NettySettings in GrpcSettings.

bson/src/main/org/bson/ByteBufNIO.java
```diff
@@ -26,6 +26,10 @@
  *
  * @since 3.0
  */
+@Deprecated
+// TODO: Can we?  Can we go even further and deprecate ByteBuf? The main public usage in in RawBsonArray and RawBsonDocument
+// and in org.bson.io package. It would be nice because ByteBuf is not a good abstraction currently and removing it from the API would
+// give us more flexibility to refactor it.
```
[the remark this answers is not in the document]
jyemin: Decided to defer this until post-5.0

jyemin:
We've learned what we needed to from this.  Closing.

driver-core/src/main/com/mongodb/connection/TransportSettings.java
stIncMale: If we can support them all (I don't remember), that may be helpful. We'll see.

PR #1186, opened 2023-08-29
Remove incorrect test-skipping logic

jyemin:
JAVA-5124

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableWritesTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableWritesTest.java
@@ -45,6 +45,8 @@ import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+import static com.mongodb.ClusterFixture.isSharded;
+import static com.mongodb.ClusterFixture.serverVersionLessThan;
 import static com.mongodb.JsonTestServerVersionChecker.skipTest;
 import static com.mongodb.client.Fixture.getDefaultDatabaseName;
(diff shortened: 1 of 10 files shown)


driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableWritesTest.java
```diff
@@ -254,7 +253,7 @@ public static Collection<Object[]> data() throws URISyntaxException, IOException
                 data.add(new Object[]{file.getName(), test.asDocument().getString("description").getValue(),
                         testDocument.getString("database_name", new BsonString(getDefaultDatabaseName())).getValue(),
                         testDocument.getArray("data"), test.asDocument(),
-                        !isDiscoverableReplicaSet() || skipTest(testDocument, test.asDocument())});
```
jyemin, line 257: I couldn't figure out why this was ever necessary, but it seems like it no longer is.

jyemin: Maybe the reason is because the tests fail on 4.2 and 4.4 sharded clusters.  Will look into that.

jyemin: I "fixed" it by bumping minServerVersion for sharded clusters to 5.0.  I'm going to check with other teams to see how they are passing these tests.

jyemin: Opened JAVA-5125 to find the root cause of the test failures, so I think we can proceed with this PR as it does increase the test coverage.

rozza (approved):
LGTM!

PR #1187, opened 2023-08-30
Add `InterruptibilityItTest`

stIncMale:
To run the tests, run `mvn verify` in the project root directory.

Note that all tests pass when run against `mongodb-driver-sync:4.10.2`. We still need to do https://jira.mongodb.org/browse/JAVA-4646, but it does not harm interruptibility.

JAVA-5109


src/test/java/com/mongodb/client/FailPoint.java
stIncMale, line 1: Had to copy this class from the `master` branch.

src/test/java/com/mongodb/InterruptibilityItTest.java
```diff
@@ -0,0 +1,277 @@
...
+                clientSettingsBuilderSupplier)) {
+            return assertThrows(expectedType, () ->
+                    inVirtualThread(() ->
+                            inInterruptedThread(interruptDelay, action))
+            );
+        }
+    }
+
+    private static void inVirtualThread(final Runnable action) throws TimeoutException {
+        try (ExecutorService ex = Executors.newVirtualThreadPerTaskExecutor()) {
```
stIncMale, line 194: If you replace `newVirtualThreadPerTaskExecutor` with `newCachedThreadPool`, all but the `serverSelection` tests start failing.

src/test/java/com/mongodb/InterruptibilityItTest.java
```diff
@@ -0,0 +1,277 @@
...
+        }
+    }
+
+    @ParameterizedTest
+    @MethodSource("arguments")
+    void serverSelection(final Duration interruptDelay) {
+        // we assume here that the time to elect a new leader is greater than `interruptDelay`
+        client.getDatabase("admin").runCommand(BsonDocument.parse(
+                "{\n"
+                // using `blockTimeout` does not give us much because it controls only the unelectability of the current primary
```
jyemin, line 148: Not sure what this comment is intended to convey.  Does `blockTimeout` give us what we need? If so, it seems unnecessary to say that it doesn't give us much.  Please clarify, or remove.

src/test/java/com/mongodb/InterruptibilityItTest.java
```diff
@@ -0,0 +1,277 @@
...
+                throw new RuntimeException(e);
+            } catch (ExecutionException e) {
+                //noinspection ThrowFromFinallyBlock
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    private static ServerAddress primary(final MongoClient client, final Duration timeout) {
+        Timeout tOut = Timeout.startNow(timeout.toNanos());
```
jyemin, line 251: tOut => timeout

(avoid most abbreviations)

jyemin (changes requested):
Looks great. Only significant request is for a potential new test of connection pool checkout interruption.

src/test/java/com/mongodb/InterruptibilityItTest.java
```diff
@@ -0,0 +1,277 @@
...
+                + "    'force': true\n"
+                + "}"));
+        MongoIterable<String> iterable = client.listDatabaseNames();
+        assertThrowsInVirtualInterruptedThread(MongoInterruptedException.class,
+                interruptDelay, null, iterable::first);
+    }
+
+    private static Stream<Arguments> arguments() {
+        return Stream.of(
+                Arguments.of(Duration.ZERO),
```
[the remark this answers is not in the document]
stIncMale: This is in case we accidentally clear the status somewhere in the code before the execution reached the point at which the thread becomes blocked. Obviously, we don't do that, but all other tests also pass at the moment, yet we have them.

src/test/java/com/mongodb/InterruptibilityItTest.java
jyemin: You could change the parameter name to `timoutDuration`

src/test/java/com/mongodb/InterruptibilityItTest.java
stIncMale: Done.

src/test/java/com/mongodb/InterruptibilityItTest.java
stIncMale: That's much better. Done.

src/test/java/com/mongodb/client/FailPoint.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * @param configureFailPointDoc A document representing {@code configureFailPoint} command to be issued as is via
+     * {@link com.mongodb.client.MongoDatabase#runCommand(Bson)}.
+     */
+    public static FailPoint enable(
+            final BsonDocument configureFailPointDoc,
+            final ServerAddress serverAddress,
+            final Supplier<MongoClientSettings.Builder> clientSettingsBuilderSupplier) {
```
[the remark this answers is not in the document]
stIncMale: Done.

src/test/java/com/mongodb/InterruptibilityItTest.java
```diff
@@ -105,8 +105,6 @@ void beforeEach() {
 
     @AfterEach
     void afterEach() {
-        // make sure the current thread is not interrupted because tests mess with this status
-        com.mongodb.assertions.Assertions.assertFalse(Thread.interrupted());
```
stIncMale: Removed this assertion because if the test framework were to interrupt tests, the assertion would have incorrectly failed.

src/test/java/com/mongodb/InterruptibilityItTest.java
```diff
@@ -0,0 +1,314 @@
...
+    @ParameterizedTest
+    @MethodSource("arguments")
+    void connectionCheckout(final Duration interruptDelay) throws InterruptedException {
+        Future<?> backgroundConnectionBlocker = ForkJoinPool.commonPool().submit(() ->
+                withBlockedCommand("listIndexes", blockTimeout, () -> collection.listIndexes().first()));
+        // Wait until the background task blocks the only connection in the pool.
+        // We need the duration the connection remains blocked after this waiting to be greater than `interruptDelay`.
+        Thread.sleep(blockTimeout.minus(interruptDelay).dividedBy(2));
+        assertThrowsInVirtualInterruptedThread(MongoInterruptedException.class,
+                interruptDelay, null, collection::drop);
```
stIncMale, line 174: If you either replace `() -> collection.listIndexes().first()` with `() -> null` (i.e., stop blocking the single connection there is), or replace `maxSize(1)` with `maxSize(2)` in the `beforeAll` method, this test fails with `org.opentest4j.AssertionFailedError: Expected com.mongodb.MongoInterruptedException to be thrown, but nothing was thrown.`, as expected.

src/test/java/com/mongodb/InterruptibilityItTest.java
```diff
@@ -0,0 +1,277 @@
...
+import java.util.function.Supplier;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import static java.lang.Math.toIntExact;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+final class InterruptibilityItTest {
```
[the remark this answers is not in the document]
stIncMale: Thanks, good idea. Done.

jyemin (commented):
> Note that all tests pass when run against mongodb-driver-sync:4.10.2. We still need to do JAVA-4646, but it does not harm interruptibility.

Is there any way to write an integration test that would detect JAVA-4646?

stIncMale:
> Is there any way to write an integration test that would detect [JAVA-4646](https://jira.mongodb.org/browse/JAVA-4646)?

Not unless we inspect exception messages / stack traces. This is because JAVA-4646 does not result in a `MongoClient` user observing an incorrect behavior.

Internally, JAVA-4646 results in the SDAM machinery handling an exception different from the `MongoInterruptedException` when an operation is interrupted. This has potential for causing the driver to retry the command, though currently that does not happen. I modified the tests to check that each when an operation is interrupted the corresponding command is not retried.

jyemin:
> This has potential for causing the driver to retry the command, though currently that does not happen

There is one other observable thing that we could check for.  SDAM requires that non-timeout socket errors result in the server being marked as UNKNOWN (and the pool cleared).  So I think we could also add a ClusterListener to the MongoClient and assert that no servers were marked as UNKNOWN during the course of the operation under test.  And/or a ConnectionPoolListener and assert no pools have been cleared.

Does that make sense?

jyemin (changes requested):

src/test/java/com/mongodb/InterruptibilityItTest.java
```diff
@@ -191,15 +202,21 @@ private static Stream<Arguments> arguments() {
...
+        return withBlockedCommand(blockCommand ? commandName : null, blockTimeout, () -> {
+            startedCommandNames.clear();
+            T t = assertThrows(expectedType, () ->
+                    inVirtualThread(() ->
+                            inInterruptedThread(interruptDelay, executeCommand)
+                    )
+            );
+            assertFalse(startedCommandNames.stream().anyMatch(startedCommandName -> !startedCommandName.equals(commandName)),
+                    "Unexpected commands in " + startedCommandNames + ", expected only " + commandName);
+            assertTrue(startedCommandNames.size() <= 1, startedCommandNames + "was attempted " + startedCommandNames.size() + " times");
```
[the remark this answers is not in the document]
stIncMale: `commandName` is never `null` here. The reason `<= 1` is used is that an operation may be interrupted either before `CommandStartedEvent`, or after.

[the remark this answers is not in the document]
jyemin: Ah, ok.  But still, doesn't each test know the expected number of commands that will be started, e.g. the server selection test expects 0, and so on?  If so, each test could pass in the expected number of commands. 

If you think it's overkill, I'll let it go.  I point it out only because I have vague memories of being burned by fuzzy assertions like this.

stIncMale:
> There is one other observable thing that we could check for. SDAM requires that non-timeout socket errors result in the server being marked as UNKNOWN (and the pool cleared). So I think we could also add a ClusterListener to the MongoClient and assert that no servers were marked as UNKNOWN during the course of the operation under test. And/or a ConnectionPoolListener and assert no pools have been cleared.
> 
> Does that make sense?

Done.

jyemin (approved):
LGTM

PR #1188, opened 2023-09-06
Deprecate SocketAddress-related methods

jyemin:
* ServerAddress#getSocketAddress
* ServerAddress#getSocketAddresses
* UnixServerAddress#getUnixSocketAddress

JAVA-4940

When we actually remove these methods and replace with some private helper, and also remove `StreamFactory` from API, we'll be able to remove the `ServerAddressWithResolver` class and just pass the `InetAddressResolver` wherever it needs to be available.

### DIFF
--- a/driver-core/src/main/com/mongodb/UnixServerAddress.java
+++ b/driver-core/src/main/com/mongodb/UnixServerAddress.java
@@ -44,6 +44,8 @@ public final class UnixServerAddress extends ServerAddress {
         isTrueArgument("The path must end in .sock", path.endsWith(".sock"));
     }
 
+    @SuppressWarnings("deprecation")
+    @Deprecated
     @Override
     public InetSocketAddress getSocketAddress() {
         throw new UnsupportedOperationException("Cannot return a InetSocketAddress from a UnixServerAddress");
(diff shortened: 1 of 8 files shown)


driver-core/src/main/com/mongodb/UnixServerAddress.java
```diff
@@ -44,14 +44,18 @@ public UnixServerAddress(final String path) {
...
     @Override
     public InetSocketAddress getSocketAddress() {
         throw new UnsupportedOperationException("Cannot return a InetSocketAddress from a UnixServerAddress");
     }
 
     /**
      * @return the SocketAddress for the MongoD unix domain socket.
+     * @deprecated Prefer {@link UnixSocketAddress#UnixSocketAddress(String)}
      */
+    @Deprecated
```
jyemin, line 58: This one seems similar enough to also deprecate at the same time

driver-core/src/main/com/mongodb/UnixServerAddress.java
```diff
@@ -44,14 +44,18 @@ public UnixServerAddress(final String path) {
         isTrueArgument("The path must end in .sock", path.endsWith(".sock"));
     }
 
+    @SuppressWarnings("deprecation")
```
jyemin, line 47: I'm not sure why the warning also has to be suppressed here, but it complains if I don't

stIncMale: It's probably because of the [`-Xlint:deprecation`](https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-deprecation) lint: the method overrides a deprecated method.

stIncMale (approved):

PR #1189, opened 2023-09-06
Handle `Socket` IO interruptibility

stIncMale:
JAVA-4646

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -422,11 +422,15 @@ final class DefaultConnectionPool implements ConnectionPool {
                         RuntimeException actualException = e instanceof MongoOpenConnectionInternalException
                                 ? (RuntimeException) e.getCause()
                                 : e;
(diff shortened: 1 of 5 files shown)


stIncMale:
@jyemin in https://github.com/jyemin/mongo-java-driver/commit/f100207fc29b4b5df6a376e96807f44c9b65985e, you made a change in `DefaultServer` to avoid masking the `MongoInterruptedException`. I'll deal with it by addressing the root cause (after I figure out why the tests are failing): the code that calls `SdamServerDescriptionManager.handleExceptionBeforeHandshake`/`handleExceptionAfterHandshake` must not lose the original exception regardless whether it was caused by an interrupt or not.

jyemin (approved):
LGTM!

stIncMale:
The tests in https://github.com/mongodb/mongo-java-driver/tree/JAVA_5109 pass when run against the changes in this PR.

stIncMale:
Looks like the tests fail because the server broke something:

- they only fail for the latest MongoDB
- they also fail in the master branch https://spruce.mongodb.com/version/64fa31ac1e2d17cdd7df67fb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC.

PR #1190, opened 2023-09-06
Improve performance monitoring

vbabanin:
This PR introduces enhancements to performance monitoring:

-  Adds dedicated distro rhel90-dbx-perf-large for performance testing.
- Uses patch-pinned v6 server (6.0.6) as v6.0-perf alias.

JAVA-5065

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -2060,15 +2060,10 @@ buildvariants:
 
 - name: "perf"
   display_name: Performance Tests
+  tags: ["perf-variant"]
   run_on: rhel90-dbx-perf-large
   tasks:
     - name: "perf"
-      vars:
-        VERSION: "v6.0-perf"
-        TOPOLOGY: "server"
-        SSL: "nossl"
-        AUTH: "noauth"
-        SKIP_LEGACY_SHELL: "true"
 
 - name: plain-auth-test
   display_name: "PLAIN (LDAP) Auth test"
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -148,7 +148,7 @@ functions:
       params:
         script: |
           ${PREPARE_SHELL}
-          REQUIRE_API_VERSION=${REQUIRE_API_VERSION} LOAD_BALANCER=${LOAD_BALANCER} MONGODB_VERSION=${VERSION} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
+          REQUIRE_API_VERSION=${REQUIRE_API_VERSION} LOAD_BALANCER=${LOAD_BALANCER} MONGODB_VERSION=${VERSION} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} SKIP_LEGACY_SHELL=${SKIP_LEGACY_SHELL} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
     # run-orchestration generates expansion file with the MONGODB_URI for the cluster
     - command: expansions.update
       params:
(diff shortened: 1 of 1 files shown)


.evergreen/.evg.yml
```diff
@@ -148,7 +148,7 @@ functions:
       params:
         script: |
           ${PREPARE_SHELL}
-          REQUIRE_API_VERSION=${REQUIRE_API_VERSION} LOAD_BALANCER=${LOAD_BALANCER} MONGODB_VERSION=${VERSION} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
+          REQUIRE_API_VERSION=${REQUIRE_API_VERSION} LOAD_BALANCER=${LOAD_BALANCER} MONGODB_VERSION=${VERSION} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} SKIP_LEGACY_SHELL=${SKIP_LEGACY_SHELL} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
```
vbabanin, line 151: The performance test distro, linux-rhel-9*, is missing a corresponding MongoDB version 5, causing a failure in downloading the legacy shell. To address this, the SKIP_LEGACY_SHELL flag has been added to `bootstrap mongo-orchestration` function. This prevents the download-mongodb.sh script from prompting to download version 5.0.

You can find the specific line of code [here](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/4be70a5bba0f9ca15d10a41aaa6a0092b398bf3b/.evergreen/download-mongodb.sh#L144).

.evergreen/.evg.yml
```diff
@@ -2052,13 +2058,17 @@ buildvariants:
...
-  display_name: "Perf Tests ${version} "
-  tags: ["perf-variant"]
+- name: "perf"
+  display_name: Performance Tests
+  run_on: rhel90-dbx-perf-large
   tasks:
-     - name: "perf"
+    - name: "perf"
+      vars:
+        VERSION: "v6.0-perf"
```
vbabanin, line 2067: Driver evergreen tools exposes the patch pinned v6 server version 6.0.6 that can be referenced via the `v6.0-perf` version alias (the analogous perf-stable v7 version will be added later)

vbabanin:
Performance trend charts can be found here: [link](https://spruce.mongodb.com/task/mongo_java_driver_perf_perf_patch_d08adffbf9837abad78029af1fcd1dfb60ff8dde_64f90e70d6d80ad7aeb7e862_23_09_06_23_42_42/trend-charts?execution=0)

.evergreen/.evg.yml
```diff
@@ -1491,6 +1491,12 @@ tasks:
       tags: ["perf"]
       commands:
         - func: "bootstrap mongo-orchestration"
+          vars:
```
jyemin, line 1494: Do you understand how this worked before adding these variables?

.evergreen/.evg.yml
```diff
@@ -2052,13 +2058,17 @@ buildvariants:
...
-  batchtime: 1440 # run once a day
-  display_name: "Perf Tests ${version} "
-  tags: ["perf-variant"]
+- name: "perf"
+  display_name: Performance Tests
+  run_on: rhel90-dbx-perf-large
   tasks:
-     - name: "perf"
+    - name: "perf"
+      vars:
```
jyemin, line 2066: Why are the vars repeated here?

.evergreen/.evg.yml
vbabanin: Based on my understanding, we're using the same set of variables from matrix_spec:

{ auth: "noauth", ssl: "nossl", jdk: "jdk17", version: ["latest"], topology: "standalone", os: "linux" }

JDK 17 is the default in javaConfig.bash, and we've updated the database version. This should maintain the original workflow. Is there something I might be overlooking?"

jyemin: No, I get it now.  I didn't put it together that we need the variables after removing the matrix spec.

jyemin (commented):
> Allows viewing performance results in Trend Charts Tab

Is running on `rhel90-dbx-perf-large` what enables this?  Or something else?

vbabanin:
Actually, it is neither. It is the `perf.send `command and Performance Plugin which enable Trend Chart. 

I have removed it from the PR  description:
> Allows viewing performance results in Trend Charts Tab

jyemin (approved):

PR #1191, opened 2023-09-08
Handle new explain output in tests

jyemin:
The response document for the explain command has changed in 7.2 sharded clusters. It introduces a top-level "shards" document that contains the explain output for each shard. This patch supports this new field in tests of explain API in the driver.

JAVA-5137

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/AbstractExplainTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/AbstractExplainTest.java
@@ -119,6 +119,8 @@ public abstract class AbstractExplainTest {
         assertFalse(explainBsonDocument.containsKey("executionStats"));
     }
 
+    // Post-MongoDB 7.0, sharded cluster responses move the explain plan document into a "shards" document, which a plan for each shard.
+    // This method grabs the explain plan document from the first shard when this new structure is present.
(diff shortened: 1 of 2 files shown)


driver-sync/src/test/functional/com/mongodb/client/AbstractExplainTest.java
```diff
@@ -99,27 +99,48 @@ public void testExplainOfAggregateWithNewResponseStructure() {
         AggregateIterable<BsonDocument> iterable = collection
                 .aggregate(singletonList(Aggregates.match(Filters.eq("_id", 1))));
 
-        Document explainDocument = iterable.explain();
-        assertNotNull(explainDocument);
+        Document explainDocument = getAggregateExplainDocument(iterable.explain());
```
katcharov, line 102: There's an explanatory comment related to such a structure change at the top of this method, should it be updated / removed?

katcharov (approved):
LGTM

driver-sync/src/test/functional/com/mongodb/client/AbstractExplainTest.java
jyemin: Good idea.  That's not quite the place for it (since it's about skipping the tests) but I added a note above the new helper method instead.

PR #1192, opened 2023-09-08
Fix how `InterruptedException`s are handled

stIncMale:
JAVA-4641

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
@@ -16,6 +16,7 @@
 
 package com.mongodb.internal.connection;
 
+import com.mongodb.MongoInterruptedException;
 import com.mongodb.MongoNamespace;
 import com.mongodb.MongoSocketException;
 import com.mongodb.ServerApi;
@@ -179,7 +180,7 @@ class DefaultServerMonitor implements ServerMonitor {
                     }
                     waitForNext();
                 }
-            } catch (InterruptedException closed) {
+            } catch (InterruptedException | MongoInterruptedException closed) {
                 // stop the monitor
             } catch (RuntimeException e) {
                 LOGGER.error(format("Server monitor for %s exiting with exception", serverId), e);
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
@@ -16,7 +16,6 @@
 
 package com.mongodb.internal.connection;
 
-import com.mongodb.MongoInterruptedException;
 import com.mongodb.MongoNamespace;
 import com.mongodb.MongoSocketException;
 import com.mongodb.ServerApi;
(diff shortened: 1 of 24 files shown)


driver-core/src/main/com/mongodb/internal/connection/tlschannel/async/AsynchronousTlsChannelGroup.java
```diff
@@ -383,7 +384,7 @@ private void waitForSocketRegistration(RegisteredSocket socket) {
         try {
             socket.registered.await();
         } catch (InterruptedException e) {
-            throw new RuntimeException(e);
+            throw interruptAndCreateMongoInterruptedException(null, e);
```
stIncMale, line 387: This is a change in the code we vendor. As far as I know, we don't have a process for tracking such changes.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -180,8 +179,8 @@ public void run() {
                     }
                     waitForNext();
                 }
-            } catch (MongoInterruptedException e) {
```
jyemin, line 183: Why remove this?  Couldn't this exception still be thrown by `lookupServerDescription`?

jyemin (changes requested):

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
stIncMale: Thanks! It may, it's not obvious which exceptions may be produced by the code in that `try` block.

driver-core/src/main/com/mongodb/internal/connection/tlschannel/async/AsynchronousTlsChannelGroup.java
stIncMale: The thing is that we'll never be able to make it the way we need: the driver is supposed to wrap `InterruptedException`s in `MongoInterruptedException`, but it won't do so when `InterruptedException` is thrown from this code.

I'll create a PR agains the `tlschannel`. If you check the history of changes in this file, you'll see we made driver-specific changes, that are not part of the original library. Given this, do you still want the change to be rolled back, and us not to wrap `InterruptedException` in `MongoInterruptedException` here?

jyemin: I guess it's ok to leave it as is.  It's just a fragile situation we're in here, since we don't have a process for re-vendoring.

stIncMale: Created https://github.com/marianobarrios/tls-channel/pull/165.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
stIncMale: Done.

jyemin (approved):

PR #1194, opened 2023-09-11
Remove Timeout, introduce StartTime and Deadline

katcharov:
JAVA-5112


driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1074,15 +1085,14 @@ private PooledConnection acquirePermitOrGetAvailableOpenedConnection(final boole
...
                         & !stateAndGeneration.throwIfClosedOrPaused()
                         & (availableConnection = tryGetAvailable ? tryGetAvailableConnection() : null) == null) {
-                    if (Timeout.expired(remainingNanos)) {
-                        throw createTimeoutException(timeout);
+                    if (timeout.hasExpired()) {
+                        throw createTimeoutException(startTime);
                     }
-                    remainingNanos = awaitNanos(permitAvailableOrHandedOverOrClosedOrPausedCondition, remainingNanos);
+                    timeout.awaitOn(permitAvailableOrHandedOverOrClosedOrPausedCondition);
                 }
```
katcharov, line 1096: Note that the timeout is now checked directly, rather than the value from await.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -1163,28 +1173,10 @@ void signalClosedOrPaused() {
                 lock.unlock();
             }
         }
-
-        /**
-         * @param timeoutNanos See {@link Timeout#started(long, TimePoint)}.
-         * @return The remaining duration as per {@link Timeout#remainingOrInfinite(TimeUnit)} if waiting ended early either
-         * spuriously or because of receiving a signal.
-         */
-        private long awaitNanos(final Condition condition, final long timeoutNanos) throws MongoInterruptedException {
```
katcharov, line 1172: This has been moved into the Deadline.

driver-core/src/main/com/mongodb/internal/time/Deadline.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+/**
+ * @see TimePoint
+ */
+public interface Deadline {
```
katcharov, line 28: I created the Deadline and StartPoint interfaces on TimePoint. These hide TimePoint methods that are not relevant or used, and also make it clear what "type of" TimePoint we are using.

driver-core/src/main/com/mongodb/internal/time/TimePoint.java
```diff
@@ -17,72 +17,171 @@
...
+    public Deadline asDeadline() {
+        return this;
+    }
+
+    /**
+     * The number of whole time units that remain until this TimePoint
+     * {@link #hasExpired()}. This should not be used to check for expiry,
+     * but can be used to supply a remaining value, in the finest-grained
+     * TimeUnit available, to some method that may time out.
+     * This method must not be used with infinite TimePoints.
```
katcharov, line 101: This changes the behaviour of the `remaining` method.

Previously, if there were 100ns remaining, and you asked how many hours were remaining, it would return 1 hour, because returning 0 might suggest that the deadline is expired. However, this does not match the behaviour of Duration, and I do not think rounding up is correct, for the expected usage. For example, I might get the remaining time in hours because that is the granularity of some method that I will invoke, and get back 1, and wait on the method for the next hour, which is not correct. As far as "hours" go, the deadline is indeed expired, and I do not think we should undertake operations that might (because of granularity differences) exceed the timeout. The same logic applies to milliseconds, though less intuitively.

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -267,20 +267,20 @@ public void monitorsSleepAtLeastMinHeartbeatFreqencyMSBetweenChecks() {
...
     }
 
     private static void assertPoll(final BlockingQueue<?> queue, @Nullable final Class<?> allowed, final Set<Class<?>> required,
-                                   final Timeout timeout) throws InterruptedException {
+                                   final Deadline deadline) throws InterruptedException {
         Set<Class<?>> encountered = new HashSet<>();
         while (true) {
             Object element;
-            if (timeout.isImmediate()) {
+            if (deadline.hasExpired()) {
```
katcharov, line 278: This removes the concept of "immediate" timepoints: they are just expired.

driver-core/src/main/com/mongodb/internal/time/StartTime.java
```diff
@@ -0,0 +1,59 @@
...
+     * the operation in question at least once.</p>
+     * <p>
+     * Note that the contract of this method is also used in some places to
+     * specify the behavior of methods that accept {@code (long timeout, TimeUnit unit)},
+     * e.g., {@link com.mongodb.internal.connection.ConcurrentPool#get(long, TimeUnit)},
+     * so it cannot be changed without updating those methods.</p>
+     *
+     * @see TimePoint#fromNowOrInfiniteIfNegative(long, TimeUnit)
+     */
+    Deadline fromNowOrInfiniteIfNegative(long timeoutValue, TimeUnit timeUnit);
```
katcharov, line 51: I think perhaps this method should not be added, but I have included it because otherwise, it seems I would need to change existing locations that take long/TimeUnit in this PR. I have at least made the name long and noticeable.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
rozza: Ack

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
rozza: Ack

driver-core/src/main/com/mongodb/internal/time/TimePoint.java
rozza: Good catch

driver-core/src/main/com/mongodb/internal/time/Deadline.java
```diff
@@ -0,0 +1,98 @@
...
+     * @param condition the condition.
+     */
+    default void awaitOn(final Condition condition) {
+        try {
+            if (isInfinite()) {
+                condition.await();
+            } else {
+                // the deadline will track this remaining time
+                //noinspection ResultOfMethodCallIgnored
+                condition.awaitNanos(remaining(NANOSECONDS));
```
rozza, line 92: I think the documentation should be clarified to ensure that `hasExpired` is checked after calling `awaitOn` as the condition may not be satisfied with in the deadline.

rozza (changes requested):
Looking good - would be worthwhile having @stIncMale review as well as he authored the original Timeout class.

driver-core/src/main/com/mongodb/internal/time/TimePoint.java
katcharov: I spoke with @stIncMale about this, and he explained that the existing rounding-up behaviour is generally consistent with the timeout interpretation elsewhere in Java, of "try at least this long and then time out as fast as possible". I agree this is not an incorrect interpretation, but we also agree that the appropriate way to address this particular issue is to use nanoseconds (as Java SE does), rather than rounding the value up and down to some granularity. I won't address this in this PR (that is, refactoring our existing methods). I will leave this rounding-up interpretation, since it is consistent with how "elapsed" works, and the inconsistency with Java SE is not crucial (and should ideally be addressed in a different way regardless).

katcharov:
(From earlier discussion, to track responses to the points above.) 

It seems beneficial, whatever we do, to have a single implementation (more maintainable and testable), and also a single abstraction (a plain point in time), and my opinion is that these are the primary benefits. The outstanding point is whether we should differentiate two "types of" TimePoint (how we decide what counts as a single API is tangential). I believe there is little harm in differentiating because there is no significant API duplication between the interfaces, and a benefit is that the compiler now prevents us from confusing the two, or using methods from one unsuitable in the other (that is, TimePoints are always strictly either Deadlines or StartTimes).

katcharov:
I renamed Deadlines to Timeouts because, while the term "deadline" is clearer, it would be burdensome to change existing usages, and would conflict with public-facing terminology. This is a naming change only - Timeouts are still trimmed down, no longer have start-times built in, etc.

Also changed target to the CSOT feature branch.

driver-core/src/main/com/mongodb/internal/time/TimePoint.java
rozza: Ack

driver-core/src/main/com/mongodb/internal/ClientSideOperationTimeout.java
```diff
@@ -75,7 +75,7 @@ public boolean hasTimeoutMS() {
      * @return true if the timeout has been set and it has expired
      */
     public boolean expired() {
```
rozza, line 77: We should change this to hasExpired at some point - just to keep the nomenclature consistent. Doesn't have to be done in this PR.

driver-core/src/main/com/mongodb/internal/time/StartTime.java
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mongodb.internal.time;
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * @see TimePoint
```
rozza, line 22: I think perhaps there should be more docs - just explaining why this interface exists.

As `TimePoint.now` could be added and all functionality is covered by Timepoint.

driver-core/src/test/unit/com/mongodb/internal/time/TimeoutTest.java
```diff
@@ -1,179 +0,0 @@
...
-
-import static java.util.concurrent.TimeUnit.NANOSECONDS;
-import static org.junit.jupiter.api.Assertions.assertAll;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-
-final class TimeoutTest {
```
rozza, line 36: Are all these test cases now in / or have equivalents in TimePointTest?

rozza (changes requested):
Looks good - main comment is to add a little more back ground to StartTime but I like it :)

driver-core/src/test/unit/com/mongodb/internal/time/TimeoutTest.java
katcharov: I went through each and moved tests from one to the other, removing those (or, certain assertions) that no longer applied.

driver-core/src/main/com/mongodb/internal/time/StartTime.java
katcharov: Added. Both StartTime and Deadline are intended to remove functionality to provide the (now-documented) typesafe guarantees, and to clarify the purpose or intended use of any value of that type.

driver-core/src/main/com/mongodb/internal/time/Deadline.java
katcharov: I added docs. It seems it is usually checked before calling await (though, this is generally in a loop).

I considered changing the method to perform the check, but there were some complications that I ran into elsewhere (that might be noted in a future PR).

rozza:
@katcharov I think you need to push a commit?

rozza (approved):
LGTM

PR #1196, opened 2023-09-12
Run changeStreamSplitLargeEvent test on >= 6.0.9

katcharov:
JAVA-5078

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/ChangeStreamProseTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/ChangeStreamProseTest.java
@@ -28,6 +28,7 @@ import com.mongodb.client.model.CreateCollectionOptions;
 import com.mongodb.client.model.changestream.ChangeStreamDocument;
 import com.mongodb.client.model.changestream.FullDocumentBeforeChange;
 import com.mongodb.client.model.changestream.SplitEvent;
+import com.mongodb.connection.ServerVersion;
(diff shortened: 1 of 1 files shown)


driver-sync/src/test/functional/com/mongodb/client/ChangeStreamProseTest.java
```diff
@@ -408,7 +410,7 @@ public void test16GetResumeTokenReturnsIdOfPreviousDocument() {
 
     @Test
     public void test19SplitChangeStreamEvents() {
-        assumeTrue(serverVersionAtLeast(7, 0));
+        assumeTrue(getServerVersion().compareTo(new ServerVersion(asList(6, 0, 9))) >= 0);
```
jyemin, line 413: Generally we assume latest patch version in these checks.  So it's fine to just check for 6.0.

Confirmed that CI is configured to use 6.0.9 already: https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/download-mongodb.sh#L61

jyemin (changes requested):

stIncMale (approved):

driver-sync/src/test/functional/com/mongodb/client/ChangeStreamProseTest.java
katcharov: Thanks, updated

jyemin (approved):

PR #1197, opened 2023-09-13
Scale double values to 4 decimal digits in geoNear test

jyemin:
This works around small variations in native math libraries used by the server.

JAVA-5097

### DIFF
--- a/driver-core/src/test/functional/com/mongodb/client/model/OperationTest.java
+++ b/driver-core/src/test/functional/com/mongodb/client/model/OperationTest.java
@@ -23,6 +23,8 @@ import com.mongodb.internal.connection.ServerHelper;
 import com.mongodb.lang.Nullable;
 import org.bson.BsonArray;
 import org.bson.BsonDocument;
+import org.bson.BsonDouble;
+import org.bson.BsonValue;
 import org.bson.Document;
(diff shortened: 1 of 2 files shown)


driver-core/src/test/functional/com/mongodb/client/model/OperationTest.java
```diff
@@ -110,6 +115,37 @@ protected void assertResults(final List<Bson> pipeline, final String expectedRes
         assertEquals(expectedResults, results);
     }
 
+    protected void assertResults(final List<Bson> pipeline, final String expectedResultsAsString,
+            final int scale, final RoundingMode roundingMode) {
+        List<BsonDocument> expectedResults = parseToList(expectedResultsAsString);
+        List<BsonDocument> results = getCollectionHelper().aggregate(pipeline);
+        assertEquals(adjustScale(expectedResults, scale, roundingMode), adjustScale(results, scale, roundingMode));
```
katcharov, line 122: Looks good. Another option would have been to perform recursive equality, with an epsilon on floating points. But I don't think we'll have an issue in these tests since the decimal places aren't close to being 0 (which would make the flooring flaky).

katcharov (approved):
LGTM!

PR #1198, opened 2023-09-13
Refactored Command Batch Cursors.

rozza:
The construction and resource management of cursors has been simplified by reducing the number of constructors used when creating the cursor.  The aim of this work was to simplify future refactorings eg CSOT work. 

Unfortunately, it expanded into a bigger piece of work than initially thought as, I took the opportunity to unify the general approaches between sync and async cursors. (With the caveat that async cursors are self closing).

The core top level changes are:

1. Added `CommandCursorResult` which replaces `QueryResult` and is used by both async and sync.

Synchronous changes:

1. `CommandBatchCursor` - Replaces `QueryBatchCursor` and is used for any commands that return a cursor object. This class contains all the logic to manage resources and issue get more calls.
2. `SingleBatchCursor` - Used for any commands that return a single list of results but not an actual cursor. There already was an `AsyncSingleBatchCursor` so brings abstractions inline.

Asynchronous changes:

1. `AsyncCommandBatchCursor` - Replaces `AsyncQueryBatchCursor` and used for any commands that return a cursor object. This class contains all the logic to manage resources and issue get more calls. This was a bigger refactor and the class now it closely follows the logic of the synchronous `SingleBatchCursor`. The core difference being its self closing and uses a callback loop rather than a while loop when calling `getMore`.

The tests were updated so that both async and sync cursors are tested equally (previously async had much more comprehensive testing). This process discovered some subtle sync bugs where resources wouldn't be released during the error scenarios. So the PR includes some potential memory leak fixes.

JAVA-5159


driver-core/src/main/com/mongodb/internal/async/AsyncBatchCursor.java
```diff
@@ -28,9 +30,9 @@
  */
 public interface AsyncBatchCursor<T> extends Closeable {
     /**
-     * Returns the next batch of results.  A tailable cursor will block until another batch exists.  After the last batch, the next call
-     * to this method will execute the callback with a null result to indicate that there are no more batches available and the cursor
-     * has been closed.
+     * Returns the next batch of results. A tailable cursor will block until another batch exists.
+     * Unlike the {@link BatchCursor} this method will automatically mark the cursor as closed when there are no more expected results.
```
rozza, line 34: The cursor no longer calls `callback.onResult(null, null)`. As this class is internal and consumed internally there is no public API breakage or behavour change.

driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    private void getMore(final AsyncConnection connection, final ServerCursor cursor, final SingleResultCallback<List<T>> callback) {
+        connection.commandAsync(namespace.getDatabaseName(),
+                getMoreCommandDocument(cursor.getId(), connection.getDescription(), namespace,
+                        limit, batchSize, count.get(), maxTimeMS, comment),
+                NO_OP_FIELD_NAME_VALIDATOR, ReadPreference.primary(),
+                CommandResultDocumentCodec.create(decoder, NEXT_BATCH), assertNotNull(resourceManager.connectionSource),
+                (commandResult, t) -> {
+                    if (t != null) {
```
rozza, line 205: I used multiple returns to make the code more readable and mimic the sync returns, even though we are using callbacks.

driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
```diff
@@ -0,0 +1,524 @@
...
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.NEXT_BATCH;
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.NO_OP_FIELD_NAME_VALIDATOR;
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.getMoreCommandDocument;
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.translateCommandException;
+import static com.mongodb.internal.operation.OperationHelper.LOGGER;
+import static java.lang.String.format;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+
+class AsyncCommandBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T> {
```
rozza, line 70: This follows the sync CommandBatchCursor approach and utilizes a ResourceManager. As such its very different to the previous `AsyncQueryBatchCursor` which contained more branching logic. 

I hope that now working on sync or async cursors in the future will be much easier on the developer.

katcharov (changes requested):
Most of the PR is a large but fairly straightforward refactor, and looks good. Still in the process of reviewing the CommandBatchCursor classes, but wanted to see what you thought about consolidating the code further.

driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
```diff
@@ -0,0 +1,524 @@
...
+
+    private final AtomicInteger count = new AtomicInteger();
+    private final AtomicBoolean processedInitial = new AtomicBoolean();
+    private final AtomicReference<CommandCursorResult<T>> commandCursorResult = new AtomicReference<>();
+    private final int maxWireVersion;
+    private final boolean firstBatchEmpty;
+    private final ResourceManager resourceManager;
+    private int batchSize;
+
+    AsyncCommandBatchCursor(
```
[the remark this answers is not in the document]
rozza: It could be made the same, however, there is no async Iterator API (which is why Publishers / Flows were invented and subsequently added to the jdk).

For async the API would be very heavy / require nested callbacks for use:

```
hasNext(SingleResultCallback<Boolean>)
next(SingleResultCallback<List<T>>)
```

Which can be replaced by just:

```
next(SingleResultCallback<List<T>>)
isClosed()
```

Mongodb cursors are self describing and have reached the end of their results.

The other main difference it the while loop used for getMores works for sync but in async it becomes a nested callback loop.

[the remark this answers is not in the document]
rozza: That said there is no reason `hasNext()` could be added instead of `isClosed()`. I'll take a look and try to increase the code reuse.

driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
```diff
@@ -0,0 +1,524 @@
...
+    private final MongoNamespace namespace;
+    private final int limit;
+    private final Decoder<T> decoder;
+    private final long maxTimeMS;
+    @Nullable
+    private final BsonValue comment;
+
+    private final AtomicInteger count = new AtomicInteger();
+    private final AtomicBoolean processedInitial = new AtomicBoolean();
+    private final AtomicReference<CommandCursorResult<T>> commandCursorResult = new AtomicReference<>();
```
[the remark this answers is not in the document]
rozza: No updated.

driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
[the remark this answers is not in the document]
rozza: I opted to leave as is, as the only consumer is internal (BatchCursorFlux) it has a check in recurse cursor:

```
if (batchCursor.isClosed()) {
  sink.complete();
} else {
  // Fetches more results.
```

driver-core/src/test/unit/com/mongodb/internal/operation/CommandBatchCursorSpecification.groovy
```diff
@@ -0,0 +1,645 @@
...
+import org.bson.Document
+import org.bson.codecs.DocumentCodec
+import spock.lang.Specification
+
+import static com.mongodb.ReadPreference.primary
+import static com.mongodb.internal.operation.OperationUnitSpecification.getMaxWireVersionForServerVersion
+
+class CommandBatchCursorSpecification extends Specification {
+
+    def 'should generate expected command with batchSize and maxTimeMS'() {
```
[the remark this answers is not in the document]
rozza: Looking into it.

rozza:
Changing to draft while I investigate the build errors.

driver-core/src/test/unit/com/mongodb/internal/operation/CommandBatchCursorSpecification.groovy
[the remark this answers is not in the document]
rozza: Done

rozza:
Ready for review again @katcharov

driver-core/src/main/com/mongodb/internal/operation/SingleBatchCursor.java
```diff
@@ -0,0 +1,95 @@
...
+    private boolean hasNext;
+
+    SingleBatchCursor(final List<T> batch, final int batchSize, final ServerAddress serverAddress) {
+        this.batch = batch;
+        this.serverAddress = serverAddress;
+        this.batchSize = batchSize;
+        this.hasNext = !batch.isEmpty();
+    }
+
+    public List<T> getBatch() {
```
jyemin, line 44: This method is unused, even in tests.  Let's remove it.

driver-core/src/main/com/mongodb/internal/operation/ListIndexesOperation.java
```diff
@@ -127,8 +124,8 @@ public BatchCursor<T> execute(final ReadBinding binding) {
                     return createReadCommandAndExecute(retryState, binding, source, namespace.getDatabaseName(), getCommandCreator(),
                             createCommandDecoder(), transformer(), connection);
                 } catch (MongoCommandException e) {
-                    return rethrowIfNotNamespaceError(e, createEmptyBatchCursor(namespace, decoder,
-                            source.getServerDescription().getAddress(), batchSize));
+                    return rethrowIfNotNamespaceError(e,
+                            SingleBatchCursor.createEmptyBatchCursor(source.getServerDescription().getAddress(), batchSize));
```
jyemin, line 128: We generally use static imports for spots like this.

driver-core/src/main/com/mongodb/internal/operation/SingleBatchCursor.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.mongodb.ServerAddress;
+import com.mongodb.ServerCursor;
+
+import java.util.List;
+import java.util.NoSuchElementException;
+
+import static java.util.Collections.emptyList;
+
+class SingleBatchCursor<T> implements BatchCursor<T> {
```
jyemin, line 26: Let's add a unit test for this class.

driver-core/src/main/com/mongodb/internal/operation/CommandBatchCursor.java
```diff
@@ -0,0 +1,383 @@
...
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.MESSAGE_IF_CLOSED_AS_CURSOR;
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.MESSAGE_IF_CLOSED_AS_ITERATOR;
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.NEXT_BATCH;
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.NO_OP_FIELD_NAME_VALIDATOR;
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.getMoreCommandDocument;
+import static com.mongodb.internal.operation.CommandBatchCursorHelper.translateCommandException;
+import static com.mongodb.internal.operation.OperationHelper.LOGGER;
+import static java.lang.String.format;
+
+class CommandBatchCursor<T> implements AggregateResponseBatchCursor<T> {
```
jyemin, line 58: I'm concerned that, as this appears as a new class in the diff, it's difficult to compare with `QueryBatchCursor`. Can you share some details about how this code was created?  Was it mostly an auto-refactoring? Are our existing tests good enough to ensure that no regressions were introduced?

jyemin (commented):
I tried to review but got stuck with the difficulty of comparing CommandBatchCursor to QueryBatchCursor without the benefit of a diff.

Is this refactoring going to make CSOT significantly simpler? It's not clear to me that it will, but if so, we should proceed. If not I would prefer to defer it.

driver-core/src/main/com/mongodb/internal/operation/CommandBatchCursor.java
rozza: Unfortunately, gits diff algorithm doesnt always get it right. 

This was done in multiple parts 
1) a simple rename. 
2) Updated tests - so both async and sync have the same tests. This lead to some fixes for cases where resources weren't released
3) A refactor to make the ResourceManager abstract and reusable across sync and async as per previous pr feedback.

driver-core/src/main/com/mongodb/internal/operation/SingleBatchCursor.java
rozza: Done

driver-core/src/main/com/mongodb/internal/operation/SingleBatchCursor.java
rozza: Done

driver-core/src/main/com/mongodb/internal/operation/ListIndexesOperation.java
rozza: Renamed the methods and done.

rozza:
Will make CSOT simpler because its now clear where timeout context is needed. 

It also brings the async and sync command cursors much more into line with lots of code reuse. 
It also brings a single batch cursor to sync also bringing async / sync into line.

PR #1200, opened 2023-09-15
Introduce builder for `$vectorSearch` aggregation stage

stIncMale:
JAVA-5117


driver-core/src/main/com/mongodb/client/model/Filters.java
```diff
@@ -86,7 +84,7 @@ public static <TItem> Bson eq(@Nullable final TItem value) {
      * @mongodb.driver.manual reference/operator/query/eq $eq
      */
     public static <TItem> Bson eq(final String fieldName, @Nullable final TItem value) {
-        return new SimpleEncodingFilter<>(fieldName, value);
+        return new OperatorFilter<>("$eq", fieldName, value);
```
stIncMale, line 87: `$vectorSearch` does not accept the short form of `$eq`, which we are currently using. We have the following options:

1. Change the implementation of `Filters.eq`, which is what I did. See more on this change below.
2. Introduce a new `Filters.fullEq` (I am not sure how to name it), and document that users are not to use `Filters.eq` with `VectorSearchOptions.filter`. This is really ugly.
3. Introduce a whole new API for filters. Even if we ever do this, we will definitely not do this as part of the current task.

This change is controversial. On one hand, nobody should care whether we use a full form or not. On the other hand, for some reason we documented that we are not using the full form. This, combined with the [regex behavior](https://www.mongodb.com/docs/manual/reference/operator/query/eq/#match-a-regular-expression), allows users to abuse `Filters.eq` to dynamically use either equality, or regex matching, even though the method is clearly intended only for equality (`Filters.regex` is intended for regex matching). This makes the change breaking, but I am still willing to do it.

stIncMale: Not supporting the short `$eq` form was intentional. Atlas will support in in the future, but not now.

driver-core/src/test/functional/com/mongodb/client/model/search/AggregatesSearchIntegrationTest.java
```diff
@@ -183,24 +201,105 @@
...
....0023810994, -0.0038648755, 0.0015467904, -0.0037136413, 0.0023458113, 0.019008467, -0.011547576, -0.010001626, 0.012347437, 0.0155267175, 0.01907568, -0.003041489, -0.0132414, 0.017449073, 0.00060073606, -0.008536334, 0.008233866, -0.0085430555, -0.02365976, 0.024089938, -0.0034615842, -0.006580371, 0.008327967, -0.01509654, 0.009692436, 0.025635887, 0.0020282194, -0.04022159, -0.0021290423, -0.012407931, -0.0021727323, 0.006506434, -0.005320085, -0.008240587, 0.020984594, -0.014491603, 0.003592654, 0.0072121937, -0.03081146, 0.043770555, 0.009302587, -0.003217929, 0.019008467, -0.011271994, 0.02917141, 0.0019576435, -0.0077431942, -0.0030448497, -0.023726976, 0.023377456, -0.006382086, 0.025716545, -0.017341528, 0.0035556855, -0.019129453, -0.004311857, -0.003253217, -0.014935223, 0.0036363439, 0.018121226, -0.0066543072, 0.02458733, 0.0035691285, 0.0039085653, -0.014209299, 0.020191453, 0.0357585, 0.007830574, -0.024130266, -0.008912739, 0.008314524, -0.0346024, -0.0014005973, -0.006788738, -0.021777734, 0.010465411, -0.004012749, -0.00679882, 0.009981461, -0.026227381, 0.027033964, -0.015567047, -0.0063115098, 0.0023071626, 0.01037131, 0.015741806, -0.020635074, -0.012945653));
```
stIncMale, line 229: The only sample data we have has 1500+ dimensions, which is why the query vector is so small.

driver-core/src/main/com/mongodb/client/model/Filters.java
jyemin: Can we annotate this as Beta, and remove it once Atlas supports the short form?

stIncMale: Alexander Lukyanchikov said that allAtlas clusters will get the updated behavior, unless they are "pinned on an older version for some reason" (I am not even sure if this can be done by a user). So this should work.

stIncMale: A ticket to remove `Filters.eqFull` https://jira.mongodb.org/browse/JAVA-5174.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -933,6 +938,90 @@ public static Bson searchMeta(final SearchCollector collector, final SearchOptio
...
+     *
+     * @mongodb.atlas.manual atlas-vector-search/vector-search-stage/ $vectorSearch
+     * @mongodb.atlas.manual atlas-search/scoring/ Scoring
+     * @mongodb.server.release 7.1
+     * @since 4.11
+     */
+    @Beta(Beta.Reason.SERVER)
+    public static Bson vectorSearch(
+            final FieldSearchPath path,
+            final Iterable<Double> queryVector,
```
katcharov, line 961: Should we use Collection? I believe an iterable can be infinite, which does not seem to fit here.

driver-core/src/test/unit/com/mongodb/client/model/AggregatesSpecification.groovy
```diff
@@ -847,6 +849,58 @@ class AggregatesSpecification extends Specification {
         }''')
     }
 
+    def 'should render $vectorSearch'() {
```
katcharov, line 852: Could we write these new tests in Java? We have been doing this (see `AggregatesTest.java`) for new test cases, and sometimes even when making changes to existing ones.

driver-core/src/test/functional/com/mongodb/client/model/search/AggregatesSearchIntegrationTest.java
```diff
@@ -183,24 +201,105 @@
...
         collectionHelpers = new HashMap<>();
         collectionHelpers.put(MFLIX_MOVIES_NS, new CollectionHelper<>(new BsonDocumentCodec(), MFLIX_MOVIES_NS));
+        collectionHelpers.put(MFLIX_EMBEDDED_MOVIES_NS, new CollectionHelper<>(new BsonDocumentCodec(), MFLIX_EMBEDDED_MOVIES_NS));
         collectionHelpers.put(AIRBNB_LISTINGS_AND_REVIEWS_NS, new CollectionHelper<>(new BsonDocumentCodec(), AIRBNB_LISTINGS_AND_REVIEWS_NS));
     }
 
     @BeforeEach
     void beforeEach() {
         assumeTrue(isAtlasSearchTest());
+        assumeTrue(serverVersionAtLeast(7, 1));
```
katcharov, line 243: Is this intended to be the case for all tests, or just the new vector search tests?

katcharov (changes requested):
Looking good

driver-core/src/test/unit/com/mongodb/client/model/AggregatesSpecification.groovy
stIncMale: `AggregatesTest` contains integration tests, it appears to be a successor of `AggregatesFunctionalSpecification`. `AggregatesSpecification.groovy` contains unit tests, and this PR added unit tests to this file.

driver-core/src/test/functional/com/mongodb/client/model/search/AggregatesSearchIntegrationTest.java
stIncMale: Thanks, it's not. Done.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -933,6 +938,90 @@ public static Bson searchMeta(final SearchCollector collector, final SearchOptio
...
     }
 
+    /**
+     * Creates a {@code $vectorSearch} pipeline stage supported by MongoDB Atlas.
+     * You may use the {@code $meta: "vectorSearchScore"} expression, e.g., via {@link Projections#metaVectorSearchScore(String)},
+     * to extract the relevance score assigned to each found document.
+     *
+     * @param queryVector The query vector.
+     * @param path The field to be searched.
+     * @param index The name of the index to use. This index predetermines the number of components of {@code queryVector}.
```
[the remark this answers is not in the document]
stIncMale: Done.

driver-core/src/main/com/mongodb/client/model/Aggregates.java
stIncMale: To me, the usefulness of restricting the type of `queryVector` from `Iterable` to `Collection` is on par with restricting the type of the argument in `Aggregates.unset` from `List` to `Set` - won't help a user, but may force him to convert a value to a different type before passing it (I don't know why it's a `List`, and not `Collection` or even `Iterable`).


P.S. An assumptions about `Collection` being finite is questionable. When `java.util.stream` was introduced, the authors decided to write _"While collections have a finite size, streams need not." in the documentation of this package. However, I fail to find something in the documentation of `Collection` in support of this. `Collection.size` is documented as _""If this collection contains more than `Integer.MAX_VALUE` elements, returns `Integer.MAX_VALUE`._, which is not inconsistent with the size being infinite.

katcharov: Alright - I see it is also consistent with what we do elsewhere.

katcharov (approved):
LGTM

driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -933,6 +938,90 @@ public static Bson searchMeta(final SearchCollector collector, final SearchOptio
...
+     * @param queryVector The query vector. The number of dimensions must match that of the {@code index}.
+     * @param path The field to be searched.
+     * @param index The name of the index to use.
+     * @param numCandidates The number of candidates.
+     * @param limit The limit on the number of documents produced by the pipeline stage.
+     * @return The {@code $vectorSearch} pipeline stage.
+     *
+     * @mongodb.atlas.manual atlas-vector-search/vector-search-stage/ $vectorSearch
+     * @mongodb.atlas.manual atlas-search/scoring/ Scoring
+     * @mongodb.server.release 7.1
```
vbabanin, line 955: I noticed that this feature was also backported to 6.0.10 as mentioned in this Jira ticket (https://jira.mongodb.org/browse/PM-3407).

I recall a similar instance where we specified the backported version in @mongodb.server.release for [one](https://github.com/mongodb/mongo-java-driver/pull/1159/commits/019922394451b92ed8a823b913c654c916e533fd) of the previous PRs. Considering this, should we include the backported version here as well?

stIncMale: Done.

vbabanin (approved):

PR #1202, opened 2023-09-22
Refactor to use Timeout for server selection

katcharov:
JAVA-5175

This uses Timeout for timeout logic, and ensures that settings are obtained from one location, for consistency with how settings will be obtained from the recently-merged TimeoutSettings (though, this is not used, as it is not yet propagated into OperationContext).

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -18,6 +18,7 @@ package com.mongodb.internal.connection;
 
 import com.mongodb.MongoClientException;
 import com.mongodb.MongoIncompatibleDriverException;
+import com.mongodb.MongoInterruptedException;
 import com.mongodb.MongoTimeoutException;
 import com.mongodb.ServerAddress;
 import com.mongodb.connection.ClusterDescription;
@@ -35,6 +36,8 @@ import com.mongodb.internal.async.SingleResultCallback;
 import com.mongodb.internal.diagnostics.logging.Logger;
 import com.mongodb.internal.diagnostics.logging.Loggers;
 import com.mongodb.internal.selector.LatencyMinimizingServerSelector;
+import com.mongodb.internal.time.StartTime;
+import com.mongodb.internal.time.Timeout;
 import com.mongodb.lang.Nullable;
 import com.mongodb.selector.CompositeServerSelector;
 import com.mongodb.selector.ServerSelector;
@@ -59,7 +62,6 @@ import static com.mongodb.connection.ServerDescription.MIN_DRIVER_WIRE_VERSION;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
 import static com.mongodb.internal.connection.EventHelper.wouldDescriptionsGenerateEquivalentEvents;
 import static com.mongodb.internal.event.EventListenerHelper.singleClusterListener;
-import static com.mongodb.internal.thread.InterruptionUtil.interruptAndCreateMongoInterruptedException;
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
 import static java.util.Comparator.comparingInt;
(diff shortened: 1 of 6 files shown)


driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -280,19 +263,20 @@ private void updatePhase() {
...
+    private Timeout startServerSelectionTimeout(final StartTime startTime) {
+        long ms = settings.getServerSelectionTimeout(MILLISECONDS);
+        return startTime.timeoutAfterOrInfiniteIfNegative(ms, MILLISECONDS);
     }
 
-    private long getMinWaitTimeNanos() {
-        return serverFactory.getSettings().getMinHeartbeatFrequency(NANOSECONDS);
+    private Timeout startMinWaitHeartbeatTimeout() {
+        long minHeartbeatFrequency = serverFactory.getSettings().getMinHeartbeatFrequency(NANOSECONDS);
+        minHeartbeatFrequency = Math.max(0, minHeartbeatFrequency);
```
rozza, line 273: Nice - I dont think there is a negative check or minimum value for this.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -169,50 +157,45 @@ public void selectServerAsync(final ServerSelector serverSelector, final Operati
...
 
-            long startTimeNanos = System.nanoTime();
-            long curTimeNanos = startTimeNanos;
-            long maxWaitTimeNanos = getMaxWaitTimeNanos();
+        while (true) {
+            CountDownLatch currentPhaseLatch = phase.get();
+            currentDescription = description;
 
-            while (curDescription.getType() == ClusterType.UNKNOWN) {
+            if (currentDescription.getType() != ClusterType.UNKNOWN) {
```
rozza, line 170: Does this check then allows for a `0` value in server selection. If ones available return it else timeout?

driver-core/src/main/com/mongodb/internal/time/Timeout.java
```diff
@@ -94,7 +112,29 @@ default void awaitOn(final Condition condition) {
...
+     * @param action supplies the name of the action, for {@link MongoInterruptedException}
+     */
+    default void awaitOn(final CountDownLatch latch, final Supplier<String> action) {
+        try {
+            if (isInfinite()) {
+                latch.await();
+            } else {
+                // the timeout will track this remaining time
+                //noinspection ResultOfMethodCallIgnored
+                latch.await(this.remaining(NANOSECONDS), NANOSECONDS);
```
rozza, line 134: The result of the latch needs to be inspected - as if it returns false it awaited the time limit but wasn't counted down.

rozza (changes requested):

rozza:
Looking good - much easier to read, couple of comments / questions.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
katcharov: Yeah, anything negative seems like it is effectively 0. I'm not sure what good a value like 0 or 1 would be, but there's nothing preventing these.

I added a TODO that I'll mark in the ultimate CSOT PR to check that we don't accept negatives for timeouts.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
katcharov: Previously, the 3 usages, all in BaseCluster, that were refactored into this method did not inspect the result. (Perhaps they should have inspected the result, but I could not readily see the error.)

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
katcharov: I made a small change to bring this getDescription in line with selectServer, but basically yes there are only two paths out of the loop. Either we throw a timeout, or we return something, and here the check is for the cluster type becoming known.

driver-core/src/main/com/mongodb/internal/time/Timeout.java
rozza: Agreed - nothing to do here.

rozza (approved):
LGTM

PR #1203, opened 2023-09-25
Handle interrupts: `Socket.connect` is interruptible in a virtual thread

stIncMale:
Notably, `Socket.close` may also block if one enables `SO_LINGER` by calling `Socket.setSoLinger`, yet there is not a word on interruptibility of `Socket.close` in the docs. Fortunately, we don't enable `SO_LINGER`, so we don't have to care about the behavior of `Socket.close`.

JAVA-5138

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
@@ -83,8 +83,7 @@ public class SocketStream implements Stream {
         } catch (IOException e) {
             close();
             throw translateInterruptedException(e, "Interrupted while connecting")
-                    .<MongoException>map(Function.identity())
-                    .orElseGet(() -> new MongoSocketOpenException("Exception opening socket", getAddress(), e));
+                    .orElseThrow(() -> new MongoSocketOpenException("Exception opening socket", getAddress(), e));
         }
     }
 
--- a/driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
@@ -16,7 +16,6 @@
 
 package com.mongodb.internal.connection;
 
-import com.mongodb.MongoException;
 import com.mongodb.MongoSocketException;
 import com.mongodb.MongoSocketOpenException;
 import com.mongodb.MongoSocketReadException;
@@ -41,7 +40,6 @@ import java.net.SocketTimeoutException;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Function;
 
 import static com.mongodb.assertions.Assertions.assertTrue;
 import static com.mongodb.assertions.Assertions.notNull;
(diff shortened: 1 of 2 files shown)


driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
```diff
@@ -79,7 +82,9 @@ public void open() {
             inputStream = socket.getInputStream();
         } catch (IOException e) {
             close();
-            throw new MongoSocketOpenException("Exception opening socket", getAddress(), e);
```
jyemin, line 82: I think it's worth a ticket to make this consistent with read/write methods to throw IOException and do the exception transation in InternalStreamConnection.  It can be dependent on the ticket to remove Stream from the API.

jyemin (approved):
LGTM

driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
stIncMale: Created https://jira.mongodb.org/browse/JAVA-5179.

driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
```diff
@@ -240,7 +245,7 @@ public void close() {
             if (socket != null) {
                 socket.close();
             }
-        } catch (IOException e) {
+        } catch (IOException | RuntimeException e) {
```
vbabanin, line 248: We have another instance where the socket is closed [KeyManagementService.java](https://github.com/mongodb/mongo-java-driver/blob/b60f8723b453d88ab14df69d1e2cfc18f09b9887/driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java#L101). Should we also implement a defensive measure here as well?

driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
```diff
@@ -79,7 +82,9 @@ public void open() {
             inputStream = socket.getInputStream();
         } catch (IOException e) {
             close();
-            throw new MongoSocketOpenException("Exception opening socket", getAddress(), e);
+            throw translateInterruptedException(e, "Interrupted while connecting")
+                    .<MongoException>map(Function.identity())
+                    .orElseGet(() -> new MongoSocketOpenException("Exception opening socket", getAddress(), e));
```
vbabanin, line 87: nit:  
We might as well omit this line `.<MongoException>map(Function.identity())` by using: 
```
  throw translateInterruptedException(e, "Interrupted while connecting")
                    .orElseThrow(() -> new MongoSocketOpenException("Exception opening socket", getAddress(), e));
```

driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
vbabanin: Found this done in another [PR](https://github.com/mongodb/mongo-java-driver/pull/1204/files).

vbabanin (approved):

driver-core/src/main/com/mongodb/internal/connection/SocketStream.java
stIncMale: Nice! Done.

PR #1204, opened 2023-09-26
Handle interrupts: synchronous `KeyManagementService` uses Socket IO (open, read, write), which is interruptible in a virtual thread

stIncMale:
This PR is very similar to https://github.com/mongodb/mongo-java-driver/pull/1203, which is why I assigned the same reviewers.

JAVA-5139

### DIFF
--- a/driver-sync/src/main/com/mongodb/client/internal/Crypt.java
+++ b/driver-sync/src/main/com/mongodb/client/internal/Crypt.java
@@ -42,7 +42,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.util.Map;
-import java.util.function.Function;
 import java.util.function.Supplier;
 
 import static com.mongodb.assertions.Assertions.assertNotNull;
--- a/driver-sync/src/main/com/mongodb/client/internal/Crypt.java
+++ b/driver-sync/src/main/com/mongodb/client/internal/Crypt.java
@@ -350,7 +350,7 @@ public class Crypt implements Closeable {
             }
             cryptContext.completeKeyDecryptors();
         } catch (Throwable t) {
-            throw translateInterruptedException(t, "Interrupted while reading")
+            throw translateInterruptedException(t, "Interrupted while doing IO")
                     .<MongoException>map(Function.identity())
                     .orElseGet(() -> wrapInMongoException(t));
         }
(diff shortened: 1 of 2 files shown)


driver-sync/src/main/com/mongodb/client/internal/Crypt.java
```diff
@@ -348,7 +350,9 @@ private void decryptKeys(final MongoCryptContext cryptContext) {
             }
             cryptContext.completeKeyDecryptors();
         } catch (Throwable t) {
-            throw wrapInClientException(t);
+            throw translateInterruptedException(t, "Interrupted while reading")
```
jyemin, line 353: Is it necessary to translate interrupted exceptions both here and in the KeyManagementService (which this is calling)?

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
```diff
@@ -98,8 +101,21 @@ private void enableHostNameVerification(final SSLSocket socket) {
...
         }
     }
+
+    /**
+     * @return Never.
+     */
+    private static RuntimeException translateInterruptedExceptionAndThrow(final IOException e, final String message) throws IOException,
+            MongoInterruptedException {
+        Optional<MongoInterruptedException> interruptedException = translateInterruptedException(e, message);
+        if (interruptedException.isPresent()) {
```
jyemin, line 115: Why can't we use `orElseGet` pattern here as in Crypt?

driver-sync/src/main/com/mongodb/client/internal/Crypt.java
stIncMale: Only if we want to add messages like "interrupted while connecting/writing/reading" which is what we do for streams, and why I did it this way here. But if it's fine to just have a single message "interrupted while doing IO", or no message at all, I'll be happy to remove the translating code from `Crypt`.

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
stIncMale: We can if we change `KeyManagementService.stream` from `throws IOException` to `throws Exception`. I don't want to do that. Fortunately, if we do https://github.com/mongodb/mongo-java-driver/pull/1204/files#r1337520600, this change in `KeyManagementService` will be gone.

driver-sync/src/main/com/mongodb/client/internal/Crypt.java
stIncMale: Done.

jyemin (approved):

vbabanin (approved):
LGTM

PR #1208, opened 2023-09-28
Fix the `ClusterListener`, `ServerListener`, `ServerMonitorListener` API docs relevant to thread-safety

stIncMale:
JAVA-5185

### DIFF
--- a/driver-core/src/main/com/mongodb/event/ClusterListener.java
+++ b/driver-core/src/main/com/mongodb/event/ClusterListener.java
@@ -21,10 +21,11 @@ import java.util.EventListener;
 /**
  * A listener for cluster-related events.
  * <p>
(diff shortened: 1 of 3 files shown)


jyemin (approved):

rozza (commented):
I'm not sure the average Java developer will know what "writes to shared memory" means in relation to their Java app.

How about something like:

> However, this guarantee does not mean that implementations automatically do not need to synchronize memory accesses to prevent data races. For example, if any data that the listener collects is accessed oustide of the normal execution of the listener by the {@code MongoClient}, then any reading and writing actions must be synchronized.

rozza (approved):
My suggestion is optional, if you have similar concerns.

So in essence this is a LGTM.

stIncMale:
@rozza, I implemented you suggestion, thank you!

Merging, because the static checks have passed, and everything else is irrelevant for the changes in this PR.

PR #1210, opened 2023-10-02
fix: include Kotlin sources into the sources jar

monosoul:
Configures `sourcesJar` task for Kotlin projects to actually include Kotlin sources, otherwise jar files are empty.

See [mongodb-driver-kotlin-sync-4.10.2-sources.jar](https://repo1.maven.org/maven2/org/mongodb/mongodb-driver-kotlin-sync/4.10.2/mongodb-driver-kotlin-sync-4.10.2-sources.jar) for example.

Also see this for context: https://youtrack.jetbrains.com/issue/KT-54207

### DIFF
--- a/bson-kotlin/build.gradle.kts
+++ b/bson-kotlin/build.gradle.kts
@@ -143,3 +143,10 @@ tasks.javadocJar.configure {
     archiveClassifier.set("javadoc")
     from(dokkaOutputDir)
 }
+
+// ===========================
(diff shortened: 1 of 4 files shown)


monosoul:
Without a proper `sources.jar` it's kinda hard to use the new driver now, since this is what you'll see in IntelliJ:

image

rozza:
Many thanks @monosoul :)

PR #1212, opened 2023-10-03
Kotlin. Support stored nulls for nullable fields

rozza:
JAVA-5134

### DIFF
--- a/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt
+++ b/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt
@@ -85,6 +85,8 @@ internal data class DataClassCodec<T : Any>(
                 if (logger.isTraceEnabled) {
                     logger.trace("Found property not present in the DataClass: $fieldName")
                 }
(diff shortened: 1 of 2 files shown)


bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt
```diff
@@ -85,6 +85,8 @@ internal data class DataClassCodec<T : Any>(
                 if (logger.isTraceEnabled) {
                     logger.trace("Found property not present in the DataClass: $fieldName")
                 }
+            } else if (propertyModel.param.type.isMarkedNullable && reader.currentBsonType == BsonType.NULL) {
+                reader.skipValue()
```
jyemin, line 89: Optional: `reader.readNull()` is a bit clearer in intent, and a bit more efficient as well.

jyemin (approved):

stIncMale (approved):

bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt
rozza: Ack

PR #1215, opened 2023-10-04
CSOT

katcharov:
### Description

This PR introduces a unified Client Side Operation Timeout (CSOT) as an Alpha feature to 
simplify the multitude of existing timeout settings. This consolidation merges various complex and interacting timeout settings into a single, overarching operation timeout option, enhancing usability and predictability. Existing timeout options will continue to be honored if the operation timeout is not set.

### Key Changes:

- Implement CSOT across various layers of the driver, applying the new operation 
  timeout configuration to govern execution time in areas such as Authentication, 
  Connection Pool and Monitoring, Server Discovery and Monitoring, CRUD, GridFS, 
  CSFLE, Transactions, Handshake, and Server Selection.

- Annotate the CSOT API as Alpha, indicating that this public API element is in 
  the early stages of development and may undergo incompatible changes.

- Centralize the setting of maxTimeMS just before sending commands to the server. 
  This optimization enhances performance and ensures more precise enforcement of timeouts.

- Introduce TimeoutContext to centralize timeout-related logic, enhancing the 
  clarity and ease of maintenance for timeout management within the driver.

- Enable a new retry mode when CSOT is active, allowing operations to retry until the timeout is reached. This improves resiliency compared to the previous behavior of retrying only once.

- Enhance MongoClient, MongoDatabase, MongoCollection, and ClientSession to support 
  setting and inheriting the operation timeout.

- Introduce MongoCluster, a new conceptual entity representing server deployments, 
  configurable with specific writeConcern, timeouts, and read preferences.

- Transform timeout errors from external sources into the new MongoOperationTimeoutException, 
  standardizing error handling when CSOT is enabled.

Note: `serverSelectionTimeoutMS` and `connectTimeoutMS` are still honored even if CSOT is enabled.

### Reviewers:
@rozza: Since you have made many of the changes in this branch, I have added you as a reviewer. I have addressed and answered some comments on this PR but would appreciate if you could double-check and comment if you have any objections. A full review of the PR is not expected.

@stIncMale: Could you please check if your previous comments have been addressed and resolve them as needed? A full review of the PR is not expected.

### For reviewers: 
- Test Verification: Ensure all tests are passing and all important build variants are tested.
- Verify that the `ClientSideOperation*` test suites does not have any ignored tests without a valid reason.

### Context:
This PR focuses on the alpha changes of CSOT and includes changes from the EPIC: [JAVA-3828](https://jira.mongodb.org/browse/JAVA-3828).

Description is updated by @vbabanin.


driver-core/src/main/com/mongodb/internal/time/Timeout.java
```diff
@@ -15,245 +15,127 @@
...
-    public boolean isImmediate() {
-        return equals(IMMEDIATE);
-    }
-
-    @Override
-    public boolean equals(final Object o) {
-        if (this == o) {
-            return true;
+    static Timeout expiresIn(final long duration, final TimeUnit unit) {
+        // TODO (CSOT) confirm that all usages in final PR are non-negative
```
katcharov, line 91: Note to address all TODOs prior to merging. (All files in github can be loaded using `document.querySelectorAll('.load-diff-button').forEach(node => node.click())` in console, and then ctrl-f for `TODO`)

driver-sync/src/test/functional/com/mongodb/client/csot/AbstractClientSideOperationsEncryptionTimeoutProseTest.java
```diff
@@ -0,0 +1,388 @@
...
+
+    private static final String MASTER_KEY = "Mng0NCt4ZHVUYUJCa1kxNkVyNUR1QURhZ2h2UzR2d2RrZzh0cFBwM3R6NmdWMDFBMUN3YkQ5a"
+            + "XRRMkhGRGdQV09wOGVNYUMxT2k3NjZKelhaQmRCZGJkTXVyZG9uSjFk";
+
+    protected abstract ClientEncryption createClientEncryption(ClientEncryptionSettings.Builder builder);
+
+    protected abstract MongoClient createMongoClient(MongoClientSettings.Builder builder);
+
+    @Test
+    @Tag("setsFailPoint")
```
stIncMale, line 96: I don't think this is a proper use of the `@Tag` functionality. We should either use something like `com.mongodb.client.FailPoint`, or implement automatic turning off of failpoints without using the `@Tag` functionality.

The original discussion is https://github.com/mongodb/mongo-java-driver/pull/1333#discussion_r1535047923.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/TimeoutHelper.java
```diff
@@ -0,0 +1,109 @@
...
+
+
+    public static MongoDatabase databaseWithTimeout(final MongoDatabase database,
+                                                    @Nullable final Timeout timeout) {
+        return databaseWithTimeout(database, DEFAULT_TIMEOUT_MESSAGE, timeout);
+    }
+
+    public static MongoDatabase databaseWithTimeout(final MongoDatabase database,
+                                                    final String message,
+                                                    @Nullable final Timeout timeout) {
```
katcharov, line 77: In these `databaseWithTimeout` methods, I would expect a null timeout to result in a database without a timeout, but below it returns the database itself.

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -0,0 +1,371 @@
...
+    public TimeoutSettings getTimeoutSettings() {
+        return timeoutSettings;
+    }
+
+    public long getMaxAwaitTimeMS() {
+        return timeoutSettings.getMaxAwaitTimeMS();
+    }
+
+    public long getMaxTimeMS() {
+        return notNull("Should never be null", maxTimeSupplier.get());
```
katcharov, line 204: This method is used only to set the `maxTimeMS` value in `CommandMessage#getExtraElements`, when non-zero. The value is:

1. By default, `calculateMaxTimeMS`: if the timeout not present, use the legacy `maxTimeMS` setting. Otherwise, shorten the timeout by the round trip time, throwing a timeout exception as needed.
2. In `CommitTransactionOperation`, the supplier is overridden to use `getMaxCommitTimeMS`, which uses the timeout if present, but otherwise uses the legacy `maxCommitTimeMS` setting.
3. Elsewhere, the supplier is overridden to return a fixed value of 0, denoting infinite, or to a fixed value like `MaxAwaitTimeMS`.

In the second path, when the `timeout` is present, it is not shortened by the round trip time. In the first, it is shortened, which seems correct. Should it be shortened in the second path?

In the third path, the value (0, or maxAwaitTime) is not overriden by timeout. In the first and second paths, fixed values are overridden by the timeout. Should they be overridden on this third path?

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -453,13 +470,17 @@ private static void attach(final RetryState retryState, final BulkWriteTracker t
...
             attempt = 0;
             attempts = retry ? RetryState.RETRIES + 1 : 1;
             this.batch = batch;
+            this.timeoutContext = timeoutContext;
         }
 
         boolean lastAttempt() {
+            if (timeoutContext.hasTimeoutMS()){
+                return timeoutContext.hasExpired();
+            }
```
katcharov, line 483: This change makes lastAttempt non-deterministic within a particular attempt. The following code calls lastAttempt:
```
if (currentBulkWriteTracker.lastAttempt()) {
    addRetryableWriteErrorLabel(writeConcernBasedError, maxWireVersion);
    addErrorLabelsToWriteConcern(result.getDocument("writeConcernError"), writeConcernBasedError.getErrorLabels());
} 
```
The intent is to add error labels on the last attempt. However, now, the above lastAttempt check can return "false", because the timeout has not yet timed out. So labels are not attached. At the end of the iteration, other code will check lastAttempt again, but now the timer will have run out, so no further iterations are performed, and labels are never attached. This appears to be a bug.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/MongoCollection.java
```diff
@@ -156,6 +188,22 @@ public interface MongoCollection<TDocument> {
...
+     *    <li>{@code > 0} The time limit to use for the full execution of an operation.</li>
+     * </ul>
+     *
+     * @param timeout the timeout, which must be greater than or equal to 0
+     * @param timeUnit the time unit
+     * @return a new MongoCollection instance with the set time limit for the full execution of an operation
+     * @since CSOT
+     * @see #getTimeout
+     */
+    MongoCollection<TDocument> withTimeout(long timeout, TimeUnit timeUnit);
```
katcharov, line 205: See https://github.com/mongodb/mongo-java-driver/pull/1349#discussion_r1538941852 - clarify regarding use of times below 1ms.

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    public int getConnectTimeoutMs() {
+        return (int) calculateMin(getTimeoutSettings().getConnectTimeoutMS());
+    }
+
+    public void resetTimeout() {
+        assertNotNull(timeout);
+        timeout = calculateTimeout(timeoutSettings.getTimeoutMS());
+    }
```
katcharov, line 246: Consider changing to `resetTimeoutIfPresent`, with an if check on null rather than the assert. All prod usages perform this check before calling.

driver-core/src/main/com/mongodb/internal/async/function/RetryState.java
```diff
@@ -150,19 +191,36 @@ private void doAdvanceOrThrow(final Throwable attemptException,
         if (onlyRuntimeExceptions) {
             assertTrue(isRuntime(attemptException));
         }
-        assertTrue(!isFirstAttempt() || exception == null);
-        Throwable newlyChosenException = callOnAttemptFailureOperator(exception, attemptException, onlyRuntimeExceptions, onAttemptFailureOperator);
-        if (isLastAttempt()) {
-            exception = newlyChosenException;
-            throw exception;
+        assertTrue(!isFirstAttempt() || previouslyChosenException == null);
+        Throwable newlyChosenException = callOnAttemptFailureOperator(previouslyChosenException, attemptException, onlyRuntimeExceptions, onAttemptFailureOperator);
```
katcharov, line 195: I believe this `previouslyChosenException` and the exception-wrapping logic below should be moved outside of RetryState, and into (I think?) the retryPredicate.

There might be a bug arising from this. This async retry API is intended to be general, and is used, for example, in OIDC retries. The exceptions thrown in that particular loop (whether auth exceptions, timeout exceptions, or otherwise) should either force a retry, or be thrown as-is. They should not interact with CSOT specific logic, like storage of previously chosen exceptions, or exception wrapping, as appears to be the case, even when an operation timeout is not set up in RetryState. This might (or might not) cause double-wrapping.

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
```diff
@@ -282,10 +287,12 @@ private void authenticationLoop(final InternalConnection connection, final Conne
     }
 
     private void authenticationLoopAsync(final InternalConnection connection, final ConnectionDescription description,
+            final OperationContext operationContext,
             final SingleResultCallback<Void> callback) {
         fallbackState = FallbackState.INITIAL;
         beginAsync().thenRunRetryingWhile(
-                c -> super.authenticateAsync(connection, description, c),
+                operationContext.getTimeoutContext(),
```
katcharov, line 294: (See prior comment).

I believe that no timeout context should be passed here, and that in general `thenRunRetryingWhile` should not take a timoutContext as a parameter.

driver-sync/src/test/functional/com/mongodb/client/csot/AbstractClientSideOperationsEncryptionTimeoutProseTest.java
vbabanin: The `@Tag` annotation was removed as part of [this PR](https://github.com/mongodb/mongo-java-driver/pull/1383/files#r1611734757). Now, we ensure failpoints are disabled idempotently after each test in the `@AfterEach` method.

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -0,0 +1,352 @@
...
+    private final TimeoutSettings timeoutSettings;
+
+    @Nullable
+    private Timeout timeout;
+    @Nullable
+    private Timeout computedServerSelectionTimeout;
+    private long minRoundTripTimeMS = 0;
+
+    public static MongoOperationTimeoutException createMongoTimeoutException() {
+        return createMongoTimeoutException("Remaining timeoutMS is less than the servers minimum round trip time.");
```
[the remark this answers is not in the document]
vbabanin: I believe we can address this as part of the upcoming async refactoring, given that it is currently used in the async code. Here's the relevant section for reference: [InternalStreamConnection.java#L640](https://github.com/mongodb/mongo-java-driver/blob/28c6df9f07f034302d1c6fe1e3113bb2a5a7859e/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java#L640).

driver-core/src/main/com/mongodb/internal/connection/netty/NettyStream.java
```diff
@@ -260,7 +252,13 @@ public void writeAsync(final List<ByteBuf> buffers, final AsyncCompletionHandler
             composite.addComponent(true, ((NettyByteBuf) cur).asByteBuf().retain());
         }
 
+        long writeTimeoutMS = operationContext.getTimeoutContext().getWriteTimeoutMS();
+        Optional<WriteTimeoutHandler> writeTimeoutHandler = writeTimeoutMS != NO_SCHEDULE_TIME
+                ? Optional.of(new WriteTimeoutHandler(writeTimeoutMS, MILLISECONDS)) : Optional.empty();
+        writeTimeoutHandler.map(w -> channel.pipeline().addBefore("ChannelInboundHandlerAdapter", "WriteTimeoutHandler", w));
```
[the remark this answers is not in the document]
vbabanin: Thank you for pointing this out. The `writeHandler` used in the lambda expression needs to be final or effectively final, which complicates using a simple if statement due to variable mutability. That’s why `Optional` was utilized here. I understand this can be confusing, so I've refactored this part to enhance readability. Please let me know if you have any suggestions on how to simplify it further without using `Optional`.

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
vbabanin: Thank you for the detailed question regarding the maxTimeMS settings across different contexts.

**First and Second Paths:** According to the "Command Execution" section in the [MongoDB client-side operations timeout specification](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#command-execution), `maxTimeMS` should be adjusted by the round trip time (RTT). This adjustment is correctly applied in the first path and should indeed be applied in the second path as well. I will create a ticket to address this oversight.

**Third Path**: The handling of `maxTimeMS` using `maxAwaitTimeMS` specifically for ChangeStreams and TailableAwait cursors is outlined as an exception in the specification:

>  If the maxAwaitTimeMS option is set, drivers MUST use it as the maxTimeMS field on getMore commands."

However, assigning `maxAwaitTimeMS` to `maxTimeMS `as a fixed value without considering RTT and the remaining timeout could lead to the connection churn problem as documented in [JIRA DRIVERS-2868](https://jira.mongodb.org/browse/DRIVERS-2868). This issue indicates a need to adjust `maxTimeMS` by RTT and the remaining timeout in these scenarios as well. We are awaiting a spec change to proceed with this adjustment.

In summary, yes, the `maxTimeMS` value should generally be shortened by the RTT to reflect accurate timeout behavior across different scenarios.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
vbabanin: Thank you for highlighting this issue. Indeed, there was non-deterministic behavior in how `lastAttempt` was handled, which we addressed in a recent update [here](https://github.com/mongodb/mongo-java-driver/pull/1418). With that change, we eliminated the race condition by ensuring `lastAttempt` consistently returns false when `timeoutMS` is present. However, following this update, error labels are never attached when CSOT is enabled. We need to investigate whether we can attach them on each iteration. I will create a ticket to examine this issue before our GA release.

driver-core/src/main/com/mongodb/MongoOperationTimeoutException.java
```diff
@@ -0,0 +1,58 @@
...
+ * <p>The key difference between {@code MongoOperationTimeoutException} and {@code MongoExecutionTimeoutException}
+ * lies in the nature of these exceptions. {@code MongoExecutionTimeoutException} indicates a server-side timeout
+ * capped by a user-specified number. These server errors are transformed into the new {@code MongoOperationTimeoutException}.
+ * On the other hand, {@code MongoOperationExecutionException} denotes a timeout during the execution of the entire operation.
+ *
+ * @see MongoClientSettings.Builder#timeout(long, TimeUnit)
+ * @see MongoClientSettings#getTimeout(TimeUnit)
+ * @since CSOT
+ */
+public final class MongoOperationTimeoutException extends MongoTimeoutException {
```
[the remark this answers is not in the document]
vbabanin: As suggested, I have reviewed and updated all timeout-related messages to ensure consistenc/. Now, they all use uniform terminology.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -680,7 +699,11 @@ private void updateSessionContext(final SessionContext sessionContext, final Res
         }
     }
 
-    private MongoException translateWriteException(final Throwable e) {
+    private MongoException translateWriteException(final Throwable e, final OperationContext operationContext) {
+        if (e instanceof MongoSocketWriteTimeoutException && operationContext.getTimeoutContext().hasExpired()) {
```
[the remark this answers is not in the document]
vbabanin: It has been changed to use `hasTimeoutMS` in this [PR](https://github.com/mongodb/mongo-java-driver/pull/1449).

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
vbabanin: Done in https://github.com/mongodb/mongo-java-driver/pull/1449.

driver-core/src/main/com/mongodb/internal/async/function/RetryState.java
vbabanin: Currently, the existing OIDC tests are passing without CSOT enabled, which suggests that we should not expect a regression. However, to address the concerns related to CSOT and its interactions with OIDC retries, we can further investigate and make necessary adjustments as part of the ongoing work in [JAVA-5337](https://jira.mongodb.org/browse/JAVA-5337).

driver-core/src/main/com/mongodb/internal/time/Timeout.java
vbabanin: I have addressed critical `//TODO` comments.

stIncMale (approved):
The only concern I left here was addressed.

rozza:
Kicked off a full build just to check.

https://spruce.mongodb.com/version/6698dcbb688ab00007faac91/

vbabanin:
The `MongoCluster` concept has been introduced in CSOT and is not marked with the `@Alpha` annotation, except for timeout-related methods like `getTimeout` and `withTimeout`. Although it originated in CSOT, it is applicable beyond this context, so the class` MongoCluster` has not been marked with `@Alpha`. If there are any objections, please share your feedback.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
```diff
@@ -114,10 +120,12 @@ public <T> Mono<T> execute(final AsyncWriteOperation<T> operation, final ReadCon
 
         return Mono.from(subscriber ->
                 clientSessionHelper.withClientSession(session, this)
-                        .map(clientSession -> getReadWriteBinding(getContext(subscriber), primary(), readConcern,
-                                clientSession, session == null && clientSession != null))
+                        .map(clientSession -> getReadWriteBinding(getContext(subscriber),
+                                primary(), readConcern, clientSession, session == null))
+                        // TODO (CSOT) - does this get called - can the above return empty?
```
rozza, line 125: As `clientSessionHelper.withClientSession` should always returns a session - either the users session or a generated one, the `switchIfEmpty` isn't needed - added #1455 to confirm.

rozza (changes requested):

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/OperationExecutorImpl.java
vbabanin: Thank you for the clarification and for addressing the TODOs. I approved the https://github.com/mongodb/mongo-java-driver/pull/1455.

driver-core/src/main/com/mongodb/internal/operation/SyncOperations.java
```diff
@@ -58,27 +60,84 @@
...
+    public TimeoutSettings createTimeoutSettings(final FindOneAndReplaceOptions options) {
+        return createTimeoutSettings(options.getMaxTime(MILLISECONDS));
+    }
+
+    @SuppressWarnings("deprecation") // MaxTime
+    public TimeoutSettings createTimeoutSettings(final FindOneAndUpdateOptions options) {
+        return timeoutSettings.withMaxTimeMS(options.getMaxTime(MILLISECONDS));
+    }
+
+    // TODO (CSOT) @SuppressWarnings("deprecation") // MaxTime
```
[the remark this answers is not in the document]
vbabanin: Thanks for pointing that out. I've committed the updates you suggested below.

rozza (commented):
LGTM

rozza (approved):

vbabanin (commented):
LGTM!

PR #1217, opened 2023-10-05
Java 5189

rozza:
JAVA-5189

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/Locks.java
+++ b/driver-core/src/main/com/mongodb/internal/Locks.java
@@ -52,6 +52,7 @@ public final class Locks {
         try {
             lock.lockInterruptibly();
         } catch (InterruptedException e) {
+            lock.unlock();
             throw interruptAndCreateMongoInterruptedException("Interrupted waiting for lock", e);
         }
     }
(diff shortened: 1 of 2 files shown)


rozza:
Patch: https://spruce.mongodb.com/version/651ebeadd1fe0742372e82b1/

I'm not 100% sure this is the correct approach

driver-core/src/main/com/mongodb/internal/Locks.java
```diff
@@ -52,6 +52,7 @@ public static void lockInterruptibly(final Lock lock) throws MongoInterruptedExc
         try {
             lock.lockInterruptibly();
         } catch (InterruptedException e) {
+            lock.unlock();
```
jyemin, line 55: If `InterruptedException` is thrown from `lockInterruptibly`, how is it not already unlocked (unless it's been locked by a different thread, in which case that thread should be responsible for unlocking)?

stIncMale: This `lock.unlock()` will cause `IllegalMonitorStateException` if `ReentrantLock`/`ReentrantReadWriteLock` is used. If `InterruptedException` is thrown from `Lock.lockInterruptibly`, then the lock hasn't been acquired by the current thread.

rozza:
Closing as it incidently worked rather than fixed the issue. #1220 implements a correct fix.

PR #1218, opened 2023-10-05
$lookup "from" must be null when using $documents pipeline

katcharov:
JAVA-5190


jyemin (approved):

jyemin:
Removed @vbabanin since I already review this.

PR #1221, opened 2023-10-11
Add server selection log messages

stIncMale:
The implemented spec changes: https://github.com/mongodb/specifications/commit/c773534925f83ecc5728971442baa63d69ecf0ce.

JAVA-4754

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -64,7 +64,7 @@ import static com.mongodb.internal.Locks.withInterruptibleLock;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
 import static com.mongodb.internal.connection.EventHelper.wouldDescriptionsGenerateEquivalentEvents;
 import static com.mongodb.internal.event.EventListenerHelper.singleClusterListener;
-import static com.mongodb.internal.logging.LogMessage.Component.SERVERSELECTION;
+import static com.mongodb.internal.logging.LogMessage.Component.SERVER_SELECTION;
 import static com.mongodb.internal.logging.LogMessage.Entry.Name.FAILURE;
 import static com.mongodb.internal.logging.LogMessage.Entry.Name.OPERATION;
 import static com.mongodb.internal.logging.LogMessage.Entry.Name.OPERATION_ID;
@@ -74,6 +74,7 @@ import static com.mongodb.internal.logging.LogMessage.Entry.Name.SERVER_HOST;
 import static com.mongodb.internal.logging.LogMessage.Entry.Name.SERVER_PORT;
 import static com.mongodb.internal.logging.LogMessage.Entry.Name.TOPOLOGY_DESCRIPTION;
 import static com.mongodb.internal.logging.LogMessage.Level.DEBUG;
+import static com.mongodb.internal.logging.LogMessage.Level.INFO;
 import static com.mongodb.internal.thread.InterruptionUtil.interruptAndCreateMongoInterruptedException;
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
@@ -120,7 +121,6 @@ abstract class BaseCluster implements Cluster {
         try {
             CountDownLatch currentPhase = phase.get();
             ClusterDescription curDescription = description;
-            // VAKOTODO think of simplifying the toString of driver selectors
             logServerSelectionStarted(operationContext, serverSelector, curDescription);
             ServerSelector compositeServerSelector = getCompositeServerSelector(serverSelector);
             ServerTuple serverTuple = selectServer(compositeServerSelector, curDescription);
@@ -502,13 +502,13 @@ abstract class BaseCluster implements Cluster {
             final ClusterDescription clusterDescription) {
         if (STRUCTURED_LOGGER.isRequired(DEBUG, clusterId)) {
             STRUCTURED_LOGGER.log(new LogMessage(
-                    SERVERSELECTION, DEBUG, "Server selection started", clusterId,
+                    SERVER_SELECTION, DEBUG, "Server selection started", clusterId,
                     asList(
-                            new Entry(OPERATION, "VAKOTODO"),
+                            new Entry(OPERATION, null),
                             new Entry(OPERATION_ID, operationContext.getId()),
                             new Entry(SELECTOR, serverSelector.toString()),
                             new Entry(TOPOLOGY_DESCRIPTION, clusterDescription.getShortDescription())),
-                    "Server selection started for operation {} with ID {}. Selector: {}, topology description: {}"));
+                    "Server selection started for operation[ {}] with ID {}. Selector: {}, topology description: {}"));
         }
     }
 
@@ -518,17 +518,17 @@ abstract class BaseCluster implements Cluster {
             final Long remainingTimeNanos,
             final ServerSelector serverSelector,
             final ClusterDescription clusterDescription) {
-        if (STRUCTURED_LOGGER.isRequired(DEBUG, clusterId)) {
+        if (STRUCTURED_LOGGER.isRequired(INFO, clusterId)) {
             STRUCTURED_LOGGER.log(new LogMessage(
-                    SERVERSELECTION, DEBUG, "Waiting for suitable server to become available", clusterId,
+                    SERVER_SELECTION, INFO, "Waiting for suitable server to become available", clusterId,
                     asList(
-                            new Entry(OPERATION, "VAKOTODO"),
+                            new Entry(OPERATION, null),
                             new Entry(OPERATION_ID, operationContext.getId()),
                             new Entry(REMAINING_TIME_MS, remainingTimeNanos == null ? null : NANOSECONDS.toMillis(remainingTimeNanos)),
                             new Entry(SELECTOR, serverSelector.toString()),
                             new Entry(TOPOLOGY_DESCRIPTION, clusterDescription.getShortDescription())),
-                    "Waiting for server to become available for operation {} with ID {}.[ Remaining time: {} ms.]" +
-                            " Selector: {}, topology description: {}."));
+                    "Waiting for server to become available for operation[ {}] with ID {}.[ Remaining time: {} ms.]"
+                            + " Selector: {}, topology description: {}."));
         }
     }
 
@@ -539,14 +539,14 @@ abstract class BaseCluster implements Cluster {
             final ClusterDescription clusterDescription) {
         if (STRUCTURED_LOGGER.isRequired(DEBUG, clusterId)) {
             STRUCTURED_LOGGER.log(new LogMessage(
-                    SERVERSELECTION, DEBUG, "Server selection failed", clusterId,
+                    SERVER_SELECTION, DEBUG, "Server selection failed", clusterId,
                     asList(
-                            new Entry(OPERATION, "VAKOTODO"),
+                            new Entry(OPERATION, null),
                             new Entry(OPERATION_ID, operationContext.getId()),
                             new Entry(FAILURE, failure.toString()),
                             new Entry(SELECTOR, serverSelector.toString()),
                             new Entry(TOPOLOGY_DESCRIPTION, clusterDescription.getShortDescription())),
-                    "Server selection failed for operation {} with ID {}. Failure: {}. Selector: {}, topology description: {}"));
+                    "Server selection failed for operation[ {}] with ID {}. Failure: {}. Selector: {}, topology description: {}"));
         }
     }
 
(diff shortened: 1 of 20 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -118,6 +119,7 @@ public final class Entities {
     private final Map<String, TestLoggingInterceptor> clientLoggingInterceptors = new HashMap<>();
     private final Map<String, TestConnectionPoolListener> clientConnectionPoolListeners = new HashMap<>();
     private final Map<String, TestServerListener> clientServerListeners = new HashMap<>();
+    private final Map<String, TestClusterListener> clientClusterListeners = new HashMap<>();
```
stIncMale, line 122: The new code using `clientClusterListeners` in this PR was written based on the existing code that uses `clientServerListeners`.

driver-core/src/test/unit/com/mongodb/internal/connection/TestClusterListener.java
```diff
@@ -20,17 +20,28 @@
...
 
 import static com.mongodb.assertions.Assertions.isTrue;
 import static com.mongodb.assertions.Assertions.notNull;
+import static com.mongodb.internal.Locks.withLock;
 
-class TestClusterListener implements ClusterListener {
-    private ClusterOpeningEvent clusterOpeningEvent;
-    private ClusterClosedEvent clusterClosingEvent;
-    private final List<ClusterDescriptionChangedEvent> clusterDescriptionChangedEvents = new ArrayList<>();
+public final class TestClusterListener implements ClusterListener {
```
stIncMale, line 37: The changes in `TestClusterListener` are needed for the spec tests added in this PR.

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -82,22 +85,37 @@ public void log(final LogMessage logMessage) {
...
-                    LogMessage.UnstructuredLogMessage unstructuredLogMessage = logMessage.toUnstructuredLogMessage();
-                    String message = unstructuredLogMessage.interpolate();
-                    Throwable exception = logMessage.getException();
-                    if (exception == null) {
-                        logger.debug(message);
-                    } else {
-                        logger.debug(message, exception);
-                    }
-                }
+                logUnstructured(logMessage, Logger::isDebugEnabled, Logger::debug, Logger::debug);
```
stIncMale, line 90: The inference magic is strong in this code, when it comes to the last two arguments.

driver-core/src/main/com/mongodb/internal/logging/LogMessage.java
```diff
@@ -41,11 +44,34 @@ public final class LogMessage {
...
+        }
+
+        public String getValue() {
+            return value;
+        }
+
+        public static Component of(final String value) {
+            Component result = INDEX.get(value);
+            return assertNotNull(result);
+        }
```
stIncMale, line 70: The `serverSelection` component is a bit more complex than the previous one-word ones, so more code is needed to match components and decode a string to `Component`.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -105,34 +121,42 @@ public ServerTuple selectServer(final ServerSelector serverSelector, final Opera
...
                 if (serverTuple != null) {
+                    logServerSelectionSucceeded(operationContext, serverTuple.getServerDescription().getAddress(), serverSelector, curDescription);
                     return serverTuple;
                 }
 
-                if (curTimeNanos - startTimeNanos > maxWaitTimeNanos) {
-                    throw createTimeoutException(serverSelector, curDescription);
+                Long remainingTimeNanos = maxWaitTimeNanos == null ? null : maxWaitTimeNanos - (curTimeNanos - startTimeNanos);
+
+                if (remainingTimeNanos != null && remainingTimeNanos <= 0) {
```
stIncMale, line 146: Note how now `remainingTimeNanos <= 0` is used, while previously `curTimeNanos - startTimeNanos > maxWaitTimeNanos` was used, which was equivalent to `remainingTimeNanos < 0`. The reason we must use `<=` instead of `<` is that when `maxWaitTimeNanos` is 0, we must not wait (as per the `getServerSelectionTimeout` docs), and `<= 0` makes sure we don't wait, while `<` is not guaranteed to give us that behavior.

driver-core/src/main/com/mongodb/internal/logging/LogMessage.java
```diff
@@ -73,6 +99,10 @@ public enum Name {
             COMMAND_NAME("commandName"),
             REQUEST_ID("requestId"),
             OPERATION_ID("operationId"),
+            /**
+             * Not supported.
+             */
+            OPERATION("operation"),
```
stIncMale, line 105: The final decision on whether to support `operation` and to what extent will be done in https://jira.mongodb.org/browse/JAVA-5197.

vbabanin (changes requested):
Looks good! Just a couple of questions and suggestions. Additionally, could you please confirm if manual testing was performed for unstructured logging?

driver-core/src/main/com/mongodb/internal/logging/LogMessage.java
vbabanin: It seems the only usage of this INDEX map is in tests. We could consider including it as part of test fixtures, perhaps within the Entities.class to keep it separate for code clarity and the main logic a bit less cluttered.

vbabanin (changes requested):

driver-core/src/main/com/mongodb/internal/logging/StructuredLogger.java
```diff
@@ -82,22 +85,37 @@ public void log(final LogMessage logMessage) {
...
+            if (exception == null) {
+                doLog.accept(logger, message);
+            } else {
+                doLogWithException.accept(logger, message, exception);
+            }
+        }
+    }
+
+    @FunctionalInterface
+    private interface TriConsumer<A, B, C> {
```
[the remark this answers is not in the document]
stIncMale: Good idea, done.

driver-core/src/main/com/mongodb/internal/logging/LogMessage.java
stIncMale: I marked the `getValue`, `of` methods with `@VisibleForTesting(otherwise = PRIVATE)` (I should have done this from the get go). If `Component` were API, it would have indeed been reasonable not to add these methods to `Component`. However, given that the code is internal, it seems to me that having this methods on `Component` itself is better, because conceptually they seem to belong on it.

vbabanin (approved):

stIncMale:
> Additionally, could you please confirm if manual testing was performed for unstructured logging?

Yes, I've just done that.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -105,34 +121,42 @@ public ServerTuple selectServer(final ServerSelector serverSelector, final Opera
...
             ServerSelector compositeServerSelector = getCompositeServerSelector(serverSelector);
             ServerTuple serverTuple = selectServer(compositeServerSelector, curDescription);
 
-            boolean selectionFailureLogged = false;
+            boolean selectionWaitingLogged = false;
 
             long startTimeNanos = System.nanoTime();
             long curTimeNanos = startTimeNanos;
-            long maxWaitTimeNanos = getMaxWaitTimeNanos();
+            Long maxWaitTimeNanos = getMaxWaitTimeNanos();
```
katcharov, line 132: Does this method need to be refactored in order to accomplish the logging spec changes? Note that much of this has been refactored already in the CSOT branch. It seems that we only need to either add methods that log messages, or remove existing logs.

katcharov (changes requested):
Looking good, just a question.

We might run into rebasing issues with CSOT that would require an understanding of this task, even without refactoring, but I don't see a way around that.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
stIncMale: - `remainingTimeNanos` has to be computed and logged, unless it's infinite. This required changing the code.
- Creating exceptions and logging them is related, which means that the line `throwIfIncompatible(curDescription);` needed to be changed.

stIncMale: > Note that much of this has been refactored already in the CSOT branch.

I won't be surprised if merging this code from `master` and CSOT will effectively require reintroducing the spec changes in the `selectServer` methods (not from scratch, though, as the methods that create exceptions and the methods that log them don't need changes)

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
jyemin: The Javadoc says:

```
A value of 0 means that it will timeout immediately if no server is available.  
A negative value means to wait indefinitely.
```

The spec only says:

```
A serverSelectionTimeoutMS of zero MAY have special meaning in some drivers; 
zero's meaning is not defined in this spec, but all drivers SHOULD document the meaning of zero.
```

So this looks good to me.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -376,24 +383,31 @@ private MongoIncompatibleDriverException createIncompatibleException(final Clust
...
-    private MongoTimeoutException createTimeoutException(final ServerSelector serverSelector, final ClusterDescription curDescription) {
-        return new MongoTimeoutException(format("Timed out after %d ms while waiting for a server that matches %s. "
-                                                + "Client view of cluster state is %s",
-                                                settings.getServerSelectionTimeout(MILLISECONDS), serverSelector,
-                                                curDescription.getShortDescription()));
+    private MongoException createAndLogTimeoutException(
+            final OperationContext operationContext,
+            final ServerSelector serverSelector,
+            final ClusterDescription clusterDescription) {
+        MongoTimeoutException exception = new MongoTimeoutException("Timed out while waiting for a suitable server");
```
jyemin, line 390: The existing message that we use for this exception has proved incredibly useful for support over the years.  I don't think we should change it.  Is there something in the spec that requires this change?

I'd be fine if the exception message more closely matches what we now log for failures.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -477,4 +495,78 @@ public void run() {
...
+            final MongoException failure,
+            final ServerSelector serverSelector,
+            final ClusterDescription clusterDescription) {
+        if (STRUCTURED_LOGGER.isRequired(DEBUG, clusterId)) {
+            STRUCTURED_LOGGER.log(new LogMessage(
+                    SERVER_SELECTION, DEBUG, "Server selection failed", clusterId,
+                    asList(
+                            new Entry(OPERATION, null),
+                            new Entry(OPERATION_ID, operationContext.getId()),
+                            new Entry(FAILURE, failure.toString()),
```
jyemin, line 546: I know it's in the specification, but it's not clear what useful information can be included in the failure string.   As far as I can see it's always the same message.  @BorisDog can you clarify the intention here?

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
jyemin: I think I see why you did this.  Is it because you log the message from this exception and don't want to repeat anything?  If so, I think we should just log the message here as a constant and leave the message as is.

jyemin (changes requested):

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
BorisDog: [Spec](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#server-selection-errors) defines different errors in different cases, mainly:
1. Timeout
2. Incompatibility (few flavors)

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
stIncMale: The spec [requires the following regarding the `failure` key](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#server-selection-failed-message):

> Drivers MUST take care to not include any information in this field that is already included in the log message; e.g. the topology description should not be duplicated within this field.

Given that both the selector and the topology description must be included in the log message according to the logging spec, the only piece of data we are allowed to still have in the `MongoTimeoutException` message of those we previously had, is the amount of time after which we timed out. This was already discussed, and everyone was happy that we don't need to track it in CSOT because the spec does not require us to log it.

> so, I think we should just log the message here as a constant and leave the message as is.

By "as is", do you mean to keep the selector and the topology description in the `MongoTimeoutException` message, or do you mean to also preserve the amount of time after which we timed out? The latter invalidates what was discussed previously in the CSOT context.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -538,12 +540,18 @@ private void logServerSelectionFailed(
             final ServerSelector serverSelector,
             final ClusterDescription clusterDescription) {
         if (STRUCTURED_LOGGER.isRequired(DEBUG, clusterId)) {
+            String failureDescription = failure instanceof MongoTimeoutException
+                    // This hardcoded message guarantees that the `FAILURE` entry for `MongoTimeoutException` does not include
+                    // any information that is specified via other entries, e.g., `SELECTOR` and `TOPOLOGY_DESCRIPTION`.
+                    // The logging spec requires us to avoid such duplication of information.
+                    ? MongoTimeoutException.class.getName() + ": Timed out while waiting for a suitable server"
```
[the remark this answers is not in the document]
stIncMale: The other branch (`failure.toString()`) includes it, so I mimicked it for consistency.

PR #1222, opened 2023-10-14
Fix automatic module name was broken

Icosider:
image

[Package name convention](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html)


Icosider:
This bug work to all Kotlin modules of MongoDb

jyemin:
This was fixed in scope of JAVA-5202.  Thanks for bringing it to our attention.

PR #1223, opened 2023-10-16
[JAVA-5203] Support stored nulls for non-primitive fields in Java records

bulbfreeman:
## Why
Same as the issue raised in https://jira.mongodb.org/browse/JAVA-5134, this change implements a mechanism to gracefully handle stored nulls for fields that are annotated with JSR305 `@Nullable` annotations.

## Changes
- Introduce a boolean field `isNullable` to `RecordCodec.ComponentModel`, which is set to true when given field is not primitive.
- When decoding stored null for a field that is considered nullable in above step, read null in stead of throwing an `BsonInvalidOperationException`.
- For any other scenarios, previous codec behaviour is preserved.

## Issue ticket
https://jira.mongodb.org/browse/JAVA-5203

### DIFF
--- a/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
+++ b/bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
@@ -16,6 +16,7 @@
 
 package org.bson.codecs.record;
 
+import org.bson.BsonInvalidOperationException;
 import org.bson.BsonReader;
 import org.bson.BsonType;
 import org.bson.BsonWriter;
@@ -277,7 +278,10 @@ final class RecordCodec<T extends Record> implements Codec<T> {
                 if (LOGGER.isTraceEnabled()) {
                     LOGGER.trace(format("Found property not present in the ClassModel: %s", fieldName));
                 }
-            } else if (reader.getCurrentBsonType() == BsonType.NULL && componentModel.isNullable) {
+            } else if (reader.getCurrentBsonType() == BsonType.NULL) {
+                if (!componentModel.isNullable) {
+                    throw new BsonInvalidOperationException(format("Null value on primitive field: %s", componentModel.fieldName));
+                }
                 reader.readNull();
             } else {
                 constructorArguments[componentModel.index] = decoderContext.decodeWithChildContext(componentModel.codec, reader);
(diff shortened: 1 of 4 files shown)


bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -156,6 +158,14 @@ private static String computeFieldName(final RecordComponent component) {
             return component.getName();
         }
 
+        private static boolean checkFieldIsNullable(final RecordComponent component) {
```
jyemin, line 161: This should be replaced with a more straightforward check of 

```
!component.getType().isPrimitive()
```

Not everyone uses the `javax.annotation.Nullable` type to declare nullable components.  If a component value shouldn't be null, the constructor can throw.

jyemin (changes requested):
Thanks for the Jira report and the pull request.  I have a change request for you if you're inclined to keep working on this.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
bulbfreeman: Make sense - updated here, test record name and PR decription.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
```diff
@@ -275,6 +277,8 @@ public T decode(final BsonReader reader, final DecoderContext decoderContext) {
                 if (LOGGER.isTraceEnabled()) {
                     LOGGER.trace(format("Found property not present in the ClassModel: %s", fieldName));
                 }
+            } else if (reader.getCurrentBsonType() == BsonType.NULL && componentModel.isNullable) {
```
jyemin, line 280: One other change request, if you don't mind: I don't think this should fall through to the else branch if the BSON type is `NULL` but the component is not nullable.  Rather, it should throw an exception directly.  `BsonInvalidOperationException` is probably best.  Add a test for it too.

jyemin (changes requested):
One more change, please.

bson-record-codec/src/main/org/bson/codecs/record/RecordCodec.java
bulbfreeman: Make sense - modified to let NULL on primitive field throw early and added one test case for it.

jyemin (approved):

jyemin:
Thanks for the contribution, @bulbfreeman

bulbfreeman:
OOI is there an expected date for the next version release that contains this change? @jyemin

bulbfreeman:
Awesome, thanks!

PR #1224, opened 2023-10-17
Remove deprecated ServerAddress methods

jyemin:
* ServerAddress#getSocketAddress
* ServerAddress#getSocketAddresses
* UnixServerAddress#getSocketAddress
* UnixServerAddress#getUnixSocketAddress

JAVA-4937

Design Notes:

- Delete `ServerAddressWithResolver`  and instead push the `InetAddressResolver` down to the `Stream` implementations 
- Add internal `DefaultInetAddressResolver` to replicate the removed `ServerAddress` method and avoid having to make the resolver nullable everywhere


driver-core/src/main/com/mongodb/internal/connection/ServerAddressHelper.java
```diff
@@ -37,6 +45,21 @@ public static ServerAddress createServerAddress(final String host, final int por
...
     }
 
+    public static InetAddressResolver getInetAddressResolver(final MongoClientSettings settings) {
+        InetAddressResolver inetAddressResolver = settings.getInetAddressResolver();
+        return inetAddressResolver == null ? new DefaultInetAddressResolver() : inetAddressResolver;
+    }
+
+    public static List<InetSocketAddress> getSocketAddresses(final ServerAddress serverAddress, final InetAddressResolver resolver) {
+        try {
+            return resolver.lookupByName(serverAddress.getHost())
```
jyemin, line 55: This was copied from `ServerAddressWithResolver` implementation

driver-core/src/test/functional/com/mongodb/connection/netty/NettyStreamSpecification.groovy
```diff
@@ -68,24 +67,6 @@ class NettyStreamSpecification extends Specification {
         thrown(MongoSocketOpenException)
     }
 
-    @IgnoreIf({ getSslSettings().isEnabled() })
-    def 'should fail AsyncCompletionHandler if name resolution fails'() {
```
stIncMale, line 72: Why was this test removed? It seems that the behavior tested here is still there, and technically we should still be able to test it. This question is also applicable to the changes in `AsyncSocketChannelStreamSpecification`.

stIncMale (changes requested):

driver-core/src/test/functional/com/mongodb/connection/netty/NettyStreamSpecification.groovy
jyemin: There was no good reason.  And not only that, the `@Slow` tests were all broken so after adding these back I also fixed those.  Sorry about that.

jyemin (commented):
Patch build: https://spruce.mongodb.com/version/6532d61757e85a1ab0fcf729/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

stIncMale (approved):
Approving, but the patch you linked has failed.

jyemin:
https://spruce.mongodb.com/version/6532de605623439ac723f8e7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC is good

PR #1228, opened 2023-10-23
Obtain waitQueue and connection timeouts from OperationContext

katcharov:
JAVA-5212

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/TimeoutContext.java
+++ b/driver-core/src/main/com/mongodb/internal/TimeoutContext.java
@@ -158,4 +158,13 @@ public class TimeoutContext {
         long ms = getTimeoutSettings().getServerSelectionTimeoutMS();
         return StartTime.now().timeoutAfterOrInfiniteIfNegative(ms, MILLISECONDS);
     }
+
+    public Timeout startWaitQueueTimeout(final StartTime checkoutStart) {
+        final long ms = getTimeoutSettings().getMaxWaitTimeMS();
+        return checkoutStart.timeoutAfterOrInfiniteIfNegative(ms, MILLISECONDS);
+    }
+
+    public int getConnectTimeoutMs() {
+        return (int) getTimeoutSettings().getConnectTimeoutMS();
+    }
 }
--- a/driver-core/src/main/com/mongodb/internal/TimeoutSettings.java
+++ b/driver-core/src/main/com/mongodb/internal/TimeoutSettings.java
@@ -37,35 +37,38 @@ public class TimeoutSettings {
     @Nullable
     private final Long defaultTimeoutMS;
 
-    private final long maxAwaitTimeMS;
-
-    // Deprecated timeouts
-    private final long readTimeoutMS;
-
-    private final long maxTimeMS;
-
-    private final long maxCommitTimeMS;
+    // Deprecated configuration timeout options
+    private final long readTimeoutMS; // aka socketTimeoutMS
+    private final long maxWaitTimeMS; // aka waitQueueTimeoutMS
     @Nullable
     private final Long wTimeoutMS;
+
+    // Deprecated options for CRUD methods
+    private final long maxTimeMS;
+    private final long maxAwaitTimeMS;
+    private final long maxCommitTimeMS;
 
     public static final TimeoutSettings DEFAULT = create(MongoClientSettings.builder().build());
 
     public static TimeoutSettings create(final MongoClientSettings settings) {
-        return new TimeoutSettings(settings.getClusterSettings().getServerSelectionTimeout(TimeUnit.MILLISECONDS),
+        return new TimeoutSettings(
+                settings.getClusterSettings().getServerSelectionTimeout(TimeUnit.MILLISECONDS),
                 settings.getSocketSettings().getConnectTimeout(TimeUnit.MILLISECONDS),
                 settings.getSocketSettings().getReadTimeout(TimeUnit.MILLISECONDS),
-                settings.getTimeout(TimeUnit.MILLISECONDS));
+                settings.getTimeout(TimeUnit.MILLISECONDS),
+                settings.getConnectionPoolSettings().getMaxWaitTime(TimeUnit.MILLISECONDS));
     }
 
     public TimeoutSettings(
-            final long serverSelectionTimeoutMS, final long connectTimeoutMS, final long readTimeoutMS, @Nullable final Long timeoutMS) {
-        this(timeoutMS, null, serverSelectionTimeoutMS, connectTimeoutMS, readTimeoutMS, 0, 0, 0, null);
+            final long serverSelectionTimeoutMS, final long connectTimeoutMS, final long readTimeoutMS,
+            @Nullable final Long timeoutMS, final long maxWaitTimeMS) {
+        this(timeoutMS, null, serverSelectionTimeoutMS, connectTimeoutMS, readTimeoutMS, 0, 0, 0, null, maxWaitTimeMS);
     }
 
     TimeoutSettings(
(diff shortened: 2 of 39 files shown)


driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -158,4 +158,13 @@ public Timeout startServerSelectionTimeout() {
...
         return StartTime.now().timeoutAfterOrInfiniteIfNegative(ms, MILLISECONDS);
     }
+
+    public Timeout startWaitQueueTimeout(final StartTime checkoutStart) {
+        final long ms = getTimeoutSettings().getMaxWaitTimeMS();
+        return checkoutStart.timeoutAfterOrInfiniteIfNegative(ms, MILLISECONDS);
+    }
+
+    public int getConnectTimeoutMs() {
+        return (int) getTimeoutSettings().getConnectTimeoutMS();
```
katcharov, line 168: This returns an int. I think it would be more consistent to return a Timeout, but it was used in a way where 0 indicated infinite timeout (rather than the usual negative is infinite, see above). I could not find what the user should use to indicate infinite connect timeout specified in our docs.

driver-core/src/main/com/mongodb/internal/TimeoutSettings.java
```diff
@@ -37,35 +37,38 @@ public class TimeoutSettings {
...
+                settings.getTimeout(TimeUnit.MILLISECONDS),
+                settings.getConnectionPoolSettings().getMaxWaitTime(TimeUnit.MILLISECONDS));
     }
 
     public TimeoutSettings(
-            final long serverSelectionTimeoutMS, final long connectTimeoutMS, final long readTimeoutMS, @Nullable final Long timeoutMS) {
-        this(timeoutMS, null, serverSelectionTimeoutMS, connectTimeoutMS, readTimeoutMS, 0, 0, 0, null);
+            final long serverSelectionTimeoutMS, final long connectTimeoutMS, final long readTimeoutMS,
+            @Nullable final Long timeoutMS, final long maxWaitTimeMS) {
+        this(timeoutMS, null, serverSelectionTimeoutMS, connectTimeoutMS, readTimeoutMS, 0, 0, 0, null, maxWaitTimeMS);
```
katcharov, line 65: I added maxWaitTimeMS, in a way that seemed consistent with readTimeoutMS, but I was also not sure whether we intended to pass these in, since both were deprecated. Should these really be passed in, or should they be set using one of the "with" methods below?

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
```diff
@@ -399,7 +401,12 @@ public void checkoutHandOverMechanism() throws InterruptedException, TimeoutExce
         provider.ready();
         List<InternalConnection> connections = new ArrayList<>();
         for (int i = 0; i < openConnectionsCount; i++) {
-            connections.add(provider.get(OPERATION_CONTEXT, 0, NANOSECONDS));
+            OperationContext operationContextWithZeroMaxWait = new OperationContext(
+                    OPERATION_CONTEXT.getRequestContext(),
+                    OPERATION_CONTEXT.getSessionContext(),
+                    new TimeoutContext(new TimeoutSettings(30_000, 10_000, 0, null, 0)),
+                    OPERATION_CONTEXT.getServerApi());
+            connections.add(provider.get(operationContextWithZeroMaxWait));
```
katcharov, line 409: Note that this constructs a new operation context, as before with a 0 wait time (which I believe is intentional in this test), rather than just using OPERATION_CONTEXT.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -41,6 +44,24 @@ public OperationContext(final RequestContext requestContext, final SessionContex
         this(NEXT_ID.incrementAndGet(), requestContext, sessionContext, timeoutContext, serverApi);
     }
 
+
+    public static OperationContext todoOperationContext() {
+        // TODO (CSOT) should be removed; used at locations that require an OC, but which do not yet have one available
```
katcharov, line 49: These TODOs are intentional, and would get merged into the CSOT branch (which we would check for TODOs before merging, as per a comment there).

I left this open, because I think this would be more appropriate to supply during the crypt work.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -41,6 +44,24 @@ public OperationContext(final RequestContext requestContext, final SessionContex
...
+        return nonUserOperationContext(null);
+    }
+
+    public static OperationContext nonUserOperationContext(final MongoClientSettings settings) {
+        // TODO (CSOT) below is placeholder, validate correctness (serverApi/timeoutSettings might
+        // TODO (CSOT) need to be passed in instead)
+        TimeoutSettings timeoutSettings = TimeoutSettings.create(settings);
```
katcharov, line 56: This PR does not implement provision of non-user operation contexts (for example, monitoring, etc.), and usages pass in null. I believe how this should be handled would be easier to determine once we have labelled all such locations (by using the this method). If others feel that this is best done in this PR, I do not feel strongly about this, just erring on the side of less work per PR.

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
rozza: This is because with Java sockets a timeout cannot be negative and 0 means infinite.  See: [socket.connect](https://docs.oracle.com/javase/8/docs/api/java/net/Socket.html#connect-java.net.SocketAddress-int-)

driver-core/src/main/com/mongodb/internal/TimeoutSettings.java
```diff
@@ -37,35 +37,38 @@ public class TimeoutSettings {
     @Nullable
     private final Long defaultTimeoutMS;
 
-    private final long maxAwaitTimeMS;
-
-    // Deprecated timeouts
-    private final long readTimeoutMS;
+    // Deprecated configuration timeout options
+    private final long readTimeoutMS; // aka socketTimeoutMS
+    private final long maxWaitTimeMS; // aka waitQueueTimeoutMS
```
rozza, line 42: 👍

driver-core/src/main/com/mongodb/internal/TimeoutSettings.java
rozza: Its needed in the constructor, so it can be copied via the with methods.
We should eventually make this constructor private or annotate visibility test only.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
rozza: ACK

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
rozza: Ack - good plan

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
rozza: Ack

rozza (approved):
LGTM

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
katcharov: Right, though I think that we should probably say the same in our own docs. (I didn't confirm that all paths used an API where 0 meant infinite, but I think this must be the case.)

PR #1230, opened 2023-10-24
Ensures the QueryBatchCursor cleans up resources

rozza:
The QueryBatchCursor now checks the cursor is operable on result of getMore.

Updated the Async test to ensure connectionB is used.

JAVA-5214

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
@@ -292,7 +292,7 @@ class QueryBatchCursor<T> implements AggregateResponseBatchCursor<T> {
                 throw translateCommandException(e, serverCursor);
             }
             resourceManager.setServerCursor(nextServerCursor);
-            if (limitReached()) {
+            if (limitReached() || !resourceManager.operable()) {
                 resourceManager.releaseServerAndClientResources(connection);
             }
         });
--- a/driver-core/src/test/unit/com/mongodb/internal/operation/QueryBatchCursorSpecification.groovy
+++ b/driver-core/src/test/unit/com/mongodb/internal/operation/QueryBatchCursorSpecification.groovy
@@ -163,9 +163,7 @@ class QueryBatchCursorSpecification extends Specification {
         }
         List<Document> firstBatch = [new Document()]
         QueryResult<Document> initialResult = new QueryResult<>(NAMESPACE, firstBatch, 1, SERVER_ADDRESS)
-        Object getMoreResponse = useCommand
-                ? emptyGetMoreCommandResponse(NAMESPACE, getMoreResponseHasCursor ? 42 : 0)
-                : emptyGetMoreQueryResponse(NAMESPACE, SERVER_ADDRESS, getMoreResponseHasCursor ? 42 : 0)
+        Object getMoreResponse = emptyGetMoreCommandResponse(NAMESPACE, getMoreResponseHasCursor ? 42 : 0)
 
         when:
         QueryBatchCursor<Document> cursor = new QueryBatchCursor<>(initialResult, 0, 0, 0, new DocumentCodec(), null, connSource, conn)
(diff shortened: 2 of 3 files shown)


jyemin:
Please update the commit message to reference JAVA-5214.

rozza:
Fixed commit message JAVA-5124 -> JAVA-5214

driver-core/src/main/com/mongodb/internal/operation/QueryBatchCursor.java
```diff
@@ -292,7 +292,7 @@ private void getMore() {
                 throw translateCommandException(e, serverCursor);
             }
             resourceManager.setServerCursor(nextServerCursor);
-            if (limitReached()) {
+            if (limitReached() || !resourceManager.operable()) {
```
stIncMale, line 295: I don't think there is currently a bug as specified in JAVA-5214, nor that this check is needed. Following are my observations, though if I am missing something, please correct me.

Currently, `ResourceManager.execute` guarantees that both server and client resources are released (via `ResourceManager.doClose`) if `QueryBatchCursor.close` is called concurrently with `QueryBatchCursor.hasNext`/`next`/`tryNext`. And if `QueryBatchCursor.close` is ordered either before or after (in the synchronization order, which in this particular case also implies the happens-before order) `QueryBatchCursor.hasNext`/`next`/`tryNext`, then `QueryBatchCursor.close` itself releases those resources (again, via `ResourceManager.doClose`).

More specifically, here the `!resourceManager.operable()` check was added to the `getMore` method. `getMore` is executed by `ResourceManager.execute`, which calls `endOperation` in a `finally` block, which in turn checks the state and calls `doClose` if needed. See also VAKOTODO on why the modified `should close cursor after getMore finishes if cursor was closed while getMore was in progress and getMore returns a response` fails.

P.S. You may see that `doHasNext` contains the `!resourceManager.operable()` check, but that is only because it calls `getMore` in a loop, and the loop must not continue if `QueryBatchCursor.close` as called. Note how `doHasNext` does nor releases anything itself, because it by design relies on `ResourceManager.execute` doing that.

P.P.S. The only time when `QueryBatchCursor` calls `ResourceManager.releaseServerAndClientResources` itself, instead of relying on `ResourceManager.execute`, is when `limitReached()`, which is again by design: `ResourceManager.execute` does not understand the semantics of the "operation" it executes, and can't know that despite `QueryBatchCursor.close` not having been called, the cursor may release resources because no more data is going to be fetched. This is just an optimization, and it had already been there before the introduction of `ResourceManager`.

driver-core/src/test/unit/com/mongodb/internal/operation/QueryBatchCursorSpecification.groovy
```diff
@@ -179,24 +177,14 @@ class QueryBatchCursorSpecification extends Specification {
...
-            }
+        // in LB mode the same connection is used to execute both `getMore` and `killCursors`
+        numberOfInvocations * conn.command(*_) >> {
+            // `getMore` command
+            cursor.close()
+            getMoreResponse
+        } >> {
+            // `killCursors` command
+            null
         }
```
stIncMale, line 188: The reason this test fails with the change made in the PR, is that it assumes that the same `conn` is used for issuing the `killCursors` regardless of whether LB mode is used (and the connection is pinned to the cursor) or not. I don't think this assumption is valid for a non-LB mode.

If the mocking in test done without the aforementioned assumption for non-LB mode, the test passes when run against `QueryBatchCursor` without requiring any more changes. See https://github.com/stIncMale/mongo-java-driver/commit/a6cad08e4fa147ca179e8dffc33e5871daf4b34d for the proposed change.

stIncMale (changes requested):

rozza:
Thanks @stIncMale - so no bug, just unclear tests.

I've simplified / clarified the initial connection and tests in the CommandBatchCursor refactor.  Closing this PR.

PR #1231, opened 2023-10-24
Remove dead code in ReplyHeader

jyemin:
The dead code is a remnant of when the driver supported the full range of OP_REPLY wire protocol message usage.  Since the driver now only runs against MongoDB releases that support OP_MSG, OP_REPLY is only used for the response to the `hello` command in the handshake, and therefore most of the OP_REPLY-handling code is no longer on any execution paths.

JAVA-5204

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
@@ -546,7 +546,7 @@ public class InternalStreamConnection implements InternalConnection {
     }
 
     private <T> T getCommandResult(final Decoder<T> decoder, final ResponseBuffers responseBuffers, final int messageId) {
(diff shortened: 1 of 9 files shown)


stIncMale (approved):

vbabanin (approved):
LGTM! I've noticed that `OP_REPLY` is unsupported as of MongoDB 5.1. Does this mean that driver 5.0 will not be compatible with MongoDB prior to 5.1?

jyemin:
No, nothing like that.  The driver (and this code) fully supports OP_MSG, which is supported by all server versions >= 3.6.  OP_REPLY is only used in very limited circumstances by the driver: when sending the first handshake message on a connection, when we don't know what server version we're connected to.  So the driver sends an "ismaster" command using OP_QUERY, and the server responds with an OP_REPLY message.  The driver will then switch to OP_MSG for all subsequent messages.

PR #1232, opened 2023-10-24
Add explicit automatic module name configuration.

vbabanin:
<h2> Description :</h2>

Some modules rely on the `publish.gradle` file to generate the Automatic Module Name in the manifest, based on the `groupName` and `archivesBaseName` of the project. However, the `archivesName` contains hyphens that are not valid according to the [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se13/html/jls-7.html#jls-7.7).

As a result, all Automatic Module Names are now explicitly specified in each build.gradle file:

`org.mongodb.mongodb-driver-kotlin-coroutine` is renamed to `org.mongodb.driver.kotlin.coroutine`
`org.mongodb.mongodb-driver-kotlin-sync` is renamed to `org.mongodb.driver.kotlin.sync`
`org.mongodb.bson-kotlin` is renamed to `org.mongodb.bson.kotlin`
`org.mongodb.bson-kotlinx` is renamed to `org.mongodb.bson.kotlinx`
`org.mongodb.mongo-scala-driver` is renamed to `org.mongodb.driver.scala`

[JAVA-5213](https://jira.mongodb.org/browse/JAVA-5213)
[JAVA-5202](https://jira.mongodb.org/browse/JAVA-5202)

### DIFF
--- a/bson-scala/build.gradle
+++ b/bson-scala/build.gradle
@@ -58,7 +58,10 @@ test {
     maxParallelForks = 1
 }
 
-jar.manifest.attributes['Import-Package'] =  [
-        '!scala.*',
-        '*'
-].join(',')
+afterEvaluate {
+    jar.manifest.attributes['Automatic-Module-Name'] = 'org.mongodb.scala.bson'
+    jar.manifest.attributes['Import-Package'] = [
+            '!scala.*',
+            '*'
+    ].join(',')
+}
--- a/bson-scala/build.gradle
+++ b/bson-scala/build.gradle
@@ -59,7 +59,7 @@ test {
 }
 
 afterEvaluate {
-    jar.manifest.attributes['Automatic-Module-Name'] = 'org.mongodb.scala.bson'
+    jar.manifest.attributes['Automatic-Module-Name'] = 'org.mongodb.bson.scala'
     jar.manifest.attributes['Import-Package'] = [
             '!scala.*',
             '*'
--- a/driver-scala/build.gradle
+++ b/driver-scala/build.gradle
@@ -117,7 +117,7 @@ ext {
 }
 
 afterEvaluate {
-    jar.manifest.attributes['Automatic-Module-Name'] = 'org.mongodb.scala.driver'
+    jar.manifest.attributes['Automatic-Module-Name'] = 'org.mongodb.driver.scala'
     jar.manifest.attributes['Import-Package'] = [
             '!scala.*',
             '*'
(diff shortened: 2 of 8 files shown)


driver-scala/build.gradle
```diff
@@ -117,7 +117,7 @@ ext {
 }
 
 afterEvaluate {
-    jar.manifest.attributes['Automatic-Module-Name'] = 'org.mongodb.scala.mongo-scala-driver'
+    jar.manifest.attributes['Automatic-Module-Name'] = 'org.mongodb.scala.driver'
```
jyemin, line 120: This one seems inconsistent with the others, where `driver` is before the language.  Should we change to `org.mongodb.driver.scala`?

bson-scala/build.gradle
```diff
@@ -58,7 +58,10 @@ test {
     maxParallelForks = 1
 }
 
-jar.manifest.attributes['Import-Package'] =  [
-        '!scala.*',
-        '*'
-].join(',')
+afterEvaluate {
+    jar.manifest.attributes['Automatic-Module-Name'] = 'org.mongodb.scala.bson'
```
jyemin, line 62: This one seems inconsistent with the others, where `bson` is before the language. Should we change to org.mongodb.bson.scala?

jyemin (changes requested):

jyemin:
Should we remove `Automatic-Module-Name` configuration from publish.gradle entirely since it's rarely correct (bson is the only module without a hyphen, and `driver-legacy` should not have a module name)?

jyemin:
Static checks are failing.

driver-scala/build.gradle
vbabanin: I aimed for consistency with the archive name, following Java's approach when the Automatic Module name is missing. However, for overall consistency, it is better to stick with having `driver` before the language. I will rename it. Thanks!

vbabanin:
> Should we remove `Automatic-Module-Name` configuration from publish.gradle entirely since it's rarely correct (bson is the only module without a hyphen, and `driver-legacy` should not have a module name)?

Agree, I've removed the `Automatic-Module-Name` configuration, as it indeed seems unnecessary.  Thank you! 
With explicit naming in place, it might be worth considering adding a validation check in `publish.gradle` to ensure `Automatic-Module-Name`'s are correctly specified and not overlooked once we need to add a new module."

rozza (approved):
LGTM

PR #1239, opened 2023-11-01
Fix server selection test performance

jyemin:
Apply test server selection timeouts to OperationContext instead of ClusterSettings

### DIFF
--- a/driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
+++ b/driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
@@ -61,6 +61,8 @@ import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Stream;
 
 import static com.mongodb.ClusterFixture.OPERATION_CONTEXT;
+import static com.mongodb.ClusterFixture.TIMEOUT_SETTINGS;
+import static com.mongodb.ClusterFixture.createOperationContext;
 import static java.lang.Long.MAX_VALUE;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.MINUTES;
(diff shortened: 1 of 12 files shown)


rozza (approved):
LGTM - thanks :)

jyemin:
Fixed the rest of the tests.  Should be green now

rozza:
Once tests are green LGTM.

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
```diff
@@ -316,7 +317,6 @@ public void concurrentUsage(final int minSize, final int maxSize, final boolean
                 ConnectionPoolSettings.builder()
                     .minSize(minSize)
                     .maxSize(maxSize)
-                    .maxWaitTime(TEST_WAIT_TIMEOUT_MILLIS, MILLISECONDS)
```
vbabanin, line 319: During my review, I noticed that in a previous PR, `maxTimeAwait()` was removed from the `concurrentUsage()` test. I understand the rationale, assuming that we no longer use the maxTimeAwait parameter from ConnectionPool in `DefaultConnectionPool` in favor of `TimeoutContext`.

However, upon further examination, I found that the sync method (as seen [here](https://github.com/mongodb/mongo-java-driver/blob/272a7c41e75c394d753fa8901998e5b0999a7646/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java#L190)) uses `maxAwaitTime` from `TimeoutContext`, while `getAsync` from `ConnectionPoolSettings`.

As a result, the `maxAwaitTime` limitation in the test applies only for the sync version of execution, but we test both sync and async. Should we consider leaving the `maxAwaitTime` parameter in the test until `TimeoutContext` is used in `getAsync`? Otherwise, the test may not be complete.

jyemin: Yeah, I missed this one because it's a Slow test that didn't run in the patch build.  Fixed in the followup PR.

PR #1244, opened 2023-11-03
Make NettyStreamFactoryFactory implement AutoCloseable

jyemin:
Since StreamFactoryFactory is now internal, just made that interface extend AutoCloseable in order to simplify MongoClients code

JAVA-5158

Note: If TransportSettings are used with the synchronous MongoClient, it currently won't close the StreamFactoryFactory.  That will require a bit more refactoring than I wanted to do in this PR, and is not likely to be encountered by actual users, as there's no good reason that I know of to use Netty with the synchronous MongoClient.

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/NettyTransportSettings.java
+++ b/driver-core/src/main/com/mongodb/connection/NettyTransportSettings.java
@@ -87,8 +87,7 @@ public final class NettyTransportSettings extends TransportSettings {
         /**
          * Sets the event loop group.
          *
-         * <p>It is highly recommended to supply your own event loop group and manage its shutdown.  Otherwise, the event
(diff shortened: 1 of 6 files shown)


stIncMale:
> there's no good reason that I know of to use Netty with the synchronous MongoClient

Using `EpollSocketChannel` / OpenSSL for TLS / `PreferredDirectByteBufAllocator` (though I don't know about the benefits of the latter, if any), sharing a thread-pool between multiple `MongoClient`s by sharing an `EventLoopGroup`.

stIncMale (approved):
Note: If TransportSettings are used with the synchronous MongoClient, it currently won't close the StreamFactoryFactory. That will require a bit more refactoring than I wanted to do in this PR

Created JAVA-5242.

PR #1245, opened 2023-11-03
Remove getCluster method

jyemin:
from the following internal interfaces:

* ClusterAwareReadWriteBinding
* AsyncClusterAwareReadWriteBinding

The interfaces remain, as they have picket up a second method.  The existing interface names arguably still make sense since implementations have to be "cluster-aware" in order to implement the remaining method.

JAVA-5217


rozza (approved):
LGTM - nice and simple

PR #1250, opened 2023-11-06
Remove deprecated MapCodec and IterableCodec

jyemin:
JAVA-5142


bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
```diff
@@ -464,15 +465,15 @@ public void testCustomRegisteredPropertyCodecOmittedValue() {
     @Test
     public void testMapStringObjectModel() {
         MapStringObjectModel model = new MapStringObjectModel(new HashMap<>(Document.parse("{a : 1, b: 'b', c: [1, 2, 3]}")));
-        CodecRegistry registry = fromRegistries(fromCodecs(new org.bson.codecs.MapCodec()),
-                fromProviders(getPojoCodecProviderBuilder(MapStringObjectModel.class).build()));
+        CodecRegistry registry = fromRegistries(fromProviders(new MapCodecProvider(), new IterableCodecProvider(), new ValueCodecProvider(),
```
stIncMale, line 468: It looks like we additionally need `IterableCodecProvider`, now that we use `MapCodecV2` instead of `MapCodec` via `MapCodecProvider`. This is probably because `MapCodec` used `ValueCodecProvider`, `BsonValueCodecProvider`, `DocumentCodecProvider`, `IterableCodecProvider` under the hood, which is not the case for `MapCodecV2`. Should this be documented somewhere to facilitate migration of user code?

jyemin: I will mark the JAVA ticket as Docs Changes Needed and mention this issue in the comment associated with that.  
In practice, I suspect most people build off whatever is in MongoClientSettings#defaultCodecRegistry and don't create a registry from scratch.

stIncMale (approved):

PR #1253, opened 2023-11-07
Add JsonEncoder and JsonDecoder implmentations to their Bson counterparts

markvogelnomad:
Kotlinx requires the encoder to implement `JsonEncoder` and `JsonDecoder` in order to encode generic `JsonElement` types.

The lack of this support for this came up many times when using [KBson](https://github.com/jershell/kbson) and [KMongo](https://github.com/Litote/kmongo) previously and I'm hoping we can quickly mitigate this now that MongoDB has official support.

I've added each implementation to support most types on decoding and determine the best possible type on encoding to insert. It is somewhat assumed that if someone is encoding a `JsonElement`, then they will probably be decoding one as well and, therefore, type information is primarily for storage optimization.

Unfortunately, Kotlinx writes all numbers as 64 bit integers first and never attempts to check for 32 bits or allow overriding the `JsonElementSerializer` behavior; this means that all integers will probably be written as `int64`. Regardless, this should be a valuable add.


markvogelnomad:
I'm unsure what checks here are failing. I see that "patch must be manually authorized" so I'm thinking this is not something I need to fix. Is that correct?

markvogelnomad:
Appreciate it!

I've pushed one more non-functional change to attempt to comply with the checks; I will leave it alone at this point to allow feedback or modification, if required.

tom-selander:
Hi @markvogelnomad thanks for the PR. We are backlogging this ticket for now, as we would like to dedicate a little more time reviewing this PR and ensuring we do our proper due diligence before merging. Please follow [JAVA-5239](https://jira.mongodb.org/browse/JAVA-5239) for updates.

rozza:
Closing in favour of #1459. I used this as the base and added you as a co-author in the commit.
The core difference is feature detection is used to see if `json` support is available, rather than taking a hard dependency.

PR #1257, opened 2023-11-09
Remove deprecated methods in WriteConcernError class

jyemin:
JAVA-5153

This requires a bit of refactoring in order to remove any usage of the errorLabels in WriteConcernError, and to consolidate error label addition in MongoWriteConcernException

### DIFF
--- a/driver-core/src/main/com/mongodb/MongoWriteConcernException.java
+++ b/driver-core/src/main/com/mongodb/MongoWriteConcernException.java
@@ -19,8 +19,8 @@ package com.mongodb;
 import com.mongodb.bulk.WriteConcernError;
 import com.mongodb.lang.Nullable;
 
+import java.util.Collection;
 import java.util.Collections;
-import java.util.Set;
 
 import static com.mongodb.assertions.Assertions.notNull;
 
@@ -43,7 +43,7 @@ public class MongoWriteConcernException extends MongoServerException {
      *
      * @param writeConcernError the non-null write concern error
      * @param serverAddress the non-null server address
-     * @deprecated Prefer {@link MongoWriteConcernException(WriteConcernError, WriteConcernResult, ServerAddress, Set)}
+     * @deprecated Prefer {@link MongoWriteConcernException(WriteConcernError, WriteConcernResult, ServerAddress, Collection)}
      */
     @Deprecated
     public MongoWriteConcernException(final WriteConcernError writeConcernError, final ServerAddress serverAddress) {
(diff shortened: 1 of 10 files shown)


driver-core/src/main/com/mongodb/MongoWriteConcernException.java
```diff
@@ -40,7 +42,9 @@ public class MongoWriteConcernException extends MongoServerException {
      *
      * @param writeConcernError the non-null write concern error
      * @param serverAddress the non-null server address
+     * @deprecated Prefer {@link MongoWriteConcernException(WriteConcernError, WriteConcernResult, ServerAddress, Set)}
      */
+    @Deprecated
```
jyemin, line 47: We have to keep these around for another major release cycle. :(

driver-core/src/main/com/mongodb/MongoWriteConcernException.java
```diff
@@ -52,24 +57,31 @@ public MongoWriteConcernException(final WriteConcernError writeConcernError, fin
...
+     * Construct an instance.
+     *
+     * @param writeConcernError the non-null write concern error
+     * @param writeConcernResult the write result
+     * @param serverAddress     the non-null server address
+     * @param errorLabels       the server errorLabels
+     * @since 5.0
+     */
+    public MongoWriteConcernException(final WriteConcernError writeConcernError, @Nullable final WriteConcernResult writeConcernResult,
+            final ServerAddress serverAddress, final Set<String> errorLabels) {
```
stIncMale, line 78: Note how `MongoException.addLabels` accepts `Collection`, not `Set`: all the method needs is the ability to iterate the given `errorLabels` (arguably, it should have been `Iterable`). Thus, I don't think there is a reason for us to restrict the type of `errorLabels` to `Set`, we can also accept `Collection` here.

This comment is also applicable to the new constructor of `MongoWriteException`.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/MongoWriteConcernException.java
jyemin: `MongoException.addLabels` probably should have been accepted a Set, since ultimately duplicates are discarded.  Since it's protected it's not a big deal to change its signature.  What do you think?

stIncMale: The absence of duplicates is important only when labels are read, and is expressed by `getErrorLabels` returning `Set`. By requiring a `Set` in `addLabels` (or in the new constructor), we are coercing a user to convert from whatever container he has into a `Set`, seemingly for no good reason: we are still relying only on being able to iterate through the elements of the argument of the `addLabels` method.

It's not an important thing, so I marked the thread as optional in case you are not convinced.

stIncMale (approved):

driver-core/src/main/com/mongodb/MongoWriteConcernException.java
jyemin: I buy the argument. Done.

PR #1260, opened 2023-11-15
Updates bson-kotlin DataClassCodec.kt to allow the usage of default values

filipeportes:
avoids providing null to parameters that are not nullable, allowing for the usage of default values. 

example: 

given a data Class: 
```
data class Item(val name:String, val category:String = "default")
```

the following objects will produce `CodecConfigurationException: Unable to invoke primary constructor of Item data class`:
```
[
 {
  name: "item1"
 },
 {
  name: "item2",
  category: null
 }
]
```

with the changes proposed in this PR, the result will be decoding the documents and using the default values in case no actual values are provided.

PS: 
* In case the parameter is required and no default value is provided, the current behaviour of throwing an exception will continue as expected.
* In case the parameter is nullable with a default value and a null value is provided, the current behaviour of using null will continue as expected


filipeportes:
Hi @jyemin could you take a look please 🙏

rozza:
The kotlin bson codec already supports default values for data classes. This PR is something slightly different, it wants to use default values where the data stored is a `BsonNull`. 

What is the correct thing to when a field is nullable and there is a `null` value in the database? eg:

```
data class Item(val name:String, val category:String? = "default")
```

Will all users want to allow a `null` value in the database to be replaced as the default value for a data class?

filipeportes:
> What is the correct thing to when a field is nullable and there is a `null` value in the database? eg:
> 
> ```
> data class Item(val name:String, val category:String? = "default")
> ```
> 
> Will all users want to allow a `null` value in the database to be replaced as the default value for a data class?

hey @rozza thanks for taking a look 🙏 

the changes on the PR will not affect the case you described, since in this case category is nullable, the filter on `isRequiredParameterWithNullValue` will not get this field and if the value is a `BsonNull` it will be passed to the constructor as usual

filipeportes:
Hi @rozza could you take another quick look?

rozza:
Hi @filipeportes,

Thanks for your perservance with this issue. I've added [JAVA-5281](https://jira.mongodb.org/browse/JAVA-5281) to track it. This will be triaged and if successful scheduled for a future release.

I will update the ticket here also, the open question is, if the change in behavior is the correct change to make and will it also apply to bson-kotlinx.|

Ross

rozza:
Just to update it looks like there is a way for kotlinx to do this with json, via configuration. 

So its now on the backlog for applying to both kotlin-bson and kotlinx-bson. However, there is no firm timeline as to when this work will be scheduled. Please watch the ticket for further details.

Ross

rozza:
Hi @filipeportes,

Just to update there is no firm timeline as to when this work will be scheduled. Please watch the ticket ([JAVA-5281](https://jira.mongodb.org/browse/JAVA-5281)) for updates.

As the solution will require further changes (eg configuration changes and bson kotlinx support), I'm going to close this ticket out.

All the best,

Ross

filipeportes:
Thanks a lot for considering the update and following up with all details, really appreciate it!

PR #1263, opened 2023-11-20
Support `authorizedCollections` option for `listCollections` helpers

stIncMale:
This PR is based on the original https://github.com/mongodb/mongo-java-driver/pull/855.

#### `MongoDatabase.listCollectionNames`

Changed the return type of a method. The change preserves source-code compatibility (existing source code that uses the changed methods does not have to be changed), breaks binary compatibility (existing source code that uses the changed methods has to be recompiled against the updated driver dependency).

| package | class/interface | method(s) | old return type | new return type |
| --- | --- | --- | --- | --- |
| `com.mongodb.client` | `MongoDatabase` | `listCollectionNames` | `MongoIterable` | `ListCollectionsIterable` |
| `com.mongodb.reactivestreams.client` | `MongoDatabase` | `listCollectionNames` | `Publisher` | `ListCollectionsPublisher` |
| `org.mongodb.scala` | `MongoDatabase` | `listCollectionNames` | `Observable` | `ListCollectionsObservable` |
| `com.mongodb.kotlin.client` | `MongoDatabase` | `listCollectionNames` | `MongoIterable` | `ListCollectionsIterable` |
| `com.mongodb.kotlin.client.coroutine` | `MongoDatabase` | `listCollectionNames` | `Flow` | `ListCollectionsFlow` |

#### `authorizedCollections`

Added a method to an interface/class. The change preserves source-code and binary compatibility when done to a class, but breaks source-code compatibility for implementors when done to an interface. For the interfaces affected, we disregard such a change as breaking nonetheless.

| package | class/interface | new method |
| --- | --- | --- |
| `com.mongodb.client` | `ListCollectionsIterable` | `authorizedCollections` |
| `com.mongodb.reactivestreams.client` | `ListCollectionsPublisher` | `authorizedCollections` |
| `org.mongodb.scala` | `ListCollectionsObservable` | `authorizedCollections` |
| `com.mongodb.kotlin.client` | `ListCollectionsIterable` | `authorizedCollections` |
| `com.mongodb.kotlin.client.coroutine` | `ListCollectionsFlow` | `authorizedCollections` |

### API design

Note how `authorizedCollections` is documented to be ignored in some situations. There seem to be two alternatives:

1. Throw a runtime exception in a situation when we ignore the new option. I didn't do this because the server ignores this situation (the behavior is documented), though it does not make sense to me. I am afraid that someone may end up having a security bug caused by this behavior.
2. Introduce subtypes for each interface/class where `authorizedCollections` is added, add this method to the new types, return these types from `listCollectionNames`. I didn't try this approach in the original PR, maybe because it seemed like more work, or maybe because the approach didn't occur to me. Will see what reviewers think about this.

### Other notes

You may see a few methods marked as `@VisibleForTesting`, which are used reflectively either from Java tests, or from Groovy. The code that effectively required the presence of these methods was already in the tests, I simply made the changes to fit into existing test infrastructure.

JAVA-4353


driver-core/src/main/com/mongodb/internal/operation/ListCollectionsOperation.java
```diff
@@ -209,6 +231,9 @@ private BsonDocument getCommand() {
         }
         if (nameOnly) {
             command.append("nameOnly", BsonBoolean.TRUE);
+            if (authorizedCollections) {
+                command.append("authorizedCollections", BsonBoolean.TRUE);
+            }
```
stIncMale, line 236: The documentation says: [_"When used without nameOnly: true, this option has no effect."_](https://docs.mongodb.com/manual/reference/command/listCollections/#:~:text=When%20used%20without%20nameOnly%3A%20true%2C%20this%20option%20has%20no%20effect.) That's why we send the option only if `nameOnly` is `true`.

driver-sync/src/main/com/mongodb/client/ListCollectionsIterable.java
```diff
@@ -39,6 +40,21 @@ public interface ListCollectionsIterable<TResult> extends MongoIterable<TResult>
      */
     ListCollectionsIterable<TResult> filter(@Nullable Bson filter);
 
+    /**
+     * Sets the {@code authorizedCollections} field of the {@code listCollections} command.
+     * This method is ignored if called on a {@link ListCollectionsIterable} obtained not via any of the
+     * {@link MongoDatabase#listCollectionNames() MongoDatabase.listCollectionNames} methods.
```
stIncMale, line 46: There are alternatives to ignoring the option, as described in the "API design" section of the PR description.

driver-core/src/main/com/mongodb/internal/operation/ListCollectionsOperation.java
rozza: If this changes in the future we'll have to make a code change, for that reason I would include it in the command if its set.

You can use the `putIfTrue` helper to ammend the command document.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ListCollectionsPublisherImpl.java
```diff
@@ -76,6 +90,77 @@ public ListCollectionsPublisher<T> comment(@Nullable final BsonValue comment) {
 
     AsyncReadOperation<AsyncBatchCursor<T>> asAsyncReadOperation(final int initialBatchSize) {
         return getOperations().listCollections(getNamespace().getDatabaseName(), getDocumentClass(), filter, collectionNamesOnly,
-                initialBatchSize, maxTimeMS, comment);
+                authorizedCollections, initialBatchSize, maxTimeMS, comment);
+    }
+
+    <U> ListCollectionsPublisher<U> map(final Function<T, U> mapper) {
```
rozza, line 96: Is a map method needed? Doesn't seem to be a pattern we use else where for publishers?

rozza (commented):
I'm not 100% sure its worth adding a ListCollection Mapper type to wrap the Observable / Iterable. As `listCollectionNames` is a helper built on top of `listCollections` it be easier if it automatically sets `authorizedCollections=true`. 

What do you think?

The listCollections and listDatabases API should be inline - so either way this may require another ticket to bring the listDatabase method inline.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/ListCollectionsPublisherImpl.java
stIncMale: ### 1) I'm not 100% sure its worth adding a ListCollection Mapper type to wrap the Observable / Iterable.

If we don't have

- `ListCollectionsIterableImpl.map` returning `ListCollectionsIterable`
- `ListCollectionsPublisherImpl.map` returning `ListCollectionsPublisher`

how will

- `com.mongodb.client.internal.MongoDatabaseImpl.listCollectionNames`
- `com.mongodb.reactivestreams.client.internal.MongoDatabaseImpl.listCollectionNames`

be implemented?

### 2) As listCollectionNames is a helper built on top of listCollections it be easier if it automatically sets authorizedCollections=true.

`listCollectionNames` can set `authorizedCollections=true` automatically only if it know when a user wants `authorizedCollections=true`, and when a user doesn't want it. But how will `listCollectionNames` know that information without a user explicitly providing it?

### 3) The listCollections and listDatabases API should be inline - so either way this may require another ticket to bring the listDatabase method inline.

1. They have already been different: there is `ListDatabasesIterable.nameOnly`, while the same option for `ListCollectionsIterable` can be set only internally by the driver when `listCollectionNames` is called.
2. The `ListCollectionsPublisherImpl.map` method added in the PR is not part of API, it's an internal method. While a similar method could be added to `ListDatabasesPublisherImpl`, it will be unused. I don't think we should add unused internal method to `ListDatabasesPublisherImpl` only because a similar method is present on `ListCollectionsPublisherImpl`.

stIncMale:
> Regarding the API design, I am inclined to go with option 2 and introduce new ListCollectionNamesIterable and related interfaces.

This turned out surprisingly non-trivial. The approach I tried yesterday was a dead end. I have another one in mind.

driver-core/src/main/com/mongodb/internal/operation/ListCollectionsOperation.java
stIncMale: Done.

stIncMale:
> Regarding the API design, I am inclined to go with option 2 and introduce new ListCollectionNamesIterable and related interfaces.

Done in https://github.com/mongodb/mongo-java-driver/pull/1270.

PR #1264, opened 2023-11-20
Simplify async Stream implementations

jyemin:
Somehow we ended up with three nearly identical classes, all of which were unnecessarily complicated.

Remove two of them and simplify the remaining one to use CompletableFuture.


driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
```diff
@@ -277,54 +273,4 @@ AsyncCompletionHandler<T> getHandlerAndClear() {
...
-
-        void getWrite() throws IOException {
-            get("Writing to");
-        }
-
-        T getRead() throws IOException {
-            return get("Reading from");
-        }
-
-        private T get(final String prefix) throws IOException {
```
jyemin, line 310: The prefix seems unnecessary so I disappeared it in the "refactoring".

PR #1274, opened 2023-12-07
Pull mongohouse image from ADL ECR repo

jyemin:
JAVA-5235


jyemin:
Just realized it would probably be better to split this into two PRs.  WDYT?

rozza (approved):
LGTM - although whats happened to the patch build?

PR #1279, opened 2023-12-12
Accept `long` instead of `int` in `SocketSettings.Builder.connectTimeout`/`readTimeout`

stIncMale:
JAVA-3897

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/SocketSettings.java
+++ b/driver-core/src/main/com/mongodb/connection/SocketSettings.java
@@ -101,7 +101,7 @@ public final class SocketSettings {
          * @return this
          */
         public Builder connectTimeout(final long connectTimeout, final TimeUnit timeUnit) {
-            this.connectTimeoutMS = toIntExact(MILLISECONDS.convert(connectTimeout, timeUnit));
+            this.connectTimeoutMS = timeoutArgumentToMillis(connectTimeout, timeUnit);
             return this;
         }
 
@@ -115,7 +115,7 @@ public final class SocketSettings {
          * @see #getReadTimeout(TimeUnit)
          */
         public Builder readTimeout(final long readTimeout, final TimeUnit timeUnit) {
-            this.readTimeoutMS = toIntExact(MILLISECONDS.convert(readTimeout, timeUnit));
+            this.readTimeoutMS = timeoutArgumentToMillis(readTimeout, timeUnit);
             return this;
         }
 
@@ -283,5 +283,14 @@ public final class SocketSettings {
         receiveBufferSize = builder.receiveBufferSize;
         sendBufferSize = builder.sendBufferSize;
         proxySettings = builder.proxySettingsBuilder.build();
+    }
+
+    private static int timeoutArgumentToMillis(final long timeout, final TimeUnit timeUnit) throws IllegalArgumentException {
+        try {
+            return toIntExact(MILLISECONDS.convert(timeout, timeUnit));
+        } catch (ArithmeticException e) {
+            throw new IllegalArgumentException(
+                    "The timeout converted to milliseconds must not be greater than `Integer.MAX_VALUE`", e);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


driver-core/src/main/com/mongodb/connection/SocketSettings.java
```diff
@@ -93,25 +95,27 @@ public Builder applySettings(final SocketSettings socketSettings) {
         /**
          * Sets the socket connect timeout.
          *
-         * @param connectTimeout the connect timeout
+         * @param connectTimeout the connect timeout.
+         * The timeout converted to milliseconds must not be greater than {@link Integer#MAX_VALUE}.
```
stIncMale, line 99: We had this requirement for forever (the `toIntExact` conversion does not, therefore, introduce a new restriction), but did not document it. Now it is documented.

driver-core/src/main/com/mongodb/connection/SocketSettings.java
```diff
@@ -260,12 +264,7 @@ public boolean equals(final Object o) {
 
     @Override
     public int hashCode() {
-        int result = (int) (connectTimeoutMS ^ (connectTimeoutMS >>> 32));
-        result = 31 * result + (int) (readTimeoutMS ^ (readTimeoutMS >>> 32));
-        result = 31 * result + receiveBufferSize;
-        result = 31 * result + sendBufferSize;
-        result = 31 * result + proxySettings.hashCode();
-        return result;
```
stIncMale, line 268: The bit shifting in the old code became irrelevant, i.e., `hashCode` had to be updated. I updated it to use `Objects.hash`.

driver-core/src/main/com/mongodb/connection/SocketSettings.java
```diff
@@ -62,8 +64,8 @@ public static Builder builder(final SocketSettings socketSettings) {
      * A builder for an instance of {@code SocketSettings}.
      */
     public static final class Builder {
-        private long connectTimeoutMS = 10000;
-        private long readTimeoutMS;
+        private int connectTimeoutMS = 10000;
+        private int readTimeoutMS;
```
stIncMale, line 68: There is no reason to have the fields internally of the `long` type, because the millisecond values were not allowed (and will continue being not allowed) to be greater than `Integer.MAX_VALUE`.

driver-core/src/main/com/mongodb/connection/SocketSettings.java
```diff
@@ -93,25 +95,27 @@ public Builder applySettings(final SocketSettings socketSettings) {
...
-         * @param connectTimeout the connect timeout
+         * @param connectTimeout the connect timeout.
+         * The timeout converted to milliseconds must not be greater than {@link Integer#MAX_VALUE}.
          * @param timeUnit the time unit
          * @return this
          */
-        public Builder connectTimeout(final int connectTimeout, final TimeUnit timeUnit) {
-            this.connectTimeoutMS = MILLISECONDS.convert(connectTimeout, timeUnit);
+        public Builder connectTimeout(final long connectTimeout, final TimeUnit timeUnit) {
+            this.connectTimeoutMS = toIntExact(MILLISECONDS.convert(connectTimeout, timeUnit));
```
jyemin, line 104: Would this be a new thing to have the driver throw `ArithmeticException` on an invalid value from the application?  Should the exception instead be wrapped in `IllegalArgumentException`, making it similar to `ConnectionString`?

jyemin, line 104: Consider adding a test for the new behavior.

jyemin (changes requested):

driver-core/src/main/com/mongodb/connection/SocketSettings.java
stIncMale: Done.

stIncMale: Thank you, done.

jyemin (approved):

vbabanin (approved):

PR #1280, opened 2023-12-13
Make `ConnectionId.serverValue`/`localValue` of the `Long`/`long` type

stIncMale:
The corresponding spec change: https://github.com/mongodb/specifications/commit/1f272b30a2e9714578b71a02b2836d5a0bd853f8.

JAVA-4846

### DIFF
--- a/driver-core/src/main/com/mongodb/connection/ConnectionId.java
+++ b/driver-core/src/main/com/mongodb/connection/ConnectionId.java
@@ -20,7 +20,7 @@ import com.mongodb.annotations.Immutable;
 import com.mongodb.lang.Nullable;
 
 import java.util.Objects;
-import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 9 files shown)


driver-core/src/main/com/mongodb/connection/ConnectionId.java
```diff
@@ -38,8 +38,9 @@ public final class ConnectionId {
     private static final AtomicInteger INCREMENTING_ID = new AtomicInteger();
```
jyemin, line 38: It's internal, but for clarity should we also change this to `AtomicLong`?

stIncMale: Thanks, we should update it.

Done.

jyemin (approved):

vbabanin (approved):

PR #1282, opened 2023-12-15
Expose `Observable[Unit]` instead of `Observable[Void]`

stIncMale:
This PR is a proof that even a single ape typing behind a keyboard can produce a PR that compiles. Vigilance from reviewers is especially needed here.

JAVA-4303

### DIFF
--- a/driver-scala/src/main/scala/org/mongodb/scala/Observable.scala
+++ b/driver-scala/src/main/scala/org/mongodb/scala/Observable.scala
@@ -49,6 +49,12 @@ object Observable {
  *
  * Extends the `Publisher` interface and adds helpers to make Observables composable and simple to Subscribe to.
  *
+ * Special parameterizations:
+ *
+ *   - `Observable[Unit]` must emit exactly one item by signalling [[Observer.onNext]]
+ *   if it terminates successfully by signalling [[Observer.onComplete]].
+ *   - `Observable[Void]` cannot emit an item. It is not exposed by the driver API because it is not convenient to work with in Scala.
+ *
  * @define forComprehensionExamples
  *         Example:
  *
@@ -464,5 +470,9 @@ trait Observable[T] extends Publisher[T] {
    * @return a single observable which emits Unit before completion.
    * @since 4.4
    */
+  @deprecated(
+    "Is no longer needed because of the `ToSingleObservableUnit` implicit class. Scheduled for removal in a major release",
+    "5.0"
+  )
   def completeWithUnit(): SingleObservable[Unit] = UnitObservable(this)
 }
--- a/driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
+++ b/driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
@@ -16,8 +16,6 @@
 
 package org.mongodb.scala
 
-import com.mongodb.reactivestreams.client.gridfs.GridFSUploadPublisher
-import org.bson.BsonValue
 import org.mongodb.scala.bson.ObjectId
 import org.mongodb.scala.gridfs.GridFSFile
 import org.mongodb.scala.internal.{ MapObservable, UnitObservable }
@@ -116,25 +114,6 @@ trait ObservableImplicits {
       extends SingleObservable[GridFSFile] {
     val publisher = pub
     override def subscribe(observer: Observer[_ >: GridFSFile]): Unit = Mono.from(publisher).subscribe(observer)
-  }
-
-  /**
-   * A `GridFSUploadPublisher`` that emits
-   *
-   *   - exactly one item, if the wrapped `Publisher` does not signal an error, even if the represented stream is empty;
-   *   - no items if the wrapped `Publisher` signals an error.
-   *
-   * @param pub A `Publisher` representing a finite stream.
-   */
-  implicit class ToGridFSUploadPublisherUnit(pub: => GridFSUploadPublisher[Void]) extends GridFSUploadPublisher[Unit] {
-    val publisher = pub
-
-    override def subscribe(observer: Subscriber[_ >: Unit]): Unit =
-      Flux.from(publisher).reduce((), (_: Unit, _: Void) => ()).subscribe(observer)
-
-    override def getObjectId: ObjectId = publisher.getObjectId
-
-    override def getId: BsonValue = publisher.getId
   }
 
   /**
--- a/driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
+++ b/driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
@@ -16,9 +16,11 @@
 
 package org.mongodb.scala
 
+import com.mongodb.reactivestreams.client.gridfs.GridFSUploadPublisher
+import org.bson.BsonValue
 import org.mongodb.scala.bson.ObjectId
 import org.mongodb.scala.gridfs.GridFSFile
-import org.mongodb.scala.internal.MapObservable
+import org.mongodb.scala.internal.{ MapObservable, UnitObservable }
 import org.reactivestreams.{ Publisher, Subscriber, Subscription => JSubscription }
 import reactor.core.publisher.{ Flux, Mono }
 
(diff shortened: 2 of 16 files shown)


driver-scala/src/main/scala/org/mongodb/scala/Observable.scala
```diff
@@ -464,5 +470,9 @@ trait Observable[T] extends Publisher[T] {
    * @return a single observable which emits Unit before completion.
    * @since 4.4
    */
+  @deprecated(
+    "Is no longer needed because of the `ToSingleObservableUnit` implicit class. Scheduled for removal in a major release",
+    "5.0"
+  )
   def completeWithUnit(): SingleObservable[Unit] = UnitObservable(this)
```
stIncMale, line 477: Note that I propose to remove this method in the future.

driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
```diff
@@ -116,17 +118,41 @@ trait ObservableImplicits {
...
+
+  /**
+   * An [[Observable]] that emits
+   *
+   *   - exactly one item, if the wrapped `Publisher` does not signal an error, even if the represented stream is empty;
+   *   - no items if the wrapped `Publisher` signals an error.
+   *
+   * @param pub A `Publisher` representing a finite stream.
+   */
+  implicit class ToSingleObservableUnit(pub: => Publisher[Void]) extends SingleObservable[Unit] {
```
stIncMale, line 148: This class is why we are able to simply replace `Observable[Void]` with `Observable[Unit]` in the codebase.

driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
```diff
@@ -116,17 +118,41 @@ trait ObservableImplicits {
...
-  implicit class ToSingleObservableVoid(pub: => Publisher[Void]) extends SingleObservable[Void] {
+  /**
+   * A `GridFSUploadPublisher`` that emits
+   *
+   *   - exactly one item, if the wrapped `Publisher` does not signal an error, even if the represented stream is empty;
+   *   - no items if the wrapped `Publisher` signals an error.
+   *
+   * @param pub A `Publisher` representing a finite stream.
+   */
+  implicit class ToGridFSUploadPublisherUnit(pub: => GridFSUploadPublisher[Void]) extends GridFSUploadPublisher[Unit] {
```
stIncMale, line 129: This class is why we are able to simply replace `GridFSUploadPublisher[Void]` with `GridFSUploadPublisher[Unit]` in the codebase.

driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
```diff
@@ -116,17 +118,41 @@ trait ObservableImplicits {
     override def subscribe(observer: Observer[_ >: GridFSFile]): Unit = Mono.from(publisher).subscribe(observer)
   }
 
-  implicit class ToSingleObservableVoid(pub: => Publisher[Void]) extends SingleObservable[Void] {
```
stIncMale, line 119: `ToSingleObservableVoid` is part of public API (https://mongodb.github.io/mongo-java-driver/4.11/apidocs/mongo-scala-driver/org/mongodb/scala/ObservableImplicits$ToSingleObservableVoid.html), but my understanding is that

1. There was never a reason for any user to use it explicitly (though I found a single explicit import here https://github.com/hmrc/event-hub/blob/main/app/uk/gov/hmrc/eventhub/service/TransactionHandlerImpl.scala#L19).
2. The implicit conversion was convenient for the driver code, and is completely unneeded for any user of the Scala driver API, because it does not expose `Publisher[Void]`. I wonder if `ToSingleObservableVoid` is part of public API simply because there is no way to remove it from there, even though users don't need it.

For the aforementioned reasons, it seems to me that removing `ToSingleObservableVoid` and introducing `ToSingleObservableUnit` is fine (they can't coexist).

rozza: Yeap its an implicit class, so using it directly is strange. However, having it in scope makes it to be usable in customers code.  Either way this is a breaking change but fixes a bug so makes sense.

driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
rozza: This doesn't need to be a global implicit so can be added here instead: `gridfs/package.scala`

driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
rozza: implicits are magic and have been changed in Scala 3 in favour of extension methods (probably more akin to Kotlin extension methods).

driver-scala/src/test/scala/org/mongodb/scala/internal/UnitObservableSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.mongodb.scala.internal
+
+import org.mongodb.scala.{ BaseSpec, Observer, Subscription }
+
+class UnitObservableSpec extends BaseSpec {
```
rozza, line 20: Lets add a chainability case to ensure this actually fixes the reported issue.

```
  it should "work with for comprehensions" in {
    val h = for {
      _ <- UnitObservable(TestObservable(1 to 2))
      _ <- UnitObservable(TestObservable(20 to 30))
    } yield List(1, 2, 3)

    val results = ArrayBuffer[Int]()
    var completed = false
    h.subscribe((s: List[Int]) => results.addAll(s), (t: Throwable) => t, () => completed = true)
    results should equal(List(1, 2, 3))
    completed should equal(true)
  }
```

rozza (changes requested):
One minor change - code move then LGTM

driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
stIncMale: Done.

driver-scala/src/test/scala/org/mongodb/scala/internal/UnitObservableSpec.scala
stIncMale: Thank you, done.

driver-scala/src/main/scala/org/mongodb/scala/Observable.scala
jyemin: When do you propose to remove it?  Prior to 5.0, from the annotation, I gather?

stIncMale: 5.0 is the version from which the method is deprecated. Once it's deprecated, we'll be able to remove it in the next major release, which is 6.0.

stIncMale: If we were to release a 4.x version with the deprecation, then we'll be able to delete the method in 5.0.

jyemin: So should the `"5.0"` become `"6.0"` in the annotation?

jyemin (commented):
Added a comment to the thread re deprecations

driver-scala/src/main/scala/org/mongodb/scala/Observable.scala
stIncMale: I don't think so, because "5.0" is the value of the `since` attribute, not a version in which we promise to remove the method.

jyemin: Well that makes a lot of sense :)

jyemin (approved):

driver-scala/src/main/scala/org/mongodb/scala/ObservableImplicits.scala
rozza: yeap 👍

rozza (approved):
LGTM

PR #1285, opened 2023-12-28
Reclaim native memory quicker when using ZlibCompressor

raelg:
Reclaim native memory quicker by overriding `close()` to invoke `end()` on the deflater, since by supplying a deflater to the constructor of `DeflaterOutputStream` `usesDefaultDeflater` is set to `false` and therefore `end()` is not invoked in [DeflaterOutputStream#close](https://github.com/openjdk/jdk/blob/2a59243cbaf3e7d5d1bfc9f247d28bc648687ea5/src/java.base/share/classes/java/util/zip/DeflaterOutputStream.java#L241).

This is related to [JDK-4797189]( https://bugs.openjdk.org/browse/JDK-4797189) bug/feature (marked as Won't Fix)

See the following blog post for more information: https://medium.com/swlh/native-memory-the-silent-jvm-killer-595913cba8e7

JAVA-5280

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ZlibCompressor.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ZlibCompressor.java
@@ -49,7 +49,7 @@ class ZlibCompressor extends Compressor {
 
     @Override
     OutputStream getOutputStream(final OutputStream source) {
-        return new DeflaterOutputStream(source, new Deflater(level)){
+        return new DeflaterOutputStream(source, new Deflater(level)) {
             @Override
             public void close() throws IOException {
                 try {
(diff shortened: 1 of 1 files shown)


jyemin:
Hi @raelg , I opened JAVA-5280 and linked this PR to it.  We'll take a look soon.

driver-core/src/main/com/mongodb/internal/connection/ZlibCompressor.java
```diff
@@ -48,6 +49,15 @@ InputStream getInputStream(final InputStream source) {
 
     @Override
     OutputStream getOutputStream(final OutputStream source) {
-        return new DeflaterOutputStream(source, new Deflater(level));
+        return new DeflaterOutputStream(source, new Deflater(level)){
```
jyemin, line 52:
```suggestion
        return new DeflaterOutputStream(source, new Deflater(level)) {
```

jyemin (approved):
Looks good. Just one whitespace issue to fix.

PR #1291, opened 2024-01-04
Fix `onErrorDropped` logged message (#1281)

rozza:
Caused by using `doOnError` rather than using `subscribe` and passing the handler in there.

JAVA-5284 JAVA-5266


stIncMale (commented):
I was confused by the PR name, because I understand backporting as cherry-picking something into the branch for an older major/minor version (e.g., from `master` to `4.11.x`). Now I see that this is effectively a cherry-pick of https://github.com/mongodb/mongo-java-driver/pull/1281 from the `CSOT` branch.

1. Could we not name the commit "backport"?
2. Why do we need to cherry-pick this change from `CSOT` to `master`? Is it because we want the fix to be released in 5.0 before CSOT is done?

rozza:
> 1. Could we not name the commit "backport"?
Sure we can
> 2. Why do we need to cherry-pick this change from CSOT to master? Is it because we want the fix to be released in 5.0 before CSOT is done?
We don't **need** to cherry-pick it but it is a fix for a bug that exists in master. 

The reason to backport it from the CSOT branch into master is because all changes from CSOT will be squashed into a single mega commit. This loses any history of certain changes that were found during CSOT but dont necessarily directly pertain to CSOT.

stIncMale (approved):
Thank you for the explanation.

PR #1293, opened 2024-01-10
Simplify customization of bson-kotlinx

rozza:
Allow configuration to be set on the KotlinSerializerCodecProvider

JAVA-5276

### DIFF
--- a/bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecProviderTest.kt
+++ b/bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecProviderTest.kt
@@ -20,6 +20,18 @@ import kotlin.test.assertEquals
 import kotlin.test.assertNotNull
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.modules.plus
+import kotlinx.serialization.modules.polymorphic
+import kotlinx.serialization.modules.subclass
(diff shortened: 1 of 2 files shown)


bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecProviderTest.kt
```diff
@@ -60,4 +72,34 @@ class KotlinSerializerCodecProviderTest {
...
+                .get(DataClassContainsOpen::class.java, Bson.DEFAULT_CODEC_REGISTRY)!!
+
+        assertTrue { codec is KotlinSerializerCodec }
+        val encodedDocument = BsonDocument()
+        val writer = BsonDocumentWriter(encodedDocument)
+        codec.encode(writer, dataClassContainsOpenB, EncoderContext.builder().build())
+        writer.flush()
+
+        assertEquals(
+            BsonDocument.parse("""{"open": {"__type": "org.bson.codecs.kotlinx.samples.DataClassOpenB", "b": 1}}"""),
```
stIncMale, line 102: [optional]
We may also test that decoding this BSON succeeds and produces an object equal to `dataClassContainsOpenB`. This way we'll know that the custom `classDiscriminator` is used not only by the encoder, but also by the decoder.

stIncMale (approved):
There is one optional suggestion, if you find it helpful.

bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecProviderTest.kt
rozza: Done

PR #1295, opened 2024-01-16
Handle kotlin / JVM erasure of types

rozza:
It has been observed that the Kotlin can behave
differently during reflection and return Type<Object> where normally it would report the correct type.

Checking for erasure allows the bson-kotlin library to handle these cases and return the expected type.

JAVA-5292


vbabanin (approved):
LGTM!

PR #1301, opened 2024-01-31
Enhance KotlinSerializer with value codecs for widening primitive conversion.

vbabanin:
JAVA-5303


rozza:
Kotlinx Json has a configuration to allow for lenient parsing: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#lenient-parsing

Should this be configurable?

jyemin:
@rozza I think in this case it's better to be consistent with other codecs and also with the way numeric values are treated in MongoDB is basically interchangeable and compatible with each other, so long as there is no loss of precision or magnitude.

jyemin:
@vbabanin can this come out of Draft now?

config/spotbugs/exclude.xml
```diff
@@ -217,6 +217,12 @@
         <Method name="~.*validateAnnotations.*"/>
         <Bug pattern="UC_USELESS_OBJECT"/>
     </Match>
+    <Match>
+        <!-- MongoDB status: "False Positive", SpotBugs rank: 17 -->
+        <Class name="org.bson.codecs.kotlinx.KotlinSerializerCodec$Companion"/>
+        <Method name="~.*validateAnnotations.*"/>
+        <Bug pattern="BC_BAD_CAST_TO_ABSTRACT_COLLECTION"/>
```
vbabanin, line 224: I've noticed that the CodecRegistry, while being passed to the get method of a CodecProvider, is not actually utilized in the current implementation. You can see this at [KotlinSerializerCodecProvider.kt Line 35](https://github.com/vbabanin/mongo-java-driver/blob/baebf9074c69300f7bd70abcd93802d2eaa0766f/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecProvider.kt#L35).

It raises a question: Should we consider leveraging this CodecRegistry to obtain value codecs instead of relying on hardcoded implementations?

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
```diff
@@ -34,8 +34,18 @@ import org.bson.BsonReader
 import org.bson.BsonReaderMark
 import org.bson.BsonType
 import org.bson.BsonValue
+import org.bson.codecs.BooleanCodec
```
rozza, line 37: I thought we discussed not using codecs (so to not confuse the two concepts of codecs and kserializers. Instead use the `NumberCodecHelper` directly (which is what the numeric codecs are doing in decode anyway).

rozza (changes requested):

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
vbabanin: Thank you for the reminder. We did discuss prioritizing clarity by not mixing codecs with serializers. I've now implemented the changes to use the codec helpers directly as we discussed.

config/spotbugs/exclude.xml
rozza: This still needded?

bson/src/main/org/bson/internal/StringCodecHelper.java
```diff
@@ -0,0 +1,41 @@
...
+ * This class is not part of the public API. It may be removed or changed at any time.
+ */
+public final class StringCodecHelper {
+
+    private StringCodecHelper(){
+        //NOP
+    }
+
+    public static char decodeChar(final BsonReader reader) {
+        String string = reader.readString();
```
rozza, line 34: Lets add a check if the bsonType isnt a string type. 

`throw new BsonInvalidOperationException(format("Invalid string type, found: %s", bsonType));`

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
```diff
@@ -154,15 +156,18 @@ internal open class DefaultBsonDecoder(
...
+    override fun decodeChar(): Char = StringCodecHelper.decodeChar(reader)
+    override fun decodeFloat(): Float = NumberCodecHelper.decodeFloat(reader)
+
+    override fun decodeShort(): Short = NumberCodecHelper.decodeShort(reader)
+    override fun decodeBoolean(): Boolean = reader.readBoolean()
+
+    override fun decodeDouble(): Double = NumberCodecHelper.decodeDouble(reader)
+    override fun decodeInt(): Int = NumberCodecHelper.decodeInt(reader)
+    override fun decodeLong(): Long = NumberCodecHelper.decodeLong(reader)
+    override fun decodeString(): String = reader.readString()
```
rozza, line 170: Lets keep the `readOrThrow({ reader.readString() }, BsonType.STRING)` so that users know if an invalid type is present for a string field.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
```diff
@@ -176,7 +181,7 @@ internal open class DefaultBsonDecoder(
         return reader.state != AbstractBsonReader.State.END_OF_DOCUMENT && reader.currentBsonType != BsonType.NULL
     }
 
-    override fun decodeObjectId(): ObjectId = readOrThrow({ reader.readObjectId() }, BsonType.OBJECT_ID)
+    override fun decodeObjectId(): ObjectId = reader.readObjectId()
```
rozza, line 184: Lets keep the `readOrThrow({ reader.readObjectId() }, BsonType.OBJECT_ID)` so that users know if an invalid type is present for the objectId field.

rozza (changes requested):
Looking good - just need to check the bson types and error accordingly.

bson/src/main/org/bson/internal/StringCodecHelper.java
vbabanin: The method readString internally calls `verifyBSONType`, which already includes a type check and throws a `BsonInvalidOperationException` if the types don't match: [link to code](https://github.com/mongodb/mongo-java-driver/blob/bf03382f6b4c3b4e5f2c796564a6d1901f600a7b/bson/src/main/org/bson/AbstractBsonReader.java#L688). The current exception message is: '`X can only be called when CurrentBSONType is Y, not when CurrentBSONType is Z.`' 

Do we want to change this exception message with additional check?

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
```diff
@@ -176,7 +181,7 @@ internal open class DefaultBsonDecoder(
         return reader.state != AbstractBsonReader.State.END_OF_DOCUMENT && reader.currentBsonType != BsonType.NULL
     }
 
-    override fun decodeObjectId(): ObjectId = readOrThrow({ reader.readObjectId() }, BsonType.OBJECT_ID)
+    override fun decodeObjectId(): ObjectId = readOrThrow({ reader.readObjectId() }, BsonType.STRING)
```
rozza, line 184:
```suggestion
    override fun decodeObjectId(): ObjectId = readOrThrow({ reader.readObjectId() }, BsonType.OBJECT_ID)
```

bson/src/main/org/bson/internal/StringCodecHelper.java
rozza: Adding the check would make it inline with the NumberCodecHelper.

rozza (changes requested):
Just a couple of minors

bson/src/main/org/bson/internal/StringCodecHelper.java
vbabanin: Thank you for clarification! I have added a check for BsonType.STRING.

rozza (approved):
LGTM!

PR #1302, opened 2024-01-31
Make MongoWriteException Serializable

govindbalaji-s:
JAVA-5305

Implement Serializable for WriteError. This is the only non-serializable field in MongoWriteException. Exceptions are supposed to be serializable.

### DIFF
--- a/driver-core/src/main/com/mongodb/WriteError.java
+++ b/driver-core/src/main/com/mongodb/WriteError.java
@@ -18,6 +18,8 @@ package com.mongodb;
 
 import org.bson.BsonDocument;
 
+import java.io.Serializable;
+
 import static com.mongodb.assertions.Assertions.notNull;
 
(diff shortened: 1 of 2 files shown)


rozza:
Hi @govindbalaji-s,

Thanks for the PR. The ticket JAVA-5305 has been superseded by JAVA-2131 which is titled:

> Make all classes referenced by MongoException subclasses implement Serializable

This is currently on the backlog but not scheduled to be worked on at this time.  Please follow / watch that ticket for updates.

In the meantime I am going to close this ticket, while we wait for JAVA-2131 to be scheduled.

All the best,

Ross

govindbalaji-s:
Sorry, I think I ran into some exception classes that had `Object` type fields or something like that, and left it at that. I will follow the other ticket, but currently won't be able to pick.

PR #1305, opened 2024-02-02
JAVA-5311

jyemin:
This is generating a new evergreen validation warning, not sure why:

```
WARNING: strict unmarshalling YAML: load project error(s): error unmarshalling yaml strict: input fields may not match the command fields: yaml: unmarshal errors:
  line 226: cannot unmarshal !!seq into struct { Function string "yaml:\"func,omitempty\" bson:\"func,omitempty\""; Type string "yaml:\"type,omitempty\" bson:\"type,omitempty\""; DisplayName string "yaml:\"display_name,omitempty\" bson:\"display_name,omitempty\""; Command string "yaml:\"command,omitempty\" bson:\"command,omitempty\""; Variants []string "yaml:\"variants,omitempty\" bson:\"variants,omitempty\""; TimeoutSecs int "yaml:\"timeout_secs,omitempty\" bson:\"timeout_secs,omitempty\""; Params map[string]interface {} "yaml:\"params,omitempty\" bson:\"params,omitempty\""; ParamsYAML string "yaml:\"params_yaml,omitempty\" bson:\"params_yaml,omitempty\""; Vars map[string]string "yaml:\"vars,omitempty\" bson:\"vars,omitempty\""; RetryOnFailure bool "yaml:\"retry_on_failure,omitempty\" bson:\"retry_on_failure,omitempty\""; Loggers *model.LoggerConfig "yaml:\"loggers,omitempty\" bson:\"loggers,omitempty\"" }
```

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -227,24 +227,29 @@ functions:
       type: test
       params:
         working_dir: "src"
+        env:
+          AWS_ACCESS_KEY_ID: ${aws_access_key_id}
+          AWS_SECRET_ACCESS_KEY: ${aws_secret_access_key}
+          AWS_DEFAULT_REGION: us-east-1
+          AWS_ACCESS_KEY_ID: ${aws_access_key_id}
+          AZURE_TENANT_ID: ${azure_tenant_id}
+          AZURE_CLIENT_ID: ${azure_client_id}
+          AZURE_CLIENT_SECRET: ${azure_client_secret}
+          GCP_EMAIL: ${gcp_email}
+          GCP_PRIVATE_KEY: ${gcp_private_key}
+          AZUREKMS_KEY_VAULT_ENDPOINT: ${testazurekms_keyvaultendpoint}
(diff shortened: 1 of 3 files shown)


dylrich (approved):
LGTM!

rozza (changes requested):
Unfortunately its still leaks in the logs.

Search the logs for gcpEmail and you can see the property in plain text in the logs.  Look for "_Starting process 'Gradle Test Executor_"

What I don't understand is that some are marked as redacted. eg: <REDACTED:variable_name>. I can't determine where / how that is done - perhaps by evergreen.

Choices seem to be:
* ensure all variables are redacted via the same mechanism.
* Turn of info logging for gradle.

jyemin:
> Ensure all variables are redacted via the same mechanism.

You're right that Evergreen is doing this for us.  The point of this PR is not to rely on that redaction.

> Turn of info logging for gradle.

Not ideal as it's often needed for debugging


How about instead we use environment variables, as those shouldn't be logged (see publish.sh for the signing keys)

rozza:
Using env's sounds like a plan

PR #1313, opened 2024-02-21
Allow configuring the monitoring protocol to use; use the polling protocol in a FaaS environment by default

stIncMale:
Reviewers:

- @vbabanin picked up automatically via @mongodb/dbx-java.
- @jyemin picked up manually as @mongodb/dbx-java does not work twice.

The initial 6 commits are organized in such a way that a reviewer may review them one by one. They don't step on each other's toes, and represent separate work items.

JAVA-4936


driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -79,11 +79,13 @@ class DefaultServerMonitor implements ServerMonitor {
...
-    private final Thread monitorThread;
-    private final RoundTripTimeRunnable roundTripTimeMonitor;
+    private final ServerMonitor monitor;
+    /**
+     * Must be guarded by {@link #lock}.
+     */
+    @Nullable
+    private RoundTripTimeMonitor roundTripTimeMonitor;
     private final ExponentiallyWeightedMovingAverage averageRoundTripTime = new ExponentiallyWeightedMovingAverage(0.2);
-    private final Thread roundTripTimeMonitorThread;
```
stIncMale, line 84: The changes required both `roundTripTimeMonitor` & `roundTripTimeMonitorThread` to become `null`-able. I simplified that by replacing these two fields with a single field. However, this change lead to an inconsistent code having a single field `roundTripTimeMonitor` and two coupled fields `monitor` & `monitorThread`. To avoid such inconsistency, I also replaced the latter with a single field.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedServerDiscoveryAndMonitoringTest.java
```diff
@@ -38,4 +42,17 @@ public UnifiedServerDiscoveryAndMonitoringTest(@SuppressWarnings("unused") final
...
+        skipTests(getDefinition());
+    }
+
+    public static void skipTests(final BsonDocument definition) {
+        String description = definition.getString("description", new BsonString("")).getValue();
+        assumeFalse("Skipping because our server monitoring events behave differently for now",
+                description.equals("connect with serverMonitoringMode=auto >=4.4"));
+        assumeFalse("Skipping because our server monitoring events behave differently for now",
+                description.equals("connect with serverMonitoringMode=stream >=4.4"));
+    }
```
stIncMale, line 57: This will be undone by JAVA-5230 as per https://jira.mongodb.org/browse/JAVA-5230?focusedCommentId=6118489&focusedId=6118489&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6118489.

driver-core/src/test/resources/unified-test-format/server-discovery-and-monitoring/serverMonitoringMode-java-specific.json
stIncMale, line 1: This will be undone by JAVA-5230 as per https://jira.mongodb.org/browse/JAVA-5230?focusedCommentId=6118489&focusedId=6118489&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6118489.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -165,8 +192,7 @@ public void run() {
                     logStateChange(previousServerDescription, currentServerDescription);
                     sdamProvider.get().update(currentServerDescription);
 
-                    if (((connection == null || shouldStreamResponses(currentServerDescription))
```
vbabanin, line 168: is this check "connection == null" not needed anymore?

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -90,28 +94,36 @@ class DefaultServerMonitor implements ServerMonitor {
...
 
     @Override
     public void start() {
-        monitorThread.start();
-        roundTripTimeMonitorThread.start();
+        monitor.start();
+    }
+
+    private void ensureRoundTripTimeMonitorStarted() {
+        lock.lock();
```
vbabanin, line 118: Could you help me understand the reason for using a lock here?

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -151,6 +174,10 @@ public void run() {
                 while (!isClosed) {
                     ServerDescription previousServerDescription = currentServerDescription;
                     currentServerDescription = lookupServerDescription(currentServerDescription);
+                    boolean shouldStreamResponses = shouldStreamResponses(currentServerDescription);
+                    if (shouldStreamResponses) {
+                        ensureRoundTripTimeMonitorStarted();
+                    }
```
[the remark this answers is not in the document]
jyemin: Update: we discussed this offline and determined that

1. The spec doesn't require it
2. It's unlikely to happen often in practice
3. If it does happen it's not too harmful

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
jyemin: I thought I had proven to myself that the check was redundant by transforming to

```
                    boolean shouldStream = currentServerDescription.getTopologyVersion() != null;
                    if (((connection == null || shouldStream)
                            && shouldStream && currentServerDescription.getType() != UNKNOWN)
                            || (connection != null && connection.hasMoreToCome())
                            || (currentServerDescription.getException() instanceof MongoSocketException
                            && previousServerDescription.getType() != UNKNOWN)) {
                        continue;
                    }
```

and then allowing Intellij to simplify the expression, but when I tried it again IntelliJ didn't offer the simplification.

This is a truly horrible conditional, and has been the source of at least one bug that I remember.

jyemin (changes requested):

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
stIncMale: > is this check "connection == null" not needed anymore?

It had no effect, so I removed it. The following demonstrates why the `connection == null` expression had no effect.

Here is the original conditional, but reformatted to simplify perception:

```java
(
    (connection == null || shouldStreamResponses(currentServerDescription))
    && currentServerDescription.getTopologyVersion() != null
    && currentServerDescription.getType() != UNKNOWN
)
||
(connection != null && connection.hasMoreToCome())
||
(
    currentServerDescription.getException() instanceof MongoSocketException
    && previousServerDescription.getType() != UNKNOWN
)
```

It consists of three Boolean expressions ORed (`||`) together. Only the first one of those is modified in the PR:

1. Let's alias the `connection == null` expression as `x`.
2. Note that the expressions `shouldStreamResponses(currentServerDescription)` and `currentServerDescription.getTopologyVersion() != null` are equivalent. Let's alias them as `s`.

Now the first ORed Boolean expression can be written as

```java
(
    (x || s)
    && s
    && currentServerDescription.getType() != UNKNOWN
)
```

If `s` is `false`, this expression evaluates to `false`. If `s` is `true`, this expression evaluates to the same value as `currentServerDescription.getType() != UNKNOWN`. As you can see, `x` affects the outcome is neither case, and the expression can be simplified to

```java
(
    s
    && currentServerDescription.getType() != UNKNOWN
)
```

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
stIncMale: 1. `DefaultServerMonitor.ensureRoundTripTimeMonitorStarted` is executed by the `ServerMonitor` thread, and may write to `roundTripTimeMonitor`.
2. `DefaultServerMonitor.close` is executed by an application thread, let's call it `A`.
3. `DefaultServerMonitor.close` implicitly checks if `roundTripTimeMonitor` is not `null`, and if it's not, calls `roundTripTimeMonitor.close`. This check-then-act logic is correct only if neither of the following can happen concurrently with this logic being executed:
    - `roundTripTimeMonitor` is being written to
    - `roundTripTimeMonitor` is being started
4. To guarantee that the aforementioned assumption holds, threads `ServerMonitor` and `A` must coordinate what they do<sup>1</sup>. For example, having them acquire the same exclusive lock guarantees that `ServerMonitor` cannot write to or start `roundTripTimeMonitor` concurrently with `A` executing its check-then-act logic.

Granted, the `roundTripTimeMonitor == null` check could have been moved out of the critical section in `ensureRoundTripTimeMonitorStarted` without introducing any correctness issues. However, doing so still has negative effects without yielding anything useful (other than flexing/posing?):

- It requires writing a lengthy comment explaining why `roundTripTimeMonitor == null` does not have to be executed as part of the critical section.
- It makes the code more fragile: as soon as we have another place that may write to `roundTripTimeMonitor` concurrently with `DefaultServerMonitor.ensureRoundTripTimeMonitorStarted` being executed (I don't see why we would do that, but that's besides the point), it's no longer correct to execute `roundTripTimeMonitor == null` outside of the critical section.

---

<sup>1</sup> Note how I only talk about the coordination here, and not the memory synchronization that affects observability of writes. This is because the latter could have been achieved without locking, and, therefore, is not relevant to the explanation.

stIncMale: It's so good you asked about this code, because I have realized that `ensureRoundTripTimeMonitorStarted` must also check `isClosed`, as otherwise we still have a race condition between `ServerMonitor` starting `roundTripTimeMonitor` and `A` shutting it down. I'll fix this. This also does not invalidate anything I said in the previous comment.

stIncMale: I also believe that the coordination described above can be done in a non-blocking way, and I remember doing so somewhere in our codebase previously. But I neither remember where it was, nor thing that would be the optimal approach here (I don't know whether the approach was optimal there).

driver-core/src/main/com/mongodb/internal/connection/DefaultClusterableServerFactory.java
```diff
@@ -62,7 +63,7 @@ public DefaultClusterableServerFactory(
             final LoggerSettings loggerSettings,
             @Nullable final CommandListener commandListener,
             @Nullable final String applicationName, @Nullable final MongoDriverInformation mongoDriverInformation,
-            final List<MongoCompressor> compressorList, @Nullable final ServerApi serverApi) {
+            final List<MongoCompressor> compressorList, @Nullable final ServerApi serverApi, final boolean faas) {
```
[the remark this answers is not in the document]
stIncMale: Done.

driver-core/src/main/com/mongodb/event/ServerHeartbeatStartedEvent.java
```diff
@@ -46,12 +62,23 @@ public ConnectionId getConnectionId() {
         return connectionId;
     }
 
+    /**
+     * Gets whether the heartbeat is for an awaitable `hello` / legacy hello.
+     *
+     * @return {@code true} if and only if the heartbeat is for an awaitable `hello` / legacy hello.
```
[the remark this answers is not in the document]
stIncMale: Not that I know of, unless the specifications count. I still added more docs and described the essence of the the streaming protocol (hopefully, correctly). Please take a look.

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -665,6 +676,9 @@ private void translateOptions(final Map<String, List<String>> optionsMap) {
                 case "heartbeatfrequencyms":
                     heartbeatFrequency = parseInteger(value, "heartbeatfrequencyms");
                     break;
+                case "servermonitoringmode":
+                    serverMonitoringMode = ServerMonitoringMode.fromString(value);
```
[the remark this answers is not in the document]
stIncMale: I haven't think about that on my own.

Done.

driver-core/src/main/com/mongodb/connection/ServerMonitoringMode.java
```diff
@@ -0,0 +1,74 @@
...
+import static java.lang.String.format;
+
+/**
+ * The server monitoring mode, which defines the monitoring protocol to use.
+ *
+ * @since 5.1
+ */
+public enum ServerMonitoringMode {
+    /**
+     * Use the streaming protocol whe the server supports it or fall back to the polling protocol otherwise.
```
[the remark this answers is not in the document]
stIncMale: @vbabanin [suggested](https://github.com/mongodb/mongo-java-driver/pull/1313#discussion_r1503604478) the same. Done.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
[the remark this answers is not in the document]
stIncMale: For the sake of simplicity, I propose not to close it in the rare case when we stop using the streaming protocol.

driver-core/src/main/com/mongodb/connection/ServerMonitoringMode.java
```diff
@@ -15,60 +15,38 @@
...
  */
 public enum ServerMonitoringMode {
     /**
-     * Use the streaming protocol whe the server supports it or fall back to the polling protocol otherwise.
+     * Use the streaming protocol when the server supports it or fall back to the polling protocol otherwise.
+     * When the streaming protocol comes into play,
+     * {@link ServerHeartbeatStartedEvent#isAwaited()}, {@link ServerHeartbeatSucceededEvent#isAwaited()},
+     * {@link ServerHeartbeatFailedEvent#isAwaited()} return {@code true} for new events.
+     * <p>
+     * The streaming protocol uses long polling for SDAM, and is intended to reduce the delay between a server change
```
jyemin, line 38:
```suggestion
     * The streaming protocol uses long polling for server monitoring, and is intended to reduce the delay between a server change
```

driver-core/src/main/com/mongodb/internal/connection/ServerMonitoringModeUtil.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mongodb.internal.connection;
+
+import com.mongodb.connection.ServerMonitoringMode;
+
+import static java.lang.String.format;
+
+public final class ServerMonitoringModeUtil {
```
jyemin, line 22:
```suggestion
/**
 * <p>This class is not part of the public API and may be removed or changed at any time</p>
 */
public final class ServerMonitoringModeUtil {
```

jyemin (changes requested):

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
vbabanin: Thanks, @stIncMale, for the detailed explanation. Great that the race condition is caught.

[Optional] We could minimize acquiring the lock when the monitor has already started or closed by checking isClosed outside the synchronized block:

private void ensureRoundTripTimeMonitorStarted() {
    if (!isClosed) {
        withLock(lock, () -> {
            if (!isClosed && roundTripTimeMonitor == null) {
                roundTripTimeMonitor = new RoundTripTimeMonitor();
                roundTripTimeMonitor.start();
            }
        });
    }
}

stIncMale: > We could minimize acquiring the lock when the monitor has already started ...

That requires moving `roundTripTimeMonitor == null` our of the block, which has the negative consequences described above, which I want to avoid. And I don't want to do the double check for `roundTripTimeMonitor`, because despite eliminating those negative consequences, int introduces another one: the second check from within the critical section would not be there for correctness (it's not needed), but to eliminate the aforementioned negative consequences. This means, the code will again need a lengthy comment describing why an totally unnecessary check is done the second time inside the critical section.

If we could gain a perceivable performance improvement from such a change, I would have probably done it despite those negative consequences, but it's clear even without measuring :trollface: that there will be no gains out of it.

> ... or closed by checking isClosed outside the synchronized block

Unless I am missing something, double checking `isClosed` saves us at best a single invocation of `lock.lock()` per instance of `DefaultServerMonitor` 😃

stIncMale: I feel like doing the dumbest and simplest implementation here (in this particular case it does not seem to have any real negative consequences), instead of being a smart ass (there were situations when I wanted the latter).

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
vbabanin: Thanks for the clarification, @stIncMale. It is indeed redundant check.

i am curious about the original purpose of this check, given it wasn't functional. My assumption was that it aimed to bypass `waitForNext()` on lookup failure as connection is nullified in lookup method. However, the presence of ` (currentServerDescription.getException() instanceof MongoSocketException
                            && previousServerDescription.getType() != UNKNOWN)`seems to cover scenario aligning with the spec, which mandates immediate retry on network errors when the server was previously in a known state, without unnecessary delays:

#. If this was a network error and the server was in a known state before the error, the client MUST NOT sleep and MUST begin the next check immediately.

Resolving the thread.

vbabanin (approved):
LGTM!

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
vbabanin: Thank you, the code looks cleaner now 👍

jyemin (approved):
LGTM!

PR #1317, opened 2024-02-26
JAVA-5342 Fix encoding nullable generics

cliffred:
Encoding a generic class object with a nullable type parameter, like `DataClassWithNullableGeneric`, will fail with `org.bson.BsonInvalidOperationException: writeString can only be called when State is VALUE, not when State is NAME` or `org.bson.BsonInvalidOperationException: writeNull can only be called when State is VALUE, not when State is NAME` depending on whether `boxed` is null.
```kotlin
@Serializable data class Box<T>(val boxed: T)
@Serializable data class DataClassWithNullableGeneric(val box: Box<String?>)
```

JAVA-5342


rozza:
Many thanks @cliffred this will be in the next release.

PR #1318, opened 2024-02-27
Adding Transaction and WithTransaction support

rozza:
JAVA-4066 JAVA-4067

### DIFF
--- a/driver-core/src/main/com/mongodb/ClientSessionOptions.java
+++ b/driver-core/src/main/com/mongodb/ClientSessionOptions.java
@@ -22,8 +22,11 @@ import com.mongodb.lang.Nullable;
 import com.mongodb.session.ClientSession;
 
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 
+import static com.mongodb.assertions.Assertions.isTrueArgument;
 import static com.mongodb.assertions.Assertions.notNull;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 /**
  * The options to apply to a {@code ClientSession}.
@@ -38,6 +41,7 @@ public final class ClientSessionOptions {
 
     private final Boolean causallyConsistent;
     private final Boolean snapshot;
+    private final Long defaultTimeoutMS;
     private final TransactionOptions defaultTransactionOptions;
 
     /**
@@ -77,6 +81,24 @@ public final class ClientSessionOptions {
         return defaultTransactionOptions;
     }
 
+    /**
+     * Gets the default time limit for the following operations executed on the session:
+     *
+     * <ul>
+     *   <li>{@code commitTransaction}</li>
+     *   <li>{@code abortTransaction}</li>
+     *   <li>{@code withTransaction}</li>
+     *   <li>{@code close}</li>
+     * </ul>
+     * @param timeUnit the time unit
+     * @return the default timeout
+     * @since CSOT
+     */
+    @Nullable
+    public Long getDefaultTimeout(final TimeUnit timeUnit) {
+        return defaultTimeoutMS == null ? null : timeUnit.convert(defaultTimeoutMS, MILLISECONDS);
+    }
+
     @Override
     public boolean equals(final Object o) {
         if (this == o) {
@@ -85,36 +107,24 @@ public final class ClientSessionOptions {
         if (o == null || getClass() != o.getClass()) {
             return false;
         }
-
-        ClientSessionOptions that = (ClientSessionOptions) o;
-
-        if (!Objects.equals(causallyConsistent, that.causallyConsistent)) {
-            return false;
-        }
-
-        if (!Objects.equals(snapshot, that.snapshot)) {
-            return false;
-        }
-        if (!Objects.equals(defaultTransactionOptions, that.defaultTransactionOptions)) {
-            return false;
-        }
-
-        return true;
+        final ClientSessionOptions that = (ClientSessionOptions) o;
+        return Objects.equals(causallyConsistent, that.causallyConsistent)
+                && Objects.equals(snapshot, that.snapshot)
+                && Objects.equals(defaultTimeoutMS, that.defaultTimeoutMS)
+                && Objects.equals(defaultTransactionOptions, that.defaultTransactionOptions);
     }
 
     @Override
     public int hashCode() {
-        int result = causallyConsistent != null ? causallyConsistent.hashCode() : 0;
-        result = 31 * result + (snapshot != null ? snapshot.hashCode() : 0);
-        result = 31 * result + (defaultTransactionOptions != null ? defaultTransactionOptions.hashCode() : 0);
-        return result;
+        return Objects.hash(causallyConsistent, snapshot, defaultTimeoutMS, defaultTransactionOptions);
     }
 
     @Override
     public String toString() {
         return "ClientSessionOptions{"
                 + "causallyConsistent=" + causallyConsistent
-                + "snapshot=" + snapshot
+                + ", snapshot=" + snapshot
+                + ", defaultTimeoutMS=" + defaultTimeoutMS
                 + ", defaultTransactionOptions=" + defaultTransactionOptions
                 + '}';
     }
@@ -141,6 +151,7 @@ public final class ClientSessionOptions {
         builder.causallyConsistent = options.isCausallyConsistent();
         builder.snapshot = options.isSnapshot();
         builder.defaultTransactionOptions = options.getDefaultTransactionOptions();
+        builder.defaultTimeoutMS = options.defaultTimeoutMS;
         return builder;
     }
 
@@ -151,6 +162,7 @@ public final class ClientSessionOptions {
     public static final class Builder {
         private Boolean causallyConsistent;
         private Boolean snapshot;
+        private Long defaultTimeoutMS;
         private TransactionOptions defaultTransactionOptions = TransactionOptions.builder().build();
 
         /**
@@ -197,6 +209,27 @@ public final class ClientSessionOptions {
         }
 
         /**
+         * Sets the default time limit for the following operations executed on the session:
+         *
+         * <ul>
+         *   <li>{@code commitTransaction}</li>
+         *   <li>{@code abortTransaction}</li>
+         *   <li>{@code withTransaction}</li>
+         *   <li>{@code close}</li>
+         * </ul>
+         * @param defaultTimeout the timeout
+         * @param timeUnit the time unit
+         * @return this
+         * @since CSOT
+         * @see #getDefaultTimeout
+         */
+        public Builder defaultTimeout(final long defaultTimeout, final TimeUnit timeUnit) {
+            isTrueArgument("timeoutMS must be >= 0", defaultTimeout >= 0);
+            this.defaultTimeoutMS = MILLISECONDS.convert(defaultTimeout, timeUnit);
+            return this;
+        }
+
+        /**
          * Build the session options instance.
          *
          * @return The {@code ClientSessionOptions}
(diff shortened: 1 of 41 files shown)


driver-core/src/main/com/mongodb/TransactionOptions.java
```diff
@@ -42,6 +42,7 @@ public final class TransactionOptions {
     private final WriteConcern writeConcern;
     private final ReadPreference readPreference;
     private final Long maxCommitTimeMS;
+    private final Long timeoutMS;
```
vbabanin, line 45: The specification seems ambiguous about the use of `timeoutMS`, particularly whether it's intended to apply solely to withTransaction. This [discussion in the specifications pull request](https://github.com/mongodb/specifications/pull/881/files#r543057367) suggests an initial intention for timeoutMS to be set/unset for withTransaction.

However, when timeoutMS is specified in `TransactionOptions`, it appears to impact the `startTransaction` method as well, not just `withTransaction`. Do we need to seek further clarification on this?

driver-core/src/main/com/mongodb/ClientSessionOptions.java
```diff
@@ -196,6 +208,27 @@ public Builder defaultTransactionOptions(final TransactionOptions defaultTransac
...
+         *   <li>{@code close}</li>
+         * </ul>
+         * @param defaultTimeout the timeout
+         * @param timeUnit the time unit
+         * @return this
+         * @since CSOT
+         * @see #getDefaultTimeout
+         */
+        public Builder defaultTimeout(final long defaultTimeout, final TimeUnit timeUnit) {
+            isTrueArgument("timeoutMS must be >= 0", defaultTimeout >= 0);
```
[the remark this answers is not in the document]
rozza: Added: JAVA-5351 to track

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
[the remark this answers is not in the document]
rozza: So according to [spec](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#sessions):

> The [SessionOptions](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#mongoclient-changes) used to construct explicit [ClientSession](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#clientsession) instances MUST accept a new defaultTimeoutMS option, which specifies the timeoutMS value for the following operations executed on the session:
>
> * commitTransaction
> * abortTransaction
> * withTransaction
> * endSession
>
> If this option is not specified for a ClientSession, it MUST inherit the timeoutMS of its parent MongoClient.


So with manual transactions each operation is a single operation so creates a new timeout and so there is nothing to reset.

With [convenient transactions](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#convenient-transactions-api) eg `withTransaction`: 

> If timeoutMS is set, drivers MUST apply it to the entire withTransaction call. To propagate the timeout to the user-supplied callback, drivers MUST store the timeout as a field on the ClientSession object.
> ...
> If the callback returns an error and the transaction must be aborted, drivers MUST refresh the timeoutMS value for the abortTransaction operation.

driver-core/src/main/com/mongodb/TransactionOptions.java
rozza: I think we follow the spec see: [TransactionOptions](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst#user-content-transactionoptions) which is described as:

> Options for {{ClientSession.startTransaction}}, as defined in the [Transactions](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.md) specification. The structure of these options MAY vary across drivers (e.g. dictionary, typed class).

rozza: Apologies, yes we do deviate from spec slightly but not really with the spirit of the [timeoutMS cannot be overridden for startSession calls]:(https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#timeoutms-cannot-be-overridden-for-startsession-calls) 

As we already share {{TransactionOptions}} with both APIs I think we can do this, without much harm. It does add a potential configuration knob but only if a user wishes to use it.

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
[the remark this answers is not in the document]
vbabanin: > So with manual transactions each operation is a single operation so creates a new timeout and so there is nothing to reset.

Thank you for explanation and apologies for confusion. By reset i meant, a new timeout which starts fresh. I've noticed that `abortTransaction `and `commitTransaction` seem to operate on the remaining timeout from the `startTransaction` operation , contrary to "each operation is a single operation so creates a new timeout".

It appears the logic in [TimeoutContext.java](https://github.com/rozza/mongo-java-driver/blob/7e0a34f64e115e90858e099c41530cc17f41877b/driver-core/src/main/com/mongodb/internal/TimeoutContext.java#L73) uses the timeout from the `ClientSession` initiated by `startTransaction`. Just wanted to clarify this behavior, as it seems the intention of the spec is to use new timeout for each:

- commitTransaction
- abortTransaction
- withTransaction
- endSession

Is this interpretation correct according to the spec?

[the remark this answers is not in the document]
rozza: Thanks good catch - will update and add some tests to cover this.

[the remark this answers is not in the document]
rozza: Done!

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -65,6 +67,30 @@ public static TimeoutContext createMaintenanceTimeoutContext(final TimeoutSettin
...
+        TimeoutContext sessionTimeoutContext = session.getTimeoutContext();
+
+        if (sessionTimeoutContext != null) {
+            TimeoutSettings sessionTimeoutSettings = sessionTimeoutContext.timeoutSettings;
+            if (timeoutSettings.getGenerationId() > sessionTimeoutSettings.getGenerationId()) {
+                throw new MongoClientException("Cannot change the timeoutMS during a transaction.");
+            }
+
+            // Check for any legacy operation timeouts
+            if (sessionTimeoutSettings.getTimeoutMS() == null) {
```
vbabanin, line 80: `maxAwaitTimeMS`, which can be set on an `Iterable`, isn't currently taken into account in this logic when user is using "withTransaction" or explicit client sessions, remaining at 0 regardless of user settings.

rozza:
Hoping to have fixed the tests - I over did the cleanup of the timeout context.

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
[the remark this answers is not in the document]
rozza: Done - added some extra tests.

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
```diff
@@ -220,7 +224,7 @@ public <T> T withTransaction(final TransactionBody<T> transactionBody, final Tra
                 if (transactionState == TransactionState.IN) {
                     abortTransaction();
```
vbabanin, line 225: It seems when one of the operations holds a lock, executing another operation on that lock in `withTransaction` might lead to a timeout breach.

For example: the insert command fails with the following error:
`Command failed with error 24 (LockTimeout): 'Unable to acquire IX lock on '{6504462206465692013: Database, 1892776188038304109}' within 5ms. opId: 11901851, op: conn653927, connId: 653927.' on server localhost:27017. The full response is {"errorLabels": ["TransientTransactionError"], "ok": 0.0, "errmsg": "Unable to acquire IX lock on '{6504462206465692013: Database, 1892776188038304109}' within 5ms. ``

The transaction does not seem to have started due to the lock being held. When this `MongoCommandException `occurs and the transaction is attempted to be aborted, a fresh `TimeoutContext` is created with a new timeout. Since the error is classified as a "TransientTransactionError," the loop continues to operate, executing `startTransaction` with a new fresh timeout one more time.

Because the timeout for `withTransaction` is created fresh for `startTransaction`, it leads to an infinite loop until the lock is released, which breaches the timeout expectation for `withTransaction` as specified in the spec: "If timeoutMS is set, drivers MUST apply it to the entire withTransaction call."

I was able to reproduce it in `test10CustomTestWithTransactionUsesASingleTimeout`.

vbabanin (commented):
Looks good overall! I've just noticed a deviation from the intended behavior, which might need addressing.

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
rozza: Fantastic - I totally overlooked that scenario. Have added a new regression test 👍

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
```diff
@@ -105,6 +105,45 @@ public void startTransaction() {
 
     @Override
     public void startTransaction(final TransactionOptions transactionOptions) {
+        startTransaction(transactionOptions, createTimeoutContext(transactionOptions));
+    }
+
+    @Override
+    public void commitTransaction() {
```
rozza, line 112: I the private methods to below the public overrides.

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
```diff
@@ -105,6 +105,45 @@ public void startTransaction() {
 
     @Override
     public void startTransaction(final TransactionOptions transactionOptions) {
+        startTransaction(transactionOptions, createTimeoutContext(transactionOptions));
```
rozza, line 108: Had to add an overload for startTransaction that takes a timeout context.

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
```diff
@@ -214,11 +218,13 @@ public <T> T withTransaction(final TransactionBody<T> transactionBody) {
...
         notNull("transactionBody", transactionBody);
         long startTime = ClientSessionClock.INSTANCE.now();
+        TimeoutContext withTransactionTimeoutContext = createTimeoutContext(options);
+
         outer:
         while (true) {
             T retVal;
             try {
-                startTransaction(options);
+                startTransaction(options, copyTimeoutContext(withTransactionTimeoutContext));
```
rozza, line 227: This ensures that on failure (which issues an abort transaction) that the original `withTransactionTimeoutContext` isn't reset.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
```diff
@@ -543,9 +543,9 @@ public void test9EndSessionCustomTesEachOperationHasItsOwnTimeoutWithCommit() {
                 + "  configureFailPoint: \"failCommand\","
                 + "  mode: { times: 1 },"
                 + "  data: {"
-                + "    failCommands: [\"insertOne\"],"
+                + "    failCommands: [\"insert\"],"
```
rozza, line 546: Fixed tests incorrectly using `insertOne` not `insert`.

vbabanin (approved):
LGTM!

PR #1319, opened 2024-02-27
Fix test failures on 8.0 sharded clusters

jyemin:
Advance cluster time on all session entities in unified tests to one after the initial data is created

JAVA-5334

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
@@ -214,7 +214,7 @@ public abstract class UnifiedTest {
         if (definition.containsKey("skipReason")) {
             throw new AssumptionViolatedException(definition.getString("skipReason").getValue());
         }
-        startingClusterTime = addInitialData();
+        startingClusterTime = addInitialDataAndGetClusterTime();
         entities.init(entitiesArray, startingClusterTime,
                 fileDescription != null && PRESTART_POOL_ASYNC_WORK_MANAGER_FILE_DESCRIPTIONS.contains(fileDescription),
                 this::createMongoClient,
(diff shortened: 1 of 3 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -890,7 +893,7 @@ private List<CommandEvent> lastTwoCommandEvents(final TestCommandListener listen
         return events.subList(events.size() - 2, events.size());
     }
 
-    private void addInitialData() {
+    private BsonDocument addInitialData() {
```
stIncMale, line 896: Let's rename the method

```suggestion
    private BsonDocument addInitialDataAndGetClusterTime() {
```

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
jyemin: That name makes me not want to return the cluster time from this method, and instead just get the cluster time separately.  But which do you prefer?

stIncMale: I am OK with returning the cluster time from `addInitialData` because it makes the dependency more explicit. If we separate `addInitialData` and `getCurrentClusterTime`, we may need a comment emphasizing the dependency, such that we don't accidentally mess up the order these methods are to be called in the future.

stIncMale (commented):
.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
jyemin: Done.

jyemin: (I didn't accept your edit because the caller needs to change too)

stIncMale (approved):

PR #1320, opened 2024-02-28
Move initial data creation back to where it was

jyemin:
JAVA-5334

When I merged the previous fix to JAVA-5334 into the JAVA-4171 PR, all the sharded variants started timing out.  I got as far as learning that, after that merge, for many of the new unified transaction tests (particular ones in `mongos-pin-auto`) the `drop` command would block for 90 seconds before completing successfully.  After ruling out all the other possibilities, I found that by moving the initial data creation below entity creation (where it was before), and then advancing the session cluster times after that, the drop command issues went away.

@ShaneHarvey wondering if you have any insight into what might be going on here.

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
@@ -71,6 +71,7 @@ import org.bson.BsonString;
 import org.bson.BsonValue;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
> the drop command would block for 90 seconds before completing successfully

This indicates the drop command is blocking waiting for an open transaction to time out (transactionLifetimeLimitSeconds). My theory is that when you moved the drop earlier it made it use a earlier clusterTime which ends up blocking on the transaction. Moving it later, the drop ends up using a later clusterTime which kills/overrides the transaction. To verify I would need to test.

jyemin:
Ah, so maybe this isn't needed, but for https://github.com/mongodb/mongo-java-driver/pull/1310/files#diff-a746e25fe11a43c11ef4cf1d65ce45fe37c6597e31f0772d4f628de5f7b36bc8R225 the call to `killAllSessions` just needs to move above the call to `addInitialData`

jyemin:
I tried it, it works locally.  Once I test successfully on the JAVA-4171 PR, I will close this PR.  

Thanks Shane.

jyemin:
Looks good on the other PR.  Closing this out.

PR #1321, opened 2024-02-28
Add Client Operation Timeout to Change Streams

vbabanin:
<h4> Changes made: </h4>

- Trigger timeout error when maxAwaitTimeMS exceeds a timeoutMS.
- Apply timeoutMS to initial aggregate and subsequent next calls, without appending maxTimeMS to getMore.
- Attempt to resume change stream on subsequent next call if previous next failed with timeout error.
- Use maxAwaitTimeMS as maxTimeMS in getMore commands.
- Preserve change stream validity upon timeout errors. 

<h4> NOTE </h4>

The [Reactive Streams specification](https://github.com/reactive-streams/reactive-streams-jvm#1.6) prevents us from allowing a subsequent next call (event in reactive terms) after a timeout error, conflicting with the CSOT spec requirement not to invalidate the change stream and to try resuming and establishing a new change stream on the server. We immediately let users know about a timeout error, which then closes the stream/publisher.

JAVA-4054

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/TimeoutContext.java
+++ b/driver-core/src/main/com/mongodb/internal/TimeoutContext.java
@@ -165,7 +165,7 @@ public class TimeoutContext {
     }
 
     public long getMaxAwaitTimeMS() {
-        return hasTimeoutMS() ? 0 : timeoutSettings.getMaxAwaitTimeMS();
+        return timeoutSettings.getMaxAwaitTimeMS();
     }
 
     public long getMaxTimeMS() {
--- a/driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
@@ -39,6 +39,21 @@ import static com.mongodb.assertions.Assertions.assertNotNull;
 import static com.mongodb.internal.operation.ChangeStreamBatchCursorHelper.isResumableError;
 import static com.mongodb.internal.operation.SyncOperationHelper.withReadConnectionSource;
 
+/**
+ * A change stream cursor that wraps {@link CommandBatchCursor} with automatic resumption capabilities in the event
+ * of timeouts or transient errors.
+ * <p>
+ * Upon encountering a resumable error during {@code hasNext()}, {@code next()}, or {@code tryNext()} calls, the {@link ChangeStreamBatchCursor}
+ * attempts to establish a new change stream on the server.
+ * </p>
+ * If an error occurring during any of these method calls is not resumable, it is immediately propagated to the caller, and the {@link ChangeStreamBatchCursor}
+ * is closed and invalidated on the server. Server errors that occur during this invalidation process are not propagated to the caller.
+ * <p>
+ * A {@link MongoOperationTimeoutException} does not invalidate the {@link ChangeStreamBatchCursor}, but is immediately propagated to the caller.
+ * Subsequent method call will attempt to resume operation by establishing a new change stream on the server, without doing {@code getMore}
+ * request first.
+ * </p>
+ */
 final class ChangeStreamBatchCursor<T> implements AggregateResponseBatchCursor<T> {
     private final ReadBinding binding;
     private final ChangeStreamOperation<T> changeStreamOperation;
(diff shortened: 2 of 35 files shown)


driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -165,7 +165,7 @@ public TimeoutSettings getTimeoutSettings() {
     }
 
     public long getMaxAwaitTimeMS() {
-        return hasTimeoutMS() ? 0 : timeoutSettings.getMaxAwaitTimeMS();
```
vbabanin, line 168: In accordance with the CSOT specifications for the tailable-await cursor:

> 
"If the maxAwaitTimeMS option is set, drivers MUST use it as the maxTimeMS field on getMore commands."

The logic to determine whether `maxAwaitTimeMS `should be utilized is located in the AggregateOperationImpl and FindOperation.

driver-core/src/main/com/mongodb/internal/operation/CommandBatchCursor.java
```diff
@@ -59,6 +59,9 @@
 class CommandBatchCursor<T> implements AggregateResponseBatchCursor<T> {
 
     private final MongoNamespace namespace;
+    /**
+     * maxAwaitTimeMS
+     */
```
rozza, line 64: Nit - no need for this javadoc

driver-core/src/main/com/mongodb/internal/operation/CommandBatchCursor.java
```diff
@@ -258,6 +261,9 @@ private CommandCursorResult<T> toCommandCursorResult(final ServerAddress serverA
         return commandCursorResult;
     }
 
+    void setCloseImmediately(final boolean closeImmediately) {
```
rozza, line 264: Would be good to explain what his is and why its needed? I dont fully understand the reasoning here.

driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
```diff
@@ -41,22 +43,31 @@ final class ChangeStreamBatchCursor<T> implements AggregateResponseBatchCursor<T
...
     private CommandBatchCursor<RawBsonDocument> wrapped;
     private BsonDocument resumeToken;
     private final AtomicBoolean closed;
 
+    /**
+     * This flag is used to manage change stream resumption logic after a timeout error.
+     * Indicates whether the last {@code next} call resulted in a {@link MongoOperationTimeoutException}. If {@code true},
+     * indicates a timeout occurred, prompting an attempt to resume the change stream on the next call.
+     */
+    private boolean lastOpTimeoutStatus;
```
rozza, line 56: The naming / use of this flag is confusing. Can you clarify in the code / docs that throwing a `MongoOperationTimeoutException` doesn't invalidate the cursor.

The reason for this is in 6 months we'll all have forgotten about this special and strange CSOT behavior.

rozza (changes requested):
A couple of questions, its looking good just mainly things to help the maintainability in the future.

Also lets document in the sync drivers that a `MongoOperationTimeoutException` is not terminal for a change stream cursor and calling `next()` again may produce results.

driver-core/src/main/com/mongodb/internal/operation/CommandBatchCursor.java
vbabanin: The current implementation is designed to automatically reset the timeoutMS for cursor closure, ensuring that the cursor has sufficient time to release resources. This behavior particularly impacts the `ChangeStreamCursor` which enforces a unified timeout duration for the entire 'next' operation, including resume attempts and the internal process of closing the cursor.

The method modifies this behavior with respect to timeouts. When enabled, it prevents the automatic reset of the cursor's timeout for closing.

I realized that the method name is not reflecting its purpose. Therefore, I have updated the name to `setCloseWithoutTimeoutReset` and added the JavaDoc documentation for clarity.

driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
```diff
@@ -192,14 +217,34 @@ <R> R resumeableOperation(final Function<AggregateResponseBatchCursor<RawBsonDoc
                     throw MongoException.fromThrowableNonNull(t);
                 }
             }
-            wrapped.close();
+            resumeChangeStream();
```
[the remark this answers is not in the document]
vbabanin: I've simplified the code to avoid using `resumeAfterTimeout`. If you have specific suggestions, please let me know.

driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
vbabanin: Acknowledged. I have renamed the field and added additional class level javaDoc. Please let me know if there is anything to add.

rozza (changes requested):
Looking really good.

Need to add to the public `ChangeStreamIterable` Javadoc so its user discoverable.

```
 * <p>
 * A {@link MongoOperationTimeoutException} does not invalidate the {@link ChangeStreamBatchCursor}, but is immediately propagated to the caller.
 * Subsequent method call will attempt to resume operation by establishing a new change stream on the server, without doing {@code getMore}
 * request first.
 * </p>
```

rozza:
I noticed that the test failures are unrelated but we should keep an eye on them as there maybe more racy tests to review / fix in the future.

PR #1324, opened 2024-03-04
Support discriminators not being the first field when decoding

rozza:
JAVA-5304

### DIFF
--- a/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
+++ b/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
@@ -69,15 +69,17 @@ internal open class DefaultBsonDecoder(
         val validKeyKinds = setOf(PrimitiveKind.STRING, PrimitiveKind.CHAR, SerialKind.ENUM)
         val bsonValueCodec = BsonValueCodec()
         const val UNKNOWN_INDEX = -10
-        fun validateCurrentBsonType(reader: AbstractBsonReader,
-                                    expectedType: BsonType,
-                                    descriptor: SerialDescriptor,
-                                    actualType: (descriptor: SerialDescriptor) -> String = { it.kind.toString() }) {
+        fun validateCurrentBsonType(
+            reader: AbstractBsonReader,
+            expectedType: BsonType,
+            descriptor: SerialDescriptor,
+            actualType: (descriptor: SerialDescriptor) -> String = { it.kind.toString() }
+        ) {
             reader.currentBsonType?.let {
                 if (it != expectedType) {
                     throw SerializationException(
-                        "Invalid data for `${actualType(descriptor)}` expected a bson ${expectedType.name.lowercase()} found: " +
-                                "${reader.currentBsonType}")
+                        "Invalid data for `${actualType(descriptor)}` expected a bson " +
+                            "${expectedType.name.lowercase()} found: ${reader.currentBsonType}")
                 }
             }
         }
--- a/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
+++ b/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
@@ -120,28 +120,14 @@ internal open class DefaultBsonDecoder(
 
     @Suppress("ReturnCount")
     override fun beginStructure(descriptor: SerialDescriptor): CompositeDecoder {
-        when (descriptor.kind) {
-            is StructureKind.LIST -> {
-                reader.readStartArray()
-                return BsonArrayDecoder(reader, serializersModule, configuration)
-            }
-            is PolymorphicKind -> {
-                return PolymorphicDecoder(reader, serializersModule, configuration)
-            }
+        return when (descriptor.kind) {
+            is StructureKind.LIST -> BsonArrayDecoder(descriptor, reader, serializersModule, configuration)
+            is PolymorphicKind -> PolymorphicDecoder(descriptor, reader, serializersModule, configuration)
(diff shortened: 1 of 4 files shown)


bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecProviderTest.kt
```diff
@@ -75,6 +79,41 @@ class KotlinSerializerCodecProviderTest {
         assertEquals(DataClassWithSimpleValues::class.java, codec.encoderClass)
     }
 
+    @Test
+    fun testDataClassWithSimpleValuesFieldOrdering() {
```
rozza, line 83: This functionality was already supported but added a regression test.

bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecProviderTest.kt
```diff
@@ -75,6 +79,41 @@ class KotlinSerializerCodecProviderTest {
...
+    fun testDataClassSealedFieldOrdering() {
+        val codec = MongoClientSettings.getDefaultCodecRegistry().get(SealedInterface::class.java)
+
+        val objectId = ObjectId("111111111111111111111111")
+        val oid = "\$oid"
+        val expected = DataClassSealedInterface(objectId, "string")
+        val actual =
+            codec.decode(
+                JsonReader(
+                    """{"name": "string", "_id": {$oid: "${objectId.toHexString()}"},
```
rozza, line 109: Here all the fields are out of order and this confirms it does support the scenario where the discriminator isn't the first field.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
```diff
@@ -213,13 +213,37 @@ private class PolymorphicDecoder(
     configuration: BsonConfiguration
 ) : DefaultBsonDecoder(reader, serializersModule, configuration) {
     private var index = 0
+    private var mark: BsonReaderMark? = reader.mark
```
rozza, line 216: Here a mark is set allowing the `PolymorphicDecoder` to iterate to the discriminator field instead of expecting it to be the first field.

robbamberg:
@rozza this seems a solution for the polymorphic serializer on root level (with _id field), but this is only part of the fix. Currently the polymorphic serializer is not called while inserting a document to Mongo, so without that also having a fix, this won't do much.

In driver-core in the Operations class we see that during insert of a document that the function documentToBsonDocument is called. It fetches the Codec for the class that is inserted (data class), but it should fetch the Codec of collection (basically the getCodec function). This is causing the wrong serializer to be used.

rozza:
Hi @robbamberg - thanks for the follow up, that is a slightly different issue to the one being fixed here. I've added #1327 for that fix.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
```diff
@@ -213,13 +213,37 @@ private class PolymorphicDecoder(
...
+            0 -> {
+                reader.readStartDocument()
+                reader.readBsonType()
+                while (reader.state != AbstractBsonReader.State.END_OF_DOCUMENT) {
+                    if (reader.readName() == configuration.classDiscriminator) {
+                        found = true
+                        break
+                    }
+                    reader.skipValue()
+                    reader.readBsonType()
```
jyemin, line 238: This can be removed, as `readName` also reads the type for you.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
```diff
@@ -125,7 +126,6 @@ internal open class DefaultBsonDecoder(
                 return BsonArrayDecoder(reader, serializersModule, configuration)
             }
             is PolymorphicKind -> {
-                reader.readStartDocument()
```
jyemin, line 128: Can you explain the apparent lack of symmetry of the various branches in this method?  It's not clear to me why in some branches we have to call `readStartArray` or `readStartDocument`, but in others we do not.

jyemin (changes requested):

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
rozza: I've cleaned these up - now all the Decoders are responsible for using the reader and validating its state.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
rozza: `reader.state` (aka `reader.getState()`) validates the type. However, I have simplified the check.

jyemin (commented):
Looking good, just one DRY-related suggestion.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
```diff
@@ -260,6 +289,17 @@ private class BsonDocumentDecoder(
     private var index = 0
     private var isKey = false
 
+    init {
+        reader.currentBsonType?.let {
```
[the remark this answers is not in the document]
rozza: Not sure its easier to read but have done so.

jyemin (approved):

PR #1325, opened 2024-03-04
Fix timing-related test failure in SessionsProseTest

jyemin:
There are two tests in AbstractSessionsProseTest that rely on a running mongocryptd process.  Each test starts and destroys the process, and it appears that occasionally the destruction in the first test interferes with the starting in the second test, and the second test fails to connect to the mongocryptd process.

This commit attempts to work around this by starting a single mongocryptd process and use it for both tests. Note that AbstractSessionsProseTest has two subclasses and therefore the mongocryptd process is still actually started and destroyed twice, it doesn't seem to be a problem in practice because the two subclasses are in different modules and there is enough time between their execution that there is interference.

JAVA-4998

Note: I ran a patch build of the 40 4.2+ standalone tasks (because they run the fastest) four times, and observed no failures of the test in question in any of them.

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/AbstractSessionsProseTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/AbstractSessionsProseTest.java
@@ -27,6 +27,8 @@ import com.mongodb.event.CommandListener;
 import com.mongodb.event.CommandStartedEvent;
 import org.bson.BsonDocument;
 import org.bson.Document;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


driver-sync/src/test/functional/com/mongodb/client/AbstractSessionsProseTest.java
```diff
@@ -55,9 +57,25 @@
...
+    public static void beforeAll() throws IOException {
+        if (serverVersionAtLeast(4, 2)) {
+            mongocryptdProcess = startMongocryptdProcess();
+        }
+    }
+
+    @AfterAll
+    public static void afterAll() {
+        if (mongocryptdProcess != null) {
+            mongocryptdProcess.destroy();
```
vbabanin, line 74: Given the empirical analysis shows no test interference across four runs, I still suggest waiting for the process to exit with `mongocryptdProcess.waitFor(TIMEOUT_DURATION.toMillis(), TimeUnit.MILLISECONDS); ` for extra certainty.

vbabanin (approved):

driver-sync/src/test/functional/com/mongodb/client/AbstractSessionsProseTest.java
jyemin: I'm going to hold off on this unless it proves necessary.

PR #1242, opened 2023-11-02
Test on Java 21

jyemin:
* Add JDK 21 to Gradle configuration
* Add JDK 21 to the test matrix
* Replace JDK 17 with JDK 21 for tasks that are ony run using latest JDK
* Update Mockito dependency

PR #1326, opened 2024-03-04
Test on Java 21

stIncMale:
This PR builds up on https://github.com/mongodb/mongo-java-driver/pull/1242.

* Add JDK 21 to Gradle configuration
* Add JDK 21 to the test matrix
* Replace JDK 17 with JDK 21 for tasks that are ony run using latest JDK
* Update dependencies (some now have to be conditional)
* Update `run-scala-tests.sh` to use `JAVA_VERSION`
* Run the `scala-tests` task with JDK 8, 17, 21 similarly to `kotlin-tests`
* Fix test code that assumed Java SE 9+

JAVA-5164

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -1789,6 +1789,10 @@ axes:
   - id: jdk
     display_name: JDK
     values:
+      - id: "jdk21"
+        display_name: JDK21
+        variables:
+          JAVA_VERSION: "21"
       - id: "jdk17"
         display_name: JDK17
         variables:
@@ -2037,7 +2041,8 @@ buildvariants:
     - name: "test"
 
 - matrix_name: "tests-jdk-secure"
-  matrix_spec: { auth: "auth", ssl: "ssl", jdk: [ "jdk8", "jdk17" ], version: [ "3.6", "4.0", "4.2", "4.4", "5.0", "6.0", "7.0", "latest" ],
+  matrix_spec: { auth: "auth", ssl: "ssl", jdk: [ "jdk8", "jdk17", "jdk21"],
+                 version: [ "3.6", "4.0", "4.2", "4.4", "5.0", "6.0", "7.0", "latest" ],
                  topology: "*", os: "linux" }
   display_name: "${version} ${topology} ${auth} ${ssl} ${jdk} ${os} "
   tags: ["tests-variant"]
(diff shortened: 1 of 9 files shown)


driver-core/src/test/functional/com/mongodb/internal/connection/SocketStreamHelperSpecification.groovy
```diff
@@ -55,10 +55,18 @@ class SocketStreamHelperSpecification extends Specification {
...
-            getOptionMethod.invoke(socket, ExtendedSocketOptions.getDeclaredField('TCP_KEEPCOUNT').get(null)) == 9
-            getOptionMethod.invoke(socket, ExtendedSocketOptions.getDeclaredField('TCP_KEEPIDLE').get(null)) == 120
-            getOptionMethod.invoke(socket, ExtendedSocketOptions.getDeclaredField('TCP_KEEPINTERVAL').get(null)) == 10
+            Method getOptionMethod
+            try {
+                getOptionMethod = Socket.getMethod('getOption', SocketOption)
+            } catch (NoSuchMethodException e) {
+                // ignore, the `Socket.getOption` method was added in Java SE 9 and does not exist in Java SE 8
+                getOptionMethod = null
+            }
```
stIncMale, line 64: This code assumed that if `ExtendedSocketOptions.TCP_KEEPCOUNT` is available, then so is `Socket.getOption`. That's not actually the case:

- `ExtendedSocketOptions.TCP_KEEPCOUNT` (not part of the Java SE) was added in Oracle JDK 8: https://docs.oracle.com/javase/8/docs/jre/api/net/socketoptions/spec/jdk/net/ExtendedSocketOptions.html#TCP_KEEPCOUNT
- `Socket.getOption` was added in the Java SE 9.
- So when I run `sdk use java 17.0.10-librca && ./gradlew -PjavaVersion=8 clean :driver-core:test` locally, `ExtendedSocketOptions.TCP_KEEPCOUNT` is present, while `Socket.getOption` is not, which leads to a failure.
- Why is this not a problem in Evergreen? My guess is that the JDK 8 it uses does not have `ExtendedSocketOptions.TCP_KEEPCOUNT`.

driver-scala/src/integration/scala/org/mongodb/scala/TestMongoClientHelper.scala
```diff
@@ -29,7 +29,7 @@ object TestMongoClientHelper {
 
   val mongoClientURI: String = {
     val uri = Properties.propOrElse(MONGODB_URI_SYSTEM_PROPERTY_NAME, DEFAULT_URI)
-    if (!uri.isBlank) uri else DEFAULT_URI
+    if (!uri.codePoints().allMatch((cp: Int) => Character.isWhitespace(cp))) uri else DEFAULT_URI
```
stIncMale, line 32: `String.isBlank` was added in Java SE 11, so when I run `sdk use java 17.0.10-librca && ./gradlew -PjavaVersion=8 clean :driver-scala:integrationTest` locally, it fails.

The are two reasons this hasn't been a problem in Evergreen:

1. `run-tests.sh` did not run the `integrationTest` task (the `test` task does not depend on `integrationTest`).
2. `run-scala-tests.sh` did not use `JAVA_VERSION`. So, despite this script running `integrationTest` implicitly, it was using the same JDK Gradle is run by.

.evergreen/run-tests.sh
```diff
@@ -144,5 +144,5 @@ else
               ${MULTI_MONGOS_URI_SYSTEM_PROPERTY} ${API_VERSION} ${GRADLE_EXTRA_VARS} ${ASYNC_TYPE} \
               ${JAVA_SYSPROP_NETTY_SSL_PROVIDER} \
               -Dorg.mongodb.test.fle.on.demand.credential.test.failure.enabled=true \
-              --stacktrace --info --continue test
+              --stacktrace --info --continue test integrationTest
```
jyemin, line 147: This was intentionally omitted here since it only affect Scala and the extra test coverage provided by running these tests against every server version was not considered beneficial enough to be worth it.  Can we get the coverage we need with Scala-specific tests?

build.gradle
```diff
@@ -256,8 +256,12 @@ configure(javaCodeCheckedProjects) {
         testImplementation platform('org.spockframework:spock-bom:2.1-groovy-3.0')
         testImplementation 'org.spockframework:spock-core'
         testImplementation 'org.spockframework:spock-junit4'
-        testImplementation("org.mockito:mockito-core:3.8.0")
-        testImplementation("org.mockito:mockito-inline:3.8.0")
+        if ('8'.equals(findProperty("javaVersion"))) {
+            testImplementation("org.mockito:mockito-core:4.6.1")
+            testImplementation("org.mockito:mockito-inline:4.6.1")
```
jyemin, line 261: Just curious why `mockito-inline` is not needed for 5.11.  Was it integrated into core?

stIncMale: I don't know, but I confirmed with experiments that it's needed when 4.6.1 is used, and is not needed when 5.11 is used.

.evergreen/run-tests.sh
stIncMale: Done. Also updated https://github.com/mongodb/mongo-java-driver/pull/1326#discussion_r1511881774 correspondingly.

jyemin (approved):

.evergreen/.evg.yml
```diff
@@ -2037,7 +2041,8 @@ buildvariants:
     - name: "test"
 
 - matrix_name: "tests-jdk-secure"
-  matrix_spec: { auth: "auth", ssl: "ssl", jdk: [ "jdk8", "jdk17" ], version: [ "3.6", "4.0", "4.2", "4.4", "5.0", "6.0", "7.0", "latest" ],
+  matrix_spec: { auth: "auth", ssl: "ssl", jdk: [ "jdk8", "jdk17", "jdk21"],
```
rozza, line 2044: As part of MongoDBs green sustainability drive we should review how much resources the matrix takes up and if we can keep coverage without duplicating testing. I'll open a ticket.

rozza (approved):
LGTM

PR #1327, opened 2024-03-05
Ensure Operations uses the supplied document class when creating BsonDocumentWrapper

rozza:
JAVA-5349


driver-sync/src/test/functional/com/mongodb/client/AbstractMongoCollectionTest.java
```diff
@@ -0,0 +1,318 @@
...
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
+import static org.bson.codecs.configuration.CodecRegistries.fromRegistries;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeFalse;
+
+public abstract class AbstractMongoCollectionTest {
```
rozza, line 65: Made abstract so that can test Operations changes via both sync and async paths

jyemin:
From the Jira ticket

> This can lead to loss of discriminator information with bson-kotlinx.

Does the same problem happen with PojoCodec?

rozza:
> Does the same problem happen with PojoCodec?
Discriminators in the PojoCodec are configured by annotations so its most likely inherited with Pojos.

rozza:
I'm concerned this could have an unintended consequences on application code. 

It may be better to ensure that bson-kotlinx handles the lookup for polymorphic classes.

robbamberg:
> That could fix both usecases, however, it could easily break existing `bson-kotlinx` users. It depends if there are many users that use `MongoCollection<Cat>` and don't want to store the discriminator but when using `MongoCollection<Pet>` they want discriminator support. Any thoughts?
> 

@rozz I can imagine you could want a Cat collection without the discriminator field, so breaking that flow feels risky for me as well. 
So I prefer your current fix, as that is very clear that it is currently not working as intended. If I insert a Cat in a Pet collection I always need a discriminator, else I would have just created a Cat collection. Especially when during deserializing the collection class is taking into account, so would make sense to do it on serializing as well.

rozza:
@jyemin - this is ready for review. I think this is the correct behavior and is of low risk.

I had thought that I just change the Codec for bson-kotlinx but that would be higher risk and limit control for users. [JAVA-5338](https://jira.mongodb.org/browse/JAVA-5338) will require more thought on how best to fix for kotlinx users.

driver-sync/src/test/functional/com/mongodb/client/AbstractMongoCollectionTest.java
```diff
@@ -0,0 +1,318 @@
...
+            @Override
+            public Class<ShapeModelAbstract> getEncoderClass() {
+                return ShapeModelAbstract.class;
+            }
+        };
+        Codec<ShapeModelCircle> circleCodec = new Codec<ShapeModelCircle>() {
+
+            @Override
+            public void encode(final BsonWriter writer, final ShapeModelCircle value, final EncoderContext encoderContext) {
+                throw new UnsupportedOperationException("Not supported");
```
jyemin, line 287: IIUC we need these two custom Codec's here because, unlike bson-kotlinx, the PojoCodec will add a discriminator no matter which codec is used, the base class or the derived class.   And the test is relying on this method to throw in order to prove that the correct codec is used.  Is that correct?

If so, let's change `"Not Supported"` to something like `"If this method is called it means the document class codec was not properly used"`

jyemin (changes requested):

driver-sync/src/test/functional/com/mongodb/client/AbstractMongoCollectionTest.java
rozza: The premise is if the collection is `MongoCollection<Pet>` but the insert is of a `Cat` instance, it should use the codec for `Pet` not for `Cat`. So the test has two custom Codecs to prove that the correct one is picked.

With kotlinx serialization discriminators are declared via a [`PolymorphicSerializer`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-polymorphic-serializer/index.html) and not on the data class itself.  So when looking up for a codec the `Pet` codec will use polymorphic serialization but the `Cat` codec won't and therefore be missing a discriminator.

rozza:
I tested this code fix with the Pojos and the PojoCodec and found that the behavior remains the same - although somethings surprised me and / or I had forgotten.


### Interface: 

```
@BsonDiscriminator
public interface Pet {
    String getType();
}

public class Cat implements Pet {

    private String name;
    // ...
}
```

Findings: 
  * MongoCollection<Pet>#insertOne(cat) Succeeds: But Discriminator field NOT stored.
  * MongoCollection<Pet>#insertMany(cats) Succeeds: But Discriminator field NOT stored.
  * MongoCollection<Pet>#find() fails: `Cannot find a public constructor for 'Pet'.  Please ensure the class has a public, empty constructor with no arguments, or else a constructor with a BsonCreator annotation`
  
###  Interface 2: 

```
public interface Pet {
    String getType();
}

@BsonDiscriminator
public class Cat implements Pet {

    private String name;
    // ...
}
```

Findings: 
  * MongoCollection<Pet>#insertOne(cat) Succeeds: Has discriminator.
  * MongoCollection<Pet>#insertMany(cats) Succeeds: Has discriminator.
  * MongoCollection<Pet>#find() fails: `Cannot find a public constructor for 'Pet'.  Please ensure the class has a public, empty constructor with no arguments, or else a constructor with a BsonCreator annotation`

### Abstract

```
@BsonDiscriminator
public abstract class Pet {
    abstract String getType();
}

public class Cat extends Pet {

    private String name;
    // ...
}
```

Findings:
  * Regardless of this change it doesn't work
  * MongoCollection<Pet>#insertOne(cat) Fails: `Can't find a codec for CodecCacheKey{clazz=class tour.Pet, types=null}.`
  * Fails regardless of the change as `getCodec` is called to check for collectible codec in both paths.

### Extends
```
@BsonDiscriminator
public class Pet {
    public Pet() {
    }

    public String getType() {
        return "PET";
    }
}

public class Cat extends Pet {

    private String name;
    // ...
}
```
Findings:
  * MongoCollection<Pet>#insertOne(cat) Succeeds: Has discriminator.
  * MongoCollection<Pet>#insertMany(cats) Succeeds: Has discriminator.
  * MongoCollection<Pet>#find() Succeeds

jyemin:
> I tested this code fix with the Pojos and the PojoCodec and found that the behavior remains the same - although somethings surprised me and / or I had forgotten.

Good to know.  Does this change anything about what we want to do with this PR, or suggest additional work that we should track going forward?

rozza:
Added [JAVA-5370](https://jira.mongodb.org/browse/JAVA-5370) and [JAVA-5371](https://jira.mongodb.org/browse/JAVA-5371) to track the PojoCodec issues.

jyemin (approved):
Seems like this is good to merge then.

LGTM

PR #1328, opened 2024-03-05
Converted legacy CRUD tests to unified format

jyemin:
JAVA-5343


driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
```diff
@@ -44,9 +45,10 @@ public void assertValuesMatch(final BsonValue expected, final BsonValue actual)
         assertValuesMatch(expected, actual, null, -1);
     }
 
-    private void assertValuesMatch(final BsonValue initialExpected, @Nullable final BsonValue actual,
+    private void assertValuesMatch(final BsonValue initialExpected, @Nullable final BsonValue initialActual,
                                    @Nullable final String keyContext, final int arrayPositionContext) {
         BsonValue expected = initialExpected;
+        BsonValue actual = initialActual == null ? BsonNull.VALUE : initialActual;
```
jyemin, line 51: This to handle the first occurrence of `expectedResult: null` in a spec test.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudTest.java
```diff
@@ -34,6 +34,13 @@ public UnifiedCrudTest(@SuppressWarnings("unused") final String fileDescription,
                            final String schemaVersion, @Nullable final BsonArray runOnRequirements, final BsonArray entities,
                            final BsonArray initialData, final BsonDocument definition) {
         super(schemaVersion, runOnRequirements, entities, initialData, definition);
+        customSkips(fileDescription, testDescription);
+    }
+
+    public static void customSkips(final String fileDescription, final String testDescription) {
+        assumeFalse(fileDescription.equals("count"));
```
jyemin, line 41: Because the Java driver has already removed the deprecated `count` method.

driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
```diff
@@ -82,8 +84,6 @@ private void assertValuesMatch(final BsonValue initialExpected, @Nullable final
                 }
             }
 
-            assertNotNull(context.getMessage("Actual value must contain key " + keyContext), actual);
```
jyemin, line 85: Given the above change this assertion can never fire, so removed it.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudTest.java
stIncMale: Each of the files skipped here:

- https://github.com/mongodb/specifications/blob/master/source/crud/tests/unified/count.json
- https://github.com/mongodb/specifications/blob/master/source/crud/tests/unified/count-collation.json
- https://github.com/mongodb/specifications/blob/master/source/crud/tests/unified/count-empty.json

has at least one test that uses the `countDocuments` operation, and not the `count` operation. Thus, it seems we should not skip these files, but only some of the tests in them.

P.S. The updated `UnifiedCrudHelper.executeCountDocuments` also suggests that at some point you intended to have at least `count.json` to be run.

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudTest.java
katcharov: Match on description contains `"Deprecated count"`?

driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
jyemin: I removed that line, and also removed handling for `$$exists` in the same switch statement, as test runs show that the case is not entered, and the assertions in the block don't make sense if `actual` is never null.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudTest.java
jyemin: I just enumerated all of them since there are only six, and it matches the pattern just below.

stIncMale (approved):

jyemin:
I was wrong about $$exists.  It is needed.  Going to need to do another patch.

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
```diff
@@ -273,9 +273,9 @@ public String toString() {
                 builder.append("   Array position: ").append(arrayPosition).append("\n");
             }
             builder.append("   Expected value:\n      ");
-            builder.append(expected.toString()).append("\n");
+            builder.append(expected).append("\n");
             builder.append("   Actual value:\n      ");
-            builder.append(actual.toString()).append("\n");
```
jyemin, line 278: This now works when `actual` is null.

driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
jyemin: This assertion is now covered is the new conditional below for BsonNull

driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
jyemin: After running the full test suite against more variants, it's clear that `$$exists` is still needed here.  I changed the logic in a different way to handle `actual == null` which I think makes more sense.

driver-sync/src/test/functional/com/mongodb/client/unified/ContextElement.java
jyemin: Done

stIncMale (approved):

PR #1330, opened 2024-03-06
Convert legacy retryable reads tests to unified format

jyemin:
JAVA-5344


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedRetryableReadsTest.java
```diff
@@ -24,12 +24,22 @@
...
-                                      final String schemaVersion, final BsonArray runOnRequirements, final BsonArray entitiesArray,
-                                      final BsonArray initialData, final BsonDocument definition) {
+    public UnifiedRetryableReadsTest(final String fileDescription, final String testDescription, final String schemaVersion,
+            final BsonArray runOnRequirements, final BsonArray entitiesArray, final BsonArray initialData, final BsonDocument definition) {
         super(schemaVersion, runOnRequirements, entitiesArray, initialData, definition);
+        customSkips(fileDescription, testDescription);
+    }
+
+    public static void customSkips(final String fileDescription, @SuppressWarnings("unused") final String testDescription) {
+        assumeFalse(fileDescription.equals("count"));
```
rozza, line 37: Perhaps add a comment why these are skipped.

driver-scala/src/integration/scala/org/mongodb/scala/RetryableReadsTest.scala
```diff
@@ -1,44 +0,0 @@
...
- * limitations under the License.
- */
-
-package org.mongodb.scala
-
-import com.mongodb.client.AbstractRetryableReadsTest
-import org.bson.{ BsonArray, BsonDocument, BsonString }
-import org.mongodb.scala.syncadapter.SyncMongoClient
-
-class RetryableReadsTests(
```
rozza, line 23: Is there a unified test for Scala now?

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedRetryableReadsTest.java
jyemin: Done

driver-scala/src/integration/scala/org/mongodb/scala/RetryableReadsTest.scala
jyemin: No.  Given that Kotlin, Kotlin co-routine, and Scala are all thin wrappers around the Java-based drivers, I don't think it's worth it to test retryable reads against all of them, as all the logic that's being tested lives in driver-core and is well-tested by the test runners for sync and reactive drivers.

Comparing with the other unified tests, I see only UnifiedCrudTest for kotlin-coroutine and kotlin-sync, so this seems in line with previous decisions.

vbabanin (approved):
LGTM!

rozza (approved):
LGTM

PR #1332, opened 2024-03-11
Convert legacy data lake tests to unified

jyemin:
JAVA-5354


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -267,7 +267,7 @@ OperationResult executeListDatabaseNames(final BsonDocument operation) {
     OperationResult executeListCollections(final BsonDocument operation) {
         MongoDatabase database = entities.getDatabase(operation.getString("object").getValue());
 
-        BsonDocument arguments = operation.getDocument("arguments");
+        BsonDocument arguments = operation.getDocument("arguments", new BsonDocument());
```
jyemin, line 270: This change was actually necessary to get the tests to pass.  The rest below I changed because this keeps happening.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -379,9 +379,9 @@ OperationResult createFindCursor(final BsonDocument operation) {
     @NonNull
     private FindIterable<BsonDocument> createFindIterable(final BsonDocument operation) {
         MongoCollection<BsonDocument> collection = entities.getCollection(operation.getString("object").getValue());
-        BsonDocument arguments = operation.getDocument("arguments");
+        BsonDocument arguments = operation.getDocument("arguments", new BsonDocument());
         ClientSession session = getSession(arguments);
-        BsonDocument filter = arguments.getDocument("filter", new BsonDocument());
```
jyemin, line 384: This matches all the other places where `filter` is required.  I had to make this change after changing the line two lines up, as the exception thrown on that line when there are no arguments was incorrectly causing the `createFindCursor fails if filter is not specified` in `unified-test-format/valid-fail` to pass.

stIncMale (changes requested):
The spec says [_"Test runners MUST NOT execute `killAllSessions` when connected to Atlas Data Lake."_](https://github.com/mongodb/specifications/commit/fd5a68d749d37d8afe20f78eeb0a2491f8fa61a0#diff-0f874d282835bde8ccd8e9f7920996c641233f6885b2a4a41cb1bb9ba644756fR3124), yet `com.mongodb.client.unified.UnifiedTest.setUp` (a superclass of `UnifiedAtlasDataLakeTest`) calls `CollectionHelper.killAllSessions` unconditionally. It seems to me that we should call `killAllSessions` only if `ClusterFixture.isDataLakeTest` returns `true`.

jyemin:
> It seems to me that we should call killAllSessions only if ClusterFixture.isDataLakeTest returns true.

I'm inclined to leave it as is.  `AbstractUnifiedTest`, which the now-deleted legacy test extends, executes `killAllSessions` without such a check, and those tests consistently passed.  The new unified tests also pass.  

@jmikola I looked at your PR for this change in the specs repo (that Valentin linked to above), and I'm not clear where this restriction on killAllSessions for Data Lake originated.  I am likely overlooking something, but it does seem unnecessary.

jyemin:
I confirmed that killAllSessions is not supported on Data Lake, and it only works in the Java driver because we catch exceptions in the helper for killAllSessions.  So I will add the check in as Valentin requested.

stIncMale (approved):

PR #1333, opened 2024-03-12
Centralize appending of maxTimeMS

vbabanin:
<h1>Description:</h1>

This PR tackles the challenge identified in JAVA-4055 related to the handling of maxTimeMS in the context of Client-Side Field Level Encryption (CSFLE) by moving appending of maxTimeMS to a centralized and lower level within the driver, specifically to the CommandMessage.

<h2>Implementation details</h2>

Operations handle maxTimeMS differently:
- A majority append maxTimeMS when it is non-zero.
- Some read operations check if cursor is tailable or non-tailable.
- Tailable await cursors append maxAwaitTimeMS as maxTimeMS to a command.

Therefore, this PR introduces `MaxTimeMsSupplier` in `TimeoutContext`. This  ensures `CommandMessage` remains agnostic to the operation's logic, allowing operations to dictate it.

JAVA-5322
JAVA-5327


driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -319,4 +324,15 @@ public Timeout getTimeout() {
         return timeout;
     }
 
+    public void addExtraElements(final List<BsonElement> extraElements) {
+        long maxTimeMS = maxTimeSupplier.get();
```
rozza, line 328: Nit: Long maxTimeMS

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -136,11 +138,8 @@ public long timeoutOrAlternative(final long alternativeTimeoutMS) {
         Long timeoutMS = timeoutSettings.getTimeoutMS();
         if (timeoutMS == null) {
             return alternativeTimeoutMS;
-        } else if (timeoutMS == 0) {
-            return timeoutMS;
-        } else {
-            return timeoutRemainingMS();
         }
```
vbabanin, line 141: `else if (timeoutMS == 0)` is removed due to a logic issue where using an original infinite timeout (timeoutMS = 0) incorrectly led to always returning 0 from TimeoutSettings in `timeoutOrAlternative`, regardless of actual `computedServerSelectionTimeout` time left. Now, `timeoutRemainingMS()` handles both infinite and finite timeouts, checking against the `timeout` object itself:

For the reference: 
`  private long timeoutRemainingMS() {
        assertNotNull(timeout);
        if (timeout.hasExpired()) {
            throw createMongoTimeoutException("The operation timeout has expired.");
        }
        return timeout.isInfinite() ? 0 : timeout.remaining(MILLISECONDS);
    }`

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -215,8 +216,14 @@ private boolean useOpMsg() {
         return getOpCode().equals(OpCode.OP_MSG);
     }
 
-    private List<BsonElement> getExtraElements(final SessionContext sessionContext) {
+    private List<BsonElement> getExtraElements(final OperationContext operationContext) {
+        SessionContext sessionContext = operationContext.getSessionContext();
+        TimeoutContext timeoutContext = operationContext.getTimeoutContext();
+
         List<BsonElement> extraElements = new ArrayList<>();
+        if (!getSettings().isCryptd()) {
```
vbabanin, line 224: From the spec:
> When sending a command to mongocryptd, drivers MUST NOT append a maxTimeMS field. This is to ensure that a maxTimeMS field can be safely appended to the command after it has been marked by mongocryptd and encrypted by libmongocrypt. To determine whether or not the server is a mongocryptd, drivers MUST check that the iscryptd field in the server's description is true.
> 
> Rationale:  https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#maxtimems-is-not-added-for-mongocryptd

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
vbabanin: Done.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -215,8 +216,14 @@ private boolean useOpMsg() {
...
+            timeoutContext.addTimeoutElements(extraElements);
```
rozza, line 225: Something about extending extraElements in `TimeoutContext` feels strange.  Feels like the logic should be here.

driver-core/src/main/com/mongodb/internal/operation/OperationHelper.java
```diff
@@ -197,16 +196,13 @@ static boolean canRetryRead(final ServerDescription serverDescription, final Ope
...
+        addMaxTimeMSToNonTailableCursor(TimeoutMode.CURSOR_LIFETIME, operationContext);
     }
 
-    static void addMaxTimeMSToNonTailableCursor(final BsonDocument commandDocument, final TimeoutMode timeoutMode,
-            final OperationContext operationContext) {
-        long maxTimeMS = timeoutMode == TimeoutMode.ITERATION ? 0 : operationContext.getTimeoutContext().getMaxTimeMS();
-        if (maxTimeMS > 0) {
-            commandDocument.put("maxTimeMS", new BsonInt64(maxTimeMS));
-        }
+    static void addMaxTimeMSToNonTailableCursor(final TimeoutMode timeoutMode, final OperationContext operationContext) {
```
rozza, line 203: Naming no longer seems appropriate: Sugges: `setNonTailableCursorMaxTimeSupplier`.

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -319,4 +323,15 @@ public Timeout getTimeout() {
         return timeout;
     }
 
+    public void addTimeoutElements(final List<BsonElement> extraElements) {
```
rozza, line 326: Feels odd having this here.

driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
```diff
@@ -90,6 +91,8 @@ class AsyncCommandBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T>
         this.maxWireVersion = connectionDescription.getMaxWireVersion();
         this.firstBatchEmpty = commandCursorResult.getResults().isEmpty();
 
+        connectionSource.getOperationContext().getTimeoutContext().setMaxTimeSupplier(() -> maxTimeMS);
```
rozza, line 94: Not sure `maxTimeMS` is needed in this class (or the sync equivalent).  Setting of MaxTimeSupplier could be done in a similar way as `addMaxTimeMSToNonTailableCursor` is used.

rozza (changes requested):
Its looking good and is much cleaner than I initially thought it would be.

Found a couple of nits but I also recommend a couple of minor changes to the API: 

### Firstly: 
Have `TimeoutContext.getMaxTimeMS()` be the thing that consumes the supplier eg:

```
    public Long getMaxTimeMS() {
       return notNull("Should never be null", maxTimeoutSupplier.get());
    }
```

Convert the current `getMaxTimeMS` to be protected and used internally eg: rename it to `defaultMaxTimeMS`.

Add a `timeoutContext.resetSupplier()` that resets the supplier to the default. Rather than doing: `timeoutContext.setMaxTimeSupplier(timeoutContext::getMaxTimeMS);`

### Secondly:

Remove `TimeoutContext#addTimeoutElements` put the logic in command message.

### Secondly:
I think `CommandOperationHelper#appendMaxTimeMs` can now be removed.

driver-core/src/main/com/mongodb/connection/ServerDescription.java
```diff
@@ -70,6 +70,10 @@ public class ServerDescription {
     private final ServerAddress address;
 
     private final ServerType type;
+    /**
+     * Identifies whether the server is a mongocryptd.
```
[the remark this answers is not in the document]
vbabanin: Removed JavaDoc from private field and added to cryptd() builder method.

driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
vbabanin: Given that `maxTime` is closely linked to the cursor's behavior, maintaining its setting within proximity to the related logic, i believe, improves clarity. Setting maxTime outside of the `CommandBatch `cursor could lead to confusion, as it might not be immediately apparent that there is `maxTime` which is configured externally from the cursor class.

vbabanin:
> Its looking good and is much cleaner than I initially thought it would be.
> 
> Found a couple of nits but I also recommend a couple of minor changes to the API:
> 
> ### A):
> Have `TimeoutContext.getMaxTimeMS()` be the thing that consumes the supplier eg:
> 
> ```
>     public Long getMaxTimeMS() {
>        return notNull("Should never be null", maxTimeoutSupplier.get());
>     }
> ```
> 
> Convert the current `getMaxTimeMS` to be protected and used internally eg: rename it to `defaultMaxTimeMS`.
> 
> Add a `timeoutContext.resetSupplier()` that resets the supplier to the default. Rather than doing: `timeoutContext.setMaxTimeSupplier(timeoutContext::getMaxTimeMS);`
> 
> ### B):
> Remove `TimeoutContext#addTimeoutElements` put the logic in command message.
> 
> ### C):
> I think `CommandOperationHelper#appendMaxTimeMs` can now be removed.
> 
> What do you think?



> Its looking good and is much cleaner than I initially thought it would be.
> 
> Found a couple of nits but I also recommend a couple of minor changes to the API:
> 
> ### A):
> Have `TimeoutContext.getMaxTimeMS()` be the thing that consumes the supplier eg:
> 
> ```
>     public Long getMaxTimeMS() {
>        return notNull("Should never be null", maxTimeoutSupplier.get());
>     }
> ```
> 
> Convert the current `getMaxTimeMS` to be protected and used internally eg: rename it to `defaultMaxTimeMS`.
> 
> Add a `timeoutContext.resetSupplier()` that resets the supplier to the default. Rather than doing: `timeoutContext.setMaxTimeSupplier(timeoutContext::getMaxTimeMS);`
> 
> ### B):
> Remove `TimeoutContext#addTimeoutElements` put the logic in command message.
> 
> ### C):
> I think `CommandOperationHelper#appendMaxTimeMs` can now be removed.
> 
> What do you think?

Thank you for the suggestions @rozza!

A) Agreed. This change will maintain consistency in the TimeoutContext API with other methods. Implemented.

B) Implemented as suggested.

C) Agreed and removed.

rozza:
@vbabanin I think one of the mocks needs updating as its causing all the tests to fail.

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/client/csot/AbstractClientSideOperationsEncryptionTimeoutProseTest.java
```diff
@@ -200,12 +199,21 @@ void shouldThrowOperationTimeoutExceptionWhenDecryptData() {
      * Not a prose spec test. However, it is additional test case for better coverage.
      */
     @Test
-    @Disabled
-    // TODO JAVA-5322. We have to reset timeoutMS for encrypted command.
+    @Tag("setsFailPoint")
```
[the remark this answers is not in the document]
rozza: This was just a pattern adopted from other code sources and allows for specific tests to manage turning off failpoints in the clean up.  Its preferred over extensions as its  done on a per test basis rather than the whole test class.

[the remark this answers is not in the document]
stIncMale: > This was just a pattern adopted from other code sources

As far as I can see, this is a pattern that was introduced in 2024 in CSOT, it does not exist in `master`.

> its done on a per test basis rather than the whole test class

The same can be achieved by raising a flag being part of test's instance state, and turning off failpoints in the `@AfterEach` method based on that flag.

> It has the benefit, should we run our test suite in the future against a topology that doesn't have perms to set fail points, we can easily filter the tests out.

Everywhere in the driver we skip tests depending on the environment by using `org.junit.jupiter.api.Assumptions`/`org.junit.Assume`, which means that the `@Tag` approach is inconsistent with the current approach and it's unlikely that we will start skipping tests depending on the environment using another mechanism.

Thus, I still don't think that it's a good idea to use `@Tag`s for turning failpoints off.

P.S. I realize that this PR is not a place to make any changes with regard to the approach, even if we eventually agree that it shouldn't be used. So I'll leave a comment in https://github.com/mongodb/mongo-java-driver/pull/1215.

driver-core/src/main/com/mongodb/internal/operation/CountOperation.java
```diff
@@ -139,7 +139,7 @@ private CommandCreator getCommandCreator() {
             putIfNotZero(document, "limit", limit);
             putIfNotZero(document, "skip", skip);
             putIfNotNull(document, "hint", hint);
-            putIfNotZero(document, "maxTimeMS", operationContext.getTimeoutContext().getMaxTimeMS());
+            // putIfNotZero.*maxTimeMs".*;
```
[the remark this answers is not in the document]
vbabanin: It was an oversight and has now been removed. Thank you for catching that!

driver-core/src/main/com/mongodb/connection/ServerDescription.java
```diff
@@ -644,6 +660,14 @@ public boolean isSecondary() {
         return ok && (type == REPLICA_SET_SECONDARY || type == SHARD_ROUTER || type == STANDALONE || type == LOAD_BALANCER);
     }
 
+    /**
+     * Returns whether this server is mongocrpytd.
```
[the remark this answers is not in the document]
vbabanin: Added.

driver-core/src/test/unit/com/mongodb/internal/connection/CommandMessageSpecification.groovy
```diff
@@ -92,35 +94,44 @@ class CommandMessageSpecification extends Specification {
...
-                            getReadConcern() >> ReadConcern.DEFAULT
-                        },
-                        Stub(SessionContext) {
-                            hasSession() >> true
-                            getClusterTime() >> new BsonDocument('clusterTime', new BsonTimestamp(42, 1))
-                            getSessionId() >> new BsonDocument('id', new BsonBinary([1, 2, 3] as byte[]))
-                            getReadConcern() >> ReadConcern.DEFAULT
-                            }
+                        new OperationContext(
+                                Stub(RequestContext),
```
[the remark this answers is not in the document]
vbabanin: Changed to `IgnorableRequestContext.INSTANCE`, thanks!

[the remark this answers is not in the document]
stIncMale: `LoggingCommandEventSenderSpecification.groovy` still needs an update.

driver-sync/src/test/functional/com/mongodb/client/csot/AbstractClientSideOperationsEncryptionTimeoutProseTest.java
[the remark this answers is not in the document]
stIncMale: The comment in the CSOT PR: https://github.com/mongodb/mongo-java-driver/pull/1215#discussion_r1536885331.

stIncMale (approved):

rozza (approved):

PR #1335, opened 2024-03-12
Convert Bson to BsonDocument for hint

jyemin:
Unlike every other instance of Bson in the CRUD API, the hint option for write operations was not converted to a BsonDocument prior to passing it down to the operation layer.  This is a problem because the operation layer doesn't have a CodecRegistry available, and so can not properly due the conversion.

In this commit, all hint options are converted to BsonDocument in the same place as all the other Bson options, where a CodecRegistry is available to properly do the conversion.

JAVA-5357

Note that no additional tests have been added.  The coverage for hint is the same as for all the other Bson-typed options (e.g. filter, projection).


katcharov (approved):
LGTM!

PR #1336, opened 2024-03-12
Use ec2.assume_role

katcharov:
DEVPROD-5197

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -111,6 +111,9 @@ functions:
           ${PROJECT_DIRECTORY}/${file}
 
   "upload mo artifacts":
+    - command: ec2.assume_role
+      params:
+        role_arn: ${ASSUME_ROLE_ARN}
     - command: shell.exec
       params:
         shell: bash
@@ -119,8 +122,9 @@ functions:
           find $MONGO_ORCHESTRATION_HOME -name \*.log | xargs tar czf mongodb-logs.tar.gz
     - command: s3.put
       params:
-        aws_key: ${aws_key}
-        aws_secret: ${aws_secret}
+        aws_key: ${AWS_ACCESS_KEY_ID}
+        aws_secret: ${AWS_SECRET_ACCESS_KEY}
+        aws_session_token: ${AWS_SESSION_TOKEN}
         local_file: mongodb-logs.tar.gz
         remote_file: ${UPLOAD_BUCKET}/${build_variant}/${revision}/${version_id}/${build_id}/logs/${task_id}-${execution}-mongodb-logs.tar.gz
         bucket: mciuploads
@@ -129,8 +133,9 @@ functions:
         display_name: "mongodb-logs.tar.gz"
     - command: s3.put
       params:
-        aws_key: ${aws_key}
-        aws_secret: ${aws_secret}
+        aws_key: ${AWS_ACCESS_KEY_ID}
+        aws_secret: ${AWS_SECRET_ACCESS_KEY}
+        aws_session_token: ${AWS_SESSION_TOKEN}
         local_file: drivers-tools/.evergreen/orchestration/server.log
         remote_file: ${UPLOAD_BUCKET}/${build_variant}/${revision}/${version_id}/${build_id}/logs/${task_id}-${execution}-orchestration.log
         bucket: mciuploads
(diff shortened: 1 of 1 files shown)


.evergreen/.evg.yml
```diff
@@ -111,6 +111,9 @@ functions:
           ${PROJECT_DIRECTORY}/${file}
 
   "upload mo artifacts":
+    - command: ec2.assume_role
```
cbiphuk, line 114: It should be a function.

```suggestion
functions:
  assume-ec2-role:
    command: ec2.assume_role
    params:
      role_arn: ${ASSUME_ROLE_ARN}
```

cbiphuk: And every task that requires iam permissions should call this function

cbiphuk (changes requested):

.evergreen/.evg.yml
katcharov: `upload mo artifacts` is a function, and I believe functions cannot call other functions. This seems to be the only function that uses `aws_key`, so I believe everything should work correctly.

.evergreen/.evg.yml
```diff
@@ -119,8 +122,9 @@ functions:
           find $MONGO_ORCHESTRATION_HOME -name \*.log | xargs tar czf mongodb-logs.tar.gz
     - command: s3.put
       params:
-        aws_key: ${aws_key}
-        aws_secret: ${aws_secret}
+        aws_key: ${AWS_ACCESS_KEY_ID}
```
jyemin, line 125: There are the same properties that we use for CSFLE/QE testing.  I'm surprised to see them used here as well.  Is this correct?

katcharov: It seems correct but suspicious. Since the [ec2.assume_role](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Commands#ec2assume_role) command is invoked directly above, and we do not use those project variables after, they should not override what we have set. I can confirm that the upload succeeds in the logs once dev prod provides us with the new `ASSUME_ROLE_ARN` value.

However, it is unclear why we have both `aws_access_key_id` and `aws_key` set in our project config (they have different values), and I would imagine that aws_access_key_id should be treated as reserved. We also have various keys that have similar values, such as `iam_auth_assume_aws_account` and `iam_auth_ecs_account`. Should these be changed to assume roles as well? [Other drivers](https://github.com/mongodb/mongo-csharp-driver/blob/27ee0b4c5c59737f3f2efc85bfea65f795f7ff25/evergreen/evergreen.yml#L134) use static credentials, but also use `ec2.assume_role` in other places.

My inclination is to merge this in, to get things working for the new project/branch, and then do any other work under a new DRIVERS ticket created by dev prod (it should be possible to identify static keys to be deprecated, for each driver, and get everyone switched over in a scheduled way; this would save the issue from arising during something like a release).

jyemin: > However, it is unclear why we have both aws_access_key_id and aws_key set in our project config (they have different values),

* `aws_key` is in pretty much every Evergreen project config and its value is unique to each project. It's generally used for pushing artifacts (e.g log files, test results) to S3.
* `aws_access_key_id` is specifically for CSFLE/QE drivers testing and its value is the same for all drivers.  While it seems possible to use it to push to S3, that's not what it's intended for.

rozza:
I removed myself from the review as there are two other active reviewers.

.evergreen/.evg.yml
katcharov: Discussed offline, we're fine to keep as is.

katcharov:
Merging. The role is [now working](https://evergreen.mongodb.com/task_log_raw/mongo_java_driver_static_checks_static_analysis_patch_5a58c4b9325ba5416637fcb048d43a7a22e2bc85_65f9abd8f71b730007b2034a_24_03_19_15_14_34/4?type=T#L7396). There is an [ensuing failure](https://evergreen.mongodb.com/task_log_raw/mongo_java_driver_static_checks_static_analysis_patch_5a58c4b9325ba5416637fcb048d43a7a22e2bc85_65f9abd8f71b730007b2034a_24_03_19_15_14_34/4?type=T#L7428), which is [not new and goes back to at least Jan](https://evergreen.mongodb.com/task_log_raw/mongo_java_driver_static_checks_static_analysis_patch_d80e9c1de594113001a47ce0e8f9db5baca37249_65b0e8cf306615ce929e752a_24_01_24_10_39_13/0?type=T#L7548).

PR #1337, opened 2024-03-13
Change mapreduce to mapReduce

jyemin:
The server accepts both but new unified spec tests require the latter.

JAVA-5363


rozza (approved):
LGTM

PR #1339, opened 2024-03-14
Fix getCodec of matching type argument bug

ht-jo:
JAVA-5365


rozza:
Thanks @ht-jo for the PR and the ticket. We'll triage and review in the near future 👍

rozza (commented):
LGTM!

PR #1348, opened 2024-03-20
Validate timeout and timeunit values

rozza:
Ensure that users can't accidently set a 0 value
which in CSOT represents infinite timeouts.

JAVA-5351

### DIFF
--- a/driver-core/src/main/com/mongodb/ClientSessionOptions.java
+++ b/driver-core/src/main/com/mongodb/ClientSessionOptions.java
@@ -24,8 +24,8 @@ import com.mongodb.session.ClientSession;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
-import static com.mongodb.assertions.Assertions.isTrueArgument;
 import static com.mongodb.assertions.Assertions.notNull;
+import static com.mongodb.internal.TimeoutSettings.convertAndValidateTimeout;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 /**
@@ -224,9 +224,7 @@ public final class ClientSessionOptions {
          * @see #getDefaultTimeout
          */
         public Builder defaultTimeout(final long defaultTimeout, final TimeUnit timeUnit) {
-            this.defaultTimeoutMS = isTrueArgument("timeoutMS must be >= 0",
-                    () -> MILLISECONDS.convert(defaultTimeout, timeUnit),
-                    (timeoutMS) -> defaultTimeout == 0 && timeoutMS == 0 || timeoutMS > 0);
+            this.defaultTimeoutMS = convertAndValidateTimeout(defaultTimeout, timeUnit, "defaultTimeout");
             return this;
         }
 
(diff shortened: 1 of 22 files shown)


driver-core/src/test/unit/com/mongodb/ClientSessionOptionsSpecification.groovy
```diff
@@ -18,6 +18,8 @@ package com.mongodb
 
 import spock.lang.Specification
 
+import java.util.concurrent.TimeUnit
+
 class ClientSessionOptionsSpecification extends Specification {
```
rozza, line 23: This class was also moved from the drivers-legacy dir. There seemed to be no reason to have it there as it tests driver-core functionality.

driver-core/src/test/unit/com/mongodb/ClientEncryptionSettingsSpecification.groovy
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package com.mongodb
+
+import spock.lang.Specification
+
+import javax.net.ssl.SSLContext
+import java.util.concurrent.TimeUnit
+
+class ClientEncryptionSettingsSpecification extends Specification {
```
rozza, line 24: There was no unit tests for this class - so added this for extra coverage.

driver-core/src/main/com/mongodb/ClientSessionOptions.java
```diff
@@ -224,8 +224,9 @@ public Builder defaultTransactionOptions(final TransactionOptions defaultTransac
          * @see #getDefaultTimeout
          */
         public Builder defaultTimeout(final long defaultTimeout, final TimeUnit timeUnit) {
-            isTrueArgument("timeoutMS must be >= 0", defaultTimeout >= 0);
-            this.defaultTimeoutMS = MILLISECONDS.convert(defaultTimeout, timeUnit);
+            this.defaultTimeoutMS = isTrueArgument("timeoutMS must be >= 0",
```
vbabanin, line 227: To enhance clarity, we could update the validation message to be more descriptive, specifying the parameter name and differentiating between the two scenarios where validation fails. Currently, we use a single message for both cases.

I propose two distinct messages:
For a too small defaultTimeout: `defaultTimeout is too small. It was rounded to 0 milliseconds, resulting in an unintended infinite timeout.`
For a negative value after conversion: `defaultTimeout, after conversion to milliseconds, must be >= 0.`

This suggestion applies to all timeout/withTimeout methods.
What do you think?

vbabanin (commented):
Looks good!

One thing: I think we should add the same validation to the `withTimeout` methods of `MongoDatabase`, `MongoCollection`, and `GridFSBucket`.

driver-core/src/main/com/mongodb/ClientSessionOptions.java
rozza: SGTM!

rozza:
Doh! My search missed all the withTimeout methods - thanks.

vbabanin (approved):

PR #1351, opened 2024-03-21
Updated the unified specs tests

rozza:
- Added createOptions support for collection creation 
- Fixed error with awaitable tailable cursors

JAVA-5374


vbabanin (approved):
Everything looks great!

it appears some static checks require attention.

PR #1352, opened 2024-03-26
Expand bounds for distinct and MongoIterable#map

rozza:
Allow for nullable types in those scenarios.

JAVA-5358


jyemin (commented):
LGTM, but just want to confirm that the signature changes are source and binary compatible.

rozza:
Running https://github.com/Kotlin/binary-compatibility-validator showed no errors :)

PR #1353, opened 2024-03-27
Use `InetAddressResolverProvider` and add tests

stIncMale:
JAVA-5390

### DIFF
--- a/driver-core/src/main/com/mongodb/MongoClientSettings.java
+++ b/driver-core/src/main/com/mongodb/MongoClientSettings.java
@@ -184,12 +184,10 @@ public final class MongoClientSettings {
     }
 
     /**
-     * Gets the {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
-     *
-     * <p>If set, it will be used to look up the {@link java.net.InetAddress} for each host, via
-     * {@link InetAddressResolver#lookupByName(String)}. Otherwise, {@link java.net.InetAddress#getAllByName(String)} will be used.
+     * Gets the explicitly set {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
      *
      * @return the {@link java.net.InetAddress} resolver
+     * @see Builder#inetAddressResolver(InetAddressResolver)
      * @since 4.10
      */
     @Nullable
@@ -660,6 +658,7 @@ public final class MongoClientSettings {
          *
          * @param inetAddressResolver the InetAddress provider
          * @return the {@link java.net.InetAddress} resolver
+         * @see #getInetAddressResolver()
          * @since 4.10
          */
         public Builder inetAddressResolver(@Nullable final InetAddressResolver inetAddressResolver) {
--- a/driver-core/src/main/resources/META-INF/native-image/org.mongodb/bson/native-image.properties
+++ b/driver-core/src/main/resources/META-INF/native-image/org.mongodb/bson/native-image.properties
@@ -13,4 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-Args = --initialize-at-run-time=com.mongodb.UnixServerAddress,com.mongodb.internal.connection.SnappyCompressor,org.bson.types.ObjectId,com.mongodb.internal.connection.ClientMetadataHelper
+Args =\
+  --initialize-at-run-time=\
+    com.mongodb.UnixServerAddress,\
+    com.mongodb.internal.connection.SnappyCompressor,\
+    com.mongodb.internal.connection.ClientMetadataHelper,\
+    com.mongodb.internal.connection.ServerAddressHelper
--- a/driver-core/src/main/resources/META-INF/native-image/resource-config.json
+++ b/driver-core/src/main/resources/META-INF/native-image/resource-config.json
@@ -1,0 +1,7 @@
+{
+  "resources":{
+  "includes":[{
+    "pattern":"\\QMETA-INF/services/com.mongodb.spi.dns.InetAddressResolverProvider\\E"
+  }]},
+  "bundles":[]
+}
--- a/graalvm-native-image-app/build.gradle
+++ b/graalvm-native-image-app/build.gradle
@@ -51,9 +51,8 @@ graalvmNative {
         configureEach {
             buildArgs.add('--strict-image-heap')
             buildArgs.add('-H:+UnlockExperimentalVMOptions')
-            // see class initialization report is generated at `graalvm/build/native/nativeCompile/reports`,
-            // informing us on the kind of initialization for each Java class
-            buildArgs.add('-H:+PrintClassInitialization')
+            // see class initialization and other reports in `graalvm/build/native/nativeCompile/reports`
+            buildArgs.add('--diagnostics-mode')
             // see the "registerResource" entries in the `native-image` built-time output,
             // informing us on the resources included in the native image being built
             buildArgs.add('-H:Log=registerResource:5')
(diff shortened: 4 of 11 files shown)


driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -184,12 +184,10 @@ public DnsClient getDnsClient() {
     }
 
     /**
-     * Gets the {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
-     *
-     * <p>If set, it will be used to look up the {@link java.net.InetAddress} for each host, via
-     * {@link InetAddressResolver#lookupByName(String)}. Otherwise, {@link java.net.InetAddress#getAllByName(String)} will be used.
```
stIncMale, line 190: This was incorrect, as "Otherwise, ..." did not take into account `InetAddressResolverProvider`. `Builder#inetAddressResolver` explains the behavior in full, so I added `@see Builder#inetAddressResolver(InetAddressResolver)`.

driver-core/src/main/resources/META-INF/native-image/org.mongodb/bson/native-image.properties
```diff
@@ -13,4 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-Args = --initialize-at-run-time=com.mongodb.UnixServerAddress,com.mongodb.internal.connection.SnappyCompressor,org.bson.types.ObjectId,com.mongodb.internal.connection.ClientMetadataHelper
+Args =\
+  --initialize-at-run-time=\
+    com.mongodb.UnixServerAddress,\
+    com.mongodb.internal.connection.SnappyCompressor,\
+    com.mongodb.internal.connection.ClientMetadataHelper,\
```
stIncMale, line 20: I checked by looking at the reports generated by `native-image` that these `\` line breaks work fine.

driver-core/src/main/resources/META-INF/native-image/org.mongodb/bson/native-image.properties
```diff
@@ -13,4 +13,9 @@
...
+    com.mongodb.internal.connection.ServerAddressHelper
```
stIncMale, line 21: I checked by looking at the reports generated by `native-image` that these `\` line breaks work fine.

driver-core/src/main/resources/META-INF/native-image/resource-config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "resources":{
+  "includes":[{
+    "pattern":"\\QMETA-INF/services/com.mongodb.spi.dns.InetAddressResolverProvider\\E"
```
stIncMale, line 4: Weirdly, everything works even without this entry, but, if I collect reachability metadata, then this entry is added there.

graalvm-native-image-app/build.gradle
```diff
@@ -51,9 +51,8 @@ graalvmNative {
         configureEach {
             buildArgs.add('--strict-image-heap')
             buildArgs.add('-H:+UnlockExperimentalVMOptions')
-            // see class initialization report is generated at `graalvm/build/native/nativeCompile/reports`,
-            // informing us on the kind of initialization for each Java class
-            buildArgs.add('-H:+PrintClassInitialization')
+            // see class initialization and other reports in `graalvm/build/native/nativeCompile/reports`
+            buildArgs.add('--diagnostics-mode')
```
stIncMale, line 55: `--diagnostics-mode` includes the functionality of `-H:+PrintClassInitialization`, and more. So we better use `--diagnostics-mode`.

driver-core/src/main/resources/META-INF/native-image/resource-config.json
jyemin: That's strange.  Did you find that for any other reachability metatada?  Can you suggest a hypothesis for what might be going on?

jyemin (changes requested):

driver-core/src/main/resources/META-INF/native-image/resource-config.json
stIncMale: > Did you find that for any other reachability metatada?

I haven't checked everything, nor do I understand how some of the automatically collected resources are used, but removing the shared library resources, or the `logback-test.xml` resource do have an effect: the former caused runtime errors, the latter results in log entries disappearing.

> Can you suggest a hypothesis for what might be going on?

I suspect, it's all about undocumented behavior that is intended to make things easier. I found the following:

- https://github.com/oracle/graal/issues/563#issuecomment-459121737
- https://javadoc.io/doc/org.graalvm.nativeimage/svm/20.3.10/com/oracle/svm/hosted/ServiceLoaderFeature.html

Of the two experimental options mentioned in the PR comment linked above, only `-H:-UseServiceLoaderFeature` is recognized, and it breaks `ServiceLoader` completely:

- if we don't have the `META-INF/services/com.mongodb.spi.dns.InetAddressResolverProvider` resource entry, `ServiceLoader` silently fails to locate `CustomInetAddressResolverProvider` and `CustomInetAddressResolverProvider.assertUsed()` fails 
- if we do have the `META-INF/services/com.mongodb.spi.dns.InetAddressResolverProvider`, the `ServiceLoader` fails with `java.util.ServiceConfigurationError: com.mongodb.spi.dns.InetAddressResolverProvider: Provider com.mongodb.internal.graalvm.CustomInetAddressResolverProvider not found` even if I re-collect the reachability metadata (it does not find anything new).

**P.S.** Before I found the above, I did and wrote the following:

> I had a **un**educated guess that maybe the entry is not needed because the `META-INF/services/com.mongodb.spi.dns.InetAddressResolverProvider` resource is in the same project as the one I am building, and if the resource were in a pre-built JAR that I depend on, then the resource entry would have been required.
> 
> Since you asked, I tested that guess: created a new Gradle module with the resource and `CustomInetAddressResolverProvider`, removed those from `:graalvm-native-image-app`, and added the `configuration:'archives'` dependency on the new module to `:graalvm-native-image-app`. The result is: the entry about `META-INF/services/com.mongodb.spi.dns.InetAddressResolverProvider` in `resource-config.json` is still not needed. Note that when there is no entry, the resource is actually not mentioned in the `registerResource` messages emitted by `native-image` when it builds, yet `CustomInetAddressResolverProvider` is still successfully located via `ServiceLoader`.
> 
> I don't have any other hypotheses. The behavior is really weird.

driver-core/src/main/resources/META-INF/native-image/org.mongodb/bson/native-image.properties
```diff
@@ -13,4 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-Args = --initialize-at-run-time=com.mongodb.UnixServerAddress,com.mongodb.internal.connection.SnappyCompressor,org.bson.types.ObjectId,com.mongodb.internal.connection.ClientMetadataHelper
+Args =\
+  --initialize-at-run-time=\
```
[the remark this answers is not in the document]
stIncMale: This was intentional (though I should have left a comment about the removal): `bson/src/main/resources/META-INF/native-image/org.mongodb/bson/native-image.properties` specifies it, and `driver-core/src/main/resources/META-INF/native-image/org.mongodb/bson/native-image.properties` was unnecessarily duplicating the entry.

graalvm-native-image-app/src/main/com/mongodb/internal/graalvm/DnsSpi.java
```diff
@@ -0,0 +1,39 @@
...
+package com.mongodb.internal.graalvm;
+
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+
+final class DnsSpi {
```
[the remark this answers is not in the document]
stIncMale: My plan is to use a custom `DnsClientProvider` in `DnsSpi` as part of JAVA-5219, and assert it is used similarly to what I've done in the current PR.

jyemin (approved):

PR #1354, opened 2024-03-28
Use `InetAddressResolverProvider` and add tests

stIncMale:
This is a backport of https://github.com/mongodb/mongo-java-driver/pull/1353.

The changes made to the `:graalvm-native-image-app` Gradle project were not included, neither were the API documentation changes made to `MongoClientSettings`, the rest of the changes was taken as is.

JAVA-5390

### DIFF
--- a/driver-core/src/main/com/mongodb/MongoClientSettings.java
+++ b/driver-core/src/main/com/mongodb/MongoClientSettings.java
@@ -184,10 +184,12 @@ public final class MongoClientSettings {
     }
 
     /**
-     * Gets the explicitly set {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
+     * Gets the {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
+     *
+     * <p>If set, it will be used to look up the {@link java.net.InetAddress} for each host, via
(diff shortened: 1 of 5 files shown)


driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -184,12 +184,10 @@ public DnsClient getDnsClient() {
     }
 
     /**
-     * Gets the {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
-     *
-     * <p>If set, it will be used to look up the {@link java.net.InetAddress} for each host, via
-     * {@link InetAddressResolver#lookupByName(String)}. Otherwise, {@link java.net.InetAddress#getAllByName(String)} will be used.
+     * Gets the explicitly set {@link InetAddressResolver} to use for looking up the {@link java.net.InetAddress} instances for each host.
```
jyemin, line 187: Just FYI, we generally don't update Javadoc in patch releases, because we don't republish the Javadoc for patch releases.  But it doesn't do any harm, so it's ok to leave this.

jyemin (approved):

driver-core/src/main/com/mongodb/MongoClientSettings.java
stIncMale: Right, thanks! It's still better to have the source code consistent with the published API documentation, so I rolled back the change.

PR #1356, opened 2024-04-01
Force unique clusterTime values by splitting insert into two commands

jyemin:
JAVA-5397


driver-core/src/test/functional/com/mongodb/internal/operation/ChangeStreamOperationSpecification.groovy
```diff
@@ -519,7 +519,13 @@ class ChangeStreamOperationSpecification extends OperationFunctionalSpecificatio
         def cursor = execute(operation, async)
 
         when:
-        def expected = insertDocuments(helper, [1, 2])
+        // split into two insert command, because starting in MongoDB 8.0 the same clusterTime is applied to all documents in a bulk
```
katcharov, line 522:
```suggestion
        // split into two insert commands, because starting in MongoDB 8.0 the same clusterTime is applied to all documents in a bulk
```

katcharov (approved):
LGTM!

rozza (approved):

PR #1360, opened 2024-04-05
CSOT error transformation

vbabanin:
JAVA-5248

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java
@@ -32,6 +32,7 @@ import com.mongodb.event.CommandListener;
 import com.mongodb.event.CommandStartedEvent;
 import com.mongodb.event.CommandSucceededEvent;
 import com.mongodb.internal.IgnorableRequestContext;
+import com.mongodb.internal.TimeoutContext;
 import com.mongodb.internal.diagnostics.logging.Logger;
 import com.mongodb.internal.diagnostics.logging.Loggers;
 import com.mongodb.lang.Nullable;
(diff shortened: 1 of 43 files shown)


driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java
```diff
@@ -240,23 +245,27 @@ static MessageSettings getMessageSettings(final ConnectionDescription connection
...
+                                                        final TimeoutContext timeoutContext) {
         if (response == null) {
             return null;
         }
         int errorCode = getErrorCode(response);
         String errorMessage = getErrorMessage(response, errorMessageFieldName);
         if (ErrorCategory.fromErrorCode(errorCode) == ErrorCategory.EXECUTION_TIMEOUT) {
-            // TODO (CSOT) JAVA-5248 when timeoutMS is set, MongoOperationTimeoutException should be thrown.
+            if (timeoutContext.hasTimeoutMS()) {
+                return TimeoutContext.createMongoTimeoutException(errorMessage);
```
rozza, line 259: We lose information here - error code / message. Is that useful information if so should the `MongoExecutionTimeoutException` be wrapped?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
```diff
@@ -132,7 +135,7 @@ public Mono<BsonBinary> createDataKey(final String kmsProvider, final DataKeyOpt
                             .getDatabase(namespace.getDatabaseName())
                             .getCollection(namespace.getCollectionName(), BsonDocument.class)
                             .withWriteConcern(WriteConcern.MAJORITY);
-                    return Mono.from(collectionWithTimeout(vaultCollection, operationTimeout) //TODO should be defered?
+                    return Mono.from(collectionWithTimeout(vaultCollection, operationTimeout, TIMEOUT_ERROR_MESSAGE_CREATE_DATA_KEY)
```
rozza, line 138: Removing the TODO - means it didn't need to be a deferred mono?

rozza (changes requested):
Couple of questions. Also `MultiServerClusterSpecification.should_not_get_server_when_closed` needs reviewing as its failing on evergreen.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/vault/ClientEncryptionImpl.java
vbabanin: Yes, because` flatMap` triggers a subscription immediately after the Mono is returned from the lambda, the execution timing of the deferred operation remains effectively the same as if a regular `Mono` were returned.

driver-core/src/main/com/mongodb/internal/connection/ProtocolHelper.java
vbabanin: Indeed, retaining an error code enhances clarity, similar to how other exceptions include one. Accordingly, I've wrapped `MongoExecutionTimeoutException` exception with a `MongoOperationTimeoutException`.

Even though the spec mentions to "transform" old exception to new one. I believe we could still use it as a cause.

rozza (approved):
LGTM!

PR #1362, opened 2024-04-10
Exclude `:graalvm-native-image-app` from being published

stIncMale:
Tested by manually doing the following:

- `rm -r ~/.m2/repository/org/mongodb`
- `rm -r ./build/docs`
- `./gradlew clean docs publishToMavenLocal -x signMavenJavaPublication`
- - check that `./build/docs/bson` exists and `./build/docs/graalvm-native-image-app` does not.
- check that `~/.m2/repository/org/mongodb/bson` exists and `~/.m2/repository/org/mongodb/graalvm-native-image-app` does not.

JAVA-5414

### DIFF
--- a/gradle/javadoc.gradle
+++ b/gradle/javadoc.gradle
@@ -17,7 +17,7 @@ import static org.gradle.util.CollectionUtils.single
  */
 
 
-def projectsThatDoNotPublishJavaDocs = project(":util").allprojects + project(":driver-benchmarks") + project("driver-workload-executor") + project("driver-lambda")
+def projectsThatDoNotPublishJavaDocs = project(":util").allprojects + project(":driver-benchmarks") + project("driver-workload-executor") + project("driver-lambda") + project(":graalvm-native-image-app")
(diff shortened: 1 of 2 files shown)


jyemin (changes requested):
There is also this line in build.gradle:

```
def javaCodeCheckedProjects = javaMainProjects.findAll { !['driver-benchmarks', 'driver-workload-executor', 'driver-lambda'].contains(it.name) }
```

and this in gradle/javadoc.gradle:

```
def projectsThatDoNotPublishJavaDocs = project(":util").allprojects + project(":driver-benchmarks") + project("driver-workload-executor") + project("driver-lambda")
```

I think `graalvm-native-image-app` should be added in those places too.

stIncMale:
Thanks! I agree about `javaCodeCheckedProjects`, as all the javadocs get generated in the same place, not in `build` subdirectories of each project. However, I don't think `javaCodeCheckedProjects` needs to be updated - there is nothing wrong with `:graalvm-native-image-app` being statically checked. The checks still seem valuable, despite the code not being intended for production.

jyemin (approved):

PR #1367, opened 2024-04-17
Direct retries to another mongos if one is available

stIncMale:
Given that the current specification wording is internally inconsistent, I confirmed the actual intent with the spec author and created [DRIVERS-2901 Clarify the intent behind the list of deprioritized mongos'es and fix the pseudocode](https://jira.mongodb.org/browse/DRIVERS-2901).

### Performance considerations

Our `ServerSelector` and `ClusterDescription` API do not allow us to implement an efficient pipeline (`CompositeServerSelector`) of `ServerSelector`s: we can neither mutate the `List<ServerDescription>` in place, nor mutate `ClusterDescription`, nor even reuse the same `ClusterDescription` if a selector did not filter anything out. This PR added one more selector required by the specification logic, and introduced two more selectors by refactoring server selection code that wasn't expressed in terms of `ServerSelector`s. As a result, it is conceivable that the PR has negative performance impact.

Additionally, due to this refactoring https://github.com/mongodb/mongo-java-driver/pull/1367/commits/d25010d928704dcb7f29921f10f70505b4d45d9f, each server selection iteration now involves copying the map of `Server`s maintained by a `Cluster`. While that copying does not entail locking (all hail the CHM!), it may still have additional negative performance impact.

If we indeed notice performance degradation, we may try mitigating the issue by introducing `InternalServerSelector extends ServerSelector` (or a subclass of `ClusterDescription` that allows mutating `getServerDescriptions`, or both), which allows for a more optimal chaining, and use it for everything but the application-specific selector. This is assuming, of course, that the would be degradation is caused to a large extent by the inefficient selector chaining, and not by the CHM copying.

JAVA-4254, JAVA-5320


driver-core/src/main/com/mongodb/internal/operation/AsyncOperationHelper.java
```diff
@@ -294,7 +299,11 @@ static <R> AsyncCallbackSupplier<R> decorateReadWithRetriesAsync(final RetryStat
 
     static <R> AsyncCallbackSupplier<R> decorateWriteWithRetriesAsync(final RetryState retryState, final OperationContext operationContext,
             final AsyncCallbackSupplier<R> asyncWriteFunction) {
-        return new RetryingAsyncCallbackSupplier<>(retryState, CommandOperationHelper::chooseRetryableWriteException,
+        BiFunction<Throwable, Throwable, Throwable> onAttemptFailure =
```
vbabanin, line 302: [Optional] We could use `BinaryOperator<Throwable> onAttemptFailure` here. This modification would also necessitate changes to the constructor of `RetryingAsyncCallbackSupplier.`

P.S. The existing code prior to the changes in this PR should function without conflicts if we make this change in `RetryingAsyncCallbackSupplier.`

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -23,12 +34,87 @@
...
+    /**
+     * @return The same {@link ServerDeprioritization} if called on the same {@link OperationContext}.
+     */
+    public ServerDeprioritization getServerDeprioritization() {
+        return serverDeprioritization;
+    }
+
+    public static final class ServerDeprioritization {
+        @Nullable
+        private ServerAddress candidate;
```
vbabanin, line 57: It appears that the asynchronous version may switch threads three times, once in `ConnectionPool` (supported by 'happens-before' in `BlockingQueue`) and again in `AsyncSocketChannel`. Should we mark this field as volatile for safety, or are the existing guarantees sufficient?

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -23,12 +34,87 @@
...
+            return new DeprioritizingSelector(selector);
+        }
+
+        void updateCandidate(final ServerAddress serverAddress, final ClusterType clusterType) {
+            candidate = isEnabled(clusterType) ? serverAddress : null;
+        }
+
+        public void onAttemptFailure(final Throwable failure) {
+            if (!(failure instanceof MongoConnectionPoolClearedException)) {
+                deprioritized.add(candidate);
```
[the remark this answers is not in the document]
stIncMale: Thanks! There was actually a bug (though, not observable): `candidate` is allowed to be `null` at this point, but `onAttemptFailure` must not add it to `deprioritized`. Fortunately for us, `HashSet` allows `null` values, so it was fine. Nevertheless, I [fixed the method and added a test](https://github.com/mongodb/mongo-java-driver/pull/1367/commits/7cba88b2fab9027d7eab35ed8dfdaddaf4d5a868), though the test passes both before and after the change, again, because of `HashSet`.

driver-core/src/main/com/mongodb/internal/operation/AsyncOperationHelper.java
stIncMale: At first, I made a similar change in `master`: https://github.com/mongodb/mongo-java-driver/pull/1374. Then I [made changes as you proposed](https://github.com/mongodb/mongo-java-driver/pull/1367/commits/e4ffab42ff04e943bd2b9fcdf2689ec3d499b7ad) in the PR at hand.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: `ServerDeprioritization` is operation-confined, and an operation is processed sequentially even by the reactive API, regardless of whether multiple threads are involved or not. That "sequential" part must implies the happens-before order between the write and read action pairs that transfer either the whole `ServerDeprioritization` state or part of it across threads.

I think, `TimeoutContext` in the CSOT branch does not use any memory synchronization for its mutable state for the same reason.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -343,14 +350,13 @@ private static List<ServerTuple> atMostNRandom(final ArrayList<ServerDescription
         return result;
     }
 
-    private ServerSelector getCompositeServerSelector(final ServerSelector serverSelector) {
+    private ServerSelector getCompleteServerSelector(final ServerSelector serverSelector, final ServerDeprioritization serverDeprioritization) {
```
katcharov, line 353: Removing the subclass of the return type from the method name is fine, but it seems "Complete" has no implication except "Actual" (there is no incomplete selector in this class?). Let's use `getServerSelector`.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -343,14 +350,13 @@ private static List<ServerTuple> atMostNRandom(final ArrayList<ServerDescription
...
                 new LatencyMinimizingServerSelector(settings.getLocalThreshold(MILLISECONDS), MILLISECONDS);
-        if (settings.getServerSelector() == null) {
-            return new CompositeServerSelector(asList(serverSelector, latencyMinimizingServerSelector));
-        } else {
-            return new CompositeServerSelector(asList(serverSelector, settings.getServerSelector(), latencyMinimizingServerSelector));
-        }
+        CompositeServerSelector compositeSelector = settings.getServerSelector() == null
+                ? new CompositeServerSelector(asList(serverSelector, latencyMinimizingServerSelector))
+                : new CompositeServerSelector(asList(serverSelector, settings.getServerSelector(), latencyMinimizingServerSelector));
+        return serverDeprioritization.apply(compositeSelector);
```
katcharov, line 359: May as well remove the rest of the duplication:

```
List<ServerSelector> selectors = Stream.of(
        serverSelector,
        settings.getServerSelector(), // can be null
        new LatencyMinimizingServerSelector(settings.getLocalThreshold(MILLISECONDS), MILLISECONDS)
).filter(v -> v != null).collect(Collectors.toList());
return serverDeprioritization.apply(new CompositeServerSelector(selectors));
```

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -23,12 +34,87 @@
...
+        private ServerAddress candidate;
+        private final Set<ServerAddress> deprioritized;
+
+        private ServerDeprioritization() {
+            candidate = null;
+            deprioritized = new HashSet<>();
+        }
+
+        ServerSelector apply(final ServerSelector selector) {
+            return new DeprioritizingSelector(selector);
```
katcharov, line 66: This `apply` naming is confusing: I expected it to perhaps apply certain settings to the selector, but actually it wraps in another selector.

It seems the purpose of CompositeServerSelector to handle such "chaining", so I would expect a `DeprioritizingSelector` to be another entry in the list in `getCompositeServerSelector`.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -23,12 +34,87 @@
...
+        }
+
+        void updateCandidate(final ServerAddress serverAddress, final ClusterType clusterType) {
+            candidate = isEnabled(clusterType) ? serverAddress : null;
```
katcharov, line 70: We should put all these conditions into the same place, `onAttemptFailure`.

Alternatively, we could skip adding a DeprioritizingSelector based on this check. (Whatever we do, I would expect disabling deprioritization in this way to just work.)

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -23,12 +34,87 @@
...
+            @Override
+            public List<ServerDescription> select(final ClusterDescription clusterDescription) {
+                if (isEnabled(clusterDescription.getType())) {
+                    List<ServerDescription> filteredServerDescriptions = ClusterDescriptionHelper.getServersByPredicate(
+                            clusterDescription, serverDescription -> !deprioritized.contains(serverDescription.getAddress()));
+                    ClusterDescription filteredClusterDescription = new ClusterDescription(
+                            clusterDescription.getConnectionMode(),
+                            clusterDescription.getType(),
+                            clusterDescription.getSrvResolutionException(),
+                            filteredServerDescriptions,
```
katcharov, line 105: Let's add a `with...` method.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -23,12 +34,87 @@
...
+            private final ServerSelector wrapped;
+
+            private DeprioritizingSelector(final ServerSelector wrapped) {
+                this.wrapped = wrapped;
+            }
+
+            @Override
+            public List<ServerDescription> select(final ClusterDescription clusterDescription) {
+                if (isEnabled(clusterDescription.getType())) {
+                    List<ServerDescription> filteredServerDescriptions = ClusterDescriptionHelper.getServersByPredicate(
```
katcharov, line 99: `filteredServerDescriptions` -> `notDeprioritized`?

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -23,12 +34,87 @@
...
+        private final class DeprioritizingSelector implements ServerSelector {
+            private final ServerSelector wrapped;
+
+            private DeprioritizingSelector(final ServerSelector wrapped) {
+                this.wrapped = wrapped;
+            }
+
+            @Override
+            public List<ServerDescription> select(final ClusterDescription clusterDescription) {
+                if (isEnabled(clusterDescription.getType())) {
```
katcharov, line 98: Is this check needed here for correctness, and is there a possible noticeable effect on performance?

This check would be clearer if inverted:
```
if (!isEnabled(clusterDescription.getType())) {
    return wrapped.select(clusterDescription);
}
// followed by the non-indented deprioritization logic
```

driver-core/src/main/com/mongodb/internal/operation/AsyncOperationHelper.java
```diff
@@ -285,7 +286,11 @@ static <D, T> void createReadCommandAndExecuteAsync(
 
     static <R> AsyncCallbackSupplier<R> decorateReadWithRetriesAsync(final RetryState retryState, final OperationContext operationContext,
             final AsyncCallbackSupplier<R> asyncReadFunction) {
-        return new RetryingAsyncCallbackSupplier<>(retryState, CommandOperationHelper::chooseRetryableReadException,
+        BinaryOperator<Throwable> onAttemptFailure =
+                (@Nullable Throwable previouslyChosenException, Throwable mostRecentAttemptException) ->
+                        CommandOperationHelper.onRetryableReadAttemptFailure(
+                                operationContext, previouslyChosenException, mostRecentAttemptException);
+        return new RetryingAsyncCallbackSupplier<>(retryState, onAttemptFailure,
```
katcharov, line 293: I find it difficult to understand what is happening in these methods, though I am unsure why. I think part of it is the amount of boilerplate: `onAttemptFailure = (priorException, currentException) ->`, and then inlining, might help.

The other, more objective issue is that `onAttemptFailure` is being used as a `failedResultTransformer`, but I would expect such a thing to have no side effects (such as deprioritizing a server in OperationContext).

katcharov (changes requested):

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
stIncMale: > there is no incomplete selector in this class?

There is - the `serverSelector` passed to the `getCompleteServerSelector` method.

vbabanin (approved):
LGTM!

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: > It seems the purpose of CompositeServerSelector to handle such "chaining"

`CompositeServerSelector` just applies one selector after the other, and that's not the logic needed for the new functionality.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: `ClusterDescription` is part of the API, and so far each time we needed more fields in such a class, we have been adding a new constructor, sometimes deprecating the previous constructors. One can certainly propose us to change that API approach not for just this class, but for all similar situations, but that warrants a team discussion. Since this PR does not add a new constructor to `ClusterDescription`, and only uses an existing constructor, it is not blocked by the decision on such a change, were it to be proposed.

katcharov: This can be a private static method.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: When we discussed this, we found that `binding.getWriteConnectionSource().getServerDescription().getClusterType()` can be used to trivially get `ClusterType` that can be passed into `ServerDeprioritization.onAttemptFailure`. However, after looking a bit deeper, I noticed that `binding.getWriteConnectionSource()` actually does server selection, so we really can't call this method when we create `BinaryOperator<Throwable> onAttemptFailure` in `decorateWriteWithRetries`.

We still may attach `ClusterType` from `binding to `RetryState` when we currently call `binding.getWriteConnectionSource()`, and then access it from the `BinaryOperator<Throwable> onAttemptFailure` object via the `RetryState`, but then the change becomes not trivial at all.

Thus, I don't see a trivial way to pass `ClusterType` into `ServerDeprioritization.onAttemptFailure`. The solution I see is not worth the potential simplification of moving the `isEnabled(clusterType)` check from `ServerDeprioritization.updateCandidate` to `ServerDeprioritization.onAttemptFailure`.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
stIncMale: [Done](https://github.com/mongodb/mongo-java-driver/pull/1367/commits/6574e241434f8c8946984ade559695e2aa9f29b3)

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: [Done](https://github.com/mongodb/mongo-java-driver/pull/1367/commits/6574e241434f8c8946984ade559695e2aa9f29b3), only in Java SE 8 we are not allowed to have `static` methods in inner classes. This is an artificial limitation (I guess, we are not the only ones who don't like to complicate things), which was removed in Java SE 16 (see https://openjdk.org/jeps/395#Static-members-of-inner-classes, https://bugs.openjdk.org/browse/JDK-8254321).

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: [Done](https://github.com/mongodb/mongo-java-driver/pull/1367/commits/6574e241434f8c8946984ade559695e2aa9f29b3)

driver-core/src/main/com/mongodb/internal/operation/AsyncOperationHelper.java
stIncMale: I [updated the docs](https://github.com/mongodb/mongo-java-driver/pull/1367/commits/171d67ed767a65156a768c9667488b13d11049eb) to reflect that side effects are allowed. The key to them being allowed is that the operator is guaranteed to be called once per failed attempt.

> I find it difficult to understand what is happening in these methods...
> ...inlining, might help...

Could you please clarify what you propose to inline and where?

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: I thought more about the possibility to implement server selection via a `DeprioritizingSelector` that is composed using `CompositeServerSelector` either before or after `settings.getServerSelector()` (regardless of whether that order is immediate). As far as I can see, that approach can't work: see counterexamples below.

`DeprioritizingSelector.select(clusterDescription)` may act in one of the three ways:

1. Filter: return a `List<ServerDescription>` that does not contain deprioritized servers, provided that there are normal servers; otherwise return everything.
2. Mark: return a `List<ServerDescription>` that contains essentially the same elements as `clusterDescription.getServerDescriptions`, with the only difference that some `ServerDescription`s are marked with the priority. This requires changing the API of the `ServerDescription`. This fact makes this approach unfavourable.
3. Sort: return a `List<ServerDescription>` that contains the same elements as `clusterDescription.getServerDescriptions`, but ordered according to the priority. Unfortunately, the order of elements in `ServerDescription.getServerDescriptions` does not seem to have any meaning, as `ServerDescription.equals` treats two `ServerDescription`s with different orders of elements in `getServerDescriptions` as equal. This means, we should not rely on the order.

Now let's consider all possibilities

# `DeprioritizingSelector` is before `settings.getServerSelector()`

## 1. Filter

- Consider `settings.getServerSelector()` that selects only servers with the `t` tag.
- `clusterDescription.getServerDescriptions` contains descriptions with server addresses `a`, `b`, only `b` has the `t` tag, only `b` is deprioritized.
- `DeprioritizingSelector` filters `b` out, and `settings.getServerSelector()` has only `a` to select from.
- `settings.getServerSelector()` selects nothing, as `a` does not have the `t` tag.
- According to the spec, server selection should have also considered `b` and selected despite it being deprioritized, because without considering this server, selection fails. However, there is no way to consider `b` in this scenario, and we failed the server selection.

## 2. Mark

- Consider `settings.getServerSelector()` that selects one server at random.
- `clusterDescription.getServerDescriptions` contains descriptions with server addresses `a`, `b`, only `b` is deprioritized.
- `DeprioritizingSelector` marks `b`'s `ServerDescription` as deprioritized, but `settings.getServerSelector()` knows nothing about such marks, and operates as if nothing were marked.
- `settings.getServerSelector()` selects `b` at random, despite it being deprioritized.
- We should have selected `a`, but we incorrectly selected the deprioritized server `b`.

## 3. Sort

The counterexample here is essentially identical to "2. Mark", as `settings.getServerSelector()` knows nothing about the order of `clusterDescription.getServerDescriptions` representing priorities. Actually, even if it new about that, it still could not have worked correctly all the time, as it needs to be able to clearly discriminate between deprioritized and normal servers when selecting a server at random.

# `settings.getServerSelector()` is before `DeprioritizingSelector`

## 1. Filter

- Consider `settings.getServerSelector()` that selects one server at random.
- `clusterDescription.getServerDescriptions` contains descriptions with server addresses `a`, `b`, only `b` is deprioritized.
- `settings.getServerSelector()` selects `b` at random, despite it being deprioritized.
- `DeprioritizingSelector` selects `b` as it has no normal servers to select.
- We should have selected `a`, but we incorrectly selected the deprioritized server `b`.

## 2. Mark

The counterexample here is essentially identical to "1. Filter", as `DeprioritizingSelector` can't do anything for `a` to be selected.

## 3. Sort

The counterexample here is essentially identical to "1. Filter", as `DeprioritizingSelector` can't do anything for `a` to be selected.

P.S. All of the counterexamples above work with the original approach taken in this PR. However, when we talked, we also identified a bug when combining this approach with the way `BaseCluster.selectServer(ServerSelector, ClusterDescription, Function<ServerAddress, Server> serverCatalog)` deals with the internal race condition (that code was introduced in https://github.com/mongodb/mongo-java-driver/commit/f19447753657d7be3a597e31731f5aedf0cc9ccd). That can be dealt with by addressing the internal race condition differently: instead of calling `getServer` and checking if the result is `null`, we can take a non-linearizable snapshot of the `Map` that `getServer` uses internally, filter the `ClusterDescription` using that snapshot, and forget about any troubles.

katcharov: The [spec](https://github.com/mongodb/specifications/blob/d3a355c299d9840bf9a758910a1739fcbc33ec64/source/server-selection/server-selection.md?plain=1#L673) provides an algorithm in which deprioritization (filtering out of deprioritized servers, when others are available) occurs in step 3 and completes before step 4 (our `settings.getServerSelector`). This PR does not adhere to the spec: it implements a different algorithm, which "wraps" steps 3-7 and repeats them when no servers are found (for any reason).

I believe this is a spec bug. If the existing latency selector comes after the user selector, then so should the deprioritizing selector. "Optimizing" selectors must be applied after any selector expressing necessary conditions, because all requirements must be met before something is selected due to preference.

The spec should be changed (to correct this, and for clarity), to list the following sequence of composed selectors:
```
serverSelector
settings.getServerSelector (ask user)
filter by latency
randomize
deprioritize stable sort
2 first
pick 1 by least ops
```
I see no issue with relying on sorting/ordering in our internal code (that is, all selectors that follow the user-provided selector).

stIncMale: Implemented the deprioritization logic according to the spec in https://github.com/mongodb/mongo-java-driver/pull/1367/commits/b3430bd0d60838fdeff00ef682857279b56d0747.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -15,20 +15,103 @@
...
+         * The returned {@link ServerSelector} tries to {@linkplain ServerSelector#select(ClusterDescription) select}
+         * only the {@link ServerDescription}s that do not have deprioritized {@link ServerAddress}es.
+         * If no such {@link ServerDescription} can be selected, then it selects {@link ClusterDescription#getServerDescriptions()}.
+         */
+        ServerSelector getServerSelector() {
+            return selector;
+        }
+
+        void updateCandidate(final ServerAddress serverAddress, final ClusterType clusterType) {
+            candidate = isEnabled(clusterType) ? serverAddress : null;
```
katcharov, line 78:
```suggestion
            candidate = serverAddress;
```
I do not think the check is needed here. This potential "server on which the operation failed" will already be correctly excluded, because the `ServerDeprioritization` server selector is already a no-op when not sharded.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -15,20 +15,103 @@
...
+                List<ServerDescription> serverDescriptions = clusterDescription.getServerDescriptions();
+                if (!isEnabled(clusterDescription.getType())) {
+                    return serverDescriptions;
+                } else {
+                    List<ServerDescription> nonDeprioritizedServerDescriptions = serverDescriptions
+                            .stream()
+                            .filter(serverDescription -> !deprioritized.contains(serverDescription.getAddress()))
+                            .collect(toList());
+                    return nonDeprioritizedServerDescriptions.isEmpty() ? serverDescriptions : nonDeprioritizedServerDescriptions;
+                }
```
katcharov, line 113:
```suggestion
                List<ServerDescription> serverDescriptions = clusterDescription.getServerDescriptions();
                if (!isEnabled(clusterDescription.getType())) {
                    return serverDescriptions;
                }
                List<ServerDescription> nonDeprioritizedServerDescriptions = serverDescriptions
                        .stream()
                        .filter(serverDescription -> !deprioritized.contains(serverDescription.getAddress()))
                        .collect(toList());
                return nonDeprioritizedServerDescriptions.isEmpty() 
                        ? serverDescriptions : nonDeprioritizedServerDescriptions;
```

driver-core/src/main/com/mongodb/internal/operation/AsyncOperationHelper.java
katcharov: I am imagining at least something like:
```
    static <R> AsyncCallbackSupplier<R> decorateReadWithRetriesAsync(final RetryState retryState, 
            final OperationContext operationContext, final AsyncCallbackSupplier<R> asyncReadFunction) {
        return new RetryingAsyncCallbackSupplier<>(retryState,
                onRetryableReadAttemptFailure(operationContext),
                CommandOperationHelper::shouldAttemptToRetryRead,
                logRetryAndGet(retryState, operationContext, asyncReadFunction));
    }
```
This is optional for this PR, and perhaps should be undertaken separately.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -302,55 +309,64 @@ private boolean handleServerSelectionRequest(final ServerSelectionRequest reques
...
+                clusterDescriptionPotentiallyInconsistentWithServerSnapshot.getServerDescriptions()
+                        .stream()
+                        .filter(serverDescription -> serversSnapshot.containsServer(serverDescription.getAddress()))
+                        .collect(toList());
+        List<ServerSelector> selectors = Stream.of(
+                raceConditionPreFiltering,
+                serverSelector,
+                serverDeprioritization.getServerSelector(),
+                settings.getServerSelector(), // may be null
+                new LatencyMinimizingServerSelector(settings.getLocalThreshold(MILLISECONDS), MILLISECONDS),
```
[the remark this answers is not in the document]
stIncMale: This PR hasn't introduced this constructor, i.e., the instance is created exactly as it was created before the PR. I don't see a reason to make the change in this PR.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -302,55 +309,64 @@ private boolean handleServerSelectionRequest(final ServerSelectionRequest reques
...
+        // To deal with this race condition, we take `serversSnapshot` of that set of `Server`s
+        // (the snapshot itself does not have to be atomic) non-atomically with reading `clusterDescription`
+        // (this means, `serversSnapshot` and `clusterDescription` are not guaranteed to be consistent with each other),
+        // and do pre-filtering to make sure that the only `ServerDescription`s we may select,
+        // are of those `Server`s that are known to both `clusterDescription` and `serversSnapshot`.
+        // This way we are guaranteed to successfully get `Server`s from `serversSnapshot` based on the selected `ServerDescription`s.
+        //
+        // The pre-filtering we do to deal with the race condition described above is achieved by this `ServerSelector`.
+        ServerSelector raceConditionPreFiltering = clusterDescriptionPotentiallyInconsistentWithServerSnapshot ->
+                clusterDescriptionPotentiallyInconsistentWithServerSnapshot.getServerDescriptions()
```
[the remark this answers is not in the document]
stIncMale: We have a local variable here that has a long name explaining what it is, and the variable is used only once. Why do you think that the generic `clusterDescription` name is better here?

driver-core/src/main/com/mongodb/internal/selector/OperationCountMinimizingServerSelector.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public List<ServerDescription> select(final ClusterDescription clusterDescription) {
+        ServerDescription selected = clusterDescription.getServerDescriptions()
+                .stream()
+                .min(comparingInt(serverDescription ->
+                        assertNotNull(serversSnapshot.getServer(serverDescription.getAddress()))
+                                .operationCount()))
+                .orElse(null);
+        return selected == null ? emptyList() : singletonList(selected);
```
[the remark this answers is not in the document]
stIncMale: I don't know how to collect an `Optional` into a `List` in Java SE 8. Do you have a way to do that in mind?

driver-core/src/main/com/mongodb/internal/connection/LoadBalancedCluster.java
```diff
@@ -181,10 +181,11 @@ public ClusterId getClusterId() {
     }
 
     @Override
-    public ClusterableServer getServer(final ServerAddress serverAddress) {
+    public ServersSnapshot getServersSnapshot() {
         isTrue("open", !isClosed());
         waitForSrv();
-        return assertNotNull(server);
+        ClusterableServer server = assertNotNull(this.server);
+        return serverAddress -> server;
```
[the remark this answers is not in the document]
stIncMale: > Should it not be an assertion failure if an incorrect address is provided?

Maybe it should, but I have maintained the behavior as it was. If we want to add an assertion there, let's do that as a different activity.

driver-core/src/main/com/mongodb/internal/connection/Cluster.java
```diff
@@ -89,4 +85,17 @@ void selectServerAsync(ServerSelector serverSelector, OperationContext operation
      * Server Discovery And Monitoring</a> specification.
      */
     void onChange(ServerDescriptionChangedEvent event);
+
+    /**
+     * A non-atomic snapshot of the servers in a {@link Cluster}.
+     */
+    @ThreadSafe
+    interface ServersSnapshot {
```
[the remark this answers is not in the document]
stIncMale: Implementing `ServersSnapshot` the way it's done now, allows to see clearly in the diff how the transition from `getServer` to `getServersSnapshot` did not change the behavior of getting a server.

We could have two different implementations: one containing a `Map`, and one containing a `Server`, but I can't see myself why this approach would be objectively better than the current one.

> the hashmap is being copied outside of the lambda - so I don't see how we can guarantee this is ThreadSafe

Could you please explain in more details what problem with the current implementation you see?

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: Done in https://github.com/mongodb/mongo-java-driver/pull/1367/commits/246353f36dae51e28aa212488444ec54cb8ad5af.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: After thinking more about this, I agree that it's fine to omit the check in the `updateCandidate` method.

Done in https://github.com/mongodb/mongo-java-driver/pull/1367/commits/246353f36dae51e28aa212488444ec54cb8ad5af.

driver-core/src/main/com/mongodb/internal/selector/AtMostTwoRandomServerSelector.java
```diff
@@ -0,0 +1,57 @@
...
+    public static AtMostTwoRandomServerSelector instance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public List<ServerDescription> select(final ClusterDescription clusterDescription) {
+        List<ServerDescription> serverDescriptions = new ArrayList<>(clusterDescription.getServerDescriptions());
+        List<ServerDescription> result = new ArrayList<>();
+        ThreadLocalRandom random = ThreadLocalRandom.current();
+        for (int i = serverDescriptions.size() - 1; i >= 0 && result.size() < TWO; i--) {
```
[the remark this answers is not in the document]
stIncMale: Done in https://github.com/mongodb/mongo-java-driver/pull/1367/commits/246353f36dae51e28aa212488444ec54cb8ad5af.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -302,55 +309,64 @@ private boolean handleServerSelectionRequest(final ServerSelectionRequest reques
...
+        // This race condition means that we are not guaranteed to observe all the servers from `clusterDescription`
+        // among the `Server`s maintained by the `Cluster`.
+        // To deal with this race condition, we take `serversSnapshot` of that set of `Server`s
+        // (the snapshot itself does not have to be atomic) non-atomically with reading `clusterDescription`
+        // (this means, `serversSnapshot` and `clusterDescription` are not guaranteed to be consistent with each other),
+        // and do pre-filtering to make sure that the only `ServerDescription`s we may select,
+        // are of those `Server`s that are known to both `clusterDescription` and `serversSnapshot`.
+        // This way we are guaranteed to successfully get `Server`s from `serversSnapshot` based on the selected `ServerDescription`s.
+        //
+        // The pre-filtering we do to deal with the race condition described above is achieved by this `ServerSelector`.
```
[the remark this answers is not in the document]
stIncMale: I think this comment should be in the code, not somewhere else, because it explains the code, that is not that obvious otherwise.

> but now that there is a solution

Note that it has been solved previously, i.e., it is solved in the `master`, just differently.

> all the facts mentioned in this comment seem like they should be evident to unfamiliar readers

This is a huge overstatement in my opinion.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -302,55 +309,64 @@ private boolean handleServerSelectionRequest(final ServerSelectionRequest reques
...
+        ServerSelector raceConditionPreFiltering = clusterDescriptionPotentiallyInconsistentWithServerSnapshot ->
+                clusterDescriptionPotentiallyInconsistentWithServerSnapshot.getServerDescriptions()
+                        .stream()
+                        .filter(serverDescription -> serversSnapshot.containsServer(serverDescription.getAddress()))
+                        .collect(toList());
+        List<ServerSelector> selectors = Stream.of(
+                raceConditionPreFiltering,
```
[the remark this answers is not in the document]
stIncMale: If `getCompleteServerSelector` gets too big in the future, that may be helpful. For now, however, it does not seem to be needed.

driver-core/src/main/com/mongodb/internal/operation/AsyncOperationHelper.java
stIncMale: It looks like a code extraction into a method, rather than inlining. Done in https://github.com/mongodb/mongo-java-driver/pull/1367/commits/69b78ffad40e07040c5f6c8c7e1b9d9e549bdc25.

driver-core/src/main/com/mongodb/internal/selector/OperationCountMinimizingServerSelector.java
```diff
@@ -0,0 +1,62 @@
...
+import static java.util.Comparator.comparingInt;
+
+/**
+ * {@linkplain #select(ClusterDescription) Selects} at most one {@link ServerDescription}
+ * corresponding to a {@link ServersSnapshot#getServer(ServerAddress) server} with the smallest {@link Server#operationCount()}.
+ *
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+@ThreadSafe
+public final class OperationCountMinimizingServerSelector implements ServerSelector {
```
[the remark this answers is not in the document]
stIncMale: Done in https://github.com/mongodb/mongo-java-driver/pull/1367/commits/246353f36dae51e28aa212488444ec54cb8ad5af.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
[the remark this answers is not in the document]
stIncMale: > there should be a test that ensures items missing from the snapshot get filtered out by this chain

Done in https://github.com/mongodb/mongo-java-driver/pull/1367/commits/246353f36dae51e28aa212488444ec54cb8ad5af.

driver-core/src/main/com/mongodb/internal/selector/OperationCountMinimizingServerSelector.java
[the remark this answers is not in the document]
katcharov: Ah I see. How about:
```
                .map(selected -> singletonList(selected))
                .orElse(emptyList());
```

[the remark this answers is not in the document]
stIncMale: Done in https://github.com/mongodb/mongo-java-driver/pull/1367/commits/05f6679e29c9826bb4fa1d9c6965c0a2ff03a8f0.

driver-core/src/main/com/mongodb/internal/connection/Cluster.java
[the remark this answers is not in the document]
stIncMale: - We talked over Zoom - there is no problem with the current implementation. There was still an idea to have a single implementation of `ServersSnapshot` that contains a `Map<ServerAddress, Server>`. However, such an implementation turned out to be impossible: `LoadBalancedCluster.server` does not have the `ServerAddress` information, which means, the method `LoadBalancedCluster.getServersSnapshot` can't create a `Map<ServerAddress, Server>`. When I agreed [here](https://github.com/mongodb/mongo-java-driver/pull/1367#discussion_r1594765696) that an assertion would have made sense, I didn't realize that the `ServerAddress` information is not available via `ClusterableServer`.

- There was also a concern that someone may in the future accidentally start using `addressToServerTupleMap` from the lambda expression that is currently in the method `AbstractMultiServerCluster.getServersSnapshot`, which will be incorrect. But the same statement is valid about any place where we write a lambda expression that can technically capture and use state it is not supposed to use; it is also valid about instance method that may technically use the instance state it is not supposed to use.

vbabanin (approved):
LGTM!

driver-core/src/main/com/mongodb/internal/connection/Cluster.java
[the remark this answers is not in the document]
katcharov: That is too bad about the assertion. The main impediment I see now is that ServerTuple is private, and I see no clean way around that. How about the following change to docs?

> Provides access to a (non-atomic) snapshot of the servers in a {@link Cluster}, mapped by their address.

katcharov (approved):
LGTM, just one optional suggestion

driver-core/src/main/com/mongodb/internal/connection/Cluster.java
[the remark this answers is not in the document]
stIncMale: I think, you advocated that we should attempt to document program elements that declare types as the things they are, as opposed to what they represent/provide/etc. For example, _"A boolean value ..."_ for `MqlBoolean` as opposed to "Represents a boolean value ...". This may not be always the most suitable way, but for `ServersSnapshot`, documenting it as _"A non-atomic snapshot ..."_ instead of as _"Provides access to a (non-atomic) snapshot ..."_ seems to be as OK as the above example. I can't on my own see why the proposed change makes the wording better. Could you please explain?

PR #1371, opened 2024-04-19
OIDC Add remaining environments (azure, gcp), evergreen testing, API naming updates

katcharov:
JAVA-5353 API naming
JAVA-5395 Atlas testing
JAVA-4834 Azure
JAVA-4932 GCP

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -12,9 +12,8 @@ stepback: true
 # Actual testing tasks are marked with `type: test`
 command_type: system
 
-# Protect ourself against rogue test case, or curl gone wild, that runs forever
-# 12 minutes is the longest we'll ever run
-exec_timeout_secs: 3600 # 12 minutes is the longest we'll ever run
+# Protect ourselves against rogue test case, or curl gone wild, that runs forever
+exec_timeout_secs: 7200
 
 # What to do when evergreen hits the timeout (`post:` tasks are run automatically)
 timeout:
@@ -967,6 +966,58 @@ tasks:
             LOAD_BALANCER: 'true'
         - func: "run load-balancer"
         - func: "run load-balancer tests"
+
+    - name: "oidc-auth-test-latest"
+      commands:
+        - command: subprocess.exec
+          type: test
+          params:
+            working_dir: "src"
+            binary: bash
+            include_expansions_in_env: ["DRIVERS_TOOLS", "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN"]
+            args:
+              - .evergreen/run-mongodb-oidc-test.sh
+
+    - name: "oidc-auth-test-azure-latest"
+      commands:
+        - command: shell.exec
+          params:
+            shell: bash
+            env:
+              JAVA_HOME: ${JAVA_HOME}
+            script: |-
+              set -o errexit
+              ${PREPARE_SHELL}
+              cd src
+              git add .
+              git commit -m "add files"
+              # uncompressed tar used to allow appending .git folder
+              export AZUREOIDC_DRIVERS_TAR_FILE=/tmp/mongo-java-driver.tar
+              git archive -o $AZUREOIDC_DRIVERS_TAR_FILE HEAD
+              tar -rf $AZUREOIDC_DRIVERS_TAR_FILE .git
+              export AZUREOIDC_TEST_CMD="OIDC_ENV=azure ./.evergreen/run-mongodb-oidc-test.sh"
+              bash $DRIVERS_TOOLS/.evergreen/auth_oidc/azure/run-driver-test.sh
+
+    - name: "oidc-auth-test-gcp-latest"
+      commands:
+        - command: shell.exec
+          params:
+            shell: bash
+            script: |-
+              set -o errexit
+              ${PREPARE_SHELL}
+              cd src
+              git add .
+              git commit -m "add files"
+              # uncompressed tar used to allow appending .git folder
+              export GCPOIDC_DRIVERS_TAR_FILE=/tmp/mongo-java-driver.tar
+              git archive -o $GCPOIDC_DRIVERS_TAR_FILE HEAD
+              tar -rf $GCPOIDC_DRIVERS_TAR_FILE .git
+              # Define the command to run on the VM.
+              # Ensure that we source the environment file created for us, set up any other variables we need,
+              # and then run our test suite on the vm.
+              export GCPOIDC_TEST_CMD="OIDC_ENV=gcp ./.evergreen/run-mongodb-oidc-test.sh"
+              bash $DRIVERS_TOOLS/.evergreen/auth_oidc/gcp/run-driver-test.sh
 
     - name: serverless-test
       commands:
@@ -2065,6 +2116,77 @@ task_groups:
     tasks:
       - test-aws-lambda-deployed
 
+  - name: testoidc_task_group
+    setup_group:
+      - func: fetch source
+      - func: prepare resources
+      - func: fix absolute paths
+      - command: ec2.assume_role
+        params:
+          role_arn: ${aws_test_secrets_role}
+      - command: subprocess.exec
+        params:
+          binary: bash
+          include_expansions_in_env: ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN"]
+          args:
+            - ${DRIVERS_TOOLS}/.evergreen/auth_oidc/setup.sh
+    teardown_task:
+      - command: subprocess.exec
+        params:
+          binary: bash
+          args:
+            - ${DRIVERS_TOOLS}/.evergreen/auth_oidc/teardown.sh
+    setup_group_can_fail_task: true
+    setup_group_timeout_secs: 1800
+    tasks:
+      - oidc-auth-test-latest
+
+  - name: testazureoidc_task_group
+    setup_group:
+      - func: fetch source
+      - func: prepare resources
+      - func: fix absolute paths
+      - command: subprocess.exec
+        params:
+          binary: bash
+          env:
+            AZUREOIDC_VMNAME_PREFIX: "JAVA_DRIVER"
+          args:
+            - ${DRIVERS_TOOLS}/.evergreen/auth_oidc/azure/create-and-setup-vm.sh
+    teardown_task:
+      - command: subprocess.exec
+        params:
+          binary: bash
+          args:
+            - ${DRIVERS_TOOLS}/.evergreen/auth_oidc/azure/delete-vm.sh
+    setup_group_can_fail_task: true
+    setup_group_timeout_secs: 1800
+    tasks:
+      - oidc-auth-test-azure-latest
+
+  - name: testgcpoidc_task_group
+    setup_group:
+      - func: fetch source
+      - func: prepare resources
+      - func: fix absolute paths
+      - command: subprocess.exec
+        params:
+          binary: bash
+          env:
+            GCPOIDC_VMNAME_PREFIX: "JAVA_DRIVER"
+          args:
+            - ${DRIVERS_TOOLS}/.evergreen/auth_oidc/gcp/setup.sh
+    teardown_task:
+      - command: subprocess.exec
+        params:
+          binary: bash
+          args:
+            - ${DRIVERS_TOOLS}/.evergreen/auth_oidc/gcp/teardown.sh
+    setup_group_can_fail_task: true
+    setup_group_timeout_secs: 1800
+    tasks:
+      - oidc-auth-test-gcp-latest
+
 buildvariants:
 
 # Test packaging and other release related routines
@@ -2215,6 +2337,27 @@ buildvariants:
   run_on: rhel80-small
   tasks:
     - name: "test_atlas_task_group_search_indexes"
+
+- name: "oidc-auth-test"
+  display_name: "OIDC Auth"
+  run_on: ubuntu2204-small
+  tasks:
+    - name: testoidc_task_group
+      batchtime: 20160 # 14 days
+
+- name: testazureoidc-variant
+  display_name: "OIDC Auth Azure"
+  run_on: ubuntu2204-small
+  tasks:
+    - name: testazureoidc_task_group
+      batchtime: 20160 # 14 days
+
+- name: testgcpoidc-variant
+  display_name: "OIDC Auth GCP"
+  run_on: ubuntu2204-small
+  tasks:
+    - name: testgcpoidc_task_group
+      batchtime: 20160 # 14 days
 
 - matrix_name: "aws-auth-test"
   matrix_spec: { ssl: "nossl", jdk: ["jdk8", "jdk17", "jdk21"], version: ["4.4", "5.0", "6.0", "7.0", "latest"], os: "ubuntu",
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -2174,6 +2174,7 @@ task_groups:
           binary: bash
           env:
             GCPOIDC_VMNAME_PREFIX: "JAVA_DRIVER"
+            GCPKMS_MACHINETYPE: "e2-medium" # comparable elapsed time to Azure; default was starved, caused timeouts
           args:
             - ${DRIVERS_TOOLS}/.evergreen/auth_oidc/gcp/setup.sh
     teardown_task:
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -13,7 +13,7 @@ stepback: true
 command_type: system
 
 # Protect ourselves against rogue test case, or curl gone wild, that runs forever
-exec_timeout_secs: 7200
+exec_timeout_secs: 3600
 
 # What to do when evergreen hits the timeout (`post:` tasks are run automatically)
 timeout:
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -967,7 +967,7 @@ tasks:
         - func: "run load-balancer"
         - func: "run load-balancer tests"
 
-    - name: "oidc-auth-test-latest"
+    - name: "oidc-auth-test"
       commands:
         - command: subprocess.exec
           type: test
@@ -975,10 +975,12 @@ tasks:
             working_dir: "src"
             binary: bash
             include_expansions_in_env: ["DRIVERS_TOOLS", "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN"]
+            env:
+              OIDC_ENV: "test"
             args:
               - .evergreen/run-mongodb-oidc-test.sh
 
-    - name: "oidc-auth-test-azure-latest"
+    - name: "oidc-auth-test-azure"
       commands:
         - command: shell.exec
           params:
@@ -998,7 +1000,7 @@ tasks:
               export AZUREOIDC_TEST_CMD="OIDC_ENV=azure ./.evergreen/run-mongodb-oidc-test.sh"
               bash $DRIVERS_TOOLS/.evergreen/auth_oidc/azure/run-driver-test.sh
 
-    - name: "oidc-auth-test-gcp-latest"
+    - name: "oidc-auth-test-gcp"
       commands:
         - command: shell.exec
           params:
@@ -2139,7 +2141,7 @@ task_groups:
     setup_group_can_fail_task: true
     setup_group_timeout_secs: 1800
     tasks:
-      - oidc-auth-test-latest
+      - oidc-auth-test
 
   - name: testazureoidc_task_group
     setup_group:
@@ -2162,7 +2164,7 @@ task_groups:
     setup_group_can_fail_task: true
     setup_group_timeout_secs: 1800
     tasks:
-      - oidc-auth-test-azure-latest
+      - oidc-auth-test-azure
 
   - name: testgcpoidc_task_group
     setup_group:
@@ -2186,7 +2188,7 @@ task_groups:
     setup_group_can_fail_task: true
     setup_group_timeout_secs: 1800
     tasks:
-      - oidc-auth-test-gcp-latest
+      - oidc-auth-test-gcp
 
 buildvariants:
 
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -240,7 +240,8 @@ import static java.util.Collections.unmodifiableList;
  * mechanism (the default).
  * </li>
  * <li>{@code authMechanismProperties=PROPERTY_NAME:PROPERTY_VALUE,PROPERTY_NAME2:PROPERTY_VALUE2}: This option allows authentication
- * mechanism properties to be set on the connection string. Property values must be percent-encoded individually, as needed. The
+ * mechanism properties to be set on the connection string. Property values must be percent-encoded individually, when
+ * separator or escape characters are used (including {@code ,} (comma), {@code =}, {@code +}, {@code &}, and {@code %}). The
  * entire substring following the {@code =} should not itself be encoded.
  * </li>
  * <li>{@code gssapiServiceName=string}: This option only applies to the GSSAPI mechanism and is used to alter the service name.
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -38,6 +38,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -923,6 +924,9 @@ public class ConnectionString {
                 }
                 String key = mechanismPropertyKeyValue[0].trim().toLowerCase();
                 String value = mechanismPropertyKeyValue[1].trim();
+                if (!decodeWholeOptionValue(mechanism)) {
+                    value = urldecode(value);
+                }
                 if (MECHANISM_KEYS_DISALLOWED_IN_CONNECTION_STRING.contains(key)) {
                     throw new IllegalArgumentException(format("The connection string contains disallowed mechanism properties. "
                             + "'%s' must be set on the credential programmatically.", key));
@@ -936,6 +940,13 @@ public class ConnectionString {
             }
         }
         return credential;
+    }
+
+    private static boolean decodeWholeOptionValue(final AuthenticationMechanism mechanism) {
+        return !AuthenticationMechanism.MONGODB_OIDC.equals(mechanism);
+    }
+    private static boolean decodeWholeOptionValue(final List<String> options) {
+        return !options.contains("authMechanism=" + AuthenticationMechanism.MONGODB_OIDC.getMechanismName());
     }
 
     private MongoCredential createMongoCredentialWithMechanism(final AuthenticationMechanism mechanism, final String userName,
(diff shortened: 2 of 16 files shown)


driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -916,7 +916,7 @@ private MongoCredential createCredentials(final Map<String, List<String>> option
 
         if (credential != null && authMechanismProperties != null) {
             for (String part : authMechanismProperties.split(",")) {
-                String[] mechanismPropertyKeyValue = part.split(":");
+                String[] mechanismPropertyKeyValue = part.split(":", 2);
```
katcharov, line 919: This ignores the second `:` (a test passes in a value that decodes to a string that includes `:`).

katcharov:
Rebased `feature-oidc` and this branch on latest master, no conflicts.

driver-core/src/main/com/mongodb/MongoCredential.java
```diff
@@ -706,9 +728,11 @@ public interface IdpInfo {
     }
 
     /**
-     * The response produced by an OIDC Identity Provider.
+     * The OIDC credential information.
+     *
+     * @since 5.1
      */
-    public static final class OidcCallbackResult {
+    public static final class OidcTokens {
```
katcharov, line 735: In the spec, an [example](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md?plain=1#L1433) is given that uses the name `OIDCCredential`, but this would be confusing here, since we use the same terminology for creating an OIDC MongoCredential, `#createOidcCredential`.

katcharov: Rename back to OidcCallbackResult

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
```diff
@@ -69,11 +73,20 @@
  */
 public final class OidcAuthenticator extends SaslAuthenticator {
 
-    private static final List<String> SUPPORTED_PROVIDERS = Arrays.asList("aws");
+    private static final String TEST_ENVIRONMENT = "test";
```
katcharov, line 76: This is only used (enabled) in unified tests. The "Entities" class could instead check if this environment is set, and supply the test callback under OIDC_CALLBACK_KEY, which is a mild abuse of mechanism properties, but would avoid the (limited) code needed in the driver to support this.

driver-core/src/test/unit/com/mongodb/ConnectionStringSpecification.groovy
```diff
@@ -601,7 +601,7 @@ class ConnectionStringSpecification extends Specification {
         new ConnectionString('mongodb://jeff@localhost/?' +
                              'authMechanism=GSSAPI' +
                              '&authMechanismProperties=' +
-                             'SERVICE_NAME:foo:bar')
+                             'SERVICE_NAMEbar') // missing =
```
katcharov, line 604: No spec was associated with the Jira ticket where this check was added. OIDC tests include a value with a second `:` (when decoded). Rather than remove, I replaced with a different invalidity check.

stIncMale (changes requested):

.evergreen/.evg.yml
```diff
@@ -968,6 +967,58 @@ tasks:
         - func: "run load-balancer"
         - func: "run load-balancer tests"
 
+    - name: "oidc-auth-test-latest"
```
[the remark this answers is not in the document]
blink1073: Yeah that makes sense to me, in hindsight I should have done the same in Python.

.evergreen/.evg.yml
```diff
@@ -2216,6 +2339,27 @@ buildvariants:
   tasks:
     - name: "test_atlas_task_group_search_indexes"
 
+- name: "oidc-auth-test"
+  display_name: "OIDC Auth"
+  run_on: ubuntu2204-small
```
blink1073, line 2344: We explicitly test on Windows and MacOS in PyMongo, but I'd consider that optional.

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
```diff
@@ -167,10 +174,47 @@ private OidcCallback getOidcCallbackMechanismProperty(final String key) {
...
+            String accessToken = readTokenFromFile();
+            return new OidcCallbackResult(accessToken);
+        };
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.AccessModifier.PRIVATE)
+    public static OidcCallback getAzureCallback(final MongoCredential credential) {
+        return (context) -> {
+            String resource = assertNotNull(credential.getMechanismProperty(TOKEN_RESOURCE_KEY, null));
+            String objectId = credential.getUserName();
```
blink1073, line 205: This is actually the clientId

katcharov: Hmm, my ensuing fetchAzure... call is calling:

```
    public static CredentialInfo fetchAzureCredentialInfo(final String resource, @Nullable final String objectId) {
        String endpoint = "http://169.254.169.254:80"
                + "/metadata/identity/oauth2/token?api-version=2018-02-01"
                + "&resource=" + resource
                + (objectId == null ? "" : "&object_id=" + objectId);
```
This must be incorrect, since the spec says to use:
```
http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=<resource>&client_id=<client_id>
```
I'll update this, but why do the tests still pass, and was this changed from object_id?

blink1073: It was changed from object_id to align with the Azure SDK that uses client_id.

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -529,30 +522,42 @@ private void initClient(final BsonDocument entity, final String id,
...
-                                        }
-                                        return new MongoCredential.OidcCallbackResult(accessToken, Duration.ZERO);
-                                    }));
-                            break;
+                        // authMechanismProperties are handled as part of authMechanism, above
+                        BsonValue authMechanism = entity
+                                .getDocument("uriOptions")
+                                .get("authMechanism");
+                        if (authMechanism.equals(new BsonString(MONGODB_OIDC.getMechanismName()))) {
+                            break; // only OIDC supports authMechanismProperties
```
stIncMale, line 560: > only OIDC supports authMechanismProperties

OIDC is not the only mechanism that supports the `authMechanismProperties` connection string option. Let's remove the comment?

This connection string option have existed before OIDC, and can be specified in `uriOptions` for other authentication mechanisms in unified tests. However, in practice, all currently existing unified tests specify `authMechanismProperties` in `uriOptions` only for the OIDC mechanism, which is why `UnsupportedOperationException` is thrown below when the mechanism is not OIDC. We do this all the time, and will implement applying `authMechanismProperties` for other auth mechanisms in unified tests on demand.

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -529,30 +522,42 @@ private void initClient(final BsonDocument entity, final String id,
...
-                            clientSettingsBuilder.credential(MongoCredential.createOidcCredential(null));
+                        if (value.equals(new BsonString(MONGODB_OIDC.getMechanismName()))) {
+                            // authMechanismProperties depends on authMechanism
+                            BsonDocument authMechanismProperties = entity
+                                    .getDocument("uriOptions")
+                                    .getDocument("authMechanismProperties");
+                            boolean hasPlaceholder = authMechanismProperties.equals(
+                                    new BsonDocument("$$placeholder", new BsonInt32(1)));
+                            if (!hasPlaceholder) {
+                                throw new UnsupportedOperationException("Unsupported authMechanism: " + value);
```
stIncMale, line 533: It's not the `authMechanism` that is not supported here, but any `authMechanismProperties` that is not equal to `{ "$$placeholder": 1 }`. The exception message, however, says `Unsupported authMechanism` and includes `value` (the authentication mechanism) instead of including `authMechanismProperties`.

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -529,30 +522,42 @@ private void initClient(final BsonDocument entity, final String id,
...
+                            }
+
+                            String env = getenv("OIDC_ENV");
+                            if (env == null) {
+                                env = "test";
+                            }
```
stIncMale, line 539: `env` is not expected to be `null`. We should assert that `env` is not `null`, instead of using `test` when this happens. This way we also won't accidentally run all tests using the `"test"` `ENVIRONMENT`.

.evergreen/.evg.yml
katcharov: All of our tests use linux/ubuntu as the `os`, so I took those out.

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
katcharov: "authMechanismProperties only supported here for OIDC", or do you feel no comment is needed at all? I am trying to note why this is `break`ing for OIDC but not others.

driver-core/src/main/com/mongodb/MongoCredential.java
```diff
@@ -248,11 +249,18 @@ public final class MongoCredential {
...
      */
     public static final List<String> DEFAULT_ALLOWED_HOSTS = Collections.unmodifiableList(Arrays.asList(
             "*.mongodb.net", "*.mongodb-qa.net", "*.mongodb-dev.net", "*.mongodbgov.net", "localhost", "127.0.0.1", "::1"));
 
+    /**
+     * The token resource.
+     *
+     * @since 5.1
+     */
+    public static final String TOKEN_RESOURCE_KEY = "TOKEN_RESOURCE";
```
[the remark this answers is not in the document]
katcharov: I am not sure about specifying the environments explicitly, since it seems easy to forget to update these, and a user configuring (for example) GCP should know that their environment needs it.

driver-sync/src/test/functional/com/mongodb/internal/connection/OidcAuthenticationProseTests.java
```diff
@@ -256,118 +253,210 @@ public void test3p1AuthFailsWithCachedToken() throws ExecutionException, Interru
...
+    public void test5p1Azure() {
+        assumeTrue(getOidcEnv().equals("azure"));
+        String oidcUri = getOidcUri();
+        assertFalse(oidcUri.contains("@"));
+        MongoClientSettings clientSettings = createSettings(oidcUri, createCallback(), null);
+        try (MongoClient mongoClient = createMongoClient(clientSettings)) {
+            // #. Perform a find operation that succeeds.
+            performFind(mongoClient);
+        }
+    }
```
stIncMale, line 376: I assume this test corresponds to https://github.com/mongodb/specifications/blob/master/source/auth/tests/mongodb-oidc.md#51-azure-with-no-username. It requires a valid `TOKEN_RESOURCE` to be set, but the test neither sets it explicitly, nor explicitly verifies that `TOKEN_RESOURCE` is set. Instead, the test relies on `getOidcUri()` having it. Such an approach means that if the assumption that `getOidcUri()` has `TOKEN_RESOURCE` is violated, the test will fail with an exception coming from the driver code (as opposed to coming from the test code), which will be confusing.

The approach taken in the PR for checking that `getOidcUri()` does not have a user name specified, is different - there is an explicit check `assertFalse(oidcUri.contains("@"))` for that. Let's do the same for `TOKEN_RESOURCE`, only checking it based on `clientSettings` is clearer and less error-prone than checking it by looking at the `oidcUri` string.

P.S. If we have other tests with similar situations (`test5p1AzureFails` is another one, maybe there are more), we should also introduce the checks there.

driver-sync/src/test/functional/com/mongodb/internal/connection/OidcAuthenticationProseTests.java
```diff
@@ -256,118 +253,210 @@ public void test3p1AuthFailsWithCachedToken() throws ExecutionException, Interru
...
+                .getDatabase("test")
+                .getCollection("test")
+                .insertOne(Document.parse("{ x: 1 }"));
+    }
+
+    @Test
+    public void test5p1Azure() {
+        assumeTrue(getOidcEnv().equals("azure"));
+        String oidcUri = getOidcUri();
+        assertFalse(oidcUri.contains("@"));
```
stIncMale, line 370: It is not obvious that this check is to make sure `getOidcUri()` does not have a user name specified. Let's rather do this check using `clientSettings`, where it will be more clear and less error-prone.

driver-sync/src/test/functional/com/mongodb/internal/connection/OidcAuthenticationProseTests.java
```diff
@@ -256,118 +253,210 @@ public void test3p1AuthFailsWithCachedToken() throws ExecutionException, Interru
...
+    @Test
+    public void test5p1AzureFails() {
+        assumeTrue(getOidcEnv().equals("azure"));
+        String oidcUri = getOidcUri();
+        oidcUri = oidcUri.replace("://", "://bad@");
+        MongoClientSettings clientSettings = createSettings(oidcUri, createCallback(), null);
+        try (MongoClient mongoClient = createMongoClient(clientSettings)) {
+            // #. Perform a find operation that succeeds.
+            performFind(mongoClient);
+        }
```
stIncMale, line 387: I assume this test corresponds to https://github.com/mongodb/specifications/blob/master/source/auth/tests/mongodb-oidc.md#52-azure-with-bad-username. The `performFind` should fail here, yet it succeeds, and even the comment states that it is supposed to succeed.

driver-sync/src/test/functional/com/mongodb/internal/connection/OidcAuthenticationProseTests.java
```diff
@@ -256,118 +253,210 @@ public void test3p1AuthFailsWithCachedToken() throws ExecutionException, Interru
...
+        assertFalse(oidcUri.contains("@"));
+        MongoClientSettings clientSettings = createSettings(oidcUri, createCallback(), null);
+        try (MongoClient mongoClient = createMongoClient(clientSettings)) {
+            // #. Perform a find operation that succeeds.
+            performFind(mongoClient);
+        }
+    }
+
+    @Test
+    public void test5p1AzureFails() {
```
stIncMale, line 379: Should it be named `test5p2AzureFails`, as the test seem to correspond to https://github.com/mongodb/specifications/blob/master/source/auth/tests/mongodb-oidc.md#52-azure-with-bad-username?

stIncMale (changes requested):

blink1073 (approved):
LGTM!

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
```diff
@@ -553,10 +589,10 @@ public static void validateOidcCredentialConstruction(
                 throw new IllegalArgumentException("source must be '$external'");
             }
 
-            Object providerName = mechanismProperties.get(PROVIDER_NAME_KEY.toLowerCase());
+            String providerName = getEnvironmentName(mechanismProperties);
             if (providerName != null) {
-                if (!(providerName instanceof String) || !SUPPORTED_PROVIDERS.contains(providerName)) {
-                    throw new IllegalArgumentException(PROVIDER_NAME_KEY + " must be one of: " + SUPPORTED_PROVIDERS);
+                if (!SUPPORTED_ENVIRONMENTS.contains(providerName)) {
+                    throw new IllegalArgumentException(ENVIRONMENT_KEY + " must be one of: " + SUPPORTED_ENVIRONMENTS);
```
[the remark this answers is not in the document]
katcharov: OidcAuthenticator is an internal class, and it supports (internally) test. Perhaps you mean that "supported" too-much suggests that it is user-facing, but then I am unsure what the alternative is... "allowed environments" for example seems to have the same problem.

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
stIncMale: I would not have the comment at all, as the fact that `break;` is executed only for OIDC is clear from the `if` condition, which checks exactly for that. But the updated comment does not have a second meaning (is not ambiguous), as the previous comment, so if you think it's needed, I have no objections.

driver-core/src/main/com/mongodb/internal/Locks.java
```diff
@@ -54,7 +54,7 @@ public static void withLockAsync(final StampedLock lock, final AsyncRunnable run
         }, callback);
     }
 
-    public static void withLock(final StampedLock lock, final Runnable runnable) {
+    public static void withInterruptibleLock(final StampedLock lock, final Runnable runnable) {
```
stIncMale, line 57:
```suggestion
    public static void withInterruptibleLock(final StampedLock lock, final Runnable runnable) throws MongoInterruptedException {
```

stIncMale, line 57: - If this method is renamed, then its asynchronous counterpart `withLockAsync(StampedLock, AsyncRunnable, SingleResultCallback<Void>)` should also be renamed similarly
- `withLockAsync should also be declared as `throws MongoInterruptedException` consistently with the rest of the `Locks` methods that use interruptible locking.

driver-core/src/test/resources/auth/legacy/connection-string.json
stIncMale, line 1: Commits https://github.com/mongodb/mongo-java-driver/pull/1371/commits/d856d84a2b6793528253cacd0c81034803e49582 and https://github.com/mongodb/mongo-java-driver/pull/1371/commits/cc1c7eccd48c2a38c9362a5fa385a4e56ec1dcf2 in this PR take changes from https://github.com/mongodb/specifications/pull/1569, which hasn't been approved and merged. Are we sure this is the way to go? I realize that we are short of time, but taking spec changes that are still work in progress (if this were not the case, they would have been merged) is not the right process, and we may end up having tests that have never been merged to the spec.

I would also expect that if there is time pressure on the driver, the same (even more, actually) should be true for the spec changes, because they have to go before the driver changes.

Maxim, if you have already confirmed this with Jeff, then OK. If not, then could you please confirm? (alternatively, I can do that)

stIncMale (changes requested):

driver-core/src/test/resources/auth/legacy/connection-string.json
katcharov: I agree that we should block merging of at least the OIDC feature PR on the approval and merging of that PR (and I have no issues with blocking this PR since nothing depends on it).

driver-core/src/main/com/mongodb/MongoCredential.java
```diff
@@ -185,31 +186,31 @@ public final class MongoCredential {
     public static final String AWS_CREDENTIAL_PROVIDER_KEY = "AWS_CREDENTIAL_PROVIDER";
 
     /**
-     * The provider name. The value must be a string.
+     * The environment. The value must be a string.
```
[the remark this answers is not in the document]
stIncMale: [moved from https://github.com/mongodb/mongo-java-driver/pull/1371#discussion_r1580045670, which is not the right location for this comment]

The spec does a good job explaining what `ENVIRONMENT` is, by saying that it's the name of a built-in integration with an OIDC application environment (the spec does not call it "the environment for OIDC", as the current Java API docs do). That description, combined with your "alternative to supplying a callback" would be a good way to describe this mechanism property.

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
```diff
@@ -588,19 +624,40 @@ public static void validateBeforeUse(final MongoCredential credential) {
...
                 }
             }
         }
     }
 
+    @Nullable
+    private static String getEnvironmentName(final Map<String, Object> mechanismProperties) {
+        Object o = mechanismProperties.get(ENVIRONMENT_KEY.toLowerCase());
+        return o instanceof String ? (String) o : null;
+    }
```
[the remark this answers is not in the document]
katcharov: I have removed this and the other method.

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
```diff
@@ -588,19 +624,40 @@ public static void validateBeforeUse(final MongoCredential credential) {
...
+
+    @Nullable
+    private static String getEnvironmentName(final MongoCredential credential) {
+        Object o = credential.getMechanismProperty(ENVIRONMENT_KEY, null);
+        return o instanceof String ? (String) o : null;
```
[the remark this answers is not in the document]
katcharov: Removed method.

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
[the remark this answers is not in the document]
katcharov: "implemented"

driver-core/src/main/com/mongodb/internal/Locks.java
katcharov: Deleted dead code

driver-sync/src/test/functional/com/mongodb/internal/connection/OidcAuthenticationProseTests.java
katcharov: I have added checks, but I do not think they are strictly needed. We generally trust that that the test runner and environment is correctly set up, so for example, an "assumption" would just disable the test. However here, I think it is fine to have these checks, because these two Azure tests are somewhat different from the rest, just to make it clearer to the reader that no mistakes were made in setup.

driver-core/src/main/com/mongodb/ConnectionString.java
stIncMale, line 1: We should at the very least document the new behavior with regard to decoding connection string option values. I don't know if this can be done without also documenting the current behavior, as it will continue coexisting with the new behavior.

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/internal/connection/OidcAuthenticationProseTests.java
katcharov: Fixed the test. Azure passing here https://spruce.mongodb.com/version/662c354f13c2570007349f91/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

It's a bit of an odd test, since it relies on a server error. For reference, in C# and Python:
* https://github.com/mongodb/mongo-csharp-driver/blob/8af9316816ca338a83d62d2812a3b7c06cb3c736/tests/MongoDB.Driver.Tests/Specifications/auth/OidcAuthenticationProseTests.cs#L306
* https://github.com/mongodb/mongo-python-driver/blob/4ea8fcbb3b8876923e92f4c98324a16249f00ab7/test/auth_oidc/test_auth_oidc.py#L1034

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -916,13 +917,16 @@ private MongoCredential createCredentials(final Map<String, List<String>> option
...
                 if (mechanismPropertyKeyValue.length != 2) {
                     throw new IllegalArgumentException(format("The connection string contains invalid authentication properties. "
                             + "'%s' is not a key value pair", part));
                 }
                 String key = mechanismPropertyKeyValue[0].trim().toLowerCase();
                 String value = mechanismPropertyKeyValue[1].trim();
+                if (!decodeWholeOptionValue(mechanism)) {
```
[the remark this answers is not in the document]
katcharov: Tried renaming

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -1034,7 +1047,10 @@ private Map<String, List<String>> parseOptions(final String optionsPart) {
                 if (valueList == null) {
                     valueList = new ArrayList<>(1);
                 }
-                valueList.add(urldecode(value));
+                if (decodeWholeOptionValue) {
+                    value = urldecode(value);
```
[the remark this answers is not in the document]
katcharov: Good catch.

driver-core/src/main/com/mongodb/ConnectionString.java
katcharov: I have tried to do this in ConnectionString (find the section `code authMechanismProperties=PROPERTY_NAME`...), discouraging the user from making use of the current behaviour. It is fine to no longer encode commas that were previously encoded. I did not mention OIDC there specifically; it seems clear when encoding would be needed, given the example string.

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
```diff
@@ -167,10 +174,47 @@ private OidcCallback getOidcCallbackMechanismProperty(final String key) {
...
+            machine = getGcpCallback(getMongoCredential());
+        } else {
+            machine = getOidcCallbackMechanismProperty(OIDC_CALLBACK_KEY);
+        }
+        OidcCallback human = getOidcCallbackMechanismProperty(OIDC_HUMAN_CALLBACK_KEY);
+        return machine != null ? machine : assertNotNull(human);
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.AccessModifier.PRIVATE)
+    public static OidcCallback getTestCallback() {
```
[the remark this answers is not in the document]
stIncMale: Looks like this change wasn't applied as suggested. The method is now package-private, but it does not have to be that.

driver-core/src/test/unit/com/mongodb/ConnectionStringUnitTest.java
```diff
@@ -34,6 +38,31 @@ void defaults() {
...
+                "mongodb://localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties="
+                        + "ENVIRONMENT:azure,TOKEN_RESOURCE:" + encode(string));
+        assertEquals(string, cs.getCredential().getMechanismProperty("TOKEN_RESOURCE", null));
+    }
+
+    @Test
+    public void mustDecodeNonOidcAsWhole()  {
+        ConnectionString cs2 = new ConnectionString(
+                "mongodb://foo:bar@example.com/?authMechanism=GSSAPI&authMechanismProperties="
+                        + encode("SERVICE_NAME:other,CANONICALIZE_HOST_NAME:true&authSource=$external"));
```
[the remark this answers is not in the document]
stIncMale: The change that has been done only introduced a space to `other`. This is different from the proposed change and does not achieve its goal. By using `encode("oth er")` as the raw value of `SERVICE_NAME`, we can verify that `ConnectionString` does not decode that value in addition to decoding `authMechanismProperties` as a whole, i.e., that there is no double-decoding.

driver-core/src/test/unit/com/mongodb/ConnectionStringUnitTest.java
```diff
@@ -41,18 +42,23 @@ void defaults() {
     @Test
     public void mustDecodeOidcIndividually() {
         String string = "abc,d!@#$%^&*;ef:ghi";
+        // encoded tags will fail parsing with an "invalid read preference tag"
+        // error if decoding is skipped.
+        String encodedTags = encode("dc:ny,rack:1");
```
stIncMale, line 47: 👌

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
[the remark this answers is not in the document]
katcharov: Right, I see (it is the error message rather than the field naming).

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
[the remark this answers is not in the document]
katcharov: Right, it is not like gcp and azure.

driver-sync/src/test/functional/com/mongodb/internal/connection/OidcAuthenticationProseTests.java
```diff
@@ -256,118 +260,223 @@ public void test3p1AuthFailsWithCachedToken() throws ExecutionException, Interru
...
+    }
+
+    @Test
+    public void test5p1AzureSucceedsWithNoUsername() {
+        assumeAzure();
+        String oidcUri = getOidcUri();
+        MongoClientSettings clientSettings = createSettings(oidcUri, createCallback(), null);
+        // Create an OIDC configured client with `ENVIRONMENT:azure` and a valid
+        // `TOKEN_RESOURCE` and no username.
+        MongoCredential credential = Assertions.assertNotNull(new ConnectionString(oidcUri).getCredential());
```
stIncMale, line 380:
```suggestion
        MongoCredential credential = Assertions.assertNotNull(clientSettings.getCredential());
```

stIncMale (changes requested):
Just a small thing

stIncMale (approved):

PR #1372, opened 2024-04-20
Mark CSOT feature as Alpha

vbabanin:
<h3>Description</h3>

This PR designates the CSOT API as Alpha for experimental use.

Additionally, it introduces the Alpha annotation. While initially utilized within the CSOT API, the annotation's Javadoc covers broader context beyond CSOT.

JAVA-5402


THIRD-PARTY-NOTICES
```diff
@@ -37,7 +37,12 @@ https://github.com/mongodb/mongo-java-driver.
     See the License for the specific language governing permissions and
     limitations under the License.
 
-3) The following files: Beta.java, UnsignedLongs.java, UnsignedLongsTest.java
+3) The following files:
+
+    Alpha.java (formerly Beta.java)
```
vbabanin, line 42: `Alpha.java` transitively utilizes some of the Javadoc wording from Guava’s `Beta.java.`

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/MongoCluster.java
```diff
@@ -165,6 +166,7 @@ public interface MongoCluster {
      * @since CSOT
      * @see #getTimeout
      */
+    @Alpha(Alpha.Reason.CLIENT)
```
jyemin, line 169: Add to `getTimeout` as well?

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/MongoDatabase.java
```diff
@@ -160,6 +161,7 @@ public interface MongoDatabase {
      * @since CSOT
      * @see #getTimeout
      */
+    @Alpha(Alpha.Reason.CLIENT)
```
jyemin, line 164: Add to `getTimeout` as well?

jyemin (changes requested):
Looking good.

I noted a few missing annotations on `getTimeout` methods, but please double-check that they are all annotated.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/MongoCluster.java
vbabanin: Done, thanks!

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/MongoDatabase.java
vbabanin: Done, thanks!

jyemin (approved):
LGTM

driver-scala/src/main/scala/org/mongodb/scala/gridfs/GridFSBucket.scala
```diff
@@ -170,6 +172,7 @@ case class GridFSBucket(private val wrapped: JGridFSBucket) {
    * @return a new GridFSBucket instance with the set time limit for operations
    * @since 4.x
```
stIncMale, line 173: Shouldn't this be `@since CSOT` similarly to line `GridFSBucket.scala:124` for `timeout`?

driver-core/src/main/com/mongodb/MongoOperationTimeoutException.java
```diff
@@ -35,6 +37,7 @@
  * @see MongoClientSettings#getTimeout(TimeUnit)
  * @since CSOT
  */
+@Alpha(Alpha.Reason.CLIENT)
```
stIncMale, line 40: It seems to me that if `MongoOperationTimeoutException` is `@Alpha`, we should not throw it unless CSOT is specified. However, `SocketStreamHelper.initialize` calls `TimeoutContext.createConnectTimeoutMs`, which throws `MongoOperationTimeoutException` when `TimeoutContext` expires, regardless of whether `TimeoutContext.timeout` is `null` or not. I.e., at the very least `SocketStreamHelper.initialize`, but maybe also some other parts of the driver, throw `MongoOperationTimeoutException` even if CSOT is not specified.

driver-core/src/main/com/mongodb/annotations/Alpha.java
```diff
@@ -0,0 +1,63 @@
...
+package com.mongodb.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Signifies that a public API (public class, method or field) is in the early stages
```
stIncMale, line 27: > public class, method or field

What about an interface or record? It seems to me that saying "a public API element" is better.

driver-core/src/main/com/mongodb/annotations/Alpha.java
```diff
@@ -0,0 +1,63 @@
...
+ * of development, subject to incompatible changes, or even removal, in a future release
+ * and may lack some intended features. An API bearing this annotation may be unstable,
+ * have potential performance implications as development progresses, and is exempt from
+ * any compatibility guarantees made by its containing library.
```
stIncMale, line 31: None of this, as far as I understand, means "is not of the same quality as ordinary or `@Beta` public API, has much higher chances of having known or unknown bugs". Should we express that?

P.S. "unstable", when it comes to API, means to me the same as "subject to incompatible changes, or even removal".

driver-core/src/main/com/mongodb/annotations/Alpha.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @see Alpha#value()
+     */
+    enum Reason {
+        /**
+         * Indicates that the driver API is either experimental or in development.
+         * Use in production environments is inadvisable due to potential API changes and possible instability.
+         */
+        CLIENT,
+    }
```
stIncMale, line 62: Let's rather extract `Beta.Reason` into a top-level class and use that in both `Alpha` and `Beta`? We can do that because `Beta.Reason` is itself beta (a class nested in a beta class can't be non-beta, as it's part of the beta class, which is allowed to be removed).

driver-core/src/main/com/mongodb/annotations/Alpha.java
```diff
@@ -0,0 +1,63 @@
...
+@Retention(RetentionPolicy.CLASS)
+@Target({
+        ElementType.ANNOTATION_TYPE,
+        ElementType.CONSTRUCTOR,
+        ElementType.FIELD,
+        ElementType.METHOD,
+        ElementType.PACKAGE,
+        ElementType.TYPE })
+@Documented
+@Alpha(Alpha.Reason.CLIENT)
```
stIncMale, line 46: I think, the `@Alpha` annotation itself should rather be marked beta.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/annotations/Alpha.java
```diff
@@ -0,0 +1,63 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Signifies that a public API (public class, method or field) is in the early stages
+ * of development, subject to incompatible changes, or even removal, in a future release
+ * and may lack some intended features. An API bearing this annotation may be unstable,
+ * have potential performance implications as development progresses, and is exempt from
```
stIncMale, line 30: Do we really want to imply that when released initially, an alpha API does not have performance issues, but may develop them "as development progresses"? I propose us to just say "... have potential performance implications, ..."

driver-core/src/main/com/mongodb/MongoOperationTimeoutException.java
vbabanin: The MongoOperationTimeoutException should indeed only be thrown when CSOT is enabled. This appears to be a newly introduced bug following the Timeout API refactoring. The exception should not be thrown here, as there is  code a few layers up specifically used to manage SocketTimeoutException and decide which exception should be relayed to the user. 

I will investigate this issue further and check that MongoOperationTimeout exception is not exposed to the user when CSOT is enabled in a separate ticket.  However, i would leave `MongoOperationTimeoutException` marked as @Alpha for now.

Good catch! Thanks!

vbabanin: The `MongoOperationTimeoutException` should indeed only be thrown when CSOT is enabled. This appears to be a newly introduced bug following the Timeout API refactoring. I will investigate this issue further in a separate ticket JAVA-5439. However, the `MongoOperationTimeoutException` should remain marked as @Alpha.

driver-core/src/main/com/mongodb/annotations/Alpha.java
vbabanin: I agree with your definition of 'unstable' as it pertains to APIs, builds or releases that are not yet finalized and may be subject to significant changes. However, this term also typically implies possibly incomplete testing, which could lead to known and unknown bugs. Although it does not explicitly state the software is untested, it suggests a higher likelihood of issues.

To avoid any confusion and manage user expectation, we could phrase it like this:

'An API bearing this annotation is considered unstable and might not have been thoroughly tested. It may contain issues that could affect performance and functionality. Additionally, it is exempt from any compatibility guarantees made by its containing library.'

What do you think?

jyemin: I don't think we should say there's lack of thorough testing.  By the standards we have for GA features of the driver, CSOT is thoroughly tested.

driver-scala/src/main/scala/org/mongodb/scala/gridfs/GridFSBucket.scala
vbabanin: Yes, these placeholders are designed to streamline the release process, whether for versions 5.1 or 5.2. We could address them in a separate PR.

driver-core/src/main/com/mongodb/annotations/Alpha.java
vbabanin: Yes, CSOT has undergone thorough testing, so it would be inaccurate to suggest otherwise in the Javadoc in this case. As Valentin highlighted, the term 'unstable' can indeed be interpreted in different ways.

I have updated the Javadoc to more accurately reflect our discussion:

```
/**
 * An API bearing this annotation may contain known issues affecting functionality,
 * performance, and stability. It is also exempt from any compatibility guarantees made by its
 * containing library.
 */
```
Please let me know if there are any further suggestions

README.md
```diff
@@ -52,6 +52,17 @@ almost always be binary compatible with prior minor releases from the same major
 Patch 4.x.y increments (such as 4.0.0 -> 4.0.1, 4.1.1 -> 4.1.2, etc) will occur for bug fixes only and will always be binary compatible
 with prior patch releases of the same minor release branch.
 
+#### @Alpha
```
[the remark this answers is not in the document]
vbabanin: I agree there's a degree of redundancy. However, users could refer to the README for quick, high-level insights, and noting that there are might be APIs marked as Alpha or Beta could prompt them to pay attention to the Javadocs. Additionally, making contributors aware of the parts of the API that are still under development might encourage their involvement. Since we rarely introduce new annotations like this, i believe the redundancy should not be problematic.

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/FindIterable.kt
```diff
@@ -53,6 +55,7 @@ public class FindIterable<T : Any>(private val wrapped: JFindIterable<T>) : Mong
      * @return this
      * @since CSOT
      */
+    @Alpha(Reason.CLIENT)
```
stIncMale, line 58: Shuldn't `com.mongodb.client.FindIterable.timeoutMode` also be marked as `Alpha`? If yes, could you please check again the rest of the codebase, as maybe there are other misses.

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -137,7 +139,8 @@
  * <li>{@code sslInvalidHostNameAllowed=true|false}: Whether to allow invalid host names for TLS connections.</li>
  * <li>{@code tlsAllowInvalidHostnames=true|false}: Whether to allow invalid host names for TLS connections. Supersedes the
  * sslInvalidHostNameAllowed option</li>
- * <li>{@code timeoutMS=ms}: Time limit for the full execution of an operation.</li>
+ * <li>{@code timeoutMS=ms}: Time limit for the full execution of an operation. Note: This parameter is part of an Alpha API and may be
```
stIncMale, line 142:
```suggestion
 * <li>{@code timeoutMS=ms}: Time limit for the full execution of an operation. Note: This parameter is part of an {@linkplain Alpha Alpha API} and may be
```

driver-core/src/main/com/mongodb/annotations/Reason.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.annotations;
+
+/**
+ * Enumerates the reasons an API element might be marked with annotations like {@link Alpha} or {@link Beta}.
+ */
+public enum Reason {
```
stIncMale, line 22: Let's mark it with `@Beta`.

stIncMale (changes requested):

README.md
```diff
@@ -54,13 +54,13 @@ with prior patch releases of the same minor release branch.
 
 #### @Alpha
 
-APIs marked with the `@Alpha` annotation at the class, method or field level are in the early stages of development, 
-subject to incompatible changes, or even removal, in a future release and may lack some intended features. These APIs 
-may be unstable, have potential performance implications as development progresses, and are exempt from any compatibility 
-guarantees made by its containing library. 
+APIs marked with the `@Alpha` annotation at a public API element are in the early stages of development, subject to
```
stIncMale, line 57: _"at a public API element"_ looks out of place.

stIncMale (changes requested):

stIncMale (approved):

PR #1373, opened 2024-04-27
Fix regression in timeout handling.

vbabanin:
- Resolve regression where CSOT exception is exposed despite CSOT being disabled.
- Correct premature decrease in connect timeout before connection initiation.
- Encapsulating logic within TimeoutContext.

JAVA-5439

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/TimeoutContext.java
+++ b/driver-core/src/main/com/mongodb/internal/TimeoutContext.java
@@ -187,7 +187,7 @@ public class TimeoutContext {
 
     public void runMaxTimeMS(final LongConsumer onRemaining) {
         if (maxTimeSupplier != null) {
-            onRemaining.accept(maxTimeSupplier.get());
+            runWithFixedTimout(maxTimeSupplier.get(), onRemaining);
         }
         if (timeout != null) {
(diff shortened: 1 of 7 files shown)


driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -185,25 +185,27 @@ public long getMaxAwaitTimeMS() {
...
+                            throw createMongoRoundTripTimeoutException();
+                        });
+
     }
 
-    private static void runWithFixedTimout(final long ms, final Runnable onInfinite, final LongConsumer onRemaining) {
-        if (ms == 0) {
-            onInfinite.run();
-        } else {
+    private static void runWithFixedTimout(final long ms, final LongConsumer onRemaining) {
```
rozza, line 207: Nit: There was a typo in the original name: 

`runWithFixedTimout` => `runWithFixedTimeout`

rozza (approved):
One typo nit but LGTM

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
vbabanin: Done, thanks!

PR #1389, opened 2024-05-13
OIDC admin credentials

katcharov:
Fixes failing tests: reverts removal of admin credentials for OIDC tests, and corrects a typo in a validation test.
JAVA-5450

### DIFF
--- a/.evergreen/run-mongodb-oidc-test.sh
+++ b/.evergreen/run-mongodb-oidc-test.sh
@@ -34,7 +34,12 @@ fi
 which java
 export OIDC_TESTS_ENABLED=true
 
-./gradlew -Dorg.mongodb.test.uri="$MONGODB_URI" \
+# use admin credentials for tests
+TO_REPLACE="mongodb://"
+REPLACEMENT="mongodb://$OIDC_ADMIN_USER:$OIDC_ADMIN_PWD@"
+ADMIN_URI=${MONGODB_URI/$TO_REPLACE/$REPLACEMENT}
+
+./gradlew -Dorg.mongodb.test.uri="$ADMIN_URI" \
   --stacktrace --debug --info --no-build-cache driver-core:cleanTest \
   driver-sync:test --tests OidcAuthenticationProseTests --tests UnifiedAuthTest \
   driver-reactive-streams:test --tests OidcAuthenticationAsyncProseTests \
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
@@ -18,6 +18,7 @@ package com.mongodb.client.unified;
 
 import com.mongodb.ClientEncryptionSettings;
 import com.mongodb.ClientSessionOptions;
+import com.mongodb.ConnectionString;
 import com.mongodb.MongoClientSettings;
 import com.mongodb.MongoCredential;
 import com.mongodb.ReadConcern;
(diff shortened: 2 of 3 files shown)


jyemin (commented):
Which tasks do we expect to pass now?  I started all the OIDC ones and two pass, while OIDC GCP still fails.

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -535,6 +536,11 @@ private void initClient(final BsonDocument entity, final String id,
                                         "Unsupported authMechanismProperties for authMechanism: " + value);
                             }
 
+                            // override the org.mongodb.test.uri connection string
```
jyemin, line 539: I don't understand why this is necessary.  Can't we always arrange to pass in the connection string as the system property?  Or are there secrets that we don't want to have logged?

.evergreen/run-mongodb-oidc-test.sh
```diff
@@ -34,7 +34,12 @@ fi
 which java
 export OIDC_TESTS_ENABLED=true
 
-./gradlew -Dorg.mongodb.test.uri="$MONGODB_URI" \
+# use admin credentials for tests
+TO_REPLACE="mongodb://"
+REPLACEMENT="mongodb://$OIDC_ADMIN_USER:$OIDC_ADMIN_PWD@"
+ADMIN_URI=${MONGODB_URI/$TO_REPLACE/$REPLACEMENT}
```
jyemin, line 40: This looks wrong. What's the intent?

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
katcharov: For OIDC, we need two different credentials - the OIDC credential for the test itself, and a credential that is able to perform admin operations, like dropping databases, that are required by our test infrastructure. Usually, the credential we use for the the test itself is also able to perform admin operations (and this must have been the case for OIDC as well, until a recent change).

Our prose tests already pull the credential for each test itself out of env variables, as per the spec, so only the unified tests needed to override this `org.mongodb.test.uri` to not use the admin credential.

.evergreen/run-mongodb-oidc-test.sh
katcharov: This creates a URI that is able to run admin commands.

driver-sync/src/test/functional/com/mongodb/internal/connection/OidcAuthenticationProseTests.java
```diff
@@ -234,6 +234,8 @@ public void test2p4InvalidClientConfigurationWithCallback() {
 
     @Test
     public void test2p5InvalidAllowedHosts() {
+        assumeTestEnvironment();
```
katcharov, line 237: There is no requirement to run this test in a non-"test" environment; disabled to prevent running on GCP. Tests [passing here](https://spruce.mongodb.com/version/664e31e50c8d75000729d6ba/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

.evergreen/run-mongodb-oidc-test.sh
jyemin: sed...

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
jyemin: Yuk, but ok.

jyemin (approved):

PR #1391, opened 2024-05-15
Remove RTT check before reading a response from the server.

vbabanin:
JAVA-5473

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
@@ -349,9 +349,6 @@ public class InternalStreamConnection implements InternalConnection {
         CommandEventSender commandEventSender;
 
         try (ByteBufferBsonOutput bsonOutput = new ByteBufferBsonOutput(this)) {
-            Timeout.onExistsAndExpired(operationContext.getTimeoutContext().timeoutIncludingRoundTrip(), () -> {
-                throw TimeoutContext.createMongoRoundTripTimeoutException();
-            });
             message.encode(bsonOutput, operationContext);
             commandEventSender = createCommandEventSender(message, bsonOutput, operationContext);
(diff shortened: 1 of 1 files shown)


driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -509,6 +502,16 @@ private <T> void sendCommandMessageAsync(final int messageId, final Decoder<T> d
                                              final SingleResultCallback<T> callback, final ByteBufferBsonOutput bsonOutput,
                                              final CommandEventSender commandEventSender, final boolean responseExpected) {
         List<ByteBuf> byteBuffers = bsonOutput.getByteBuffers();
+
+        boolean[] shouldReturn = {false};
```
vbabanin, line 506: I've made changes to move the log message from "after the write is complete" to "before the write is complete."

According to the CSOT spec:
> After wire message construction, drivers MUST check for timeout before writing the message to the server.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -349,9 +349,6 @@ public <T> T sendAndReceive(final CommandMessage message, final Decoder<T> decod
         CommandEventSender commandEventSender;
 
         try (ByteBufferBsonOutput bsonOutput = new ByteBufferBsonOutput(this)) {
-            Timeout.onExistsAndExpired(operationContext.getTimeoutContext().timeoutIncludingRoundTrip(), () -> {
```
vbabanin, line 352: According to the specification: 

> After wire message construction, drivers MUST check for timeout before writing the message to the server.


I have removed this check from its previous position as it was occurring before the wire message was constructed, not after, which is contrary to the spec. We still perform this necessary timeout check later in the code within the trySendMessage method, just before writing the message to the socket.

rozza (approved):
LGTM

PR #1392, opened 2024-05-16
Augment `config/spotbugs/exclude.xml` with finding status and rank

stIncMale:
To get the rank data, I had enabled SpotBugs XML reports, then was commenting out parts of `exclude.xml` and manually get the rank data from the XML reports.

JAVA-5431

### DIFF
--- a/bson/src/main/org/bson/types/ObjectId.java
+++ b/bson/src/main/org/bson/types/ObjectId.java
@@ -57,7 +57,7 @@ public final class ObjectId implements Comparable<ObjectId>, Serializable {
     private static final int RANDOM_VALUE1;
     private static final short RANDOM_VALUE2;
 
-    private static final AtomicInteger NEXT_COUNTER;
+    private static final AtomicInteger NEXT_COUNTER = new AtomicInteger(new SecureRandom().nextInt());
 
     private static final char[] HEX_CHARS = {
             '0', '1', '2', '3', '4', '5', '6', '7',
@@ -409,7 +409,6 @@ public final class ObjectId implements Comparable<ObjectId>, Serializable {
             SecureRandom secureRandom = new SecureRandom();
             RANDOM_VALUE1 = secureRandom.nextInt(0x01000000);
             RANDOM_VALUE2 = (short) secureRandom.nextInt(0x00008000);
-            NEXT_COUNTER = new AtomicInteger(secureRandom.nextInt());
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
--- a/bson/src/main/org/bson/types/ObjectId.java
+++ b/bson/src/main/org/bson/types/ObjectId.java
@@ -57,7 +57,7 @@ public final class ObjectId implements Comparable<ObjectId>, Serializable {
     private static final int RANDOM_VALUE1;
     private static final short RANDOM_VALUE2;
 
-    private static final AtomicInteger NEXT_COUNTER = new AtomicInteger(new SecureRandom().nextInt());
+    private static final AtomicInteger NEXT_COUNTER;
 
     private static final char[] HEX_CHARS = {
             '0', '1', '2', '3', '4', '5', '6', '7',
@@ -409,6 +409,7 @@ public final class ObjectId implements Comparable<ObjectId>, Serializable {
             SecureRandom secureRandom = new SecureRandom();
             RANDOM_VALUE1 = secureRandom.nextInt(0x01000000);
             RANDOM_VALUE2 = (short) secureRandom.nextInt(0x00008000);
+            NEXT_COUNTER = new AtomicInteger(secureRandom.nextInt());
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
--- a/config/spotbugs/exclude.xml
+++ b/config/spotbugs/exclude.xml
@@ -121,7 +121,7 @@
     </Match>
 
     <Match>
-        <!-- MongoDB status: "No Fix Needed", SpotBugs rank: 18 -->
+        <!-- MongoDB status: "False Positive", SpotBugs rank: 18 -->
         <!-- Spotbugs assumes that SSLSocket#getSSLParameters never returns null, when that is not the case for all JDKs -->
         <Class name="com.mongodb.internal.connection.SslHelper"/>
         <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE"/>
--- a/config/spotbugs/exclude.xml
+++ b/config/spotbugs/exclude.xml
@@ -142,7 +142,7 @@
     </Match>
 
     <Match>
-        <!-- MongoDB status: "TODO Ross", SpotBugs rank: 11 -->
+        <!-- MongoDB status: "False Positive", SpotBugs rank: 11 -->
         <Class name="org.bson.codecs.kotlin.DataClassCodec$Companion"/>
         <Method name="getCodec"/>
         <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE"/>
--- a/config/spotbugs/exclude.xml
+++ b/config/spotbugs/exclude.xml
@@ -14,7 +14,7 @@
   ~ limitations under the License.
   -->
 
-<!-- See https://github.com/spotbugs/spotbugs/blob/master/docs/filter.rst#rank
+<!-- See https://spotbugs.readthedocs.io/en/latest/filter.html#rank
     for the mapping from SpotBugs rank to verbal labels.
 -->
 <FindBugsFilter>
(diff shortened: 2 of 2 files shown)


config/spotbugs/exclude.xml
```diff
@@ -14,251 +14,194 @@
...
+-->
 <FindBugsFilter>
-    <!-- specific excludes which should be fixed -->
     <Match>
-        <Class name="org.bson.JSONWriterSettings"/>
-    </Match>
-    <Match>
-        <Bug code="EI,EI2"/>
+        <!-- MongoDB status: "No Fix Needed", SpotBugs rank: 18 -->
+        <Bug pattern="EI_EXPOSE_REP,EI_EXPOSE_REP2"/>
```
stIncMale, line 23: Used a mor specific filter here than the previous `code="EI,EI2"`.

config/spotbugs/exclude.xml
```diff
@@ -14,251 +14,194 @@
...
-
-    <Match>
-        <Class name="com.mongodb.DatabaseTestCase"/>
-        <Bug pattern="MS_PKGPROTECT"/>
-    </Match>
-
-    <Match>
-        <Class name="org.bson.types.StringRangeSet$NumberStringComparator"/>
-        <Bug pattern="SE_COMPARATOR_SHOULD_BE_SERIALIZABLE"/>
-    </Match>
```
stIncMale, line 99: I wonder if when building with JDK 8, we will still get these findings.

bson/src/main/org/bson/types/ObjectId.java
stIncMale, line 1: I thought this change will resolve at least one finding - it didn't. Yet, the proposed code seems more reasonable.

config/spotbugs/exclude.xml
stIncMale: Ah, we don't seem to run static checks with JDK 8.

config/spotbugs/exclude.xml
```diff
@@ -14,251 +14,186 @@
...
         <Method name="retain"/>
         <Bug pattern="RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT"/>
     </Match>
 
-    <!-- Caused by Java 11 generating a null check on try with resources.
-         See https://github.com/spotbugs/spotbugs/issues/756 for details. -->
     <Match>
+        <!-- MongoDB status: "TODO Ross", SpotBugs rank: 11 -->
+        <Class name="org.bson.codecs.kotlin.DataClassCodec$Companion"/>
+        <Method name="getCodec"/>
```
stIncMale, line 147: @rozza, could you please look at this Kotlin finding and tell whether it is a "False Positive" or rather "No Fix Needed" one?

config/spotbugs/exclude.xml
```diff
@@ -14,251 +14,194 @@
...
-    </Match>
-
-    <Match>
-        <Class name="~com.mongodb.async.client.internal.AsyncCryptConnection.*"/>
-        <Bug pattern="RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT"/>
-    </Match>
-
-    <!-- Spotbugs assumes that SSLSocket#getSSLParameters never returns null, when that is not the case for all JDKs -->
-    <Match>
+        <!-- MongoDB status: "No Fix Needed", SpotBugs rank: 18 -->
```
jyemin, line 124: Above on line 55 the same condition is marked as "False Positive".  That seems more correct.

config/spotbugs/exclude.xml
```diff
@@ -14,251 +14,194 @@
...
     <Match>
+        <!-- MongoDB status: "False Positive", SpotBugs rank: 17 -->
         <Class name="com.mongodb.client.internal.CryptConnection"/>
         <Method name="retain"/>
         <Bug pattern="RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT"/>
     </Match>
 
     <Match>
-        <Class name="~com.mongodb.reactivestreams.client.internal.crypt.*"/>
+        <!-- MongoDB status: "False Positive", SpotBugs rank: 17 -->
```
jyemin, line 138: `No Fix Needed` seems better.

config/spotbugs/exclude.xml
```diff
@@ -14,251 +14,194 @@
...
+        <Class name="com.mongodb.reactivestreams.client.internal.crypt.CryptConnection"/>
         <Method name="retain"/>
```
jyemin, line 140: Is this exclusion necessary any more?  Looking at all calls to `ReferenceCounted#retain` almost all of them ignore the return value, so not clear to me what's different about this one.

config/spotbugs/exclude.xml
```diff
@@ -14,251 +14,194 @@
...
         <Bug pattern="RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT"/>
     </Match>
 
-    <!-- Caused by Java 11 generating a null check on try with resources.
-         See https://github.com/spotbugs/spotbugs/issues/756 for details. -->
     <Match>
+        <!-- MongoDB status: "TODO Ross", SpotBugs rank: 11 -->
+        <Class name="org.bson.codecs.RepresentationConfigurable"/>
+        <Method name="withRepresentation"/>
+        <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE"/>
```
jyemin: I can't tell what this is warning about, as this is an interface method

jyemin (changes requested):

config/spotbugs/exclude.xml
stIncMale: This one was accidentally not removed previously. It's not longer in the PR: https://github.com/mongodb/mongo-java-driver/pull/1392/commits/d06e1ff418ab2a03c138b9cb1edaeac622798f0b.

config/spotbugs/exclude.xml
stIncMale: It is needed. I don't know why SpotBugs is so sure that the `retain` method has no side-effects.

config/spotbugs/exclude.xml
stIncMale: I think this is actually "False Positive", because the finding description is _"This code calls a method and ignores the return value. However our analysis shows that the method (including its implementations in subclasses if any) **does not produce any effect other than return value**. Thus this call can be removed."_ The highlighted part is false.

bson/src/main/org/bson/types/ObjectId.java
```diff
@@ -409,6 +409,7 @@ private Object readResolve() {
             SecureRandom secureRandom = new SecureRandom();
             RANDOM_VALUE1 = secureRandom.nextInt(0x01000000);
             RANDOM_VALUE2 = (short) secureRandom.nextInt(0x00008000);
+            NEXT_COUNTER = new AtomicInteger(secureRandom.nextInt());
```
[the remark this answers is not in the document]
stIncMale: Rolled back in https://github.com/mongodb/mongo-java-driver/pull/1392/commits/d949b608f5cb12b4532d36d22132bd89197c1e77.

config/spotbugs/exclude.xml
stIncMale: Yes, done in https://github.com/mongodb/mongo-java-driver/pull/1392/commits/d949b608f5cb12b4532d36d22132bd89197c1e77.

config/spotbugs/exclude.xml
jyemin: And only in this one place.  That's strange.

jyemin (approved):
LGTM

config/spotbugs/exclude.xml
rozza: So this is a legit bug, not sure why it wasn't picked up by Spotbugs before.  Created a fix in #1395

rozza (approved):
LGTM - not sure if you want #1395 merged first. If you are happy with the PR feel free to merge.

config/spotbugs/exclude.xml
stIncMale: It was picked up, but it was ignored because we were ignoring all `RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE` findings, due to there being too many false positives related to try-with-resources statements. We don't have those false positives anymore (I am guessing, that's because we are not running SpotBugs with JDK 8 anymore, SpotBugs looks at the bytecode, and the bytecode denerated by different `javac` versions is different), I removed the corresponding exclusion, and we now can focus on those more reasonable `RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE` findings.

stIncMale: I prefer this PR wait for https://github.com/mongodb/mongo-java-driver/pull/1395 to be merged in to `master, then incorporate the change from `master`, get rid of the corresponding SpotBugs exclusion, then be merged.

jyemin: This seems like a False Positive.  The CodecRegistry contract requires that a `CodecConfigurationException` is thrown, and null is not returned, and there are many seemingly identical matches that SpotBugs does not report, e.g. `RecordCodec.ComponentModel#computeCodec`.

PR #1395, opened 2024-05-17
Ensure bson-kotlin handles possible NPE

rozza:
If the user provides their own CodecRegistry it could return `null` from CodecRegistry#get(clazz)

JAVA-5477

### DIFF
--- a/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt
+++ b/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt
@@ -236,6 +236,8 @@ internal data class DataClassCodec<T : Any>(
                 } else {
                     this.get(clazz, types)
                 }
+            codec ?: throw CodecConfigurationException("Can't find a codec for $clazz.")
+
             return kParameter.findAnnotation<BsonRepresentation>()?.let {
                 if (codec !is RepresentationConfigurable<*>) {
                     throw CodecConfigurationException(
--- a/bson-kotlin/src/test/kotlin/org/bson/codecs/kotlin/DataClassCodecTest.kt
+++ b/bson-kotlin/src/test/kotlin/org/bson/codecs/kotlin/DataClassCodecTest.kt
@@ -19,10 +19,12 @@ import kotlin.test.assertEquals
 import org.bson.BsonDocument
 import org.bson.BsonDocumentReader
 import org.bson.BsonDocumentWriter
+import org.bson.codecs.Codec
 import org.bson.codecs.DecoderContext
 import org.bson.codecs.EncoderContext
 import org.bson.codecs.configuration.CodecConfigurationException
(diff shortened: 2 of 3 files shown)


bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt
```diff
@@ -236,6 +236,8 @@ internal data class DataClassCodec<T : Any>(
                 } else {
                     this.get(clazz, types)
                 }
+            codec ?: throw CodecConfigurationException("Can't find a codec for $clazz.")
```
stIncMale, line 239: I think we should not do this ad hoc check. Both of the `CodecRegistry.get` methods don't allow returning `null`s, so when a custom `CodecRegistry.get` returns `null`, it violates its contract. There are three ways we can go about that:

1. [the current approach in the driver, though I can't guarantee that it is used consistently] Do nothing, assuming `null`s can't be returned, but also not asserting that values are non-`null`, because we use assertions only for driver bugs, and here a bug may be in the application code.
2. [the approach proposed in this PR, but the PR applies it to a single place of many] Everywhere we call `CodecRegistry.get`, we check if the result is `null`, and do something with it. This approach is error-prone, and we should avoid it.
3. [the approach I propose] Wrap all application-provided `CodecRegistry` objects in `CheckedCodecRegistry`, which checks the return value of the `get` methods for `null`, and either converts it into the `CodecConfigurationException` with accordance to the contract, or throws an NPE, informing an application that it violated the contract at runtime.
  3.1. If we were just now designing this API, I would have suggested the "throw NPE from `CheckedCodecRegistry`" approach. However, since we already have it implemented, and there is a chance we sometimes may do the ad hoc checks somewhere in the driver, I propose to use the "convert `null` into the `CodecConfigurationException`" approach.
  3.2 Following are all the places where a user may specify a custom `CodecRegistry`:
    * `MongoClientSettings.Builder.codecRegistry`
    * `MongoCollection.withCodecRegistry`
    * `CodecRegistries.fromRegistries`
  3.3 We can optimize wrapping by omitting it when the instance given to us is of the concrete type we own and know users may not extend: `ChildCodecRegistry`, `ProvidersCodecRegistry`, `CheckedCodecRegistry`.

stIncMale (changes requested):

bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt
rozza: The adhoc check is needed to inform the user if they are able to encode / decode the Kotlin DataClass and all its properties (fields). If it were to return `null` it means that there is no `Codec` available in the registry for the type.

Note: `CodecRegistry.getCodec` here is an extension function to `CodecRegistry` and is only for the use within the `DataClassCodec`. The call to `this.get` is a call to  `codecRegistry#get` which uses the users configured `CodecRegistry`.

PR #1396, opened 2024-05-18
Java 5474

vbabanin:


driver-core/src/main/com/mongodb/internal/connection/AbstractProtocolExecutor.java
```diff
@@ -0,0 +1,42 @@
...
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public abstract class AbstractProtocolExecutor implements ProtocolExecutor {
+
+    protected boolean isMongoSocketException(final Throwable e) {
+        return e instanceof MongoSocketException;
+    }
+
+    protected boolean isOperationTimeoutFromSocketException(final Throwable e) {
+        return e instanceof MongoOperationTimeoutException && e.getCause() instanceof MongoSocketException;
```
rozza, line 33: Can `e.getCause()` return null?

driver-core/src/main/com/mongodb/internal/connection/AbstractProtocolExecutor.java
```diff
@@ -0,0 +1,42 @@
...
+package com.mongodb.internal.connection;
+
+import com.mongodb.MongoOperationTimeoutException;
+import com.mongodb.MongoSocketException;
+import com.mongodb.internal.session.SessionContext;
+
+/**
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public abstract class AbstractProtocolExecutor implements ProtocolExecutor {
```
rozza, line 26: Does this abstract class need to exist? Can these helper methods just be added to `ProtocolExecutor`?

driver-core/src/test/functional/com/mongodb/internal/operation/AsyncCommandBatchCursorTest.java
```diff
@@ -0,0 +1,186 @@
...
+        });
+
+        commandBatchCursor.close();
+
+        //then
+        verify(mockConnection, times(1)).commandAsync(any(),
+                any(), any(), any(), any(), any(), any());
+        verify(mockConnection, times(1)).commandAsync(eq(NAMESPACE.getDatabaseName()),
+                argThat(bsonDocument -> bsonDocument.containsKey("getMore")), any(), any(), any(), any(), any());
+        verify(mockConnection, never()).commandAsync(eq(NAMESPACE.getDatabaseName()),
```
rozza, line 169: 👍

rozza (commented):
One question - but looking good.

driver-core/src/main/com/mongodb/internal/connection/AbstractProtocolExecutor.java
vbabanin: Yes, it can return null in cases when `MongoOperationTimeoutException` was not caused by any [blocking operation ](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#blocking-sections-for-operation-execution).

However, `instanceOf `will return `false` in this case.

driver-core/src/main/com/mongodb/internal/connection/AbstractProtocolExecutor.java
vbabanin: Technically, the methods could be moved to the interface since they are stateless, but this would make them public. I've opted to keep them `protected` within the abstract class to maintain better encapsulation and clearer boundaries in our codebase. If you feel that we should move them to the interface, I am open to doing that. Please let me know your thoughts.

I've also relocated the `isMongoSocketException `and `isOperationTimeoutFromSocketException` to `ExceptionUtils` for broader reuse.

rozza: This is all `internal` code, it just adds an extra layer for helper style methods. Looks like this abstract class could be removed and `shouldMarkSessionDirty` be moved to a helper.

However, if you are keen to keep it protected then its ok with me.

rozza (changes requested):
Just a checkstyle issue to be fixed in the tests.

rozza (approved):
LGTM!

PR #1398, opened 2024-05-22
Upgrade to Spotbugs 6.0.15

jyemin:
* Exclude CT_CONSTRUCTOR_THROW
* Exclude PA_PUBLIC_PRIMITIVE_ATTRIBUTE
* Add assertion to handle NP warning

JAVA-5480

### DIFF
--- a/config/spotbugs/exclude.xml
+++ b/config/spotbugs/exclude.xml
@@ -19,6 +19,17 @@
 -->
 <FindBugsFilter>
     <Match>
+        <!-- MongoDB status: "No Fix Needed", SpotBugs rank: * -->
+        <Bug pattern="CT_CONSTRUCTOR_THROW"/>
+    </Match>
+
+    <Match>
+        <!-- MongoDB status: "No Fix Needed", SpotBugs rank: * -->
+        <Class name="com.mongodb.internal.connection.tlschannel.impl.BufferHolder"/>
+        <Bug pattern="PA_PUBLIC_PRIMITIVE_ATTRIBUTE"/>
+    </Match>
+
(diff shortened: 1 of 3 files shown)


config/spotbugs/exclude.xml
```diff
@@ -18,6 +18,17 @@
     for the mapping from SpotBugs rank to verbal labels.
 -->
 <FindBugsFilter>
+    <Match>
+        <!-- MongoDB status: "No Fix Needed", SpotBugs rank: * -->
```
jyemin, line 22: @stIncMale how did you determine the ranks?  Above comment says to look at https://spotbugs.readthedocs.io/en/latest/filter.html#rank but I don't see anything useful at that location.

stIncMale: I determined the ranks as described in https://github.com/mongodb/mongo-java-driver/pull/1392:

> To get the rank data, I had enabled SpotBugs XML reports, then was commenting out parts of exclude.xml and manually get the rank data from the XML reports.

stIncMale: > Above comment says to look at https://spotbugs.readthedocs.io/en/latest/filter.html#rank but I don't see anything useful at that location.

It says to look at the page "for the mapping from SpotBugs rank to verbal labels", and the mapping is there. This is in case we decide to map the numbers to something more human-friendly.

jyemin: Ah, got it. I added the ranks and a comment saying how to find them, for our future selves.

stIncMale (approved):

PR #1400, opened 2024-05-24
JAVA-5479: Add GitHub Actions based release automation

alcaeus:
This PR adds a GitHub Actions Workflow to create releases. The process is as follows:

* Users with write access trigger a new workflow run for the "Release new version" workflow, entering the version to be released
* The automation does some sanity checks on the branch name:
** Release `<major>.<minor>.0` must be released from the maintenance branch (i.e. `<major>.<minor>.x`) or the `master` branch. If the `master` branch is selected, the automation will attempt to create a new maintenance branch
** Release `<major>.<minor>.<patch>` must be released from the maintenance branch (i.e. `<major>.<minor>.x`) - releasing from any other branches is not allowed
* The automation then updates the build version to `<major>.<minor>.<patch>` in `build.gradle`, commits this change and creates a tag
* After tagging, the version in `build.gradle` is updated to `<major>.<minor>.<patch+1>-SNAPSHOT` and again committed
* The maintenance branch and tag are pushed to the repository
* A draft release with generated release notes is created

I've tested the flow in my fork where I've confirmed it works. However, a couple of changes may be appropriate still:
- The automation currently uses the GitHub token of the user triggering the release. Since commits are being pushed directly, branch protection rules may interfere with the release process. In this case, we should add the `mongodb-dbx-release-bot` app to the repository and generate a token for the app. We can then allow the app to bypass branch protections
- If not all users with write access should have permission to create releases, we can set up an environment in GitHub settings and require that an authorised user approves the workflow run
- The release commits and tag are not signed; this could be added using the [git-sign action](https://github.com/mongodb-labs/drivers-github-tools/tree/main?tab=readme-ov-file#git-sign)
- Other release requirements related to the SSDLC effort (e.g. generating a SAST report, uploading SBOM, etc.) are not added yet and can be integrated based on the shared driver's GitHub Actions once they are ready.

Any additional build tasks (e.g. creating a package to upload to a package manager) can be done in a separate workflow that runs when a tag is pushed.

### DIFF
--- a/.github/workflows/bump-and-tag.sh
+++ b/.github/workflows/bump-and-tag.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-# TODO: should this always assume that release X.Y.Z will always be created from X.Y.Z-SNAPSHOT"?
+# This script assumes that release X.Y.Z will always be created from X.Y.Z-SNAPSHOT"
 echo "Replace snapshot version with release version ${RELEASE_VERSION} in build.gradle"
 sed --in-place "s/version = '.*-SNAPSHOT'/version = '${RELEASE_VERSION}'/g" build.gradle
 
--- a/.github/workflows/bump-and-tag.sh
+++ b/.github/workflows/bump-and-tag.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+set -e
+
+# TODO: should this always assume that release X.Y.Z will always be created from X.Y.Z-SNAPSHOT"?
+echo "Replace snapshot version with release version ${RELEASE_VERSION} in build.gradle"
+sed --in-place "s/version = '.*-SNAPSHOT'/version = '${RELEASE_VERSION}'/g" build.gradle
+
+echo "Create package commit for ${RELEASE_VERSION}"
+git commit -m "Version: bump ${RELEASE_VERSION}" build.gradle
+
+echo "Create release tag for ${RELEASE_VERSION}"
+git tag -a -m "${RELEASE_VERSION}" r${RELEASE_VERSION}
+
+echo "Bump to snapshot version for ${NEXT_VERSION}"
+sed --in-place "s/version = '${RELEASE_VERSION}'/version = '${NEXT_VERSION}-SNAPSHOT'/g" build.gradle
+
+echo "Create commit for version bump to ${NEXT_VERSION}"
+git commit -m "Version: bump ${NEXT_VERSION}-SNAPSHOT" build.gradle
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,6 +17,7 @@ env:
 
 jobs:
   prepare-release:
+    environment: release
     name: "Prepare release"
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+          echo '🆕 Creating new release branch ${RELEASE_BRANCH} from ${{ github.ref_name }}' >> $GITHUB_STEP_SUMMARY
+          git checkout -b ${RELEASE_BRANCH}
+
+      - name: "Set git author information"
+        run: |
+          git config user.name "${GIT_AUTHOR_NAME}"
+          git config user.email "${GIT_AUTHOR_EMAIL}"
+
+      # This step bumps version numbers in build.gradle and creates git artifacts for the release
+      # TODO: this could leverage the git-sign action with some minor changes
```
jyemin, line 82: Let's defer.

.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "The version to be released (e.g. 1.2.3)"
+        required: true
+        type: "string"
+
+env:
+  GH_TOKEN: ${{ github.token }}
+  # TODO: Adding the mongodb-dbx-release-automation app to the repository will allow fetching a one-time token and pushing
```
jyemin, line 14: This seems like a necessary thing to enable in the initial version, as we do have branch protection rules enabled.

.github/workflows/bump-and-tag.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -e
+
+# TODO: should this always assume that release X.Y.Z will always be created from X.Y.Z-SNAPSHOT"?
```
jyemin, line 4: I believe our current scripts assume this.  Seems safe.

.github/workflows/release.yml
alcaeus: Correct. In that case, you should follow [these instructions (internal link)](https://wiki.corp.mongodb.com/display/SYSENG/GitHub#GitHub-RequestingaGitHubApplication) to request the `mongodb-dbx-release-bot` application to be added to all relevant repositories.

.github/workflows/release.yml
alcaeus: Removed the TODO for now.

.github/workflows/release.yml
jyemin: Requested

jyemin:
Is there anything the blocks taking this out of Draft?

alcaeus:
> Is there anything the blocks taking this out of Draft?

Not really. I've rebased on top of master and added the release environment, which allows us to better protect the necessary credentials and also allows for adding additional checks (such as directly restricting which branches can be released).
Once the app has been added to the repository, we're good to start testing.

.github/workflows/release.yml
rozza: 🍿 Is this the last blocker?

alcaeus: It is - just waiting for our IT to enable the app here, then we're good to go :)

jyemin (approved):
LGTM.  Thank you!

PR #1401, opened 2024-05-27
Add `ssdlc-report.sh` that uses SpotBugs to create SARIF files

stIncMale:
In the future this script may do more work for us.

JAVA-5431

### DIFF
--- a/.evergreen/ssdlc-report.sh
+++ b/.evergreen/ssdlc-report.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -o errexit
+
+############################################
+#            Main Program                  #
+############################################
+RELATIVE_DIR_PATH="$(dirname "${BASH_SOURCE[0]:-$0}")"
+source "${RELATIVE_DIR_PATH}/javaConfig.bash"
+
+echo "Creating SSLDC reports"
+./gradlew -version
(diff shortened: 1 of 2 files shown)


.evergreen/ssdlc-report.sh
```diff
@@ -0,0 +1,15 @@
...
+
+############################################
+#            Main Program                  #
+############################################
+RELATIVE_DIR_PATH="$(dirname "${BASH_SOURCE[0]:-$0}")"
+source "${RELATIVE_DIR_PATH}/javaConfig.bash"
+
+echo "Creating SSLDC reports"
+./gradlew -version
+./gradlew -PssdlcReport.enabled=true --continue -x test -x integrationTest -x spotlessApply clean check scalaCheck kotlinCheck testClasses || true
```
stIncMale, line 13: This command most likely fails due to SpotBugs findings, because it ignores `spotbugs/exclude.xml`. To prevent the whole script from failing, we use `|| true`.

jyemin:
How do you envision this script being used in the future (as currently there is no Evergreen task that executes it)?

I could see it being called as part of `publish-release` and the SARIF files being uploaded to S3.

jyemin (approved):

PR #1408, opened 2024-06-06
Disallow comma character in authMechanismProperties

katcharov:
JAVA-5486


jyemin (approved):
I compared these changes to the commit before the one for OIDC, and it looks good.  The only thing that remains that is not directly related to OIDC is the addition of the `limit` parameter (with value `2`) to `String.split`.  

Assuming that is your intention, then LGTM.

katcharov:
Yes, intentional. The split/2 allows a colon to be used in option key-value values, like in [this test](https://github.com/mongodb/mongo-java-driver/pull/1408/files#diff-46bac68daec38295a33abfd5a7f427d891fc2824c3c254d9f78b16de2fa4c574R26).

PR #1418, opened 2024-06-11
Remove hasExpired and expiry checks in retries

katcharov:
JAVA-5379

This removes `hasExpired`, and expiry checks in retries (retries rely exclusively on timeout exceptions being thrown by blocking operations, to signal that the retry should halt).

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/async/function/RetryState.java
+++ b/driver-core/src/main/com/mongodb/internal/async/function/RetryState.java
@@ -197,8 +197,7 @@ public final class RetryState {
          * A MongoOperationTimeoutException indicates that the operation timed out, either during command execution or server selection.
          * The timeout for server selection is determined by the computedServerSelectionMS = min(serverSelectionTimeoutMS, timeoutMS).
          *
-         * The isLastAttempt() method checks if the timeoutMS has expired, which could be greater than the computedServerSelectionMS.
-         * Therefore, it's important to check if the exception is an instance of MongoOperationTimeoutException to detect a timeout.
+         * It is important to check if the exception is an instance of MongoOperationTimeoutException to detect a timeout.
          */
         if (isLastAttempt() || attemptException instanceof MongoOperationTimeoutException) {
             previouslyChosenException = newlyChosenException;
--- a/driver-core/src/main/com/mongodb/internal/async/function/RetryState.java
+++ b/driver-core/src/main/com/mongodb/internal/async/function/RetryState.java
@@ -51,8 +51,7 @@ public final class RetryState {
 
     private final LoopState loopState;
     private final int attempts;
-    @Nullable
-    private final TimeoutContext timeoutContext;
+    private final boolean retryUntilTimeoutThrowsException;
     @Nullable
     private Throwable previouslyChosenException;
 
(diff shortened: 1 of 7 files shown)


driver-core/src/test/unit/com/mongodb/internal/async/function/RetryStateTest.java
```diff
@@ -260,77 +177,14 @@ void breakAndCompleteIfRetryAndPredicateThrows(final TimeoutContext timeoutConte
...
-        MongoOperationTimeoutException actualTimeoutException =
-                assertThrows(expectedTimeoutException.getClass(), () -> retryState.advanceOrThrow(expectedTimeoutException,
-                        (e1, e2) -> expectedTimeoutException,
-                        (rs, e) -> false));
-
-        Assertions.assertEquals(actualTimeoutException, expectedTimeoutException);
-        Assertions.assertNull(actualTimeoutException.getCause(),
-                "Original timeout exception should not have a cause if transformer already returned timeout exception.");
-    }
-
```
katcharov, line 333: Note the removal of the above 3 tests. I have removed all references to timeouts from RetryState and these tests: RetryState should now not be aware of timeouts, and should have no special behaviours associated with timeouts. Please confirm that this is expected.

driver-core/src/main/com/mongodb/internal/async/function/RetryState.java
```diff
@@ -203,15 +193,6 @@ private void doAdvanceOrThrow(final Throwable attemptException,
          */
         if (isLastAttempt() || attemptException instanceof MongoOperationTimeoutException) {
             previouslyChosenException = newlyChosenException;
-            /*
-             * The function of isLastIteration() is to indicate if retrying has been explicitly halted. Such a stop is not interpreted as
-             * a timeout exception but as a deliberate cessation of retry attempts.
-             */
-            if (hasTimeoutMs() && !loopState.isLastIteration()) {
```
katcharov, line 210: The logic for removing these is that it is safe to behave as if a timeout has not expired (even if it has), because it will be detected and thrown by an ensuing blocking operation.

This relies on the assumption that all blocking operations will throw when given an expired timeout (this should be the case, because all blocking operations should now be wrapped in Timeout's branching "call" methods).

driver-core/src/test/unit/com/mongodb/internal/async/function/RetryStateTest.java
katcharov: Reverted

driver-core/src/main/com/mongodb/internal/async/function/RetryState.java
katcharov: Tests were failing. The logic related to timeout exceptions in RetryState is necessary, but complicated to extract. I've restored it, but the isExpired check is still removed (so this still relies on the ensuing blocking operation throwing a timeout exception).

driver-core/src/main/com/mongodb/internal/async/function/RetryState.java
```diff
@@ -204,10 +203,11 @@ private void doAdvanceOrThrow(final Throwable attemptException,
         if (isLastAttempt() || attemptException instanceof MongoOperationTimeoutException) {
```
vbabanin, line 203: As we discussed on the call, only the following comment should be retained above this code block:

'It's important to check if the exception is an instance of MongoOperationTimeoutException to detect a timeout.'

driver-core/src/test/unit/com/mongodb/internal/async/function/RetryStateTest.java
```diff
@@ -439,30 +397,6 @@ void advanceOrThrowTransformAfterFirstAttempt(final TimeoutContext timeoutContex
                 }));
     }
 
-    @Test
-    void advanceOrThrowTransformThrowsTimeoutExceptionAfterFirstAttempt() {
```
vbabanin, line 443: This test asserts the behavior of error transformation. Despite the removal of timeout checks within the `RetryState`, we should ensure that the logic related to error transformation is still tested.

vbabanin: As discussed on the call, those tests will be ignored and updated later.

vbabanin (approved):

PR #1420, opened 2024-06-12
Forward slash in connection string is optional 

jyemin:
JAVA-5166

### DIFF
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -368,16 +368,19 @@ public class ConnectionString {
 
         // Split out the user and host information
         String userAndHostInformation;
-        int idx = unprocessedConnectionString.indexOf("/");
-        if (idx == -1) {
-            if (unprocessedConnectionString.contains("?")) {
-                throw new IllegalArgumentException("The connection string contains options without trailing slash");
+        int firstForwardSlashIdx = unprocessedConnectionString.indexOf("/");
+        int firstQuestionMarkIdx = unprocessedConnectionString.indexOf("?");
+        if (firstForwardSlashIdx == -1 || (firstQuestionMarkIdx != -1 && firstQuestionMarkIdx < firstForwardSlashIdx)) {
+            if (firstQuestionMarkIdx == -1) {
+                userAndHostInformation = unprocessedConnectionString;
+                unprocessedConnectionString = "";
+            } else {
+                userAndHostInformation = unprocessedConnectionString.substring(0, firstQuestionMarkIdx);
+                unprocessedConnectionString = unprocessedConnectionString.substring(firstQuestionMarkIdx);
             }
-            userAndHostInformation = unprocessedConnectionString;
-            unprocessedConnectionString = "";
         } else {
-            userAndHostInformation = unprocessedConnectionString.substring(0, idx);
-            unprocessedConnectionString = unprocessedConnectionString.substring(idx + 1);
+            userAndHostInformation = unprocessedConnectionString.substring(0, firstForwardSlashIdx);
+            unprocessedConnectionString = unprocessedConnectionString.substring(firstForwardSlashIdx + 1);
         }
 
         // Split the user and host information
@@ -385,7 +388,7 @@ public class ConnectionString {
         String hostIdentifier;
         String userName = null;
         char[] password = null;
-        idx = userAndHostInformation.lastIndexOf("@");
+        int idx = userAndHostInformation.lastIndexOf("@");
         if (idx > 0) {
             userInfo = userAndHostInformation.substring(0, idx).replace("+", "%2B");
             hostIdentifier = userAndHostInformation.substring(idx + 1);
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -370,14 +370,13 @@ public class ConnectionString {
         String userAndHostInformation;
         int firstForwardSlashIdx = unprocessedConnectionString.indexOf("/");
         int firstQuestionMarkIdx = unprocessedConnectionString.indexOf("?");
-        if (firstForwardSlashIdx == -1 || (firstQuestionMarkIdx != -1 && firstQuestionMarkIdx < firstForwardSlashIdx)) {
-            if (firstQuestionMarkIdx == -1) {
-                userAndHostInformation = unprocessedConnectionString;
-                unprocessedConnectionString = "";
-            } else {
-                userAndHostInformation = unprocessedConnectionString.substring(0, firstQuestionMarkIdx);
-                unprocessedConnectionString = unprocessedConnectionString.substring(firstQuestionMarkIdx);
-            }
+        if (firstQuestionMarkIdx == -1 && firstForwardSlashIdx == -1) {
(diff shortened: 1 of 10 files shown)


driver-core/src/test/unit/com/mongodb/ConnectionStringSpecification.groovy
```diff
@@ -141,6 +141,29 @@ class ConnectionStringSpecification extends Specification {
                 .withWTimeout(5, MILLISECONDS).withJournal(true)
     }
 
+    @Unroll
```
jyemin, line 144: I wasn't confident in the test coverage from the spec so I added a few more tests here.

driver-core/src/test/resources/connection-string/valid-options.json
```diff
@@ -21,9 +21,26 @@
...
+      ],
+      "auth": null,
+      "options": {
+        "tls": true
+      }
+    },
     {
       "description": "Colon in a key value pair",
-      "uri": "mongodb://example.com/?authMechanism=MONGODB-OIDC&authMechanismProperties=TOKEN_RESOURCE:mongodb://test-cluster",
+      "uri": "mongodb://example.com?authMechanism=MONGODB-OIDC&authMechanismProperties=TOKEN_RESOURCE:mongodb://test-cluster",
```
jyemin, line 43: This works now

driver-core/src/test/resources/connection-string/valid-warnings.json
```diff
@@ -63,6 +63,51 @@
       "options": {
         "wtimeoutms": 10
       }
+    },
+    {
+      "description": "Empty integer option values are ignored",
```
jyemin, line 68: This throws an exception.  I don't think it's worth fixing so just skipping the test in code.

driver-core/src/test/resources/connection-string/valid-options.json
katcharov: Updated in https://github.com/mongodb/specifications/pull/1599 to include the slash (so that it tests only one thing - test above covers the missing `/`). Update valid-warnings/options files to latest?

katcharov (changes requested):

driver-core/src/test/resources/connection-string/valid-options.json
jyemin: Funny, because this is the one spec test that led me to check whether the first forward slash precedes the first question mark.  

But I'll change it back since we have the new ConnectionStringSpecification tests that test that code.

driver-core/src/test/unit/com/mongodb/ConnectionStringSpecification.groovy
```diff
@@ -141,6 +141,29 @@ class ConnectionStringSpecification extends Specification {
...
+        new ConnectionString('mongodb://mongodb.com/?')                  | null    | null
+        new ConnectionString('mongodb://mongodb.com/test')               | null    | "test"
+        new ConnectionString('mongodb://mongodb.com/test?')              | null    | "test"
+        new ConnectionString('mongodb://mongodb.com/?appName=a1')        | "a1"    | null
+        new ConnectionString('mongodb://mongodb.com?appName=a1')         | "a1"    | null
+        new ConnectionString('mongodb://mongodb.com/?appName=a1/a2')     | "a1/a2" | null
+        new ConnectionString('mongodb://mongodb.com?appName=a1/a2')      | "a1/a2" | null
+        new ConnectionString('mongodb://mongodb.com/test?appName=a1/a2') | "a1/a2" | "test"
+        new ConnectionString('mongodb://mongodb.com/test?appName=a1')    | "a1"    | "test"
+        new ConnectionString('mongodb://mongodb.com/test?appName=a1/a2') | "a1/a2" | "test"
```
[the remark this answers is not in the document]
jyemin: I think a duplicate, as I can't think of another test that I intended to add.  I removed it, but lmk if you can think of any other good cases.

driver-core/src/test/resources/connection-string/valid-options.json
jyemin: Done

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -368,24 +368,27 @@ public ConnectionString(final String connectionString, @Nullable final DnsClient
...
+                unprocessedConnectionString = unprocessedConnectionString.substring(firstQuestionMarkIdx);
             }
-            userAndHostInformation = unprocessedConnectionString;
-            unprocessedConnectionString = "";
         } else {
-            userAndHostInformation = unprocessedConnectionString.substring(0, idx);
-            unprocessedConnectionString = unprocessedConnectionString.substring(idx + 1);
+            userAndHostInformation = unprocessedConnectionString.substring(0, firstForwardSlashIdx);
+            unprocessedConnectionString = unprocessedConnectionString.substring(firstForwardSlashIdx + 1);
         }
```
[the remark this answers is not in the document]
jyemin: Much clearer, thanks!  

Done.

driver-core/src/test/resources/connection-string/valid-warnings.json
```diff
@@ -63,6 +63,51 @@
...
+          "host": "localhost",
+          "port": null
+        }
+      ],
+      "auth": null,
+      "options": null
+    },
+    {
+      "description": "Comma in a key value pair causes a warning",
+      "uri": "mongodb://localhost?authMechanism=MONGODB-OIDC&authMechanismProperties=TOKEN_RESOURCE:mongodb://host1%2Chost2",
```
jyemin, line 99: This test now fails so it's skipped in code.  Previously it was only passing due to the missing `authMechanism` property, which was causing the `authMechanismProperties` to be ignored.  Now that they are not ignored, this connection string throws an exception at this spot:

```
                if (mechanismPropertyKeyValue.length != 2) {
                    throw new IllegalArgumentException(format("The connection string contains invalid authentication properties. "
                            + "'%s' is not a key value pair", part));
                }
```

The message is`'host2' is not a key value pair`.

Should that be filed as a bug?

driver-core/src/test/unit/com/mongodb/ConnectionStringSpecification.groovy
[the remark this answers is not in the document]
katcharov: This seems fine for this set.

driver-core/src/test/resources/connection-string/valid-warnings.json
katcharov: I think so.

It seems we should notify the user via an error if there is a possible user error, such as forgetting a value. However, the spec says that we ignore at least missing numbers (I don't know the rationale for this - perhaps forward-compatibility?). The spec (according to this test) now also says that it is ok to have a stray mechanism property (called `host2`). I'm not sure this should be allowed, at the spec level - what if the user is trying to enable a security feature that we do not support in an earlier version? It seems that we should fail. In any case, there's either a bug in the spec, or a (minor) bug in the driver.

katcharov (approved):
LGTM!

driver-core/src/test/resources/connection-string/valid-warnings.json
jyemin: Will file

jyemin: Filed JAVA-5504

PR #1422, opened 2024-06-14
Fix release workflow

alcaeus:
This fixes a couple of minor issues that snuck in when I updated the release workflow:
- The git author information was inadvertently removed from the environment
- The inputs for the secure-checkout actions were wrong


jyemin (approved):
LGTM

PR #1423, opened 2024-06-18
Allow generic base classes for POJOs

jyemin:
This change fixes a regression which prevents the driver from encoding and decoding concrete classes which extend generic base classes, when the base class is specified as the generic type of the MongoCollection.

JAVA-5173

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
+++ b/bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
@@ -173,6 +173,9 @@ class LazyPropertyModelCodec<T> implements Codec<T> {
 
         @Override
         public void encode(final BsonWriter writer, final T value, final EncoderContext encoderContext) {
+            if(value.getClass().equals(classModel.getType())) {
+                throw exception();
+            }
             tryEncode(codecRegistry.get(value.getClass()), writer, value, encoderContext);
         }
 
--- a/bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
+++ b/bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
@@ -163,19 +163,41 @@ class LazyPropertyModelCodec<T> implements Codec<T> {
     static final class NeedSpecializationCodec<T> extends PojoCodec<T> {
         private final ClassModel<T> classModel;
         private final DiscriminatorLookup discriminatorLookup;
+        private final CodecRegistry codecRegistry;
 
-        NeedSpecializationCodec(final ClassModel<T> classModel, final DiscriminatorLookup discriminatorLookup) {
+        NeedSpecializationCodec(final ClassModel<T> classModel, final DiscriminatorLookup discriminatorLookup, final CodecRegistry codecRegistry) {
             this.classModel = classModel;
             this.discriminatorLookup = discriminatorLookup;
-        }
-
-        @Override
-        public T decode(final BsonReader reader, final DecoderContext decoderContext) {
-            throw exception();
+            this.codecRegistry = codecRegistry;
         }
 
         @Override
         public void encode(final BsonWriter writer, final T value, final EncoderContext encoderContext) {
+            tryEncode(codecRegistry.get(value.getClass()), writer, value, encoderContext);
+        }
+
+        @Override
+        public T decode(final BsonReader reader, final DecoderContext decoderContext) {
+            return tryDecode(reader, decoderContext);
+        }
+
+        @SuppressWarnings("unchecked")
+        private <A> void tryEncode(final Codec<A> codec,  final BsonWriter writer, final T value, final EncoderContext encoderContext) {
+            try {
+                codec.encode(writer, (A) value, encoderContext);
+            } catch (Exception e) {
+                throw exception();
+            }
+        }
+
+        @SuppressWarnings("unchecked")
+        public T tryDecode(final BsonReader reader, final DecoderContext decoderContext) {
+            Codec<T> codec = PojoCodecImpl.<T>getCodecFromDocument(reader, classModel.useDiscriminator(), classModel.getDiscriminatorKey(),
+                    codecRegistry, discriminatorLookup, null, classModel.getName());
+            if (codec != null) {
+                return codec.decode(reader, decoderContext);
+            }
+
             throw exception();
         }
 
--- a/bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
+++ b/bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
@@ -173,7 +173,7 @@ class LazyPropertyModelCodec<T> implements Codec<T> {
 
         @Override
         public void encode(final BsonWriter writer, final T value, final EncoderContext encoderContext) {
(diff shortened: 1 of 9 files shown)


bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
```diff
@@ -163,19 +163,41 @@ private <V> PropertyModel<V> getSpecializedPropertyModel(final PropertyModel<V>
...
         @Override
-        public T decode(final BsonReader reader, final DecoderContext decoderContext) {
-            throw exception();
+        public void encode(final BsonWriter writer, final T value, final EncoderContext encoderContext) {
+            tryEncode(codecRegistry.get(value.getClass()), writer, value, encoderContext);
         }
 
         @Override
-        public void encode(final BsonWriter writer, final T value, final EncoderContext encoderContext) {
+        public T decode(final BsonReader reader, final DecoderContext decoderContext) {
```
jyemin, line 180: I took this straight from your suggestion, just DRY'd it up.

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
```diff
@@ -545,15 +548,27 @@ public void testInvalidDiscriminatorInNestedModel() {
                         + "'simple': {'_t': 'FakeModel', 'integerField': 42, 'stringField': 'myString'}}"));
     }
 
+    @Test
+    public void testGenericBaseClass() {
```
jyemin, line 552: Not sure I needed a new set of classes here, but couldn't find any others that did quite what I wanted to test.

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
```diff
@@ -545,15 +548,27 @@ public void testInvalidDiscriminatorInNestedModel() {
...
+        CodecRegistry registry = fromProviders(new ValueCodecProvider(), PojoCodecProvider.builder().automatic(true).build());
+
+        ConcreteModel model = new ConcreteModel(new ConcreteField("name1"));
+
+        String json = "{\"_t\": \"org.bson.codecs.pojo.entities.ConcreteModel\", \"field\": {\"name\": \"name1\"}}";
+        roundTrip(PojoCodecProvider.builder().automatic(true), GenericBaseModel.class, model, json);
+    }
+
+
+    // TODO: this now fails with a StackOverflowError, which is not desirable
```
jyemin, line 562: This is undesirable, but not sure how to address it.  Perhaps another check to make sure that the actual class being encoding/decoded has no generic parameters?  Where to put that check though...

bson/src/test/unit/org/bson/codecs/pojo/PojoTestCase.java
```diff
@@ -90,8 +90,12 @@ <T> void roundTrip(final T value, final String json) {
     }
 
     <T> void roundTrip(final PojoCodecProvider.Builder builder, final T value, final String json) {
-        encodesTo(getCodecRegistry(builder), value, json);
-        decodesTo(getCodecRegistry(builder), json, value);
+        roundTrip(builder, value.getClass(), value, json);
+    }
+
+    <T> void roundTrip(final PojoCodecProvider.Builder builder, final Class<?> clazz, final T value, final String json) {
```
jyemin, line 96: This is needed so the class of the Codec can be different from the class of the entity, as it is in the regression described on the JAVA ticket.

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
NathanQingyangXu: I took some look. The issue is a cyclic dependency was found and ended up with LazyCodec as the codec which will lead to infinitive codec lookup loop. Seems the LazyCodec is mainly for breaking cyclic dependency without full-fledged recursive encoding feature. Given its current limit, seems we could throw another exception once LazyCodec was returned from child reigstry.

jyemin: Right, it's an interaction between LazyPropertyModelCodec and the recursive use of types in GenericTreeModel

rozza: I've commented in `LazyPropertyModelCodec`

bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
```diff
@@ -163,19 +163,41 @@ private <V> PropertyModel<V> getSpecializedPropertyModel(final PropertyModel<V>
...
+            return tryDecode(reader, decoderContext);
+        }
+
+        @SuppressWarnings("unchecked")
+        private <A> void tryEncode(final Codec<A> codec,  final BsonWriter writer, final T value, final EncoderContext encoderContext) {
+            try {
```
rozza, line 186: Looks like `codec.encode` can be cyclical if `codecRegistry.get(value.getClass())` returns a `LazyCodec` - eg there is no codec available for that class at the time.

This leads to a stackoverflow of:

```
(LazyCodec) codecRegistry.get(value.getClass())
LazyCodec.encode(...) // Creates a new LazyPropertyModelCodec and round we go again.
```

This can be fixed by checking here if the class of the value and the type of the classModel is the same - if it is it needs specialization, so throw.

bson/src/main/org/bson/codecs/pojo/LazyPropertyModelCodec.java
```diff
@@ -163,19 +163,41 @@ private <V> PropertyModel<V> getSpecializedPropertyModel(final PropertyModel<V>
...
             this.classModel = classModel;
             this.discriminatorLookup = discriminatorLookup;
+            this.codecRegistry = codecRegistry;
         }
 
         @Override
-        public T decode(final BsonReader reader, final DecoderContext decoderContext) {
-            throw exception();
+        public void encode(final BsonWriter writer, final T value, final EncoderContext encoderContext) {
+            tryEncode(codecRegistry.get(value.getClass()), writer, value, encoderContext);
```
rozza, line 176:
```suggestion
            if(value.getClass().equals(classModel.getType())) {
                throw exception();
            }
            tryEncode(codecRegistry.get(value.getClass()), writer, value, encoderContext);
```

bson/src/test/unit/org/bson/codecs/pojo/PojoCustomTest.java
```diff
@@ -545,15 +548,27 @@ public void testInvalidDiscriminatorInNestedModel() {
...
     @Test
     public void testCannotEncodeUnspecializedClasses() {
         CodecRegistry registry = fromProviders(getPojoCodecProviderBuilder(GenericTreeModel.class).build());
-        assertThrows(CodecConfigurationException.class, () ->
+        assertThrows(StackOverflowError.class, () ->
```
rozza, line 566:
```suggestion
        assertThrows(CodecConfigurationException.class, () ->
```

rozza (changes requested):
Looks good - suggested a simple check to prevent the recursion.

rozza (approved):
LGTM

PR #1406, opened 2024-06-03
Add kmip tests, use mongoCrypt snapshot

katcharov:
JAVA-5300

Copies tests, updates docs, and uses the (erroneously named) 1.8.0-SNAPSHOT which has the `delegated` feature available.

See also, update/fix in: https://github.com/mongodb/specifications/pull/1591

PR #820, opened 2021-11-09
Add SSLContext configuration per KMS provider

jyemin:
JAVA-4374

PR #1424, opened 2024-06-20
Add unified, legacy, and prose tests for CSFLE multi-KMS support.

vbabanin:
### Summary
This PR implements the changes described in the "Downstream Changes Summary" of [DRIVERS-2731](https://jira.mongodb.org/browse/DRIVERS-2731).

- Updates JavaDoc.
- Implements required spec tests.
- Passes new AWS credentials to test a rewrapManyDataKey with two different AWS accounts using Evergreen expansions: AWS_ACCESS_KEY_ID_2 and AWS_SECRET_ACCESS_KEY_2, configured in the Evergreen [Variables page](https://spruce.mongodb.com/project/mongo-java-driver/settings/variables).

**NOTE:** No changes to the CSFLE API are necessary, as it is forward-compatible because of using Map [AutoEncryptionSettings.java#L126)](https://github.com/mongodb/mongo-java-driver/blob/50a01d4539c42e81193525d08b85016a59fcf2d4/driver-core/src/main/com/mongodb/AutoEncryptionSettings.java#L126). All relevant logic is encapsulated within `libmongocrypt`. This PR builds on the updated dependency of `mongocrypt version 1.10.0-SNAPSHOT `which was updated in [PR 1406](https://github.com/mongodb/mongo-java-driver/pull/1406/files).

### Background
Named KMS providers is further described in [DBX Scope: Support Named KMS Providers](https://docs.google.com/document/d/1IUxkgGFx5VLjMr7HWxBOXerN0pUZwcqWb11gLoRtcJs/edit?usp=sharing).

The KMS provider identified by a string. Previously supported KMS providers were only: "aws", "azure", "gcp", "kmip", and "local". The KMS provider is now expanded to support name suffixes. (e.g. "local:myname").

Named KMS providers enables more than one of each KMS provider type to be configured.

```
kms_providers = {
    "local": {"key": local_kek1},        # Unnamed KMS provider.
    "local:myname": {"key": local_kek2 } # Named KMS provider with name "myname".
}
```


driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsTlsTest.java
```diff
@@ -120,6 +123,10 @@ public void testInvalidKmsCertificate() {
         }
     }
 
+    /**
+     * See <a href="https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#11-kms-tls-options-tests">
+     * 11. KMS TLS Options Tests</a>.
+     */
     @Test()
     public void testThatCustomSslContextIsUsed() {
```
vbabanin, line 131: A new use-case related to named providers was recently added to the KMS TLS Options specification prose tests. While the existing tests for KMS TLS Options do not precisely align with the spec tests, they have been purposefully designed for simplicity and effectiveness. In line with the approach discussed in [this original comment](https://github.com/mongodb/mongo-java-driver/pull/820/files#r745607434), this test has been extended to include named providers.

vbabanin, line 131: A new use-case related to named providers was recently added to the KMS TLS Options specification prose tests. While the existing tests for KMS TLS Options do not precisely align with the spec tests, they have been purposefully designed for simplicity and effectiveness. In line with the approach discussed in [this original comment](https://github.com/mongodb/mongo-java-driver/pull/820/files#r745607434), this test has been extended to include named providers.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionAwsCredentialFromEnvironmentTest.java
```diff
@@ -192,6 +196,60 @@ public void shouldThrowMongoConfigurationIfSupplierReturnsDoesSomethingUnexpecte
         }
     }
 
+    @Test
+    public void shouldThrowMongoCryptIfNamedKmsProviderHasEmptyParametersInEncryptionSettings() {
+        assumeTrue(serverVersionAtLeast(4, 2));
+        assumeFalse(System.getenv().containsKey("AWS_ACCESS_KEY_ID"));
+        assumeTrue(isClientSideEncryptionTest());
```
stIncMale, line 203: Unless I am crazy, these two assumptions contradict each other, and can never be both satisfied: the first one requires that the environment does not have the `AWS_ACCESS_KEY_ID` variable, while the second one requires the value of the `AWS_ACCESS_KEY_ID` variable to be non-empty.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionAwsCredentialFromEnvironmentTest.java
```diff
@@ -192,6 +196,60 @@ public void shouldThrowMongoConfigurationIfSupplierReturnsDoesSomethingUnexpecte
...
+
+        Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>() {{
+            put("aws:name", new HashMap<>());
+        }};
+
+        Map<String, Supplier<Map<String, Object>>> kmsProviderPropertySuppliers = Mockito.mock(Map.class);
```
stIncMale, line 209: [optional]

If we use `MongoMockito.mock` instead, the mock itself will throw as soon as it is being interacted with. This way we avoid ever having silent incorrect behavior with no clear indication of what caused it (the default behavior of mocks may not cause an immediate exception, but instead cause the code to behave weirdly, or throw from a place that is far away from interacting with a mock).

The same applies to `shouldThrowMongoCryptIfNamedKmsProviderHasEmptyParametersInAutoSettings`.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionAwsCredentialFromEnvironmentTest.java
```diff
@@ -192,6 +196,60 @@ public void shouldThrowMongoConfigurationIfSupplierReturnsDoesSomethingUnexpecte
...
+                .kmsProviderPropertySuppliers(kmsProviderPropertySuppliers)
+                .keyVaultMongoClientSettings(Fixture.getMongoClientSettings())
+                .build();
+
+        MongoCryptException e = assertThrows(MongoCryptException.class, () -> {
+            try (ClientEncryption ignore = createClientEncryption(settings)) {//NOP
+            }
+        });
+        assertTrue(e.getMessage().contains("On-demand credentials are not supported for named KMS providers."));
+        verifyNoInteractions(kmsProviderPropertySuppliers);
```
stIncMale, line 222: [optional]

It seems that there should be no interactions with the suppliers, while the interactions with `kmsProviderPropertySuppliers` itself are not conceptually prohibited. I understand, that the current implementation happens to avoid any interactions with `kmsProviderPropertySuppliers`, but it would have been better to make the assertions less depending on the implementation detail.

The same applies to `shouldThrowMongoCryptIfNamedKmsProviderHasEmptyParametersInAutoSettings`.

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionAwsCredentialFromEnvironmentTest.java
vbabanin: I have adjusted the test behavior to verify that only `Supplier.get` has not been invoked. This is done using the following setup:

```
 Map<String, Supplier<Map<String, Object>>> kmsProviderPropertySuppliers = new HashMap<>();
        kmsProviderPropertySuppliers.put("aws:name", () -> Assertions.fail("Supplier should not be called"));
```

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionAwsCredentialFromEnvironmentTest.java
vbabanin: Good catch! Removed `assumeFalse(System.getenv().containsKey("AWS_ACCESS_KEY_ID"));`, thanks!

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedClientEncryptionHelper.java
```diff
@@ -207,6 +240,38 @@ OperationResult executeRewrapManyDataKey(final BsonDocument operation) {
         return resultOf(() ->  toExpected(clientEncryption.rewrapManyDataKey(filter, rewrapManyDataKeyOptions)));
     }
 
+    OperationResult executeEncrypt(final BsonDocument operation) {
+        ClientEncryption clientEncryption = entities.getClientEncryption(operation.getString("object").getValue());
+        BsonDocument arguments = operation.getDocument("arguments", new BsonDocument());
+        BsonDocument options = arguments.getDocument("opts", new BsonDocument());
```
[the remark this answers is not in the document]
vbabanin: I've updated the implementation to fail when there are no arguments or options.

stIncMale (changes requested):

stIncMale (approved):

PR #1427, opened 2024-06-27
Extend Socks5 TLD regex validation to support up to 63 characters

vbabanin:
### Changes:

Modified the regex pattern to support top-level domain (TLD) lengths of up to 63 characters, in accordance with RFC 1034 standards.

### Purpose:
This update ensures our regex can handle the 'internal' TLD used in EC2 guest OS hostnames, thus increasing the flexibility of TLD validation.

JAVA-5490

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
@@ -22,7 +22,7 @@ import java.util.regex.Pattern;
  */
 public class DomainNameUtils {
     private static final Pattern DOMAIN_PATTERN =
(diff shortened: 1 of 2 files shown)


rozza (approved):
LGTM

rozza:
Only question: Should there be a drivers ticket for this change (for those drivers that regex validate TLDs)? Maybe a prose test?

vbabanin:
> Only question: Should there be a drivers ticket for this change (for those drivers that regex validate TLDs)? Maybe a prose test?

Thanks for your question. There’s no need for a driver's ticket as the regex validation of domain names is only implemented in Java driver.

PR #1437, opened 2024-07-04
Attribute release commit and tag to user triggering the workflow

alcaeus:
With this change, commits made by the release automation and the git tag itself will be attributed to the user triggering the workflow. We can do so without knowing the email address of a user by using GitHub's anonymous email address format `<id>+<name>@users.noreply.github.com`.

This fixes the issue where the tag run in Evergreen does not know which employee created the release, which is relevant for the internal Papertrail service.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,8 +71,9 @@ jobs:
       # Set commit author information to the user that triggered the release workflow
       - name: "Set git author information"
         run: |
+          GITHUB_USER_NAME=$(gh api users/${{ github.actor }} --jq '.name')
           GITHUB_USER_ID=$(gh api users/${{ github.actor }} --jq '.id')
-          git config user.name "${{ github.actor }}"
+          git config user.name "${GITHUB_USER_NAME}}"
           git config user.email "${GITHUB_USER_ID}+${{ github.actor }}@users.noreply.github.com"
 
       # This step bumps version numbers in build.gradle and creates git artifacts for the release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,10 +9,6 @@ on:
         required: true
         type: "string"
 
-env:
-  GIT_AUTHOR_EMAIL: "167856002+mongodb-dbx-release-bot[bot]@users.noreply.github.com"
-  GIT_AUTHOR_NAME: "mongodb-dbx-release-bot[bot]"
-
 jobs:
   prepare-release:
     environment: release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -72,10 +68,12 @@ jobs:
           echo '🆕 Creating new release branch ${RELEASE_BRANCH} from ${{ github.ref_name }}' >> $GITHUB_STEP_SUMMARY
           git checkout -b ${RELEASE_BRANCH}
 
+      # Set commit author information to the user that triggered the release workflow
       - name: "Set git author information"
         run: |
-          git config user.name "${GIT_AUTHOR_NAME}"
-          git config user.email "${GIT_AUTHOR_EMAIL}"
+          GITHUB_USER_ID=$(gh api users/${{ github.actor }} --jq '.id')
+          git config user.name "${{ github.actor }}"
```
jyemin, line 75: I might be misunderstanding the intent here, but I thought something like this would be needed here: 

```
git config user.name  $(gh api users/${{ github.actor }} --jq '.name')`
git config user.email  $(gh api users/${{ github.actor }} --jq '.email')`
```

stIncMale: Based on the Slack discussions, @alcaeus was concerned about users not necessarily having email exposed publicly. THe approach in the PR works around that issue.

jyemin: I don't understand the concern.  Can you explain here?

alcaeus: > I don't think we need user email at all, as Evergreen does not send to Papertrail. So I would have just left `git config user.email untouched`.
> On top of #1, I don't know what GitHub will use as `user.email` if we don't set it explicitly.

`git` will refuse to create a commit object if `user.email` is not configured properly, so this would fail long before GitHub sees the commit. Once the commit is pushed, GitHub attributes it to a user based on the email address. So if we were to only change the name of the commit author but use the app's email address, the commits would still be attributed to the app, which is why I need to change both name and email.

> @alcaeus wants to set user.email explicitly, and if a user does not have a publicly available email, then user.email will likely be set to an empty string (I don't know how $(gh api users/${{ github.actor }} --jq '.email') works in that case).
> Given that GitHub creates a public email for every user, and it is known how that email is constructed (this is all new to me), we can compute it on our own for any user, and know that we will always have a correct value in user.email.

That's correct. For example, the [API response](https://api.github.com/users/jyemin) for @jyemin lists `null` for the `email` field. On top of that, users can verify multiple email addresses, so using the `email` field doesn't necessarily give us the correct email address, it just gives us the primary one. That's why I decided to use GitHub's noreply address: it's predictable, always exists, and will forever be associated with that user.

jyemin (approved):
LGTM

PR #1440, opened 2024-07-09
Fix `CursorResourceManager.close`

stIncMale:
This PR is an alternative to https://github.com/mongodb/mongo-java-driver/pull/1439.

The commit https://github.com/mongodb/mongo-java-driver/pull/1440/commits/b7be8c856f8e500dcc10ca39757c8dccf33769c3 introduces a draft reproducer of the bug reported in JAVA-5516. Then the following commit fixes it. The reproducer is not guaranteed to reproduce the error, but usually does so in just a few attempts on my machine. With the fix I was unable to reproduce the error.

The commit https://github.com/mongodb/mongo-java-driver/pull/1440/commits/510548421e663d391a89557a9940afeacba2dc9c introduces a reproducer that we can actually check into the codebase.

JAVA-5516

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/CursorResourceManager.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/CursorResourceManager.java
@@ -182,7 +182,7 @@ abstract class CursorResourceManager<CS extends ReferenceCounted, C extends Refe
     void close() {
         boolean doClose = withLock(lock, () -> {
             State localState = state;
-            if (localState == State.OPERATION_IN_PROGRESS) {
+            if (localState.inProgress()) {
                 state = State.CLOSE_PENDING;
             } else if (localState != State.CLOSED) {
                 state = State.CLOSED;
--- a/driver-core/src/test/functional/com/mongodb/internal/operation/AsyncCommandBatchCursorFunctionalTest.java
+++ b/driver-core/src/test/functional/com/mongodb/internal/operation/AsyncCommandBatchCursorFunctionalTest.java
@@ -45,7 +45,6 @@ import org.junit.jupiter.params.provider.MethodSource;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
@@ -298,24 +297,6 @@ public class AsyncCommandBatchCursorFunctionalTest extends OperationTest {
 
         assertDoesNotThrow(() -> checkReferenceCountReachesTarget(connectionSource, 1));
         assertTrue(cursor.isClosed());
-    }
-
-    @Test
-    void attemptJava5516() {
-        BsonDocument commandResult = executeFindCommand(5, 2);
-        cursor = new AsyncCommandBatchCursor<>(commandResult, 2, 0, DOCUMENT_DECODER,
-                                               null, connectionSource, connection);
-        assertNotNull(cursorNext());
-        // Calling `close` twice is the key to reproducing.
-        // It does not matter whether we call `close` twice from the same thread or not.
-        ForkJoinPool.commonPool().execute(() -> cursor.close());
-        ForkJoinPool.commonPool().execute(() -> cursor.close());
-        try {
-            assertNotNull(cursorNext());
-            assertNotNull(cursorNext());
-        } catch (IllegalStateException e) {
-            // one of the expected outcomes, because we call `cursorNext` concurrently with `close`
-        }
     }
 
     @Test
--- a/driver-core/src/test/functional/com/mongodb/internal/operation/AsyncCommandBatchCursorFunctionalTest.java
+++ b/driver-core/src/test/functional/com/mongodb/internal/operation/AsyncCommandBatchCursorFunctionalTest.java
@@ -45,6 +45,7 @@ import org.junit.jupiter.params.provider.MethodSource;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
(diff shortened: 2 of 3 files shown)


driver-core/src/main/com/mongodb/internal/operation/CursorResourceManager.java
```diff
@@ -182,7 +182,7 @@ void endOperation() {
     void close() {
         boolean doClose = withLock(lock, () -> {
             State localState = state;
-            if (localState == State.OPERATION_IN_PROGRESS) {
+            if (localState.inProgress()) {
```
stIncMale, line 185: The problem with the original code:

If `AsyncCommandBatchCursor.close` is called concurrently more than once while `AsyncCommandBatchCursor.next` is running (not necessarily the `next` method itself, but the asynchronous logic of this method), then the following state transitions were possible:

1. One of the `close` calls observes `State.OPERATION_IN_PROGRESS`, and changes it to `State.CLOSE_PENDING`. So far so good.
2. The other `close` observes `State.CLOSE_PENDING` and changes it to `State.CLOSED`, which breaks one of the invariant that are supposed to be maintained by `CursorResourceManager`: while an operation is in progress, `state.inProgress()` must be `true`, i.e., `state` must be either `OPERATION_IN_PROGRESS` or `CLOSE_PENDING` (implies `OPERATION_IN_PROGRESS`).

driver-core/src/test/functional/com/mongodb/internal/operation/AsyncCommandBatchCursorFunctionalTest.java
```diff
@@ -299,6 +300,24 @@ void testLimitWithGetMore() {
...
+    @Test
+    void attemptJava5516() {
+        BsonDocument commandResult = executeFindCommand(5, 2);
+        cursor = new AsyncCommandBatchCursor<>(commandResult, 2, 0, DOCUMENT_DECODER,
+                                               null, connectionSource, connection);
+        assertNotNull(cursorNext());
+        // Calling `close` twice is the key to reproducing.
+        // It does not matter whether we call `close` twice from the same thread or not.
+        ForkJoinPool.commonPool().execute(() -> cursor.close());
+        ForkJoinPool.commonPool().execute(() -> cursor.close());
```
stIncMale, line 312: The issue this PR fixes is 100% a bug. And it manifests itself with an error that is very similar to the user-reported error (following is the error I reproduced locally):

```
java.lang.AssertionError: null
	at com.mongodb.assertions.Assertions.assertTrue(Assertions.java:158)
	at com.mongodb.internal.operation.CursorResourceManager.setServerCursor(CursorResourceManager.java:225)
	at com.mongodb.internal.operation.AsyncCommandBatchCursor.lambda$getMoreLoop$2(AsyncCommandBatchCursor.java:186)
	at com.mongodb.internal.async.ErrorHandlingResultCallback.onResult(ErrorHandlingResultCallback.java:47)
	at com.mongodb.internal.connection.DefaultServer$DefaultServerProtocolExecutor.lambda$executeAsync$0(DefaultServer.java:249)
	at com.mongodb.internal.async.ErrorHandlingResultCallback.onResult(ErrorHandlingResultCallback.java:47)
	at com.mongodb.internal.connection.CommandProtocolImpl.lambda$executeAsync$0(CommandProtocolImpl.java:88)
	at com.mongodb.internal.connection.DefaultConnectionPool$PooledConnection.lambda$sendAndReceiveAsync$1(DefaultConnectionPool.java:774)
	at com.mongodb.internal.connection.UsageTrackingInternalConnection.lambda$sendAndReceiveAsync$1(UsageTrackingInternalConnection.java:150)
	at com.mongodb.internal.async.ErrorHandlingResultCallback.onResult(ErrorHandlingResultCallback.java:47)
	at com.mongodb.internal.async.AsyncSupplier.lambda$finish$0(AsyncSupplier.java:65)
	at com.mongodb.internal.async.SingleResultCallback.complete(SingleResultCallback.java:67)
	at com.mongodb.internal.async.AsyncSupplier.lambda$onErrorIf$5(AsyncSupplier.java:122)
	at com.mongodb.internal.async.AsyncSupplier.lambda$finish$0(AsyncSupplier.java:65)
	at com.mongodb.internal.async.AsyncSupplier.lambda$finish$0(AsyncSupplier.java:65)
	at com.mongodb.internal.connection.InternalStreamConnection.lambda$sendCommandMessageAsync$7(InternalStreamConnection.java:613)
	at com.mongodb.internal.connection.InternalStreamConnection$MessageHeaderCallback$MessageCallback.onResult(InternalStreamConnection.java:894)
```

However, none of this guarantees that the bug the PR fixes is the bug the user is facing, because theoretically, I can imagine other bugs leading to the same error. However, I was unable to find them in our code, so I am saying that the fix proposed in this PR is my current best solution to JAVA-5516.

rozza (approved):
LGTM

driver-core/src/test/unit/com/mongodb/internal/operation/CursorResourceManagerTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+            @Override
+            void doClose() {
+            }
+        };
+        cursorResourceManager.tryStartOperation();
+        try {
+            assertDoesNotThrow(() -> {
+                cursorResourceManager.close();
+                cursorResourceManager.close();
```
stIncMale, line 52: A good example of `close` being executed concurrently with an operation being in progress without any parallelism involved. Examples like this emphasize how [concurrency and parallelism are different](https://www.kovalenko.link/blog/parallelism-vs-concurrency).

driver-core/src/test/unit/com/mongodb/internal/operation/CursorResourceManagerTest.java
```diff
@@ -0,0 +1,59 @@
...
+            void markAsPinned(final ReferenceCounted connectionToPin, final Connection.PinningMode pinningMode) {
+            }
+
+            @Override
+            void doClose() {
+            }
+        };
+        cursorResourceManager.tryStartOperation();
+        try {
+            assertDoesNotThrow(() -> {
```
stIncMale, line 50: Without the fix in `CursorResourceManager`, this assertion fails.

stIncMale:
@rozza, I replaced the reproducer with a simpler one that is also deterministic. Please take a look.

thjaeckle:
Thanks for the fix.
Looking forward to get rid of the assertion errors 👍

Guillaume-Mayer:
If it can help anyone, I think this bug would cause conection not to be freed, which could lead to the saturation of the pool. We had this problem using Spring 3.3.4 (WebFlux) fixed in Spring Boot 3.4.0 when limiting find result with Flux.take(n).

PR #1448, opened 2024-07-15
Use JCA provider for implementation of PBKDF2

jyemin:
Previously the driver implemented the PBKDF2 (Password-based-Key-Derivative-Function) function itself. With this change it instead looks up an implementation of either PBKDF2WithHmacSHA1 or PBKDF2WithHmacSHA256 from the JCA provider via SecretKeyFactory#getInstance.

JAVA-5507

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ScramShaAuthenticator.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ScramShaAuthenticator.java
@@ -28,6 +28,8 @@ import org.bson.BsonDocument;
 import org.bson.BsonString;
 
 import javax.crypto.Mac;
+import javax.crypto.SecretKeyFactory;
+import javax.crypto.spec.PBEKeySpec;
 import javax.crypto.spec.SecretKeySpec;
 import javax.security.sasl.SaslClient;
 import javax.security.sasl.SaslException;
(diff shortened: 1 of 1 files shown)


rozza (approved):
LGTM - iirc we had to manually implement the algorithm so to support Java 6 & 7

driver-core/src/main/com/mongodb/internal/connection/ScramShaAuthenticator.java
```diff
@@ -246,25 +254,15 @@ private byte[] h(final byte[] data) throws SaslException {
...
-                }
-                return result;
+                SecretKeyFactory secretKeyFactory = SecretKeyFactory.getInstance(pbeAlgorithm);
+                PBEKeySpec spec = new PBEKeySpec(password.toCharArray(), salt, iterations, keyLength);
+                return secretKeyFactory.generateSecret(spec).getEncoded();
             } catch (NoSuchAlgorithmException e) {
-                throw new SaslException(format("Algorithm for '%s' could not be found.", hmacAlgorithm), e);
-            } catch (InvalidKeyException e) {
-                throw new SaslException(format("Invalid key for %s", hmacAlgorithm), e);
+                throw new SaslException(format("Algorithm for '%s' could not be found.", pbeAlgorithm), e);
```
vbabanin, line 263: [nit] This String is repeated in the code. It could be a candidate for a constant.

vbabanin (approved):
LGTM!

jyemin:
> iirc we had to manually implement the algorithm so to support Java 6 & 7

Oh, good memory!  I checked and found that PBKDF2WithHmacSHA1 is supported back to Java 6 but PBKDF2WithHmacSHA256 came in Java 8.

PR #1449, opened 2024-07-16
Unify timeout checks in exception handling methods.

vbabanin:
JAVA-5526

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/TimeoutContext.java
+++ b/driver-core/src/main/com/mongodb/internal/TimeoutContext.java
@@ -31,6 +31,7 @@ import static com.mongodb.assertions.Assertions.isTrue;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
 import static com.mongodb.internal.time.Timeout.ZeroSemantics.ZERO_DURATION_MEANS_INFINITE;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 /**
  * Timeout Context.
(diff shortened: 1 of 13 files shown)


driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -139,7 +138,7 @@ public boolean hasTimeoutMS() {
      * @param onExpired the runnable to run
      */
     public void onExpired(final Runnable onExpired) {
-        Timeout.nullAsInfinite(timeout).onExpired(onExpired);
+        Timeout.nullAsInfinite(timeout).onExpired(onExpired, MILLISECONDS);
```
vbabanin, line 141: The unit is changed to MILLISECONDS to align with other logic in the `TimeoutContext`, which also calculates in MILLISECONDS.

driver-core/src/main/com/mongodb/internal/TimeoutContext.java
```diff
@@ -253,9 +253,11 @@ public int getConnectTimeoutMs() {
                 () -> throwMongoTimeoutException("The operation exceeded the timeout limit.")));
     }
 
-    public void resetTimeout() {
-        assertNotNull(timeout);
-        timeout = startTimeout(timeoutSettings.getTimeoutMS());
+    public void resetTimeoutIfPresent() {
+        if (hasTimeoutMS()) {
```
rozza, line 257: Comment: The method naming makes the `assertNotNull` seem restrictive. Should it be part of the if statement?

I guess its a bug if this is called if there isnt a timeout set - its just slightly unclear thats the case.

rozza (approved):
LGTM - had one optional comment / observation.

PR #1450, opened 2024-07-16
Ensure Sink.contextView is propagated

rozza:
When creating a Mono via Mono.create any nested
Monos should have the higher level sinks context view propogated down.

JAVA-5345


rozza:
I was unable to create a test to replicate the tracing issue reported in: spring-projects/spring-data-mongodb#4650. However, I manually tested against the repo in ticket and confirmed it works.

stIncMale:
Shouldn't we also do a similar change [here](https://github.com/mongodb/mongo-java-driver/blob/96540ca32d7837c9ed9d36da433d6c480ed319ef/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/Crypt.java#L307-L308)? We have a `sink` there that is given to us, and `collectionInfoRetriever.filter` creates a new `Mono`. I.e., shouldn't we do

```java
collectionInfoRetriever.filter(databaseName, cryptContext.getMongoOperation())
        .contextWrite(sink.contextView())
        .doOnSuccess...
        ...
        .subscribe();
```

rozza:
@stIncMale I agree its not helped by not being able to replicate the original test without the full stack.  Let me investigate so more.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
```diff
@@ -103,21 +103,25 @@ public void subscribe(final Subscriber<? super Void> s) {
             sink.onCancel(() -> createCancellationMono(terminated).subscribe());
```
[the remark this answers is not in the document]
rozza: Done

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorPublisher.java
```diff
@@ -106,6 +106,7 @@ public Publisher<T> first() {
                 .flatMap(batchCursor -> Mono.create(sink -> {
                     batchCursor.setBatchSize(1);
                     Mono.from(batchCursor.next())
+                            .contextWrite(sink.contextView())
```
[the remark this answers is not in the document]
rozza: Done

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
```diff
@@ -227,6 +233,7 @@ private Mono<Long> createSaveChunksMono(final AtomicBoolean terminated) {
             AtomicLong lengthInBytes = new AtomicLong(0);
             AtomicInteger chunkIndex = new AtomicInteger(0);
             new ResizingByteBufferFlux(source, chunkSizeBytes)
+                    .contextWrite(sink.contextView())
                     .flatMap((Function<ByteBuffer, Publisher<InsertOneResult>>) byteBuffer -> {
                         if (terminated.get()) {
                             return Mono.empty();
```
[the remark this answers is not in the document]
rozza: I don't know why the tracing in the repo example was happy here. I also don't fully understand how reactor propagates it 🤷 

So it appears to automatically be propogated within scope - but I have added your suggestion to make it explicit.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorPublisher.java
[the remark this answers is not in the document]
stIncMale: > Should the existing contextWrite call on line 119 be removed?

I thought it's better to remove the existing call, and leave the new call that comes right after creating the nested `Mono`, because in all other places `contextWrite` is called right after creating a nested `Mono`.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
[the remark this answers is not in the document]
stIncMale: The change seems to having been done only to the `Publisher`s created via `insertOne`, but "this `Mono`", i.e., the `Mono` on line 239 where I originally left the comment (that line is now 238) still does not have `contextWrite` called on it.

> it appears to automatically be propogated within scope

What do you mean by "scope" here?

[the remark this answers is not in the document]
rozza: > a Context is tied to each Subscriber in a chain. It uses the Subscription propagation mechanism to make itself available to each operator, starting with the final subscribe and moving up the chain.

So that makes sense why our code broke that.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorPublisher.java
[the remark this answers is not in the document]
rozza: Removed the use of `subscribe` from within the `BatchCursorPublisher`

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -48,9 +46,9 @@ public void subscribe(final Subscriber<? super T> subscriber) {
                 if (calculateDemand(demand) > 0 && inProgress.compareAndSet(false, true)) {
                     if (batchCursor == null) {
                         int batchSize = calculateBatchSize(sink.requestedFromDownstream());
-                        Context initialContext =  subscriber instanceof CoreSubscriber<?>
-                                ? ((CoreSubscriber<?>) subscriber).currentContext() : null;
-                        batchCursorPublisher.batchCursor(batchSize).subscribe(bc -> {
+                        batchCursorPublisher.batchCursor(batchSize)
+                                .contextWrite(sink.contextView())
```
rozza, line 50: Simplfy getting and setting the context for this Mono.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorFlux.java
```diff
@@ -86,6 +84,7 @@ private void recurseCursor(){
             } else {
                 batchCursor.setBatchSize(calculateBatchSize(sink.requestedFromDownstream()));
                 Mono.from(batchCursor.next(() -> sink.isCancelled()))
+                        .contextWrite(sink.contextView())
```
rozza, line 87: propagate the context to this Mono as we control the subscription of it.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/BatchCursorPublisher.java
```diff
@@ -123,21 +123,17 @@ public TimeoutMode getTimeoutMode() {
 
     public Publisher<T> first() {
         return batchCursor(this::asAsyncFirstReadOperation)
-                .flatMap(batchCursor -> Mono.create(sink -> {
```
rozza, line 126: No need for the sink

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
```diff
@@ -120,7 +120,7 @@ public void subscribe(final Subscriber<? super Void> s) {
                                         return originalError;
                                     })
                                     .then(Mono.error(originalError)))
-                    .doOnCancel(() -> createCancellationMono(terminated, timeout).subscribe())
+                    .doOnCancel(() -> createCancellationMono(terminated, timeout).contextWrite(ctx).subscribe())
```
rozza, line 123: Cancellation has a side effect - so we need to propagate context there.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
```diff
@@ -228,40 +205,37 @@ private <T> Mono<String> createIndexMono(final MongoCollection<T> collection, fi
     }
 
     private Mono<Long> createSaveChunksMono(final AtomicBoolean terminated, @Nullable final Timeout timeout) {
-        return Mono.create(sink -> {
```
rozza, line 231: Remove the use of sink and use the project reactor api to declare how the data is processed.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
```diff
@@ -149,38 +149,15 @@ public void subscribe(final Subscriber<? super ObjectId> subscriber) {
...
-                                .first());
-                    }
-                });
+        return collectionWithTimeoutDeferred(filesCollection.withDocumentClass(Document.class).withReadPreference(primary()), timeout)
+                .map(collection -> clientSession != null ? collection.find(clientSession) : collection.find())
+                .flatMap(findPublisher -> Mono.from(findPublisher.projection(PROJECTION).first()))
+                .switchIfEmpty(Mono.defer(() ->
+                        checkAndCreateIndex(filesCollection.withReadPreference(primary()), FILES_INDEX, timeout)
+                                .then(checkAndCreateIndex(chunksCollection.withReadPreference(primary()), CHUNKS_INDEX, timeout))
+                                .then(Mono.fromCallable(Document::new))
```
stIncMale, line 158: It seems we can easily avoid having `Mono.fromCallable(Document::new)` by flat mapping `Mono<Document>` to `Mono<Void>` via `then` before we call `switchIfEmpty`:

```suggestion
                .flatMap(findPublisher -> Mono.from(findPublisher.projection(PROJECTION).first()))
                .then()
                .switchIfEmpty(Mono.defer(() ->
                        checkAndCreateIndex(filesCollection.withReadPreference(primary()), FILES_INDEX, timeout)
                                .then(checkAndCreateIndex(chunksCollection.withReadPreference(primary()), CHUNKS_INDEX, timeout))
                                .then()
```

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
stIncMale: Isn't us calling `subscribe()` here not with a subscriber that was given to us (the `s` argument), the only thing that requires us to propagate the context from the subscriber `s`? I suspect that the fact that `createCancellationMono` has a side-effect (it mutates `terminated`) is irrelevant. If the side-effect is relevant to context propagation, then could you please explain why?

stIncMale (changes requested):
Just a small code suggestion

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
rozza: As `then()` will never have a result `switchIfEmpty` will always be called.

> Mono<Void> should be used for Publisher that just completes without any value.
https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
rozza: On cancellation of the main `Publisher` we are calling [subscribe()](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#subscribe--) on a async clean up `Publisher` and we ignore any errors, results / completion.

> If the side-effect is relevant to context propagation, then could you please explain why?

The user may have tracing and want to trace from the web app to the driver. Without adding the context here we break that chain and they cannot tie the cleanup operations to a web request.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
rozza: Updated to use `.then(Mono.empty())` instead of `.then(Mono.fromCallable(Document::new))`

rozza:
@stIncMale answered your questions - not sure why the reply to `createCancellationMono` isn't showing inline - but if you click on the code change you see it.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/gridfs/GridFSUploadPublisherImpl.java
stIncMale: > As `then()` will never have a result `switchIfEmpty` will always be called.

This was a stupid suggestion from me. Sorry, Ross.

stIncMale (approved):

PR #1453, opened 2024-07-17
Fix bson-kotlinx `encodeNullableSerializableValue` null handling

rozza:
Ensures that the `deferredElement` name is reset as it does in `encodeSerializableValue`. 
Test case ported to bson-kotlin

JAVA-5524

### DIFF
--- a/bson-kotlin/src/test/kotlin/org/bson/codecs/kotlin/DataClassCodecTest.kt
+++ b/bson-kotlin/src/test/kotlin/org/bson/codecs/kotlin/DataClassCodecTest.kt
@@ -25,6 +25,7 @@ import org.bson.codecs.configuration.CodecConfigurationException
 import org.bson.codecs.configuration.CodecRegistries.fromProviders
 import org.bson.codecs.kotlin.samples.Box
 import org.bson.codecs.kotlin.samples.DataClassEmbedded
+import org.bson.codecs.kotlin.samples.DataClassLastItemDefaultsToNull
 import org.bson.codecs.kotlin.samples.DataClassListOfDataClasses
 import org.bson.codecs.kotlin.samples.DataClassListOfListOfDataClasses
 import org.bson.codecs.kotlin.samples.DataClassListOfSealed
@@ -51,6 +52,7 @@ import org.bson.codecs.kotlin.samples.DataClassWithEnum
 import org.bson.codecs.kotlin.samples.DataClassWithEnumMapKey
 import org.bson.codecs.kotlin.samples.DataClassWithFailingInit
 import org.bson.codecs.kotlin.samples.DataClassWithInvalidBsonRepresentation
+import org.bson.codecs.kotlin.samples.DataClassWithListThatLastItemDefaultsToNull
 import org.bson.codecs.kotlin.samples.DataClassWithMutableList
 import org.bson.codecs.kotlin.samples.DataClassWithMutableMap
 import org.bson.codecs.kotlin.samples.DataClassWithMutableSet
@@ -131,6 +133,20 @@ class DataClassCodecTest {
 
         val withStoredNulls = BsonDocument.parse("""{"boolean": null, "string": null, "listSimple": null}""")
         assertDecodesTo(withStoredNulls, dataClass)
+    }
+
+    @Test
+    fun testDataClassWithListThatLastItemDefaultsToNull() {
+        val expected =
+            """{
+            | "elements": [{"required": "required"}, {"required": "required"}],
+            |}"""
+                .trimMargin()
+
+        val dataClass =
+            DataClassWithListThatLastItemDefaultsToNull(
+                listOf(DataClassLastItemDefaultsToNull("required"), DataClassLastItemDefaultsToNull("required")))
+        assertRoundTrips(expected, dataClass)
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonEncoder.kt
```diff
@@ -156,6 +156,8 @@ internal class DefaultBsonEncoder(
             if (value != null || configuration.explicitNulls) {
                 encodeName(it)
                 super.encodeNullableSerializableValue(serializer, value)
+            } else {
```
jyemin, line 159: It seems worth it to add tests for the corresponding `else` clause in the `encodeSerializableValue` method just above (I removed the clause and no tests fails, so it's not covered).

jyemin (changes requested):

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonEncoder.kt
rozza: For me the new test: `KotlinSerializerCodecTest.testDataClassWithListThatLastItemDefaultsToNull` fails if I remove this else statement.

bson-kotlin/src/test/kotlin/org/bson/codecs/kotlin/DataClassCodecTest.kt
```diff
@@ -133,6 +135,20 @@ class DataClassCodecTest {
         assertDecodesTo(withStoredNulls, dataClass)
     }
 
+    @Test
+    fun testDataClassWithListThatLastItemDefaultsToNull() {
```
rozza, line 139: This is just a copy from `bson-kotlinx` - to ensure the `bson-kotlin` library also works as expected.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonEncoder.kt
rozza: nm - will update the tests.

rozza: I can't concoct a test that triggers this.  That said it doesn't guarantee that there isn't a scenario where it could be triggered. 

As we are skipping this value it makes sense to reset the defferred name if it exists.
However, I think I can clean up the usage.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonEncoder.kt
```diff
@@ -170,14 +171,7 @@ internal class DefaultBsonEncoder(
...
     override fun encodeLong(value: Long) = writer.writeInt64(value)
-    override fun encodeNull() {
-        deferredElementName?.let {
-            if (configuration.explicitNulls) {
-                encodeName(it)
-            }
-        }
-        writer.writeNull()
-    }
+    override fun encodeNull() = writer.writeNull()
```
rozza, line 174: Not sure how this was missed in a PR - but encoding null will never require the deferred element name - as encode(Nullable)SerializableValue will be used

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonEncoder.kt
```diff
@@ -206,7 +200,6 @@ internal class DefaultBsonEncoder(
 
     private fun encodeName(value: Any) {
         writer.writeName(value.toString())
-        deferredElementName = null
```
rozza, line 209: No longer required - as its cleaned up in the `SerializableValue` usages

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonEncoder.kt
rozza: @jyemin I cleaned up the usage and encapsulated the handling of deferred element names.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonEncoder.kt
```diff
@@ -140,25 +140,25 @@ internal class DefaultBsonEncoder(
...
-        deferredElementName?.let {
-            if (value != null || configuration.explicitNulls) {
-                encodeName(it)
-                super.encodeSerializableValue(serializer, value)
-            } else {
-                deferredElementName = null
-            }
-        }
-            ?: super.encodeSerializableValue(serializer, value)
+        deferredElementHandler.with(
```
jyemin, line 143: It's unclear to me why value is

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonEncoder.kt
```diff
@@ -140,25 +140,25 @@ internal class DefaultBsonEncoder(
...
+            {
+                if (value != null || configuration.explicitNulls) {
```
jyemin, line 145: Just a point of clarification:

It's unclear to me why `value` is ever `null` here.  If it is, why isn't `encodeNullableSerializableValue` called instead?

jyemin (approved):
Thanks.  LGTM

PR #1455, opened 2024-07-19
Remove redundant switchIfEmptyStage

rozza:
As ` clientSessionHelper.withClientSession` will always produce a session, there is no need for the `switchIfEmpty` branch.


vbabanin (approved):
LGTM!

PR #1456, opened 2024-07-20
Support any number of Document Sequence Sections in CommandMessage#getCommandDocument

jyemin:
JAVA-5536

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonDocument.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonDocument.java
@@ -53,38 +53,26 @@ final class ByteBufBsonDocument extends BsonDocument {
 
     private final transient ByteBuf byteBuf;
 
-    static List<ByteBufBsonDocument> createList(final ByteBufferBsonOutput bsonOutput, final int startPosition) {
-        List<ByteBuf> duplicateByteBuffers = bsonOutput.getByteBuffers();
-        CompositeByteBuf outputByteBuf = new CompositeByteBuf(duplicateByteBuffers);
-        outputByteBuf.position(startPosition);
+    // Create a list of ByteBufBsonDocument from a ByteBuf positioned at the start of the first document of an OP_MSG Section
+    // of type Kind 1: Document Sequence
+    // The provided ByteBuf will be positioned at the end of the section upon normal completion of the method
+    static List<ByteBufBsonDocument> createList(final ByteBuf outputByteBuf) {
         List<ByteBufBsonDocument> documents = new ArrayList<>();
-        int curDocumentStartPosition = startPosition;
         while (outputByteBuf.hasRemaining()) {
-            int documentSizeInBytes = outputByteBuf.getInt();
-            ByteBuf slice = outputByteBuf.duplicate();
-            slice.position(curDocumentStartPosition);
-            slice.limit(curDocumentStartPosition + documentSizeInBytes);
-            documents.add(new ByteBufBsonDocument(slice));
-            curDocumentStartPosition += documentSizeInBytes;
-            outputByteBuf.position(outputByteBuf.position() + documentSizeInBytes - 4);
-        }
-        for (ByteBuf byteBuffer : duplicateByteBuffers) {
-            byteBuffer.release();
+            ByteBufBsonDocument curDocument = createOne(outputByteBuf);
+            documents.add(curDocument);
         }
         return documents;
     }
 
-    static ByteBufBsonDocument createOne(final ByteBufferBsonOutput bsonOutput, final int startPosition) {
-        List<ByteBuf> duplicateByteBuffers = bsonOutput.getByteBuffers();
-        CompositeByteBuf outputByteBuf = new CompositeByteBuf(duplicateByteBuffers);
-        outputByteBuf.position(startPosition);
+    // Create a ByteBufBsonDocument from a ByteBuf positioned at the start of a BSON document.
+    // The provided ByteBuf will be positioned at the end of the section upon normal completion of the method
+    static ByteBufBsonDocument createOne(final ByteBuf outputByteBuf) {
         int documentSizeInBytes = outputByteBuf.getInt();
         ByteBuf slice = outputByteBuf.duplicate();
-        slice.position(startPosition);
-        slice.limit(startPosition + documentSizeInBytes);
-        for (ByteBuf byteBuffer : duplicateByteBuffers) {
-            byteBuffer.release();
-        }
+        slice.position(slice.position() - 4);
+        slice.limit(slice.position() + documentSizeInBytes);
+        outputByteBuf.position(outputByteBuf.position() + documentSizeInBytes - 4);
         return new ByteBufBsonDocument(slice);
     }
 
@@ -136,10 +124,6 @@ final class ByteBufBsonDocument extends BsonDocument {
         }
 
         return finder.notFound();
-    }
-
-    int getSizeInBytes() {
-        return byteBuf.getInt(byteBuf.position());
     }
 
     BsonDocument toBaseBsonDocument() {
--- a/driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonDocument.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonDocument.java
@@ -73,11 +73,11 @@ final class ByteBufBsonDocument extends BsonDocument {
      * The provided buffer will be positioned at the end of the document upon normal completion of the method
      */
     static ByteBufBsonDocument createOne(final ByteBuf outputByteBuf) {
+        int documentStart = outputByteBuf.position();
         int documentSizeInBytes = outputByteBuf.getInt();
-        ByteBuf slice = outputByteBuf.duplicate();
-        slice.position(slice.position() - 4);
-        slice.limit(slice.position() + documentSizeInBytes);
-        outputByteBuf.position(outputByteBuf.position() + documentSizeInBytes - 4);
+        int documentEnd = documentStart + documentSizeInBytes;
+        ByteBuf slice = outputByteBuf.duplicate().position(documentStart).limit(documentEnd);
+        outputByteBuf.position(documentEnd);
         return new ByteBufBsonDocument(slice);
     }
 
--- a/driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonDocument.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonDocument.java
@@ -53,9 +53,12 @@ final class ByteBufBsonDocument extends BsonDocument {
 
     private final transient ByteBuf byteBuf;
 
-    // Create a list of ByteBufBsonDocument from a ByteBuf positioned at the start of the first document of an OP_MSG Section
-    // of type Kind 1: Document Sequence
-    // The provided ByteBuf will be positioned at the end of the section upon normal completion of the method
+    /**
+     * Create a list of ByteBufBsonDocument from a buffer positioned at the start of the first document of an OP_MSG Section
+     * of type Document Sequence (Kind 1).
+     * <p>
+     * The provided buffer will be positioned at the end of the section upon normal completion of the method
+     */
     static List<ByteBufBsonDocument> createList(final ByteBuf outputByteBuf) {
         List<ByteBufBsonDocument> documents = new ArrayList<>();
         while (outputByteBuf.hasRemaining()) {
@@ -65,8 +68,10 @@ final class ByteBufBsonDocument extends BsonDocument {
         return documents;
     }
 
-    // Create a ByteBufBsonDocument from a ByteBuf positioned at the start of a BSON document.
-    // The provided ByteBuf will be positioned at the end of the section upon normal completion of the method
+    /**
+     * Create a ByteBufBsonDocument from a buffer positioned at the start of a BSON document.
+     * The provided buffer will be positioned at the end of the document upon normal completion of the method
+     */
     static ByteBufBsonDocument createOne(final ByteBuf outputByteBuf) {
         int documentSizeInBytes = outputByteBuf.getInt();
         ByteBuf slice = outputByteBuf.duplicate();
--- a/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
@@ -139,7 +139,7 @@ public final class CommandMessage extends RequestMessage {
                         int sequenceStart = byteBuf.position();
                         int sequenceSizeInBytes = byteBuf.getInt();
                         int sectionEnd = sequenceStart + sequenceSizeInBytes;
-                        
+
                         String fieldName = getSequenceIdentifier(byteBuf);
                         // If this assertion fires, it means that the driver has started using document sequences for nested fields.  If
                         // so, this method will need to change in order to append the value to the correct nested document.
--- a/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
@@ -134,12 +134,24 @@ public final class CommandMessage extends RequestMessage {
                     // Each loop iteration processes one Document Sequence
                     // When there are no more bytes remaining, there are no more Document Sequences
                     while (byteBuf.hasRemaining()) {
-                        byteBuf.position(byteBuf.position() + 1 /* payload type */ + 4 /* payload size */);
+                        // skip reading the payload type, we know it is 1
+                        byteBuf.position(byteBuf.position() + 1);
+                        int sequenceStart = byteBuf.position();
+                        int sequenceSizeInBytes = byteBuf.getInt();
+                        int sectionEnd = sequenceStart + sequenceSizeInBytes;
+                        
                         String fieldName = getSequenceIdentifier(byteBuf);
                         // If this assertion fires, it means that the driver has started using document sequences for nested fields.  If
-                        // so, the line following this assertion would have to change in order to handle that situation.
+                        // so, this method will need to change in order to append the value to the correct nested document.
                         assertFalse(fieldName.contains("."));
-                        commandBsonDocument.append(fieldName, new BsonArray(createList(byteBuf)));
+
+                        ByteBuf documentsByteBufSlice = byteBuf.duplicate().limit(sectionEnd);
+                        try {
+                            commandBsonDocument.append(fieldName, new BsonArray(createList(documentsByteBufSlice)));
+                        } finally {
+                            documentsByteBufSlice.release();
+                        }
+                        byteBuf.position(sectionEnd);
                     }
                     return commandBsonDocument;
                 } else {
--- a/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
@@ -30,12 +30,14 @@ import org.bson.BsonDocument;
 import org.bson.BsonElement;
 import org.bson.BsonInt64;
 import org.bson.BsonString;
+import org.bson.ByteBuf;
 import org.bson.FieldNameValidator;
 import org.bson.io.BsonOutput;
 
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.IntStream;
 
 import static com.mongodb.ReadPreference.primary;
 import static com.mongodb.ReadPreference.primaryPreferred;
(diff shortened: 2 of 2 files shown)


driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -107,30 +111,49 @@ public final class CommandMessage extends RequestMessage {
...
+        List<ByteBuf> duplicateByteBuffers = bsonOutput.getByteBuffers();
+        try {
+            CompositeByteBuf outputByteBuf = new CompositeByteBuf(duplicateByteBuffers);
+            outputByteBuf.position(getEncodingMetadata().getFirstDocumentPosition());
+            ByteBufBsonDocument byteBufBsonDocument = createOne(outputByteBuf);
+
+            if (outputByteBuf.hasRemaining()) {
+                BsonDocument commandBsonDocument = byteBufBsonDocument.toBaseBsonDocument();
+
+                while (outputByteBuf.hasRemaining()) {
```
jyemin, line 127: This while loop is how the method supports multiple document sequences.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -107,30 +111,49 @@ public final class CommandMessage extends RequestMessage {
...
+                    outputByteBuf.position(outputByteBuf.position() + 1 /* payload type */ + 4 /* payload size */);
+                    String payloadName = getPayloadName(outputByteBuf);
```
jyemin, line 129: Now we get the payload name directly from the OP_MSG instead of from the SplittablePayload, making it easier to support any number of document sequence independent of any future changes to SplittablePayload

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -107,30 +111,49 @@ public final class CommandMessage extends RequestMessage {
...
-       return commandBsonDocument;
     }
 
-    boolean containsPayload() {
-        return payload != null;
+    // Get the field name from a ByteBuf positioned at the start of the document sequence identifier of an OP_MSG Section of type
+    // Kind 1: Document Sequence.
+    // Upon normal completion of the method, the ByteBuf will be positioned at the start of the first BSON object in the sequence.
+    private String getPayloadName(final ByteBuf outputByteBuf) {
+        List<Byte> payloadNameBytes = new ArrayList<>();
```
jyemin, line 146: Is there a cleaner way to do this?  I couldn't think of one.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -107,30 +111,49 @@ public final class CommandMessage extends RequestMessage {
...
-            int payloadStartPosition = getEncodingMetadata().getFirstDocumentPosition()
-                    + byteBufBsonDocument.getSizeInBytes()
-                    + 1 // payload type
-                    + 4 // payload size
-                    + payload.getPayloadName().getBytes(StandardCharsets.UTF_8).length + 1;  // null-terminated UTF-8 payload name
-            commandBsonDocument.append(payload.getPayloadName(),
-                    new BsonArray(ByteBufBsonDocument.createList(bsonOutput, payloadStartPosition)));
-        } else {
-            commandBsonDocument = byteBufBsonDocument;
+        List<ByteBuf> duplicateByteBuffers = bsonOutput.getByteBuffers();
```
jyemin, line 118: The refactoring in ByteBufBsonDocument allows this method to create the list of byte buffers once and re-use them both for creating the BsonDocument from the Kind 0 Section and the List<BsonDocument> for each Kind 1 Sections.

driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonDocument.java
```diff
@@ -53,38 +53,26 @@ final class ByteBufBsonDocument extends BsonDocument {
 
     private final transient ByteBuf byteBuf;
 
-    static List<ByteBufBsonDocument> createList(final ByteBufferBsonOutput bsonOutput, final int startPosition) {
-        List<ByteBuf> duplicateByteBuffers = bsonOutput.getByteBuffers();
-        CompositeByteBuf outputByteBuf = new CompositeByteBuf(duplicateByteBuffers);
-        outputByteBuf.position(startPosition);
+    // Create a list of ByteBufBsonDocument from a ByteBuf positioned at the start of the first document of an OP_MSG Section
```
jyemin, line 56: This is all refactoring.  No functional changes in this class.  The main benefit is that the caller is now responsible for creating a single ByteBuf instance and passing it in.

driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonDocument.java
```diff
@@ -53,38 +53,26 @@ final class ByteBufBsonDocument extends BsonDocument {
...
-            ByteBuf slice = outputByteBuf.duplicate();
-            slice.position(curDocumentStartPosition);
-            slice.limit(curDocumentStartPosition + documentSizeInBytes);
-            documents.add(new ByteBufBsonDocument(slice));
-            curDocumentStartPosition += documentSizeInBytes;
-            outputByteBuf.position(outputByteBuf.position() + documentSizeInBytes - 4);
-        }
-        for (ByteBuf byteBuffer : duplicateByteBuffers) {
-            byteBuffer.release();
+            ByteBufBsonDocument curDocument = createOne(outputByteBuf);
```
jyemin, line 62: Now this method can simply delegate to `createOne`, removing some code duplication.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: The issue is that `ByteBuffer`, like `byte[]`, doesn't grow.  You have to know the capacity ahead of time.

Another approach would be to save the position, loop until the zero byte to find out how big an array is needed, position the buffer back to the starting position, and fill the byte array.

NathanQingyangXu: Yeah. Seems no silver bullet without knowing the size beforehand. Guava library has a Bytes.toArray util method for the last step but saw it is not in classpath. The stream usage might be much slower than vanilla loop.

NathanQingyangXu: Your second approach makes lots of sense. If we could reposition to previous location, why the class didn't expose a method to provide the size?

NathanQingyangXu: I think the most straightforward way is ByteArrayOutputStream: https://docs.oracle.com/javase/6/docs/api/java/io/ByteArrayOutputStream.html

NathanQingyangXu: Out of curiosity, I compared the per for the following three approaches:

- approach in this PR
- approach same as in PR except that using vanilla loop instead of stream usage
- approach based on ByteArrayOutputStream usage

Below is the time spent for a huge byte stream input:

- approach #1: 93408 seconds
- apporach #2: 20667 seconds
- approach #3: 11342 seconds

It is no wonder that stream usage is 4 or 5 times slower than normal loop; ByteArrayOutputStream usage is fastest.

NathanQingyangXu: As a digression, typically stream usage is 5 times slower than pre-jdk8 way. It is used in non-critical path to enhance readability (another pro is parralelStream() to tap into multi core CPU, though seldom used). As an extreme example, stream usage is only allowed in testing case in Hibernate library. It is always a good idea to inspect whether its usage is really necessary in face of its serious perf overhead, especially in library development.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
stIncMale: If we now get the "payload name" (which is actually, the [sequence identifier](https://www.mongodb.com/docs/manual/reference/mongodb-wire-protocol/#kind-1--document-sequence)), why do we still use `SplittablePayload.getPayloadName`, including using it in `CommandMessage`?

I don't yet have a good understanding of what's going on, but the above seems quite confusing.

jyemin: When writing the OP_MSG Kind 1 Section, we need to get the payload name from somewhere.  It comes from `SplittablePayload`.  This code is just getting it out of the OP_MSG after it's been written, so it no longer has to depend on `SplittablePayload`, which is now free to evolve independent of this method (for example, to support multiple payloads).

jyemin: If you are suggesting that the `getPayloadName` method should be renamed to something like `getSequenceIdentifier`, I agree.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -107,30 +111,62 @@ public final class CommandMessage extends RequestMessage {
...
+                if (byteBuf.hasRemaining()) {
+                    BsonDocument commandBsonDocument = byteBufBsonDocument.toBaseBsonDocument();
+
+                    // Each loop iteration processes one Document Sequence
+                    // When there are no more bytes remaining, there are no more Document Sequences
+                    while (byteBuf.hasRemaining()) {
+                        byteBuf.position(byteBuf.position() + 1 /* payload type */ + 4 /* payload size */);
+                        String payloadName = getPayloadName(byteBuf);
+                        assertFalse(payloadName.contains("."));
+                        commandBsonDocument.append(payloadName, new BsonArray(createList(byteBuf)));
```
stIncMale, line 138: `createList(byteBuf)` never stops, until it reaches the end of `byteBuf`. That is a problem, given that we are trying to decode multiple sections, as opposed to decoding a single section. Given that we expect the payload of all sections to be sequences, we need to read the sequence size and use `createList` to decode the documents in the sequence only up to the position where the sequence ends.

Unfortunately, given that `CommandMessage` can't currently have multiple `SplittablePayload`s, we can neither test the PR code, nor the proposed fix for it. 

Following is how I propose to fix the code:

```suggestion
                        // skip reading the payload type, we know it is 1
                        byteBuf.position(byteBuf.position() + 1);
                        int sequenceStart = byteBuf.position();
                        int sequenceSizeInBytes = byteBuf.getInt();
                        int sectionEnd = sequenceStart + sequenceSizeInBytes;
                        String sequenceIdentifier = getSequenceIdentifier(byteBuf);
                        assertFalse(sequenceIdentifier.contains("."));
                        ByteBuf documentsByteBufSlice = byteBuf.duplicate().limit(sectionEnd);
                        byteBuf.position(sectionEnd);
                        commandBsonDocument.append(sequenceIdentifier, new BsonArray(createList(documentsByteBufSlice)));
```

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -107,30 +111,62 @@ public final class CommandMessage extends RequestMessage {
...
+        while (curByte != 0) {
+            payloadNameBytes.add(curByte);
+            curByte = outputByteBuf.get();
+        }
+
+        // Convert List<Byte> to byte[]
+        byte[] byteArray = new byte[payloadNameBytes.size()];
+        IntStream.range(0, payloadNameBytes.size()).forEachOrdered(i -> byteArray[i] = payloadNameBytes.get(i));
+        return new String(byteArray, StandardCharsets.UTF_8);
     }
```
stIncMale, line 170: I agree with @NathanQingyangXu's [suggestion](https://github.com/mongodb/mongo-java-driver/pull/1456/files?show-viewed-files=true&file-filters%5B%5D=#r1688958770). The code will look as follows, I believe (unfortunately, we have to deal with `UnsupportedEncodingException` because the `toString` method that does not throw it exists only since Java SE 10):

```suggestion
    private String getSequenceIdentifier(final ByteBuf byteBuf) {
        ByteArrayOutputStream sequenceIdentifierBytes = new ByteArrayOutputStream();
        byte curByte = byteBuf.get();
        while (curByte != 0) {
            sequenceIdentifierBytes.write(curByte);
            curByte = byteBuf.get();
        }
        try {
            return sequenceIdentifierBytes.toString(StandardCharsets.UTF_8.name());
        } catch (UnsupportedEncodingException e) {
            throw new RuntimeException(e);
        }
    }
```

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -107,30 +111,62 @@ public final class CommandMessage extends RequestMessage {
...
+                // If true, it means there is at least one Kind 1:Document Sequence in the OP_MSG
+                if (byteBuf.hasRemaining()) {
+                    BsonDocument commandBsonDocument = byteBufBsonDocument.toBaseBsonDocument();
+
+                    // Each loop iteration processes one Document Sequence
+                    // When there are no more bytes remaining, there are no more Document Sequences
+                    while (byteBuf.hasRemaining()) {
+                        byteBuf.position(byteBuf.position() + 1 /* payload type */ + 4 /* payload size */);
+                        String payloadName = getPayloadName(byteBuf);
+                        assertFalse(payloadName.contains("."));
```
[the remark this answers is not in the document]
stIncMale: I think, I figured this out: dots are allowed in principle, but this code is not prepared to correctly decode them, which is why we have the assertion. Is this understanding correct?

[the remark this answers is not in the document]
jyemin: Yes, exactly.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -107,30 +111,62 @@ public final class CommandMessage extends RequestMessage {
...
+        List<ByteBuf> byteBuffers = bsonOutput.getByteBuffers();
+        try {
+            CompositeByteBuf byteBuf = new CompositeByteBuf(byteBuffers);
+            try {
+                byteBuf.position(getEncodingMetadata().getFirstDocumentPosition());
+                ByteBufBsonDocument byteBufBsonDocument = createOne(byteBuf);
+
+                // If true, it means there is at least one Kind 1:Document Sequence in the OP_MSG
+                if (byteBuf.hasRemaining()) {
+                    BsonDocument commandBsonDocument = byteBufBsonDocument.toBaseBsonDocument();
```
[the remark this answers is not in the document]
jyemin: I don't think it's a performance issue.  Consider two cases:

1.  A `findAndModify` command containing a large document.  In this case there is only a single Kind 0 section.  In this case, the check for `byteBuf.hasRemaining()` on line 129 will be false, and so `toBaseBsonDocument()` will not be called.  In this case, the optimization in  `ByteBufBsonDocument.getFirstKey` is helpful.
2. An `insert` command containing one or more large documents in a Kind 1 section.  In this case the check for `byteBuf.hasRemaining()` will be true, and  `toBaseBsonDocument()` will be called and the BsonDocument will be hydrated into a Map.  But the large documents in the Kind 1 Section will not.  The optimization in `ByteBufBsonDocument.getFirstKey` is no longer used, but that's ok.  We've avoided hydration of the large documents and only hydrated the small number of fields in the Kind 0 document.  It's true that this is relying on what we know to be true about how and when we use Kind 1 Sections, but I think that's ok.

> It seems that we can skip decoding all but the very first section with kind 0 (which in turn can be fake decoded via createOne(byteBuf), just like it's done now) when we decode for the purpose of getting the value of the first key. Moreover, that can be further simplified by just having a method that literally reads the first BSON key from ByteBufferBsonOutput starting at getEncodingMetadata().getFirstDocumentPosition().

Yes, I think you're correct about this.  I can work on this as part of this PR or we can ticket it separately.

> I was also wondering why we have to do this complex decoding with optimizations that fake decode, just to get to the command name, when we could have just propagated the command name from the place where we know it via CommandMessage to the place in InternalStreamConnection where we need it. Is it because we have MongoDatabase.runCommand, where we don't know what the command is until we encode the command?

Yes, we don't know the command name until we encode it when it's coming from runCommand, because runCommand can take any `Bson`.

The complex decoding was built not to just get the command name in `InternalStreamConnection` but rather to get the entire command in `LoggingCommandEventSender`for the purpose of publishing it in a `CommandStartedEvent`.  That's where we're really trying to avoid hydrating the whole document, when often consumers of that event just want to look at a couple of fields.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: I committed and pushed this change locally, as I wanted to change one small thing about the suggestion: generally I think we throw MongoInternalException in these situations instead of RuntimeException.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: Resolving based on Valentin's suggestion below, which has been committed.

driver-core/src/main/com/mongodb/internal/connection/ByteBufBsonDocument.java
```diff
@@ -53,38 +53,31 @@ final class ByteBufBsonDocument extends BsonDocument {
...
-        slice.position(startPosition);
-        slice.limit(startPosition + documentSizeInBytes);
-        for (ByteBuf byteBuffer : duplicateByteBuffers) {
-            byteBuffer.release();
-        }
+        slice.position(slice.position() - 4);
+        slice.limit(slice.position() + documentSizeInBytes);
+        outputByteBuf.position(outputByteBuf.position() + documentSizeInBytes - 4);
         return new ByteBufBsonDocument(slice);
     }
```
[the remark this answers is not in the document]
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: This all goes a long way back, to when there were many more subclasses of RequestMessage (we used to have subclasses for OP_INSERT, OP_UPDATE, OP_DELETE, etc).  But now, there is only CommandMessage, and except for the annoyance of having to use OP_QUERY for the very first message we send on a connection, CommandMessage only encodes OP_MSG.  Which makes `EncodingMetadata#getFirstDocumentPosition` effectively a constant.  

When we removed support for all the other op codes, we simplified a lot, but there is more that can be done.  Just not sure it should be done in this PR.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: I'm going to grab this suggestion and commit it locally, since I've made further changes that conflict.

jyemin: I basically used this as is, except I added a release of `documentsByteBufSlice` in a finally block.

Thanks for find this.  It was careless of me.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -107,30 +113,76 @@ public final class CommandMessage extends RequestMessage {
...
-    boolean containsPayload() {
-        return payload != null;
+    /**
+     * Get the field name from a buffer positioned at the start of the document sequence identifier of an OP_MSG Section of type
+     * Document Sequence (Kind 1).
+     * <p>
+     * Upon normal completion of the method, the buffer will be positioned at the start of the first BSON object in the sequence.
+    */
+    private String getSequenceIdentifier(final ByteBuf byteBuf) {
+        ByteArrayOutputStream sequenceIdentifierBytes = new ByteArrayOutputStream();
```
NathanQingyangXu, line 175: thanks for adopting my suggestion! A further improvement is to give a good guess of 'default size' to avoid resizing again and again, but it is out of my knowledge scope to do that. If no clue, it is good enough.

jyemin: The default for ByteArrayOutputStream is 32, which is large enough for all known sequence identifiers.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
stIncMale: > I added a release of `documentsByteBufSlice`

🤦‍♂️ Thank you for spotting that)

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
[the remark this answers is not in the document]
stIncMale: Thank you for the explanations. I agree with everything but the very conclusion:

> We've avoided hydration of the large documents and only hydrated the small number of fields in the Kind 0 document.

yes, though compared to just decoding the first key this is still quite a lot.

> I can work on this as part of this PR or we can ticket it separately.

Let's do that separately from this PR: https://jira.mongodb.org/browse/JAVA-5553.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
stIncMale: Thank you, I now understand why it was varying previously, but not the reasons behind the design that I described previously. I created https://jira.mongodb.org/browse/JAVA-5554 about improving this.

stIncMale (approved):

PR #1460, opened 2024-07-25
Kotlin updated to 1.9.x

rozza:
JAVA-5538


driver-kotlin-sync/src/test/kotlin/com/mongodb/kotlin/client/MongoIterableTest.kt
```diff
@@ -93,7 +93,7 @@ class MongoIterableTest {
         assertContentEquals(documents.subList(0, 2), iterable.use { it.take(2) }.toList())
 
         verify(delegate, times(1)).cursor()
-        verify(cursor, times(2)).hasNext()
+        verify(cursor, times(3)).hasNext()
```
rozza, line 96: Due to a change in the kotlin internals - it now checks `hasNext` 3 times. But only calls `next` twice.

rozza:
Closing for now - as will need to plan when we upgrade.

PR #1461, opened 2024-07-25
Kotlin: Updated driver metadata

rozza:
Updated the driver name to include which driver type. Added Kotlin version to the platform metadata.

JAVA-5539


rozza:
I reverted the changes to the driver name as they were implicit. Here is an example of the metadata:

ashni-mongodb (approved):

PR #1464, opened 2024-07-30
Add ANNOTATION_TYPE for other annotation to inherite this @BsonId

godotg:
I want to extend the functionality of @BsonId by inheriting them, but currently custom annotations can't inherit @BsonId. This can make the code bloated.

https://github.com/zfoo-project/zfoo/blob/main/orm/src/test/java/com/zfoo/orm/entity/MailEntity.java

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/annotations/BsonId.java
+++ b/bson/src/main/org/bson/codecs/pojo/annotations/BsonId.java
@@ -33,6 +33,6 @@ import java.lang.annotation.Target;
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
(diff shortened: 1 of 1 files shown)


rozza:
Thank you @godotg for the PR. I've added [JAVA-5556](https://jira.mongodb.org/browse/JAVA-5556) to track this request and will triage in the near future.

rozza:
@godotg I'm closing because even if you were to annotate `@Id` with `@BsonId` the convention logic is not recursive to check the annotations annotations. As such custom conventions code would still be required.

Your code that adds its own convention to add support for `@Id` is the recommended way to extend Pojo functionality.

Closing won't fix.

PR #1465, opened 2024-07-31
Support Range Indexes as GA.

vbabanin:
JAVA-5537
JAVA-5441


driver-core/src/main/com/mongodb/client/model/vault/RangeOptions.java
```diff
@@ -81,6 +81,7 @@ public RangeOptions max(@Nullable final BsonValue max) {
      * @return the trim factor value if set
      * @since 5.2
      */
+    @Nullable
```
NathanQingyangXu, line 84: How to guard against ripple effect after the method respone were made nullable? Like some direct or indirect caller might end up with NPE. The concern might be invalid in this scenario and nullness annotations might be enough. I have no idea.
How to guard against ripple effect after the method respone were made nullable? Like some direct or indirect caller might end up with NPE. The concern might be invalid in this scenario and nullness annotations might be enough. You know better. I chimed in only for learning purpose.

In my recent open source project I started experimenting the Checker framework usage for at such early stage its adoption is still possible. In theory it could catch all possible NPE during compiling phase.
In my recent open source project I started experimenting the Checker framework usage for at such early stage its adoption is still possible. In theory it could catch all possible NPE during compiling phase and I am willing to invest for that grand goal.

driver-sync/src/test/functional/com/mongodb/fixture/EncryptionFixture.java
vbabanin, line 1: I've consolidated the common initialization logic, previously repeated across encryption tests, into a new EncryptionFixture.

driver-sync/src/test/functional/com/mongodb/client/JsonPoweredCrudTestHelper.java
```diff
@@ -305,11 +305,10 @@ BsonDocument getRunCommandResult(final BsonDocument collectionOptions, final Bso
                 response = database.runCommand(clientSession, command, readPreference, BsonDocument.class);
             }
         }
-        response.remove("ok");
```
vbabanin, line 308: According to the [legacy test format specification](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/legacy-test-format.md#use-as-integration-tests): 

> If the operation returns a raw command response, e.g., from runCommand, then compare only the fields present in the expected result document.

Therefore, we should not remove any fields from the response. Instead, we should only compare the fields that are explicitly expected and ignore any additional fields that are not specified in the expected results.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionRangeDefaultExplicitEncryptionTest.java
```diff
@@ -0,0 +1,131 @@
...
+                        .min(new BsonInt32(0))
+                        .max(new BsonInt32(1000))
+                );
+    }
+
+    @AfterEach
+    public void cleanUp() {
+        try (ClientEncryption ignored = clientEncryption) {
+            // just using try-with-resources to ensure they all get closed, even in the case of exceptions
+        }
```
stIncMale, line 91:
```
AbstractClientSideEncryptionRangeDefaultExplicitEncryptionTest.java:89: warning: [try] auto-closeable resource ignored is never referenced in body of corresponding try statement
        try (ClientEncryption ignored = clientEncryption) {
                              ^
```

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionRangeDefaultExplicitEncryptionTest.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Validates that the omission of options trimFactor and sparsity leads to libmongocrypt-provided defaults being used instead.
+     */
+    @Test
+    @DisplayName("Case 1: Uses libmongocrypt defaults")
+    void shouldUserDefaultsWhenNotSpecified() {
```
stIncMale, line 99:
```suggestion
    void shouldUsesDefaultsWhenNotSpecified() {
```

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionRangeDefaultExplicitEncryptionTest.java
```diff
@@ -0,0 +1,131 @@
...
+import org.bson.BsonInt32;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import static com.mongodb.ClusterFixture.isServerlessTest;
+import static com.mongodb.ClusterFixture.isStandalone;
+import static com.mongodb.ClusterFixture.serverVersionAtLeast;
+import static com.mongodb.assertions.Assertions.assertTrue;
```
stIncMale, line 38: We should use `org.junit.jupiter.api.Assertions.assertTrue` instead for checks in tests.

P.S. In the rare occasion when we need an assertion about the testing code itself, we should indeed use `com.mongodb.assertions.Assertions`, but here we don't have such a situation.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionRangeDefaultExplicitEncryptionTest.java
```diff
@@ -0,0 +1,131 @@
...
+ * <a href="https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.md#22-range-explicit-encryption-applies-defaults">
+ */
+
+public abstract class AbstractClientSideEncryptionRangeDefaultExplicitEncryptionTest {
+    private static final BsonInt32 VALUE_TO_ENCRYPT = new BsonInt32(123);
+    private ClientEncryption clientEncryption;
+    private BsonBinary keyId;
+    private BsonBinary payloadDefaults;
+
+    protected abstract MongoClient createMongoClient(MongoClientSettings settings);
```
stIncMale, line 55: This method is not used and is not needed.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ClientSideEncryptionRangeDefaultExplicitEncryptionTest.java
```diff
@@ -0,0 +1,40 @@
...
+public class ClientSideEncryptionRangeDefaultExplicitEncryptionTest extends AbstractClientSideEncryptionRangeDefaultExplicitEncryptionTest {
+
+    @Override
+    protected MongoClient createMongoClient(final MongoClientSettings settings) {
+        return new SyncMongoClient(MongoClients.create(settings));
+    }
+
+    @Override
+    protected ClientEncryption createClientEncryption(final ClientEncryptionSettings settings) {
+        return ClientEncryptions.create(settings);
```
stIncMale, line 38: Reactive test must use `com.mongodb.reactivestreams.client.vault.ClientEncryptions.create`, and then wrap the result in `SyncClientEncryption`.

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/fixture/EncryptionFixture.java
```diff
@@ -0,0 +1,83 @@
...
+ * Helper class for the CSFLE/QE tests.
+ */
+public final class EncryptionFixture {
+
+    private EncryptionFixture() {
+        //NOP
+    }
+
+    public static Map<String, Map<String, Object>> getKmsProviders(final KmsProviderType... kmsProviderTypes) {
+        return new HashMap<String, Map<String, Object>>() {{
```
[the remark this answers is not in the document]
vbabanin: I've reused this approach for readability purposes as it seems less bulky. Since the suggestion is optional, I'll maintain the current implementation. However, I'm open to revisiting this if you feel strongly about it.

driver-sync/src/main/com/mongodb/client/vault/ClientEncryption.java
```diff
@@ -98,9 +96,6 @@ public interface ClientEncryption extends Closeable {
      *     </ul>
      *     {@code $gt} may also be {@code $gte}. {@code $lt} may also be {@code $lte}.
      *
-     * <p>Only supported when queryType is "range" and algorithm is "Range".
-     * <p>Note: The Range algorithm is unstable. It is subject to breaking changes.
```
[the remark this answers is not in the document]
vbabanin: Good catch—it was removed by accident. I've updated it now.

driver-core/src/main/com/mongodb/client/model/vault/RangeOptions.java
```diff
@@ -81,18 +77,21 @@ public RangeOptions max(@Nullable final BsonValue max) {
...
+     * Set the number of top-level edges stored per record.
+     * <p>
+     * The trim factor may be used to tune performance.
+     *
      * @param trimFactor the trim factor
      * @return this
      * @since 5.2
      */
-    public RangeOptions setTrimFactor(final Integer trimFactor) {
+    public RangeOptions setTrimFactor(@Nullable final Integer trimFactor) {
```
rozza, line 94: Nit: Can we deprecate this add a `trimFactor` method. Using the API having the `set` prefix really sticks out.

vbabanin: Absolutely, good catch! I've renamed it to `trimFactor()`.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionRangeDefaultExplicitEncryptionTest.java
stIncMale: When trying to fix a typo I made another one myself - way to go 👍

stIncMale (approved):

rozza (approved):
LGTM!

PR #1467, opened 2024-08-01
Connection String

rozza:
Don't output the host and port information if the port is invalid. Reduces risk of leaking password information if the password has not been correctly urlencoded.

JAVA-5560

### DIFF
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -1209,7 +1209,7 @@ public class ConnectionString {
                 }
                 int idx = host.indexOf("]:");
                 if (idx != -1) {
-                    validatePort(host, host.substring(idx + 2));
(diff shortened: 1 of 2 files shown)


rozza:
Note: In the unlikely case of the password looking like a port eg: within numeric range and ending in a slash `mongodb://foo:1234/@hostname` the password would be interpreted as the port.

jyemin (commented):
Is there anything similarly dangerous in `ConnectionString#parseHosts`?

rozza:
@jyemin I think all is covered. An unescaped `@` sign eg: `mongodb://foo:bar@mypass@localhost` is already handled and errors with:

> The connection string contains invalid user information. If the username or password contains a colon (:) or an at-sign (@) then it must be urlencoded

If the user were to accidently escape the `@` sign then it would fail with the new error message about the port.

I added a couple of extra regression tests.

jyemin (approved):
LGTM!

PR #1470, opened 2024-08-01
Add optional casting methods

jeffalder:
Because `BsonValue` has boolean `is` methods, and throwing `as` methods, a common pattern is:

```java
BsonValue value;
if (value.isSomeType()) {
  BsonSomeType valueSomeType = value.asSomeType();
  // .. operate on valueSomeType
}
```

Nested conditionals are even harder:
```java
BsonValue value;
if (value.isDocument() && value.asDocument().get("key").isInt32()) {
  //... dig in and get it
}
```

By using Java 8's `Optional`, we can simplify these with a fluent syntax:
```java
BsonValue value;
value.asSomeTypeOpt().ifPresent(valueSomeType => {
  // valueSomeType is guaranteed to be BsonSomeType
});

Optional<BsonInt32> nested = value.asDocumentOpt()
  .flatMap(doc => doc.get("key").asInt32Opt());
```

I can also add `get*Opt` methods to BsonDocument if they make sense.


NathanQingyangXu:
Interesting idea. I want to provide 2 cwnts as food of thought:

* the naming of 'getxxxOpt()' might not be most intuitive (if you google 'opt', first reaction is 'make a choice'). There is no well-defined naming convention for usually Optional is advised to be used as internal variable. But if we have to return it, other naming might be considered, e.g. getOptionalXxx() seems a reasonable choice as well (more verbose but less confusing)
* other method design without exposing Optional is possible. Like 'ifXXX(Consumer<> consumer)

My 2 cents only

jeffalder:
> the naming of 'getxxxOpt()' might not be most intuitive (if you google 'opt', first reaction is 'make a choice'). There is no well-defined naming convention for usually Optional is advised to be used as internal variable. But if we have to return it, other naming might be considered, e.g. getOptionalXxx() seems a reasonable choice as well (more verbose but less confusing)

I grant full naming rights to the code owners.

> other method design without exposing Optional is possible. Like 'ifXXX(Consumer<> consumer)' or even 'ifXXX(Consumer<> consumer, Runnable else)'.

The methods you propose are `Optional::ifPresent` and `Optional::ifPresentOrElse`: implementing them here would start replicating the `Optional` interface... which just demonstrates that `Optional` _is_ the right choice.

Truly, this is a bridge anyway -- once you get to Java 17, you can use pattern matching on `instanceof` and all the is/as methods can be ignored.

```java
BsonValue value;
if (value instanceof BsonInt32 int32Value) {
  // ... act on int32Value
}
```

NathanQingyangXu:
> > the naming of 'getxxxOpt()' might not be most intuitive (if you google 'opt', first reaction is 'make a choice'). There is no well-defined naming convention for usually Optional is advised to be used as internal variable. But if we have to return it, other naming might be considered, e.g. getOptionalXxx() seems a reasonable choice as well (more verbose but less confusing)
> 
> I grant full naming rights to the code owners.
> 
> > other method design without exposing Optional is possible. Like 'ifXXX(Consumer<> consumer)' or even 'ifXXX(Consumer<> consumer, Runnable else)'.
> 
> The methods you propose are `Optional::ifPresent` and `Optional::ifPresentOrElse`: implementing them here would start replicating the `Optional` interface... which just demonstrates that `Optional` _is_ the right choice.
> 
> Truly, this is a bridge anyway -- once you get to Java 17, you can use pattern matching on `instanceof` and all the is/as methods can be ignored.
> 
> ```java
> BsonValue value;
> if (value instanceof BsonInt32 int32Value) {
>   // ... act on int32Value
> }
> ```

Good point on JDK's instanceof usage. So it seems such syntax sugars are not fully justified?

jeffalder:
> So it seems such syntax sugars are not fully justified?

I think they are justified, as long as this library supports Java 8 or 11.

PR #1472, opened 2024-08-07
PojoCodec - use both BsonDiscriminator key and value for lookup

rozza:
Previously, only the value of a discriminator was used as the key for discriminator lookups. Now both key and values are used.

Updated the PojoCodecProvider to use a LinkedHashMap to keep the registration order of classes / classModels.

Added an ERROR level log if duplicate BsonDiscriminators are discovered.

JAVA-5566

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/DiscriminatorLookup.java
+++ b/bson/src/main/org/bson/codecs/pojo/DiscriminatorLookup.java
@@ -17,34 +17,38 @@
 package org.bson.codecs.pojo;
 
 import org.bson.codecs.configuration.CodecConfigurationException;
+import org.bson.diagnostics.Logger;
+import org.bson.diagnostics.Loggers;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import static java.lang.String.format;
 
 final class DiscriminatorLookup {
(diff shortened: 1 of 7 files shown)


bson/src/main/org/bson/codecs/pojo/DiscriminatorLookup.java
```diff
@@ -57,7 +61,12 @@ public Class<?> lookup(final String discriminator) {
 
     void addClassModel(final ClassModel<?> classModel) {
         if (classModel.getDiscriminator() != null) {
-            discriminatorClassMap.put(classModel.getDiscriminator(), classModel.getType());
+            Discriminator discriminator = Discriminator.create(classModel);
+            Class<?> cachedClass = discriminatorClassMap.computeIfAbsent(discriminator, (cachedDiscriminator) -> classModel.getType());
+            if (cachedClass != classModel.getType()) {
+                LOGGER.error(format("Duplicate %s registered for both: `%s` and `%s`",
```
rozza, line 67: Opted to not throw an exception as this may be discovered during runtime. 

I'm not 100% sure if throwing an exception would be considered a breaking change. In theory if a user was just writing data then there is potential that the code reading the data could be configured differently (multiple PojoCodecProviders) to allow for conflict free reading. 🤷 

So opted to log an error.

cc @jyemin - any thoughts?

rozza:
Closing as it would be better to link the `BsonDiscriminator` with a relevant class hierarchy.

PR #1476, opened 2024-08-09
Fix most of the build warnings

stIncMale:
The following commit introduced many warnings https://github.com/mongodb/mongo-java-driver/commit/01aff5a0789f71b9d0b56190d4996e8ac5436827. Many of them were then fixed in https://github.com/mongodb/mongo-java-driver/commit/cd297a13d1868e6a22b88529016a17fda618363d. This PR fixes most of the ones that are still present.

### DIFF
--- a/driver-core/src/test/unit/com/mongodb/internal/operation/OperationUnitSpecification.groovy
+++ b/driver-core/src/test/unit/com/mongodb/internal/operation/OperationUnitSpecification.groovy
@@ -68,7 +68,7 @@ class OperationUnitSpecification extends Specification {
             [9, 0]: 25,
     ]
 
-    static int getMaxWireVersionForServerVersion(List<Integer> serverVersion) {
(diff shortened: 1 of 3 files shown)


driver-core/src/test/unit/com/mongodb/internal/operation/OperationUnitSpecification.groovy
```diff
@@ -68,7 +68,7 @@ class OperationUnitSpecification extends Specification {
             [9, 0]: 25,
     ]
 
-    static int getMaxWireVersionForServerVersion(List<Integer> serverVersion) {
+    static Integer getMaxWireVersionForServerVersion(List<Integer> serverVersion) {
```
stIncMale, line 71: I don't understand the warning we had for this method, but the change removed that warning. I don't want to spend more time on investigating the weirdness of Groovy.

vbabanin (approved):
LGTM! Thanks for addressing those issues!

PR #1477, opened 2024-08-09
Do minor improvements to how we use field validators

stIncMale:
A few improvements I noticed while working on the new `bulkWrite` command support. Doing them separately here to avoid drive-by changes in the implementation of `bulkWrite`.


jyemin (approved):
LGTM

PR #1480, opened 2024-08-14
change modifiedCount to type long

danibs:
`modifiedCount` should be `long` instead of `Long`. If not:

```java
public long getModifiedCount() {
    return modifiedCount;
}
```

could throws exception.

**NOTE: The `public int hashCode()` method must to be updated accordingly**


rozza:
Hi @danibs thanks for the PR.

Apologies for the delay in responding. I have opened [JAVA-5621](https://jira.mongodb.org/browse/JAVA-5621) to track this and will triage it shortly.

All the best,

Ross

jyemin:
See linked ticket for rationale.

PR #1481, opened 2024-08-15
Refactor `MixedBulkWriteOperation` a bit and change `CommandMessage.isResponseExpected` such that it accounts for ordered/unordered bulk writes

stIncMale:


driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
```diff
@@ -101,12 +101,7 @@ static BulkWriteBatch createBulkWriteBatch(final MongoNamespace namespace,
                                                       final List<? extends WriteRequest> writeRequests,
                                                       final OperationContext operationContext,
                                                       @Nullable final BsonValue comment, @Nullable final BsonDocument variables) {
-        SessionContext sessionContext = operationContext.getSessionContext();
-        if (sessionContext.hasSession() && !sessionContext.isImplicitSession() && !sessionContext.hasActiveTransaction()
-                && !writeConcern.isAcknowledged()) {
-            throw new MongoClientException("Unacknowledged writes are not supported when using an explicit session");
-        }
```
stIncMale, line 108: This validation was moved to `MixedBulkWriteOperation.validateAndGetEffectiveWriteConcern`.

driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
```diff
@@ -169,9 +164,8 @@ private BulkWriteBatch(final MongoNamespace namespace, final ConnectionDescripti
         if (!payloadItems.isEmpty()) {
             command.put(getCommandName(batchType), new BsonString(namespace.getCollectionName()));
             command.put("ordered", new BsonBoolean(ordered));
-            if (!writeConcern.isServerDefault() && !sessionContext.hasActiveTransaction()) {
-                command.put("writeConcern", writeConcern.asDocument());
-            }
```
stIncMale, line 174: This logic needs to be reused in the implementation of the new `bulkWrite` command, so I extracted it into `MixedBulkWriteOperation.commandWriteConcern`.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -405,31 +414,47 @@ private boolean handleMongoWriteConcernWithResponseExceptionAsync(final RetrySta
...
-                : getAppliedWriteConcern(sessionContext).isAcknowledged();
+    static Optional<WriteConcern> commandWriteConcern(final WriteConcern effectiveWriteConcern, final SessionContext sessionContext) {
+        return effectiveWriteConcern.isServerDefault() || sessionContext.hasActiveTransaction()
+                ? Optional.empty()
+                : Optional.of(effectiveWriteConcern);
+    }
+
+    private boolean shouldExpectResponse(final BulkWriteBatch batch, final WriteConcern effectiveWriteConcern) {
+        return effectiveWriteConcern.isAcknowledged() || (ordered && batch.hasAnotherBatch());
     }
```
stIncMale, line 458: `shouldExpectResponse` replaces the `shouldAcknowledge` method. The new method has simpler body and named in a way that avoids erroneous allusions to an unacknowledged (`w: 0`) write concern.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -253,7 +253,7 @@ private boolean requireOpMsgResponse() {
         if (responseExpected) {
```
jyemin, line 253: We can simplify this class bit further by adding the following assertion in the constructor:

```
assertFalse(getOpCode() == OpCode.OP_QUERY && !responseExpected);
```

this assertion should never fire because OP_QUERY effectively is only used for the initial `hello` command on a new connection, which should always be expecting a response.

Then we can remove the `!useOpMsg()` condition in `isResponseExpected`, and inline this method.

driver-core/src/main/com/mongodb/internal/operation/BulkWriteBatch.java
jyemin: Will it eventually move to some common "helper" class?

stIncMale: I had no such plans, as I don't see any value in creating a somewhat useless class just to hold this method, such that both `MixedBulkWriteOperation` and `ClientBulkWriteOperation` could call `commandWriteConcern` on that class. Calling `commandWriteConcern` on `MixedBulkWriteOperation` seems to be not worse in any way, especially given that when it is statically imported, one can't even notice immediately see where the method comes from.

driver-sync/src/test/functional/com/mongodb/client/MongoClientSessionSpecification.groovy
```diff
@@ -350,9 +351,11 @@ class MongoClientSessionSpecification extends FunctionalSpecification {
 
     void waitForInsertAcknowledgement(MongoCollection<Document> collection, ObjectId id) {
         Document document = collection.find(Filters.eq(id)).first()
+        Timeout timeout = Timeout.expiresIn(5, TimeUnit.SECONDS, Timeout.ZeroSemantics.ZERO_DURATION_MEANS_INFINITE)
```
jyemin, line 354: Is this change relevant to the rest of the PR?

jyemin (approved):

driver-sync/src/test/functional/com/mongodb/client/MongoClientSessionSpecification.groovy
stIncMale: Sorry, I thought I posted my comments, but they were pending. Yes, here are the relevant comments:
- https://github.com/mongodb/mongo-java-driver/pull/1481#discussion_r1718712117
- https://github.com/mongodb/mongo-java-driver/pull/1481#discussion_r1718764877

PR #1486, opened 2024-08-22
Implement Java sync improved bulk write API and unified spec tests

stIncMale:
The first three commits are well-organized to allow reviewing them one by one.

This PR depends on https://github.com/mongodb/mongo-java-driver/pull/1458.

The following test runners execute the unified and prose tests added in this PR:

- `com.mongodb.client.unified.VersionedApiTest`
- `com.mongodb.client.unified.UnifiedTransactionsTest`
- `com.mongodb.client.unified.ServerSelectionLoggingTest`
- `com.mongodb.client.unified.UnifiedRetryableWritesTest`
- `com.mongodb.client.unified.CommandMonitoringTest`
- `com.mongodb.client.unified.UnifiedCrudTest`
- `com.mongodb.client.CrudProseTest`

JAVA-5528
JAVA-5609


driver-core/src/main/com/mongodb/internal/operation/BsonDocumentWrapperHelper.java
```diff
@@ -25,7 +26,12 @@ final class BsonDocumentWrapperHelper {
 
     @SuppressWarnings("unchecked")
     static <T> List<T> toList(final BsonDocument result, final String fieldContainingWrappedArray) {
-        return ((BsonArrayWrapper<T>) result.getArray(fieldContainingWrappedArray)).getWrappedArray();
+        // BULK-TODO why does the expectation of this code fails?
```
rozza, line 29: If the document doesn't contain a wrapped array why would you use this helper?

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -0,0 +1,1051 @@
...
+            // and if it is the last attempt failure, `RetryingSyncSupplier` also may not have a chance
+            // to add the label. So we do that explicitly.
+            shouldAttemptToRetryWriteAndAddRetryableLabel(retryState, e, false);
+            resultAccumulator.onBulkWriteCommandErrorWithoutResponse(e);
+            throw e;
+        }
+    }
+
+    /**
+     * @throws MongoWriteConcernWithResponseException This internal exception must be handled to avoid it being observed by an application.
```
jyemin, line 245: Please remind me why we have to both throw and catch this weird exception within even the same class?

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -0,0 +1,1051 @@
...
+            writer.writeName("document");
+            Object document = model.getDocument();
+            @SuppressWarnings("unchecked")
+            Encoder<Object> documentEncoder = (Encoder<Object>) codecRegistry.get(document.getClass());
+            assertNotNull(encodedBatchInfo).insertModelDocumentIds.compute(modelIndexInBatch, (k, knownModelDocumentId) -> {
+                IdHoldingBsonWriter documentIdHoldingBsonWriter = new IdHoldingBsonWriter(
+                        writer,
+                        // Reuse `knownModelDocumentId` if it may have been generated by `IdHoldingBsonWriter` in a previous attempt.
+                        // If its type is not `BsonObjectId`, we know it could not have been generated.
+                        knownModelDocumentId instanceof BsonObjectId ? knownModelDocumentId.asObjectId() : null);
```
jyemin, line 943: Running unified tests with coverage enabled, there are no occurrences of  ` knownModelDocumentId instanceof BsonObjectId` evaluating to `true`.   Is there a prose test for this, or if not should we create one?  Or could a unified test be written for it?

jyemin (changes requested):
Given the complexity, the code is quite straightforward to follow.  Really nice.

driver-core/src/main/com/mongodb/internal/operation/BsonDocumentWrapperHelper.java
stIncMale: I am simply constructing a new `CommandBatchCursor`. And for some reason, the `result` document, which I received from a `Connection.command` call that executed `bulkWrite`, contains a normal `BsonArray` in its `cursor.firstBatch` field, instead of containing a `BsonArrayWrapper` as the code in `BsonDocumentWrapperHelper.toList` expected. This does not seem right, given that in all other cases the expectation holds.

driver-sync/src/main/com/mongodb/client/internal/MongoClusterImpl.java
```diff
@@ -365,6 +381,14 @@ private <TResult> ChangeStreamIterable<TResult> createChangeStreamIterable(@Null
                 retryReads, timeoutSettings);
     }
 
+    private ClientBulkWriteResult executeBulkWrite(
+            @Nullable final ClientSession clientSession,
+            final List<? extends ClientNamespacedWriteModel> clientWriteModels,
+            @Nullable final ClientBulkWriteOptions options) {
+        isTrue("`autoEncryptionSettings` is null, as automatic encryption is not yet supported", autoEncryptionSettings == null);
```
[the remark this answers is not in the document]
stIncMale: Auto-encryption is will not be supported in `bulkWrite` at first, as per https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#auto-encryption. All TODO's are expressed as TODO's for now.

driver-core/src/main/com/mongodb/ClientBulkWriteException.java
```diff
@@ -56,16 +58,23 @@ public final class ClientBulkWriteException extends MongoServerException {
...
             @Nullable final Map<Long, WriteError> writeErrors,
             @Nullable final ClientBulkWriteResult partialResult,
             final ServerAddress serverAddress) {
-        super(message(error, writeConcernErrors, writeErrors, partialResult, serverAddress), serverAddress);
+        super(
+                message(
+                        error, writeConcernErrors, writeErrors, partialResult,
+                        notNull("serverAddress", serverAddress)),
+                validateServerAddress(error, serverAddress));
         // BULK-TODO Should ClientBulkWriteException.getCode be the same as error.getCode,
```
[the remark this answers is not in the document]
stIncMale: We should decide this here https://github.com/mongodb/mongo-java-driver/pull/1458#discussion_r1688758112.

driver-core/src/main/com/mongodb/internal/operation/BsonDocumentWrapperHelper.java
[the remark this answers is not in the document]
stIncMale: I explained that here https://github.com/mongodb/mongo-java-driver/pull/1486#discussion_r1732120952.

> not sure how it relates to the PR

Neither do I. I would expect things to just work, but in this case they didn't, which is suspicious. I changed the code in `BsonDocumentWrapperHelper.toList` such that it works, but why did I have to do that, why does it work for all other commands? Those are the questions I don't have answers to at the moment.

driver-core/src/main/com/mongodb/internal/operation/CommandOperationHelper.java
```diff
@@ -153,7 +153,14 @@ static boolean shouldAttemptToRetryRead(final RetryState retryState, final Throw
...
 
-    static boolean shouldAttemptToRetryWrite(final RetryState retryState, final Throwable attemptFailure) {
+    static boolean shouldAttemptToRetryWriteAndAddRetryableLabel(final RetryState retryState, final Throwable attemptFailure) {
+        return shouldAttemptToRetryWriteAndAddRetryableLabel(retryState, attemptFailure, true);
+    }
+
+    static boolean shouldAttemptToRetryWriteAndAddRetryableLabel(
+            final RetryState retryState,
+            final Throwable attemptFailure,
+            final boolean logIfDecideToNotRetry) {
```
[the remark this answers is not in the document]
stIncMale: Done in bfbc1cc26685f2e0a76832a9d05d2e78dca94ac7.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -0,0 +1,1051 @@
...
+    }
+
+    private BsonDocumentWrapper<?> createBulkWriteCommand(
+            final boolean effectiveRetryWrites,
+            final WriteConcern effectiveWriteConcern,
+            final SessionContext sessionContext,
+            final List<? extends ClientNamespacedWriteModel> unexecutedModels,
+            final BatchEncoder batchEncoder,
+            final Runnable ifCommandIsRetryable) {
+        // BULK-TODO This implementation must limit the number of `models` it includes in a batch if needed.
```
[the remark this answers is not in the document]
stIncMale: There are multiple restrictions to comply with, but conceptually that's all the same thing - "limit the number of models in a batch", and it better be done the same way in the same place. So the answer is "yes".

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
stIncMale: 1) Our retry logic is built around exceptions: they are used to signal that an attempt has failed, to carry the error codes and the retryable label. A write concern error is, however, a weird one:

2) `Connection.command` does not throw when a write concern error happens.
3) There may be a useful server response accompanying the write concern error.

`1)` and `2)` mean that we must throw an "artificial" exception when a write concern error happens, if we want retries. `3)` means that we also must preserve the result when doing so, as it becomes useful if our "artificial" exception turns out to be the failure chosen as the outcome of all attempts. That's why we attach the response to the exception. And given that the exception is "artificial", we must ensure that once the retrying logic has done its work, we destroy the exception and never expose it to an application.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
stIncMale: Done in 1c3c590ce10bb5f4682ddaf2e3596636ca023700.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -0,0 +1,1051 @@
...
+                        assertFalse(unexecutedModels.isEmpty());
+                        LinkedHashMap<MongoNamespace, Integer> indexedNamespaces = new LinkedHashMap<>();
+                        writer.writeStartArray("ops");
+                        boolean commandIsRetryable = effectiveRetryWrites;
+                        for (int modelIndexInBatch = 0; modelIndexInBatch < unexecutedModels.size(); modelIndexInBatch++) {
+                            ConcreteClientNamespacedWriteModel modelWithNamespace = getModelWithNamespace(unexecutedModels, modelIndexInBatch);
+                            ClientWriteModel model = modelWithNamespace.getModel();
+                            if (commandIsRetryable && !modelSupportsRetries.apply(model)) {
+                                commandIsRetryable = false;
+                                logWriteModelDoesNotSupportRetries();
```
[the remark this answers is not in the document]
stIncMale: We can. I think, that's how it is done.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -435,7 +435,7 @@ operationContext, shouldExpectResponse(batch, effectiveWriteConcern),
                 batch.getPayload(), batch.getFieldNameValidator(), callback);
     }
 
-    private static WriteConcern validateAndGetEffectiveWriteConcern(final WriteConcern writeConcernSetting, final SessionContext sessionContext)
+    static WriteConcern validateAndGetEffectiveWriteConcern(final WriteConcern writeConcernSetting, final SessionContext sessionContext)
```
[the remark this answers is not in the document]
stIncMale: Done in 1c9c19e7344d410f4adf6b53bd59bdadd757cd41. But let's not embark on the journey of refactoring other operations.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -1760,6 +1778,226 @@ public OperationResult createChangeStreamCursor(final BsonDocument operation) {
         });
     }
 
+    public OperationResult clientBulkWrite(final BsonDocument operation) {
+        Set<String> unexpectedOperationKeys = singleton("saveResultAsEntity");
```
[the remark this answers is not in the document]
stIncMale: This [operation field](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#operation) is handled (or not) by each operation, and not by a piece of code common to all operations. It seems reasonable for all operations that don't implement it to fail if the field is encountered. I don't know why other operations don't do that. Maybe because it is not expected to be encountered for them.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -1760,6 +1778,226 @@ public OperationResult createChangeStreamCursor(final BsonDocument operation) {
...
+        });
+    }
+
+    private static ClientWriteModelWithNamespace toClientWriteModelWithNamespace(final BsonDocument model) {
+        String modelType = model.getFirstKey();
+        BsonDocument arguments = model.getDocument(modelType);
+        MongoNamespace namespace = new MongoNamespace(arguments.getString("namespace").getValue());
+        switch (modelType) {
+            case "insertOne":
+                Set<String> expectedArguments = new HashSet<>(asList("namespace", "document"));
```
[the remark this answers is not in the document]
stIncMale: Done in 061e605188f9e065fc2085a1c56f519d8a7b225e.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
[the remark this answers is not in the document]
jyemin: It seems like it will log once per model.  Am I misunderstanding the code?

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -0,0 +1,1051 @@
...
+            final BsonDocumentWrapper<?> lazilyEncodedCommand,
+            final List<? extends ClientNamespacedWriteModel> unexecutedModels,
+            final WriteConcern effectiveWriteConcern,
+            final OperationContext operationContext) throws MongoWriteConcernWithResponseException {
+        BsonDocument bulkWriteCommandOkResponse = connection.command(
+                "admin",
+                lazilyEncodedCommand,
+                FieldNameValidators.createUpdateModsFieldValidator(unexecutedModels),
+                null,
+                codecRegistry.get(BsonDocument.class),
```
jyemin, line 265:
```suggestion
                CommandResultDocumentCodec.create(codecRegistry.get(BsonDocument.class), 
                        CommandBatchCursorHelper.FIRST_BATCH)
```

driver-core/src/main/com/mongodb/internal/operation/BsonDocumentWrapperHelper.java
jyemin: My suggested edit in ClientBulkWriteOperation should fix this. That's what all the other cursor-initiating command do.

jyemin (changes requested):

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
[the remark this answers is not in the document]
stIncMale: Here is how I understand the code:

1. When `!modelSupportsRetries.apply(model)` is true, i.e., `model` does not support retries, we enter the `if` block.
2. There we do `commandIsRetryable = false` and log.
3. The `if` condition can never be `true` again for the same batch, because it starts with `commandIsRetryable && ...`.
4. Thus, we log at most once per batch.

driver-core/src/main/com/mongodb/internal/operation/BsonDocumentWrapperHelper.java
stIncMale: Thank you!, it worked indeed.

stIncMale:
@jyemin Requested to implement all the prose tests that are implementable at this point in this PR. I will do that.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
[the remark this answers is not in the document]
jyemin: Ah, ok

stIncMale:
The spec authors keep mutating existing DRIVERS tickets instead of creating new ones, which is detrimental, because it drastically increases the chances of a change being missed. The following unified test changes need to be included in this PR: https://github.com/mongodb/specifications/commit/0984b0942b9d8aaa11610184d0be16b27a263ec3.

driver-sync/src/test/functional/com/mongodb/client/CrudProseTest.java
```diff
@@ -16,181 +16,411 @@
...
+                    ClientNamespacedWriteModel.insertOne(namespace(), new Document("a", "b"))));
+            assertEquals(maxWriteBatchSize + 1, result.getInsertedCount());
+            assertEquals(2, startedBulkWriteCommandEvents.size());
+            CommandStartedEvent firstEvent = startedBulkWriteCommandEvents.get(0);
+            CommandStartedEvent secondEvent = startedBulkWriteCommandEvents.get(1);
+            assertEquals(maxWriteBatchSize, firstEvent.getCommand().getArray("ops").size());
+            assertEquals(1, secondEvent.getCommand().getArray("ops").size());
+            assertEquals(firstEvent.getOperationId(), secondEvent.getOperationId());
         }
+    }
```
stIncMale, line 222: I realized this test can't pass for now only after implementing it. Decided to leave the implementation, though the test is skipped.

driver-sync/src/test/functional/com/mongodb/client/CrudProseTest.java
```diff
@@ -16,181 +16,411 @@
...
+                    model = ClientNamespacedWriteModel.replaceOne(namespace(), Filters.empty(), document);
+                    break;
+                }
+                default: {
+                    throw Assertions.fail(operationType);
+                }
             }
+            assertThrows(MongoClientException.class, () -> client.bulkWrite(singletonList(model)));
+        }
+    }
```
stIncMale, line 259: I realized this test can't pass for now only after implementing it. Decided to leave the implementation, though the test is skipped.

stIncMale:
@jyemin, @rozza, please re-review.

jyemin (approved):
LGTM!

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -0,0 +1,1052 @@
...
+    private BsonDocumentWrapper<?> createBulkWriteCommand(
+            final boolean effectiveRetryWrites,
+            final WriteConcern effectiveWriteConcern,
+            final SessionContext sessionContext,
+            final List<? extends ClientNamespacedWriteModel> unexecutedModels,
+            final BatchEncoder batchEncoder,
+            final Runnable ifCommandIsRetryable) {
+        // BULK-TODO This implementation must limit the number of `models` it includes in a batch if needed.
+        // Each batch re-selects a server and re-checks out a connection because this is simpler and it is allowed,
+        // see https://mongodb.slack.com/archives/C035ZJL6CQN/p1722265720037099?thread_ts=1722264610.664109&cid=C035ZJL6CQN.
```
rozza, line 333: Nit: These archives will be lost over time, so recommend adding a tldr here instead.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
```diff
@@ -700,6 +700,14 @@ public void test10CustomTestWithTransactionUsesASingleTimeoutWithLock() {
         }
     }
 
+    @DisplayName("11. Multi-batch bulkWrites")
+    @Test
+    void test11MultiBatchBulkWrites() {
+        assumeTrue(serverVersionAtLeast(8, 0));
+        assumeFalse(isServerlessTest());
+        assumeTrue(Runtime.getRuntime().availableProcessors() < 1, "BULK-TODO implement prose test https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/tests/README.md#11-multi-batch-bulkwrites");
```
rozza, line 708: Does this have a jira ticket - if not lets add one and then that can be tracked there.

driver-sync/src/test/functional/com/mongodb/client/CrudProseTest.java
```diff
@@ -16,181 +16,411 @@
...
+
+        }
+    }
+
+    @DisplayName("3. MongoClient.bulkWrite batch splits a writeModels input with greater than maxWriteBatchSize operations")
+    @Test
+    void testBulkWriteSplitsWhenExceedingMaxWriteBatchSize() {
+        assumeTrue(serverVersionAtLeast(8, 0));
+        assumeFalse(isServerlessTest());
+        assumeTrue(Runtime.getRuntime().availableProcessors() < 1, "BULK-TODO implement prose test https://github.com/mongodb/specifications/blob/master/source/crud/tests/README.md#3-mongoclientbulkwrite-batch-splits-a-writemodels-input-with-greater-than-maxwritebatchsize-operations");
```
rozza, line 165: Same as previous - in CSOT we'd add a jira ticket to track the todos there.

rozza (changes requested):
Looks really good.

Just had a couple of suggestions. Less worried about the TODO's but I do recommend summarizing the slack discussion in the comment.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
stIncMale: Done in 14bb86ef3b832c743962f3cb1cb2c565284e3ec6. I moved the comment, and linked the relevant `DRIVERS` ticket there (the ticket will not be done, but it has the relevant discussion).

driver-sync/src/test/functional/com/mongodb/client/CrudProseTest.java
stIncMale: The reply is also the same:) https://github.com/mongodb/mongo-java-driver/pull/1486#discussion_r1752076281

driver-core/src/test/resources/unified-test-format/server-selection/logging/operation-id.json
```diff
@@ -224,6 +227,192 @@
...
+                "newDescription": {
+                  "type": "Unknown"
+                }
+              }
+            },
+            "count": 1
+          }
+        },
+        {
+          "name": "bulkWrite",
```
jyemin, line 348: Shouldn't this be `clientBulkWrite`?  When I run locally this test fails looking for a collection named `client`.

stIncMale: Thank you, it should. It haven't noticed this because I've been only running tests against a replica set, where this test is skipped. This is how it's in the [spec|https://github.com/mongodb/specifications/blob/10919c92cb784e9730551841e57541bee2e1af93/source/server-selection/tests/logging/operation-id.json#L346], so I'll file a relevant DRIVERS ticket

stIncMale: Created https://jira.mongodb.org/browse/DRIVERS-2978 with a [PR](https://github.com/mongodb/specifications/pull/1659).

stIncMale: @jyemin Could you please move https://jira.mongodb.org/browse/DRIVERS-2978 to "Ready for work"? I don't have the permission to do that.

PR #1487, opened 2024-08-27
Integrate mongodb-crypt module 

vbabanin:
## Description
This PR relocates the mongocrypt Java bindings from the libmongocrypt repository to the mongo-java-driver repository to streamline maintenance and the release process.

Key changes to the Gradle build script, originally copied from the libmongocrypt repository, include:

- Removal of Maven publication configuration; now managed by the publish.gradle script.
- Removal of signing configurations; managed by the the publish.gradle script.
- Removal of explicit Slf4j dependency, as it is inherited from the root build script located [here](https://github.com/vbabanin/mongo-java-driver/blob/ceb18e5e59d917195c1a4463704ca01b530f17a6/build.gradle#L178).

### To aid the review process, I have taken several steps:

- Downloaded the mongocrypt version 1.11 from [Maven Central](https://mvnrepository.com/artifact/org.mongodb/mongodb-crypt/1.11.0) and unzipped the source, Javadoc, runtime jar, and pom into the git folder.
- Run a Maven local publication specifically for the mongocrypt module using the command` gradle mongocrypt:clean mongocrypt:publishToMavenLocal -x signMavenJavaPublication` to produce a publication. Subsequently, I unzipped and applied a git diff over the ones from version 1.11. The diff can be viewed [here](https://github.com/vbabanin/crypt-content-test/pull/1/files#).

Notable differences include many Javadoc variations due to the mongocrypt module inheriting configurations from javadoc.gradle. Key areas to focus on are the pom and manifest files, along with some source files modified by the Spotless plugin, which corrected redundant imports.

I have built the jar files using JDK 11.0.24 (Homebrew), which is the same version used to build 1.11 currently on Maven Central.

## Note for Reviewers:
I would appreciate it if reviewers could be particularly vigilant when reviewing the build configurations and produced artifacts in this PR. If you have any doubts or cannot be certain that the changes do not impact previously guaranteed functionalities, please do not hesitate to raise your concerns.

NOTE: Publishing snapshot to Sonatype has not been tested yet.

JAVA-5582
JAVA-5583


mongocrypt/.gitignore
```diff
@@ -0,0 +1,39 @@
...
+*.idea
+workspace.xml
+atlassian-ide-plugin.xml
+
+# gradle
+.gradle
+
+# local settings
+**/gradle.properties
+local.properties
```
NathanQingyangXu, line 32: This is super minor, just for fun of tech discussion. In .gitignorr, "filrname" means the same as "**/filename" but the latter takes effect only after git v1.8.2

The only reason to use "**" is for such example: "a/**/b". 

No need to change if it has been a pattern elsewhere already. Feel free to ignore, without saying.

mongocrypt/build.gradle.kts
```diff
@@ -0,0 +1,191 @@
...
+base.archivesBaseName = "mongodb-crypt"
+description = "MongoDB client-side crypto support"
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+dependencies {
+    api(project(path = ":bson", configuration = "default"))
```
vbabanin, line 46: The latest version of BSON is now used, as opposed to the version specified in the libmongocrypt repository. For reference, see the previous dependency listed here: [libmongocrypt build.gradle.kts](https://github.com/mongodb/libmongocrypt/blob/f9adce53f6e676cb226b33b411501fe0b6faeca0/bindings/java/mongocrypt/build.gradle.kts#L61).

mongocrypt/build.gradle.kts
```diff
@@ -0,0 +1,196 @@
...
+   dependsOn("downloadJnaLibs")
+}
+
+tasks.javadoc {
+    if (JavaVersion.current().isJava9Compatible) {
+        (options as StandardJavadocDocletOptions).addBooleanOption("html5", true)
+    }
+}
+
+afterEvaluate {
```
vbabanin, line 181: Manifest is preserved as it is in libmongocrypt, except `"Import-Package" to "org.bson.*",`.

mongocrypt/build.gradle.kts
```diff
@@ -0,0 +1,196 @@
...
+                "Import-Package" to "org.bson.*",
+                "Build-Version" to gitVersion,
+                "Bundle-Version" to gitVersion,
+                "Bundle-Name" to "MongoCrypt",
+                "Bundle-SymbolicName" to "com.mongodb.crypt.capi",
+                "Private-Package" to ""
+            )
+        }
+    }
+}
```
vbabanin, line 196: Manifest is preserved as it is from libmongocrypt, except "Import-Package" to "org.bson.*",.

mongocrypt/.gitignore
vbabanin: Thanks for the insightful comment! We currently lack consistency in the root `.gitignore`. We could consider addressing this if it becomes a source of confusion.

As for now, I've removed the `.gitignore` from this module since we don’t need an additional one; there’s already a `.gitignore` in the root folder of our repo.

mongocrypt/build.gradle.kts
```diff
@@ -0,0 +1,199 @@
...
+afterEvaluate {
+    tasks.jar {
+        manifest {
+            attributes(
+                "-exportcontents" to "com.mongodb.crypt.capi.*;-noimport:=true",
+                "Automatic-Module-Name" to "com.mongodb.crypt.capi",
+                "Import-Package" to "org.slf4j.*;resolution:=optional,org.bson.*",
+                "Bundle-Name" to "MongoCrypt",
+                "Bundle-SymbolicName" to "com.mongodb.crypt.capi",
+                "Private-Package" to ""
```
vbabanin, line 195: Manifest configuration is preserved as it is from libmongocrypt, except for:

- `"Import-Package" `now uses a wildcard `"org.slf4j.*;resolution:=optional,org.bson.*"`,
- `"Build-Version" to gitVersion` is handled by [publish.gradle#L67](https://github.com/mongodb/mongo-java-driver/blob/c4af1c5fe20a88807bf4fa9c0071f6126424ade3/gradle/publish.gradle#L67).
- `"Bundle-Version" to gitVersion` is handled by [publish.gradle#L68](https://github.com/vbabanin/mongo-java-driver/blob/c4af1c5fe20a88807bf4fa9c0071f6126424ade3/gradle/publish.gradle#L68).

mongocrypt/build.gradle.kts
```diff
@@ -0,0 +1,197 @@
...
+    eachFile {
+        path = "${jnaMapping[path.substringBefore("/")]}/${name}"
+    }
+    into(jnaResourcesDir)
+    mustRunAfter("downloadJava")
+
+    doLast {
+        println("jna.library.path contents: \n  ${fileTree(jnaResourcesDir).files.joinToString(",\n  ")}")
+    }
+}
```
vbabanin, line 134: Gradle tests whether task `inputs` or `outputs` have changed since the last build as part of the incremental build process. If no changes are detected, it considers the task up-to-date and skips execution.

There are two time-consuming tasks now using incremental builds to improve build efficiency:

- downloadJava: Downloads `libmongocrypt` binaries from S3. Enhanced with the `onlyIfModified(true)` option, this task avoids redundant downloads by fetching binaries only if they are absent from the build output.

The `localBinariesArchiveName` variable in the `./build` folder, containing the zip archive file name with the revision number, ensures that binaries are downloaded just once per revision/tag unless a `gradle clean` command explicitly clears the build folder.

- unzipJava: Unzips binaries into the temporary resource folder. Previously, this task was configured with `outputs.upToDateWhen { false }` to always execute. Now, it triggers only under these conditions:
    - The input file name changes, detected from the `downloadRevision` variable, prompting the unzipping of the latest or previously saved binaries.
    - Any alteration or corruption in the output file structure within `jnaResourcesDir`. To ensure correct unzipping, it first cleans `jnaResourcesDir` and then proceeds to unzip the binaries.

With `mongocrypt` as a `project dependency` of `driver-core`, the incremental build process conserves time when running or testing code unrelated to the `mongocrypt module` in sync/async modules. It avoids unnecessary rebuilding of the `mongocrypt jar` for the `driver-core module`. This is also useful during active development as `libmongocrypt` commits may update frequently.

driver-benchmarks/src/main/com/mongodb/benchmark/framework/MongoCryptBenchmarkRunner.java
vbabanin, line 1: Please note, the benchmark runner is directly copied from mongocrypt with minor adjustments (added MongoCryptBenchmarkResult). It has not been refactored to better align with the benchmark framework; we may consider refactoring in a separate scope. Functionally, it executes the tests, which can be verified by running performance tests on this build and checking the trends graph.

jyemin (changes requested):
Note that static checks are failing: 

```
The class name com.mongodb.crypt.capi.CAPI$cstring doesn't start with an upper case letter
```
so we'll have to suppress that.  This is actually a spotbugs error not checkstyle, which surprises me.

mongocrypt/build.gradle.kts
```diff
@@ -0,0 +1,188 @@
...
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+dependencies {
+    api(project(path = ":bson", configuration = "default"))
+    api("net.java.dev.jna:jna:5.11.0")
+
+    // Tests
+    testImplementation("org.junit.jupiter:junit-jupiter")
+    testRuntimeOnly("ch.qos.logback:logback-classic:1.2.13")
```
jyemin, line 51: Can this use `$logbackVersion`?

jyemin, line 51: Everywhere else this is `testImplementation`.  Does it need to be different here?

jyemin (changes requested):

rozza (approved):
LGTM - fixed a couple of nits, I had discovered doing a diff between the jar contents.

Also, fixed the spotbugs issue.

Only thing that could be improved is the javadoc warnings for the CAPI file.

mongocrypt/build.gradle.kts
rozza: This line can be removed as its set in the main `build.gradle`

mongocrypt/build.gradle.kts
```diff
@@ -0,0 +1,188 @@
...
+        | gitRevision    : Optional Git Revision to download the built resources for from s3.
+    """.trimMargin()
+}
+
+tasks.jar {
+    //NOTE this enables depending on the mongocrypt from driver-core
+   dependsOn("downloadJnaLibs")
+}
+
+tasks.javadoc {
```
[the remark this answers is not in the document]
vbabanin: Yes, this will be covered under JAVA-5584, which includes removing Javadoc generation for this module. However, there's a `MongoCryptException` that might not be moved to the internal package, so we could consider retaining the Javadoc because of this.

mongocrypt/build.gradle.kts
vbabanin: I removed this line as per @rozza suggestion here: https://github.com/mongodb/mongo-java-driver/pull/1487#discussion_r1763156734

jyemin (approved):
LGTM!

vbabanin:
I'm holding off on merging this as I noticed an issue after running the publish snapshot. The `driver-core` POM in the snapshot no longer marks `mongodb-crypt` as an optional dependency, despite it being specified as such in the build configuration: `implementation project(path: ':mongocrypt', configuration: 'default'), optional`. The setting isn’t being applied correctly. [Here's the POM for reference](https://oss.sonatype.org/content/repositories/snapshots/org/mongodb/mongodb-driver-core/5.2.0-SNAPSHOT/mongodb-driver-core-5.2.0-20240919.011153-130.pom).

rozza:
Hoping to have fixed the 'optional' support - will review the build in the morning

jyemin:
@rozza do you expect/desire that the optional mongodb-crypt dependency appears in driver-core, driver-sync, and driver-reactive-streams? Currently it's only in driver-core (even though driver-core doesn't actually use it!).

jyemin:
I think the latest changes might have broken the build.  I see a lot of

```
FAILURE: java.lang.NoClassDefFoundError: com/mongodb/crypt/capi/MongoCryptOptions (java.lang.NoClassDefFoundError)
```

vbabanin:
Thanks, @rozza for your help on this issue! 

I identified the root cause of the issue with the existing nebula plugin for optional dependencies. The mismatch between the `mongocrypt `submodule name and the artifactId `mongodb-crypt` used during publishing caused the plugin to skip appending the optional flag as detailed in [this plugin code](https://github.com/nebula-plugins/gradle-extra-configurations-plugin/blob/c14cbeb060ee23f56fba62fa189c427dd32ee652/src/main/groovy/nebula/plugin/extraconfigurations/OptionalBasePlugin.groovy#L66).

To address this, I renamed the module name to `mongodb-crypt` while keeping the directory as `mongocrypt`. All tests pass, and the artifact now correctly includes the optional flag.

I've pushed these changes to a separate branch to avoid overriding current changes and I've run the full set of Evergreen tests, and all tests have passed successfully. You can review the results [here](https://spruce.mongodb.com/version/66eca01a31bb9e00072b1000/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

The commit with the fix is here: [6e3e7068f1b0f2648a2fa0610e2c13a3f1dd2e3f](https://github.com/vbabanin/mongo-java-driver/commit/6e3e7068f1b0f2648a2fa0610e2c13a3f1dd2e3f)

@jyemin @rozza please review it, and if everything looks good, feel free to cherry-pick it into this PR.

rozza:
@vbabanin your change looks good, great catch with the naming.  I'm thinking renaming the directory to match the archive name also makes sense. Then we don't have to worry about changing the projects name in settings.gradle

> @rozza do you expect/desire that the optional mongodb-crypt dependency appears in driver-core, driver-sync, and driver-reactive-streams? Currently it's only in driver-core (even though driver-core doesn't actually use it!).

I was trying to workaround the feature support limitations but seems it failed anyway. driver-core has to take a dependency as some of the options classes shared between sync and reactivestreams use the library.  Thankfully, @vbabanin discovered the cause of the plugin failing 👍

jyemin (approved):
LGTM again!

jyemin:
evergreen retry

PR #1488, opened 2024-08-28
JAVA-5505 minor refactoring

katcharov:
JAVA-5505

Mostly-automated refactoring to remove IDE nags. I've left unused methods like `getCurrentRegistrationCount` in `AsynchronousTlsChannelGroup`.

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/tlschannel/async/AsynchronousTlsChannelGroup.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/tlschannel/async/AsynchronousTlsChannelGroup.java
@@ -159,7 +159,7 @@ public class AsynchronousTlsChannelGroup {
 
     private final Selector selector;
 
-    final ExecutorService executor;
+    private final ExecutorService executor;
 
     private final ScheduledThreadPoolExecutor timeoutExecutor =
             new ScheduledThreadPoolExecutor(
@@ -226,6 +226,10 @@ public class AsynchronousTlsChannelGroup {
     /** Creates an instance of this class, using as many thread as available processors. */
     public AsynchronousTlsChannelGroup() {
         this(Runtime.getRuntime().availableProcessors());
+    }
+
+    void submit(final Runnable r) {
+        executor.submit(r);
     }
 
     RegisteredSocket registerSocket(TlsChannel reader, SocketChannel socketChannel) {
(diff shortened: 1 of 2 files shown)


stIncMale (commented):
I am not sure it's wise to refactor `tlschannel` code, which at some point we may have to synchronize with the upstream again. Detecting which discrepancies are functional and must not be lost when we synchronize, and which were done for other reasons, may be more difficult if we change the `tlschannel` code without necessity.

driver-core/src/main/com/mongodb/internal/connection/tlschannel/async/AsynchronousTlsChannelGroup.java
```diff
@@ -159,7 +159,7 @@ static final class WriteOperation extends Operation {
 
     private final Selector selector;
 
-    final ExecutorService executor;
+    private final ExecutorService executor;
```
katcharov, line 162: @stIncMale I can roll back the finals. The reduced visibility refactor would ensure that (in the near future) the user-set executor is not shut down from outside this class. Up to you - is this worth the potential future merge conflict?

stIncMale: This looks like something worth discussing the team on the catchup meeting, or asking Jeff/Ross here, as they are likely the only ones who have idea on our process of syncing the `tlschannel` code.

katcharov: Force rebased. Removed finals commit, kept visibility commit.

driver-core/src/main/com/mongodb/internal/connection/tlschannel/async/AsynchronousTlsChannelGroup.java
```diff
@@ -702,7 +706,7 @@ public void shutdownNow() {
      *
      * @return whether the channel is terminated
      */
-    public boolean isTerminated() {
+    private boolean isTerminated() {
```
stIncMale, line 709: This change seems irrelevant to making sure that the rest of the `tlschannel` code can't terminate a user-supplied executor. If that's so, let's not do the change.

stIncMale (changes requested):

stIncMale (approved):

katcharov:
Test failures appear to be unrelated.

PR #1495, opened 2024-09-09
Implement splitting and encoding `ops`, `nsInfo` as separate `OP_MSG` sections, implement prose tests

stIncMale:
The first four commits are well-organized to allow reviewing them one by one.

This PR depends on https://github.com/mongodb/mongo-java-driver/pull/1486.

The following test runners execute the unified and prose tests added in this PR:

- `com.mongodb.client.CrudProseTest`
- `com.mongodb.client.AbstractClientSideOperationsTimeoutProseTest`

JAVA-5529
JAVA-5610
JAVA-5695


driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -309,34 +314,30 @@ private List<List<BsonDocument>> exhaustBulkWriteCommandOkResponseCursor(
...
-        try (BatchCursor<BsonDocument> cursor = cursorDocumentToBatchCursor(
+        try (CommandBatchCursor<BsonDocument> cursor = cursorDocumentToBatchCursor(
                 TimeoutMode.CURSOR_LIFETIME,
                 response,
                 serverDefaultCursorBatchSize,
                 codecRegistry.get(BsonDocument.class),
                 options.getComment().orElse(null),
                 connectionSource,
                 connection)) {
+            cursor.setCloseWithoutTimeoutReset(true);
```
stIncMale, line 325: `BULK-TODO` Why is this needed?

P.S. If undoing this, roll `SyncOperationHelper.cursorDocumentToBatchCursor` back.

stIncMale: Asked a relevant question here https://mongodb.slack.com/archives/C035ZJL6CQN/p1726019394102369.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -675,7 +538,7 @@ ClientBulkWriteResult build(@Nullable final MongoException topLevelError, final
                         writeConcernErrors.add(writeConcernException.getWriteConcernError());
                     }
                     int batchStartModelIndex = batchResult.getBatchStartModelIndex();
-                    ExhaustiveBulkWriteCommandOkResponse response = batchResult.getResponse();
+                    ExhaustiveClientBulkWriteCommandOkResponse response = batchResult.getResponse();
                     haveSuccessfulIndividualOperations = haveSuccessfulIndividualOperations
                             || response.getNErrors() < batchResult.getBatchModelsCount();
```
stIncMale, line 543: `BULK-TODO` This logic seems incorrect. See https://mongodb.slack.com/archives/C035ZJL6CQN/p1726074791410459.

jyemin (changes requested):
I was unable to envision the design I would like to see without actually implementing it, so I did that on a [branch](https://github.com/jyemin/mongo-java-driver/commits/JAVA-5529-jeff/).  Please take a look, and if it looks mostly like a reasonable way forward we can merge it into your branch and review it as part of this PR.

driver-core/src/test/unit/com/mongodb/internal/connection/CommandMessageSpecification.groovy
```diff
@@ -149,7 +151,10 @@ class CommandMessageSpecification extends Specification {
     def 'should get command document'() {
```
stIncMale, line 151: `BULK-TODO` Test `OpsAndNsInfo` here.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
stIncMale: https://jira.mongodb.org/browse/DRIVERS-2975

jyemin (changes requested):

driver-sync/src/main/com/mongodb/client/internal/CryptConnection.java
```diff
@@ -87,13 +93,19 @@ public ConnectionDescription getDescription() {
...
-            @Nullable final SplittablePayload payload, @Nullable final FieldNameValidator payloadFieldNameValidator) {
+            final OperationContext operationContext, final boolean responseExpected, final OpMsgSequences sequences) {
 
         if (serverIsLessThanVersionFourDotTwo(wrapped.getDescription())) {
             throw new MongoClientException("Auto-encryption requires a minimum MongoDB version of 4.2");
         }
 
+        SplittablePayload payload = null;
+        FieldNameValidator payloadFieldNameValidator = null;
+        if (sequences instanceof ValidatableSplittablePayload) {
```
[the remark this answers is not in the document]
stIncMale: Done in ffd4f75e3bba4060c578e009e0b34d9e1d35319b. I added assertion instead of throwing `MongoClientException`: `CryptConnection` cannot observe anything but `ValidatableSplittablePayload`, `EmptyMessageSequences` because of the `isTrue("`autoEncryptionSettings` is null, as bulkWrite does not currently support automatic encryption", autoEncryptionSettings == null)` check in `MongoClusterImpl`.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
stIncMale: Moved to https://github.com/mongodb/mongo-java-driver/pull/1458/files?show-viewed-files=true&file-filters%5B%5D=#r1765904656.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
stIncMale: `BULK-TODO` We need to implement https://jira.mongodb.org/browse/JAVA-5615 in `master` for this PR.

stIncMale:
@jyemin I incorporated the [changes|https://github.com/jyemin/mongo-java-driver/commits/JAVA-5529-jeff/] you proposed. See ea0633e1a9d336cc49fdbc0115d7cddfacdae717, 9fe35e4c9d4e44d4c930bc9cbbd03442a3fb99c3.

However, I did not include https://github.com/jyemin/mongo-java-driver/commit/7d7f3eb9c9bb7b44e35890af065ef9c3200c3014, because I understand it. `CommandMessage.encodeMessageBodyWithMetadata` in the `main` branch is already concerned with `getSettings().getMaxDocumentSize()`; `BsonWriterHelper.getPayloadMessageSettings` is concerned with `DOCUMENT_HEADROOM`; `RequestMessage.addDocument` creates a `BsonBinaryWriter` that checks that the document size is not greater than `settings.getMaxDocumentSize() + DOCUMENT_HEADROOM`. Given that, what is the purpose of moving similar validation checks to the operation layer specifically when it comes to `DualMessageSequences`?

jyemin:
> Given that, what is the purpose of moving similar validation checks to the operation layer specifically when it comes to DualMessageSequences?

I think the difference is that the checks you enumerated apply equally to all commands: no command document can be larger than maxDocumentSize + DOCUMENT_HEADROOM, period.  Whereas for `bulkWrite` there is specific "business logic" that applies only to that command.  That's why it seemed more appropriate to consolidate that logic with the operation.  Let me know your thoughts on this.

Also, wanted to check about https://github.com/jyemin/mongo-java-driver/commit/a2805a6bd1041c61299f94e5311dc1f37923dd43, since you didn't mention that commit.

jyemin (commented):
Waiting for feedback from Valentin

stIncMale:
@jyemin 

Each time I loot at this, I not only can't see a problem with the current approach, but even fail to convince myself that the proposed change makes things better overall.

> for bulkWrite there is specific "business logic" that applies only to that command: the check applies to each document in the array, but only if it's a replace or insert

The logic that is aware of which documents are stored, and which ones are not, is in `ClientBulkWriteOperation`. Neither `CommandMessage` nor `BsonWriterHelper` are aware of it.

> and only if write concern is unacknowledged. That's why it seemed more appropriate to consolidate that logic with the operation.

The logic that is aware acknowledged writes and of what an application has requested, is in `ClientBulkWriteOperation`. `CommandMessage` knows only whether a response needs be requested from the server or not, and it's a reasonable thing to be aware of at this level of abstraction. `CommandMessage` treats different `MessageSequences` differently regardless of the document size validation, and treating them differently when it comes to making a decision on whether size validation should be done at all is no different (if we have no way to learn about exceeded size from the server, we have to validate the size ourselves).

If for some reason we want `CommandMessage` to not treat different `MessageSequences` differently when it comes to size validation, we would have to move the logic of encoding the command document into an operation, like we do when encoding extra elements. This way `ClientBulkWriteOperation` would be able to decide whether to validate the size of the command document, which can exceed the limit, for example, because of a large `comment` or `let`.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -201,27 +222,49 @@ MongoNamespace getNamespace() {
...
-                bsonOutput.writeInt32(0);         // size
-                bsonOutput.writeCString(payload.getPayloadName());
-                writePayload(new BsonBinaryWriter(bsonOutput, payloadFieldNameValidator), bsonOutput, getSettings(),
-                        messageStartPosition, payload, getSettings().getMaxDocumentSize());
-
-                int payloadBsonOutputLength = bsonOutput.getPosition() - payloadBsonOutputStartPosition;
-                bsonOutput.writeInt32(payloadBsonOutputStartPosition, payloadBsonOutputLength);
+            ArrayList<BsonElement> extraElements = getExtraElements(operationContext);
+            // `DualMessageSequences` requires validation only if no response is expected, otherwise we must rely on the server validation
+            boolean validateDocumentSizeLimits = !(sequences instanceof DualMessageSequences) || !responseExpected;
```
stIncMale, line 235: This exception makes the new `bulkWrite` command behavior regarding BSON size validation inconsistent with all other commands. I am trying to figure this out in https://mongodb.slack.com/archives/C035ZJL6CQN/p1730774917892009.

jyemin (commented):
Latest commit LGTM.

Not sure if any more are coming in this PR so not approving the PR yet.

driver-core/src/test/unit/com/mongodb/internal/connection/CommandMessageSpecification.groovy
stIncMale: Done in 6f68c0e3fe7d1591d43583c78244c5384e3640b0.

stIncMale:
@vbabanin This PR is ready for your review.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
stIncMale: This is resolved by https://jira.mongodb.org/browse/JAVA-5695, which is part of the current PR.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -807,18 +652,233 @@ void onBulkWriteCommandErrorWithoutResponse(final MongoException exception) {
...
+            /**
+             * The server supports only the {@code update} individual write operation in the {@code ops} array field, while the driver supports
+             * {@link ClientNamespacedUpdateOneModel}, {@link ClientNamespacedUpdateOneModel}, {@link ClientNamespacedReplaceOneModel}.
+             * The difference between updating and replacing is only in the document specified via the {@code updateMods} field:
+             * <ul>
+             *     <li>if the name of the first field starts with {@code '$'}, then the document is interpreted as specifying update operators;</li>
+             *     <li>if the name of the first field does not start with {@code '$'}, then the document is interpreted as a replacement.</li>
+             * </ul>
+             */
+            private static final class OpsFieldNameValidator implements FieldNameValidator {
```
vbabanin, line 753: As discussed on the call, let's add a link to the [spec requirement](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#update-vs-replace-document-validation) for clarity.

driver-core/src/main/com/mongodb/internal/connection/BsonWriterHelper.java
```diff
@@ -65,16 +96,86 @@ static void writePayload(final BsonWriter writer, final BsonOutput bsonOutput, f
...
+    static EncodeDocumentsResult writeDocumentsOfDualMessageSequences(
+            final DualMessageSequences dualMessageSequences,
+            final int commandDocumentSizeInBytes,
+            final BsonOutput firstOutput,
+            final BsonOutput secondOutput,
+            final MessageSettings messageSettings) {
+        BsonBinaryWriter firstWriter = createBsonBinaryWriter(firstOutput, dualMessageSequences.getFirstFieldNameValidator(), null);
+        BsonBinaryWriter secondWriter = createBsonBinaryWriter(secondOutput, dualMessageSequences.getSecondFieldNameValidator(), null);
+        // the size of operation-agnostic command fields (a.k.a. extra elements) is counted towards `messageOverheadInBytes`
+        int messageOverheadInBytes = 1000;
```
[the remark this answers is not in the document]
stIncMale: I am not sure if limitation is the right word here. But yes, this is unique to client bulk writes:

1. A mixed bulk write does not mix different kinds of operations in a batch because it is technically impossible there. As a result, we know in advance what kind of operations is in a batch, thus knowing whether it supports retries or not. That allows us to encode `extraElements` before encoding the `PAYLOAD_TYPE_1_DOCUMENT_SEQUENCE` section. Thus, when we are encoding the document sequence, we know exactly how much space we have left available before reaching `MessageSettings.getMaxMessageSize`.
2. A client bulk write may mix different kinds of operations in a batch, which means that we know whether the batch supports retries only after encoding its `PAYLOAD_TYPE_1_DOCUMENT_SEQUENCE` sections. That is, we may need to write something after writing those sections. That, in turn, means we can't know exactly how much space we have left when we encode document sequences. But whatever we write after writing the sequences, its size is bounded, and [1000 bytes is used in the spec as the value that is definitely not smaller than that bound](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#total-message-size).

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -341,6 +383,26 @@ private void addReadConcernDocument(final List<BsonElement> extraElements, final
...
+     * @param sequenceId The identifier of the sequence contained in the {@code OP_MSG} section to be written.
+     * @param writeDocumentsAction The action that writes the documents of the sequence.
+     * @see <a href="https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.md">OP_MSG</a>
+     */
+    private <R> R writeOpMsgSectionWithPayloadType1(
+            final ByteBufferBsonOutput bsonOutput,
+            final String sequenceId,
+            final Supplier<R> writeDocumentsAction) {
+        // payload type
+        bsonOutput.writeByte(1);
```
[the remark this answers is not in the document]
stIncMale: Done in 38c880d988c2c7f9d4a8716b4155ef36f43de77b.

driver-core/src/main/com/mongodb/internal/connection/BsonWriterHelper.java
```diff
@@ -65,16 +96,86 @@ static void writePayload(final BsonWriter writer, final BsonOutput bsonOutput, f
...
+        int maxSizeInBytes = messageSettings.getMaxMessageSize() - (messageOverheadInBytes + commandDocumentSizeInBytes);
+        int firstStart = firstOutput.getPosition();
+        int secondStart = secondOutput.getPosition();
+        int maxBatchCount = messageSettings.getMaxBatchCount();
+        return dualMessageSequences.encodeDocuments(write -> {
```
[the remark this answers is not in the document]
stIncMale: Done in 38c880d988c2c7f9d4a8716b4155ef36f43de77b.

driver-core/src/main/com/mongodb/internal/connection/BsonWriterHelper.java
```diff
@@ -16,32 +16,63 @@
...
+        bsonOutputWithDocument.truncateToPosition(appendFrom);
+        if (bsonElements != null) {
+            for (BsonElement element : bsonElements) {
+                String name = element.getName();
+                BsonValue value = element.getValue();
+                writer.writeName(name);
+                encodeUsingRegistry(writer, value);
+            }
         }
+        // write the BSON document ending
```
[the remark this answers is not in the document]
stIncMale: Done in 38c880d988c2c7f9d4a8716b4155ef36f43de77b.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -942,18 +1002,19 @@ void encodeWriteModel(
...
+                final int modelIndexInBatch) {
             writer.writeName("document");
             Object document = model.getDocument();
-            @SuppressWarnings("unchecked")
-            Encoder<Object> documentEncoder = (Encoder<Object>) codecRegistry.get(document.getClass());
             assertNotNull(encodedBatchInfo).insertModelDocumentIds.compute(modelIndexInBatch, (k, knownModelDocumentId) -> {
                 IdHoldingBsonWriter documentIdHoldingBsonWriter = new IdHoldingBsonWriter(
                         writer,
                         // Reuse `knownModelDocumentId` if it may have been generated by `IdHoldingBsonWriter` in a previous attempt.
                         // If its type is not `BsonObjectId`, we know it could not have been generated.
```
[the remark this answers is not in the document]
stIncMale: Done in 38c880d988c2c7f9d4a8716b4155ef36f43de77b.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -265,11 +307,11 @@ private boolean useOpMsg() {
         return getOpCode().equals(OpCode.OP_MSG);
     }
 
-    private List<BsonElement> getExtraElements(final OperationContext operationContext) {
```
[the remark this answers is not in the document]
stIncMale: There is no dependency specifically on the `ArrayList` type. Replaced with `List`.

Done in 38c880d988c2c7f9d4a8716b4155ef36f43de77b.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -293,7 +335,7 @@ private List<BsonElement> getExtraElements(final OperationContext operationConte
         assertFalse(sessionContext.hasActiveTransaction() && sessionContext.isSnapshot());
         if (sessionContext.hasActiveTransaction()) {
             checkServerVersionForTransactionSupport();
-            extraElements.add(new BsonElement("txnNumber", new BsonInt64(sessionContext.getTransactionNumber())));
+            extraElements.add(new BsonElement(TXN_NUMBER_KEY, new BsonInt64(sessionContext.getTransactionNumber())));
```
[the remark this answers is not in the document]
stIncMale: There was probably such an intention, but nothing like that happened. Removed the constant.

Done in 38c880d988c2c7f9d4a8716b4155ef36f43de77b.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
stIncMale: Done in 38c880d988c2c7f9d4a8716b4155ef36f43de77b.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -201,27 +222,47 @@ MongoNamespace getNamespace() {
     }
 
     @Override
-    protected EncodingMetadata encodeMessageBodyWithMetadata(final BsonOutput bsonOutput, final OperationContext operationContext) {
+    protected EncodingMetadata encodeMessageBodyWithMetadata(final ByteBufferBsonOutput bsonOutput, final OperationContext operationContext) {
```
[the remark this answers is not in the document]
stIncMale: Done in 015613505607cccee42d7626db9acd17177ef455.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -201,27 +222,47 @@ MongoNamespace getNamespace() {
...
+                try (ByteBufferBsonOutput.Branch bsonOutputBranch2 = bsonOutput.branch();
+                     ByteBufferBsonOutput.Branch bsonOutputBranch1 = bsonOutput.branch()) {
+                    DualMessageSequences.EncodeDocumentsResult encodeDocumentsResult = writeOpMsgSectionWithPayloadType1(
+                            bsonOutputBranch1, dualMessageSequences.getFirstSequenceId(), () ->
+                                    writeOpMsgSectionWithPayloadType1(bsonOutputBranch2, dualMessageSequences.getSecondSequenceId(), () ->
+                                            writeDocumentsOfDualMessageSequences(
+                                                    dualMessageSequences, commandDocumentSizeInBytes, bsonOutputBranch1,
+                                                    bsonOutputBranch2, getSettings())
+                                    )
+                    );
```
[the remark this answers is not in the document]
stIncMale: This is a great improvement.

Done in f07ff6f4c2e664890286797290a3be4ceaefc48a.

vbabanin (approved):
LGTM!

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
stIncMale: Moved to https://github.com/mongodb/mongo-java-driver/pull/1509/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#r1861036281.

PR #1500, opened 2024-09-20
JAVA-5584: Move mongodb-crypt classes to internal package

jyemin:
* Remove dependency in MongoCryptException on non-public class
* Move mongodb-crypt classes to internal package

JAVA-5584


mongodb-crypt/src/main/com/mongodb/crypt/capi/MongoCryptException.java
```diff
@@ -47,17 +53,6 @@ public MongoCryptException(final String msg, final Throwable cause) {
         this.code = -1;
     }
 
-    /**
-     * Construct an instance from a {@code mongocrypt_status_t}.
-     *
-     * @param status the status
-     */
-    MongoCryptException(final mongocrypt_status_t status) {
```
jyemin, line 55: Note this is not a breaking change because it's not public.

rozza (approved):
LGTM

PR #1506, opened 2024-09-23
Fix OSGi `:mongodb-crypt` manifest entries

stIncMale:


driver-core/build.gradle
```diff
@@ -89,7 +89,7 @@ afterEvaluate {
             'com.github.luben.zstd.*;resolution:=optional',
             'org.slf4j.*;resolution:=optional',
             'jnr.unixsocket.*;resolution:=optional',
-            'com.mongodb.crypt.capi.*;resolution:=optional',
```
jyemin, line 92: driver-core doesn't reference `MongoCryptException` (the only class left in this package), so this needed to be removed or else we get a warning.

driver-reactive-streams/build.gradle
```diff
@@ -75,6 +75,7 @@ afterEvaluate {
     jar.manifest.attributes['Bundle-SymbolicName'] = 'org.mongodb.driver-reactivestreams'
     jar.manifest.attributes['Import-Package'] = [
             'com.mongodb.crypt.capi.*;resolution:=optional',
+            'com.mongodb.internal.crypt.capi.*;resolution:=optional',
```
jyemin, line 78: reactive-streams (and sync) reference `MongoCryptException`, and also the classes that were moved into the internal package, so both packages are now needed.

mongodb-crypt/build.gradle.kts
```diff
@@ -175,7 +175,7 @@ afterEvaluate {
     tasks.jar {
         manifest {
             attributes(
-                "-exportcontents" to "com.mongodb.crypt.capi.*;-noimport:=true",
+                "-exportcontents" to "com.mongodb.*;-noimport:=true",
```
jyemin, line 178: This is a weird one.  driver-core uses `*` for this, but `*` causes a warning with mongodb-crypt because the jar file contains directories for each of the shared library files, and some of those (e.g. `linux-aarch64`) are not valid Java package names.  Using `com.mongodb.*` does the trick (it could also just be `com.*`).

jyemin (approved):
Left a few comments for other reviewers.

vbabanin (approved):
LGTM!

PR #1510, opened 2024-09-24
Revert making BsonEncoder / BsonDecoder internal

rozza:
Made BsonEncoder / Decoder internal as part of the JsonElement support, however, this reduces the flexibility of the API and that change should be reverted

JAVA-5623

### DIFF
--- a/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
+++ b/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
@@ -51,10 +51,10 @@ import org.bson.types.ObjectId
  * For custom serialization handlers
  */
 @ExperimentalSerializationApi
-internal sealed interface BsonDecoder : Decoder, CompositeDecoder {
+public sealed interface BsonDecoder : Decoder, CompositeDecoder {
 
     /** Factory helper for creating concrete BsonDecoder implementations */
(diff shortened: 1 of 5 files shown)


rozza:
In 7e4661c I made the `BsonEncoder` / `BsonDecoder` interfaces `internal` which wasn't a good idea. This PR reverts that by making them public again.

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonDecoder.kt
```diff
@@ -66,7 +66,7 @@ internal sealed interface BsonDecoder : Decoder, CompositeDecoder {
             }
         }
 
-        fun createBsonDecoder(
+        internal fun createBsonDecoder(
```
nhachicha, line 69: Shouldn't this be `public` since [createBsonEncoder](https://github.com/mongodb/mongo-java-driver/pull/1510/files#diff-6ca11d6fb183c8e6441055f9fe4a6a804ae6676b009f6b6ba567b3d5fc757cbbR56) is public ?

rozza: Thanks good catch - I opted to add an internal `BsonEncoderDecoderUtils` object. So that the public interfaces are as simple as can be and don't leak a useless companion object (from outside the library).

PR #1515, opened 2024-09-26
Adding Kotlin extensions methods for projection.

nhachicha:
Fixes JAVA-5603.
Based on kmongo-property with more tests and a bug fix in the path extension method throwing an NPE for calculated properties without a Kotlin backing field.

### DIFF
--- a/driver-kotlin-extensions/src/main/kotlin/com/mongodb/kotlin/client/model/Projections.kt
+++ b/driver-kotlin-extensions/src/main/kotlin/com/mongodb/kotlin/client/model/Projections.kt
@@ -1,0 +1,257 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ * Copyright (C) 2016/2022 Litote
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @custom-license-header
+ */
+package com.mongodb.kotlin.client.model
+
+import com.mongodb.annotations.Beta
+import com.mongodb.annotations.Reason
+import com.mongodb.client.model.Aggregates
+import com.mongodb.client.model.Projections
+import org.bson.conversions.Bson
+import kotlin.reflect.KProperty
+import kotlin.reflect.KProperty1
+
+/**
+ * The projection of the property.
+ * This is used in an aggregation pipeline to reference a property from a path.
+ */
+public val <T> KProperty<T>.projection: String get() = path().projection
+
+/**
+ * The projection of the property.
+ */
+public val String.projection: String get() = "\$$this"
+
+/**
+ * In order to write `$p.p2`
+ */
+
+@JvmSynthetic
+public infix fun <T0, T1> KProperty1<T0, T1?>.projectionWith(p2: String): String = "$projection.$p2"
+
+/**
+ * Creates a projection of a property whose value is computed from the given expression. Projection with an expression can be used in the
+ * following contexts:
+ * <ul>
+ *   <li>$project aggregation pipeline stage.</li>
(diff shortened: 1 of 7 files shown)


driver-kotlin-extensions/src/main/kotlin/com/mongodb/kotlin/client/model/Properties.kt
```diff
@@ -84,12 +83,19 @@ internal fun <T> KProperty<T>.path(): String {
             // If no path (serialName) then check for BsonId / BsonProperty
             if (path == null) {
                 val originator = if (this is CustomProperty<*, *>) this.previous.property else this
-                val constructorProperty =
-                    originator.javaField!!.declaringClass.kotlin.primaryConstructor?.findParameterByName(this.name)
-
-                // Prefer BsonId annotation over BsonProperty
-                path = constructorProperty?.annotations?.filterIsInstance<BsonId>()?.firstOrNull()?.let { "_id" }
-                path = path ?: constructorProperty?.annotations?.filterIsInstance<BsonProperty>()?.firstOrNull()?.value
+                // If this property is calculated (doesn't have a backing field) ex
```
rozza, line 86: Good catch 👍

driver-kotlin-extensions/src/test/kotlin/com/mongodb/kotlin/client/model/ProjectionTest.kt
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    @Test
+    fun firstElem() {
+        assertEquals(""" {"name.${'$'}" : 1} """, Person::name.elemMatchProj)
+    }
+
+    @Test
+    fun elemMatch() {
+        // Given the following document:
```
rozza, line 72: I get why you added this, but for me it reduces the readibility of the test cases. Especially as we're not testing the server and how it does projection. Also, this wasn't done for the Filters so that also adds the cognitive load.

driver-kotlin-extensions/src/main/kotlin/com/mongodb/kotlin/client/model/Projections.kt
```diff
@@ -0,0 +1,236 @@
...
+ * </li> </ul>
+ *
+ * @param expression the expression
+ * @param <T> the expression type
+ * @return the projection
+ * @see #computedSearchMeta(String)
+ * @see Aggregates#project(Bson)
+ */
+@JvmSynthetic
+public infix fun <T> KProperty<T>.computedFrom(expression: Any): Bson =
```
rozza, line 62: In Filters we added both extension methods (where possible) as well as simple methods eg: `computedFrom(property: KProperty<T>, expression: Any)`.  Although, this adds two ways of doing things, it does mean that users can read the driver docs and easily port the Java examples.

We should also add them here, to keep this inline with the `Filters` class.

rozza (changes requested):
Looking good - couple of recommended changes.

driver-kotlin-extensions/src/main/kotlin/com/mongodb/kotlin/client/model/Projections.kt
```diff
@@ -0,0 +1,313 @@
...
+public fun fields(vararg projections: Bson): Bson = Projections.fields(*projections)
+
+/**
+ * Creates a projection that combines the list of projections into a single one. If there are duplicate keys, the last
+ * one takes precedence.
+ *
+ * @param projections the list of projections to combine
+ * @return the combined projection @mongodb.driver.manual
+ */
+public fun fields(projections: List<Bson>): Bson = Projections.fields(projections)
```
rozza, line 313: This method needs a test

driver-kotlin-extensions/src/main/kotlin/com/mongodb/kotlin/client/model/Projections.kt
```diff
@@ -0,0 +1,313 @@
...
+ *
+ * </li> </ul>
+ *
+ * @param property the data class property
+ * @param expression the expression
+ * @return the projection
+ * @see #computedSearchMeta(String)
+ * @see Aggregates#project(Bson)
+ */
+public fun computedFrom(property: String, expression: Any): Bson = property.computedFrom(expression)
```
rozza, line 136: This method needs a test

rozza (changes requested):
Looking good. 

Remember to add a test to `ExtensionsApiTest` to ensure we capture all Projections methods and have a regression test incase future ones are added.

rozza (approved):
I fixed some `@since` nits and pushed to your repo @nhachicha as they were only minor.

LGTM!

PR #1525, opened 2024-10-04
Allow valid SRV hostnames with less than 3 parts

NathanQingyangXu:
https://jira.mongodb.org/browse/JAVA-5542


driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
```diff
@@ -83,6 +89,12 @@ public List<String> resolveHostFromSrvRecords(final String srvHost, final String
             for (String srvRecord : srvAttributeValues) {
                 String[] split = srvRecord.split(" ");
                 String resolvedHost = split[3].endsWith(".") ? split[3].substring(0, split[3].length() - 1) : split[3];
+                if (isShortSrvHost && resolvedHost.equals(srvHost)) {
+                    throw new MongoConfigurationException(
+                            format("The SRV host name '%s' has less than three parts and the resolved host '%s' is identical.",
```
katcharov, line 94: Does this (user-facing) error message discuss implementation details, for example the number of parts? The [spec](https://github.com/mongodb/specifications/commit/8e9d768dc2840eafdb9e2cb4788c608247c56a96#diff-6de763a7f8f4413bfd2a70cdfe5c2237e9ced55b9d14fb8e1714152a54957d44R43) potentially suggests that this would be described as "...host name ... lacks a subdomain, and the resolved..."? I tried to compare against the other drivers, and I believe this corresponds to the following:

Node: https://github.com/mongodb/node-mongodb-native/pull/4197/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R1181-R1185
Rust: https://github.com/mongodb/mongo-rust-driver/pull/1211/files#diff-32a39d356e1edde3d94581275b42e994ad1780ce4de8d6935ec5596fe4493ffdR39

However, I am not sure, because the error messages are different. It would have been useful if the spec specified the wording of the error messages.

driver-core/src/test/unit/com/mongodb/internal/connection/InitialDnsSeedListDiscoveryProseTest.java
```diff
@@ -0,0 +1,147 @@
...
+    @Mock
+    private DnsSrvRecordMonitorFactory dnsSrvRecordMonitorFactory;
+
+    private DnsMultiServerCluster cluster;
+
+    @BeforeEach
+    void setUp() throws Exception {
+        try (AutoCloseable ignored = openMocks(this)) {
+            when(serverFactory.getSettings()).thenReturn(ServerSettings.builder().build());
+            when(serverFactory.create(any(Cluster.class), any(ServerAddress.class))).thenReturn(mock(ClusterableServer.class));
```
katcharov, line 61: I have not seen `openMocks` before - is there a benefit to using this here, rather than using `Mockito.mock(` in `doTest`?

driver-core/src/test/unit/com/mongodb/internal/connection/InitialDnsSeedListDiscoveryProseTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+        cluster = new DnsMultiServerCluster(clusterId, settingsBuilder.build(),
+                serverFactory,
+                dnsSrvRecordMonitorFactory);
+        try {
+            Thread.sleep(100); // racy
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            // ignore
+        }
```
katcharov, line 137: `ClusterFixture.sleep(100);`?

driver-core/src/test/unit/com/mongodb/internal/connection/InitialDnsSeedListDiscoveryProseTest.java
```diff
@@ -0,0 +1,147 @@
...
+        doTest(srvHost, resolvedHost, true);
+    }
+
+    private void doTest(final String srvHost, final String resolvedHost, final boolean throwException) {
+        final ClusterId clusterId = new ClusterId();
+
+        final DnsResolver dnsResolver = new DefaultDnsResolver((name, type) -> singletonList(String.format("10 5 27017 %s",
+                resolvedHost)));
+
+        when(dnsSrvRecordMonitorFactory.create(eq(srvHost), eq(SRV_SERVICE_NAME), any(DnsSrvRecordInitializer.class))).thenAnswer(
```
katcharov, line 120: Perhaps linebreak before `.thenAnswer` to keep lines under 120 chars?

katcharov (changes requested):
Looking good, just a few comments/questions.

driver-core/src/test/unit/com/mongodb/internal/connection/InitialDnsSeedListDiscoveryProseTest.java
NathanQingyangXu: It is the preferred counterpart (see https://javadoc.io/doc/org.mockito/mockito-core/3.4.0/org/mockito/MockitoAnnotations.html#openMocks-java.lang.Object-) of the previous common mockito usage as below:

```java
import org.mockito.MockitoAnnotations;

MockitoAnnotations.initMocks(this);
```

A better alternative is definitively the following test class level annotation:
```java

@ExtendWith(MockitoExtension.class)
```

but its usage currently requires our gradle dependency change.
How do you think which way to persue:
1) use `MockitoAnnotations.initMocks(this)`
2) add MockitoExtension related testing scope dependency and use `@ExtendWith(MockitoExtension.class)`
3) current status quo, which avoids deprecation warning but ends up with lots of confusion

driver-core/src/test/unit/com/mongodb/internal/connection/InitialDnsSeedListDiscoveryProseTest.java
NathanQingyangXu: the coding pattern in this class was mainly copied from another testing class, i.e. com.mongodb.internal.connection.SrvPollingProseTests.

I can use ClusterFixture.sleep(100) for sure.

but should we add `Thread.currentThread().interrupt()` in its code?

driver-core/src/test/unit/com/mongodb/internal/connection/InitialDnsSeedListDiscoveryProseTest.java
NathanQingyangXu: from .editorConfig file, our max line length is 140?

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
```diff
@@ -64,13 +64,19 @@ public DefaultDnsResolver(@Nullable final DnsClient dnsClient) {
...
     */
     @Override
     public List<String> resolveHostFromSrvRecords(final String srvHost, final String srvServiceName) {
-        String srvHostDomain = srvHost.substring(srvHost.indexOf('.') + 1);
+        List<String> srvHostParts = asList(srvHost.split("\\."));
+
+        String srvHostDomain;
+        boolean isShortSrvHost = srvHostParts.size() < 3;
+        if (isShortSrvHost) {
+            srvHostDomain = srvHost; // when dot separated parts less than 3, domain name is the host per se
```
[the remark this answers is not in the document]
NathanQingyangXu: totally agree. I chose `srvHostHasLessThanThreeParts`

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
NathanQingyangXu: This exception is mainly for this prose testing case: https://github.com/mongodb/specifications/commit/8e9d768dc2840eafdb9e2cb4788c608247c56a96#diff-dfcf948e6a987b880ea85e1907d5b7ce8fb98c9dab752f42ccdaf1e656b8848fR30

NathanQingyangXu: thanks for referring to other driver's implementation. will investigate further.

driver-core/src/test/unit/com/mongodb/internal/connection/InitialDnsSeedListDiscoveryProseTest.java
katcharov: By "current status quo" in point 3, are you referring to `Mockito.mock(`? Could you elaborate on what the confusion is?

NathanQingyangXu: It means the current approach as below:
```
try (AutoCloseable ignored = openMocks(this)) {
    // mock behaviour
}
```

It is a programmatic way to ensure each field annotation with `@Mock` has been gone through `Mockito.mock()` so we could focus on mocking their behavour in the try block.

All the three approaches are for the same purposes so we don't need to invoke `Mockito.mock()` manually one by one (which is tedious, less readable and error-prone).

NathanQingyangXu: As we discussed, I reverted back Mock annotation usage to manually mock inside method.

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
NathanQingyangXu: as per [the following code snippet](https://github.com/mongodb/mongo-java-driver/blob/main/driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java#L96), the exception message won't be exposed to end user but end up with a generic exception: 
```
throw new MongoConfigurationException(format("Failed looking up SRV record for '%s'.", resourceName), e);
```

NathanQingyangXu: I changed the logic to accommodate the possibility that 'resolvedHost' could be something like `localhost` (as in Prosetest case #3) which doesn't contain '.' separator and so it has no `parent domain`. After the code refactoring, no new exception or logic branch is required.

katcharov: Were the prose tests passing previously? I thought you were testing against `localhost`.

> exception message won't be exposed to end user

I believe all exception messages are exposed to the user. It may be in "caused by", but this is still visible.

> no new exception or logic branch is required

I am not sure I follow the logic well enough to understand this - why is it that the other drivers have introduced 1 or 2 new messages, such as "Server record does not have at least one more domain level than parent URI", but we do not need to?

NathanQingyangXu: yeah, you are totally correct in that the exception will be exposed to end user as exception's cause. I missed that. Now after new code change (converging the newly -added logic branch to the existing only flow), we reuse the old same exception mesage as below:
```
throw new MongoConfigurationException(
                            format("The SRV host name '%s' resolved to a host '%s 'that is not in a sub-domain of the SRV host.",
                                    srvHost, resolvedHost));
```
It is simply another way to express the same meaning, from my understanding. Maybe it is desirable to maintain backward compatibility so I didn't change it.

NathanQingyangXu: I talked with Aditi or the node developer and clarified lots of doubts. I also changed some verbiages which is out of sync with the new terminology in the spec. Now it is ready for a new review. As in node PR, two exceptions are created in similar manner.

driver-core/src/main/com/mongodb/internal/dns/DefaultDnsResolver.java
```diff
@@ -64,13 +64,19 @@ public DefaultDnsResolver(@Nullable final DnsClient dnsClient) {
 
       The priority and weight are ignored, and we just concatenate the host (after removing the ending '.') and port with a
       ':' in between, as expected by ServerAddress.
-
-      It's required that the srvHost has at least three parts (e.g. foo.bar.baz) and that all of the resolved hosts have a parent
-      domain equal to the domain of the srvHost.
```
katcharov, line 69: Optional since this is internal, but consider replacing this with the revised conditions.

katcharov (approved):
LGTM!

PR #1528, opened 2024-10-10
Add BSON Binary Subtype 9 support for vector storage and retrieval.

vbabanin:
## Description
This PR introduces the implementation of [BSON Binary Subtype 9 - Vector](https://bsonspec.org/spec.html), based on the proposed specification for efficiently storing and retrieving densely packed arrays of numbers. 

#### Key Changes:
BSON Binary subtype: Implements a new BSON binary subtype (value "\x09") for vectors to better manage dense numeric arrays.

API enhancements: 
Updates include idiomatic APIs that facilitate the translation between arrays of numbers and their BSON binary representations.

Data type support: Supports various data types (dtypes), including `INT8`, `FLOAT32`, and `PACKED_BIT`.

#### Implementation details
Encoding and decoding: Adds methods to encode and decode vectors according to the specified dtype and padding.

## Draft status and review:
The specification this PR depends on is under review. You can review the specification here: [BSON Binary Subtype 9 Specification PR](https://github.com/mongodb/specifications/pull/1658). Substantial changes to the specification are not expected, therefore your feedback on this preliminary implementation would be valuable.

JAVA-5544


bson/src/test/unit/org/bson/vector/VectorGenericBsonTest.java
```diff
@@ -0,0 +1,265 @@
...
+                        Vector.packedBitVector(expectedVectorPackedBitData, expectedPadding),
+                        testKey,
+                        actualDecodedDocument,
+                        expectedCanonicalBsonHex,
+                        description);
+                break;
+            case FLOAT32:
+                float[] actualFloatVector = actualVector.asFloatVectorData();
+                float[] expectedFloatVector = toFloatArray(arrayVector);
+//                assertVectorDecoding(
```
jyemin, line 154: I was just reviewing the specification, and noticed the line

> For floating point number types, numerical values need not match exactly.

So I looked at your implementation and saw that this assertion is commented out.  But when I uncomment it the tests all still pass. So I'm curious why you have it commented out. Do you think anything needs to change about that line in the spec?  Does the spec need to define more precisely what "not match exactly" means in practice.

bson/src/test/unit/org/bson/vector/VectorGenericBsonTest.java
```diff
@@ -0,0 +1,265 @@
...
+        assertTrue(runtimeException instanceof IllegalArgumentException || runtimeException instanceof IllegalStateException);
+    }
+
+    private static void assertThatVectorCreationResultsInCorrectBinary(final Vector expectedVectorData,
+                                                                       final String testKey,
+                                                                       final BsonDocument actualDecodedDocument,
+                                                                       final String expectedCanonicalBsonHex,
+                                                                       final String description) {
+        BsonDocument documentToEncode = new BsonDocument(testKey, new BsonBinary(expectedVectorData));
+        assertEquals(documentToEncode, actualDecodedDocument);
```
jyemin, line 180: Does the specification include all the required assertions that you think are actually necessary?  I see eleven assertions in this test, but only four required assertions in the README.

bson/src/main/org/bson/Vector.java
```diff
@@ -0,0 +1,268 @@
...
+        return result;
+    }
+
+    /**
+     * Represents the data type (dtype) of a vector.
+     * <p>
+     * Each dtype determines how the data in the vector is stored, including how many bits are used to represent each element
+     * in the vector.
+     */
+    public enum Dtype {
```
jyemin, line 234: How about just call it "Type".  The "D" isn't going to help anyone, and might confuse people.  Also, we generally avoid abbreviations unless they are well-known terms of art, e.g. "html"

bson/src/main/org/bson/Vector.java
```diff
@@ -0,0 +1,268 @@
...
+     */
+    public static Vector floatVector(final float[] vectorData) {
+        notNull("vectorData", vectorData);
+        return new Vector(vectorData);
+    }
+
+    /**
+     * Returns the {@link Dtype#PACKED_BIT} vector data as a byte array.
+     *
+     * <p> This method is used to retrieve the underlying underlying byte array representing the {@link Dtype#PACKED_BIT} vector, where
```
jyemin, line 123:
```suggestion
     * <p> This method is used to retrieve the underlying byte array representing the {@link Dtype#PACKED_BIT} vector, where
```

bson/src/main/org/bson/Vector.java
```diff
@@ -0,0 +1,268 @@
...
+ * vectors and methods to retrieve their underlying data.
+ * <p>
+ * Vectors are densely packed arrays of numbers, all the same type, which are stored efficiently
+ * in BSON using a binary format.
+ *
+ * @mongodb.server.release 6.0
+ * @see BsonBinary
+ * @since BINARY_VECTOR
+ */
+public final class Vector {
```
jyemin, line 40: Instead of having unused fields depending on the type of the vector, should we instead 
have private subclasses of Vector, one for each type, that has only the fields needed for that type?

Another option would be to make the subclasses public, and put the "as" conversion methods on the subclasses.  This would be more similar to the design of the `BsonValue` type hierarchy.

bson/src/main/org/bson/Vector.java
```diff
@@ -0,0 +1,268 @@
...
+     *
+     * <p> This method is used to retrieve the underlying underlying byte array representing the {@link Dtype#PACKED_BIT} vector, where
+     * each bit represents an element of the vector (either 0 or 1).
+     *
+     * @return the packed bit vector data.
+     * @throws IllegalStateException if this vector is not of type {@link Dtype#PACKED_BIT}. Use {@link #getDataType()} to check the vector type before
+     *                               calling this method.
+     * @see #getPadding() getPadding() specifies how many least-significant bits in the final byte should be ignored.
+     */
+    public byte[] asPackedBitVectorData() {
```
jyemin, line 131: Let's bikeshed on the names of these methods a bit.  Generally I don't like including generic words like "data" in names, because everything is data.  What about something like "asPackedBitArray"? That would let us add new methods that return more convenient types, like "BitSet asBitSet()"

bson/src/main/org/bson/Vector.java
jyemin, line 40: Let's bikeshed on the name and consider more description options like

* `BinaryVector`
* `BsonBinaryVector`

This corresponds a bit better to the ` BsonValue` type hierarchy, which has the `Bson` prefix for all the types.

bson/src/main/org/bson/BsonBinary.java
```diff
@@ -89,6 +92,20 @@ public BsonBinary(final UUID uuid) {
         this(uuid, UuidRepresentation.STANDARD);
     }
 
+    /**
```
jyemin, line 95: I'm wondering if we should add similar support for vectors to `org.bson.types.Binary`, since that's the class more typically used with `org.bson.Document`.

I know we didn't do that for UUIDs, but can't remember the rationale.

What do you think?

bson/src/main/org/bson/Vector.java
jyemin, line 40: Consider adding a codec for Vector, so that they can be included directly in a POJO or record without having to manually wraps it in `BsonBinary`.

jyemin (changes requested):
Just giving this an initial review as part of my review of the spec.

bson/src/test/unit/org/bson/vector/VectorGenericBsonTest.java
vbabanin: Thank you for bringing this up! Commenting it out was an oversight on my part. I've now uncommented it.

As regards to the spec line:
When starting with a `float` in Java (which is already stored as an approximation, like 127.699..... for 127.7), converting it to a byte array of 4 bytes and back to float should not introduce any additional approximation error. The round-trip preserves the same approximated value because we're directly encoding and decoding the binary representation of the float. In this case, the values should match exactly, as no further approximation should occur.  So precise matching seems important for drivers that support float32 to catch any mistakes in the `float to bytes & bytes to float` algorithm itself. Perhaps this is something that could be clarified in the spec as well.

The "not match exactly" clause could benefit from specifying a small margin of error for comparing floating-point values. This would likely be helpful for drivers where 64-bit floats (e.g., in Python or JavaScript) are converted to FLOAT32, where some precision loss can happen. Otherwise, the term "not match exactly" might feel a bit vague—large differences like 4.1 and 4.9 would clearly point to a problem in the conversion.
 
So the spec could be more precise in explaining when and why exact matching might not be expected.

bson/src/test/unit/org/bson/vector/VectorGenericBsonTest.java
vbabanin: Yes, it should be sufficient. I removed some duplicated assertions in our tests to avoid redundancy.

bson/src/main/org/bson/Vector.java
vbabanin: After reconsidering, creating a subclass for each type leads to a more user-friendly API. Initially, I was concerned about class proliferation, given the number of potential vector types in the future:

```
INT4, INT8, INT16, INT32, INT64
UINT8, UINT16, UINT32, UINT64
FLOAT16, FLOAT32, FLOAT64
BOOLEAN
PACKED_BIT
DECIMAL128, DECIMAL32, DECIMAL64
COMPLEX64, COMPLEX128
```

However, having a subclass per type helps keep the logic isolated where it belongs. For example, padding isn’t applicable to float32/64 vectors or Int8/16/32/64 vectors, so having particular subclass without padding makes the API more user-friendly and focused. This is especially useful when using `PojoCodec`, where users can directly specify `Float32Vector` or `PackedBitVector`, improving clarity and usability, similar to how we handle public subclasses in `BsonValue`.

I've updated the implementation to use subclasses. Please, let me know your thoughts.

bson/src/main/org/bson/Vector.java
vbabanin: Yes, I agree. With the introduction of subclasses, I've renamed the methods to more specific names like `asFloat32Vector`, `asInt8Vector`, etc. This should make the method names more intuitive and aligned with the vector types.   

Now `asBitSet()` could be added on `PackedBitVector`.

bson/src/main/org/bson/BsonBinary.java
vbabanin: Yes, we should, as it would otherwise limit users who work with the `org.bson.Document` class. I've added the support for vectors in `org.bson.types.Binary.` Thank you!

bson/src/main/org/bson/Vector.java
vbabanin: Thanks for the suggestion! I've added codecs for each vector subclass, along with the corresponding tests.

bson/src/main/org/bson/Vector.java
vbabanin: Good point. 'DType' originates from NumPy, but I agree we should avoid abbreviations unless they're well-known. Many libraries and frameworks (for example TensorFlow and DL4J) use 'DataType' [DL4J DataType](https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/buffer/DataType.java)). Using 'DataType' for the enum could maintain consistency and make the terminology clearer.

I'm open to naming it `Type `as well. Would `DataType` work, or do you prefer `Type`?

jyemin: DataType sgtm

bson/src/main/org/bson/Vector.java
jyemin: I like it.

bson/src/main/org/bson/Vector.java
vbabanin: Renamed. Thanks!

bson/src/main/org/bson/Vector.java
vbabanin: I understand the idea of aligning the name with the `BsonValue` hierarchy, however since` Vector` isn't a direct part of that hierarchy (similar to how UUID isn't), I believe it would be more consistent to avoid the `Bson` prefix.

Regarding `BinaryVector`, the term "binary vector" seems to be commonly used in machine learning to refer to binary quantized vectors, like our `PackedBitVector`, which could lead to confusion. For reference, here’s an example: [Binary Vector in Machine Learning](https://machinelearningcatalogue.com/input_data_type/idt_binary-vector.html#:~:text=A%20binary%20vector%20is%20a,vectors%20and%20vectors%20of%20variables).

I personally think `Vector` works, but there is the potential collision with the `Vector collection class`, and IDEs often auto-import it, which could be problematic. I think we should brainstorm further on the naming. Perhaps `NumericVector` could be an option, but it’s worth giving it more thought.

jyemin: If it wasn't for `java.util.Vector` I'd probably let this go.  But let's get a second opinion before resolving.

bson/src/test/unit/org/bson/codecs/VectorCodecTest.java
```diff
@@ -0,0 +1,162 @@
...
+        when(mockReader.peekBinarySubType()).thenReturn(BsonBinarySubType.VECTOR.getValue());
+        when(mockReader.readBinaryData()).thenReturn(bsonBinary);
+
+        // when
+        Vector decodedVector = vectorCodec.decode(mockReader, DecoderContext.builder().build());
+
+        // then
+        assertNotNull(decodedVector);
+        assertEquals(vectorToDecode, decodedVector);
+    }
```
stIncMale, line 114: - Given that we have `shouldRoundTripVectors`, what value do these two tests add, and why do they need to use `Mockito`?
- If it turns out that these tests are actually useful and mocking is necessary, we should use `MongoMockito`.

bson/src/test/unit/org/bson/codecs/VectorCodecTest.java
```diff
@@ -0,0 +1,162 @@
...
+        when(mockReader.peekBinarySubType()).thenReturn(subType.getValue());
+
+        Stream.of(new Float32VectorCodec(), new Int8VectorCodec(), new PackedBitVectorCodec())
+                .forEach(codec -> {
+                    // when & then
+                    BSONException exception = assertThrows(BSONException.class, () ->
+                            codec.decode(mockReader, DecoderContext.builder().build()));
+                    assertEquals("Unexpected BsonBinarySubType", exception.getMessage());
+                });
+    }
```
stIncMale, line 131: I think we can decode data with invalid subtype without using mocking. Let's try and avoid resorting to mocking when it's not needed.

bson/src/test/unit/org/bson/codecs/VectorCodecTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @ParameterizedTest
+    @MethodSource("provideVectorsCodec")
+    void shouldConvertToString(final Codec<Vector> codec) {
+        // when
+        String result = codec.toString();
+
+        // then
+        assertEquals(codec.getClass().getSimpleName() + "{}", result);
+    }
```
stIncMale, line 152: It's probably the first `toString` test I've seen. I don't think we need it even if the new vector codecs need to override `toString`, but on top of that, I don't think they should do that, as other codecs usually don't.

bson/src/test/unit/org/bson/vector/VectorGenericBsonTest.java
```diff
@@ -0,0 +1,272 @@
...
+            }
+        }
+        return data.stream();
+    }
+
+    private static String createTestCaseDescription(final BsonDocument testDocument,
+                                                    final BsonDocument testCaseDocument) {
+        boolean isValidTestCase = testCaseDocument.getBoolean("valid").getValue();
+        String testSuiteDescription = testDocument.getString("description").getValue();
+        String testCaseDescription = testCaseDocument.getString("description").getValue();
```
stIncMale, line 269: We usually call these `fileDescription` and `testDescription`. Let's use the same names here, especially given that https://github.com/mongodb/specifications/tree/master/source/bson-binary-vector/tests does not mention "suite".

bson/src/test/unit/org/bson/vector/VectorGenericBsonTest.java
```diff
@@ -0,0 +1,272 @@
...
+        String testKey = testDefinition.getString("test_key").getValue();
+        boolean isValidVector = testCase.getBoolean("valid").getValue();
+        if (isValidVector) {
+            runValidTestCase(testKey, testCase);
+        } else {
+            runInvalidTestCase(testCase);
+        }
+    }
+
+    private void runInvalidTestCase(final BsonDocument testCase) {
```
stIncMale, line 86: Let's make `static` this and all the other methods in `VectorGenericBsonTest` except for `shouldPassAllOutcomes`.

stIncMale (changes requested):
Note that static checks don't pass. Please make sure there are no new warnings when building the project.

bson/src/test/unit/org/bson/codecs/ValueCodecProviderSpecification.groovy
```diff
@@ -32,6 +36,7 @@ import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicLong
 import java.util.regex.Pattern
 
+@SuppressWarnings("VectorIsObsolete")
```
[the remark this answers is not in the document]
vbabanin: It’s actually CodeNarc analysis that fails to identify that the `Vector` we use is not from `java.util.Vector`. I’ve added a comment above the `SuppressWarnings` annotation to clarify this.

bson/src/main/org/bson/Vector.java
jyemin: @stIncMale I saw your thumbs up but not sure what it means. Let it go or continue to bikeshed?

stIncMale: I haven't thumbsed up, that's not mine. However, I don't have any suggestions, nor do I object to `Vector`.

P.S. I do object to both Java SE and MongoDB calling vector components "elements", but I can't do anything with people disregarding actual terms and using arbitrary words instead.

bson/src/test/unit/org/bson/internal/vector/VectorHelperTest.java
```diff
@@ -0,0 +1,239 @@
...
+
+    @ParameterizedTest(name = "{index}: {0}")
+    @MethodSource("provideFloatVectors")
+    void shouldDecodeFloatVector(final Float32Vector expectedFloatVector, final byte[] bsonEncodedVector) {
+        // when
+        Float32Vector decodedVector = (Float32Vector) VectorHelper.decodeBinaryToVector(bsonEncodedVector);
+
+        // then
+        assertEquals(Vector.DataType.FLOAT32, decodedVector.getDataType());
+        assertArrayEquals(expectedFloatVector.getVectorArray(), decodedVector.getVectorArray());
```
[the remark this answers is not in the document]
vbabanin: Good catch—that was an oversight, as I wrote the tests before equals was implemented on Vector. I've corrected it to use assertEquals for comparison now. Thanks!

bson/src/main/org/bson/BsonBinary.java
jyemin: Good questions, @vbabanin.  

> Why do those types exist

These types go way back to 2009, five year before we decided to create the BsonValue type hierarchy.  Back then, the way to create BSON documents was with a class called BasicDBObject (the precursor to Document), which is a `Map<String, Object>`.  Like Document, you could add Integer/Long/Date, but then we created classes like ObjectId, Binary, etc. for the types that didn't have a good analog in the JDK.

The types in the BsonValue hierarchy, on the other hand, were meant to be used with BsonDocument, though in practice they can be used with Document as well.  However, in places like https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/data-formats/documents/ we list Binary and not BsonBinary as what an application would typically use with Document.

> Not only it does not add anything to the functionality of org.bson.BsonBinary, but it does not support UUIDs similarly to how BsonBinary does. Given this, why does it need to support Vector?

I think the reason we didn't add UUID support to Binary is that you can just put a UUID directly into a Document, whereas you can't do that with a BsonDocument.

jyemin: I now realize that we could do the same thing for Vector as we do for UUID: allow it to be serialized from and deserialized into a Document.  If we do that, then we could remove the Vector support from Binary.  But that won't just work out of the box.  The issue is round-tripping.  You can add a Vector to a Document and it will be encoded properly, but what happens during decoding? The `BsonTypeClassMap#DEFAULT_BSON_TYPE_CLASS_MAP` used by `DocumentCodec` will decode it as `Binary`, not Vector.  To make round-tripping work for UUIDs, there is an ugly switch statement in `ContainerCodecHelper#readValue` which would have to be adjusted to account for Vector.

So we have to decide whether that's the behavior that we want for Vector, or whether we should require users to manually convert between Binary and Vector.  UUID has worked this way since the dawn of time, but we can make a different decision for Vector if we want.

stIncMale: We talked about that exact idea with @vbabanin yesterday, and we decided to have the same behavior of `Vector` as it is for `UUID`. The issue is that round-tripping is broken either way:

- option one: encode `BsonBinary`/`Binary`/`Vector`, but decode `Vector`;
- option two: encode `BsonBinary`/`Binary`/`Vector`, but decode `BsonBinary`/`Binary`.

It looks like there are always situations when we cannot decode the Java type that was used when encoding.

jyemin: Yes, there is not enough information encoded in BSON to round trip.  We have to choose a canonical type to decode into regardless of what was used to encode.

> option two: encode BsonBinary/Binary/Vector, but decode BsonBinary/Binary.

Given that all other BSON binary are decoded as `Binary`, I think it would be unexpected to decode vectors as `BsonBinary`.

bson/src/main/org/bson/Int8Vector.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Retrieve the underlying byte array representing this {@link Int8Vector} vector, where each byte represents
+     * an element of a vector.
+     * <p>
+     * NOTE: The underlying byte array is not copied; changes to the returned array will be reflected in this instance.
+     *
+     * @return the underlying byte array representing this {@link Int8Vector} vector.
+     */
+    public byte[] getVectorArray() {
```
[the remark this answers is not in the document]
vbabanin: @stIncMale  and I discussed this on a call, and I agree with the mentioned above. I've renamed it to `data`, so the method to get raw vector array is now named `getData`.

Looking ahead, if additional representations are needed, we could introduce methods like `getDataAsBitSet` to get/wrap/copy the underlying array. If we convert, for example `PackedBitVector`,  to another structure in the future (potentially one from libraries we may incorporate for ML), we’d use `asSomeStructure()`. This way, methods prefixed with `"as"` would focus on converting the object into a new representation (preserving padding and data), while methods with `getData` would retrieve the vector data without the padding.

@jyemin , I know you had [concerns](https://github.com/mongodb/mongo-java-driver/pull/1528#discussion_r1796878940) about naming methods `getData`, however since `BsonBinary` also uses `getData`, this naming should align well. Technically, a `Vector` is represented by its `data` and `data type`, so I think this approach keeps things consistent and intuitive. (This was discussed with @stIncMale, and he had no objections to this naming—@stIncMale, please correct me if I’m mistaken.)

@jyemin what do you think?

bson/src/test/unit/org/bson/codecs/VectorCodecTest.java
vbabanin: Good point. Upon reviewing it, the round-trip test should actually be in `DocumentCodecTest` rather than `VectorCodecTest`, as it’s specifically testing that encoding and decoding a `Document` uses a `Vector` codecs by default when `Vector` object is present. I have already extended the existing round-trip test in `DocumentCodecTest` for that case.

Regarding the tests in `VectorCodecTest`, I’ve kept encode and decode but replaced the behavior verification approach (mocking) with a state verification approach that doesn’t rely on mocks.

bson/src/main/org/bson/BsonBinary.java
jyemin: OK, looks like you with the latest patch, any representation of Vector can be encoded b DocumentCodec: Vector subclasses, Binary, BsonBinary.  But vectors are always decoded to one of the Vector subclasses.

sgtm, resolving.

bson/src/main/org/bson/Int8Vector.java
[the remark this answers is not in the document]
jyemin: LGTM

bson/src/main/org/bson/codecs/VectorCodec.java
```diff
@@ -0,0 +1,62 @@
...
+    public void encode(final BsonWriter writer, final Vector vectorToEncode, final EncoderContext encoderContext) {
+        writer.writeBinaryData(new BsonBinary(vectorToEncode));
+    }
+
+    @Override
+    public Vector decode(final BsonReader reader, final DecoderContext decoderContext) {
+        byte subType = reader.peekBinarySubType();
+
+        if (subType != BsonBinarySubType.VECTOR.getValue()) {
+            throw new BSONException("Unexpected BsonBinarySubType");
```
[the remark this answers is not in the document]
vbabanin: You make a valid point. While `UUIDCodec` utilizes `BSONException` when the BSON subtype differs, most codecs throw `BsonInvalidOperationException` when there is a mismatch in BSON type.

I've updated `VectorCodec` to throw `BsonInvalidOperationException` to maintain consistency with most codecs.

bson/src/main/org/bson/Vector.java
```diff
@@ -0,0 +1,198 @@
...
+ * vectors.
+ * <p>
+ * Vectors are densely packed arrays of numbers, all the same type, which are stored efficiently
+ * in BSON using a binary format.
+ *
+ * @mongodb.server.release 6.0
+ * @see BsonBinary
+ * @since 5.3
+ */
+public abstract class Vector {
```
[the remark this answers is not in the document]
stIncMale: Can't we just use the wording from `@Sealed` (`This class should be treated as sealed: it must not be extended or implemented by consumers of the library.`), instead of using quite a different wording that is currently present in the PR?

bson/src/main/org/bson/BsonBinary.java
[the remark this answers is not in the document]
stIncMale: All the changes in `Binary.java` were rolled back.

bson/src/test/unit/org/bson/codecs/CodecTestCase.java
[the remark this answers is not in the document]
stIncMale: The changes were rolled back, but a new method `encode` was added, and it is not used anywhere. Is it really needed?

bson/src/main/org/bson/Vector.java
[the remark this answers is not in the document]
vbabanin: I considered adding more context on the potential consequences; however, i am fine with keeping it consistent. I’ve updated it to use the `@Sealed` wording.

bson/src/test/unit/org/bson/codecs/CodecTestCase.java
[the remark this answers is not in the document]
vbabanin: Removed, thanks!

driver-sync/src/test/functional/com/mongodb/client/vector/VectorAbstractFunctionalTest.java
```diff
@@ -0,0 +1,359 @@
...
+                .getDatabase(getDatabaseName())
+                .getCollection(getCollectionName());
+    }
+
+    @AfterEach
+    public void afterEach() {
+        super.afterEach();
+        if (mongoClient != null) {
+            mongoClient.close();
+        }
```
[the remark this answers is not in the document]
stIncMale: The change made does not resolve the issue, `mongoClient` may still be left not closed.

bson/src/main/org/bson/internal/vector/VectorHelper.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    /**
+     * Decodes a vector from a binary representation.
+     * <p>
+     * encodedVector is not mutated nor stored in the returned {@link Vector}.
+     */
+    public static Vector decodeBinaryToVector(final byte[] encodedVector) {
+        isTrue("Vector encoded array length must be at least 2.", encodedVector.length >= METADATA_SIZE);
```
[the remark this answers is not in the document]
stIncMale: While other places were changed, this particular error message still does not include the actual length.

I also missed requesting to change the following error message in `VectorHelper` in a similar way:
```java
isTrue("Byte array length must be a multiple of 4 for FLOAT32 data type.",
        (encodedVector.length - METADATA_SIZE) % Float.BYTES == 0);
```

stIncMale (changes requested):

bson/src/main/org/bson/codecs/ContainerCodecHelper.java
```diff
@@ -42,28 +44,53 @@ static Object readValue(final BsonReader reader, final DecoderContext decoderCon
...
+
+        if (binarySubType == BsonBinarySubType.VECTOR.getValue()) {
+            Codec<Vector> vectorCodec = registry.get(Vector.class, registry);
+            if (vectorCodec != null) {
+                return vectorCodec;
             }
-            return valueTransformer.transform(codec.decode(reader, decoderContext));
         }
+
+        if (reader.peekBinarySize() == 16) {
```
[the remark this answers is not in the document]
vbabanin: Do you mean that using else if could make the logical flow slightly more explicit? Functionally, it behaves the same due to the early return and separate check.

I’ve changed it to `else if ` to improve clarity.

stIncMale (approved):

PR #1531, opened 2024-10-16
Fix pom scope for Scala projects

rozza:
Required the `api` dependency method from the `java-library` plugin. Now the scope is `compile` instead of runtime.

JAVA-5647

### DIFF
--- a/bson-scala/build.gradle
+++ b/bson-scala/build.gradle
@@ -18,7 +18,7 @@ description = "A Scala wrapper / extension to the bson library"
 archivesBaseName = 'mongo-scala-bson'
 
 dependencies {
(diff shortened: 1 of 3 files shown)


rozza:
To check the generated poms run: 
```
./gradlew bson-scala:publishToMavenLocal driver-scala:publishToMavenLocal
```

Then the built jars & poms will then be in your local maven repository: `~/.m2/repository/org/mongodb/scala`.  

The `mongo-scala-bson_2.13` pom should have the bson library scoped to `compile`.
The `mongo-scala-driver_2.13` pom should have the scala bson library as well as the reactive streams driver scoped to `compile`.

NathanQingyangXu (approved):
LGTM

PR #1535, opened 2024-10-24
Bump Scala patch versions

rozza:
Updated to: 2.11.12, 2.12.20, 2.13.15

JAVA-5649

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -158,7 +158,7 @@ configure(scalaProjects) {
                     "-feature",
                     "-unchecked",
                     "-language:reflectiveCalls",
-                    "-Wconf:cat=deprecation:ws,any:e",
+                    "-Wconf:cat=deprecation:ws",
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -158,7 +158,7 @@ configure(scalaProjects) {
                     "-feature",
                     "-unchecked",
                     "-language:reflectiveCalls",
-                    "-Wconf:cat=deprecation:ws,any:e",
```
nhachicha, line 161: Any reason why we're removing the previous flag to report other warnings as error ?

nhachicha (approved):
LGTM, looks like `2.13.15` is binary compatible with all `2.13` series https://github.com/scala/scala/releases/tag/v2.13.15

build.gradle
rozza: For some reason that now turned all warnings into errors - so that caused the compile to fail. Where as in 2.13.6 it didnt.

PR #1538, opened 2024-10-29
sync up command logging specs

NathanQingyangXu:


driver-core/src/test/resources/unified-test-format/command-monitoring/find.json
```diff
@@ -390,7 +390,6 @@
     },
     {
       "description": "A successful find event with a getmore and the server kills the cursor (<= 4.4)",
-      "comment": "UPDATED final batchSize to 3 as batchSize is no longer calculated see: DRIVERS-1448 ",
       "runOnRequirements": [
```
NathanQingyangXu, line 393: I am not sure about the change but latest spec removes the comment and breaks the testing currently.
I added skipping mechanism to this testing case with the above comment there.

jyemin: The issue is that Java driver has already done the work for DRIVERS-1448, and change the test accordingly.  So let's just leave this as is and not skip it.

Another option is to ask the spec owner whether you can pick up DRIVERS-1448 as part of Spec Fest, and then we would be in sync once again.

driver-core/src/test/resources/unified-test-format/command-monitoring/unacknowledged-client-bulkWrite.json
```diff
@@ -0,0 +1,220 @@
+{
```
jyemin, line 1: Let's remove this file since Valentin is going to pull it it.

driver-core/src/test/resources/unified-test-format/command-monitoring/command.json
```diff
@@ -52,159 +54,25 @@
...
         }
       ],
-      "expectLogMessages": [
+      "expectEvents": [
         {
           "client": "client",
-          "messages": [
+          "events": [
             {
-              "level": "debug",
```
jyemin, line 60: What DRIVERS ticket is this large diff associated with?

jyemin (changes requested):

driver-core/src/test/resources/unified-test-format/command-monitoring/command.json
NathanQingyangXu: This large diff seems our previous overlooking. The command.json file under `command-monitoring` used to be the exactly the same as the `command.json` under `command-logging`.

driver-core/src/test/resources/unified-test-format/command-monitoring/find.json
jyemin: This is a great example, actually, of a problem we will have if we move to something like git submodules.  This is a test that a driver should pass prior to implementing DRIVERS-1448, and fail afterwards.  So instead of modifying this test as part of DRIVERS-1448, we should instead create a _new_ test that looks like this one.  Then a driver can pull in _both_ tests, and skip one or the other.

FYI @alcaeus

NathanQingyangXu: I reached out to spec author. hopefully tomorrow I can get some response.

NathanQingyangXu: Shane Harvey responded that given the spec change will involve other driver's changes (ours is fine), so he is not sure whether it is in the scope of the Spec Fest.
Not sure what to do here. Reverting back to the old spec so our evergreen could be green?

jyemin: Yes, then please revert the change

jyemin: @JamesKovacs ok for Nathan to proceed with DRIVERS-1448?

NathanQingyangXu: Yesterday I asked him directly and he confirmed. I'll do that separately. Now this PR is ready to be reviewed again.

jyemin (approved):
LGTM

PR #1539, opened 2024-10-29
Sync up transaction specs

NathanQingyangXu:


driver-core/src/test/resources/unified-test-format/transactions/mongos-pin-auto.json
```diff
@@ -2004,6 +2004,104 @@
         }
       ]
     },
+    {
```
jyemin, line 2007: Let's revert this change (and the associated skip of the tests) to save Valentin the hassle of a merge conflict.

jyemin (changes requested):

jyemin (approved):
LGTM

PR #1543, opened 2024-10-29
Sync CRUD unified tests

nhachicha:
- Removed all client bulkWrite tests until @stIncMale finishes the implementation 
- Added some skips in [driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudTest.java](https://github.com/nhachicha/mongo-java-driver/blob/8185ecdaad26a1f426d7891bd4eff5e335bd443a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudTest.java)


jyemin (changes requested):
While there are some advantages to including the YAML for the readability, we have so far decided to copy over only the JSON files, not the YAML ones.  Let's not change that here.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudTest.java
```diff
@@ -43,6 +43,12 @@ public static void doSkips(final String fileDescription, final String testDescri
             assumeFalse(testDescription.equals("Aggregate with $out includes read preference for 5.0+ server"));
             assumeFalse(testDescription.equals("Database-level aggregate with $out includes read preference for 5.0+ server"));
         }
+        assumeFalse(testDescription.toLowerCase().contains("updateone with sort option"), "Skipping until JAVA-5622 is implemented");
+        assumeFalse(testDescription.toLowerCase().contains("replaceone with sort option"), "Skipping until JAVA-5622 is implemented");
+        assumeFalse(testDescription.equals("BulkWrite with string comment"), "Skipping until we decide "
+                + "whether to update the spec or test https://mongodb.slack.com/archives/C72LB5RPV/p1730225657999979");
+        assumeFalse(testDescription.equals("BulkWrite with document comment"), "Skipping until we decide "
```
jyemin, line 50: Let's only include links that are accessible to the public. Slack doesn't fit the bill.

driver-core/src/test/resources/unified-test-format/crud/bulkWrite-comment.json
```diff
@@ -122,7 +122,6 @@
       "expectEvents": [
         {
           "client": "client0",
-          "note": "Events split due to : JAVA-4537 ",
```
jyemin, line 125: Since we had previously decided to change this test, let's not reverse this decision until we have a better alternative.  I'd prefer to revert the change to this file and leave it as is.

jyemin (changes requested):

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudTest.java
```diff
@@ -43,6 +43,22 @@ public static void doSkips(final String fileDescription, final String testDescri
             assumeFalse(testDescription.equals("Aggregate with $out includes read preference for 5.0+ server"));
             assumeFalse(testDescription.equals("Database-level aggregate with $out includes read preference for 5.0+ server"));
         }
+        if (fileDescription.equals("updateOne-sort")) {
+            assumeFalse(testDescription.equals("UpdateOne with sort option"));
```
jyemin, line 47: Let's add a second param for each of these, like `"Skipped until implementation of JAVA-XXXX"`, where XXXX is the issue number for adding sort to update/replace.

jyemin (changes requested):
LGTM after the requested change.

jyemin (approved):
Thanks.  LGTM

PR #1548, opened 2024-10-30
Sync up client-side-operation-timeout specs

NathanQingyangXu:


driver-core/src/test/resources/unified-test-format/client-side-operation-timeout/deprecated-options.json
```diff
@@ -1,6 +1,5 @@
 {
   "description": "operations ignore deprecated timeout options if timeoutMS is set",
-  "comment": "Manually changed session to use defaultTimeoutMS when testing socket / maxCommit overrides",
```
NathanQingyangXu, line 3: @katcharov here it is.

PR #1551, opened 2024-10-31
Clean up unified tests via skipping API

katcharov:
This is a lightweight wrapper API for organizing assumptions ("skips") in unified tests. Offers a consistent and clearer mechanism for specifying why tests are being skipped.

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedServerDiscoveryAndMonitoringTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedServerDiscoveryAndMonitoringTest.java
@@ -37,24 +37,20 @@ public final class UnifiedServerDiscoveryAndMonitoringTest extends UnifiedSyncTe
 
     public static void doSkips(final TestDef def) {
         def.skipJira("https://jira.mongodb.org/browse/JAVA-5230")
-                .file("server-discovery-and-monitoring", "connect with serverMonitoringMode=auto >=4.4")
-                .file("server-discovery-and-monitoring", "connect with serverMonitoringMode=stream >=4.4");
-
+                .test("server-discovery-and-monitoring", "serverMonitoringMode", "connect with serverMonitoringMode=auto >=4.4")
+                .test("server-discovery-and-monitoring", "serverMonitoringMode", "connect with serverMonitoringMode=stream >=4.4");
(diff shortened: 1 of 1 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestSkips.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        /**
+         * Test is skipped because it is pending implementation, and there is
+         * a Jira ticket tracking this which has more information.
+         *
+         * @param skip reason for skipping the test; must start with a Jira URL
+         */
+        public Skip skipJira(final String skip) {
+            Assertions.assertTrue(skip.startsWith("https://jira.mongodb.org/browse/JAVA-"));
```
nhachicha, line 49: Can the skip be due to a DRIVER-XXX ticket? I think checking against `https://jira.mongodb.org/browse/` is enough to make sure a valid JIRA link is provided

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestSkips.java
```diff
@@ -0,0 +1,113 @@
...
+                    && file.equals(testDef.file)
+                    && test.equals(testDef.test);
+            assumeFalse(match, reason);
+            return this;
+        }
+
+        /**
+         * Test skipped if the test description contains the fragment as a substring
+         */
+        public Skip testContains(final String dir, final String fragment) {
```
nhachicha, line 106: Is there a use case where we want to filter by dir and test description as well? Thinking whether we should add an optional `testDef.file` parameter to the method?

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestSkips.java
katcharov: I considered this, but I think we need to have something our team is tracking. If a good counterexample comes up, we can remove the constraint at the time?

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestSkips.java
katcharov: Let's add it if we run into it (hopefully we don't, since as you say the contains methods seem risky).

nhachicha (approved):
Nice 👍

PR #1555, opened 2024-11-01
Sync BSON specification tests

jyemin:
There are some changes we made long ago to double.json to account for how Java formats exponents, so double.json is not synced.  The diffs are in these two tests: https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/double.json#L31-L41.  Java formats the exponent as `E18` rather that `E+18`.  Both are perfectly valid so we just changed the test.


NathanQingyangXu:
The diff is just as you mentioned. Only double.json not synced up, and if we synced up, our `GenericBsonTest` would fail due to the difference between `E18` and `E+18`.
LGTM

NathanQingyangXu (approved):

PR #1404, opened 2024-05-29
Remove legacy shell from test scripts

jyemin:
The legacy shell was only used in AWS authentication tests, so updating those gets rid of the last remaining use of the legacy shell.

JAVA-4791

PR #1560, opened 2024-11-05
Remove legacy shell from test scripts (#1404)

vbabanin:
The legacy shell was only used in AWS authentication tests, so updating those gets rid of the last remaining use of the legacy shell.

Backport of (https://github.com/mongodb/mongo-java-driver/pull/1404)

JAVA-4791


katcharov (approved):
LGTM. Did not review in detail, assuming it is a no-conflict cherrypick of the other commit.

PR #1564, opened 2024-11-15
Fix static checker warnings in `AggregatesSpecification.groovy`

stIncMale:


vbabanin (approved):
Thank you very much for fixing that!

PR #1565, opened 2024-11-20
Retry flaky unified tests

katcharov:
JAVA-5795

See [WRITING-28651](https://jira.mongodb.org/browse/WRITING-28651)

This creates N copies (attempts) of the unified tests, and places a try-catch around shouldPassAllOutcomes. If an attempt passes, ensuing attempts are ignored. Failures are ignored, except the last attempt.

This is not intended to fully resolve the issue, but is a low-cost fix that should reduce failing tests.

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
@@ -57,7 +57,6 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
-import org.opentest4j.AssertionFailedError;
 import org.opentest4j.TestAbortedException;
 
 import java.io.File;
@@ -96,6 +95,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assumptions.abort;
 import static org.junit.jupiter.api.Assumptions.assumeFalse;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import static util.JsonPoweredTestHelper.getTestDocument;
@@ -107,7 +107,7 @@ public abstract class UnifiedTest {
             "wait queue timeout errors include details about checked out connections");
 
     public static final int ATTEMPTS = 3;
-    private static Set<String> completed = new HashSet<>();
+    private static Set<String> ignoreRemaining = new HashSet<>();
 
     @Nullable
     private String fileDescription;
@@ -348,8 +348,9 @@ public abstract class UnifiedTest {
             final BsonDocument definition) {
         boolean forceFlaky = totalAttempts < 0;
         if (!forceFlaky) {
-            assumeFalse(completed.contains(testName), "Skipping retryable test that succeeded");
-            completed.add(testName);
+            boolean ignoreThisTest = ignoreRemaining.contains(testName);
+            assumeFalse(ignoreThisTest, "Skipping a retryable test that already succeeded");
+            ignoreRemaining.add(testName);
         }
         try {
             BsonArray operations = definition.getArray("operations");
@@ -376,20 +377,21 @@ public abstract class UnifiedTest {
                 }
                 compareLogMessages(rootContext, definition, tweaks);
             }
-        } catch (AssertionFailedError e) {
-            assertTrue(testDef.wasAssignedModifier(Modifier.RETRY));
-            if (!testDef.matchesError(e)) {
-                // if the error is not matched, test definitions were not intended to apply; throw it
+        } catch (Throwable e) {
+            if (forceFlaky) {
                 throw e;
             }
-
-            completed.remove(testName);
-            boolean lastAttempt = attemptNumber == Math.abs(totalAttempts);
-            if (forceFlaky || lastAttempt) {
+            if (!testDef.matchesThrowable(e)) {
+                // if the throwable is not matched, test definitions were not intended to apply; rethrow it
                 throw e;
-            } else {
-                assumeFalse(completed.contains(testName), "Ignoring failure and retrying attempt " + attemptNumber);
             }
+            boolean isLastAttempt = attemptNumber == Math.abs(totalAttempts);
+            if (isLastAttempt) {
+                throw e;
+            }
+            
+            ignoreRemaining.remove(testName);
+            abort("Ignoring failure and retrying attempt " + attemptNumber);
         }
     }
 
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
@@ -107,7 +107,7 @@ public abstract class UnifiedTest {
             "wait queue timeout errors include details about checked out connections");
 
     public static final int ATTEMPTS = 3;
-    private static Set<String> ignoreRemaining = new HashSet<>();
+    private static final Set<String> ATTEMPTED_TESTS_TO_HENCEFORTH_IGNORE = new HashSet<>();
 
     @Nullable
     private String fileDescription;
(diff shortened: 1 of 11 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
```diff
@@ -30,13 +32,27 @@
...
+        def.retry("TODO reason")
+                .whenFailureContains("ChangeCipherSpec message sequence violation")
+                .test("client-side-encryption", "namedKMS-createDataKey", "create datakey with named KMIP KMS provider");
+
+        def.retry("TODO reason")
+                .whenFailureContains("Number of checked out connections must match expected")
+                .test("load-balancers", "cursors are correctly pinned to connections for load-balanced clusters", "pinned connections are returned after a network error during a killCursors request");
+
+        def.retry("TODO reason")
+                .test("client-side-encryption", "namedKMS-rewrapManyDataKey", "rewrap to kmip:name1");
```
katcharov, line 55: These are flaky failures from evergreen. Reasons should be specified, or these 3 retries can be removed and added separately, after examining the failures.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -298,40 +325,66 @@ protected boolean isReactive() {
...
+                }
+                compareLogMessages(rootContext, definition, tweaks);
+            }
+        } catch (Throwable e) {
+            if (forceFlaky) {
+                throw e;
+            }
+            if (!testDef.matchesThrowable(e)) {
+                // if the throwable is not matched, test definitions were not intended to apply; rethrow it
+                throw e;
```
katcharov, line 379: I tested this by failing an assertion within the try of `shouldPassAllOutcomes`. When the assertion message matches what is specified in a def, it will ignore, ignore, and then fail with that assertion failure. When it does not match, the first instance will fail, and the remaining will be ignored.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -156,32 +164,47 @@ public Entities getEntities() {
     }
 
     @NonNull
-    protected static Collection<Arguments> getTestData(final String directory) throws URISyntaxException, IOException {
```
katcharov, line 159: This should be a straightforward inline, along with the changes to the test parameters.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -218,9 +244,9 @@ public void setUp(
         ignoreExtraEvents = false;
         if (directoryName != null && fileDescription != null && testDescription != null) {
```
vbabanin, line 245: I'm not sure if it's actually possible for these parameters to be null.

- directory is always passed to getTestData and cannot be null.
- fileDescription and testDescription are required fields in unified tests, as specified [here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#test).
- This runner is specifically for unified tests.

Given this, I think we could safely remove these null checks, or alternatively, handle this and Nullable removal later. What do you think?

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
vbabanin: Since we have a separate ticket for this, i think we should handle it there to better understand and document why these tests are flaky.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -305,40 +332,66 @@ protected boolean isReactive() {
...
-        BsonArray operations = definition.getArray("operations");
-        for (int i = 0; i < operations.size(); i++) {
-            BsonValue cur = operations.get(i);
-            assertOperation(rootContext, cur.asDocument(), i);
+        boolean forceFlaky = totalAttempts < 0;
+        if (!forceFlaky) {
+            boolean ignoreThisTest = ignoreRemaining.contains(testName);
+            assumeFalse(ignoreThisTest, "Skipping a retryable test that already succeeded");
+            ignoreRemaining.add(testName);
         }
```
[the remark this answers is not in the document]
vbabanin: Ah, I see! Thanks for the clarification—makes sense now.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -156,32 +164,47 @@ public Entities getEntities() {
...
-                data.add(UnifiedTest.createTestData(directory, fileDocument, cur.asDocument()));
+
+                final BsonDocument testDocument = cur.asDocument();
+                String testDescription = testDocument.getString("description").getValue();
+                String fileDescription = fileDocument.getString("description").getValue();
+                TestDef testDef = testDef(directory, fileDescription, testDescription, isReactive);
+                applyCustomizations(testDef);
+
+                boolean forceFlaky = testDef.wasAssignedModifier(Modifier.FORCE_FLAKY);
+                boolean retry = forceFlaky || testDef.wasAssignedModifier(Modifier.RETRY);
```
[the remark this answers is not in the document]
katcharov: I clarified the below `if`. (Though it was not previously used, it should still represent whether we plan to retry.)

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -156,32 +164,47 @@ public Entities getEntities() {
...
+                            fileDescription,
+                            testDescription,
+                            directory,
+                            attempt,
+                            attempts * (forceFlaky ? -1 : 1),
+                            fileDocument.getString("schemaVersion").getValue(),
+                            fileDocument.getArray("runOnRequirements", null),
+                            fileDocument.getArray("createEntities", new BsonArray()),
+                            fileDocument.getArray("initialData", new BsonArray()),
+                            testDocument));
```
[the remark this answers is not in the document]
katcharov: I do not understand where testDocument is being changed. I've followed the various paths in `shouldPassAllOutcomes`, but these seem to not modify. Could you link me to where you found the modification happening? 

Or which tests you ran? I ran UnifiedGridFSTest and did not see a failure.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
katcharov: @stIncMale added this logic. I do not remember all the details, but the message is:

> When `com.mongodb.client.unified.UnifiedTest` is used by `com.mongodb.workload.WorkloadExecutor` for running tests with Astrolabe, `TestDef` shouldn't be created. JAVA-5716

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
katcharov: In the time since the draft of this ticket, we began to use the "TODO-JAVA-N" pattern for TODOs, and (if I am recalling correctly) we also decided to disable flaky tests (as above), but track the issue you are describing in Jira. This way, the flaky tests do not obscure evergreen, or force us to manually retry, but we still must decide what to do with them.

I have created a ticket to do this here, https://jira.mongodb.org/browse/JAVA-5809, which will come up in triage, and have added the relevant TODO comments (as well as some others). (In any case, there are existing flaky annotations that are currently not marked with reasons, which I have now marked with TODOs.)

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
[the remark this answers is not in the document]
katcharov: 👍 
(Feel free to resolve.)

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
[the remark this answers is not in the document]
katcharov: (I did change this to use clone, but I would still like to understand what is going on.)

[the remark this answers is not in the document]
vbabanin: One of the tests where this can be observed is `timeoutMS can be overridden for upload` [Code reference](https://github.com/vbabanin/mongo-java-driver/blob/01aff5a0789f71b9d0b56190d4996e8ac5436827/driver-core/src/test/resources/unified-test-format/client-side-operation-timeout/gridfs-upload.json#L98).

You could try adding these lines to `UnifiedTestModifications` to check its behavior:

```
def.modify(FORCE_FLAKY).onMatch(def.test.contains("timeoutMS can be overridden for upload"));
def.skipDeprecated("").onMatch(!def.test.contains("timeoutMS can be overridden for upload"));
```
Then, running `ClientSideOperationTimeoutTest` should reflect the behavior where subsequent retries are currently failing, marking them reported as flaky, while they should all pass.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
katcharov: Removed

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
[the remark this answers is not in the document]
katcharov: Thank you! I will leave clone.

vbabanin (approved):
LGTM! Great addition to enhance test stability, @katcharov !

PR #1578, opened 2024-12-06
Make sure `TestDef` is created and used only when there is enough data to do that

stIncMale:
When `com.mongodb.client.unified.UnifiedTest` is used by `com.mongodb.workload.WorkloadExecutor` for running tests with Astrolabe, `TestDef` shouldn't be created.

JAVA-5716


nhachicha (approved):
Nice👌

PR #1581, opened 2024-12-11
Add index hint support for distinct command

joykim1005:
### Ticket
[JAVA-5686](https://jira.mongodb.org/browse/JAVA-5686?filter=-1)

### Description
This PR adds index hint support for `distinct` command for java, kotlin, and scala. It also syncs CRUD unified spec tests to [77b1f78458cf25b525cde2274e33acd5db2581d2](https://github.com/mongodb/specifications/commit/77b1f78458cf25b525cde2274e33acd5db2581d2).

### Testing
- [X] ran `./gradlew check`
- [X] local testing - ran `UnifiedCRUDTest.java`

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/DistinctOperation.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/DistinctOperation.java
@@ -54,6 +54,8 @@ public class DistinctOperation<T> implements AsyncReadOperation<AsyncBatchCursor
     private BsonDocument filter;
     private Collation collation;
     private BsonValue comment;
+    private BsonValue hint;
+    private String hintString;
 
     public DistinctOperation(final MongoNamespace namespace, final String fieldName, final Decoder<T> decoder) {
         this.namespace = notNull("namespace", namespace);
@@ -97,6 +99,24 @@ public class DistinctOperation<T> implements AsyncReadOperation<AsyncBatchCursor
         return this;
     }
 
+    public BsonValue getHint() {
+        return hint;
+    }
+
+    public DistinctOperation<T> hint(@Nullable final BsonValue hint) {
+        this.hint = hint;
+        return this;
+    }
+
+    public String getHintString() {
+        return hintString;
+    }
+
+    public DistinctOperation<T> hintString(@Nullable final String hint) {
+        this.hintString = hint;
+        return this;
+    }
+
     @Override
     public BatchCursor<T> execute(final ReadBinding binding) {
         return executeRetryableRead(binding, namespace.getDatabaseName(), getCommandCreator(), createCommandDecoder(),
@@ -124,6 +144,7 @@ public class DistinctOperation<T> implements AsyncReadOperation<AsyncBatchCursor
                 commandDocument.put("collation", collation.asDocument());
             }
             putIfNotNull(commandDocument, "comment", comment);
+            putIfNotNull(commandDocument, "hint", hint);
             return commandDocument;
         };
     }
--- a/driver-core/src/main/com/mongodb/internal/operation/DistinctOperation.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/DistinctOperation.java
@@ -55,7 +55,6 @@ public class DistinctOperation<T> implements AsyncReadOperation<AsyncBatchCursor
     private Collation collation;
     private BsonValue comment;
     private BsonValue hint;
-    private String hintString;
 
     public DistinctOperation(final MongoNamespace namespace, final String fieldName, final Decoder<T> decoder) {
         this.namespace = notNull("namespace", namespace);
@@ -105,15 +104,6 @@ public class DistinctOperation<T> implements AsyncReadOperation<AsyncBatchCursor
 
     public DistinctOperation<T> hint(@Nullable final BsonValue hint) {
         this.hint = hint;
-        return this;
-    }
-
-    public String getHintString() {
-        return hintString;
-    }
-
-    public DistinctOperation<T> hintString(@Nullable final String hint) {
-        this.hintString = hint;
         return this;
     }
 
--- a/driver-core/src/main/com/mongodb/internal/operation/Operations.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/Operations.java
@@ -217,13 +217,14 @@ final class Operations<TDocument> {
     }
 
     <TResult> DistinctOperation<TResult> distinct(final String fieldName, @Nullable final Bson filter, final Class<TResult> resultClass,
-            final Collation collation, final BsonValue comment) {
+            final Collation collation, final BsonValue comment, @Nullable final Bson hint, @Nullable final String hintString) {
         return new DistinctOperation<>(assertNotNull(namespace),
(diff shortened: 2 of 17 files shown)


driver-core/src/main/com/mongodb/internal/operation/Operations.java
```diff
@@ -217,13 +217,14 @@ private <TResult> FindOperation<TResult> createFindOperation(final MongoNamespac
...
-            final Collation collation, final BsonValue comment) {
+            final Collation collation, final BsonValue comment, @Nullable final Bson hint, @Nullable final String hintString) {
         return new DistinctOperation<>(assertNotNull(namespace),
                 fieldName, codecRegistry.get(resultClass))
                 .retryReads(retryReads)
                 .filter(filter == null ? null : filter.toBsonDocument(documentClass, codecRegistry))
                 .collation(collation)
-                .comment(comment);
+                .comment(comment)
+                .hint(hint != null ? toBsonDocument(hint) : (hintString != null ? new BsonString(hintString) : null));
```
NathanQingyangXu, line 227: It is fine to implement via one-liner, but I think the similar way in the above `createFindOperation()` is cleaner and more readable.

NathanQingyangXu:
```suggestion
        DistinctOperation<TResult> operation = new DistinctOperation<>(assertNotNull(namespace),
                fieldName, codecRegistry.get(resultClass))
                .retryReads(retryReads)
                .filter(filter == null ? null : filter.toBsonDocument(documentClass, codecRegistry))
                .collation(collation)
                .comment(comment);
        if (hint != null) {
            operation.hint(toBsonDocument(hint));
        } else if (hintString != null) {
            operation.hintString(hintString);
        }
        return operation;
```

driver-core/src/main/com/mongodb/internal/operation/DistinctOperation.java
```diff
@@ -97,6 +99,24 @@ public DistinctOperation<T> comment(final BsonValue comment) {
...
+
+    public String getHintString() {
+        return hintString;
+    }
+
+    public DistinctOperation<T> hintString(@Nullable final String hint) {
+        this.hintString = hint;
+        return this;
+    }
+
```
NathanQingyangXu, line 119: I think `hint` field suffices which is of `BsonValue` type that will accomodate both `BsonDocument` and `BsonString`.

Similar scenario applies in `FindOperation` within the same package which has a `hint` field as well.

driver-kotlin-coroutine/src/main/kotlin/com/mongodb/kotlin/client/coroutine/DistinctFlow.kt
```diff
@@ -103,5 +104,21 @@ public class DistinctFlow<T : Any>(private val wrapped: DistinctPublisher<T>) :
...
+    public fun hint(@Nullable hint: Bson?): DistinctFlow<T> = apply { wrapped.hint(hint) }
+
+    /**
+     * Sets the hint for which index to use. A null value means no hint is set.
+     *
+     * @param hint the name of the index which should be used for the operation
+     * @return this
+     */
+    public fun hintString(@Nullable hint: String?): DistinctFlow<T> = apply { wrapped.hintString(hint) }
+
```
NathanQingyangXu, line 122: It seems the doc is consistent with the counterpart of `FindFlow.kt`, except for the following note for `hintString()`:
```
* Note: If [DistinctFlow.hint] is set that will be used instead of any hint string.
```
could we add this detail here as well?

driver-kotlin-sync/src/main/kotlin/com/mongodb/kotlin/client/DistinctIterable.kt
```diff
@@ -99,4 +100,20 @@ public class DistinctIterable<T : Any?>(private val wrapped: JDistinctIterable<T
...
+     */
+    public fun hint(@Nullable hint: Bson?): DistinctIterable<T> = apply { wrapped.hint(hint) }
+
+    /**
+     * Sets the hint for which index to use. A null value means no hint is set.
+     *
+     * @param hint the name of the index which should be used for the operation
+     * @return this
+     */
+    public fun hintString(@Nullable hint: String?): DistinctIterable<T> = apply { wrapped.hintString(hint) }
```
NathanQingyangXu, line 118: again, could we add the following detail to above to be consistent with `FindIterable.kt`?
```
     * Note: If [DistinctIterable.hint] is set that will be used instead of any hint string.
```

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/syncadapter/SyncDistinctIterable.java
```diff
@@ -71,6 +71,18 @@ public DistinctIterable<T> comment(@Nullable final BsonValue comment) {
...
     }
 
+    @Override
+    public DistinctIterable<T> hint(@Nullable final Bson hint) {
+        wrapped.hint(hint);
+        return this;
+    }
+
+    @Override
+    public DistinctIterable<T> hintString(final String hint) {
```
NathanQingyangXu, line 81: seems a `@Nullable` is missing above

driver-scala/src/main/scala/org/mongodb/scala/DistinctObservable.scala
```diff
@@ -111,6 +111,30 @@ case class DistinctObservable[TResult](private val wrapped: DistinctPublisher[TR
...
+    wrapped.hint(hint)
+    this
+  }
+
+  /**
+   * Sets the hint for this operation. A null value means no hint is set.
+   *
+   * @param hint the name of the index which should be used for the operation
+   * @return this
+   * @since 5.3
```
NathanQingyangXu, line 131: again, we'd better add the following key detail in line with `FindObservable`:
```
   * @note if [[hint]] is set that will be used instead of any hint string.
```

driver-sync/src/main/com/mongodb/client/DistinctIterable.java
```diff
@@ -92,6 +92,24 @@ public interface DistinctIterable<TResult> extends MongoIterable<TResult> {
...
+    DistinctIterable<TResult> hint(@Nullable Bson hint);
+
+    /**
+     * Sets the hint to apply.
+     *
+     * @param hint the name of the index which should be used for the operation
+     * @return this
+     * @since 5.3
+     */
+    DistinctIterable<TResult> hintString(@Nullable String hint);
```
NathanQingyangXu, line 111: again, it would be great to add the following detail above:
```
 * <p>Note: If {@link DistinctIterable#hint(Bson)} is set that will be used instead of any hint string.</p>
 ```

NathanQingyangXu (changes requested):

driver-core/src/main/com/mongodb/internal/operation/Operations.java
joykim1005: Thank you for the suggestion, will update!

NathanQingyangXu (approved):

PR #1585, opened 2024-12-18
Allow updateOne and replaceOne to supply sort option

joykim1005:
### Ticket
[JAVA-5722](https://jira.mongodb.org/browse/JAVA-5722)

### Description
This PR adds a sort option for `updateOne` and `replaceOne` commands so if it matches more than one candidate document, the first one matched by sort order will be updated. It also syncs CRUD unified spec tests to [this](https://github.com/mongodb/specifications/pull/1644/files).

### Testing
- [X] ran `./gradlew check`
- [X] local testing - ran `UnifiedCRUDTest.java`

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/BulkWriteOptions.java
+++ b/driver-core/src/main/com/mongodb/client/model/BulkWriteOptions.java
@@ -17,6 +17,7 @@
 package com.mongodb.client.model;
 
 import com.mongodb.lang.Nullable;
+import org.bson.BsonDocument;
 import org.bson.BsonString;
 import org.bson.BsonValue;
 import org.bson.conversions.Bson;
@@ -31,6 +32,7 @@ public final class BulkWriteOptions {
     private Boolean bypassDocumentValidation;
     private BsonValue comment;
     private Bson variables;
+    private BsonDocument sort;
 
     /**
      * If true, then when a write fails, return without performing the remaining
@@ -147,6 +149,32 @@ public final class BulkWriteOptions {
         return this;
     }
 
+    /**
+     * Gets the sort criteria to apply to the query. The default is null, which means that the documents will be returned in an undefined
+     * order.
+     *
+     * @return a document describing the sort criteria
+     * @since 5.3
+     * @mongodb.driver.manual reference/method/cursor.sort/ Sort
+     */
+    @Nullable
+    public BsonDocument getSort() {
+        return sort;
+    }
+
+    /**
+     * Sets the sort criteria to apply to the query.
+     *
+     * @param sort the sort criteria, which may be null.
+     * @return this
+     * @since 5.3
+     * @mongodb.driver.manual reference/method/cursor.sort/ Sort
+     */
+    public BulkWriteOptions sort(@Nullable final BsonDocument sort) {
+        this.sort = sort;
+        return this;
+    }
+
     @Override
     public String toString() {
         return "BulkWriteOptions{"
@@ -154,6 +182,7 @@ public final class BulkWriteOptions {
                 + ", bypassDocumentValidation=" + bypassDocumentValidation
                 + ", comment=" + comment
                 + ", let=" + variables
+                + ", sort=" + sort
                 + '}';
     }
 }
--- a/driver-core/src/main/com/mongodb/client/model/BulkWriteOptions.java
+++ b/driver-core/src/main/com/mongodb/client/model/BulkWriteOptions.java
@@ -17,7 +17,6 @@
 package com.mongodb.client.model;
 
 import com.mongodb.lang.Nullable;
-import org.bson.BsonDocument;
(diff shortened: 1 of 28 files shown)


driver-core/src/main/com/mongodb/client/model/UpdateOptions.java
```diff
@@ -256,6 +258,32 @@ public UpdateOptions let(final Bson variables) {
         return this;
     }
 
+    /**
+     * Gets the sort criteria to apply to the query. The default is null, which means that the documents will be returned in an undefined
+     * order.
+     *
+     * @return a document describing the sort criteria
+     * @since 5.3
+     * @mongodb.driver.manual reference/method/cursor.sort/ Sort
```
vbabanin, line 267: Should we remove the manual reference here and from the `getSort` method, given that it seems more relevant to the cursor?

driver-core/src/main/com/mongodb/client/model/UpdateOptions.java
```diff
@@ -256,6 +258,32 @@ public UpdateOptions let(final Bson variables) {
...
+
+    /**
+     * Sets the sort criteria to apply to the query.
+     *
+     * @param sort the sort criteria, which may be null.
+     * @return this
+     * @since 5.3
+     * @mongodb.driver.manual reference/method/cursor.sort/ Sort
+     */
+    public UpdateOptions sort(@Nullable final BsonDocument sort) {
```
vbabanin, line 282: Let's leave a note in Javadoc about sort not applying to `updateMany`.

driver-core/src/main/com/mongodb/client/model/UpdateOptions.java
```diff
@@ -256,6 +258,32 @@ public UpdateOptions let(final Bson variables) {
...
+     * @mongodb.driver.manual reference/method/cursor.sort/ Sort
+     */
+    @Nullable
+    public BsonDocument getSort() {
+        return sort;
+    }
+
+    /**
+     * Sets the sort criteria to apply to the query.
+     *
```
vbabanin, line 276: Could we clarify the purpose of this sort option in the Javadoc? It would be helpful to explain how it determines which document is updated when multiple matches are found.

```suggestion
     *
     * When multiple documents match the query, the sort order specifies the document 
     * to be updated first.
     *
```

driver-core/src/main/com/mongodb/client/model/BulkWriteOptions.java
```diff
@@ -31,6 +32,7 @@ public final class BulkWriteOptions {
     private Boolean bypassDocumentValidation;
     private BsonValue comment;
     private Bson variables;
+    private BsonDocument sort;
```
vbabanin, line 35: Judging by the [spec](https://github.com/qingyang-hu/specifications/blob/master/source/crud/crud.md), BulkWriteOptions does not seem to require a sort option.

vbabanin (changes requested):
Looking good overall! I noticed in the spec changes here: https://github.com/mongodb/specifications/pull/1644/files#diff-d6a7b61abed64b14a2e1eb4827556e8ed74c84bc8ae1f635ff8de67a4cfbcfba that `ReplaceOneOptions` might also need to be updated. Should we include that in this PR?

joykim1005:
> Looking good overall! I noticed in the spec changes here: https://github.com/mongodb/specifications/pull/1644/files#diff-d6a7b61abed64b14a2e1eb4827556e8ed74c84bc8ae1f635ff8de67a4cfbcfba that `ReplaceOneOptions` might also need to be updated. Should we include that in this PR?

I originally didn't include it because the ticket specifically dealt with `updateOne`, but I can go ahead and include that in this PR and update the ticket as it is adding the same option!

driver-core/src/main/com/mongodb/client/model/BulkWriteOptions.java
joykim1005: I overlooked that part, will remove it!

driver-core/src/main/com/mongodb/client/model/UpdateOptions.java
joykim1005: This is a good point, will go ahead and add it!

driver-core/src/main/com/mongodb/client/model/UpdateOptions.java
joykim1005: Will do!

driver-core/src/main/com/mongodb/client/model/UpdateOptions.java
joykim1005: I referenced the `findOptions` and the other `findOneAnd..Options` and thought it might be helpful to have a reference! I don't have a strong opinion here though.

vbabanin (approved):

PR #1587, opened 2024-12-23
Add regex operator to Atlas Search

joykim1005:
### Ticket
[JAVA-5726](https://jira.mongodb.org/browse/JAVA-5726)

### Description
There are several Atlas Search query operators that are not implemented with first class support in the Java driver. This PR adds the `regex` operator to Atlas Search.

### Testing
- [X] ran `./gradlew check`
- [X] ran all [Evergreen patch ](https://spruce.mongodb.com/version/6769aa5a191fe8000776f2d5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


driver-core/src/main/com/mongodb/client/model/search/SearchConstructibleBsonElement.java
```diff
@@ -88,6 +88,11 @@ public TextSearchOperator synonyms(final String name) {
         });
     }
 
+    @Override
+    public RegexSearchOperator allowAnalyzedField(final boolean isAllowed) {
+        return newWithAppendedValue("allowAnalyzedField", notNull("isAllowed", isAllowed));
+    }
```
katcharov, line 94:
```suggestion
    public RegexSearchOperator allowAnalyzedField(final boolean allowAnalyzedField) {
        return newWithAppendedValue("allowAnalyzedField", allowAnalyzedField);
    }
```
Redundant null check. For naming, a similar method uses the same param name as the method itself:

```
    public SearchOptions returnStoredSource(final boolean returnStoredSource) {
        return newAppended("returnStoredSource", new BsonBoolean(returnStoredSource));
    }
```

This can be changed in the interfaces as well.

driver-core/src/main/com/mongodb/client/model/search/RegexSearchOperator.java
```diff
@@ -0,0 +1,43 @@
...
+@Sealed
+@Beta(Reason.CLIENT)
+public interface RegexSearchOperator extends SearchOperator {
+    @Override
+    RegexSearchOperator score(SearchScore modifier);
+
+    /**
+     * Creates a new {@link RegexSearchOperator} that uses allowAnalyzedField.
+     *
+     * @param isAllowed The boolean to run the query against an analyzed field.
```
katcharov, line 37:
```suggestion
     * @param isAllowed Must be set to true if the query is run against an analyzed field.
```

katcharov (changes requested):
Just a couple of minor nits.

driver-core/src/main/com/mongodb/client/model/search/SearchConstructibleBsonElement.java
joykim1005: Got it.

driver-core/src/main/com/mongodb/client/model/search/RegexSearchOperator.java
joykim1005: Done

katcharov (approved):

katcharov (changes requested):

driver-core/src/main/com/mongodb/client/model/search/RegexSearchOperator.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Creates a new {@link RegexSearchOperator} that uses allowAnalyzedField.
+     *
+     * @param allowAnalyzedField Must be set to true if the query is run against an analyzed field.
+     * @return A new {@link RegexSearchOperator}.
+     *
+     * @mongodb.atlas.manual atlas-search/allowAnalyzedField/ allowAnalyzedField mappings
+     */
+    RegexSearchOperator allowAnalyzedField(boolean allowAnalyzedField);
```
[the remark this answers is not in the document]
stIncMale: Could you please clarify what the suggestion is?

> set it to true as part of the plain regex builder

I am guessing, the "plain regex builder" means the `SearchOperator.regex` methods. Does that mean that the suggestion is to just always set `allowAnalyzedField` to `true` when `SearchOperator.regex` is called, thus "hiding" the existence of this field, and not allowing it to be `false`/absent?

> The operation will not change the original data

What operation does this refer to? The [`$search`](https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/search/) aggregation pipeline indeed does not have the capacity to change the data being searched, but I am not sure why it is important here.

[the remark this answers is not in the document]
joykim1005: Synced offline and concluded that we should delete `allowAnalyzedField` from `regex` and `wildcard` operators as it could be deprecated in the future, meaning the users will have no "supported" way to set it. However, the users can set the field manually or use raw Bson.

joykim1005:
`com.mongodb.client.ClientSideEncryptionE…estExternal_withExternalKeyVault__false` is failing and is unlikely caused by this PR and is happening in others as well.

katcharov (approved):
LGTM

PR #1588, opened 2024-12-26
Add queryString operator to Atlas Search

joykim1005:
### Ticket
[JAVA-5727](https://jira.mongodb.org/browse/JAVA-5727)

### Description
There are several Atlas Search query operators that are not implemented with first class support in the Java driver. This PR adds the queryString operator to Atlas Search.

### Testing
- [X] ran ./gradlew check
- [X] ran atlas-search-test on Evergreen
- [X] ran SearchOperatorTest
- [X] ran [everygreen patch](https://spruce.mongodb.com/version/676cd53717f01a0007ccdddf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) on all tests


driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +292,23 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
 
+    /**
+     * Returns a {@link SearchOperator} that supports querying a combination of indexed fields and values.
+     *
+     * @param defaultPath The field to be searched by default.
+     * @param query One or more indexed fields and values to search.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/queryString/ queryString operator
+     */
+    static QueryStringSearchOperator queryString(final SearchPath defaultPath, final String query) {
```
katcharov, line 303: I think it should not be possible to write `queryString(wildcardPath(...` - though the query supports wildcards, I believe the defaultPath itself does not. Could you confirm?

katcharov (changes requested):

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
joykim1005: You are right, I mistook query supporting wildcards with the defaultPath being able to support it too. Will go ahead and change it!

joykim1005: You're right, will change `SearchPath` to `FieldSearchPath` to limit defaultPath to use wildcardPath!

driver-scala/src/main/scala/org/mongodb/scala/model/search/SearchOperator.scala
```diff
@@ -228,6 +228,17 @@ object SearchOperator {
...
 
+  /**
+   * Returns a `SearchOperator` that supports querying a combination of indexed fields and values.
+   *
+   * @param defaultPath The field to be searched by default.
+   * @param query One or more indexed fields and values to search.
+   * @return The requested `SearchOperator`.
+   * @see [[https://www.mongodb.com/docs/atlas/atlas-search/queryString/ queryString operator]]
+   */
+  def queryString(defaultPath: SearchPath, query: String): QueryStringSearchOperator =
```
nhachicha, line 239: Shouldn't we also include the `score` parameter? (see example https://www.mongodb.com/docs/atlas/atlas-search/queryString/)

joykim1005: I left out the `score` as a parameter since it's an optional field and instead have it in the interface. All the other search operators also include `score` in the interface with overrides!

katcharov (approved):

PR #1589, opened 2024-12-30
JAVA-5736: Add bsonNamingStrategy option to support snake_case naming strategy

leesungbin:
## Ticket
[JAVA-5736](https://jira.mongodb.org/browse/JAVA-5736)

## Description
This PR adds bsonNamingStrategy property to BsonConfiguration to support snake_case naming strategy.

## Testing
 ran ./gradlew check


nhachicha (changes requested):
Hi @leesungbin 
Thank you for the PR! We really appreciate your contribution.

However, the current Regexp-based approach doesn’t fully account for all use cases. According to the [Kotlin language specification](https://kotlinlang.org/spec/syntax-and-grammar.html#identifiers), variable names can include Unicode characters, such as Cyrillic characters. For example, a variable name like имяПеременной would not be handled correctly.

Additionally, backtick names (e.g., ``` `my Class` ```) won’t be transformed as expected.

A possible refactor could leverage the existing implementation of SnakeCase in kotlinx.serialization. You can refer to this implementation:
[SnakeCase in kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization/blob/f9f160a680da9f92c3bb121ae3644c96e57ba42e/formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt#L142C21-L174).

Also, please ensure the tests align with the transformation rules and acronyms defined in:
[JsonNamingStrategy Transformation Rules](https://github.com/Kotlin/kotlinx.serialization/blob/f9f160a680da9f92c3bb121ae3644c96e57ba42e/formats/json/commonMain/src/kotlinx/serialization/json/JsonNamingStrategy.kt#L76-L94).

Looking forward to your updates! Let me know if you need any clarification.

leesungbin:
Hello @nhachicha, thanks for providing such detailed comments.

I have copied the `convertCamelCase` logic from the kotlinx.serialization library and used it to convert camelCase to snake_case.

Additionally, I introduced an error condition for scenarios where decoding snake_case to camelCase is ambiguous (for instance, `my_http_auth` could be interpreted as `myHttpAuth` or `myHTTPAuth`).

Please give it a look. Thanks!

nhachicha:
Thanks for iterating on this @leesungbin can you please rebase against `JAVA-5736` branch?

PR #1590, opened 2024-12-31
Improve selection criteria for srvMaxHosts

joykim1005:
### Ticket
[JAVA-4400](https://jira.mongodb.org/browse/JAVA-4400)

### Description
This PR improves the selection criteria for srvMaxHosts. The net result is that it will keep existing nodes, discard deleted nodes, and fill up to srvMaxHosts from a random shuffle of newly discovered nodes from the SRV response. It also updates the prose test to match the [spec](https://github.com/mongodb/specifications/blob/master/source/polling-srv-records-for-mongos-discovery/tests/README.md#12-new-dns-records-are-randomly-selected-srvmaxhosts--0).

### Testing
- [X] ran ./gradlew check
- [X] ran SrvPollingProseTests
- [X] ran [everygreen patch](https://spruce.mongodb.com/version/67743252483afe00076f3554/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


jyemin:
I removed @stIncMale since he is currently overloaded.  @katcharov please let me know if you think a second reviewer is needed for this.

driver-core/src/main/com/mongodb/internal/connection/DnsMultiServerCluster.java
```diff
@@ -58,13 +60,27 @@ public void initialize(final Collection<ServerAddress> hosts) {
...
+                    ServerAddress host = iterator.next();
+                    if (!hosts.contains(host)) {
+                        iterator.remove();
+                        numRemovedEndPoints++;
+                    }
+                }
                 Integer srvMaxHosts = getSettings().getSrvMaxHosts();
                 if (srvMaxHosts != null && srvMaxHosts > 0) {
                     if (srvMaxHosts < hosts.size()) {
                         List<ServerAddress> newHostsList = new ArrayList<>(hosts);
```
katcharov, line 79: This is a bit confusing, because we have a newHosts above, but they have different meanings (one means, the new hosts based on the srv response, while the other means, hosts added in the srv response relative to prior hosts). We could call this `addedHosts` or `novelHosts` or something of the sort?

driver-core/src/main/com/mongodb/internal/connection/DnsMultiServerCluster.java
```diff
@@ -58,13 +60,27 @@ public void initialize(final Collection<ServerAddress> hosts) {
...
+                        newHostsList.removeAll(existingHostList);
```
katcharov, line 80: I am not sure that this is correct. The first step according to the Jira is:

"Remove all nodes no longer present in the SRV response"

The existing host list is being removed here. This is no longer the existing host list, but rather the existing (prior) host list, minus "dead" hosts. But we are removing it, rather than removing the dead hosts.

Perhaps I am misreading? If I am not, then shouldn't the test catch this?

driver-core/src/main/com/mongodb/internal/connection/DnsMultiServerCluster.java
```diff
@@ -58,13 +60,27 @@ public void initialize(final Collection<ServerAddress> hosts) {
...
                         Collections.shuffle(newHostsList, ThreadLocalRandom.current());
-                        newHosts = newHostsList.subList(0, srvMaxHosts);
+                        newHosts = existingHostList;
+                        newHosts.addAll(newHostsList.subList(0, srvMaxHosts - numCurrentEndPoints + numRemovedEndPoints));
```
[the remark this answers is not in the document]
katcharov: The number of current will never be over the max hosts, since this setting cannot change.

driver-core/src/main/com/mongodb/internal/connection/DnsMultiServerCluster.java
katcharov: This is fine, it's being removed from the new hosts list, which will then be shuffled and added to the results.

driver-core/src/main/com/mongodb/internal/connection/DnsMultiServerCluster.java
```diff
@@ -57,17 +59,35 @@ public void initialize(final Collection<ServerAddress> hosts) {
...
-                    }
+                if (srvMaxHosts == null || srvMaxHosts <= 0 || latestSrvHosts.size() <= srvMaxHosts) {
+                    return new ArrayList<>(latestSrvHosts);
                 }
-                return newHosts;
+                List<ServerAddress> activePriorHosts = getActivePriorHosts(latestSrvHosts);
+                int numNewHostsToAdd = srvMaxHosts - activePriorHosts.size();
+                List<ServerAddress> result = getShuffledLatestSrvHosts(latestSrvHosts, activePriorHosts, numNewHostsToAdd);
+
+                return result;
```
katcharov, line 71: This is cleaner than the original, and the naming (active, latest, etc.) is clear. Nice!

driver-core/src/main/com/mongodb/internal/connection/DnsMultiServerCluster.java
```diff
@@ -57,17 +59,35 @@ public void initialize(final Collection<ServerAddress> hosts) {
...
+
+            private List<ServerAddress> getShuffledLatestSrvHosts(final Collection<ServerAddress> latestSrvHosts,
+                    List<ServerAddress> activePriorHosts, int numNewHostsToAdd) {
+                List<ServerAddress> addedHosts = new ArrayList<>(latestSrvHosts);
+                addedHosts.removeAll(activePriorHosts);
+                Collections.shuffle(addedHosts, ThreadLocalRandom.current());
+                // add the shuffled latestSrvHosts to the activePriorHosts
+                activePriorHosts.addAll(addedHosts.subList(0, numNewHostsToAdd));
+
+                return activePriorHosts;
```
katcharov, line 90: (minor) This method takes activePriorHosts through the parameter, but also returns it - this is not much of an issue here, but it is good practice to make a method pure (here, no mutations of params), if possible. You could either create a new list based on activePriorHosts and return that, or you could remove the return and rename the method to something that suggests that something is added.

driver-core/src/main/com/mongodb/internal/connection/DnsMultiServerCluster.java
```diff
@@ -57,17 +59,35 @@ public void initialize(final Collection<ServerAddress> hosts) {
...
+
+            private List<ServerAddress> getActivePriorHosts(Collection<ServerAddress> latestSrvHosts) {
+                List<ServerAddress> priorHosts = DnsMultiServerCluster.this.getCurrentDescription().getServerDescriptions().stream()
+                        .map(ServerDescription::getAddress).collect(Collectors.toList());
+                priorHosts.removeIf(host -> !latestSrvHosts.contains(host));
+
+                return priorHosts;
+            }
+
+            private List<ServerAddress> getShuffledLatestSrvHosts(final Collection<ServerAddress> latestSrvHosts,
```
katcharov, line 82: (minor) The name of the method suggests that this returns just the shuffled hosts. If that is your intent, I think you could `return addedHosts.subList(0, numNewHostsToAdd)` and then add it in the caller. You could call it something like `addShuffledHosts`, or `withShuffledHosts`, or something like this.

katcharov (approved):
LGTM, just a couple of minor optional comments

driver-core/src/main/com/mongodb/internal/connection/DnsMultiServerCluster.java
joykim1005: Got it, will update!

driver-core/src/main/com/mongodb/internal/connection/DnsMultiServerCluster.java
joykim1005: Thank you for the suggestion!

PR #1593, opened 2025-01-06
Divide ClientUpdateOptions and ClientDeleteOptions 

vbabanin:
This PR adds `BaseClientUpdateOptions` and `BaseClientDeleteOptions` as shared base interfaces for their respective client option types. While duplication of methods in sub-interfaces is still required, this approach ensures a more error-resistant API evolution. 

JAVA-5737


driver-core/src/test/unit/com/mongodb/MongoAssertions.java
```diff
@@ -0,0 +1,68 @@
...
+                        subInterface,
+                        method.getReturnType(),
+                        String.format(
+                                "Method %s in %s does not return %s. "
+                                        + "The return type must match the defining class.",
+                                method.getName(),
+                                subInterface.getName(),
+                                subInterface.getName()
+                        )
+                );
```
stIncMale, line 64: We don't really care about the methods being overridden, we only care about their return types. Thus, I propose the following change:

```suggestion
                Method method;
                try {
                    method = subInterface.getMethod(baseMethod.getName(), baseMethod.getParameterTypes());
                } catch (NoSuchMethodException e) {
                    throw fail(e.toString());
                }

                assertEquals(
                        subInterface,
                        method.getReturnType(),
                        String.format(
                                "Method `%s` in `%s` does not return `%s`. "
                                        + "The return type must match the defining class.",
                                method,
                                subInterface,
                                subInterface
                        )
                );
```

1. Because we call `getMethod` instead of `getDeclaredMethod` on a sub-interface, we are guaranteed to find the method, so we use our `fail` assertion to express that `NoSuchMethodException` can't happen.
2. We don't need to call `getName` on `method`/`subInterface`, we can rely on `toString` - that's why it is there.

stIncMale (changes requested):

driver-core/src/test/unit/com/mongodb/MongoAssertions.java
```diff
@@ -0,0 +1,68 @@
...
+    private MongoAssertions() {
+        //NOP
+    }
+
+    public static <T> void assertSubInterfaceReturnTypes(final String packageName,
+                                                         final Class<T> baseClass) {
+        Reflections reflections = new Reflections(packageName);
+        Set<Class<? extends T>> subInterfaces = reflections.getSubTypesOf(baseClass).stream()
+                .filter(Class::isInterface)
+                .collect(Collectors.toSet());
```
stIncMale, line 39: I am not sure this approach generalizes well:

1. If we do https://github.com/mongodb/mongo-java-driver/pull/1593#discussion_r1904816539, then the tests will start failing because `BaseClientDeleteOptions.collation` will be returning `BaseClientWriteModelOptions`, not `BaseClientDeleteOptions`.
2. If we want to use the same approach with classes, rather than interfaces (think of `ClientDeleteOneOptions`, `ClientDeleteManyOptions` being classes), then the `assertSubInterfaceReturnTypes` method would not be suitable for that. Though, arguably, its name explicitly mentions that it is intended only for sub-intefraces.

I currently don't have ideas on how to solve these issues without changing the method such that it accepts explicitly the sub-classes/interfaces that need to be checked.

driver-core/src/test/unit/com/mongodb/MongoAssertions.java
vbabanin: As discussed on the call:

1. We are leaving it as is to help catch mistakes when the return type in `BaseClientOptions` is specified as a subclass of `BaseClientOptions` when only one implementation exists.

2. Updated to use toString for better troubleshooting.

driver-core/src/test/unit/com/mongodb/MongoAssertions.java
vbabanin: As discussed on the call, filtering has been added to generalize it to public classes or interfaces outside of the internal package. The approach assumes that the first ancestor node in the class hierarchy tree for a public class or interface should be used for testing (passed in this method).

driver-core/src/main/com/mongodb/client/model/bulk/BaseClientDeleteOptions.java
```diff
@@ -0,0 +1,35 @@
...
+package com.mongodb.client.model.bulk;
+
+import com.mongodb.annotations.Sealed;
+import com.mongodb.client.model.Collation;
+import com.mongodb.lang.Nullable;
+import org.bson.conversions.Bson;
+
+/**
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
```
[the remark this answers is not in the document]
stIncMale: The "Instead, we should document" part is missing.

driver-core/src/main/com/mongodb/internal/client/model/bulk/ConcreteClientUpdateManyModel.java
```diff
@@ -15,26 +15,32 @@
...
+                options == null ? ConcreteClientUpdateManyOptions.MUTABLE_EMPTY : (ConcreteClientUpdateManyOptions) options);
     }
 
     @Override
-    String getToStringDescription() {
-        return "ClientUpdateManyModel";
+    public String toString() {
+        return "ClientUpdateManyModel{"
+                + "filter=" + getFilter()
+                + ", update=" + (getUpdate().isPresent() ? getUpdate() : getUpdatePipeline().orElseThrow(Assertions::fail))
```
stIncMale, line 42: With this code, we will add `Optional.toString` as the value of `update`. I propose

```suggestion
                + ", update=" + getUpdate().map(Object::toString).orElseGet(() ->
                        getUpdatePipeline().map(Object::toString).orElseThrow(Assertions::fail))
```

Formatting does not look good in the GitHub UI, but it's just fine in the code, when lines are not wrapped forcefully.

driver-core/src/main/com/mongodb/internal/client/model/bulk/ConcreteClientUpdateOneModel.java
```diff
@@ -15,26 +15,32 @@
...
+                options == null ? ConcreteClientUpdateOneOptions.MUTABLE_EMPTY : (ConcreteClientUpdateOneOptions) options);
     }
 
     @Override
-    String getToStringDescription() {
-        return "ClientUpdateOneModel";
+    public String toString() {
+        return "ClientUpdateOneModel{"
+                + "filter=" + getFilter()
+                + ", update=" + (getUpdate().isPresent() ? getUpdate() : getUpdatePipeline().orElseThrow(Assertions::fail))
```
stIncMale, line 42: The same thing I suggested in `ConcreteClientUpdateManyModel`:

```suggestion
                + ", update=" + getUpdate().map(Object::toString).orElseGet(() ->
                        getUpdatePipeline().map(Object::toString).orElseThrow(Assertions::fail))
```

driver-core/src/test/unit/com/mongodb/MongoBaseInterfaceAssertions.java
```diff
@@ -0,0 +1,70 @@
...
+    private MongoBaseInterfaceAssertions() {
+        //NOP
+    }
+
+    public static <T> void assertSubtypeReturn(final String packageName,
+                                               final Class<T> baseClass) {
+        Reflections reflections = new Reflections(packageName);
+        Set<Class<? extends T>> subtypes = reflections.getSubTypesOf(baseClass).stream()
+                .filter(aClass -> Modifier.isPublic(aClass.getModifiers()))
+                .filter(aClass -> !aClass.getPackage().getName().contains("internal"))
```
stIncMale, line 40: Let's make it more specific?

```suggestion
                .filter(aClass -> !aClass.getPackage().getName().contains(".internal"))
```

driver-core/src/test/unit/com/mongodb/client/model/bulk/BaseClientDeleteOptionsTest.java
```diff
@@ -0,0 +1,28 @@
...
+package com.mongodb.client.model.bulk;
+
+import com.mongodb.MongoBaseInterfaceAssertions;
+import org.junit.jupiter.api.Test;
+
+class BaseClientDeleteOptionsTest {
+
+    @Test
+    void testAllSubInterfacesOverrideMethods() {
+        MongoBaseInterfaceAssertions.assertSubtypeReturn("com.mongodb", BaseClientDeleteOptions.class);
```
stIncMale, line 26: Do we really need to pass `"com.mongodb"` here? Maybe we could hardcode it in `assertSubtypeReturn`, as we will never pass anything else.

stIncMale (changes requested):

stIncMale (approved):

PR #1595, opened 2025-01-07
Rename Vector class names.

vbabanin:
Primary reviewer: @katcharov 
Secondary reviewer (optional): @stIncMale 

JAVA-5710


bson/src/main/org/bson/BinaryVector.java
```diff
@@ -33,10 +33,10 @@
  * @see BsonBinary
  * @since 5.3
  */
-public abstract class Vector {
+public abstract class BinaryVector {
```
katcharov, line 36: Should update Vectors in the javadoc to Binary Vectors (and other classes, in case they mention Vector in isolation?)

```Binary Vectors are densely packed arrays of numbers, all the same type, which are stored efficiently in BSON using the BSON Binary Subtype 9 format. This class supports multiple vector {@link DataType}'s and provides static methods to create vectors.```

The first sentence of the javadoc and second paragraph seem partly redundant, perhaps merge together, as in the above?

katcharov (approved):
LGTM! I did not confirm all individual changes, as I believe the rename is fully automated IDE refactoring.

bson/src/main/org/bson/BinaryVector.java
vbabanin: Good point, updated! Thanks!

katcharov (approved):

PR #1598, opened 2025-01-09
Make `AbstractClientUpdateModel` implement `ClientWriteModel`

stIncMale:
JAVA-4586


driver-core/src/main/com/mongodb/internal/client/model/bulk/AbstractClientUpdateModel.java
```diff
@@ -26,7 +26,7 @@
 /**
  * This class is not part of the public API and may be removed or changed at any time.
  */
-public abstract class AbstractClientUpdateModel<O extends AbstractClientUpdateOptions> {
+public abstract class AbstractClientUpdateModel<O extends AbstractClientUpdateOptions> implements ClientWriteModel {
```
stIncMale, line 29: `AbstractClientDeleteModel` does this, so should `AbstractClientUpdateModel`. This way it's much easier to explore the type hierarchy from `ClientWriteModel` down to concrete types.

PR #1605, opened 2025-01-13
Add in operator to Atlas Search

joykim1005:
### Ticket
[JAVA-5744](https://jira.mongodb.org/browse/JAVA-5744)

### Description
There are several Atlas Search query operators that are not implemented with first class support in the Java driver. This PR adds the `in` operator to Atlas Search.

### Testing
- [X] ran `./gradlew check`
- [ ] ran all Evergreen patch


joykim1005:
`com.mongodb.client.ClientSideEncryptionE…estExternal_withExternalKeyVault__false` is failing and is unlikely caused by this PR and is happening in others as well.

katcharov (changes requested):

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +300,121 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+    /**
+     * Returns a {@link SearchOperator} that searches for an array of values at the given path and returns documents where the value of
+     * the field equals any value in the specified array.
+     *
+     * @param path The indexed field to be searched.
+     * @param values The non-empty values to search for. Value can be either a single value or an array of values of only one of the supported BSON types and can't be a mix of different types.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/in/ in operator
+     */
+    static <T> InSearchOperator in(final FieldSearchPath path, final Iterable<? extends T> values) {
```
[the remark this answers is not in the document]
joykim1005: I was basing the implementation off of Valentin's comment when we talked offline: 

"Also, we don't need overloads for in that take a single value. Instead, we need overloads that take Iterable<? extends T> and overloads that take T, T..., as the latter also allow passing a single value. It will be similar to SearchOperator.near accepting either Iterable<? extends FieldSearchPath> or FieldSearchPath, FieldSearchPath... "

Would it be better if I just return a new `SearchConstructibleBsonElement` in each of the overloads and remove `static <T> InSearchOperator in(final FieldSearchPath path, final Iterable<? extends T> values)`?

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +300,121 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+     */
+    static InSearchOperator in(final FieldSearchPath path, final UUID value, final UUID... values) {
+        List<BsonBinary> bsonValues = new ArrayList<>();
+        bsonValues.add(new BsonBinary(value));
+        if (values != null) {
+            for (UUID uuid : values) {
+                bsonValues.add(new BsonBinary(uuid));
+            }
+        }
+        return in(notNull("path", path), notNull("value", bsonValues));
```
stIncMale, line 377: `bsonValues` is guaranteed to be not `null`, we should not check it with `notNull("value", bsonValues)`.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +300,121 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+     * @param values More fields to be searched.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/in/ in operator
+     */
+    static InSearchOperator in(final FieldSearchPath path, final UUID value, final UUID... values) {
+        List<BsonBinary> bsonValues = new ArrayList<>();
+        bsonValues.add(new BsonBinary(value));
+        if (values != null) {
+            for (UUID uuid : values) {
+                bsonValues.add(new BsonBinary(uuid));
```
stIncMale, line 374: Do we actually have to "manually" convert `UUID` into `BsonBinary` here, given that we have `org.bson.codecs.UuidCodec`?

stIncMale: If this was done only to make the `CodecConfigurationException: The uuidRepresentation has not been specified, so the UUID cannot be encoded.` exception disappear when running `SearchOperatorTest.in`, then I don't think it's a good reason. If encoding UUID requires an explicitly set representation, then instead of calling `InSearchOperator.toBsonDocument()` in the test, we should call `toBsonDocument(Document.class, CodecRegistries.withUuidRepresentation(Bson.DEFAULT_CODEC_REGISTRY, UuidRepresentation.STANDARD))`.

stIncMale: The complete change I am proposing with regard to this method and its test is https://github.com/stIncMale/mongo-java-driver/commit/42da7affff1c1c432a5e36026417ee9a53847098.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +300,121 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+     * Returns a {@link SearchOperator} that searches for an array of values at the given path and returns documents where the value of
+     * the field equals any value in the specified array.
+     *
+     * @param path The indexed field to be searched.
+     * @param value The objectId value to search for.
+     * @param values More fields to be searched.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/in/ in operator
+     */
+    static InSearchOperator in(final FieldSearchPath path, final ObjectId value, final ObjectId... values) {
```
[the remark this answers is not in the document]
stIncMale: Let's check if the server actually allows passing `value: []`.

- If it does not, then following the approach taken in the rest of the search building API, we should introduce `in(final FieldSearchPath path, final ObjectId value, final ObjectId... values)`. We may follow that with adding another thing to be re-reviewed to [JAVA-5752: Re-visit some aspects of the Atlas search API design before moving it out of beta
](https://jira.mongodb.org/browse/JAVA-5752).
- If the server indeed allows passing `value: []`, then we should rather introduce `in(final FieldSearchPath path, final ObjectId... values)`.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +300,121 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+    static <T> InSearchOperator in(final FieldSearchPath path, final Iterable<? extends T> values) {
+        Iterator<T> valueIterator = (Iterator<T>) notNull("values", values).iterator();
+        isTrueArgument("values must not be empty", valueIterator.hasNext());
+        T firstValue = valueIterator.next();
+
+        Iterator<T> typeIterator = (Iterator<T>) notNull("values", values).iterator();
+        while (typeIterator.hasNext()) {
+            Object element = typeIterator.next();
+            isTrueArgument("values must be of same type", firstValue.getClass().isInstance(element));
+        }
```
stIncMale, line 412: I don't think we should iterate `values` here to check that they are of the same type as `firstValue`, because [_"Where possible, depend on server to return errors
"_](https://github.com/mongodb/specifications/blob/master/source/driver-mantras.md#where-possible-depend-on-server-to-return-errors).

If we see the the search building API currently violating that mantra somewhere else, we may add one more item to re-visit to [JAVA-5752: Re-visit some aspects of the Atlas search API design before moving it out of beta
](https://jira.mongodb.org/browse/JAVA-5752).

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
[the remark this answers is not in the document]
katcharov: It does not allow. See https://search-playground.mongodb.com/tools/code-sandbox/snapshots/6793c3ebe68ae0bae860e843

Added comment to 5752

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +300,121 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+     * Returns a {@link SearchOperator} that searches for an array of values at the given path and returns documents where the value of
+     * the field equals any value in the specified array.
+     *
+     * @param path The indexed field to be searched.
+     * @param value The boolean value to search for.
+     * @param values More fields to be searched.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/in/ in operator
+     */
+    static InSearchOperator in(final FieldSearchPath path, final Boolean value, final Boolean... values) {
```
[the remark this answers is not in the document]
katcharov: `"in.value[0]" value type cannot be null.`
`"in.value" value type cannot be null.`

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
katcharov: Applied

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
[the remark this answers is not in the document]
katcharov: Updated

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +300,117 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+     */
+    static InSearchOperator in(final FieldSearchPath path, final boolean value, final boolean... values) {
+        Assertions.notNull("values", values);
+        List<BsonValue> list = new ArrayList<>();
+        list.add(new BsonBoolean(value));
+        for (boolean n : values) {
+            list.add(new BsonBoolean(n));
+        }
+        BsonArray bsonArray = new BsonArray(list);
+        return in(notNull("path", path), bsonArray);
```
stIncMale, line 321: - Despite the API not allowing arrays to be `null`, we don't enforce that with runtime checks in other places of the search API (obviously, a mistake). If we want to start doing that, we should do that everywhere. Or, we may keep not enforcing that.
- `Assertions.notNull` is called as just `notNull` everywhere else in `SearchOperator.java`.
- There is no need for the intermediate `List<BsonValue>`.
- `n` is a surprising variable name in this case. Let's use `v`, for example?

```suggestion
        BsonArray bsonArray = new BsonArray();
        bsonArray.add(new BsonBoolean(value));
        for (boolean v : values) {
            bsonArray.add(new BsonBoolean(v));
        }
        return in(notNull("path", path), bsonArray);
```

driver-core/src/test/unit/com/mongodb/client/model/search/SearchOperatorTest.java
```diff
@@ -581,6 +591,96 @@ void near() {
...
+        UUID uuid = UUID.randomUUID();
+        assertAll(
+                () -> assertThrows(IllegalArgumentException.class, () ->
+                        // paths must not be empty
+                        SearchOperator.in(null, true)
+                ),
+                () -> assertThrows(IllegalArgumentException.class, () ->
+                        // values should be of one supported BSON types and can't be a mix of different types
+                        SearchOperator.in(fieldPath("fieldName1"), asList(true, "value"))
+                ),
```
stIncMale, line 606: This test does not pass. It can't, unless we inspect every element of the `values` argument at run time, which no longer happens as of 65840ebcafc785839823775eda11a8d11dd2ea1c (I don't think we should do such a check).

stIncMale (changes requested):
The last reviewed commit is 65840ebcafc785839823775eda11a8d11dd2ea1c. I did not fully review all commits before it.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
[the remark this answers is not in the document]
stIncMale: Thank you for checking!

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
katcharov: 1. Since it is not a potential design decision, but a bug (I say this because I agree it is an obvious mistake), to not nullcheck varargs, I think we should not remove the fix from this method, just because it is present everywhere else. I will leave it, and I have added a comment to the tracking ticket.
2. ok
3. I think the primary reason for this is performance, which I'm not sure matters much here. But, I have changed it here and elsewhere in the codebase for consistency.
4. ok

stIncMale: > I think the primary reason for this is performance, which I'm not sure matters much here.

That's not the objection. There was unnecessary code that improved neither readability, nor performance, nor anything else as far as I can tell. We should not have such code.

I don't know how you found that pattern also in ``BsonArrayCodec.java`, but its cool that you did 👍

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -300,6 +299,117 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+        return in(notNull("path", path), concat(notNull("value", value), values));
+    }
+
+    /**
+     * Returns a {@link SearchOperator} that searches for documents where the
+     * value or array of values at a given path contains any of the specified values
+     *
+     * @param path The indexed field to be searched.
+     * @param values The non-empty values to search for. Value can be either a single value or an array of values of only one of the supported BSON types and can't be a mix of different types.
+     * @param <T> the type.
```
stIncMale, line 398: `T` is a type parameter, so it's always "the type". I think, the documentation should answer the question "the type of what?" For example,

```suggestion
     * @param <T> the type of elements in {@code values}.
```

stIncMale (changes requested):
The last reviewed commit is d0dfa2b2de7c7bbe8eb055425405449c96c13707.

stIncMale (approved):

PR #1606, opened 2025-01-14
Add equals operator to Atlas Search

joykim1005:
### Ticket
[JAVA-5729](https://jira.mongodb.org/browse/JAVA-5729)

### Description
There are several Atlas Search query operators that are not implemented with first class support in the Java driver. This PR adds the `equals` operator to Atlas Search.

### Testing
- [X] ran `./gradlew check`
- [ ] ran all Evergreen patch


joykim1005:
com.mongodb.client.ClientSideEncryptionE…estExternal_withExternalKeyVault__false is failing and is unlikely caused by this PR and is happening in others as well.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +299,96 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+    }
+
+    /**
+     * Returns a {@link SearchOperator} that checks whether a field matches null.
+     *
+     * @param path The indexed field to be searched.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/equals/ equals operator
+     */
+    static EqualsSearchOperator equalsNull(final FieldSearchPath path) {
```
katcharov, line 387: We could also use `isNull`, similar to `public <R extends MqlValue> BranchesIntermediary<T, R> isNull`. This would be consistent with the MQL API, but less consistent with the other equals, possibly making it harder to see in intellisense, but this might not be significant because there are not many methods.

I'm ambivalent about making this change. @stIncMale for secondary.

katcharov (commented):
Just a couple of comments - possibly no changes needed.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +299,96 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+    /**
+     * Returns a {@link SearchOperator} that checks whether a field matches a value you specify.
+     *
+     * @param path The indexed field to be searched.
+     * @param value The string value to query for.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/equals/ equals operator
+     */
+    static EqualsSearchOperator equals(final FieldSearchPath path, final String value) {
+        return new SearchConstructibleBsonElement("equals", new Document("path", notNull("path", path).toValue())
```
katcharov, line 363: Perhaps we should be allowing String (and others) to be null, which would check if the value is null. I cannot recall if the type system on the server considers types to be nullable like in Java. If they are nullable, it seems we should allow, if not, probably separate?

@stIncMale for secondary

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +299,96 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+                .append("value", notNull("value", value)));
+    }
+
+    /**
+     * Returns a {@link SearchOperator} that checks whether a field matches a value you specify.
```
katcharov, line 368: The docs from our docs page are confusing. I think we should say, `...that searches for documents where a field matches the specified value.` in all of these

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
joykim1005: I talked to Valentin offline previously, and he said that having a separate overload for a BSON type null and naming it `equalsNull` would be a better approach! I think that `isNull` might be slightly confusing due to its naming. Curious on what Valentin thinks though.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
joykim1005: I think Valentin mentioned that he prefers having a separate overload for BSON type null over having a null check for the values. I can go ahead and check if the server considers types to be nullable though!

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
katcharov: Ok, `equalsNull` sounds fine.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
katcharov: Looks like we treat null separately elsewhere. So let's continue to enforce this here.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
```diff
@@ -292,6 +299,96 @@ static GeoNearSearchOperator near(final Point origin, final Number pivot, final
...
+     *
+     * @param path The indexed field to be searched.
+     * @param value The number value to query for.
+     * @return The requested {@link SearchOperator}.
+     * @mongodb.atlas.manual atlas-search/equals/ equals operator
+     */
+    static EqualsSearchOperator equals(final FieldSearchPath path, final Number value) {
```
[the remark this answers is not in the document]
katcharov: There is another reason to use Number, which is that having two Number params would require too many methods if we needed to have all permutations of the types. In any case, Number is fine for now.

driver-core/src/main/com/mongodb/client/model/search/SearchOperator.java
stIncMale, line 387: Based on experience, MongoDB official documentation is not always exhaustive or correct. So sometimes, when I suspect that it may be missing something, or that it may not reflect the reality, I check the actual server behavior.

Could we please check whether the `equals` search operator matches not only the BSON null values, but also unset fields? That is, does the operator

```
{
    "equals": {
        "path": "job_title",
        "value": null
    }
}
```

match, for example, the empty document `{}`, which does not have the field `job_title` set (or, in other words, does not contain the `job_title` key). The best way to check this seem to be by introducing an integration test to the PR, as then we will review the code that checks the behavior, and have the answer to the question with us for as long as we have the integration test.

stIncMale: If we discover that documents with the field being not set match, then we should rename the `equalsNull` method into something like `equalsNullOrNotSet`. I don't have a preference between "not set", "unset", "missing", "absent" (we may want to see if any of these or other words are already used in the Java driver API for expressing the same), but let's not use "undefined" to avoid confusion, as there is a single-value BSON Undefined type.

katcharov: If we need to rename it, probably equalsNullOrNotExists (because of the $exists operator)

katcharov: `equals: null` returns just null, and omits non-existent fields, see [this sandbox](https://search-playground.mongodb.com/tools/code-sandbox/snapshots/67916ec26ce5459b887ede3b). (To get equals to work with strings in this sandbox, the title type in the index must be `token` rather than `string`.)

I tried, unsuccessfully, to get that $search-based test we discussed working (see [gist](https://gist.github.com/katcharov/d0c7e142446bafa3fe5c3f5d9d238fbe) for the class). I used `exists`, but even this returned no results (the results are present, and show up with an empty agg pipeline; the index is present and ready; etc.). However, a [sandbox](https://search-playground.mongodb.com/tools/code-sandbox/snapshots/67916be96ce5459b887ede37) that uses the same query, index, and data (I copied the string contents via intellij) returns the expected records. I'm using atlas cli locally, and this is the standard/recommended way to test.

It seems that we should integration test everything against atlas cli.

katcharov: There are no records because the index named "default" is used when no index is specified, and it does not exist, so no records are returned.

1. https://jira.mongodb.org/browse/CLOUDP-296469 Created, to have the server error when non-existent index is used.
2. I will write integration tests, but separately from this PR https://jira.mongodb.org/browse/JAVA-5771

stIncMale: 👍

stIncMale (approved):
The last reviewed commit is 013310aad0e4bdbea8f2560afcac55a06b80ec13. I have not fully reviewed the commits before it.

luketn:
@joykim1005 thanks for this! It'll get rid of annoying helper methods I had to add to my code for the purpose like:
```
    private static SearchOperator equals(String fieldName, Object value) {
        return SearchOperator.of(
                new Document("equals", new Document()
                        .append("path", fieldName)
                        .append("value", value)
                ));
    }
```

So much nicer to have the specific method and overloads for each supported type.

PR #1608, opened 2025-01-15
Fix test cleanup to prevent conflicts between tests.

vbabanin:
JAVA-5761

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/ClientSideEncryptionExternalKeyVaultTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/ClientSideEncryptionExternalKeyVaultTest.java
@@ -20,9 +20,11 @@ import com.mongodb.AutoEncryptionSettings;
 import com.mongodb.ClientEncryptionSettings;
 import com.mongodb.MongoClientSettings;
 import com.mongodb.MongoCredential;
+import com.mongodb.MongoNamespace;
 import com.mongodb.MongoSecurityException;
 import com.mongodb.WriteConcern;
 import com.mongodb.client.model.vault.EncryptOptions;
+import com.mongodb.client.test.CollectionHelper;
 import com.mongodb.client.vault.ClientEncryption;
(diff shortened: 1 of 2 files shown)


jyemin (approved):
Approved, but wondering whether `ClientSideEncryptionExternalKeyVaultTest` should also do some cleanup, e.g. drop the collection.

driver-sync/src/test/functional/com/mongodb/client/ClientSideEncryptionExternalKeyVaultTest.java
```diff
@@ -57,6 +59,7 @@ public class ClientSideEncryptionExternalKeyVaultTest {
     private MongoClient client, clientEncrypted;
     private ClientEncryption clientEncryption;
     private final boolean withExternalKeyVault;
+    private static final MongoNamespace NAMESPACE = new MongoNamespace("db", ClientSideEncryptionExternalKeyVaultTest.class.getName());
```
jyemin, line 62: Let's make the corresponding change in CrudProseTest while we are at it.

jyemin (changes requested):

driver-sync/src/test/functional/com/mongodb/client/ClientSideEncryptionExternalKeyVaultTest.java
vbabanin: Changed namespace. Thanks!

jyemin (approved):

jyemin:
Let me know why the collection name change had to be reverted.

vbabanin:
> Let me know why the collection name change had to be reverted.

Initially, the calculation of bytes to be subtracted from the available bytes for the message in the batch-splitting test was defined as 1122 in the spec tests. After changing the namespace, the test started failing. I reverted the change but later found a way to recalculate it. The test and namespace have now been updated accordingly.

jyemin (approved):

PR #1610, opened 2025-01-16
fix JAVA-5764

raffaeleflorio:
This PR is about the https://jira.mongodb.org/browse/JAVA-5764 issue. It changes the `encode` method by skipping the encode of a property if it's already encoded as the discriminator.


raffaeleflorio:
@rozza I've just added the test case. The `testDiscriminatorEncodedOnceWhenItIsAlsoAGetter` fails without the change because the discriminator will be encoded two times when encoding a `DiscriminatorWithModelGetter` object.

raffaeleflorio:
@rozza I also added the `testDiscriminatorEncodedOnceWhenItIsAlsoAProperty` in order to test a getter with the `readName` as the discriminationKey.

raffaeleflorio:
Hello @rozza @jyemin, what else could I improve about this PR? Thanks

bson/src/test/unit/org/bson/codecs/pojo/entities/DiscriminatorWithGetterModel.java
```diff
@@ -0,0 +1,49 @@
...
+    public DiscriminatorWithGetterModel() {
+    }
+
+    public String getDiscriminatorKey() {
+        return "discriminatorValue";
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (o == null || getClass() != o.getClass()) return false;
```
rozza:
```suggestion
        if (o == null || getClass() != o.getClass()) {
            return false;
        }
```

bson/src/test/unit/org/bson/codecs/pojo/entities/DiscriminatorWithProperty.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @BsonProperty("discriminatorKey")
+    public String getDiscriminator() {
+        return "discriminatorValue";
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (o == null || getClass() != o.getClass()) return false;
```
rozza:
```suggestion
        if (o == null || getClass() != o.getClass()){
            return false;
        }
```

rozza:
Looks good - just going to run the CI again then should be able to merge.

Thanks @raffaeleflorio

PR #1612, opened 2025-01-23
Client Bulk Write sort option.

vbabanin:
- Add sort option to updateOne.
- Add sort option to replaceOne.

JAVA-5723

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
@@ -1250,8 +1250,7 @@ public final class ClientBulkWriteOperation implements WriteOperation<ClientBulk
 
         private void encodeWriteModelInternals(final BsonWriter writer, final ConcreteClientUpdateOneModel model) {
             encodeWriteModelInternals(writer, (AbstractClientUpdateModel<?>) model);
-            ConcreteClientUpdateOneOptions options = model.getOptions();
-            options.getSort().ifPresent(value -> {
+            model.getOptions().getSort().ifPresent(value -> {
                 writer.writeName("sort");
                 encodeUsingRegistry(writer, value);
             });
--- a/driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
@@ -75,7 +75,6 @@ import com.mongodb.internal.client.model.bulk.ConcreteClientReplaceOneModel;
 import com.mongodb.internal.client.model.bulk.ConcreteClientReplaceOneOptions;
 import com.mongodb.internal.client.model.bulk.ConcreteClientUpdateManyModel;
 import com.mongodb.internal.client.model.bulk.ConcreteClientUpdateOneModel;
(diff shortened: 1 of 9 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
```diff
@@ -1938,7 +1942,13 @@ private static ClientReplaceOneOptions getClientReplaceOneOptions(final BsonDocu
     }
 
     private static ClientUpdateOneOptions getClientUpdateOneOptions(final BsonDocument arguments) {
-        return fillAbstractClientUpdateOptions(new ConcreteClientUpdateOneOptions(), arguments);
+        ConcreteClientUpdateOneOptions options = new ConcreteClientUpdateOneOptions();
+
+        ofNullable(arguments.remove("sort"))
+                .map(BsonValue::asDocument)
+                .ifPresent(options::sort);
```
katcharov, line 1949: I think we should avoid using Optional. Return values and the occasional "orElse" seems fine, but it becomes quite confusing when chaining operations are used - even in this small example, I think the following is much clearer:
```suggestion
        if (arguments.containsKey("sort")) {
            options.sort(arguments.get("sort").asDocument());
        }
```
This does not use "remove" - I think that was not intended, since it modifies arguments?

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -1247,6 +1248,15 @@ private void encodeWriteModelInternals(
             });
         }
 
+        private void encodeWriteModelInternals(final BsonWriter writer, final ConcreteClientUpdateOneModel model) {
+            encodeWriteModelInternals(writer, (AbstractClientUpdateModel<?>) model);
+            ConcreteClientUpdateOneOptions options = model.getOptions();
+            options.getSort().ifPresent(value -> {
```
katcharov, line 1254:
```suggestion
            model.getOptions().getSort().ifPresent(value -> {
```

katcharov (changes requested):
Just a couple of nits.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java
vbabanin: I see your point, and I used Optional here for conciseness, as I thought it kept the example readable. I noticed we use method chaining with streams and Optional in other parts of the codebase, so it might be worth discussing this as a broader code style topic. However, I’ve changed it to use an `if - else` for now.

Regarding `arguments.remove`, it is used because `fillAbstractClientUpdateOptions` is a shared method between `UpdateOneOptions` and `UpdateManyOptions`, and it throws an exception for unknown options. Since `sort` is not a shared option, it is removed to allow processing. We use a similar pattern here:
[UnifiedCrudHelper example](https://github.com/mongodb/mongo-java-driver/blob/01eaf33e1d34e1e5c3e32d54fed3d23f5df0db99/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedCrudHelper.java#L1756).

katcharov (approved):
LGTM!

PR #1617, opened 2025-01-29
JAVA-5776 Make KProperty<T>.path() public

jenssuhr:
As mentioned in JAVA-5776, I think this function should be public so custom functions can use the same path cache and name resolution as those provided by the driver.


rozza (approved):
LGTM

PR #1619, opened 2025-01-29
JAVA-5774 Allow all property types for Filter.regex()

jenssuhr:
While it makes some sense to only allow the `regex` function on String properties, this also precludes using it with properties that wrap the String in a value class.

To allow for that use case, I think it would make sense to relax the property type and allow anything there.

### DIFF
--- a/driver-kotlin-extensions/src/main/kotlin/com/mongodb/kotlin/client/model/Filters.kt
+++ b/driver-kotlin-extensions/src/main/kotlin/com/mongodb/kotlin/client/model/Filters.kt
@@ -434,7 +434,7 @@ public object Filters {
      */
     @JvmSynthetic
     @JvmName("regexExt")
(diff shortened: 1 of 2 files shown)


rozza:
I'm not sure I fully understand the usecase could you please add some test cases? That should help.

jenssuhr:
> I'm not sure I fully understand the usecase could you please add some test cases? That should help.

Sure, I can do that

jenssuhr:
Oops, forgot that renaming the source branch closes the PR

jenssuhr:
I've added an example for filtering on a property with a value class type, but it would also be useful for data classes with a custom codec that serialize into a string.

jenssuhr:
@rozza Fair point, I think this PR can be closed. If the `path()` function is made public (see https://jira.mongodb.org/browse/JAVA-5776), it will be easy to create a custom function for this. I agree that it could be confusing if `regex` is allowed for e.g. an `Int` property.

PR #1621, opened 2025-02-04
Evergreen: Update to use stop-orchestration.sh

rozza:
JAVA-5778

### DIFF
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -794,7 +794,7 @@ functions:
         shell: bash
         script: |
           ${PREPARE_SHELL}
-          sh ${DRIVERS_TOOLS}/.evergreen/stop-orchestration.sh || true
+          bash ${DRIVERS_TOOLS}/.evergreen/stop-orchestration.sh || true
           rm -rf $DRIVERS_TOOLS || true
 
   "fix absolute paths":
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -794,15 +794,7 @@ functions:
         shell: bash
         script: |
           ${PREPARE_SHELL}
-          cd "$MONGO_ORCHESTRATION_HOME"
-          # source the mongo-orchestration virtualenv if it exists
-          if [ -f venv/bin/activate ]; then
-            . venv/bin/activate
-          elif [ -f venv/Scripts/activate ]; then
-            . venv/Scripts/activate
-          fi
-          mongo-orchestration stop || true
-          cd -
+          sh ${DRIVERS_TOOLS}/.evergreen/stop-orchestration.sh || true
           rm -rf $DRIVERS_TOOLS || true
 
   "fix absolute paths":


rozza:
@NathanQingyangXu you may need this update for mongodb/mongo-hibernate as well.

.evergreen/.evg.yml
```diff
@@ -794,15 +794,7 @@ functions:
...
-          cd "$MONGO_ORCHESTRATION_HOME"
-          # source the mongo-orchestration virtualenv if it exists
-          if [ -f venv/bin/activate ]; then
-            . venv/bin/activate
-          elif [ -f venv/Scripts/activate ]; then
-            . venv/Scripts/activate
-          fi
-          mongo-orchestration stop || true
-          cd -
+          sh ${DRIVERS_TOOLS}/.evergreen/stop-orchestration.sh || true
```
jyemin, line 797: We use `bash` everywhere else.  Should we do the same here?

.evergreen/.evg.yml
```diff
@@ -794,15 +794,7 @@ functions:
...
         script: |
           ${PREPARE_SHELL}
-          cd "$MONGO_ORCHESTRATION_HOME"
-          # source the mongo-orchestration virtualenv if it exists
-          if [ -f venv/bin/activate ]; then
-            . venv/bin/activate
-          elif [ -f venv/Scripts/activate ]; then
-            . venv/Scripts/activate
-          fi
-          mongo-orchestration stop || true
```
katcharov, line 804: `bash: line 20: mongo-orchestration: command not found`

From the linked log, it looks like this did not cause a system failure, presumably because of the `|| true`. Would there be a problem if we removed this from this command?

katcharov (changes requested):

.evergreen/.evg.yml
rozza: yeah that stops the system failure error that would happen without it.

.evergreen/.evg.yml
rozza:
```suggestion
          bash ${DRIVERS_TOOLS}/.evergreen/stop-orchestration.sh || true
```

rozza: Done!

.evergreen/.evg.yml
rozza: To clarify - we could remove the `|| true` part but if the command fails then there is no more clean up. Which I think is the reason it was added in the first place.

jyemin: Yes, iirc it's because in some variants we don't start mongo orchestration but we always attempt to stop it.

NathanQingyangXu: The reason we always stop it is because if we put it in task which starts mongo-orchestration, it might be skipped if the testing failed; to avoid that ugliness, we put it into the `clearup` post step to ensure it is run always, even for `static-analysis` task.

NathanQingyangXu: In `mongo-hibernate` project, a separate post step is created (rather than as a step in `clearup`), so no need to use the `|| true` hack.

katcharov: @rozza I see. If this is corresponds only to `"bootstrap mongo-orchestration"`, then perhaps we could set an env var in that func, and then check it here? I think we can ensure that we `rm -rf` (which is the last part of cleanup) with the following, though it's a bit long:

```

if [ -n "$MONGO_ORCHESTRATION_STARTED" ]; then
  MONGO_ORCHESTRATION_OUTPUT=$(mongo-orchestration stop)
  MONGO_ORCHESTRATION_EXIT_CODE=$?
  echo "$MONGO_ORCHESTRATION_OUTPUT"
fi

cd -
rm -rf "$DRIVERS_TOOLS"

if [ $MONGO_ORCHESTRATION_EXIT_CODE -ne 0 ]; then
  exit $MONGO_ORCHESTRATION_EXIT_CODE
fi
```

(Don't know if I have the `cd -` in the right place.)

Not sure if this is worth it - up to you. LGTM.

katcharov (approved):

.evergreen/.evg.yml
NathanQingyangXu: there must be some elegant way for sure; but conservativeness usually reigns in devops process.

I created my PR for `mongo-hibernate` at https://github.com/mongodb/mongo-hibernate/pull/36. I assigned @rozza as one reviewer. Thanks for your reminder.

PR #1623, opened 2025-02-05
JAVA-5781 Add env var override

katcharov:
JAVA-5781

Fixes tests breaking due to `java.lang.reflect.InaccessibleObjectException`

### DIFF
--- a/driver-core/src/test/functional/com/mongodb/client/WithWrapper.java
+++ b/driver-core/src/test/functional/com/mongodb/client/WithWrapper.java
@@ -32,7 +32,7 @@ public interface WithWrapper {
 
     default WithWrapper withEnvironmentVariable(final String name, @Nullable final String value) {
         return runnable -> {
-            Map<String, String> innerMap = getEnvInnerMap();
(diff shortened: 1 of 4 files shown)


driver-core/src/test/functional/com/mongodb/client/WithWrapper.java
```diff
@@ -66,15 +64,6 @@ default WithWrapper withSystemProperty(final String name, final String value) {
...
-            Map<String, String> env = getenv();
-            Field field = env.getClass().getDeclaredField("m");
-            field.setAccessible(true);
-            @SuppressWarnings("unchecked")
-            Map<String, String> result = (Map<String, String>) field.get(env);
-            return result;
-        } catch (IllegalAccessException | NoSuchFieldException e) {
-            throw new RuntimeException(e);
-        }
+        return FaasEnvironmentAccessor.getFaasEnvMap();
```
jyemin, line 67: This looks good.  Only change i would make is to inline this method, since the name no longer makes sense.

jyemin (changes requested):

jyemin (approved):

PR #1627, opened 2025-02-07
JAVA-5736 Add bsonNamingStrategy option to support snake_case

nhachicha:
Fixes Java-5736 
This PR adds some refactoring to https://github.com/mongodb/mongo-java-driver/pull/1589

### DIFF
--- a/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonConfiguration.kt
+++ b/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonConfiguration.kt
@@ -31,4 +31,9 @@ public data class BsonConfiguration(
     val encodeDefaults: Boolean = true,
     val explicitNulls: Boolean = false,
     val classDiscriminator: String = "_t",
+    val bsonNamingStrategy: BsonNamingStrategy? = null
 )
+
+public enum class BsonNamingStrategy {
+    SNAKE_CASE,
+}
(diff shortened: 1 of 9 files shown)


bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecTest.kt
```diff
@@ -1184,6 +1221,7 @@ class KotlinSerializerCodecTest {
         serializersModule: SerializersModule = defaultSerializersModule,
         configuration: BsonConfiguration = BsonConfiguration()
     ): T {
+        println("Deserializing: ${value.toJson()}")
         val codec = KotlinSerializerCodec.create(T::class, serializersModule, configuration)!!
```
rozza, line 1225:
```suggestion
        val codec = KotlinSerializerCodec.create(T::class, serializersModule, configuration)!!
```

bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/BsonConfiguration.kt
```diff
@@ -31,4 +31,9 @@ public data class BsonConfiguration(
     val encodeDefaults: Boolean = true,
     val explicitNulls: Boolean = false,
     val classDiscriminator: String = "_t",
+    val bsonNamingStrategy: BsonNamingStrategy? = null
 )
+
+public enum class BsonNamingStrategy {
```
rozza, line 37: Does this need docs?

rozza (changes requested):
Is there a way to hook directly to use `serialNameForJson` then this could support all strategies for naming.

If not then we should add a section to `THIRD-PARTY-NOTICES` for the `convertCamelCase` method.

rozza (approved):

PR #1628, opened 2025-02-10
Eliminate temporary buffer allocations during bson serialization

Edarke:
Fixes: [JAVA-5786](https://jira.mongodb.org/browse/JAVA-5786)

* Override `Outputbuffer.toByteArray()` to avoid extraneous array copy
* Modify `BsonBinaryWriter` to call `OutputBuffers.writeObjectId` so implementations can opt to use `putToByteBuffer` rather than allocating a temporary byte[]
* Wrap array in ByteBuffer so we can write numeric values 8 bytes at a time

**Motivation**

Atlas Search uses the Java driver to communicate with the Mongodb Server. We serialize up to ~300k documents per batch, primarily consisting of ObjectID and numbers. For these large batches, serialization has measurable overhead. This PR proposes some simple optimizations for `BasicOutputBuffer` without altering the API.

[JMH Benchmarks](https://gist.github.com/Edarke/2ce0750feca70c668a571175134cca9a)
| Benchmark         	| Main  (ops/ms) 	| GC    (B/op) 	| This PR  (ops/ms) 	| GC    (B/op) 	|
|-------------------	|----------------	|------------	|:---------------:	|------------	|
| toByteArray_large 	|          0.907 	| 26,777,857 	|          2.269  	| 13,388,928 	|
| toByteArray_small 	|        110,515 	|        160 	|         184,816 	|         80 	|
| encodeId          	|         23,774 	|        134 	|          32,696 	|        104 	|
| write_id          	|         28,255 	|        168 	|          36,467 	|         80 	|
| write_numbers     	|         16,327 	|         80 	|          18,125 	|         80 	|


jyemin:
@Edarke please open a Jira issue in the JAVA project for this PR, as per our contribution guide.

Thanks!

bson/src/main/org/bson/io/BasicOutputBuffer.java
```diff
@@ -35,6 +38,12 @@ public class BasicOutputBuffer extends OutputBuffer {
     private byte[] buffer;
     private int position;
 
+    /**
```
jyemin, line 41: It seems error-prone (and confusing) to add this field without completely replacing use of the two existing fields with the new one.  For instance, you have to always remember to use absolute position when writing to `buf`, so that its own position is ignored in favor of `position` field.

What do you think about getting rid of `buffer` and `position` fields entirely?

Edarke: Agreed. Updated in [0385da6](https://github.com/mongodb/mongo-java-driver/pull/1628/commits/0385da6e479a135bc844d9292d15a9f9ac4f093d)

vbabanin (commented):
Looks good! I just have a few minor comments.

bson/src/main/org/bson/io/BasicOutputBuffer.java
```diff
@@ -58,50 +66,76 @@ public BasicOutputBuffer(final int initialSize) {
...
     public void writeByte(final int value) {
         ensureOpen();
 
         ensure(1);
-        buffer[position++] = (byte) (0xFF & value);
+        buffer.put((byte) (0xFF & value));
     }
 
     @Override
     protected void write(final int absolutePosition, final int value) {
```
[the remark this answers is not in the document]
Edarke: Good catch, updated in 9a187bd533f57afa1b8ad3ab7b5c963d8911e9df

vbabanin (approved):
LGTM! Thank you for the improvement, @Edarke!

PR #1629, opened 2025-02-11
JAVA-5788 Improve ByteBufferBsonOutput::writeCharacters

franz1981:
Fixes [JAVA-5788](https://jira.mongodb.org/browse/JAVA-5788)

While running few benchmarks vs https://github.com/quarkusio/quarkus-super-heroes/blob/main/README.md using native image (graal CE) we found

image

`writeCharacters` heap-based code path to be bloated by:
- many not inlined calls: which Graal CE should be blamed since it has a very simple and naive inliner
- an excessive usage of logic to handle the current buffer to be used to write chars into
- zero loop unrolling due to the complexity of the main loop; since native image cannot profile branches cannot "shape" the assembly to grant latin to be a fast path

This patch is manually inlining the fast paths for both latin chars and heap buffers, assuming them to have enough available space.
This should help immensely JVM mode as well (i.e. JDK Hotspot) - so if there's any JMH bench I can re-use or run to verify it would be great.

NOTE: i could change the assumption which make the optimization to be triggered to be more "elastic" and use the heap buffers (byte[]) remaining capacity and perform checkpoints to refill it, too, but it would make the logic way more complex - still more elastic. 
Let me know what you prefer :pray:

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
@@ -23,6 +23,7 @@ import org.bson.io.OutputBuffer;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


franz1981:
@jyemin I see I have to open an issue there but I need a bit of help to find if there is already a JMH benchmark which already cover this; if there is not I can provides a new one myself

franz1981:
@jyemin I have just verified that since we are using the Netty driver as well, the core path there will uses (direct) Netty buffers too, so I think I will expand this optimization to include them, checking which optimization path is more appropriate there e g.
- branchless ASCII check with 8 chars (using 2 longs storing 4 UTF-16 chars each)
- branchless zero check leveraging ASCII bytes 
- batch buffer writes

franz1981:
I have added an extra commit to improve the case for the Netty buffers as well.
I can't modify the original JIRA, to explain and include such, how I can do it?

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
```diff
@@ -273,6 +276,161 @@ public void close() {
...
+        // MSB is 1 iff the byte is 0x00, 0 otherwise
+        if (tmp != 0) {
+            // there's some 0x00 in the word
+            int firstZero = Long.numberOfTrailingZeros(tmp) >> 3;
+            throw new BsonSerializationException(format("BSON cstring '%s' is not valid because it contains a null character "
+                    + "at index %d", str, i + firstZero));
+        }
+    }
+
+    private int writeCharactersOnNettyByteBuf(String str, boolean checkForNullCharacters, ByteBuf buf) {
```
franz1981, line 317: I have tried hard to reduce the burden of Netty accessibility checks and the CPU cost of writing byte per byte with this, to not mention amortize the cost of branch mispredict for less common cases (non ASCII String).

franz1981:
I have performed some JMH benchmarks and the Netty approach, despite is way better than the original code, is still much slower than the plain array case, byte per byte; I will experiment a bit more by using some Netty internalNioBuffer's API - although, since we create a Netty swapped wrapper (not needed actually) it could not work

franz1981:
I've added ff360c0c2752c7fdc84b2f60fa9981d7e44c4c31 to show what means using the internal NIO buffer from Netty
I'll post some JMH bench results soon - although native image CE won't benefit a lot from this since NIO ByteBuffer(s) have some "wide" enough OOP hierarchy which can still make single byte access to be too costly - which is something the "batchy" version instead was saving.

franz1981:
@jyemin any chance I can help moving this forward? Anything I can help with? 🙏 
Using this has a very large performance impact to both netty and non-netty users...

jyemin:
Hi @franz1981 we're actually working on a larger performance epic and one of the areas we are investigating is the `writeCharacters` method.  @vbabanin will follow up soon.

franz1981:
thanks @jyemin and I'm happy (or any in my team) to help - since i'm a Netty commiter, whatever I can help with re my knowledge there I'll do it :pray:

vbabanin:
Hi @franz1981, apologies for the late response. Thank you for the PR - i like the ideas! We’ve been working on similar improvements as part of our performance optimization investigations.

We added additional tests to guard against regressions and included enhancements for multi-byte encoding to reduce overhead from method calls and bounds checks. I’ll be proceeding with that implementation in https://github.com/mongodb/mongo-java-driver/pull/1651 with `ByteBuf` abstraction, that could later be iterated on.

That said, your Netty-specific optimizations go beyond what we’ve implemented so far and are definitely something we want to explore. 

We don’t yet have a Netty baseline to evaluate their impact, but I’ve added them to our driver-benchmark suite, currently under review in https://github.com/mongodb/mongo-java-driver/pull/1629. Once that’s merged and we can measure those changes, I’ll revisit the Netty-specific optimizations and follow up here.

One concern I have is that `internalNioBuffer` is documented as 
> Internal use only: Exposes the internal NIO buffer.
We aim to avoid using non-public APIs, as there’s no guarantee they’ll remain stable, and we aim to preserve compatibility for our users.

Regarding JMH: we don’t have any JMH benchmarks set up for this area. Instead, we use a custom benchmarking framework shared across drivers to measure performance. You could consider using `BenchmarkSuite` from there and running it locally to verify the performance of the `Netty` improvements. To enable `Netty`, configure `NettyTransportSettings` via `MongoClientSettings` on `MongoClient` created in `Benchmark` class.

franz1981:
@vbabanin 

It's fair enough related the internal NIO api, although (I am a Netty committer) I know for sure this won't go away for 4.2 as well (which means other N years before Netty 5 which have a much different Buffer API regardless).
I still suggest to use the ByteBuf::get(index) API since it will cut further the cost of moving the reader/writer offsets and further improving performance 

Said that, I have sent some time ago #1630 which improve another "hot spot" found in production, so please check if it fits

vbabanin:
Hi @franz1981, apologies for the late follow-up. The heap/array `writeCharacters` path was addressed in #1651, which is now on `main`.

Now that we have benchmarks using Buffer Provider for Netty, we tested adding a Netty-specific ASCII fast-path in `writeCharacters` and compared two approaches:

1. **With `internalNioBuffer`**  - unwraps the Netty buffer and writes ASCII bytes via the internal NIO buffer using indexed puts **nioBuffer.put(index, byte)**(the approach from this PR)  
2. **Without `internalNioBuffer`** - writes ASCII bytes directly on the Netty buffer using the public API **nettyBuf.setByte(index, value)**.

Head-to-head benchmark results (Netty transport, insert workloads):

| Benchmark                             | internalNioBuffer (ops/s) | Public API (ops/s) | Change  |
|---------------------------------------|---------------------------:|-------------------:|--------:|
| Large doc Client BulkWrite insert     |                      63.87 |              64.39 |  +0.81% |
| Large doc Collection BulkWrite insert |                      85.93 |              84.52 |  -1.64% |
| Large doc bulk insert                 |                      70.27 |              69.90 |  -0.53% |
| Large doc insertOne                   |                      73.57 |              73.36 |  -0.29% |
| Small doc Client BulkWrite insert     |                      30.84 |              31.32 |  +1.54% |
| Small doc Collection BulkWrite insert |                      34.98 |              35.11 |  +0.36% |
| Small doc bulk insert                 |                      30.47 |              30.46 |  -0.01% |
| Small doc insertOne                   |                       1.63 |               1.65 |  +1.35% |

Across these benchmarks, the two approaches perform very similarly; the differences are within typical run-to-run variance (CV 1–7%).

We then compared the Netty ASCII fast-path (using ByteBuff public API) against current `main` (where Netty buffers have no ASCII fast-path):

| Benchmark                             | Main (ops/s) | With fast-path (ops/s) | Change   |
|---------------------------------------|-------------:|------------------------:|---------:|
| Large doc bulk insert                 |        62.03 |                   71.27 | +14.89%  |
| Large doc insertOne                   |        64.87 |                   73.94 | +13.99%  |
| Large doc Client BulkWrite insert     |        57.91 |                   64.04 | +10.59%  |
| Small doc Collection BulkWrite insert |        33.51 |                   35.47 |  +5.83%  |
| Small doc bulk insert                 |        29.00 |                   30.64 |  +5.66%  |
| Small doc Client BulkWrite insert     |        29.96 |                   31.38 |  +4.74%  |

5/8 insert benchmarks show statistically significant improvement (z-score > 2). No regressions.

Benchmarks follow the MongoDB driver benchmarking specification (https://github.com/mongodb/specifications/blob/master/source/benchmarking/benchmarking.md), running with Netty transport on dedicated perf infrastructure. Results are mean-to-mean from 11 executions across different hosts.

Since **internalNioBuffer** has not showed a **Z-score** higher than 1 over the public API, and we generally aim to avoid using non-public APIs, I'm closing this PR. I've created JAVA-6199 to track the fast-path implementation using public APIs.

Thank you for your contribution and effort to explore performance improvements.

franz1981:
Yep, i am one of the authors of the Netty itself ASCII fast path (I have recently improved the performance of it without Unsafe, for Netty 4.2); the only downside is for native image as Netty buffers impls are way more one and native image CE is not very good to monomorphize the virtual calls paying per each byte the cost of a indirect call 😢

PR #1630, opened 2025-02-12
JAVA-5789 Implements batch ByteBuf::indexOf

franz1981:
Fixes [JAVA-5789](https://jira.mongodb.org/browse/JAVA-5789)


franz1981:
Similarly to #1629 this have some large performance impact to existing users 🙏

vbabanin:
Hi @franz1981,

As replied on the associated ticket [JAVA-5789](https://jira.mongodb.org/browse/JAVA-5789), thank you again for your contribution and for taking the time to explore performance improvements.

I really liked your suggestion and the direction it suggested. However, we ended up implementing an alternative SWAR-based solution as part of [JAVA-5842](https://jira.mongodb.org/browse/JAVA-5842). This approach resulted in approximately ~10% better performance according to our benchmarks. For reference, our benchmarks are based on the framework described in the [benchmarking specification](https://github.com/mongodb/specifications/blob/master/source/benchmarking/benchmarking.md) and implemented in the [driver-benchmarks module](https://github.com/mongodb/mongo-java-driver/tree/main/driver-benchmarks).

Given this, I’ll go ahead and close this PR as superseded by the implementation in [JAVA-5842](https://jira.mongodb.org/browse/JAVA-5842).

franz1981:
Hi @vbabanin 

I've taken a quick look at the SWAR approach you used there and it is sounds - but for the Netty has case I still suggest to leverage what is offered by ByteBuf::indexOf since within Netty I could have used bound check free (and without position change too!) getLong - which directly become as simple as a "mov" in x86.
In term of absolute performance increase, IDK if the original benchmark exercise branch misses and if the new SWAR implementation is as effective as the one I mentioned in Netty - which, accordingly to my past measurements, deliver a good 30% boost alone in Http 1.1 decoding (it is used for both line and headed name/values parsing).

Said that, many thanks for the consideration and keep the good work ❤️

PR #1635, opened 2025-02-20
Optimize writing numeric values.

vbabanin:
## Summary
This PR improves BSON encoding by replacing byte manipulations and bound checks with ByteBuffer’s optimized methods (putInt, putLong, putDouble). These methods are optimized for bulk operations and often intrinsified by the JVM using efficient CPU opcodes, leading to reduced overhead and better performance. 

## Key Changes:

-  Reduces constant factor of buffer lookups in `writeInt32(int absolutePosition, int value)` from 4 to 1 by preferring a single putInt call when capacity allows, minimizing overhead. Adds a fallback for edge cases (e.g., buffer boundaries) with O(1) buffer switches instead of repeated O(n) lookups.
- Interface Extensions: Adds `putInt(int)`, `putInt(int, int)`, `putDouble(double)`, and `putLong(long)` to `ByteBuf`, implemented in `ByteBufNIO` and `NettyByteBuf`, with no-op stubs in `CompositeByteBuf`.
- Updates `ByteBufferBsonOutput` to: Use `putInt`, `putLong`, and `putDouble` for single-call writes when capacity allows, reducing overhead.

## Performance analysis

To ensure accurate performance comparison and reduce noise, 11 versions (Comparison Versions) were aggregated and compared against a stable region of data around the Base Mainline Version. The percentage difference and z-score of the mean of the Comparison Versions were calculated relative to the Base Mainline Version’s stable region mean. 

| Test                 | Base Stable Mean (ops/sec) | Comparison Versions Mean (ops/sec) | Diff | Z-Score |
|----------------------|-----------------|------------------|-------------|---------|
| **Deep BSON Encoding**   | 47.45           | **60.15**        | **+26.8%**  | 5.02    |
| **Flat BSON Encoding**   | 104.94          | **131.79**       | **+25.6%**  | 1.33    |
| **Full BSON Encoding**   | 110.44          | **141.49**       | **+28.1%**  | 1.98    |
| **Large Doc Bulk Insert** | 46.53           | **64.59**        | **+38.8%**  | 3.44    |
| **Large Doc InsertOne**   | 50.38           | **67.20**        | **+33.4%**  | 3.07    |
| **Small Doc Bulk Insert** | 25.07           | **30.08**        | **+20.0%**  | 3.19    |

Perf analyzer results: [Link](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=67c89770b940b82855b1f6e2&selected_tab=data-table&percent_filter=0%7C%7C100&z_filter=0%7C%7C10&filter_type=Default)

JAVA-5800


franz1981:
I like this, it looks similar to my PR at https://github.com/mongodb/mongo-java-driver/pull/1629 which try writing byte per byte if there is available space to straight use the same buffer ❤️

bson/src/main/org/bson/ByteBuf.java
```diff
@@ -106,6 +106,50 @@ public interface ByteBuf  {
...
+    /**
+     * Writes the given int value into this buffer at the current position,
+     * using the current byte order, and increments the position by 4.
+     *
+     * @param b the int value to be written
+     * @return this buffer
+     * @throws java.nio.BufferOverflowException if there are fewer than 4 bytes remaining in this buffer
+     * @throws java.nio.ReadOnlyBufferException if this buffer is read-only
+     */
+    ByteBuf putInt(int b);
```
vbabanin, line 118: Given that the interface includes the warning: 

> This interface is not frozen yet, and methods may be added in a minor release, so beware implementing this yourself we are justified in adding new methods to this class.

 Therefore, it is permissible to add new methods to this class.

rozza (changes requested):
Like the changes - just a couple of nits.

jyemin (changes requested):
I ran `ByteBufferBsonOutputTest` with Coverage, and noticed that `ByteBufferBsonOutput#write(int absolutePosition, int value)` is no longer covered by test due to the optimizations that were introduced.  This method should either be covered by tests or removed (it seems like it will never be called)

Here's one way to do it: https://github.com/jyemin/mongo-java-driver/pull/new/jeff-wip

bson/src/main/org/bson/ByteBuf.java
vbabanin, line 1: [Note] As @stIncMale suggested, lets put Evolving annotation here.

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufTest.java
```diff
@@ -0,0 +1,101 @@
...
+            buffer.flip();
+            assertEquals(42.0D, buffer.getDouble());
+        } finally {
+            buffer.release();
+        }
+    }
+
+    @ParameterizedTest
+    @MethodSource("bufferProviders")
+    void shouldPutIntAtIndex(final BufferProvider provider) {
```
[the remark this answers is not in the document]
rozza: Apologies, nothing to do here.

rozza (changes requested):
Looking good - just javadocs and jeffs comment to address.

bson/src/main/org/bson/ByteBufNIO.java
```diff
@@ -160,8 +184,13 @@ public ByteBuf get(final byte[] bytes, final int offset, final int length) {
 
     @Override
     public ByteBuf get(final int index, final byte[] bytes, final int offset, final int length) {
-        for (int i = 0; i < length; i++) {
-            bytes[offset + i] = buf.get(index + i);
+        if (buf.hasArray()) {
+            System.arraycopy(buf.array(), index, bytes, offset, length);
+        } else {
+            // Fallback to per-byte copying if no backing array is available.
```
franz1981, line 190: Actually Netty ByteBufs already have optimized methods with this signature - and most of the time the buffers are off-heap

vbabanin: Just to clarify, do you mean  Netty ByteBuf's `ByteBuf.getBytes(index, bytes, offset, length)?`

I believe we’re already using that one in our `NettyByteBuf` wrapper:
```
@Override
public ByteBuf get(final int index, final byte[] bytes, final int offset, final int length) {
    proxied.getBytes(index, bytes, offset, length);
    return this;
}
```

This particular method is targeting the Java NIO `ByteBuffer` case.

vbabanin:
> I ran `ByteBufferBsonOutputTest` with Coverage, and noticed that `ByteBufferBsonOutput#write(int absolutePosition, int value)` is no longer covered by test due to the optimizations that were introduced. This method should either be covered by tests or removed (it seems like it will never be called)
> 
> Here's one way to do it: https://github.com/jyemin/mongo-java-driver/pull/new/jeff-wip

Since it's a public API method, there's still a small chance it could be used. To be safe, I added test coverage for it and marked `writeInt32` with absolute position as @Deprecated in `OutputBuffer` for now. Let me know what you think.

bson/src/main/org/bson/ByteBuf.java
vbabanin: @stIncMale I will add it in follow-up PR as we don't have this annotation in `Bson` package.

jyemin (approved):
LGTM!

PR #1642, opened 2025-03-05
Gradle modernization

rozza:
This PR comprises of 3 separate commits for the purpose of aiding reviewability. 

The commits themselves aren't designed to be standalone (they are not) but rather compose separate areas of work that updating the build has required.


rozza:
https://spruce.mongodb.com/version/67c88fbc80d1df00071d5175/tasks

rozza:
TODO: 

Compare: each projects `./gradlew dependencies` between main and this version.
Compare: each projects `pom.xml` between main and this version.
Compare: each projects `MANIFEST.MF` between main and this version.

bom/build.gradle.kts
```diff
@@ -21,3 +45,109 @@ dependencies {
...
+                     * The <scope> and <optional> tags should be omitted in BOM dependencies.
+                     * This ensures that consuming projects have the flexibility to decide whether a dependency is optional in their context.
+                     *
+                     * The BOM's role is to provide version information, not to dictate inclusion or exclusion of dependencies.
+                     */
+                    assert(it.getNode("scope") == null) {
+                        "BOM must not contain <scope> elements in dependency:\n$destination"
+                    }
+                    assert(it.getNode("optional") == null) {
+                        "BOM must not contain <scope> elements in dependency:\n$destination"
```
NathanQingyangXu, line 139: should be "BOM must not contain <optional> elements ..."?

jyemin:
> driver-legacy now has a manifest - side effect of the single publishing config

It was intentional that driver-legacy doesn't have a manifest (and also can never be a Java module).  The reason is that OSGi requires that Java packages are segregated between modules, i.e. you can't have more than module containing types in the same Java package.  And for various reasons, we were forced to do that with driver-legacy, as there are types in both core and legacy in the "com.mongodb" package (and same for "org.bson" in the bson module.

Is there a straightforward way to exclude the manifest for driver-legacy?  I know there is at least one place where our module names are referenced in buildSrc.  Maybe this is another one?

jyemin (changes requested):
I haven't reviewed the buildSrc directory yet.

buildSrc/src/main/kotlin/conventions/testing-integration.gradle.kts
```diff
@@ -0,0 +1,50 @@
...
+//
+// See:
+// https://docs.gradle.org/current/samples/sample_jvm_multi_project_with_additional_test_types.html
+plugins { id("java-library") }
+
+val integrationTest by sourceSets.creating
+
+configurations[integrationTest.implementationConfigurationName].extendsFrom(configurations.testImplementation.get())
+
+configurations[integrationTest.runtimeOnlyConfigurationName].extendsFrom(configurations.testRuntimeOnly.get())
```
NathanQingyangXu, line 29: If you are tired of `"integrationTestImplementation"(...)`, you might well replace the above two statements with

```
val integrationTestImplementation: Configuration by
configurations.getting { extendsFrom(configurations.implementation.get()) }
val integrationTestRuntimeOnly: Configuration by
configurations.getting { extendsFrom(configurations.runtimeOnly.get()) }
```

then you can use `integrationTestImplementation(...)` directly

NathanQingyangXu: nvm. Further more configs are required. maybe not worth while.

buildSrc/settings.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+plugins {
+    // Add support for `libs.versions.toml` within `buildSrc`
+    // https://github.com/radoslaw-panuszewski/typesafe-conventions-gradle-plugin
+    // https://github.com/gradle/gradle/issues/15383
+    id("dev.panuszewski.typesafe-conventions") version "0.4.1"
+}
```
NathanQingyangXu, line 21: We might well simply put the following content to this file to reuse the toml file in the main build (which seems a lightweight way without a third-party gradle plugin)
```
dependencyResolutionManagement {
    versionCatalogs {
        create("libs") {
            from(files("../gradle/libs.versions.toml"))
        }
    }
}
```

buildSrc/src/main/kotlin/conventions/Companion.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.getByType
+
+// Adds the `libs` value for use in conventions
+internal val Project.libs: LibrariesForLibs
+    get() = extensions.getByType()
+
+/** Extension function to determine if a project property has been set. */
+fun Project.buildingWith(name: String): Boolean {
+    return this.hasProperty(name) && this.property(name).toString().toBoolean()
```
NathanQingyangXu, line 28: seems the following would be better in terms of robustness (property value could still be null if it exists) and consistency with other similar usages:
```
return findProperty(name)?.toString().toBoolean()
```

bom/build.gradle.kts
```diff
@@ -21,3 +45,109 @@ dependencies {
...
+assert(!scalaVersions.isNullOrEmpty()) {
+    "Scala versions must be provided as a comma-separated list in the 'supportedScalaVersions' project property"
+}
+
+/*
+ * Apply the Java Platform plugin to create the BOM
+ * Modify the generated POM to include all supported versions of Scala for driver-scala or bson-scala.
+ */
+configureMavenPublication {
+    components.findByName("javaPlatform")?.let { from(it) }
```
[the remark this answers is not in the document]
rozza: This is a port of the existing BOM code which was in `publishing.gradle`. So it applies all java platform configurations - essentially to include all projects into this BOM.

bom/build.gradle.kts
rozza, line 139:
```suggestion
                        "BOM must not contain <optional> elements in dependency:\n$destination"
```

bson-scala/src/test/scala/org/mongodb/scala/bson/BaseSpec.scala
```diff
@@ -15,10 +15,7 @@
  */
 package org.mongodb.scala.bson
 
-import org.junit.runner.RunWith
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
-import org.scalatestplus.junit.JUnitRunner
 
-@RunWith(classOf[JUnitRunner])
```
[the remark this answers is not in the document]
rozza: Updated to use junit 5

graalvm-native-image-app/src/main/com/mongodb/internal/graalvm/NativeImageApp.java
```diff
@@ -18,7 +18,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.annotation.Nullable;
+import org.jetbrains.annotations.Nullable;
```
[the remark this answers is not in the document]
rozza: We need a dependency to be able to import the annoation. 

Previously, in the jumble of dependencies and configuration I think one of the plugins put it on the path. So it was only there implicitly. I tried added javax as a dependency for this project and fell into other issues.  Simplest solution that I could get to work was to add jetbrains annotations as a compile dependency.

buildSrc/src/main/kotlin/conventions/testing-integration.gradle.kts
rozza: Gotcha - in the actual builds we can reference them without the quotes.  However, in conventions we leave them to be lazy configurations that are resolved at runtime.

buildSrc/src/main/java/com/mongodb/doclet/DocTaglet.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.List;
+import java.util.Set;
+import javax.lang.model.element.Element;
+import jdk.javadoc.doclet.Taglet;
+
+public abstract class DocTaglet implements Taglet {
+
+    @Override
+    public Set<Location> getAllowedLocations() {
+        return new HashSet<>(asList(CONSTRUCTOR, METHOD, FIELD, OVERVIEW, PACKAGE, TYPE));
```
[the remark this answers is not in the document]
rozza: Agreed, but out of scope for this review as its just a move of directory.

buildSrc/src/main/kotlin/conventions/Companion.kt
rozza, line 28:
```suggestion
    return this.findProperty(name)?.toString()?.toBoolean() ?: false
```

rozza: Done

bson/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
+/*
```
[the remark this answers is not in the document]
rozza: In 3ead368

```git show 3ead3681110d4aa00c55fbd8dfa36435ef1a8f92 -- ./bson/build.gradle   ```

**Comment by rozza** [comment]
> > driver-legacy now has a manifest - side effect of the single publishing config
> 
> It was intentional that driver-legacy doesn't have a manifest (and also can never be a Java module). The reason is that OSGi requires that Java packages are segregated between modules, i.e. you can't have more than module containing types in the same Java package. And for various reasons, we were forced to do that with driver-legacy, as there are types in both core and legacy in the "com.mongodb" package (and same for "org.bson" in the bson module.
> 
> Is there a straightforward way to exclude the manifest for driver-legacy? I know there is at least one place where our module names are referenced in buildSrc. Maybe this is another one?

Done - See [d39ce81](https://github.com/mongodb/mongo-java-driver/pull/1642/commits/d39ce8119af90b4625fc7c5b678438fee2c81792)

File: driver-legacy/build.gradle.kts [hunk 10]
```diff
@@ -0,0 +1,55 @@
...
+        name.set("The Legacy MongoDB Driver")
+        description.set("The Legacy MongoDB Driver")
+    }
+}
+
+// Disable the manifest for driver-legacy as its not a valid OSGI package
+configureJarManifest {
+    attributes["-nomanifest"] = true
+    afterEvaluate {
+        attributes.remove("Bundle-Version")
```
jyemin, line 49: It's surprising that these need to be removed explicitly even when `-nomanifest` is specified just above.  What happens if you don't do this?  Or does `-nomanifest` not mean what I think it does?

rozza: In the publishing convention we set up some attributes outside of what the bnd plugin auto configures.

rozza: See 68689fd for an improved approach.

jyemin: I like it.

buildSrc/src/main/kotlin/conventions/detekt.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+import io.gitlab.arturbosch.detekt.Detekt
+
+// Static code analysis for Kotlin
+// https://plugins.gradle.org/plugin/io.gitlab.arturbosch.detekt
+plugins { id("io.gitlab.arturbosch.detekt") }
+
+detekt {
+    allRules = true // fail build on any finding
+    buildUponDefaultConfig = true // preconfigure defaults
+    config = rootProject.files("config/detekt/detekt.yml") // point to your custom config defining rules to run,
```
jyemin, line 27: Any reason while this is `.files` but the one below is `.file`?  If not, let's make them consistent.

buildSrc/src/main/kotlin/conventions/spotbugs.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+
+dependencies {
+    compileOnly(libs.findbugs.jsr)
+
+    testImplementation(libs.findbugs.jsr)
+}
+
+spotbugs {
+    if (!project.buildingWith("ssdlcReport.enabled")) {
+        excludeFilter.set(file("${rootProject.rootDir}/config/spotbugs/exclude.xml"))
```
jyemin, line 37: Can this use `rootProject.file` function as in `detekt.gradle.kts`?

buildSrc/src/main/kotlin/conventions/optional.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+
+// Provides the optional dependencies support eg: optionalApi, optionalImplementation
+plugins {
+    id("java-library")
+    id("maven-publish")
+}
+
+java { registerFeature("optional") { usingSourceSet(sourceSets["main"]) } }
+
+// Suppress POM warnings for the optional features (eg: optionalApi, optionalImplementation)
```
jyemin, line 26: What are these POM warnings warning about?

buildSrc/src/main/kotlin/conventions/optional.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+ */
+package conventions
+
+// Provides the optional dependencies support eg: optionalApi, optionalImplementation
+plugins {
+    id("java-library")
+    id("maven-publish")
+}
+
+java { registerFeature("optional") { usingSourceSet(sourceSets["main"]) } }
```
jyemin, line 24: Where is this "optional" feature actually defined?  It's not clear to me how this works :)

buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
```diff
@@ -0,0 +1,176 @@
...
+                    password = nexusPassword.get()
+                }
+            }
+        }
+
+        // publish to local dir, for artifact tracking and testing
+        // `./gradlew publishMavenPublicationToLocalBuildRepository`
+        maven {
+            url = uri(localBuildRepo.get())
+            name = "LocalBuild"
```
jyemin, line 72: Why does this one need a name?

buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
```diff
@@ -0,0 +1,176 @@
...
+                credentials {
+                    username = nexusUsername.get()
+                    password = nexusPassword.get()
+                }
+            }
+        }
+
+        // publish to local dir, for artifact tracking and testing
+        // `./gradlew publishMavenPublicationToLocalBuildRepository`
+        maven {
```
jyemin, line 70: How does publishing to the local repository, i.e. `~/.m2/repository`, work?  Is it this repository?  If so, I'm not clear on what `localBuildRepo` is (I've never seen a "repo" directory).

buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
```diff
@@ -0,0 +1,176 @@
...
+            url = uri(localBuildRepo.get())
+            name = "LocalBuild"
+        }
+    }
+
+    publications.create<MavenPublication>("maven") {
+        components.findByName("scala")?.let { from(it) }
```
jyemin, line 77: I'm not clear what "scala" and "java" are referring to in this context.  I would also expect to see "kotlin", so I'm even less sure what this is about.

buildSrc/src/main/kotlin/conventions/testing-base.gradle.kts
```diff
@@ -0,0 +1,107 @@
...
+            ))
+    }
+
+    // Allow testing with an alternative JDK version
+    val testJavaVersion: Int = findProperty("javaVersion")?.toString()?.toInt() ?: DEFAULT_JAVA_VERSION
+    javaLauncher.set(javaToolchains.launcherFor { languageVersion = JavaLanguageVersion.of(testJavaVersion) })
+}
+
+// Pretty test output
+testlogger {
```
jyemin, line 89: Is this new?

buildSrc/src/main/kotlin/project/base.gradle.kts
```diff
@@ -13,3 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+package project
+
+plugins { id("conventions.git-version") }
+
+group = "org.mongodb"
+
+// TODO update Github publish action.
```
jyemin, line 22: Reminder to remove this TODO after updating the GH action.  Can it be done as part of this PR?

buildSrc/src/main/kotlin/project/java.gradle.kts
```diff
@@ -0,0 +1,50 @@
...
+    id("conventions.bnd")
+    id("conventions.javadoc")
+    id("conventions.optional")
+    id("conventions.publishing")
+    id("conventions.spotbugs")
+    id("conventions.spotless")
+    id("conventions.testing-junit")
+}
+
+dependencies { "optionalApi"(libs.slf4j) }
```
jyemin, line 31: Remind me what "optionalApi" means?  I would think this is implementation, not api.

buildSrc/src/main/kotlin/project/kotlin.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+    id("conventions.spotless")
+    id("conventions.testing-integration")
+    id("conventions.testing-junit")
+}
+
+/* Compiling */
+logger.info("Compiling ${project.name} using JDK${DEFAULT_JAVA_VERSION}")
+
+kotlin {
+    explicitApi()
```
jyemin, line 39: What does this mean?

jyemin (changes requested):

buildSrc/src/main/kotlin/conventions/detekt.gradle.kts
rozza: Yes, its because config requires a `ConfigurableFileCollection`.

buildSrc/src/main/kotlin/conventions/spotbugs.gradle.kts
rozza: Done

buildSrc/src/main/kotlin/conventions/optional.gradle.kts
rozza: > Silences the compatibility warnings for the Maven publication for the specified variant. Warnings are emitted when Gradle features are used that cannot be mapped completely to Maven POM.

Essentially Gradle can support extra metadata that maven doesn't support.

buildSrc/src/main/kotlin/conventions/optional.gradle.kts
rozza: We're registering the feature here. See: https://docs.gradle.org/current/userguide/feature_variants.html for more information.

buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
rozza: Its a task thats added by the maven publish plugin. (https://docs.gradle.org/current/userguide/publishing_maven.html) 

The local build repo was added at some point to the main java driver here: [publish.gradle](https://github.com/mongodb/mongo-java-driver/blob/42b1a373017f9709a5c0ac6f2ddd90dccbb45c90/gradle/publish.gradle#L59-L61)

buildSrc/src/main/kotlin/project/base.gradle.kts
rozza: I don't know how the action works - will look.

buildSrc/src/main/kotlin/conventions/testing-base.gradle.kts
rozza: No just expanded. Here it is in main - https://github.com/mongodb/mongo-java-driver/blob/42b1a373017f9709a5c0ac6f2ddd90dccbb45c90/build.gradle#L301-L306

buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
rozza: Turns out scala is not needed / doesn't exist - I just ported it but testing shows it can be removed.

Java components are as essentially that the Jar that the java/scala/kotlin plugins create.

buildSrc/src/main/kotlin/project/java.gradle.kts
rozza: I had it as "optionalImplementation" but then it changes the scope to runtime in the pom.xml files and prior to this change it was compile. 

This keeps it the same as it has been even though I don't believe we surface and SLF4J API to userland code.

buildSrc/src/main/kotlin/project/base.gradle.kts
rozza: Done 👍

.github/workflows/bump-version.sh
```diff
@@ -9,5 +9,5 @@ fi
 FROM_VERSION=$1
 TO_VERSION=$2
 
-sed --in-place "s/version = '${FROM_VERSION}'/version = '${TO_VERSION}'/g" build.gradle
-git commit -m "Version: bump ${TO_VERSION}" build.gradle
+sed --in-place "s/version=${FROM_VERSION}/version=${TO_VERSION}/g" gradle.properties
+git commit -m "Version: bump ${TO_VERSION}" gradle.properties
```
NathanQingyangXu, line 13: Is the commit message 100% clear? 

Seems `Version: bump ${FROM_VERSION} to ${TO_VERSION}` will make the intention cystal clear.

NathanQingyangXu: nvm. Such change is not in the scope of this PR for it is legacy issue (if it is).

buildSrc/src/main/kotlin/project/kotlin.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+    testImplementation(libs.kotlin.reflect)
+    testImplementation(libs.junit.kotlin)
+    testImplementation(libs.bundles.mockito.kotlin)
+    testImplementation(libs.assertj)
+    testImplementation(libs.classgraph)
+
+    "integrationTestImplementation"(libs.junit.kotlin)
+}
+
+sourceSets["integrationTest"].kotlin.srcDir("src/integration/kotlin")
```
NathanQingyangXu, line 69: Did you intend 
```
sourceSets["integrationTest"].kotlin.srcDir("src/integrationTest/kotlin")
```
?
It seems inconsistent with the directories in `conventions/test-integration.gradle`.
Also, you did use `integrationTest` in scala.project.kts.

buildSrc/src/main/kotlin/conventions/testing-spock.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package conventions
+
+import org.gradle.api.tasks.testing.Test
+import org.gradle.kotlin.dsl.dependencies
+import org.gradle.kotlin.dsl.withType
+import project.libs
```
NathanQingyangXu, line 21: from my IDE, there are some gray import entries unused. Are they there intentionally? Why our kts spotless can't fix them?

NathanQingyangXu: found `removeUnusedImports()` is only applicable to Java, not to Kotlin.

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    java {
+        palantirJavaFormat()
+        target("src/*/java/**/*.java")
+        removeUnusedImports()
+        trimTrailingWhitespace()
+        indentWithSpaces()
+        endWithNewline()
+        licenseHeaderFile(rootProject.file("../config/mongodb.license"))
```
NathanQingyangXu, line 71: I have no difficulty adding license for Java source code within `buildSrc`, but I fail to see how to add license in other subproject, like `driver-core`. There is spotless tasks for both kotlin and scala, but none for Java.

Also, when I created a new Java class in `driver-core` and run `spotlessApply` task, ending up with no lience header added to the file. Anything is missing?

NathanQingyangXu: seems Java spotless configuration is missing from `conventions.spotless` plugin.

driver-core/build.gradle.kts
```diff
@@ -0,0 +1,99 @@
...
+ * limitations under the License.
+ */
+import ProjectExtensions.configureJarManifest
+import ProjectExtensions.configureMavenPublication
+
+plugins {
+    id("project.java")
+    id("conventions.test-artifacts")
+    id("conventions.testing-mockito")
+    id("conventions.testing-spock-exclude-slow")
```
NathanQingyangXu, line 23: so from my understanding, the above will bring in transitively  the plugin `conventions.testing-junit-vintage`, but not `conventions.testing-junit`. But it seems JUnit 5 is used extensively together with Spock (which is based on JUnit 4). Should we add `conventions.testing-junit` as well?

buildSrc/src/main/kotlin/project/kotlin.gradle.kts
rozza: The inconsistency exists in main, so it was ported here. Now that JAVA-5754 isn't going to be done. I think its worth doing this here to reduce any cognitive load in the future 👍

buildSrc/build.gradle.kts
rozza: Correct, we haven't adopted spotless for java code. JAVA-3809 tracks the adoption but has been on the backlog for 5 years now.

Note checkstyle confirms the presence of a license for java files.

buildSrc/src/main/kotlin/project/kotlin.gradle.kts
rozza: Done

buildSrc/src/main/kotlin/conventions/testing-spock.gradle.kts
rozza: Good catch - done.

driver-core/build.gradle.kts
rozza: Done :)

jyemin (commented):
Why didn't the integration test directory name change require a corresponding change in Gradle scripts?

jyemin:
Also, static-analysis task is faling now.

rozza:
@jyemin static tests will now pass - a side effect of moving integration tests. Also removed the unneeded extra source directory for kotlin integration tests from the kotlin project.

jyemin:
> Also removed the unneeded extra source directory for kotlin integration tests from the kotlin project.

Ah, so it was looking in both places, so they were found in either location.  Thanks.

jyemin (approved):
LGTM

Since the validation of the correctness of changes were largely manual, please re-validate before final merge. :)

NathanQingyangXu (approved):
Looks great and I learned a lot from the PR. Thanks.

.github/workflows/bump-version.sh
```diff
@@ -9,5 +9,5 @@ fi
 FROM_VERSION=$1
 TO_VERSION=$2
 
-sed --in-place "s/version = '${FROM_VERSION}'/version = '${TO_VERSION}'/g" build.gradle
-git commit -m "Version: bump ${TO_VERSION}" build.gradle
+sed --in-place "s/version=${FROM_VERSION}/version=${TO_VERSION}/g" gradle.properties
```
jyemin, line 12: I'm not able to run this script locally for some reason (sed error), but curious why you removed the spaces and single quotes.  Are you able to test it?

rozza: The string in `build.gradle` that was replaced looks like:

```
version = '5.4.0-SNAPSHOT'
```

Where as in `gradle.properties` it looks like:

```
version=5.4.0-SNAPSHOT
```

I tested without the `--in-place` flag thats not supported on mac osx sed eg:

```sed  "s/version=5.4.0-SNAPSHOT/version=5.4.0/g" gradle.properties```

Comment (jyemin) [reply .github/workflows/bump-version.sh:12]: Ah, I see.  Thanks!

Comment (NathanQingyangXu) [reply .github/workflows/bump-version.sh:12]: but should `-i` make more sense for it is universally supported? `--in-place` is a GNU alternative long argument but `-i` is always working (and MAC user could test it without trouble).

Comment (NathanQingyangXu) [reply .github/workflows/bump-version.sh:12]: nvm. Just found on mac we need special treatment as:
```
sed -i '' 's/STRING_TO_REPLACE/STRING_TO_REPLACE_IT/g' filename
```
then all good. no need to change.

Comment (rozza) [reply .github/workflows/bump-version.sh:12]: Even with `-i` on mac it still needs work.

I believe this will work: 
```
sed -i '' 's/version=5.4.0-SNAPSHOT/version=5.4.0/g' gradle.properties
```

However, I don't have the ability to test on both.

File: buildSrc/build.gradle.kts [hunk 28]
```diff
@@ -0,0 +1,83 @@
...
+
+plugins {
+    id("java-library")
+    `kotlin-dsl`
+    alias(libs.plugins.spotless)
+    alias(libs.plugins.detekt) apply false
+}
+
+repositories {
+    gradlePluginPortal()
```
NathanQingyangXu, line 26: This thread is only for discussion. Did you see scenario where some plugin could only be found in `gradlePluginPortal()`, not in `mavenCentral()`? I asked for I struggled on whehter I need to use `gradlePluginPortal()` in `mongo-hibernate` project as well.

bom/build.gradle.kts
```diff
@@ -21,3 +45,109 @@ dependencies {
...
+
+            existingScalaDeps.forEach {
+                val groupId: String = it.getNode("groupId")!!.text()
+                val originalArtifactId: String = it.getNode("artifactId")!!.text()
+                val artifactVersion: String = it.getNode("version")!!.text()
+
+                // Add multiple versions with Scala suffixes for each Scala-related dependency.
+                scalaVersions!!.forEach { scalaVersion ->
+                    if (scalaVersion != defaultScalaVersion) {
+                        // Replace scala version suffix
```
NathanQingyangXu, line 95: super minor. `scala version` -> `Scala version`? Don't wanna make a fuss about it. Feel free to ignore.

buildSrc/build.gradle.kts
rozza: Just following the official recommendations. I imagine any third party plugins would be on maven but any gradle specific ones might not be.

bom/build.gradle.kts
```diff
@@ -21,3 +45,109 @@ dependencies {
...
+    }
+}
+
+/*
+ * Validate the BOM file.
+ */
+tasks.withType<GenerateMavenPom> {
+    doLast {
+        pom.withXml {
+            val pomXml: Node = asNode()
```
vbabanin, line 114: It looks like the validation isn’t currently being executed, because`pom.withXml {}` runs during the configuration phase (before the POM file is written), placing it inside a `doLast` block causes it to be skipped.

One alternative is to remove `pom.WithXml` block and parse the generated POM file directly using `XmlParser`, as we did previously: 
`val pomXml: Node = XmlParser().parse(destination)`

The rest of the logic should not be affected after the change.

bom/build.gradle.kts
```diff
@@ -21,3 +45,109 @@ dependencies {
...
+            val dependenciesNode = pomXml.getNode("dependencyManagement").getNode("dependencies")
+            assert(dependenciesNode!!.children().isNotEmpty()) {
```
vbabanin, line 116: These assertions only take effect when the JVM is run with the -ea flag, which could lead to silent failures. We could replace `assert()` with `require()` to ensure the validation always runs.

PR #1646, opened 2025-03-13
Fix Netty event loop resource leak on MongoClient close.

vbabanin:
JAVA-5812

### DIFF
--- a/driver-sync/src/main/com/mongodb/client/internal/MongoClientImpl.java
+++ b/driver-sync/src/main/com/mongodb/client/internal/MongoClientImpl.java
@@ -33,8 +33,8 @@ import com.mongodb.client.MongoDatabase;
 import com.mongodb.client.MongoIterable;
 import com.mongodb.client.SynchronousContextProvider;
 import com.mongodb.client.model.bulk.ClientBulkWriteOptions;
-import com.mongodb.client.model.bulk.ClientNamespacedWriteModel;
 import com.mongodb.client.model.bulk.ClientBulkWriteResult;
+import com.mongodb.client.model.bulk.ClientNamespacedWriteModel;
 import com.mongodb.connection.ClusterDescription;
 import com.mongodb.connection.SocketSettings;
 import com.mongodb.internal.TimeoutSettings;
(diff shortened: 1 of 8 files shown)


driver-sync/src/main/com/mongodb/client/internal/MongoClientImpl.java
```diff
@@ -75,14 +73,22 @@ public final class MongoClientImpl implements MongoClient {
...
     private final MongoClusterImpl delegate;
     private final AtomicBoolean closed;
+    private final AutoCloseable externalResourceCloser;
 
-    public MongoClientImpl(final MongoClientSettings settings, final MongoDriverInformation mongoDriverInformation) {
-        this(createCluster(settings, mongoDriverInformation), mongoDriverInformation, settings, null);
+    public MongoClientImpl(final Cluster cluster,
+                           final MongoClientSettings settings,
+                           final MongoDriverInformation mongoDriverInformation,
+                           @Nullable final AutoCloseable externalResourceCloser) {
```
vbabanin, line 81: This change aligns with the implementation in `MongoClientImpl` for the reactive-streams driver, as detailed here: [MongoClientImpl Code](https://github.com/mongodb/mongo-java-driver/blob/355c6ea795633ba2d4070c77ed93d3a994cfb66b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/MongoClientImpl.java#L86).

driver-sync/src/main/com/mongodb/client/internal/MongoClientImpl.java
```diff
@@ -114,6 +120,13 @@ public void close() {
             }
             delegate.getServerSessionPool().close();
             delegate.getCluster().close();
+            if (externalResourceCloser != null) {
+                try {
+                    externalResourceCloser.close();
+                } catch (Exception e) {
+                    LOGGER.warn("Exception closing resource", e);
+                }
+            }
```
katcharov, line 129: Is it correct that this is the functional change, and the rest is effectively refactoring?

vbabanin: Yes, that’s correct — this is the only functional change. The rest is refactoring to align with the async `MongoClient` implementation and simplify testing, so we can inject a `SocketStreamFactoryFactory` and verify it's properly closed without relying on static mocks.

rozza (approved):
LGTM

katcharov (approved):

PR #1650, opened 2025-03-13
Adjust maxAwaitTimeMS by remaining timeout.

vbabanin:
JAVA-5720


driver-core/src/test/resources/unified-test-format/client-side-operation-timeout/tailable-awaitData.json
```diff
@@ -419,6 +420,141 @@
...
+                  "maxTimeMS": {
+                    "$$lte": 70
+                  }
+                }
+              }
+            }
+          ]
+        }
+      ]
+    },
```
vbabanin, line 509: The failpoint is used here to delay the first `getMore` by 30ms, reducing raciness and allowing verification that the second `getMore` occurs with `maxTimeMs` being less than 70ms.

PR #1651, opened 2025-03-18
Optimize String write

vbabanin:
## Summary

This PR optimizes string writing in BSON by minimizing redundant checks and internal method indirection in hot paths.

## Key Changes:

- Removed internal method indirection (e.g., write() wrappers) that introduced extra bound checks.
- Avoided repeated calls to ensureOpen(), hasRemaining(), and related checks by caching position and limit values inside hot paths.
- When writing ASCII, write directly to the underlying ByteBuffer array if capacity allows. This avoids the extra bounds and range checks in `put()`. Logic follows the same fast path approach used in `String.getBytes()` for ASCII.
- Used `str.charAt()` instead of `Character.toCodePoint()` to avoid unnecessary surrogate checks when not needed (e.g., for characters within the ASCII or 2-byte UTF-8 code unit range). Fall back only when multi-unit characters (e.g., 3-byte UTF-8) are detected.

## Performance analysis
To ensure accurate performance comparison and reduce noise, 11 versions (Comparison Versions) were aggregated and compared against a stable region of data around the Base Mainline Version. The percentage difference and z-score of the mean of the Comparison Versions were calculated relative to the Base Mainline Version’s stable region mean. 

The following tables show improvements across two configurations:

- The first uses ASCII-only strings from the regular benchmark suite.
- The second uses an augmented benchmark dataset with UTF-8 strings containing 3-byte characters.

### ASCII Benchmark Suite (Regular JSON Workloads)
Test | Base Mean (MB/s) | Patched Mean (ops/s) | Diff | Z-Score
-- | -- | -- | -- | --
Deep BSON Encoding | 47.02 | 56.97 | +21.2% | 2.97
Flat BSON Encoding | 105.49 | 238.03 | +125.6% | 6.65
Full BSON Encoding | 110.68 | 176.65 | +59.6% | 4.21
Large Doc Bulk Insert | 46.75 | 62.48 | +33.6% | 2.99
Large Doc InsertOne | 50.52 | 67.18 | +33.0% | 3.03
Small Doc Bulk Insert | 25.09 | 28.30 | +12.8% | 2.07

Perf analyzer results: [Link](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=67e060e522c4af821b4590c7&selected_tab=data-table&percent_filter=0%7C%7C100&z_filter=0%7C%7C10&filter_type=Default)

### Augmented Benchmark Suite (UTF-8 Strings with 3-Byte Characters)

To evaluate performance on multi-byte UTF-8 input, the large_doc.json, small_doc.json, and tweet.json datasets were modified to use UTF-8 characters encoded with 3 bytes (code units). These changes were introduced on mainline via an Evergreen patch, and benchmark results were collected from:

- 11 runs on patched mainline (baseline with UTF-8 input)
- 11 runs on this branch (UTF-8 input + string write optimizations)

Test | Base Mean (MB/s) | Patched Mean (ops/s) | Diff | Z-Score
-- | -- | -- | -- | --
Deep BSON Encoding | 91.31 | 124.3 | +35.8% | 6.94
Flat BSON Encoding | 95.79 | 193.47 | +100.2% | 13.2
Full BSON Encoding | 86.3 | 151.8 | +73.4% | 9.76
Large Doc Bulk Insert | 35.5 | 46.0 | +31.8% | 8.79
Large Doc InsertOne | 42.8 | 53.28 | +25.9% | 8.06
Small Doc Bulk Insert | 43.6 | 53.85 | +23.6% | 8.45

Perf analyzer results: [Link](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=67e254aca0fff1c53cd70117&selected_tab=data-table&determination_filter=Improved&percent_filter=0%7C%7C100&z_filter=0%7C%7C10&filter_type=Default)

JAVA-5816


franz1981:
That's related to https://github.com/mongodb/mongo-java-driver/pull/1629 🙏

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
```diff
@@ -282,4 +305,129 @@ private static final class BufferPositionPair {
...
+            int c = str.charAt(sp);
+
+            if (checkNullTermination && c == 0x0) {
+                throw new BsonSerializationException(
+                        format("BSON cstring '%s' is not valid because it contains a null character " + "at index %d", str, sp));
+            }
+
+            if (c < 0x80) {
+                if (remaining == 0) {
+                    buf = getNextByteBuffer();
```
franz1981, line 347: I still suggest to give at shot at the PR I made which use a separate index to access single bytes in the internalNio Buffer within the Netty buffers, for two reasons:
- NIO ByteBuffer can benefit from additional optimizations from the JVM since it's a known type to it
- Netty buffer read/write both move forward the internal indexes and force Netty to verify accessibility of the buffer for each operation, which have some Java Memory Model effects (.e.g. any subsequent load has to happen for real, each time!)

franz1981 (changes requested):

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
vbabanin: Good point - accessing the NIO buffer directly sounds like a potential win. I’m aiming to keep this PR focused and incremental for easier review and integration. We could consider Netty-specific optimizations in a follow-up PR/ scope, once we have Netty benchmarks running in CI

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
```diff
@@ -282,4 +305,129 @@ private static final class BufferPositionPair {
             this.position = position;
         }
     }
+
+    protected int writeCharacters(final String str, final boolean checkNullTermination) {
```
rozza, line 309: This is quite complex to understand.  It could do with some inline comments to explain the code flow, or potentially have the while loop in another method named after why its used as a potential fall back.

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonOutputTest.java
```diff
@@ -82,7 +149,7 @@ void positionAndSizeShouldBe0AfterConstructor(final String branchState) {
                     break;
                 }
                 default: {
-                    throw Assertions.fail(branchState);
+                    throw com.mongodb.assertions.Assertions.fail(branchState);
```
rozza, line 152: Could this not just use `org.junit.jupiter.api.fail` ?

rozza (changes requested):
Looks good a couple of comments and one suggestion to help improve the readability of `writeCharacters`

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonOutputTest.java
stIncMale: This code was originally introduced by me. The reason I use `com.mongodb.assertions.Assertions` in tests sometimes instead of the assertions from `org.junit.jupiter.api` is that I want to separate the assertions that are part of what a test is supposed to check, from the assertions that are about the test code itself. The latter are just like the assertions we have in the production code - we can remove them without affecting the usefulness of the tests, but they are there to make the test code itself clearer and more robust.

I am not sure, though, we have the discipline to have such separation.

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
```diff
@@ -155,14 +174,18 @@ public int pipe(final OutputStream out) throws IOException {
...
+                while (cur.hasRemaining()) {
+                    int numBytesToCopy = Math.min(cur.remaining(), tmp.length);
+                    cur.get(tmp, 0, numBytesToCopy);
+                    out.write(tmp, 0, numBytesToCopy);
+                }
+                total += cur.limit();
             }
-            total += dup.limit();
+        } finally {
+            byteBuffers.forEach(ByteBuf::release);
```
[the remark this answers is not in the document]
vbabanin: Yes, that was a bug. The `getByteBuffers()` method is documented to return `ByteBufs` that must eventually be released:

> The returned {@link ByteBuf}s must eventually be {@linkplain ByteBuf#release() released}.

However, we never released them in `pipe` method, and the external caller wouldn’t have been able to either. In tests, `Netty` reported a resource leak since the reference count remained above zero when the buffers were garbage collected.

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
vbabanin: I’ve added some inline comments to clarify the flow and moved the fallback path to a separate method to make the structure a bit more readable. Let me know if you have any other suggestions.

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
```diff
@@ -100,11 +103,17 @@ private ByteBuf getCurrentByteBuffer() {
...
+        assertFalse(bufferList.get(curBufferIndex).hasRemaining());
+        return getByteBufferAtIndex(++curBufferIndex);
+    }
+
     private ByteBuf getByteBufferAtIndex(final int index) {
         if (bufferList.size() < index + 1) {
-            bufferList.add(bufferProvider.getBuffer(index >= (MAX_SHIFT - INITIAL_SHIFT)
-                                                            ? MAX_BUFFER_SIZE
-                                                            : Math.min(INITIAL_BUFFER_SIZE << index, MAX_BUFFER_SIZE)));
+            ByteBuf buffer = bufferProvider.getBuffer(index >= (MAX_SHIFT - INITIAL_SHIFT)
```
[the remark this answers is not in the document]
vbabanin: I moved it to a separate variable to improve profiling readability in the IDE. Since we might explore this area of buffer management further, having it separated from the `list.add(...)` call helps with visibility.

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonOutputTest.java
vbabanin: I kept it for the same reason - to separate test logic assertions from behavioral ones. I think this distinction could be useful if the test is ever refactored or extended. Perhaps it's something we could discuss more broadly as a test code style convention.

rozza: Sound good. Ensuring the fail message clarifies its a test implementation bug and not a driver behaviour failure would help.

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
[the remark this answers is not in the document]
rozza: Great catch :)

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
rozza: Thank you :)

rozza (approved):
LGTM!

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
```diff
@@ -282,4 +305,165 @@ private static final class BufferPositionPair {
...
+                    return sp + 1;
+                }
+                // Otherwise, update the position to reflect the partial write.
+                position += sp;
+                curBuffer.position(curBufferPos);
+            }
+        }
+
+        // We get here, when the buffer is not backed by an array, or when the string contains at least one non-ASCII characters.
+        return writeOnBuffers(str,
```
franz1981, line 340: Can't we have within this a fast PATH for ASCII too - it will grant more chances to get inlined (since is a smaller method) and be more unrolled...
If we can have a JMH bench it would be fairly easy (I can do it) to peek into the assembly produced to verify it

franz1981 (changes requested):

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
franz1981: > Are you suggesting we add a fast path similar to writeOnArrayAscii, but using dynamic buffer allocation and falling back to writeOnBuffers/writeUtf8OnBuffers when a UTF-8 character is encountered?

Yep, since I see the the ascii path there is already taking care to change the buffer to write against instead of performing a lookup per each byte to write.
A tighter loop increase the chance it to be loop unrolled, although...the fact we can change the buffer where to write during the loop, can affect this - both for the array case and this.

vbabanin: Yeah, it makes sense. Thanks for the suggestion. I implemented `writeOnBuffersAsccii` and ran local benchmarks.

<details>
<summary> writeOnBuffersAsccii </summary>

```
private int writeOnBuffersAsccii(final String str,
                                     final boolean checkNullTermination,
                                     final int stringPointer,
                                     final int bufferLimit,
                                     final int bufferPos,
                                     final ByteBuf buffer) {
        int remaining;
        int sp = stringPointer;
        int curBufferPos = bufferPos;
        int curBufferLimit = bufferLimit;
        ByteBuf curBuffer = buffer;
        final int length = str.length();

        while (sp < length) {
            remaining = curBufferLimit - curBufferPos;
            char c = str.charAt(sp);
            if (checkNullTermination && c == 0) {
                throw new BsonSerializationException(
                        format("BSON cstring '%s' is not valid because it contains a null character " + "at index %d", str, sp));
            }
            if (c >= 0x80) {
                break;
            }
            if (remaining == 0) {
                curBuffer = getNextByteBuffer();
                curBufferPos = 0;
                curBufferLimit = curBuffer.limit();
            }
            curBuffer.put((byte) c);
            position++;
            sp++;
            curBufferPos++;
        }
        return sp;
    }
```
</details>

I printed the assembly to see the JIT’s behavior. It looks like the loop wasn’t unrolled by JIT - there’s only one `charAt` and `put` per iteration in the main loop. However, there seems to be an additional `charAt` in the buffer allocation path (after `getNextByteBuffer`), but it’s a separate code path.

I’ve shared a GitHub Gist with the shortened assembly (keeping the key parts) and a pseudo-Java interpretation to show how the assembly could map back to the logic: [Gist](https://gist.github.com/vbabanin/e20301d879a0acfccefd53aec0584f05). Local perf showed modest gains likely limited by dynamic buffer allocation, as you noted. I’ll run more tests on a dedicated perf instance to confirm. If I missed anything in the assembly, please let me know!

I’m merging this PR for its improvements, but I agree tighter loops or manual unrolling could be further explored, if the performance gains justify the maintainability trade-off.

franz1981: I mean something which uses compile command too as https://github.com/TechEmpower/FrameworkBenchmarks/pull/9800#issuecomment-2796302625

vbabanin: You’re right - my earlier Gist showed raw hex. I recompiled on Oracle JDK 17.0.7 with hsdis for readable assembly. Thanks!

The main loop (0x0000000113a12c40–0x0000000113a12d3c) seems to have no unrolling:
Condition: 0x0000000113a12c40 (cmp w4, w15, line 453: sp < length).
charAt: One instance at 0x0000000113a12c60–0x0000000113a12c88 (line 455).
Checks: Null termination (0x0000000113a12c8c, line 456), ASCII (0x0000000113a12c94, line 460), buffer space (0x0000000113a12c98, line 463).
put: One instance at 0x0000000113a12ca4–0x0000000113a12d20 (line 468).
Index Updates: 0x0000000113a12d24–0x0000000113a12d34 (lines 469–471).
Branch Back: 0x0000000113a12d3c: b.lt 0x0000000113a12c60, looping to 0x0000000113a12c40.

A second charAt appears in the getNextByteBuffer path, not the main loop. I’ve created new [Gist](https://gist.github.com/vbabanin/1e21a96e7d042ba03772d6aacf45071c) with the readable assembly

vbabanin:
I've added additional tests for code points represented by surrogate pairs:
https://github.com/mongodb/mongo-java-driver/pull/1651/commits/dd5fe4de84e04e5b155080d4a759e4363671b4f8

Adding a second reviewer to look at just this commit -  the rest has already been reviewed by @rozza.

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonOutputTest.java
```diff
@@ -536,15 +628,23 @@ void shouldTruncateToPosition(final boolean useBranch) {
...
+            Consumer<ByteBufferBsonOutput> assertByteBuffers = effectiveOut -> {
+                List<ByteBuf> byteBuffers = new ArrayList<>();
+                try {
+                    byteBuffers = effectiveOut.getByteBuffers();
+                    assertEquals(
+                            asList(1 << 10, 1 << 11, 1 << 12, 1 << 13, 1 << 14, 1 << 15, 1 << 16, 1 << 17, 1 << 18, 1 << 19, 1 << 20,
+                                    1 << 21, 1 << 22, 1 << 23, 1 << 24, 1 << 24),
+                            byteBuffers.stream().map(ByteBuf::capacity).collect(toList()));
+                } finally {
+                    byteBuffers.forEach(ByteBuf::release);
```
vbabanin, line 645: Netty was logging warnings due to a direct ByteBuf not being released properly.

franz1981: Yep leaving leak detection paranoid enabled for test is key to save leaks to happen

driver-core/src/main/com/mongodb/internal/connection/ByteBufferBsonOutput.java
franz1981: Mmm It looks to me that the unrolling was having a factor of two, I will re read it again since I am more used to x86 asm :)

Anyway, I suggest to look at the PR I sent for this same optimization: having the check for remaining space in the loop would bloat the loop body, reducing the chances that C2 will unroll it many times.
You should keep the loop as simple and branch free as possible

franz1981: Another reason why the Netty version loop body is too fat is because the JIT doesn't trust final fields and since you can get a new mongo ByteBuf at each iteration, the required amount of pointer chase (mongo buf, Netty swapped big, Netty buf, Unsafe..) is way to much; this prevent massively to have unrolling.
The reason why I have unwrapped till Netty or NIO, the buffers, is to be as per byte[], at the level in which the loop body can just use whatever is saved into a register (the reference of the address field in the Netty buffer) assuming it to be constant, and trust it.

nhachicha (approved):

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufSpecification.groovy
```diff
@@ -249,11 +249,7 @@ class ByteBufSpecification extends Specification {
         @Override
         ByteBuf getBuffer(final int size) {
             io.netty.buffer.ByteBuf buffer = allocator.directBuffer(size, size)
-            try {
-                new NettyByteBuf(buffer.retain())
-            } finally {
-                buffer.release();
-            }
+            new NettyByteBuf(buffer)
```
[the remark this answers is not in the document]
vbabanin: The previous logic retained a reference and then immediately released it in the finally block, which seemed redundant. The `ByteBuffer` is already released in the cleanup section of each Spock test.

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonOutputTest.java
```diff
@@ -929,13 +1047,553 @@ void shouldWriteDoubleWithinSpanningBuffers(
...
+        @DisplayName("should write surrogate CString across buffers")
+        @ParameterizedTest
+        @MethodSource("com.mongodb.internal.connection.ByteBufferBsonOutputTest#bufferProviders")
+        void shouldWriteCStringWithSurrogatePairs(final BufferProvider bufferProvider) throws IOException {
+            for (Integer surrogateCodePoint : ALL_UTF_16_CODE_POINTS_FORMED_BY_SURROGATE_PAIRS) {
+                String stringToEncode = new String(toSurrogatePair(surrogateCodePoint));
+                byte[] expectedEncoding = stringToEncode.getBytes(StandardCharsets.UTF_8);
+                int bufferAllocationSize = expectedEncoding.length + "\u0000".length();
+
+                testWriteCStringAcrossBufferWithBranch(
```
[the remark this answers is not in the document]
vbabanin: Good catch! Changed.

PR #1652, opened 2025-03-19
Support placeholder values in BsonReader and BsonWriter

jyemin:
This is an attempt to add support for placeholder values in JSON, so that our Hibernate dialect can support native queries like

```
{ 
   aggregate: "book",
   pipeline: [
       {
          $match: {
              title: ?
          }
       }
   ]
}
```
instead of the less intuitive: 

```
{ 
   aggregate: "book",
   pipeline: [
       {
          $match: {
              title: {$undefined: true}
           }
       }
   ]
}
```

It could also be used for non-native queries, but it's less important there.  It would be more to make the query logs look more sensible, i.e.

```
DEBUG o.h.p.entity.AbstractEntityPersister -  Insert (0): {"insert": "books", "documents": [{"author": ?, "publishYear": ?, "title": ?, "_id": ?}]}
DEBUG o.h.p.entity.AbstractEntityPersister -  Update (0): {"update": "books", "updates": [{"q": {"_id": {"$eq": ?}}, "u": {"$set": {"author": ?, "publishYear": ?, "title": ?}}, "multi": true}]}
DEBUG o.h.p.entity.AbstractEntityPersister -  Delete (0): {"delete": "books", "deletes": [{"q": {"_id": {"$eq": ?}}, "limit": {"$numberInt": "0"}}]}
```

instead of 

```
DEBUG o.h.p.entity.AbstractEntityPersister -  Insert (0): {"insert": "books", "documents": [{"authorFirstName": {"$undefined": true}, "authorLastName": {"$undefined": true}, "publishYear": {"$undefined": true}, "title": {"$undefined": true}, "_id": {"$undefined": true}}]}
DEBUG o.h.p.entity.AbstractEntityPersister -  Update (0): {"update": "books", "updates": [{"q": {"_id": {"$eq": {"$undefined": true}}}, "u": {"$set": {"authorFirstName": {"$undefined": true}, "authorLastName": {"$undefined": true}, "publishYear": {"$undefined": true}, "title": {"$undefined": true}}}, "multi": true}]}
DEBUG o.h.p.entity.AbstractEntityPersister -  Delete (0): {"delete": "books", "deletes": [{"q": {"_id": {"$eq": {"$undefined": true}}}, "limit": {"$numberInt": "0"}}]}
```

The problem is that in order to fit in the BsonReader/BsonWriter framework, the change is quite invasive, including

* Adding `writePlaceholder` methods to `BsonWriter`
* Adding `readPlaceholder` methods to `BsonReader`
* Adding a `BsonPlaceholder` subclass to `BsonValue`
* Adding a `PLACEHOLDER` value to the `BsonType` enum
* Adding a `BsonPlaceholderCodec`
* Added `asPlaceholder` and `isPlaceholder` to `BsonValue`
* Adding `doWritePlaceholder` method to `AbstractBsonWriter` and subclasses, including one in `BsonBinaryWriter` that necessarily throws `UnsupportedOperationException` since placeholder is not a real BSON type
* Adding `doReadPlaceholder` method to `AbstractBsonReader` and subclasses, including one in `BsonBinaryReader` that necessarily throws `UnsupportedOperationException` since placeholder is not a real BSON type

It would be hard to justify all these changes, but I thought it would be interesting for people to look at.

### DIFF
--- a/bson/src/main/org/bson/json/JsonReader.java
+++ b/bson/src/main/org/bson/json/JsonReader.java
@@ -23,6 +23,7 @@ import org.bson.BsonBinarySubType;
 import org.bson.BsonContextType;
 import org.bson.BsonDbPointer;
 import org.bson.BsonInvalidOperationException;
+import org.bson.BsonPlaceholder;
 import org.bson.BsonReaderMark;
 import org.bson.BsonRegularExpression;
 import org.bson.BsonTimestamp;
@@ -250,6 +251,9 @@ public class JsonReader extends AbstractBsonReader {
                     currentValue = visitUUIDConstructor();
                 } else if ("new".equals(value)) {
                     visitNew();
+                } else if ("?".equals(value)) {
+                    setCurrentBsonType(BsonType.PLACEHOLDER);
+                    currentValue = new BsonPlaceholder();
                 } else {
                     noValueFound = true;
                 }
@@ -406,6 +410,11 @@ public class JsonReader extends AbstractBsonReader {
 
     @Override
     protected void doReadUndefined() {
+    }
+
+    @Override
+    protected BsonPlaceholder doReadPlaceholder() {
+        return (BsonPlaceholder) currentValue;
     }
 
     @Override
--- a/bson/src/main/org/bson/AbstractBsonReader.java
+++ b/bson/src/main/org/bson/AbstractBsonReader.java
@@ -265,6 +265,8 @@ public abstract class AbstractBsonReader implements BsonReader {
      */
     protected abstract void doReadUndefined();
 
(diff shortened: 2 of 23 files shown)


bson/src/main/org/bson/json/JsonReader.java
```diff
@@ -250,6 +251,9 @@ public BsonType readBsonType() {
                     currentValue = visitUUIDConstructor();
                 } else if ("new".equals(value)) {
                     visitNew();
+                } else if ("?".equals(value)) {
```
jyemin, line 254: This part was easy, since we already support raw strings dues to the shell syntax

PR #1656, opened 2025-03-24
Eliminate unnecessary killCursors command when batchSize == limit

NathanQingyangXu:
https://jira.mongodb.org/browse/JAVA-5667


jyemin (changes requested):
There's a lot of apparent code duplication here.  I suggest you have a look at the `FindOperation` class and see if you can consolidate all the logic there, at the point where the action command document is created.

driver-reactive-streams/src/test/unit/com/mongodb/reactivestreams/client/internal/FindPublisherImplTest.java
```diff
@@ -88,7 +88,7 @@ void shouldBuildTheExpectedOperation() {
                 .retryReads(true)
                 .filter(new BsonDocument())
                 .allowDiskUse(false)
-                .batchSize(100)
+                .batchSize(101)
```
jyemin, line 91: I suspect that if you move the logic to FindOperation, you can revert the test changes.

PR #1661, opened 2025-03-28
Record both FaaS and container metadata when both are present

NathanQingyangXu:
https://jira.mongodb.org/browse/JAVA-5541

### DIFF
--- a/driver-core/src/test/functional/com/mongodb/internal/connection/ClientMetadataHelperProseTest.java
+++ b/driver-core/src/test/functional/com/mongodb/internal/connection/ClientMetadataHelperProseTest.java
@@ -197,11 +197,10 @@ public class ClientMetadataHelperProseTest {
     @Test
     void testKubernetesMetadataIncluded() {
         withWrapper()
-                .withEnvironmentVariable("AWS_EXECUTION_ENV", "AWS_Lambda_java8")
                 .withEnvironmentVariable("KUBERNETES_SERVICE_HOST", "kubernetes.default.svc.cluster.local")
                 .run(() -> {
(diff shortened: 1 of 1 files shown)


driver-core/src/test/functional/com/mongodb/internal/connection/ClientMetadataHelperProseTest.java
```diff
@@ -175,6 +175,23 @@ public void test08NotLambda() {
...
+                    BsonDocument actual = createActualClientMetadataDocument();
+                    assertEquals(expected, actual);
+
+                    performHello();
+                });
+    }
+
     // Additional tests, not specified as prose tests:
 
     @Test
```
vbabanin, line 197: I think we can now remove `testKubernetesMetadataIncluded`, as its logic is effectively covered by the newly added `test09ValidContainerAndFaasProvider`.

NathanQingyangXu: should we do similar change for the other two methods as well?

vbabanin: It'd be great to apply the same approach to the other two as well. However, it is totally optional though!

NathanQingyangXu: I updated the other two methods to avoid testing duplication and potential confusion in the future.

vbabanin (approved):

rozza (approved):
LGTM!

PR #1663, opened 2025-03-28
Remove support for MongoDB 4.0

jyemin:
* Remove branching code in the driver based on 4.0 version checks
* Remove testing of 4.0
* Clean up tests

JAVA-5831


.evergreen/.evg.yml
```diff
@@ -285,7 +285,7 @@ functions:
           export AWS_TEMP_SESSION_TOKEN=$CSFLE_AWS_TEMP_SESSION_TOKEN
           export CRYPT_SHARED_LIB_PATH=${CRYPT_SHARED_LIB_PATH}
 
-          AUTH="${AUTH}" SSL="${SSL}" MONGODB_URI="${MONGODB_URI}" SAFE_FOR_MULTI_MONGOS="${SAFE_FOR_MULTI_MONGOS}" TOPOLOGY="${TOPOLOGY}" \
```
jyemin, line 288: All supported server versions are safe for multi mongos now, so removing the variable entirely

.evergreen/.evg.yml
```diff
@@ -599,16 +599,6 @@ functions:
           LOGIN_CONTEXT_NAME=${LOGIN_CONTEXT_NAME} \
           .evergreen/run-gssapi-auth-test.sh
 
-  "run mmapv1 storage test":
```
jyemin, line 602: 4.0 is the last server version that supports this storage engine, so removing the task.

driver-core/src/main/com/mongodb/connection/ServerDescription.java
```diff
@@ -55,12 +55,12 @@ public class ServerDescription {
      * The minimum supported driver server version
      * @since 3.8
      */
-    public static final String MIN_DRIVER_SERVER_VERSION = "3.6";
+    public static final String MIN_DRIVER_SERVER_VERSION = "4.2";
```
jyemin, line 58: Looks like we missed updating this to 4.0 when 4.6 support was removed.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -416,7 +408,7 @@ private static OpCode getOpCode(final MessageSettings settings, final ClusterCon
     }
 
     private static boolean isServerVersionKnown(final MessageSettings settings) {
-        return settings.getMaxWireVersion() >= FOUR_DOT_ZERO_WIRE_VERSION;
+        return settings.getMaxWireVersion() >= EARLIEST_WIRE_VERSION;
```
jyemin, line 411: I added this new constant so we don't keep needing to update this code every time the earliest supported version changes.

jyemin: Though perhaps this should just be `> 0`

rozza: Yeah I think it can be as its validated in BaseCluster and this is only used to set the opcode. So can use `return settings.getMaxWireVersion() > MIN_WIRE_VERSION`.  Then `EARLIEST_WIRE_VERSION` isn't needed.

driver-core/src/main/com/mongodb/internal/operation/ServerVersionHelper.java
```diff
@@ -25,22 +25,19 @@
...
-    public static final int FOUR_DOT_ZERO_WIRE_VERSION = 7;
     public static final int FOUR_DOT_TWO_WIRE_VERSION = 8;
     public static final int FOUR_DOT_FOUR_WIRE_VERSION = 9;
-    public static final int FIVE_DOT_ZERO_WIRE_VERSION = 12;
+    public static final int FIVE_DOT_ZERO_WIRE_VERSION = 13;
     public static final int SIX_DOT_ZERO_WIRE_VERSION = 17;
     public static final int SEVEN_DOT_ZERO_WIRE_VERSION = 21;
-    public static final int LATEST_WIRE_VERSION = SEVEN_DOT_ZERO_WIRE_VERSION;
+    public static final int EIGHT_DOT_ZERO_WIRE_VERSION = 25;
+    public static final int EARLIEST_WIRE_VERSION = FOUR_DOT_TWO_WIRE_VERSION;
```
rozza, line 34: Not sure these are needed -`LATEST_WIRE_VERSION` is only used in tests and `EARLIEST_WIRE_VERSION` could be replaced by `MIN_WIRE_VERSION`.

jyemin: I removed EARLIEST_WIRE_VERSION and renamed MIN_WIRE_VERSION to UNKNOWN_WIRE_VERSION, which is more accurate.  I left LATEST_WIRE_VERSION since it's used quite a lot in tests and I wasn't sure what to replace it with.  Since this class is internal, this seems an ok spot to leave this constant.  What do you think?

rozza (approved):
LGTM!

PR #1664, opened 2025-03-29
Add String cache for BsonArray indexes.

vbabanin:
JAVA-5836


bson/src/main/org/bson/BsonBinaryWriter.java
```diff
@@ -37,8 +37,16 @@ public class BsonBinaryWriter extends AbstractBsonWriter {
 
     private final BsonOutput bsonOutput;
     private final Stack<Integer> maxDocumentSizeStack = new Stack<>();
+    private static final int ARRAY_INDEXES_CACHE_SIZE = 1 << 8;
```
rozza, line 40: Setting it to 256 explicitly reduces the cognitive load.

```suggestion
    private static final int ARRAY_INDEXES_CACHE_SIZE = 256;
```

rozza (approved):
LGTM!

One optional suggestion.

ShaneHarvey:
FYI we cache the first 1000 keys in Python: https://github.com/mongodb/mongo-python-driver/blob/4.11.3/bson/__init__.py#L646

PR #1671, opened 2025-04-02
Updated the bom validation to ensure it runs.

rozza:
JAVA-5822

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -53,7 +53,7 @@ dependencies {
 val defaultScalaVersion: String = project.findProperty("defaultScalaVersion")!!.toString()
 val scalaVersions: List<String>? = project.findProperty("supportedScalaVersions")?.toString()?.split(",")
 
-assert(!scalaVersions.isNullOrEmpty()) {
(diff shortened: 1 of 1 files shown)


bom/build.gradle.kts
```diff
@@ -53,7 +53,7 @@ dependencies {
 val defaultScalaVersion: String = project.findProperty("defaultScalaVersion")!!.toString()
 val scalaVersions: List<String>? = project.findProperty("supportedScalaVersions")?.toString()?.split(",")
 
-assert(!scalaVersions.isNullOrEmpty()) {
+require(!scalaVersions.isNullOrEmpty()) {
     "Scala versions must be provided as a comma-separated list in the 'supportedScalaVersions' project property"
 }
```
vbabanin, line 59: I suggest adding a check for the expected version format - this way, if the version is accidentally changed during development (e.g., an extra patch version is added), it can be caught here in the BOM.

```suggestion
scalaVersions?.forEach { version ->
    require(version.matches(Regex("\\d\\.\\d{2}"))) {
        "Scala version '$version' must be in the format X.YY"
    }
}
```

vbabanin (approved):
LGTM!

PR #1673, opened 2025-04-03
Optimize BsonArray Index encoding

vbabanin:
Instead of caching precomputed `String` values, we now cache their corresponding **byte[] representations**. This avoids repeated ASCII string decoding. 

The single contiguous buffer approach provides better cache locality in loops, reduced memory overhead and better performance. 

The array index caching implementation consumes approximately `12 KB of memory in total`:
```
ARRAY_INDEXES_BUFFER: ~3,890 bytes
Numbers 0-9: 10 × (1 byte + null terminator) = 20 bytes
Numbers 10-99: 90 × (2 byte + null terminator) = 270 bytes
Numbers 100-999: 900 × (3 bytes + null terminator) = 3,600 bytes
```

```
ARRAY_INDEXES_OFFSETS: 4,000 bytes
1,000 integers × 4 bytes per int = 4,000 bytes
```

```
ARRAY_INDEXES_LENGTHS: 4,000 bytes
1,000 integers × 4 bytes per int = 4,000 bytes
```

`Array object headers: ~48 bytes`

BsonArrayCodecBenchmark results:

| Metric                        | Before        | After         | Change       |
|------------------------------|---------------|---------------|--------------|
| ops/s                        | 13583.483     | 20046.810     | **+47.5%** |

JAVA-5836

### DIFF
--- a/bson/src/main/org/bson/BsonBinaryWriter.java
+++ b/bson/src/main/org/bson/BsonBinaryWriter.java
@@ -37,13 +37,34 @@ public class BsonBinaryWriter extends AbstractBsonWriter {
 
     private final BsonOutput bsonOutput;
     private final Stack<Integer> maxDocumentSizeStack = new Stack<>();
-    private static final int ARRAY_INDEXES_CACHE_SIZE = 256;
-    private static final String[] ARRAY_INDEXES_CACHE = new String[ARRAY_INDEXES_CACHE_SIZE];
+    private static final int ARRAY_INDEXES_CACHE_SIZE = 1000;
+    private static final byte[] ARRAY_INDEXES_BUFFER;
+    private static final int[] ARRAY_INDEXES_OFFSETS;
+    private static final int[] ARRAY_INDEXES_LENGTHS;
     private Mark mark;
 
     static {
+        int totalSize = 0;
+        int[] lengths = new int[ARRAY_INDEXES_CACHE_SIZE];
         for (int i = 0; i < ARRAY_INDEXES_CACHE_SIZE; i++) {
-            ARRAY_INDEXES_CACHE[i] = Integer.toString(i);
+            String string = Integer.toString(i);
+            lengths[i] = string.length() + 1; // +1 for null terminator
+            totalSize += lengths[i];
+        }
+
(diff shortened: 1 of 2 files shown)


vbabanin:
As @ShaneHarvey noted in the previous PR (https://github.com/mongodb/mongo-java-driver/pull/1664#issuecomment-2773351852), Python caches 1000 consecutive indexes. To align with that behavior, I’ve adjusted the cache size to 1000 in the Java driver.

bson/src/main/org/bson/BsonBinaryWriter.java
```diff
@@ -37,13 +38,34 @@ public class BsonBinaryWriter extends AbstractBsonWriter {
 
     private final BsonOutput bsonOutput;
     private final Stack<Integer> maxDocumentSizeStack = new Stack<>();
-    private static final int ARRAY_INDEXES_CACHE_SIZE = 256;
-    private static final String[] ARRAY_INDEXES_CACHE = new String[ARRAY_INDEXES_CACHE_SIZE];
+    private static final int ARRAY_INDEXES_CACHE_SIZE = 1000;
+    private static final byte[] ARRAY_INDEXES_BUFFER;
```
vbabanin, line 42: In prior benchmarks, the reference-chasing layout (e.g., `byte[][]`) showed ~25% lower throughput compared to a flat `byte[]` layout, primarily due to fragmented sequential locality and indirect memory access.

bson/src/test/unit/org/bson/BsonBinaryWriterTest.java
```diff
@@ -66,7 +66,7 @@ public void shouldThrowWhenMaxDocumentSizeIsExceeded() {
             writer.writeEndDocument();
             fail();
         } catch (BsonMaximumSizeExceededException e) {
-            assertEquals("Document size of 1037 is larger than maximum of 1024.", e.getMessage());
+            assertEquals("Document size of 12917 is larger than maximum of 12904.", e.getMessage());
```
rozza, line 69: Did this get missed last time?

rozza (approved):
LGTM!

PR #1685, opened 2025-04-24
Optimize String length computation.

vbabanin:
JAVA-5842


bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +182,58 @@ public void skipCString() {
...
+
+               mask:
+               00000000 00000000 11111111 00000000 00000001 00000001 00000000 00000111
+
+               ANDing mask with 0x8080808080808080 isolates the high bit (0x80) in positions where
+               the original byte was 0x00, setting the high bit to 1 only at the 0x00 byte position.
+
+               result:
+               00000000 00000000 10000000 00000000 00000000 00000000 00000000 00000000
+                                           ^^^^^^^^
```
jyemin:
```suggestion
                                          ^^^^^^^^
```

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +182,58 @@ public void skipCString() {
...
+                          The high bit is set only at the 0x00 byte position.
```
jyemin, line 222:
```suggestion
               The high bit is set only at the 0x00 byte position.
```

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +182,58 @@ public void skipCString() {
         buffer.position(pos + length);
     }
 
+    /**
+     * Detects the position of the first NULL (0x00) byte in a 64-bit word using SWAR technique.
+     * <a href="https://graphics.stanford.edu/~seander/bithacks.html#ZeroInWord">
```
jyemin, line 187: Do you have reason to believe that this link will still be around in, say, 5 years?  If not, maybe just go with Wikipedia.

jyemin (changes requested):
Overall looks good, just a few minor changes requests.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
jyemin: Not sure if you meant to center this...

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +182,58 @@ public void skipCString() {
...
+            mask &= ~word;
+            /*
+               0x8080808080808080:
+               10000000 10000000 10000000 10000000 10000000 10000000 10000000 10000000
+
+               mask:
+               00000000 00000000 11111111 00000000 00000001 00000001 00000000 00000111
+
+               ANDing mask with 0x8080808080808080 isolates the high bit (0x80) in positions where
+               the original byte was 0x00, setting the high bit to 1 only at the 0x00 byte position.
```
NathanQingyangXu, line 217: Optional. maybe the following verbiage is little bit clearer:
```
, by setting the high bit to 1 only at the 0x00 byte position.
```

bson/src/main/org/bson/io/ByteBufferBsonInput.java
NathanQingyangXu: Yeah, seems Wiki URL like https://en.wikipedia.org/wiki/SWAR is better

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +182,58 @@ public void skipCString() {
         buffer.position(pos + length);
     }
 
+    /**
+     * Detects the position of the first NULL (0x00) byte in a 64-bit word using SWAR technique.
+     * <a href="https://graphics.stanford.edu/~seander/bithacks.html#ZeroInWord">
+     */
     private int computeCStringLength(final int prevPos) {
         ensureOpen();
```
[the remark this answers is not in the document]
vbabanin: Removed. I also added `ensureOpen` to `readString`, as it previously relied on `computeCStringLength` to perform the check.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
vbabanin: I'm not 100% sure if the link will stay available long-term, however, it has been maintained for over 24 years. In my opinion, it explains SWAR optimization and its alternatives more clearly. In any case, it would remain accessible through the Wayback Machine. 

I’ve switched to the Wikipedia link for better stability.

jyemin: 24 years seems like a good bet to continue.  Maybe keep both?  Up to you.

jyemin (approved):
LGTM!

stIncMale (changes requested):
I still have not internalized the technique, and, therefore, can't check that it's correct (but I don't think spending any more time on the review is warranted).

The code does look like https://jameshfisher.com/2017/01/24/bitwise-check-for-zero-byte/ on its surface. However, based on both Wikipedia and the above link, it may matter whether little- or big-endian byte order is used (`ByteBufferBsonInput` explicitly uses the little-endian byte order). Though when I tried playing with the big-endian byte order, the code seemed to still work.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +183,57 @@ public void skipCString() {
         buffer.position(pos + length);
     }
 
+    /**
+     * Detects the position of the first NULL (0x00) byte in a 64-bit word using SWAR technique.
+     * <a href="https://en.wikipedia.org/wiki/SWAR">
+     */
     private int computeCStringLength(final int prevPos) {
-        ensureOpen();
         int pos = buffer.position();
```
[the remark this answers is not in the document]
jyemin: Alternatively, should we remove the parameter if `buffer.position()` is always what's passed in?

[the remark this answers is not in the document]
stIncMale: We may, but that will result in reading `buffer.position` twice in `skipCString`, instead of once. However, I would not be surprised if in reality there is only one read, and the second time the value is used, it is not being re-read. I am fine with either change.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +183,57 @@ public void skipCString() {
...
         int limit = buffer.limit();
 
+        int chunks = (limit - pos) >>> 3;
+        // Process 8 bytes at a time.
+        for (int i = 0; i < chunks; i++) {
+            long word = buffer.getLong(pos);
+            /*
+              Subtract 0x0101010101010101L to cause a borrow on 0x00 bytes.
+              if original byte is 00000000, then 00000000 - 00000001 = 11111111 (borrow causes high bit set to 1).
```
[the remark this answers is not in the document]
vbabanin: Yes, "[high bit](https://en.wiktionary.org/wiki/high_bit)" refers to the most significant bit - it's an alternative term sometimes used in [CS literature](https://graphics.stanford.edu/~seander/bithacks.html). I’ve changed it to “MSB” for clarity.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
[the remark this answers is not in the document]
vbabanin: Good catch! Yes, `prevPos` was additionally supposed to avoid reading `buffer.position()` twice. In some of my experiments, JIT didn’t always inline those calls - though of course that can vary depending on load or runtime conditions. Given that this is a hot path, I’d lean toward keeping `prevPos`.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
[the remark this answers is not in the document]
stIncMale: Many do not do the distinction between bit and byte when abbreviating, but in https://en.wikipedia.org/wiki/Bit_numbering it is done, and most/least significant **bit** is abbreviated MS**b**/LS**b**. I like how that makes things clear, compared to just using MSB/LSB for both bits and bytes. If you also find that helpful, let's adopt the approach.

stIncMale (approved):
Left an optional suggestion, feel free to ignore.

The last reviewed commit is 364a8108a47fb282e18f0de80eb7237833bb46a8.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +183,58 @@ public void skipCString() {
...
+
+               result:
+               00000000 00000000 10000000 00000000 00000000 00000000 00000000 00000000
+                                 ^^^^^^^^
+               The MSB is set only at the 0x00 byte position.
+             */
+            mask &= 0x8080808080808080L;
+            if (mask != 0) {
+                /*
+                 *  Performing >>> 3 (i.e., dividing by 8) gives the byte offset from the LSB.
```
NathanQingyangXu, line 229: LSB is used to denote the bit in a byte; but here we are computing the `offset in terms of byte`, so
```
... gives the byte offset.
```
is enough and less confusing.

stIncMale: > LSB is used to denote the bit in a byte

LSB may mean both least significant byte and bit, which is why I advocated for using MSb in https://github.com/mongodb/mongo-java-driver/pull/1685#discussion_r2068514237. The phrase

```
Performing >>> 3 (i.e., dividing by 8) gives the byte offset from the LSB.
```

Is correct and clear, assuming that LSB means "least significant **byte**".

The new next is also good, and explains things in more details.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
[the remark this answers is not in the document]
stIncMale: Another alternative is to just stop using abbreviations. After thinking about this more, I think, it's the only viable alternative.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
vbabanin: I’ll update the terminology as suggested by @stIncMale : LSb for least significant bit and LSB for least significant byte for better clarify. 

If there are no objections, I’ll also add a detailed comment, suggested above, explaining why we use `trailingZeros` instead of `leadingZeros`.

vbabanin:
I’ve re-requested review since previous approvals became stale after recent changes.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +183,64 @@ public void skipCString() {
...
+        int toPos = pos + (chunks << 3);
+        for (; pos < toPos; pos += Long.BYTES) {
+            long chunk = buffer.getLong(pos);
+            /*
+              Subtract 0x0101010101010101L to cause a borrow on 0x00 bytes.
+              if original byte is 00000000, then 00000000 - 00000001 = 11111111 (borrow causes the most significant bit set to 1).
+             */
+            long mask = chunk - 0x0101010101010101L;
+            /*
+              mask will only have the most significant bit set iff it was a 0x00 byte (0x00 becomes 0xFF because of the borrow).
```
stIncMale, line 206: "most significant bit" may refer to the most significant bit in a byte, or in a long (the chunk), and without clarified the text is still unclear. Note also how you used "significant byte **of the long**" below.

```suggestion
              mask will only have the most significant bit in each byte set iff it was a 0x00 byte (0x00 becomes 0xFF because of the borrow).
```

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +183,64 @@ public void skipCString() {
...
+              ~chunk will have bits that were originally 0 set to 1.
+              mask & ~chunk will have the most significant bit set iff original byte was 0x00.
```
stIncMale, line 208: "most significant bit" may refer to the most significant bit in a byte, or in a long (the chunk), and without clarified the text is still unclear. Note also how you used "significant byte **of the long**" below.

```suggestion
              mask & ~chunk will have the most significant bit set in each byte iff original byte was 0x00.
```

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +183,64 @@ public void skipCString() {
...
+            mask &= ~chunk;
+            /*
+               0x8080808080808080:
+               10000000 10000000 10000000 10000000 10000000 10000000 10000000 10000000
+
+               mask:
+               00000000 00000000 11111111 00000000 00000001 00000001 00000000 00000111
+
+               ANDing mask with 0x8080808080808080 isolates the most significant bit (0x80) in positions where
+               the original byte was 0x00, thereby setting the most significant bit to 1 only at the 0x00 byte position.
```
stIncMale, line 219: "most significant bit" may refer to the most significant bit in a byte, or in a long (the chunk), and without clarified the text is still unclear. Note also how you used "significant byte **of the long**" below.

```suggestion
               ANDing mask with 0x8080808080808080 isolates the most significant bit in each byte where
               the original byte was 0x00, thereby setting the most significant bit to 1 in each 0x00 original byte.
```

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -182,11 +183,64 @@ public void skipCString() {
...
+
+               result:
+               00000000 00000000 10000000 00000000 00000000 00000000 00000000 00000000
+                                 ^^^^^^^^
+               The most significant bit is set only at the 0x00 byte position.
```
stIncMale, line 224: "most significant bit" may refer to the most significant bit in a byte, or in a long (the chunk), and without clarified the text is still unclear. Note also how you used "significant byte **of the long**" below.

```suggestion
               The most significant bit is set in each 0x00 byte, and only there.
```

stIncMale (changes requested):

bson/src/main/org/bson/io/ByteBufferBsonInput.java
stIncMale: Btw, I am not so sure that "and only there" is true. I suspect, after the first (left-most, closest to the least significant byte) `0x00` byte, bits may be set on other bytes, but we don't care about that, as we are looking only for the left-most one. That's why (again, I suspect) Wikipedia mentions that the technique is applicable only if _"the goal is limited to finding the first zero byte on a little-endian processor"_.

vbabanin: Interesting, the claim on Wikipedia about "little-endian only" is misleading (maybe a documentation error or not complete). The algorithm, before `numberOfTrailingZeros`, works on both systems (little or big endian), as the zero-byte detection relies on bitwise operations that are endian-agnostic. The only place where byte endianness matters is in `numberOfTrailingZeros` or `numberOfLeadingZeros` to correctly identify the offset, but Wikipedia does not seem to mention that.

stIncMale: > the zero-byte detection relies on bitwise operations that are endian-agnostic

I am not sure that's true, as `chunk - 0x0101010101010101L` is not a bitwise operation, and we actually treat it as if it's a **byte**wise operation. It may not matter, of course, however https://jameshfisher.com/2017/01/24/bitwise-check-for-zero-byte/ mentions:

> But why were we able to treat the subtraction as byte-wise subtraction? The subtraction algorithm doesn’t work like that! Well, it does work like this for our particular case where there are no zero bytes and we are subtracting 1. It is only when doing 00000000 - 00000001 that the carry bit will be set when crossing the byte boundary.

As I mentioned in https://github.com/mongodb/mongo-java-driver/pull/1685#pullrequestreview-2801172065, I don't really understand what's going on (for that, I'd have to spend much more time playing with the code and observing intermediate results, and also learn how subtraction works), but it is not impossible that _"our particular case where there are no zero bytes and we are subtracting 1"_ holds only because we are looking for the left-most zero byte in a long, i.e., there are no other zero bytes before it when the `-` operation is being executed, and so the carry bits (I don't actually know what they are) from them do not exist and can't affect the outcome. The zero byte we need is going to be the left-most in a long only if the byte order is little endian.

stIncMale (approved):

stIncMale:
Approved. But I don't know if the tests fail because `main` is broken, or because of the changes in this PR.

The last reviewed commit is 176db5b7b4ce98fbe3804437e3df5e32ef44be4d.

jyemin (approved):
LGTM.  I like the updated comments.

NathanQingyangXu (approved):

PR #1687, opened 2025-04-28
Tests: Add support for crud tests bypassDocumentationValidation.json

rozza:
JAVA-5827

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
@@ -168,8 +168,6 @@ public final class UnifiedTestModifications {
                 .when(() -> def.isReactive() && UnifiedTest.Language.KOTLIN.equals(def.getLanguage()))
                 .file("crud", "findOne");
 
-        def.skipNoncompliant("Find with batchSize equal to limit - batchSize is set to limit if limit is smaller")
-                .test("crud", "find", "Find with batchSize equal to limit");
         def.skipNoncompliant("Updates and Replace bulk operations are split in the java driver")
                         .file("crud", "bulkWrite-comment");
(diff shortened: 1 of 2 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
```diff
@@ -168,8 +168,8 @@ public static void applyCustomizations(final TestDef def) {
                 .when(() -> def.isReactive() && UnifiedTest.Language.KOTLIN.equals(def.getLanguage()))
                 .file("crud", "findOne");
 
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5827")
-                        .file("crud", "bypassDocumentValidation");
+        def.skipNoncompliant("Find with batchSize equal to limit - batchSize is set to limit if limit is smaller")
+                .test("crud", "find", "Find with batchSize equal to limit");
```
katcharov, line 172: I'm not sure I understand the rationale - in [the test](https://github.com/mongodb/specifications/blob/335848d3f4577724b7b82cb9fa0c12defba54647/source/crud/tests/unified/find.json#L292), the limit is not smaller, but is equal to the batchSize. What is the actual expected behaviour in the spec, is it to set batchSize to 1 over the limit, or something else? And why is the driver not doing that?

katcharov (commented):
Looking good, just a question.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
rozza: Updated - I'm not sure why that was there, must have been an unneeded artifact left over when developing the initial submodule code.

katcharov (approved):
LGTM!

PR #1688, opened 2025-04-28
Ensure socks5 proxy option parsing is to specification

rozza:
Duplicates are not permitted, where as elsewhere in the connection string they are.

JAVA-5834

### DIFF
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -56,6 +56,7 @@ import static com.mongodb.MongoCredential.ALLOWED_HOSTS_KEY;
 import static com.mongodb.internal.connection.OidcAuthenticator.OidcValidator.validateCreateOidcCredential;
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static java.util.Collections.unmodifiableList;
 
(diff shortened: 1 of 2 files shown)


driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -1259,6 +1259,19 @@ private void validateProxyParameters() {
             throw new IllegalArgumentException(
                     "Both proxyUsername and proxyPassword must be set together. They cannot be set individually");
         }
+
+        if (proxyHost != null && optionsMap.get("proxyhost").size() > 1) {
```
stIncMale, line 1263: `proxyHost != null` is used here (and similarly below) as a way to check that `optionsMap.get("proxyhost")` is not `null`. Let's not do that, because it is not only non-obvious, but also fragile. We may introduce a `static` method for checking that a `Collection` has more than one element, which checks for `null`s, instead of what we do now.

driver-core/src/test/unit/com/mongodb/UriOptionsTest.java
```diff
@@ -47,9 +47,6 @@ public void shouldPassAllOutcomes() {
         // No CANONICALIZE_HOST_NAME support https://jira.mongodb.org/browse/JAVA-4278
         assumeFalse(getDescription().equals("Valid auth options are parsed correctly (GSSAPI)"));
 
-        // https://jira.mongodb.org/browse/JAVA-5834
-        assumeFalse(getFilename().equals("proxy-options.json"));
```
stIncMale, line 51: `proxy-options.json` was not added to our codebase by https://github.com/mongodb/mongo-java-driver/pull/1180 (in fact, no spec tests were added), despite them having been added in https://jira.mongodb.org/browse/DRIVERS-1357. I wonder if @vbabanin remembers the reason, and if it was done on purpose.

stIncMale (changes requested):

driver-core/src/test/unit/com/mongodb/UriOptionsTest.java
stIncMale: Got a response - it wasn't the case that we omitted them intentionally, because they did not fit our behavior.

driver-core/src/main/com/mongodb/ConnectionString.java
rozza: Good shout - updated.

stIncMale (approved):

PR #1689, opened 2025-04-28
Added option to configure DEK cache lifetime

rozza:
Added the `AutoEncryptionSettings.Builder#keyExpiration` method and `ClientEncryptionSettings.Builder#keyExpiration` method to configure the cache expiration time for data encryption keys.

JAVA-5547


nhachicha (commented):
LGTM 👍 
1- Does this use the updated spec tests? https://github.com/mongodb/specifications/commit/787bbe67bdffb614270654372112e93fc2cd3eb3 

2- What happens when the key expires? Are we testing this via unified or prose test?

rozza:
@nhachicha thanks for the review:

> 1- Does this use the updated spec tests?
Yes 👍 
> 2- What happens when the key expires? Are we testing this via unified or prose test?
This is handled by libmongocrypt and is covered by unified and their tests

nhachicha (approved):

PR #1692, opened 2025-04-28
Added regression tests for "number" alias in $$type operator.

rozza:
JAVA-5682

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
@@ -24,6 +24,7 @@ import org.bson.BsonValue;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import static java.util.Arrays.asList;
 import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/ValueMatcher.java
```diff
@@ -158,7 +159,12 @@ private void assertValuesMatch(final BsonValue initialExpected, @Nullable final
     private void assertExpectedType(final BsonValue actualValue, final BsonValue expectedTypes) {
         List<String> types;
         if (expectedTypes.isString()) {
-            types = singletonList(expectedTypes.asString().getValue());
+            String expectedType = expectedTypes.asString().getValue();
+            if (expectedType.equals("number")) {
```
NathanQingyangXu, line 163: Discussion only. Nothing needs to be done.

I know in this specific case, `expectedfType` can't be `null`, but generally speaking, it is a good practice to put string constant to the left hand side of `equals()` method to avoid NPE.

```
if ("number".equals(expectedType)) {
```
Not a big deal for sure.

NathanQingyangXu (approved):

NathanQingyangXu:
LGTM. Tested successfully locally.

PR #1695, opened 2025-04-29
Add tracing support using Micrometer

nhachicha:
https://jira.mongodb.org/browse/JAVA-5733

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
@@ -178,8 +178,13 @@ public final class UnifiedTestModifications {
         // OpenTelemetry
         def.skipJira("https://jira.mongodb.org/browse/JAVA-5991")
                 .file("open-telemetry/tests", "operation find")
+                .file("open-telemetry/tests", "operation find_one_and_update")
+                .file("open-telemetry/tests", "operation update")
+                .file("open-telemetry/tests", "operation bulk_write")
+                .file("open-telemetry/tests", "operation drop collection")
                 .file("open-telemetry/tests", "transaction spans")
                 .file("open-telemetry/tests", "convenient transactions")
+                .file("open-telemetry/tests", "operation atlas_search")
                 .file("open-telemetry/tests", "operation insert")
                 .file("open-telemetry/tests", "operation map_reduce")
                 .file("open-telemetry/tests", "operation find without db.query.text")


driver-core/src/main/com/mongodb/tracing/MicrometerTracer.java
```diff
@@ -0,0 +1,102 @@
...
+ * limitations under the License.
+ */
+
+package com.mongodb.tracing;
+
+import com.mongodb.internal.tracing.Span;
+import com.mongodb.internal.tracing.TraceContext;
+
+public class MicrometerTracer implements Tracer {
+    private final io.micrometer.tracing.Tracer tracer;
```
rozza, line 23: I believe these can be imported and you don't need to use full path names. Even though this dependency is optional, it will be there at compile time (and in testing) so would only be an issue if a user tried to import the class without taking the dependency.

rozza (commented):
I'm liking the approach. 

Does this meet the spec for transactions and retryable operations?

I'm hoping so as both `OperationContext#withSessionContext` and `OperationContext#withTimeoutContext` reuse the operation id.

However, it would be worth clarifying before reviewing further.

driver-core/src/main/com/mongodb/tracing/MicrometerTracer.java
nhachicha: We need the FQN otherwise, this will conflict with `com.mongodb.internal.tracing.Tracer`

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -723,6 +728,20 @@ Builder heartbeatSocketTimeoutMS(final int heartbeatSocketTimeoutMS) {
...
 
+        /**
+         * Sets the tracer to use for creating Spans for operations and commands.
+         *
+         * @param tracer the tracer
+         * @see com.mongodb.tracing.MicrometerTracer
+         * @return this
+         * @since 5.5
+         */
+        @Alpha(Reason.CLIENT)
```
rozza: Q: Is this going to be released as alpha?

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -186,6 +186,10 @@ BsonDocument getCommandDocument(final ByteBufferBsonOutput bsonOutput) {
         }
     }
 
+    BsonDocument getCommand() {
```
rozza: recommend adding a javadoc to help disambiguate this from the `getCommandDocument` method.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -432,15 +446,31 @@ public boolean reauthenticationIsTriggered(@Nullable final Throwable t) {
     private <T> T sendAndReceiveInternal(final CommandMessage message, final Decoder<T> decoder,
             final OperationContext operationContext) {
         CommandEventSender commandEventSender;
+        Span tracingSpan = createTracingSpan(message, operationContext);
+
         try (ByteBufferBsonOutput bsonOutput = new ByteBufferBsonOutput(this)) {
             message.encode(bsonOutput, operationContext);
-            commandEventSender = createCommandEventSender(message, bsonOutput, operationContext);
+            BsonDocument commandDocument = message.getCommandDocument(bsonOutput);
```
rozza: This always hydrates the `commandDocument`, even if its potentially not used, this will impact performance. So it hydrated if:
 
`tracingSpan == null` and

```
 boolean listensOrLogs = commandListener != null || COMMAND_PROTOCOL_LOGGER.isRequired(DEBUG, getClusterId());
        if (!recordEverything && (isMonitoringConnection || !opened() || !authenticated.get() || !listensOrLogs)) {
            return new NoOpCommandEventSender();
```

We should avoid this and only create it if its used.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -952,19 +983,87 @@ public void onResult(@Nullable final ByteBuf result, @Nullable final Throwable t
...
+     * If command payload tracing is enabled, the command document is also attached as a tag.
+     *
+     * @param message          the command message to trace
+     * @param operationContext the operation context containing tracing and session information
+     * @return the created {@link Span}, or {@code null} if tracing is not enabled or the command is security-sensitive
+     */
+    @Nullable
+    private Span createTracingSpan(final CommandMessage message, final OperationContext operationContext) {
+        TracingManager tracingManager = operationContext.getTracingManager();
+        BsonDocument command = message.getCommand();
```
rozza: Now that `message.getCommand` is available does it make sense to use it everywhere where in `InternalStreamConnection` for example where: `SECURITY_SENSITIVE_COMMANDS.contains(message.getCommandDocument(bsonOutput).getFirstKey())` is used.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -952,19 +983,87 @@ public void onResult(@Nullable final ByteBuf result, @Nullable final Throwable t
...
+            parentContext = tracingManager.getCursorParentContext(cursorId);
+        } else {
+            parentContext = tracingManager.getParentContext(operationContext.getId());
+        }
+
+        Span span = tracingManager
+                .addSpan("Command " + commandName, parentContext)
+                .tag(SYSTEM, "mongodb")
+                .tag(NAMESPACE, message.getNamespace().getDatabaseName())
+                .tag(QUERY_SUMMARY, command.toString());
```
rozza: Should this be `command.toJson()` ?

driver-core/src/main/com/mongodb/internal/operation/AggregateOperationImpl.java
```diff
@@ -242,9 +243,14 @@ BsonDocument getCommand(final OperationContext operationContext, final int maxWi
...
-        return (result, source, connection) ->
-                new CommandBatchCursor<>(getTimeoutMode(), result, batchSize != null ? batchSize : 0,
-                        getMaxTimeForCursor(source.getOperationContext().getTimeoutContext()), decoder, comment, source, connection);
+        return (result, source, connection) -> {
+            OperationContext operationContext = source.getOperationContext();
+
+            // register cursor id with the operation context, so 'getMore' commands can be folded under the 'find' operation
+            TracingManager.linkCursorWithOperation(result, operationContext);
+            return new CommandBatchCursor<>(getTimeoutMode(), result, batchSize != null ? batchSize : 0,
+                    getMaxTimeForCursor(source.getOperationContext().getTimeoutContext()), decoder, comment, source, connection);
```
rozza:
```suggestion
                    getMaxTimeForCursor(operationContext.getTimeoutContext()), decoder, comment, source, connection);
```

driver-core/src/main/com/mongodb/internal/tracing/Span.java
```diff
@@ -0,0 +1,117 @@
...
+ *   <li><b>Operation Spans</b>: Trace higher-level operations, which may include multiple commands or internal steps.</li>
+ *   <li><b>Transaction Spans</b>: Trace the lifecycle of a transaction, including all operations and commands within it.</li>
+ * </ul>
+ * </p>
+ *
+ * @since 5.6
+ */
+public interface Span {
+    /**
+     * A no-op implementation of the Span interface.
```
rozza:
```suggestion
     * An empty / no-op implementation of the Span interface.
```

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -723,6 +728,20 @@ Builder heartbeatSocketTimeoutMS(final int heartbeatSocketTimeoutMS) {
             return this;
         }
 
+        /**
+         * Sets the tracer to use for creating Spans for operations and commands.
+         *
+         * @param tracer the tracer
+         * @see com.mongodb.tracing.MicrometerTracer
+         * @return this
+         * @since 5.5
```
rozza:
```suggestion
         * @since 5.6
```

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -1040,6 +1059,17 @@ public ContextProvider getContextProvider() {
         return contextProvider;
     }
 
+    /**
+     * Get the tracer to create Spans for operations and commands.
+     *
+     * @return this
+     * @since 5.5
```
rozza:
```suggestion
     * @since 5.6
```

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -126,14 +135,16 @@ public OperationContext(final long id,
                             final TimeoutContext timeoutContext,
                             final ServerDeprioritization serverDeprioritization,
                             @Nullable final ServerApi serverApi,
-                            @Nullable final String operationName) {
+                            @Nullable final String operationName,
+                            final TracingManager tracingManager) {
```
rozza: Optional: reorder params as `TracingManager` is not nullable, so that it comes before the nullable params.

driver-core/src/main/com/mongodb/internal/operation/CommandBatchCursor.java
```diff
@@ -169,6 +173,9 @@ public void remove() {
 
     @Override
     public void close() {
+        if (cursorId != -1) {
+            operationContext.getTracingManager().removeCursorParentContext(cursorId);
```
rozza: 👍

driver-core/src/main/com/mongodb/internal/operation/AggregateOperationImpl.java
```diff
@@ -242,9 +243,14 @@ BsonDocument getCommand(final OperationContext operationContext, final int maxWi
     }
 
     private CommandReadTransformer<BsonDocument, CommandBatchCursor<T>> transformer() {
-        return (result, source, connection) ->
-                new CommandBatchCursor<>(getTimeoutMode(), result, batchSize != null ? batchSize : 0,
-                        getMaxTimeForCursor(source.getOperationContext().getTimeoutContext()), decoder, comment, source, connection);
+        return (result, source, connection) -> {
+            OperationContext operationContext = source.getOperationContext();
+
+            // register cursor id with the operation context, so 'getMore' commands can be folded under the 'find' operation
```
rozza:
```suggestion
            // register cursor id with the operation context, so 'getMore' commands can be folded under the 'aggregate' operation
```

driver-scala/build.gradle.kts
```diff
@@ -36,6 +36,9 @@ dependencies {
 
     // Encryption testing
     integrationTestImplementation(project(path = ":mongodb-crypt", configuration = "default"))
+
+    // Micrometer dependency for tracing
+    api(libs.micrometer)
```
rozza: Should this be a `testImplementation` like reactive streams and kotlin?

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
```diff
@@ -317,4 +341,52 @@ private TimeoutContext createTimeoutContext(final TransactionOptions transaction
                 TransactionOptions.merge(transactionOptions, getOptions().getDefaultTransactionOptions()),
                 operationExecutor.getTimeoutSettings()));
     }
+
+    private void handleTransactionSpanError(final Throwable e) {
```
rozza: It feels wrong that these are in `ClientSessionImpl` is there a way to abstract them out to the tracing package?

driver-sync/src/test/functional/com/mongodb/client/tracing/ZipkinTracer.java
```diff
@@ -0,0 +1,95 @@
...
+ * This tracer can be used to send spans to a Zipkin server.
+ * <p>
+ * Spans are visible in the Zipkin UI at <a href="http://localhost:9411">...</a>.
+ * <p>
+ * To Start Zipkin server, you can use the following command:
+ * <pre>{@code
+ * docker run -d -p 9411:9411 openzipkin/zipkin
+ * }</pre>
+ */
+public final class ZipkinTracer {
```
rozza: Lets add a MongoDB property to auto include use of this and document that.

eg: `org.mongodb.test.zipkin=true`

driver-sync/src/test/functional/com/mongodb/client/unified/Entities.java
```diff
@@ -604,6 +612,22 @@ private void initClient(final BsonDocument entity, final String id,
             }
             clientSettingsBuilder.serverApi(serverApiBuilder.build());
         }
+
+        if (entity.containsKey("tracing")) {
+            boolean enableCommandPayload = entity.getDocument("tracing").get("enableCommandPayload", BsonBoolean.FALSE).asBoolean().getValue();
+            /* To enable Zipkin backend, uncomment the following lines and ensure you have the server started
```
rozza: See previous recommendation - lets make this based off a system property so it can be added without a code change.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: Why was this method added when there is already `getCommandDocument`, which provides a correct and performant command document (The `command` field does not, since it is most likely an instance of `BsonDocumentWrapper`, and it will not contain fields from the `PAYLOAD_TYPE_1_DOCUMENT_SEQUENCE` used in bulk writes)?

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -1040,6 +1059,17 @@ public ContextProvider getContextProvider() {
...
+     */
+    @Alpha(Reason.CLIENT)
+    public TracingManager getTracingManager() {
```
jyemin: It would be more consistent with the rest of the various settings classes to return the same type that was used to configure, rather than a wrapping manager class, i.e.

```
public Tracer getTracer() {
    return tracer;
}
```

It's easy enough to create the TracingManager within the `MongoClientImpl` constuctors.

Also, clients are allowed to create multiple `MongoClient` instances from the same `MongoClientSettings`, and if that happens then we'll have two multiple `MongoClient` instances sharing the same `TracingManager` instance.

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -723,6 +728,20 @@ Builder heartbeatSocketTimeoutMS(final int heartbeatSocketTimeoutMS) {
...
+        /**
+         * Sets the tracer to use for creating Spans for operations and commands.
+         *
+         * @param tracer the tracer
+         * @see com.mongodb.tracing.MicrometerTracer
+         * @return this
+         * @since 5.5
+         */
+        @Alpha(Reason.CLIENT)
+        public Builder tracer(final Tracer tracer) {
```
jyemin: Given that `Tracer` and related types are part of the API, they should not be in `com.mongodb.internal`

driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java
```diff
@@ -0,0 +1,291 @@
...
+        this(tracer, tracer.currentContext());
+    }
+
+    /**
+     * Constructs a new TracingManager with the specified tracer and parent context.
+     *
+     * @param tracer        The tracer to use for tracing operations.
+     * @param parentContext The parent trace context.
+     */
+    public TracingManager(final Tracer tracer, final TraceContext parentContext) {
```
jyemin: This constructor is only used by the constructor just above.  Is that intentional, and if so, it can be inlined.

driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java
```diff
@@ -0,0 +1,291 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+import static com.mongodb.internal.operation.OperationHelper.LOGGER;
+import static com.mongodb.internal.tracing.Tags.CURSOR_ID;
+import static com.mongodb.internal.tracing.Tags.SYSTEM;
+import static java.lang.String.format;
+
+/**
+ * Manages tracing spans for MongoDB driver activities.
```
jyemin, line 53: I'm not clear why a class like this is necessary.  As I understand it, there is a single instance created at the scope of `MongoClientSettings`, and treated as effectively a singleton by the `MongoClient` instance.  Would it be sufficient, and also simpler, to create a new instance of some class at the point where we start execution of an operation or transaction, and allow it to garbage collected when the operation or transaction completes?

driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java
```diff
@@ -0,0 +1,291 @@
...
+public class TracingManager {
+    /**
+     * A no-op instance of the TracingManager used when tracing is disabled.
+     */
+    public static final TracingManager NO_OP = new TracingManager(Tracer.NO_OP);
+
+    private final Tracer tracer;
+    private final TraceContext parentContext;
+    private final Map<String, TraceContext> transactions = new ConcurrentHashMap<>();
+    private final Map<Long, TraceContext> cursors = new ConcurrentHashMap<>();
```
jyemin: Cursor ids are not guaranteed to be unique across a cluster, so they don't make a very good key for a global map.

driver-core/src/main/com/mongodb/MongoNamespace.java
```diff
@@ -190,7 +190,7 @@ private static String getCollectionNameFullName(final String namespace) {
         return namespace.substring(firstDot + 1);
     }
 
-    private static String getDatatabaseNameFromFullName(final String namespace) {
+    private static String getDatabaseNameFromFullName(final String namespace) {
```
jyemin: This seems like a good change, but how did it become part of this PR, as it's the only change to this file.

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
jyemin, line 1: There is a corresponding class for the reactive streams driver called `com.mongodb.reactivestreams.client.internal.ClientSessionPublisherImpl`.  I suspect similar changes are required in that class as well.

driver-sync/src/main/com/mongodb/client/internal/MongoClientImpl.java
jyemin, line 1: Do we need to apply similar changes to `com.mongodb.reactivestreams.client.internal.MongoClientImpl`?

driver-sync/src/main/com/mongodb/client/internal/MongoClusterImpl.java
jyemin, line 1: Do we need to apply similar changes to `com.mongodb.reactivestreams.client.internal.MongoClusterImpl`?

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -432,15 +446,31 @@ public boolean reauthenticationIsTriggered(@Nullable final Throwable t) {
     private <T> T sendAndReceiveInternal(final CommandMessage message, final Decoder<T> decoder,
             final OperationContext operationContext) {
         CommandEventSender commandEventSender;
+        Span tracingSpan = createTracingSpan(message, operationContext);
```
jyemin: Consider pushing all the changes made to this class down into `LoggingCommandEventSender`, which already contains a lot of similar logic.  `LoggingCommandEventSender` might need some refactoring as a result, but it seems better than messing too much with this class, which already has enough responsibilities as it is.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
nhachicha: This does return the original `BsonDocument` passed to the `CommandMessage` (this is needed to get the command name for example without parsing the document via getCommandDocument).

driver-core/src/main/com/mongodb/MongoNamespace.java
nhachicha: It's just fixing a typo

driver-sync/src/main/com/mongodb/client/internal/ClientSessionImpl.java
nhachicha: This implementation is limited to Sync driver only.

driver-sync/src/main/com/mongodb/client/internal/MongoClientImpl.java
nhachicha: This implementation is limited to Sync driver only.

driver-sync/src/main/com/mongodb/client/internal/MongoClusterImpl.java
nhachicha: This implementation is limited to Sync driver only.

driver-core/src/main/com/mongodb/MongoNamespace.java
jyemin: Oh, I thought CoPilot actually suggested it, and wasn't just commenting on it.

Still, our practice generally is to avoid including drive-by changes like this into PRs, especially when they are in files that are otherwise uninvolved in the PR.

Feel free to open a new PR with just this change though.  No ticket required.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: Right, but in some cases, when the command itself is an instance of `BsonDocumentWrapper`, this will result in hydrating into a Map, which is something that we assiduously try to avoid during encoding.

jonatan-ivanov:
Hi,

I'm one of the Maintainers of Micrometer and Micrometer Tracing.
I'm happy you want to use the Tracing API! I can share some feedback on this PR if you would like to receive some.

I think the most important feedback I'd like to give is recommending to use the Observation API instead of "just" the Tracing API. The reason behind it is that this PR adds tracing capabilities but if you also want to add metrics support, this needs to be instrumented again. Also if users want (custom) logs or any custom output that they can figure out based on the instrumentation in these components, that also need to happen through a separate mechanism.
The idea behind the Observation API, is to put these Observability needs behind one API so that you need to instrument your components once and you can get any output you like from that instrumentation (tracing, metrics, logs, anything that your users can come up with) without modifying the instrumentation itself.
The usage of it is similar to the Tracing API so what you already instrumented should be relatively easy to migrate but it's much more powerful than that and if you register tracing, it will call the Micrometer Tracing API to create spans. The whole Spring Portfolio is instrumented using this API including most of the popular frameworks/libraries.
If you want to learn more, here's a talk here we explain some basics: https://www.youtube.com/watch?v=Qyku6cR6ADY (starts at 14:00 but I recommend watching the whole talk).

Regarding to the Tracing API usage, I think the most important thing I would like to mention is that it is a Tracing Facade. It is what SLF4J is for loggers. As SLF4J can delegate to logging libraries, Micrometer Tracing can delegate to tracing libraries (Brave and OpenTelemetry right now but custom ones can be implemented by users). Because of this, it's usually not needed to pull another facade in front of it (just like it is usually not needed for SLF4J).

Also, the Tracing and Observation APIs provide the features that are implemented here: `NOOP`, unit test support (spans in-memory), integration test support (with Brave/Zipkin, OTel/OTLP, ...), etc.

The Observation API also has a nice feature that the Tracing API does not and it seems you could benefit from that (based on the `Tags` class and span names): it has a component called `ObservationConvention` which lets the MongoDB team define their own convention(s) but the users can define theirs too and decide which one they want to use. So users can use the MongoDB team's convention (default), the OpenTelemetry naming convention, or their company's convention (regardless of which Tracing library is under the hood). You can see this in action in the video I linked above.

driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java
nhachicha: New implementation uses a single instance per MongoClient.

driver-core/src/main/com/mongodb/internal/tracing/MicrometerTracer.java
```diff
@@ -0,0 +1,228 @@
...
+import static com.mongodb.tracing.MongodbObservation.LowCardinalityKeyNames.EXCEPTION_MESSAGE;
+import static com.mongodb.tracing.MongodbObservation.LowCardinalityKeyNames.EXCEPTION_STACKTRACE;
+import static com.mongodb.tracing.MongodbObservation.LowCardinalityKeyNames.EXCEPTION_TYPE;
+import static com.mongodb.tracing.MongodbObservation.MONGODB_OBSERVATION;
+import static java.lang.System.getenv;
+import static java.util.Optional.ofNullable;
+
+
+/**
+ * A {@link Tracer} implementation that delegates tracing operations to a Micrometer {@link io.micrometer.observation.ObservationRegistry}.
```
vbabanin:
```suggestion
 * A {@link Tracer} implementation that delegates tracing operations to a Micrometer {@link ObservationRegistry}.
```

driver-core/src/main/com/mongodb/internal/tracing/MicrometerTracer.java
```diff
@@ -0,0 +1,228 @@
...
+                Observation observation = getObservation(name).parentObservation(parentObservation);
+                return new MicrometerSpan(observation.start(), namespace);
+            }
+        }
+        Observation observation = getObservation(name);
+        return new MicrometerSpan(observation.start(), namespace);
+    }
+
+    @Override
+    public boolean enabled() {
```
vbabanin: [nit] We already have these existing naming conventions in the codebase:

```
  In interfaces:
  - isCausallyConsistent() - ClientSession.java:112
  - hasActiveTransaction() - ClientSession (multiple variants):46

  In classes with getter methods:
  - getRetryWrites() - MongoClientSettings.java:802
  - getRetryReads() - MongoClientSettings.java:813
  - isServerDefault() - ReadConcern.java:101
  - isBypassAutoEncryption() - AutoEncryptionSettings.java:473
```

Also `TracingManager` wraps this method and exposes another method named `isEnabled` [TracingManager.java#L101](https://github.com/mongodb/mongo-java-driver/blob/0628386af4508d749b018674181ce19e049f0c8f/driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java#L101).

Should we keep it consistent and name it `isEnabled()` and `isCommandPayloadIncluded()`?

driver-sync/src/test/functional/com/mongodb/client/tracing/MicrometerProseTest.java
```diff
@@ -0,0 +1,182 @@
...
+        MongoClientSettings clientSettings = Fixture.getMongoClientSettingsBuilder()
+                .tracer(tracer).build();
+
+        try (MongoClient client = MongoClients.create(clientSettings)) {
+            MongoDatabase database = client.getDatabase(getDefaultDatabaseName());
+            MongoCollection<Document> collection = database.getCollection("test");
+            collection.find().first();
+
+            // Assert that no OpenTelemetry tracing spans are emitted for the operation.
+            assertTrue(inMemoryOtel.getFinishedSpans().isEmpty(), "Spans should not be emitted when instrumentation is enabled.");
```
vbabanin:
```suggestion
            assertTrue(inMemoryOtel.getFinishedSpans().isEmpty(), "Spans should not be emitted when instrumentation is disabled.");
```

driver-sync/src/test/functional/com/mongodb/client/tracing/MicrometerProseTest.java
```diff
@@ -0,0 +1,182 @@
...
+        setEnv(ENV_OTEL_ENABLED, "true");
+        clientSettings = Fixture.getMongoClientSettingsBuilder()
+                .tracer(tracer).build();
+        try (MongoClient client = MongoClients.create(clientSettings)) {
+            MongoDatabase database = client.getDatabase(getDefaultDatabaseName());
+            MongoCollection<Document> collection = database.getCollection("test");
+            collection.find().first();
+
+            // Assert that OpenTelemetry tracing spans are emitted for the operation.
+            assertEquals(2, inMemoryOtel.getFinishedSpans().size(), "Spans should be emitted when instrumentation is disabled.");
```
vbabanin:
```suggestion
            assertEquals(2, inMemoryOtel.getFinishedSpans().size(), "Spans should be emitted when instrumentation is enabled.");
```

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -182,6 +183,12 @@ public String getCommandName() {
         return "bulkWrite";
     }
 
+    @Override
+    public MongoNamespace getNamespace() {
+        // The bulkWrite command is executed on the "admin" database.
+        return new MongoNamespace("admin", COMMAND_COLLECTION_NAME);
```
rozza, line 189: Nit: Make this whole namespace static.

Also COMMAND_COLLECTION_NAME is deprecated - so will need an ignore warning deprecated - or we could remove the deprecation and mark the static as internal.

driver-core/src/main/com/mongodb/internal/operation/CommitTransactionOperation.java
```diff
@@ -116,6 +117,11 @@ public String getCommandName() {
         return COMMAND_NAME;
     }
 
+    @Override
+    public MongoNamespace getNamespace() {
+        return new MongoNamespace("admin", MongoNamespace.COMMAND_COLLECTION_NAME);
```
rozza: Nit: Make this whole namespace static. (as previous).

driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java
```diff
@@ -0,0 +1,241 @@
...
+ * <p>
+ * This class provides methods to create and manage spans for commands, operations and transactions.
+ * It integrates with a {@link Tracer} to propagate tracing information and record telemetry.
+ * </p>
+ */
+public class TracingManager {
+    /**
+     * A no-op instance of the TracingManager used when tracing is disabled.
+     */
+    public static final TracingManager NO_OP = new TracingManager(ObservationRegistry.NOOP, false);
```
rozza: This adds a hard dependency on ObservationRegistry - so can use null here instead.

driver-core/src/main/com/mongodb/MongoClientSettings.java
```diff
@@ -723,6 +739,47 @@ Builder heartbeatSocketTimeoutMS(final int heartbeatSocketTimeoutMS) {
...
+         * captured.
+         *
+         * @param observationRegistry the observation registry
+         * @param enableCommandPayload whether command payloads should be captured in tracing spans. This may have performance
+         *                             implications so should be used with care.
+         * @return this
+         * @since 5.7
+         */
+        @Alpha(Reason.CLIENT)
+        public Builder observationRegistry(final ObservationRegistry observationRegistry, final boolean enableCommandPayload) {
```
rozza: This would be the only place where we have two settings combined. Elsewhere we'd use a separate settings class.

driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java
```diff
@@ -0,0 +1,241 @@
...
+     */
+    public TracingManager(@Nullable final ObservationRegistry observationRegistry, final boolean enableCommandPayload) {
+        String envOtelInstrumentationEnabled = getenv(ENV_OTEL_ENABLED);
+        boolean enableTracing = true;
+        if (envOtelInstrumentationEnabled != null) {
+            enableTracing = Boolean.parseBoolean(envOtelInstrumentationEnabled);
+        }
+        tracer = (observationRegistry == null) ? Tracer.NO_OP
+                : (enableTracing) ? new MicrometerTracer(observationRegistry, enableCommandPayload)
+                : Tracer.NO_OP;
```
vbabanin: [nit] Just to reduce cyclomatic complexity.
```suggestion
        tracer = enableTracing && observationRegistry != null ?
                new MicrometerTracer(observationRegistry, enableCommandPayload)
                : Tracer.NO_OP;
```

driver-core/src/main/com/mongodb/internal/tracing/Span.java
```diff
@@ -0,0 +1,141 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.internal.tracing;
+
+import com.mongodb.MongoNamespace;
+import com.mongodb.lang.Nullable;
+import io.micrometer.common.KeyValue;
+import io.micrometer.common.KeyValues;
```
vbabanin: The abstraction layer (interfaces like Span and Tracer) currently depend on concrete implementation details of Micrometer types, making it harder in the future to switch to a different tracing implementation.

We have two similar methods in Span:
` void tagHighCardinality(String keyName, BsonDocument value);`
  `void tagLowCardinality(KeyValue keyValue);`

The former is implementation independent, and i suggest we use this signature
//todo

driver-sync/src/test/functional/com/mongodb/client/tracing/MicrometerProseTest.java
```diff
@@ -0,0 +1,174 @@
...
+        setEnv(ENV_OTEL_ENABLED, "true");
+        try (MongoClient client = MongoClients.create(clientSettings)) {
+            MongoDatabase database = client.getDatabase(getDefaultDatabaseName());
+            MongoCollection<Document> collection = database.getCollection("test");
+            collection.find().first();
+
+            // Assert that OpenTelemetry tracing spans are emitted for the operation.
+            assertEquals(2, inMemoryOtel.getFinishedSpans().size(), "Spans should be emitted when instrumentation is disabled.");
+            assertEquals("find", inMemoryOtel.getFinishedSpans().get(0).getName());
+            assertEquals("find " + getDefaultDatabaseName() + ".test", inMemoryOtel.getFinishedSpans().get(1).getName());
```
vbabanin, line 110: Currently, finished spans in this test don’t include SpanKind.CLIENT (`inMemoryOtel.getFinishedSpans().get(1).getSpanKind()`), which is required by the spec.

It seems we don’t explicitly enforce this when creating spans in the driver.

1. We should add a test to ensure that SpanKind.CLIENT is always set.

2. One possible way to address this could be to use an additional `SenderContext` in `getObservation()`. I’m not sure if this is the best approach, but it seems to produce the expected `SpanKind.CLIENT`:

```
private Observation getObservation(final String name) {
    Observation observation = MONGODB_OBSERVATION.observation(observationRegistry,
            () -> new SenderContext((carrier, key, value) -> {}, Kind.CLIENT))
            .contextualName(name);
    observation.getContext().put(QUERY_TEXT_LENGTH_CONTEXT_KEY, textMaxLength);
    return observation;
}
```

This setup produces a span with `SpanKind.CLIENT`, which can be verified through the prose test.

driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java
```diff
@@ -0,0 +1,241 @@
...
+            final Predicate<String> isSensitiveCommand,
+            final Supplier<ServerAddress> serverAddressSupplier,
+            final Supplier<ConnectionId> connectionIdSupplier
+            ) {
+
+        BsonDocument command = commandDocumentSupplier.get();
+        String commandName = command.getFirstKey();
+        if (!isEnabled() || isSensitiveCommand.test(commandName)) {
+            return null;
+        }
```
vbabanin, line 193: We currently perform [document deserialization from `BsonOutput`](https://github.com/mongodb/mongo-java-driver/blob/9a3696edeaf1cb4d36a1c501786852aa416e5999/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java#L445) regardless of whether tracing is enabled. Let’s move the `isEnabled` check before invoking the supplier.

```suggestion
       if (!isEnabled()) {
            return null;
        }
        BsonDocument command = commandDocumentSupplier.get();
        String commandName = command.getFirstKey();
        if (isSensitiveCommand.test(commandName)) {
            return null;
        }
```

vbabanin (changes requested):

driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java
nhachicha: Good catch 👍

driver-sync/src/test/functional/com/mongodb/client/tracing/MicrometerProseTest.java
nhachicha: Good catch 👍

driver-core/src/main/com/mongodb/MongoNamespace.java
```diff
@@ -40,9 +41,12 @@ public final class MongoNamespace {
...
      */
-    @Deprecated
+    @Internal
     public static final String COMMAND_COLLECTION_NAME = "$cmd";
 
+    @Internal
+    public static final MongoNamespace ADMIN_DB_COMMAND_NAMESPACE = new MongoNamespace("admin", COMMAND_COLLECTION_NAME);
+
     private static final Set<Character> PROHIBITED_CHARACTERS_IN_DATABASE_NAME =
             new HashSet<>(asList('\0', '/', '\\', ' ', '"', '.'));
```
rozza:
```suggestion
    private static final Set<Character> PROHIBITED_CHARACTERS_IN_DATABASE_NAME =
            new HashSet<>(asList('\0', '/', '\\', ' ', '"', '.'));

    @Internal
    public static final String COMMAND_COLLECTION_NAME = "$cmd";

    @Internal
    public static final MongoNamespace ADMIN_DB_COMMAND_NAMESPACE = new MongoNamespace("admin", COMMAND_COLLECTION_NAME);
```

driver-core/src/main/com/mongodb/internal/tracing/TracingManager.java
```diff
@@ -0,0 +1,261 @@
...
+    public static final TracingManager NO_OP = new TracingManager(null);
+    private final Tracer tracer;
+    private final boolean enableCommandPayload;
+
+    /**
+     * Constructs a new TracingManager with the specified observation registry.
+     *
+     * @param observationRegistry The observation registry to use for tracing operations, may be null.
+     * @param enableCommandPayload Whether to enable command payload tracing.
+     */
```
vbabanin, line 70: [nit] 
```suggestion
    /**
     * Constructs a new TracingManager with the specified observation registry.
     * @param observationRegistry The observation registry to use for tracing operations, may be null.
     */
```

driver-core/src/main/com/mongodb/observability/ObservabilitySettings.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package com.mongodb.observability;
+
+import com.mongodb.annotations.Immutable;
+import com.mongodb.annotations.Sealed;
+
+@Sealed
+@Immutable
```
vbabanin, line 31: Let's mark this interface with Alpha status as well.
```suggestion
/**
 * Observability settings for the driver.
 *
 * @since 5.7
 */
@Alpha(Reason.CLIENT)
@Sealed
@Immutable
```

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
jyemin: @nhachicha what was the justification for not taking this suggestion?

nhachicha: That wasn't possible within the scope of this PR. https://jira.mongodb.org/browse/JAVA-5999

PR #1699, opened 2025-04-30
Added extra test to ignore for reactive streams

rozza:
Related to: 6bd6a19 and DRIVERS-2868 - I think there was an additional test that was missing an ignore.

JAVA-5720


nhachicha (approved):
No JIRA ticket needed for the skip?

rozza:
> No JIRA ticket needed for the skip?

Unfortunately, its a limitation of using Publishers. I believe this test was missed when the previous skips were added due the lack of tryNexts

nhachicha (approved):

PR #1705, opened 2025-05-02
Fix release version validation to use gradle.properties.

vbabanin:


vbabanin:
The change has been applied in: https://github.com/mongodb/mongo-java-driver/commit/03c9c55bc371ed9b8998bc7a7e85aa3b1f73aaf4.

PR #1708, opened 2025-05-06
Add Client Metadata Update Support.

vbabanin:


buildSrc/src/main/kotlin/conventions/testing-base.gradle.kts
```diff
@@ -105,3 +105,7 @@ testlogger {
     showFailedStandardStreams = true
     logLevel = LogLevel.LIFECYCLE
 }
+
+dependencies {
+    testImplementation(libs.assertj)
+}
```
vbabanin, line 111: AssertJ has been used in this PR and added to test-base as it is a useful library that could be shared across all modules.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -0,0 +1,49 @@
...
+import static com.mongodb.internal.connection.ClientMetadataHelper.createClientMetadataDocument;
+import static com.mongodb.internal.connection.ClientMetadataHelper.updateClientMedataDocument;
+
+/**
+ * Represents metadata of the current MongoClient.
+ *
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public class ClientMetadata {
+    private volatile BsonDocument clientMetadataBsonDocument;
```
jyemin: I don't think that a volatile fields is enough.  If `append` is called concurrently by multiple threads, then one of the updates to metadata will likely be lost.

jyemin (commented):
Initial feedback

driver-sync/src/main/com/mongodb/client/MongoClient.java
```diff
@@ -61,4 +62,18 @@ public interface MongoClient extends MongoCluster, Closeable {
...
+     * <p>
+     * This enables frameworks and libraries to include identifying metadata (e.g., name, version, platform) which might be visible in
+     * the MongoD/MongoS logs. This can assist with diagnostics by making client identity visible to the server.
+     * <p>
+     * <strong>Note:</strong> Metadata is limited to 512 bytes; any excess will be truncated.
+     *
+     * @param mongoDriverInformation the driver information to append to the existing metadata
+     * @since 5.6
+     */
+    void updateMetadata(MongoDriverInformation mongoDriverInformation);
```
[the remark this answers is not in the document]
vbabanin: Just conceptually, `MongoCluster` acts as a view with overridden configurations (e.g., codec registry, read preference) and doesn’t own the client’s lifecycle or global state.

In my opinion, placing `updateMetadata` on `MongoCluster `raises the question: does it update metadata for just that instance or for all instances? That behavior would require clear documentation.

`MongoClient` already handles driver-wide operations (e.g., close), so `updateMetadata` aligns with its role.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
vbabanin: You're right  - volatile alone isn't enough for concurrent updates. I initially didn’t add locking because I expected concurrent metadata updates to be extremely rare in practice, but I’ve now added a lock to ensure consistency.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadataHelper.java
```diff
@@ -179,6 +180,43 @@ static boolean clientMetadataDocumentTooLarge(final BsonDocument document) {
...
+
+        updatedDriverVersions.add(driverInformation.getString("version").getValue());
+        updatedDriverVersions.addAll(driverVersionsToAppend);
+
+        updateDriverPlatforms.add(clientMetadataDocument.getString("platform").getValue());
+        updateDriverPlatforms.addAll(driverPlatformsToAppend);
+
+        tryWithLimit(clientMetadataDocument, d -> {
+            putAtPath(d, "driver.name", listToString(updatedDriverNames));
+            putAtPath(d, "driver.version", listToString(updatedDriverVersions));
```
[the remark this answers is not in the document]
vbabanin: By the spec, driver name and version must be updated atomically. If both can't fit, then both must be omitted - so splitting them in a `tryWithLimit` wouldn't comply with that requirement.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadataHelper.java
```diff
@@ -179,6 +180,43 @@ static boolean clientMetadataDocumentTooLarge(final BsonDocument document) {
...
+        updatedDriverNames.add(driverInformation.getString("name").getValue());
+        updatedDriverNames.addAll(driverNamesToAppend);
+
+        updatedDriverVersions.add(driverInformation.getString("version").getValue());
+        updatedDriverVersions.addAll(driverVersionsToAppend);
+
+        updateDriverPlatforms.add(clientMetadataDocument.getString("platform").getValue());
+        updateDriverPlatforms.addAll(driverPlatformsToAppend);
+
+        tryWithLimit(clientMetadataDocument, d -> {
```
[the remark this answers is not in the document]
vbabanin: Since fields have variable length, even if the driver name doesn’t fit, the platform field might still fit, so we need to attempt adding each field individually.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public ClientMetadata(@Nullable final String applicationName, final MongoDriverInformation mongoDriverInformation) {
+        this.clientMetadataBsonDocument = createClientMetadataDocument(applicationName, mongoDriverInformation);
+    }
+
+    /**
+     * Returns mutable BsonDocument that represents the client metadata.
+     */
+    public BsonDocument getBsonDocument() {
+        return clientMetadataBsonDocument;
```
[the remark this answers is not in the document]
vbabanin: We frequently pass `BsonDocuments` through layers, and since they're mutable, I think it's acceptable not to clone here. Cloning also introduces some overhead during connection establishment - though to be fair, I'm speculating about the performance impact, as I haven't measured it. What do you think?

driver-core/src/main/com/mongodb/internal/connection/ClientMetadataHelper.java
```diff
@@ -179,6 +180,43 @@ static boolean clientMetadataDocumentTooLarge(final BsonDocument document) {
...
+            putAtPath(d, "driver.name", listToString(updatedDriverNames));
```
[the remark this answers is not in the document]
vbabanin: That's a fair point. In this case, I’d choose to follow the existing pattern in `ClientMetadataHelper` to keep the code consistent and neat, as this pattern exists elsewhere in the class. Since `updateMetadata` is expected to be called most likely once per application lifecycle and isn’t on a hot path or GC-critical, I’d prefer to keep it as is for now. Let me know what you think.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadataHelper.java
```diff
@@ -179,6 +180,43 @@ static boolean clientMetadataDocumentTooLarge(final BsonDocument document) {
...
         return buffer.getPosition() > MAXIMUM_CLIENT_METADATA_ENCODED_SIZE;
     }
+/**
+ * Modifies the given client metadata document by appending the driver information.
+ * Driver name and version are appended atomically to the existing driver name and version if they do not exceed
+ * {@value MAXIMUM_CLIENT_METADATA_ENCODED_SIZE} bytes.
+ *
+ * Platform is appended separately to the existing platform if it does not exceed {@value MAXIMUM_CLIENT_METADATA_ENCODED_SIZE} bytes.
+ */
+    public static BsonDocument updateClientMedataDocument(final BsonDocument clientMetadataDocument,
```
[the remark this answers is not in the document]
vbabanin: Good catch! Renamed, thanks!

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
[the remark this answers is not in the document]
vbabanin: We often pass mutable `BsonDocument`s through layers (except at the public API). To keep consistency in the codebase, I think it’s fine not to clone here. I’m open to defensive copy as well . Let me know what you think.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadataHelper.java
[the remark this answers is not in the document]
nhachicha: +1 for consistency

driver-core/src/main/com/mongodb/internal/connection/ClientMetadataHelper.java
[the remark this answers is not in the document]
nhachicha: Good point 👍

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -20,23 +20,25 @@
...
+import static com.mongodb.internal.connection.ClientMetadataHelper.updateClientMetadataDocument;
 
 /**
  * Represents metadata of the current MongoClient.
  *
  * <p>This class is not part of the public API and may be removed or changed at any time</p>
  */
 public class ClientMetadata {
-    private final ReentrantLock updateLock = new ReentrantLock();
-    private volatile BsonDocument clientMetadataBsonDocument;
```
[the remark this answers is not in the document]
vbabanin: Yes, I intended to use the `readLock(`) here. Thanks for catching that - updated accordingly!

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -36,7 +36,7 @@ public class ClientMetadata {
     private BsonDocument clientMetadataBsonDocument;
 
     public ClientMetadata(@Nullable final String applicationName, final MongoDriverInformation mongoDriverInformation) {
-        withLock(readWriteLock.writeLock(), () -> {
+        withLock(readWriteLock.readLock(), () -> {
```
nhachicha: why switching to a `readLock()`?

vbabanin: Ah, that was an oversight on my part - I meant to apply the change in `getBsonDocument `. Fixed here: 
[Fix readLock issue.](https://github.com/mongodb/mongo-java-driver/pull/1708/commits/8a582946b81c8bee9f55e394dd63bda86d444d38)

vbabanin:
`I switched from volatile` (for reads) and `volatile + lock` (for writes) to a `ReadWriteLock`. While the original approach was sufficient for thread safety, `ReadWriteLock` should be easier to understand at a glance without sacrificing performance, as the overhead is negligible.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "specifications"]
 	path = driver-core/src/test/resources/specifications
-	url = https://github.com/mongodb/specifications
+	url = https://github.com/vbabanin/specifications
+	branch = DRIVERS-2985
```
vbabanin: A link to the fork was added to connect the specification PR under review with this PR, letting unified tests to run. This should be removed once the specification PR is merged and before merging this PR.

vbabanin: Reverted back: https://github.com/mongodb/mongo-java-driver/pull/1708/commits/31ef18ef42dbb3129dd0e71218cbf2a2d6e31465

Synced with https://github.com/mongodb/specifications upstream master: https://github.com/mongodb/mongo-java-driver/pull/1708/commits/38ba5e6aeca0acb80a863f00e567b712fa263bdf

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
```diff
@@ -269,7 +270,7 @@ private boolean handleReplicaSetMemberChanged(final ServerDescription newDescrip
         }
 
         if (isStalePrimary(newDescription)) {
-            invalidatePotentialPrimary(newDescription);
+            invalidatePotentialPrimary(newDescription, new MongoStalePrimaryException("Primary marked stale due to electionId/setVersion mismatch"));
```
nhachicha: Is there a test asserting this exception?

driver-core/src/test/unit/com/mongodb/internal/connection/DefaultServerMonitorSpecification.groovy
nhachicha, line 1: why it show up empty in the diff?

nhachicha (commented):
Not sure why the changes of SDAM/Connection/Monitoring is part of this PR? The client metadata aspect LGTM but didn't review the SDAM changes

buildSrc/src/main/kotlin/conventions/testing-base.gradle.kts
jyemin: A decision to use a new assertion library seems like something that should be debated a bit more widely, rather than being brought in as part of a PR like this one. How important do you think it is to the assertions required for this PR?

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -0,0 +1,57 @@
...
+package com.mongodb.internal.connection;
+
+import com.mongodb.MongoDriverInformation;
+import com.mongodb.lang.Nullable;
+import org.bson.BsonDocument;
+
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import static com.mongodb.internal.Locks.withLock;
+import static com.mongodb.internal.connection.ClientMetadataHelper.createClientMetadataDocument;
```
jyemin: Now that we have a proper `ClientMetadata` class that is meant to be instantiated, it seems like we no longer need a separate `ClientMetadataHelper` class, instead moving the methods of that class into this one.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadataHelper.java
```diff
@@ -180,6 +181,44 @@ static boolean clientMetadataDocumentTooLarge(final BsonDocument document) {
...
     }
 
+    /**
+     * Modifies the given client metadata document by appending the driver information.
+     * Driver name and version are appended atomically to the existing driver name and version if they do not exceed
+     * {@value MAXIMUM_CLIENT_METADATA_ENCODED_SIZE} bytes.
+     * <p>
+     * Platform is appended separately to the existing platform if it does not exceed {@value MAXIMUM_CLIENT_METADATA_ENCODED_SIZE} bytes.
+     */
+    public static BsonDocument updateClientMetadataDocument(final BsonDocument clientMetadataDocument,
```
jyemin: Let's consider a slightly different algorithm for updating client metadata that shared more logic with `createClientMetadataDocument`.  

Now that we have a `ClientMetadata` class that is meant to be instantiated, we can store more than just the resulting `BsonDocument` representing the metadata.  We can also store the data from which it was created: the application name and the `MongoDriverInformation`.  If we have that, we can share all the create logic by first merging the original `MongoDriverInformation` with the appended one, and then calling `createClientMetadataDocument`.  We might need a new `MongoDriverInformation#merge` method, or it could be added to the `MongoDriverInformation.Builder`, but otherwise it reduces the code complexity a bit.

What do you think?

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
jyemin: It's also surprising that `ClientMetadataHelper.createClientMetadataDocument` is called directly in multiple places, e.g. `com.mongodb.MongoClient`, instead of just getting the `ClientMetadata` from the cluster or delegate.  Looks like it's done just to log the metadata.  Could we simplify that?

buildSrc/src/main/kotlin/conventions/testing-base.gradle.kts
vbabanin: I agree that adopting a new assertion or testing approach should generally be a broader team decision. This isn’t a critical for this PR, so I can remove it from this PR. AssertJ has already been used in the Kotlin coroutine module ([reference](https://github.com/mongodb/mongo-java-driver/blob/21065b58f60a4cce4007e747ff1c41d01994d6ff/driver-kotlin-coroutine/src/test/kotlin/com/mongodb/kotlin/client/coroutine/MockitoHelper.kt#L46)) and is present in the toml file, though it’s not widely used in the codebase. This PR made the library visible to all modules, based on its usefulness in the Hibernate repository, which was the motivation for extending its usage further.

For now, I will revert this change and we can revisit the discussion as a team.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -53,5 +75,272 @@ public void append(final MongoDriverInformation mongoDriverInformation) {
...
+        new BsonDocumentCodec().encode(new BsonBinaryWriter(buffer), document, EncoderContext.builder().build());
+        return buffer.getPosition() > MAXIMUM_CLIENT_METADATA_ENCODED_SIZE;
+    }
+
+    private enum ContainerRuntime {
+        DOCKER("docker") {
+            @Override
+            boolean isCurrentRuntimeContainer() {
+                try {
+                    return Files.exists(get(File.separator + ".dockerenv"));
```
nhachicha: The file is not always guaranteed to be created. we can add other heuristics based on `cgroup`  and environment variables (set by Kubernetes for example)
```Java
boolean isRunningInContainer() {
    // Check for /.dockerenv
    if (Files.exists(Paths.get("/.dockerenv"))) return true;

    // Check cgroup for Docker/container patterns
    try {
        List<String> lines = Files.readAllLines(Paths.get("/proc/1/cgroup"));
        for (String line : lines) {
            if (line.contains("docker") || line.contains("kubepods") || line.contains("containerd")) {
                return true;
            }
        }
    } catch (IOException ignored) {}

    // Check for common container environment variable
    if (System.getenv("container") != null) return true;

    return false;
}
```

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -53,5 +75,272 @@ public void append(final MongoDriverInformation mongoDriverInformation) {
...
+                } catch (Exception e) {
+                    return false;
+                    // NOOP. This could be a SecurityException.
+                }
+            }
+        },
+        UNKNOWN(null);
```
nhachicha: Are we interested to track other containers like Podman?

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -53,5 +75,272 @@ public void append(final MongoDriverInformation mongoDriverInformation) {
...
+            }
+            return UNKNOWN;
+        }
+    }
+
+    private enum Orchestrator {
+        K8S("kubernetes") {
+            @Override
+            boolean isCurrentOrchestrator() {
+                return FaasEnvironment.getEnv("KUBERNETES_SERVICE_HOST") != null;
```
nhachicha: Maybe also check for `/etc/kubernetes` per node if available?

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -53,5 +75,272 @@ public void append(final MongoDriverInformation mongoDriverInformation) {
...
+                : MongoDriverInformation.builder();
+        return builder
+                .driverName(MongoDriverVersion.NAME)
+                .driverVersion(MongoDriverVersion.VERSION)
+                .driverPlatform(format("Java/%s/%s", getProperty("java.vendor", "unknown-vendor"),
+                        getProperty("java.runtime.version", "unknown-version")))
+                .build();
+    }
+
+    private static String listToString(final List<String> listOfStrings) {
```
nhachicha, line 268: Can't we use 
```Java 
String result = listOfStrings.stream()
    .collect(Collectors.joining(SEPARATOR));
```
 Instead of this method?

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
vbabanin: Those are valid points and I agree. However, these changes were just moved from `ClientMetadataHelper` to consolidate them in the `ClientMetadata` instance; they were already present in the codebase and were not introduced in this PR.

You can see the diff here:
https://github.com/mongodb/mongo-java-driver/pull/1708/files#diff-ad857628989f20f23a2f410c3a2c9b49d1fc664d06a2998a489393adb5bd8e8dR161

The methods `getOperatingSystemType`, `getOperatingSystemName`, and `nameStartsWith` were also moved from `ClientMetadataHelper`. They show up as new additions because I relocated them to the bottom of the class for better organization - separating utility methods from the core logic.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
vbabanin: Replied in https://github.com/mongodb/mongo-java-driver/pull/1708#discussion_r2163082362.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
vbabanin: Replied in https://github.com/mongodb/mongo-java-driver/pull/1708#discussion_r2163082362.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
vbabanin: Replied in https://github.com/mongodb/mongo-java-driver/pull/1708#discussion_r2163082362.

driver-core/src/test/unit/com/mongodb/internal/connection/DefaultServerMonitorSpecification.groovy
vbabanin: Replied in https://github.com/mongodb/mongo-java-driver/pull/1708#issuecomment-2986875119.

driver-core/src/main/com/mongodb/internal/connection/AbstractMultiServerCluster.java
vbabanin: Replied in https://github.com/mongodb/mongo-java-driver/pull/1708#issuecomment-2986875119.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadataHelper.java
vbabanin: After our offline discussion, I checked the spec and see that we’re actually allowed to omit any values outside of `client.driver` if the appended driver name, version, and platform take up the full space.

I agree that removing `updateClientMetadata` makes code easier to reason about.

I’ve updated the code to reuse `createClientMetadataDocument`, but I think we should avoid adding public API elements purely for internal purposes. To address that, I added an internal class to encapsulate the metadata construction logic - similar to what we did with `SearchIndexRequest` as an internal representation of `SearchIndexModel`.

I pushed a commit to a separate branch https://github.com/vbabanin/mongo-java-driver/commit/a41ad6180a46014797ce4f11991b086d8d6d76d8.  What do you think?

jyemin: I like the approach.

vbabanin: I added the changes to this PR in this commit: https://github.com/mongodb/mongo-java-driver/pull/1708/commits/66a59131c2309c55c225eb5054a5f946ba7b45b2. Thank you! Let me know if you have any further suggestions.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -48,10 +79,294 @@ public BsonDocument getBsonDocument() {
...
+
+        public String getInitialDriverName() {
+            return MongoDriverVersion.NAME;
+        }
+
+        public String getInitialDriverVersion() {
+            return MongoDriverVersion.VERSION;
+        }
+
+        public List<String> getAllDriverNames() {
```
jyemin, line 360: Nit: I was initially confused by the names of these methods, as I was thinking:  we already added the initial ones, why are we now adding all of them, thinking that "all" included "initial".  But I see that it doesn't.  So consider whether this should be "Remaining" instead of "All".

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/MongoClientImpl.java
```diff
@@ -117,7 +116,8 @@ private MongoClientImpl(final MongoClientSettings settings, final MongoDriverInf
         this.externalResourceCloser = externalResourceCloser;
         this.settings = settings;
         this.closed = new AtomicBoolean();
-        BsonDocument clientMetadataDocument = createClientMetadataDocument(settings.getApplicationName(), mongoDriverInformation);
+
+        BsonDocument clientMetadataDocument = delegate.getCluster().getClientMetadata().getBsonDocument();
         LOGGER.info(format("MongoClient with metadata %s created with settings %s", clientMetadataDocument.toJson(), settings));
```
jyemin, line 122: Should we be similarly logging that new metadata was appended?

vbabanin: Yes, that makes sense, added.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
vbabanin: Just to align on the current logic:

In the current logic, we first try to add the `initial` driver name/version/platform (in this example it is `'platform'`) because it’s prioritized, and then we attempt to replace "platform" field with “all”, if it fits within the size limit:

```
tryWithLimit(client, d -> putAtPath(d, "platform", driverInformation.getInitialDriverPlatform()));
tryWithLimit(client, d -> putAtPath(d, "platform", listToString(driverInformation.getAllDriverPlatforms())));
```

The `getAllDriverPlatforms()` method does return `all` items, including the `initial` one. In `DriverInformation` constructor, we always add the `initial `items first to the lists, then `append` the rest to the same lists, if needed. See [this section](https://github.com/mongodb/mongo-java-driver/blob/66a59131c2309c55c225eb5054a5f946ba7b45b2/driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java#L320-L329) for reference.  The distinction can be confusing at first glance, but the name matches the behavior.

If renaming would improve clarity here, I’m open to making the change. 

What do you think?

jyemin: Oh... I didn't realize it was replacing, rather than appending.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientMetadataProseTest.java
```diff
@@ -0,0 +1,238 @@
...
+                .driverVersion("1.2")
+                .driverPlatform("Library Platform")
+                .build();
+
+        try (MongoClient mongoClient = createMongoClient(initialWrappingLibraryDriverInformation, getMongoClientSettingsBuilder()
+                .applyToConnectionPoolSettings(builder ->
+                        builder.maxConnectionIdleTime(1, TimeUnit.MILLISECONDS))
+                .build())) {
+
+            //TODO change get() to orElseThrow
```
nhachicha: Still relevant?

driver-sync/src/test/functional/com/mongodb/client/AbstractClientMetadataProseTest.java
```diff
@@ -0,0 +1,238 @@
...
+            String generatedDriverName = driverInformation.get("name").asString().getValue();
+            String generatedVersionName = driverInformation.get("version").asString().getValue();
+            String generatedPlatformName = initialClientMetadata.get("platform").asString().getValue();
+
+            //when
+            sleep(5); // wait for connection to become idle
+            updateClientMetadata(driverVersion, driverName, driverPlatform, mongoClient);
+
+            //then
+            //TODO change get() to orElseThrow
```
nhachicha: ditto

driver-sync/src/test/functional/com/mongodb/client/AbstractClientMetadataProseTest.java
```diff
@@ -0,0 +1,238 @@
...
+    void shouldAppendToDefaultClientMetadataWhenUpdatedAfterInitialization(@Nullable final String driverVersion,
+                                                                           @Nullable final String driverName,
+                                                                           @Nullable final String driverPlatform) {
+        //given
+        try (MongoClient mongoClient = createMongoClient(null, getMongoClientSettingsBuilder()
+                .applyToConnectionPoolSettings(builder ->
+                        builder.maxConnectionIdleTime(1, TimeUnit.MILLISECONDS))
+                .build())) {
+
+            //TODO change get() to orElseThrow
```
nhachicha: ditto

driver-sync/src/test/functional/com/mongodb/client/AbstractClientMetadataProseTest.java
```diff
@@ -0,0 +1,238 @@
...
+            String generatedDriverName = generatedDriverInformation.get("name").asString().getValue();
+            String generatedVersionName = generatedDriverInformation.get("version").asString().getValue();
+            String generatedPlatformName = initialClientMetadata.get("platform").asString().getValue();
+
+            //when
+            sleep(5); // wait for connection to become idle
+            updateClientMetadata(driverVersion, driverName, driverPlatform, mongoClient);
+
+            //then
+            //TODO change get() to orElseThrow
```
nhachicha: ditto

driver-sync/src/test/functional/com/mongodb/client/AbstractClientMetadataProseTest.java
```diff
@@ -0,0 +1,238 @@
...
+                .driverName("library")
+                .driverVersion("1.2")
+                .driverPlatform("Library Platform")
+                .build();
+
+        try (MongoClient mongoClient = createMongoClient(initialWrappingLibraryDriverInformation, getMongoClientSettingsBuilder()
+                .build())) {
+
+            //when
+            //TODO change get() to orElseThrow
```
nhachicha: ditto

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -723,6 +723,8 @@ private OperationResult executeOperation(final UnifiedTestContext context, final
                     return clientEncryptionHelper.executeEncrypt(operation);
                 case "decrypt":
                     return clientEncryptionHelper.executeDecrypt(operation);
+                case "appendMetadata":
+                    return crudHelper.executeUpdateClientMetadata(operation);
```
nhachicha, line 727: Currently, other drivers can't update their submodule deps on the sepc until they update their runner? or are they excluding the new client metadata tests?

vbabanin: Other drivers can update their submodule dependencies, however they will have to skip/ignore the new client metadata tests in their test runner until after they implemented the relevant changes from the epic.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientMetadataProseTest.java
vbabanin: Looks like Gradle build issues are now resolved (there were compilation issues with `orElseThrow`), so I’ve removed the TODOs and replaced the previous code with `orElseThrow`.

PR #1709, opened 2025-05-06
Gradle: Ensure maven artifacts are signed.

rozza:
Log a warning if the signing keys are missing.

JAVA-5872

### DIFF
--- a/buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
+++ b/buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
@@ -124,11 +124,11 @@ tasks.withType<Jar> {
 
 signing {
     if (signingKey.isPresent && signingPassword.isPresent) {
-        logger.debug("[${project.displayName}] Signing is enabled")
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
```diff
@@ -126,6 +126,9 @@ signing {
     if (signingKey.isPresent && signingPassword.isPresent) {
         logger.debug("[${project.displayName}] Signing is enabled")
         useInMemoryPgpKeys(signingKey.get(), signingPassword.get())
+        sign(publishing.publications["maven"])
+    } else {
+        logger.warn("[${project.displayName}] No Signing keys found, skipping signing configuration")
```
jyemin, line 131: Let's make both the logging statements `info`, since it's normal to publish to the local repository without signing, and Sonatype won't let us release without signatures anyway.

jyemin (approved):
LGTM after minor change to logging

PR #1716, opened 2025-05-22
Update README.md

rishitb-mongodb:
Added mention of Semantic Versioning to the README.md file. Addresses [DRIVERS-3015](https://jira.mongodb.org/browse/DRIVERS-3105).


README.md
```diff
@@ -42,6 +42,8 @@ MongoDB project, please report it according to the [instructions here](https://w
 
 ## Versioning
 
+The MongoDB Java, Kotlin and Scala drivers follow [semantic versioning](https://semver.org/) for their releases.
```
stIncMale, line 45: - We produce more libraries than just the drivers from this repository, and there is no reason to list them all here in the "Versioning" section.
- We try to follow semantic versioning, but I don't believe we have never violated it, or guarantee that we will never violate it.

Let's rephrase to something like

```suggestion
We generally follow [semantic versioning(https://semver.org/) when releasing.
```

stIncMale (changes requested):

README.md
stIncMale: 1. In that case, let's use the option B suggested in the ticket, specifying 5.6.0 as the "since" version. And say that historically the driver generally followed semantic versioning, but may not always had followed it strictly. I vaguely remember us talking about some changes technically requiring a more severe version change, but deciding not to do that.
  1.1. We also should remove all the text from the "Versioning" section of the readme, leaving there only the information about us following semantic versioning. Having that text will only introduce confusion, unless the text is identical to the semantic versioning specification.
  1.2. Semantic versioning specification itself has versions, and since the intent is to promise that we'll follow it strictly, we should specify the semantic versioning version and link to it (https://semver.org/spec/v2.0.0.html), instead of vaguely referring to semantic versioning as an idea. DRIVERS-3105 fails to require this, but I don't think it is possible to talk about strict compliance without specifying the spec version being complied with.
  1.3. I guess, we'll have to be much more careful with releases, and carefully read all the rules of the versioning scheme.
2. The suggestion about not listing all the libraries we produce from this codebase is still relevant.

jyemin: I think we can safely say that we've been strictly following semantic versioning since at least the 3.0 release.

rishitb-mongodb: Should I change this to "The MongoDB Java, Kotlin and Scala drivers follow [semantic versioning](https://semver.org/spec/v2.0.0.html) since v3.0.0 of their releases."? The original ticket was focused only on Drivers (We don't need it for other libraries/ODMs etc hence they weren't added).

jyemin: Since 3.0 was released over 10 years ago, I don't think it matters anymore whether we say that explicitly.

katcharov: Are we fine with the current PR wording,
> The MongoDB Java, Kotlin and Scala drivers follow [semantic versioning](https://semver.org/) for their releases.
then? And remove the rest of the text?

stIncMale: @rishitb-mongodb 

> The original ticket was focused only on Drivers (We don't need it for other libraries/ODMs etc hence they weren't added).

All the artifacts we release from the https://github.com/mongodb/mongo-java-driver are related to the driver. I am certain the "follow semantic versioning" rule should be applied to all of them, but we should not list all the artifacts, because
- there are quite a few of them, and just saying "Java, Kotlin and Scala drivers" does not cover them all;
- the list is not fixed in stone, and if we start producing more artifacts in the future, we will likely not update the list in the readme part about the versioning, i.e., if we list the artifacts, the list may become stale.

> 1.2 - This is to help clarify this at a high level and hence pointing to the home page feels sufficient (This was taken from our internal policy)

I am not sure what you mean. Given that there are two different versions of semantic versioning scheme, saying/implying that we strictly follow semantic versioning is meaningless, unless the scheme version itself is also specified.

If the above seems like I am getting into the unhelpful details, then we should not say/imply that we strictly follow that specific versioning scheme.

----

@katcharov 

> Are we fine with the current PR wording,
> > The MongoDB Java, Kotlin and Scala drivers follow semantic versioning for their releases.

I am proposing us not to use that wording, see above.

rishitb-mongodb: @stIncMale What other artefacts are you alluding to? We are focused on making this change ONLY for drivers and hence the drivers that are a part of this repo were added to the statement. 

Could you clarify what the suggestion is then? If I understood correctly, it is to: 
1. Remove all current text under the "Versioning" section of the Readme and
2. Replace it with "We generally follow [semantic versioning](https://semver.org/spec/v2.0.0.html) when releasing."

stIncMale: > Could you clarify what the suggestion is then?

1. Avoid listing all the artifacts we release from the https://github.com/mongodb/mongo-java-driver repository.
2. Link to https://semver.org/spec/v2.0.0.html, which actually specifies the semantic versioning scheme, instead of linking to https://semver.org, which makes it ambiguous which version of the semantic versioning scheme we follow.
3. Remove the text highlighted [here](https://github.com/mongodb/mongo-java-driver/blob/afe3600187e6508b8882902129280a62694e6f72/README.md?plain=1#L45-L55).

> What other artefacts are you alluding to?

Following is the list of all artifacts (`groupId:artifactId`) we produce (one may also see them in our [BOM](https://repo.maven.apache.org/maven2/org/mongodb/mongodb-driver-bom/5.5.0/mongodb-driver-bom-5.5.0.pom)):
```
org.mongodb:bson
org.mongodb:bson-record-codec
org.mongodb:mongodb-driver-core
org.mongodb:mongodb-crypt
org.mongodb:mongodb-driver-sync
org.mongodb:mongodb-driver-reactivestreams
org.mongodb:mongodb-driver-legacy
org.mongodb.scala:mongo-scala-bson_2.11
org.mongodb.scala:mongo-scala-bson_2.12
org.mongodb.scala:mongo-scala-bson_2.13
org.mongodb.scala:mongo-scala-driver_2.11
org.mongodb.scala:mongo-scala-driver_2.12
org.mongodb.scala:mongo-scala-driver_2.13
org.mongodb:bson-kotlin
org.mongodb:bson-kotlinx
org.mongodb:mongodb-driver-kotlin-sync
org.mongodb:mongodb-driver-kotlin-coroutine
org.mongodb:mongodb-driver-kotlin-extensions
org.mongodb:mongodb-driver-bom
```

Different intersecting subsets of them represent the following products we provide:

- Java legacy driver
- Java synchronous driver
- Java reactive streams driver
- Scala driver
- Kotlin synchronous driver
- Kotlin coroutine driver
- BSON library
  - One may argue that this is not considered a product because there is no documentation page dedicated to it. But "Java legacy driver" also does not have a dedicated page. Furthermore, "BSON library" is an implementation of https://bsonspec.org/spec.html, which may be considered a self-sufficient product.
  - **Update**: Actually, I found that we list this library [here](https://bsonspec.org/implementations.html) among other BSON implementations, and explicitly call it a "standalone BSON library", i.e., it is a product.

rishitb-mongodb: Updated based on suggestions

stIncMale: I proposed the "generally" phrasing only in the beginning of the thread, before I knew that the idea was to express strict adherence. The [answer](https://github.com/mongodb/mongo-java-driver/pull/1716#discussion_r2122465489) to the _"Could you clarify what the suggestion is then?"_ question did not propose to say "generally". Thus, let's remove it from text.

rishitb-mongodb: I misunderstood that, fixed.

stIncMale (approved):

PR #1736, opened 2025-06-10
Add namespace type support to ChangeStreamDocument.

vbabanin:
JAVA-5769

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
+++ b/driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
@@ -46,6 +46,19 @@ public final class ChangeStreamDocument<TDocument> {
     @BsonId()
     private final BsonDocument resumeToken;
     private final BsonDocument namespaceDocument;
+
+    /**
+     * Only present for ops of type 'create'.
+     * Only present when the `showExpandedEvents` change stream option is enabled.
+     *
+     * The type of the newly created object.
+     *
+     * @since 8.1.0
+     */
+    @BsonProperty("nsType")
+    private final String namespaceTypeString;
+    @BsonIgnore
+    private final NamespaceType namespaceType;
     private final BsonDocument destinationNamespaceDocument;
     private final TDocument fullDocument;
     private final TDocument fullDocumentBeforeChange;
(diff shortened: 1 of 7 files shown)


driver-sync/src/test/functional/com/mongodb/client/ChangeStreamProseTest.java
```diff
@@ -355,6 +358,72 @@ public void test19SplitChangeStreamEvents() {
         }
     }
 
+    /**
+     * Not a prose spec test. However, it is additional test case for better coverage.
+     */
+    @Test
```
vbabanin, line 364: Unified tests validate the raw `BsonDocument` returned from a `MongoCursor`, while these prose tests add additional integration coverage for working with `ChangeStreamDocument`.

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
```diff
@@ -101,12 +108,14 @@ public ChangeStreamDocument(
...
         this.namespaceDocument = namespaceDocument;
+        this.namespaceTypeString = namespaceType;
+        this.namespaceType = namespaceTypeString == null ? null : NamespaceType.fromString(namespaceType);
         this.destinationNamespaceDocument = destinationNamespaceDocument;
         this.fullDocumentBeforeChange = fullDocumentBeforeChange;
         this.documentKey = documentKey;
         this.fullDocument = fullDocument;
         this.clusterTime = clusterTime;
-        this.operationTypeString = operationTypeString;
+        this.operationTypeString = operationType;
```
katcharov, line 118: (minor) Should the field be renamed?

driver-core/src/main/com/mongodb/client/model/changestream/OperationType.java
```diff
@@ -95,9 +97,9 @@ public String getValue() {
      * Returns the ChangeStreamOperationType from the string value.
      *
      * @param operationTypeName the string value.
-     * @return the read concern
+     * @return the operation type.
```
katcharov, line 100: 👍

katcharov (approved):
LGTM

vbabanin:
> Just noticed that ChangeStreamDocumentCodec#OPERATION_TYPE_CODEC is unused and so OperationTypeCodec is dead code. Should we remove both (in this PR or a subsequent one)?

Good catch! I crated a separate PR for removing unused codec: https://github.com/mongodb/mongo-java-driver/pull/1746

driver-core/src/main/com/mongodb/client/model/changestream/ChangeStreamDocument.java
vbabanin: We have two internal fields: operationType and operationTypeString. I renamed the parameter since its type already makes it clear it’s a String, but internally we need both names to distinguish between them.

PR #1740, opened 2025-06-16
Added logging specification tests

rozza:
JAVA-4770


driver-core/src/main/com/mongodb/internal/connection/SingleServerCluster.java
```diff
@@ -58,9 +58,9 @@ public SingleServerCluster(final ClusterId clusterId, final ClusterSettings sett
         // synchronized in the constructor because the change listener is re-entrant to this instance.
         // In other words, we are leaking a reference to "this" from the constructor.
         withLock(() -> {
-            server.set(createServer(settings.getHosts().get(0)));
             publishDescription(ServerDescription.builder().state(CONNECTING).address(settings.getHosts().get(0))
```
rozza, line 62: The tests expect the toplogy description changed to connecting & unknown to be logged before the starting the server monitor log message.  


Creating the server first meant that they were the wrong way around.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -515,4 +540,94 @@ private void waitForNext() throws InterruptedException {
...
+                    TOPOLOGY, DEBUG, "Server heartbeat started", serverId.getClusterId(),
+                    asList(
+                            new LogMessage.Entry(SERVER_HOST, serverId.getAddress().getHost()),
+                            new LogMessage.Entry(SERVER_PORT, serverId.getAddress().getPort()),
+                            new LogMessage.Entry(DRIVER_CONNECTION_ID, connectionDescription.getConnectionId().getLocalValue()),
+                            new LogMessage.Entry(SERVER_CONNECTION_ID, connectionDescription.getConnectionId().getServerValue()),
+                            new LogMessage.Entry(TOPOLOGY_ID, serverId.getClusterId()),
+                            new LogMessage.Entry(AWAITED, awaited)),
+                    "Heartbeat started for {}:{} on connection with driver-generated ID {} and server-generated ID {} "
+                    + "in topology with ID {}. Awaited: {}"));
```
katcharov: This is extremely minor, but the + is not indented here, but is below.

driver-sync/src/test/functional/com/mongodb/client/unified/EventMatcher.java
```diff
@@ -336,7 +336,7 @@ public <T> void waitForServerMonitorEvents(final String client, final Class<T> e
         BsonDocument expectedEventContents = getEventContents(expectedEvent);
         try {
             serverMonitorListener.waitForEvents(expectedEventType,
-                    event -> serverMonitorEventMatches(expectedEventContents, event, null), count, Duration.ofSeconds(10));
+                    event -> serverMonitorEventMatches(expectedEventContents, event, null), count, Duration.ofSeconds(30));
```
katcharov: Why?

katcharov (approved):
LGTM!

driver-sync/src/test/functional/com/mongodb/client/unified/EventMatcher.java
rozza: Updated to 15 - I just wanted to ensure that the server monitor through runs at least once. I think it defaults to every 10 seconds.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
rozza: Nit removed 🐛

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
```diff
@@ -235,14 +235,15 @@ public static void applyCustomizations(final TestDef def) {
 
         // server-discovery-and-monitoring (SDAM)
 
+        def.retry("Flaky test,due to a race waiting for the failed heartbeat.")
```
jyemin: I hate to add new flaky tests off the bat like this.  Do you understand the root cause of the flakiness?  I'm curious also if they are flaky for other drivers.

jyemin (commented):
Strangely, when I run the tests locally, both the sync and reactive runners consistently fail the `standalone-logging: Topology lifecycle` test.  

```
java.lang.AssertionError: Expected BSON values to be equal

Assertion Context:

Value Matching Context:
   Key: message
   Expected value:
      BsonString{value='Topology description changed'}
   Actual value:
      BsonString{value='Stopped server monitoring'}

```

I imagine you're not seeing that?

rozza:
@jyemin - I do see it locally. I recall it as the result of a race condition - the test fails before the description is changed.  

I'll review again and see if there is a way to reduce the flakyness. As 5 of the 6 failures hit a similar race when heartbeats are expected to fail.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -221,66 +242,25 @@ public void run() {
 
         private ServerDescription lookupServerDescription(final ServerDescription currentServerDescription) {
             try {
-                boolean shouldStreamResponses = shouldStreamResponses(currentServerDescription);
```
katcharov, line 224: I did not do a line-by-line comparison of https://github.com/mongodb/mongo-java-driver/pull/1740/commits/7c0efa360fbc59a189f61ed41a01692e1444db51 , to ensure the refactor was correct. Is this effectively equivalent to an IntelliJ "extract" refactoring?

katcharov (approved):
Still LGTM.

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
rozza: It was using copilot to clarify the code as it wasn't very readable.

PR #1741, opened 2025-06-17
Add CSOT to OIDC.

vbabanin:
JAVA-5357


driver-core/src/main/com/mongodb/internal/connection/SaslAuthenticator.java
```diff
@@ -117,7 +117,8 @@ void authenticateAsync(final InternalConnection connection, final ConnectionDesc
 
     public abstract String getMechanismName();
 
-    protected abstract SaslClient createSaslClient(ServerAddress serverAddress);
+    //protected abstract SaslClient createSaslClient(ServerAddress serverAddress);
+    protected abstract SaslClient createSaslClient(ServerAddress serverAddress, OperationContext operationContext);
```
rozza:
```suggestion
    protected abstract SaslClient createSaslClient(ServerAddress serverAddress, OperationContext operationContext);
```

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
```diff
@@ -118,8 +124,22 @@ public OidcAuthenticator(final MongoCredentialWithCache credential,
...
-        return isHumanCallback() ? HUMAN_CALLBACK_TIMEOUT : CALLBACK_TIMEOUT;
+    private Duration getCallbackTimeout(final TimeoutContext timeoutContext) {
+        if (isHumanCallback()) {
+            return HUMAN_CALLBACK_TIMEOUT;
+        }
+
+        if (timeoutContext.hasTimeoutMS()) {
+            return assertNotNull(timeoutContext.getTimeout()).call(TimeUnit.MILLISECONDS,
+                    () -> {
+                        Assertions.fail();
```
rozza: From the tests it looks like it should fall back to `serverSelectionTimeoutMS` - if theres an infinite timeout - so how does that logic work?

Also it might be best to put an assertion message incase a future regression means this code path is hit - potentially would make debugging it easier.

rozza (changes requested):
One question and one nit

driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java
vbabanin: I initially assumed, based on the spec, that `serverSelectionTimeoutMS` couldn’t be infinite, so I added the assertion. However, our implementation allows infinite timeouts for `serverSelectionTimeoutMS`. Given this, i see two options:

1. If both `timeoutMs` and `serverSelectionTimeoutMS` are infinite, throw an exception.
2. The OIDC spec doesn’t specify behavior for infinite timeouts, but we could pass the maximum `Duration` value (using `ChronoUnit.FOREVER.getDuration()`) to the OIDC callback. While not technically infinite, it’s large enough to represent an infinite timeout, aligning with both the OIDC spec (as it is technically `Long.MAX_VALUE` and not infinite) and our current API for infinite `serverSelectionTimeoutMS`.

I think option 2 is preferable, so I’ve added it in this commit: https://github.com/mongodb/mongo-java-driver/pull/1741/commits/cc5469d183dee15a464bd54047aa910a7a5fd0d5. Let me know your thoughts.

PR #1743, opened 2025-06-17
Improve test clarity and performance.

vbabanin:
JAVA-5898

### DIFF
--- a/driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonOutputTest.java
+++ b/driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonOutputTest.java
@@ -70,38 +70,76 @@ final class ByteBufferBsonOutputTest {
     private static final List<Integer> ALL_CODE_POINTS_EXCLUDING_SURROGATES = Stream.concat(
                     range(1, MIN_HIGH_SURROGATE).boxed(),
                     rangeClosed(MAX_LOW_SURROGATE + 1, MAX_CODE_POINT).boxed())
+            .filter(codePoint -> codePoint < 128 || codePoint % 30 == 0) // only subset of code points to speed up testing
             .collect(toList());
+
     private static final List<Integer> ALL_SURROGATE_CODE_POINTS = Stream.concat(
             range(MIN_LOW_SURROGATE, MAX_LOW_SURROGATE).boxed(),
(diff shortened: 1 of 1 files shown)


driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonInputTest.java
```diff
@@ -54,37 +54,68 @@ class ByteBufferBsonInputTest {
...
+                            public int arrayOffset() {
+                                return Assertions.fail("arrayOffset() is called, when isBackedByArray() returns false");
+                            }
+                        }
+                )
         );
     }
 
-    @ParameterizedTest
+    private static BufferProvider createBufferProvider(final String bufferName, final BufferProvider bufferProvider) {
```
NathanQingyangXu: I am not sure about my point. Is it true that bufferDescription is more appropriate naming? Saw it is used to overwrite toString.

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonInputTest.java
```diff
@@ -660,7 +691,7 @@ void shouldSkipCStringWhenMultipleNullTerminationPresent(final byte[] cStringByt
         }
     }
 
-    @ParameterizedTest
+    @ParameterizedTest(name = "should read skip CString when multiple null termination present within buffer. BufferProvider={0}")
```
NathanQingyangXu: Is "multiple null terminations" preferable?

NathanQingyangXu (changes requested):

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonInputTest.java
vbabanin: I agree, changed to `bufferDescription`.

driver-core/src/test/unit/com/mongodb/internal/connection/ByteBufferBsonInputTest.java
vbabanin: Yes, it has to be plural. And now I think the correct term here is “terminators,” not “termination,” so I changed it to “terminators.”

NathanQingyangXu (approved):

PR #1744, opened 2025-06-19
Ignore maxWaitTime when CSOT is enabled.

vbabanin:
JAVA-5409

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
@@ -42,6 +42,7 @@ import com.mongodb.event.ConnectionPoolCreatedEvent;
 import com.mongodb.event.ConnectionPoolListener;
 import com.mongodb.event.ConnectionPoolReadyEvent;
 import com.mongodb.event.ConnectionReadyEvent;
+import com.mongodb.internal.TimeoutContext;
 import com.mongodb.internal.VisibleForTesting;
 import com.mongodb.internal.async.SingleResultCallback;
 import com.mongodb.internal.connection.SdamServerDescriptionManager.SdamIssue;
@@ -98,6 +99,7 @@ import static com.mongodb.assertions.Assertions.notNull;
 import static com.mongodb.event.ConnectionClosedEvent.Reason.ERROR;
 import static com.mongodb.internal.Locks.lockInterruptibly;
 import static com.mongodb.internal.Locks.withLock;
+import static com.mongodb.internal.TimeoutContext.createMongoTimeoutException;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
 import static com.mongodb.internal.async.ErrorHandlingResultCallback.errorHandlingCallback;
 import static com.mongodb.internal.connection.ConcurrentPool.INFINITE_SIZE;
@@ -193,7 +195,7 @@ final class DefaultConnectionPool implements ConnectionPool {
         Timeout waitQueueTimeout =  operationContext.getTimeoutContext().startWaitQueueTimeout(checkoutStart);
         try {
             stateAndGeneration.throwIfClosedOrPaused();
-            PooledConnection connection = getPooledConnection(waitQueueTimeout, checkoutStart);
+            PooledConnection connection = getPooledConnection(waitQueueTimeout, checkoutStart, operationContext);
             if (!connection.opened()) {
                 connection = openConcurrencyLimiter.openOrGetAvailable(operationContext, connection, waitQueueTimeout, checkoutStart);
             }
@@ -225,13 +227,13 @@ final class DefaultConnectionPool implements ConnectionPool {
             eventSendingCallback.onResult(null, e);
             return;
         }
-        asyncWorkManager.enqueue(new Task(maxWaitTimeout, checkoutStart, t -> {
+        asyncWorkManager.enqueue(new Task(maxWaitTimeout, checkoutStart, operationContext, t -> {
             if (t != null) {
                 eventSendingCallback.onResult(null, t);
             } else {
                 PooledConnection connection;
                 try {
-                    connection = getPooledConnection(maxWaitTimeout, checkoutStart);
+                    connection = getPooledConnection(maxWaitTimeout, checkoutStart, operationContext);
                 } catch (Exception e) {
                     eventSendingCallback.onResult(null, e);
                     return;
@@ -330,7 +332,9 @@ final class DefaultConnectionPool implements ConnectionPool {
         return stateAndGeneration.generation();
     }
 
-    private PooledConnection getPooledConnection(final Timeout waitQueueTimeout, final StartTime startTime) throws MongoTimeoutException {
+    private PooledConnection getPooledConnection(final Timeout waitQueueTimeout,
+                                                 final StartTime startTime,
+                                                 final OperationContext operationContext) throws MongoTimeoutException {
         try {
             UsageTrackingInternalConnection internalConnection = waitQueueTimeout.call(NANOSECONDS,
                     () -> pool.get(-1L, NANOSECONDS),
@@ -345,7 +349,7 @@ final class DefaultConnectionPool implements ConnectionPool {
             }
             return new PooledConnection(internalConnection);
         } catch (MongoTimeoutException e) {
-            throw createTimeoutException(startTime);
+            throw createTimeoutException(startTime, operationContext.getTimeoutContext());
         }
     }
 
@@ -359,13 +363,14 @@ final class DefaultConnectionPool implements ConnectionPool {
         return internalConnection == null ? null : new PooledConnection(internalConnection);
     }
 
-    private MongoTimeoutException createTimeoutException(final StartTime startTime) {
+    private MongoTimeoutException createTimeoutException(final StartTime startTime, final TimeoutContext timeoutContext) {
         long elapsedMs = startTime.elapsed().toMillis();
         int numPinnedToCursor = pinnedStatsManager.getNumPinnedToCursor();
         int numPinnedToTransaction = pinnedStatsManager.getNumPinnedToTransaction();
         if (numPinnedToCursor == 0 && numPinnedToTransaction == 0) {
-            return new MongoTimeoutException(format("Timed out after %d ms while waiting for a connection to server %s.",
-                    elapsedMs, serverId.getAddress()));
+            String errorMessage = format("Timed out after %d ms while waiting for a connection to server %s.",
+                    elapsedMs, serverId.getAddress());
+            return timeoutContext.hasTimeoutMS() ? createMongoTimeoutException(errorMessage) : new MongoTimeoutException(errorMessage);
         } else {
             int maxSize = pool.getMaxSize();
             int numInUse = pool.getInUseCount();
@@ -394,12 +399,13 @@ final class DefaultConnectionPool implements ConnectionPool {
             int numOtherInUse = numInUse - numPinnedToCursor - numPinnedToTransaction;
             assertTrue(numOtherInUse >= 0);
             assertTrue(numPinnedToCursor + numPinnedToTransaction + numOtherInUse <= maxSize);
-            return new MongoTimeoutException(format("Timed out after %d ms while waiting for a connection to server %s. Details: "
+            String errorMessage = format("Timed out after %d ms while waiting for a connection to server %s. Details: "
                             + "maxPoolSize: %s, connections in use by cursors: %d, connections in use by transactions: %d, "
                             + "connections in use by other operations: %d",
                     elapsedMs, serverId.getAddress(),
                     sizeToString(maxSize), numPinnedToCursor, numPinnedToTransaction,
-                    numOtherInUse));
+                    numOtherInUse);
+            return timeoutContext.hasTimeoutMS() ? createMongoTimeoutException(errorMessage) : new MongoTimeoutException(errorMessage);
         }
     }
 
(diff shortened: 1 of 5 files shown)


driver-core/src/main/com/mongodb/internal/time/Timeout.java
```diff
@@ -40,6 +40,8 @@
 /**
  * A Timeout is a "deadline", point in time by which something must happen.
  *
+ * Implementations of this interface must be immutable.
+ *
  * @see TimePoint
  */
 public interface Timeout {
```
stIncMale, line 47: - Is this change needed because of a change done in this PR, or is it simply something that you noticed and improved?
- Shouldn't the same be documented for `StartTime`?

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -208,7 +210,7 @@ public InternalConnection get(final OperationContext operationContext) {
     @Override
     public void getAsync(final OperationContext operationContext, final SingleResultCallback<InternalConnection> callback) {
         StartTime checkoutStart = connectionCheckoutStarted(operationContext);
-        Timeout maxWaitTimeout = checkoutStart.timeoutAfterOrInfiniteIfNegative(settings.getMaxWaitTime(NANOSECONDS), NANOSECONDS);
+        Timeout maxWaitTimeout = operationContext.getTimeoutContext().startWaitQueueTimeout(checkoutStart);
```
stIncMale: `maxWaitTimeout` is called `waitQueueTimeout` in the `get` method. They should have equal names, and given that we obtain them by calling `startWaitQueueTimeout`, the name `waitQueueTimeout` seems to be more suitable. If it's not, maybe we should rename the method also?

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -330,7 +332,9 @@ public int getGeneration() {
         return stateAndGeneration.generation();
     }
 
-    private PooledConnection getPooledConnection(final Timeout waitQueueTimeout, final StartTime startTime) throws MongoTimeoutException {
+    private PooledConnection getPooledConnection(final Timeout waitQueueTimeout,
+                                                 final StartTime startTime,
+                                                 final OperationContext operationContext) throws MongoTimeoutException {
```
stIncMale: Given that this methods uses only `operationContext.getTimeoutContext()`, let's accept `TimeoutContext` instead of `OperationContext`? The same applies to the `acquirePermitOrGetAvailableOpenedConnection` method and the constructor of `DefaultConnectionPool.Task`.

Interestingly, `createTimeoutException` rightly accepts `TimeoutContext`, not `OperationContext`.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
```diff
@@ -359,13 +363,14 @@ private PooledConnection getPooledConnectionImmediate() {
...
+    private MongoTimeoutException createTimeoutException(final StartTime startTime, final TimeoutContext timeoutContext) {
         long elapsedMs = startTime.elapsed().toMillis();
         int numPinnedToCursor = pinnedStatsManager.getNumPinnedToCursor();
         int numPinnedToTransaction = pinnedStatsManager.getNumPinnedToTransaction();
         if (numPinnedToCursor == 0 && numPinnedToTransaction == 0) {
-            return new MongoTimeoutException(format("Timed out after %d ms while waiting for a connection to server %s.",
-                    elapsedMs, serverId.getAddress()));
+            String errorMessage = format("Timed out after %d ms while waiting for a connection to server %s.",
+                    elapsedMs, serverId.getAddress());
+            return timeoutContext.hasTimeoutMS() ? createMongoTimeoutException(errorMessage) : new MongoTimeoutException(errorMessage);
```
stIncMale: Let's extract the `errorMessage` variable to a place before the `if` block, initialize it in the `if` and the `else` blocks, and then do `return timeoutContext.hasTimeoutMS() ? createMongoTimeoutException(errorMessage) : new MongoTimeoutException(errorMessage);` as the last statement in the `createTimeoutException` method. This way we will avoid duplicating this last statement.

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
```diff
@@ -133,6 +135,33 @@ public void shouldThrowOnTimeout() throws InterruptedException {
         assertTrue(connectionGetter.isGotTimeout());
     }
 
+    @Test
+    public void shouldNotUseMaxAwaitTimeMSOnWhenTimeoutMsIsSet() throws InterruptedException {
```
stIncMale:
```suggestion
    public void shouldNotUseMaxAwaitTimeMSWhenTimeoutMsIsSet() throws InterruptedException {
```

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
```diff
@@ -781,6 +783,118 @@ public void shouldIgnoreWtimeoutMsOfWriteConcernToInitialAndSubsequentCommitTran
...
+            assertThrows(MongoOperationTimeoutException.class, () -> collection.find().first());
+        }
+    }
+
+    /**
+     * Not a prose spec test. However, it is additional test case for better coverage.
+     */
+    @Test
+    @DisplayName("Should ignore waitQueueTimeoutMS when timeoutMS is set")
+    public void shouldUseWaitQueueTimeoutMSWhenTimeoutIsNotSet() {
```
stIncMale, line 855: `@DisplayName` does not match the name/behavior of the test.

driver-core/src/test/functional/com/mongodb/internal/connection/DefaultConnectionPoolTest.java
```diff
@@ -133,6 +135,33 @@ public void shouldThrowOnTimeout() throws InterruptedException {
...
+        provider.ready();
+        TimeoutSettings timeoutSettings = TIMEOUT_SETTINGS
+                .withTimeout(100L, MILLISECONDS)
+                .withMaxWaitTimeMS(50);
+
+        InternalConnection internalConnection = provider.get(createOperationContext(timeoutSettings));
+
+        // when
+        TimeoutTrackingConnectionGetter connectionGetter = new TimeoutTrackingConnectionGetter(provider, timeoutSettings);
+        new Thread(connectionGetter).start();
```
stIncMale: Instead of starting a new thread and never making sure it terminates, let's use `cachedExecutor`, which is properly terminated in `cleanup`.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
```diff
@@ -781,6 +783,118 @@ public void shouldIgnoreWtimeoutMsOfWriteConcernToInitialAndSubsequentCommitTran
...
+                .timeout(500, TimeUnit.MILLISECONDS)
+                .applyToConnectionPoolSettings(builder -> builder
+                        .maxWaitTime(1, TimeUnit.MILLISECONDS)
+                        .maxSize(1)
+                ))) {
+            MongoCollection<Document> collection = mongoClient.getDatabase(namespace.getDatabaseName())
+                    .getCollection(namespace.getCollectionName());
+
+            collectionHelper.runAdminCommand("{"
+                    + "    configureFailPoint: \"failCommand\","
```
stIncMale, line 798: We have `com.mongodb.client.FailPoint` for creating fail points in a way that cleans them up. Let's use it here and in other new tests.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
```diff
@@ -781,6 +783,118 @@ public void shouldIgnoreWtimeoutMsOfWriteConcernToInitialAndSubsequentCommitTran
...
+                    + "    mode: { times: 1},"
+                    + "    data: {"
+                    + "        failCommands: [\"find\" ],"
+                    + "        blockConnection: true,"
+                    + "        blockTimeMS: " + 300
+                    + "    }"
+                    + "}");
+
+            ExecutorService executor = Executors.newSingleThreadExecutor();
```
stIncMale: Let's properly shut `executor` down, and wait for shutting down to complete. We may also do that in an `@AfterEach` method, whatever seems more appropriate. This suggestion applies to all new tests in this file.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
```diff
@@ -781,6 +783,118 @@ public void shouldIgnoreWtimeoutMsOfWriteConcernToInitialAndSubsequentCommitTran
...
+    @Test
+    @DisplayName("Should throw MongoOperationTimeoutException when connection is not available and timeoutMS is set")
+    public void shouldThrowOperationTimeoutExceptionWhenConnectionIsNotAvailableAndTimeoutMSIsSet() {
+        assumeTrue(serverVersionAtLeast(4, 4));
+
+        //given
+        try (MongoClient mongoClient = createMongoClient(getMongoClientSettingsBuilder()
+                .timeout(100, TimeUnit.MILLISECONDS)
+                .applyToConnectionPoolSettings(builder -> builder
+                        .maxWaitTime(1, TimeUnit.MILLISECONDS)
```
stIncMale: Why do we set `maxWaitTime` here?

driver-core/src/main/com/mongodb/internal/time/Timeout.java
vbabanin: This wasn’t required by changes in this PR, just a small improvement I noticed while reviewing the Timeout code. I thought documenting immutability in the Javadoc would be useful for anyone double-checking this detail, without needing to examine the implementations.

I’ve also added the note for `StartTime` as you suggested.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
vbabanin: Good catch! Even though it is ignored,  we don't need it here. It was a leftover from coping settings from the test above.  Removed.

driver-core/src/main/com/mongodb/internal/connection/DefaultConnectionPool.java
vbabanin: Renamed to `maxWaitTimeout` everywhere in `DefaultConnectionPool`, as that’s the prevailing naming convention in the codebase. There are still some discrepancies in the javadocs, but this naming keeps it consistent with most usages internally.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
vbabanin: There are cases where the failpoint must be configured after a specific command has executed (see, for example, commit transactions [here](https://github.com/mongodb/mongo-java-driver/blob/820b131797bac6e3627133743db4b1e0aa502ed2/driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java#L1001-L1023)). In that case, using the FailPoint class would require another nested `try`, which would make the test a bit harder to read. I do agree that using `FailPoint` in a `try` at the beginning of the test is more explicit and clear, which i assume the majority of use-cases do.

In current tests case, we could certainly use `FailPoint` to make it explicit and clear, but it would introduce inconsistency in this class and might raise questions of what to use for others adding new tests to this class.

Ideally, it would be beneficial if a `FailPoint` instance could be injected with an annotation before each test and cleaned up automatically, so that we could enable it at any point in the test as needed without `try/catch` via `FailPoint.enable()` method.

I suggest we address this in a separate ticket to unify the approach in this (and probably other test classes), and to consider automatically injected `FailPoint`.

What do you think?

stIncMale: I originally did not notice that `AbstractClientSideOperationsTimeoutProseTest.tearDown` cleans the failpoints. Given that, `FailPoint` is not needed here.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
```diff
@@ -979,6 +1090,10 @@ public void tearDown() {
                 // ignore
             }
         }
+
+        executor.shutdownNow();
+        //noinspection ResultOfMethodCallIgnored
+        executor.awaitTermination(MAX_VALUE, NANOSECONDS);
```
stIncMale: `executor` may be `null` at this point, as it is initialized in the `setUp` method, and not in the constructor:


```suggestion
        if (executor != null) {
            executor.shutdownNow();
            //noinspection ResultOfMethodCallIgnored
            executor.awaitTermination(MAX_VALUE, NANOSECONDS);
        }
```

vbabanin: Good catch!  Applied, thank you!

stIncMale (approved):

PR #1747, opened 2025-06-23
Added nexus-publish plugin

rozza:
The first commit simplifies use of plugins in buildSrc by upgrading the `dev.panuszewski.typesafe-conventions` plugin to: `0.7.3`.

The second commit introduces the `nexus-publish` to handle publishing to sonatype. Once the migration to central.sonatype.com it should be a case of updating the username and password.

JAVA-5899

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -17,4 +17,25 @@
 plugins {
     id("eclipse")
     id("idea")
+    alias(libs.plugins.nexus.publish)
+}
+
+val nexusUsername: Provider<String> = providers.gradleProperty("nexusUsername")
+val nexusPassword: Provider<String> = providers.gradleProperty("nexusPassword")
+
+nexusPublishing {
+    packageGroup = "org.mongodb"
+    repositories {
+        sonatype {
+            username = nexusUsername
+            password = nexusPassword
+
+            snapshotRepositoryUrl.set(uri("https://oss.sonatype.org/content/repositories/snapshots/"))
+            nexusUrl.set(uri("https://oss.sonatype.org/service/local/"))
+
+            // central portal URLs
+//            nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
+//            snapshotRepositoryUrl.set(uri("https://central.sonatype.com/repository/maven-snapshots/"))
+        }
+    }
 }
--- a/buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
+++ b/buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
@@ -28,8 +28,6 @@ plugins {
 
 val signingKey: Provider<String> = providers.gradleProperty("signingKey")
 val signingPassword: Provider<String> = providers.gradleProperty("signingPassword")
-val nexusUsername: Provider<String> = providers.gradleProperty("nexusUsername")
-val nexusPassword: Provider<String> = providers.gradleProperty("nexusPassword")
 @Suppress("UNCHECKED_CAST") val gitVersion: Provider<String> = project.findProperty("gitVersion") as Provider<String>
 
 tasks.withType<AbstractPublishToMaven>().configureEach {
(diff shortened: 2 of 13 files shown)


build.gradle.kts
```diff
@@ -17,4 +17,21 @@
...
+val nexusPassword: Provider<String> = providers.gradleProperty("nexusPassword")
+
+nexusPublishing {
+    repositories {
+        sonatype {
+            username = nexusUsername
+            password = nexusPassword
+
+            // central portal URLs
+//            nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
```
jyemin, line 37: With this commented out, what happens if we attempt to publish?  Will it publish to existing Sonatype portal, or just fail?

build.gradle.kts
```diff
@@ -17,4 +17,21 @@
 plugins {
     id("eclipse")
     id("idea")
+    alias(libs.plugins.nexus.publish)
```
jyemin, line 20: Curious why all this can't be in `publishing.gradle.kts`.  Can you explain?

buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
```diff
@@ -141,7 +130,7 @@ tasks.register("publishSnapshots") {
     description = "Publishes snapshots to Sonatype"
 
     if (version.toString().endsWith("-SNAPSHOT")) {
-        dependsOn(tasks.withType<PublishToMavenRepository>())
+        dependsOn(tasks.named("publishToSonatype"))
```
jyemin, line 125: Curious why this change is needed.

I'm also not clear how it "knows" to use the snapshot repo.  Is it keying off the version string?

buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
```diff
@@ -168,7 +157,7 @@ tasks.register("publishArchives") {
         }
     }
     if (gitVersionMatch) {
-        dependsOn(tasks.withType<PublishToMavenRepository>())
+        dependsOn(tasks.named("publishToSonatype"))
```
jyemin, line 152: Should we add `closeAndReleaseSonatypeStagingRepository` or at least `closeSonatypeStagingRepository`?

build.gradle.kts
rozza: It will publish to the deprecated sonatype urls.

build.gradle.kts
rozza: Its hidden a little in the second commit message:

> As the nexus-publish plugin can only be configured in the root project
the username and password logic was moved out of conventions/publishing.gradle.kts
and into the root buid.gradle.kts.
> 
> Note: The nexus-publish plugin builds upon the gradle maven-publish
plugin so even though the configuration is in the root project,
it will only publish projects that use the publishing convention.

buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
rozza: Its to use the plugins tasks - rather than the default maven task. (the sonatype url was removed).

buildSrc/src/main/kotlin/conventions/publishing.gradle.kts
rozza: We can in the future - perhaps after the initial release?

rozza:
@jyemin & @katcharov  Just to clarify this PR can be accepted before the migration to the central portal. It keeps the existing behaviour (I tested via publishing snapshots). 

Once the namespace migration is done the work then would be to update the sonatype urls and potentially auto close and release the artifacts.

jyemin:
How far were you able to get with testing these changes without actually cutting over to `central.sonatype.com`?  Were you able to test both snapshot and regular publish to legacy Sonatype?

rozza:
Manually tested an alpha release from my local machine - had to explictly set the `publishingGroup` - so was worth testing - force pushed an update.

Only difference of note is it created 3 staged groups (rather than just one) - one for the initial full driver and then one for each of the scala 2.11 and 2.12 variants (which are separate calls to `publishArchives`)

jyemin (approved):
LGTM!

PR #1749, opened 2025-06-24
Ensure local build artifacts are created

rozza:
When migrating to the nexus-publish plugin I changed the dependency to just use `publishSonatype` this stopped the local build artifacts being created and tracing artifacts in evergreen relies on them.

JAVA-5904


jyemin (approved):
LGTM

PR #1758, opened 2025-07-03
Added operation (aka command name) to structured logging

rozza:
JAVA-5197

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -89,6 +89,10 @@ public class OperationContext {
         return new OperationContext(id, requestContext, sessionContext, timeoutContext, serverDeprioritization, serverApi, operationName);
     }
 
+    public OperationContext withOperationName(final String operationName) {
+        return new OperationContext(id, requestContext, sessionContext, timeoutContext, serverDeprioritization, serverApi, operationName);
+    }
+
     public long getId() {
         return id;
     }
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -49,10 +49,17 @@ public class OperationContext {
     private final TimeoutContext timeoutContext;
     @Nullable
     private final ServerApi serverApi;
+    @Nullable
+    private final String commandName;
 
     public OperationContext(final RequestContext requestContext, final SessionContext sessionContext, final TimeoutContext timeoutContext,
             @Nullable final ServerApi serverApi) {
-        this(NEXT_ID.incrementAndGet(), requestContext, sessionContext, timeoutContext, new ServerDeprioritization(), serverApi);
+        this(requestContext, sessionContext, timeoutContext, serverApi, null);
+    }
+
+    public OperationContext(final RequestContext requestContext, final SessionContext sessionContext, final TimeoutContext timeoutContext,
+            @Nullable final ServerApi serverApi, @Nullable final String commandName) {
+        this(NEXT_ID.incrementAndGet(), requestContext, sessionContext, timeoutContext, new ServerDeprioritization(), serverApi, commandName);
     }
 
     public static OperationContext simpleOperationContext(
@@ -61,7 +68,8 @@ public class OperationContext {
                 IgnorableRequestContext.INSTANCE,
                 NoOpSessionContext.INSTANCE,
                 new TimeoutContext(timeoutSettings),
-                serverApi);
+                serverApi,
+                null);
     }
 
     public static OperationContext simpleOperationContext(final TimeoutContext timeoutContext) {
@@ -69,15 +77,16 @@ public class OperationContext {
                 IgnorableRequestContext.INSTANCE,
                 NoOpSessionContext.INSTANCE,
                 timeoutContext,
+                null,
                 null);
     }
 
     public OperationContext withSessionContext(final SessionContext sessionContext) {
-        return new OperationContext(id, requestContext, sessionContext, timeoutContext, serverDeprioritization, serverApi);
+        return new OperationContext(id, requestContext, sessionContext, timeoutContext, serverDeprioritization, serverApi, commandName);
     }
 
     public OperationContext withTimeoutContext(final TimeoutContext timeoutContext) {
-        return new OperationContext(id, requestContext, sessionContext, timeoutContext, serverDeprioritization, serverApi);
+        return new OperationContext(id, requestContext, sessionContext, timeoutContext, serverDeprioritization, serverApi, commandName);
     }
 
     public long getId() {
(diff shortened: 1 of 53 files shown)


driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -49,10 +49,17 @@ public class OperationContext {
     private final TimeoutContext timeoutContext;
     @Nullable
     private final ServerApi serverApi;
+    @Nullable
+    private final String commandName;
```
nhachicha: Shouldn't this be named `operationName` instead? since one operation can issue multiple commands potentially...

driver-core/src/main/com/mongodb/internal/operation/AbortTransactionOperation.java
```diff
@@ -31,6 +31,7 @@
  * <p>This class is not part of the public API and may be removed or changed at any time</p>
  */
 public class AbortTransactionOperation extends TransactionOperation {
+    private static final String COMMAND_NAME = "abortTransaction";
```
nhachicha, line 34: Are the names of operations standardised via the spec? (since we will enable spec tests for those)

driver-core/src/main/com/mongodb/internal/operation/FindAndDeleteOperation.java
```diff
@@ -95,5 +95,4 @@ protected FieldNameValidator getFieldNameValidator() {
     protected void specializeCommand(final BsonDocument commandDocument, final ConnectionDescription connectionDescription) {
         commandDocument.put("remove", BsonBoolean.TRUE);
     }
-
```
nhachicha, line 98: `@overrides` missing for `getFieldNameValidator` &  `specializeCommand` btw

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -84,17 +85,20 @@ public class MixedBulkWriteOperation implements AsyncWriteOperation<BulkWriteRes
...
     private Boolean bypassDocumentValidation;
+    private String commandName;
     private BsonValue comment;
     private BsonDocument variables;
 
     public MixedBulkWriteOperation(final MongoNamespace namespace, final List<? extends WriteRequest> writeRequests,
             final boolean ordered, final WriteConcern writeConcern, final boolean retryWrites) {
+        notNull("writeRequests", writeRequests);
+        isTrueArgument("writeRequests is not an empty list", !writeRequests.isEmpty());
+        this.commandName = notNull("commandName", writeRequests.get(0).getType().toString().toLowerCase(Locale.ROOT));
```
nhachicha, line 96: Why using the first request only (there could be a mix of `Insert/Update/Delete`)? Why can't we use `bulkWrite` as a name (or something similar) as described in [ReadOperation.java](https://github.com/mongodb/mongo-java-driver/pull/1758/files#diff-b4ecd17ed39bce3aff4da4d50cf2ba168ef741e368cbdf9591ca6fb8a48e9ed5)?

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -420,6 +429,7 @@ private BsonDocument executeCommand(
             final OperationContext operationContext,
             final Connection connection,
             final BulkWriteBatch batch) {
+        commandName = batch.getCommand().getFirstKey();
```
nhachicha, line 432: ditto

driver-legacy/src/main/com/mongodb/LegacyMixedBulkWriteOperation.java
```diff
@@ -79,29 +75,30 @@ static LegacyMixedBulkWriteOperation createBulkWriteOperationForDelete(final Mon
 
     private LegacyMixedBulkWriteOperation(final MongoNamespace namespace, final boolean ordered, final WriteConcern writeConcern,
             final boolean retryWrites, final List<? extends WriteRequest> writeRequests, final WriteRequest.Type type) {
-        isTrueArgument("writeRequests not empty", !writeRequests.isEmpty());
-        this.writeRequests = notNull("writeRequests", writeRequests);
+        notNull("writeRequests", writeRequests);
```
nhachicha, line 78: These checks are done in `MixedBulkWriteOperation` already, should we remove them?

driver-core/src/main/com/mongodb/internal/operation/AbortTransactionOperation.java
rozza: Yeah the spec says it should be the command name sent to the server.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
rozza: This is the odd one - updateOne, InsertOne all boil down to a bulk write operation.  We only send the name of the first command sent. There is a `bulkWrite` command however, that uses the new bulk write command.

Note: Each command will also issue its command name (it already does). This is for the server selection part - which is triggered by the initial command.

driver-legacy/src/main/com/mongodb/LegacyMixedBulkWriteOperation.java
rozza: No harm in leaving - we generally don't touch too much legacy code.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
rozza: Its really the initial command name - I did originally have it as `operationName` but changed it. Happy to rename if you think that works.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
rozza: Fixed - updated to use `operationContext.withOperationName` on each execute.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
rozza: > The name of the operation for which a server is being selected. When server selection is being performed to select a server for a command, this MUST be the command name.

Have now plumbed this in correctly (see previous).

driver-core/src/main/com/mongodb/internal/operation/FindAndDeleteOperation.java
rozza: Done

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
rozza: Updated - it matches more with `operation` in the logging spec even though its the current command name.

nhachicha (approved):
LGTM 👍

PR #1761, opened 2025-07-07
Propagate timeout errors to callback.

vbabanin:
JAVA-5906


driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
```diff
@@ -88,7 +89,7 @@ protected void setChannel(final ExtendedAsynchronousByteChannel channel) {
 
     @Override
     public void writeAsync(final List<ByteBuf> buffers, final OperationContext operationContext,
-            final AsyncCompletionHandler<Void> handler) {
+                           final AsyncCompletionHandler<Void> handler) {
```
NathanQingyangXu, line 92: Do we really need the above change? Seems the aim is to vertical alignment? Usually parameters are put into a new  line because the previous parameters are too long, and maintaining vertical alignment won't help when the remaining parameters are long and need to be split again (the above approach shrink the real estate as a downside).

NathanQingyangXu: nvm. saw it has been done consistently when the second-line parameters don't need to be split

driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
```diff
@@ -189,8 +190,11 @@ public void failed(final Throwable t) {
 
     private class AsyncWritableByteChannelAdapter {
         void write(final ByteBuffer src, final OperationContext operationContext, final AsyncCompletionHandler<Void> handler) {
-            getChannel().write(src, operationContext.getTimeoutContext().getWriteTimeoutMS(), MILLISECONDS, null,
-                    new AsyncWritableByteChannelAdapter.WriteCompletionHandler(handler));
+            beginAsync().thenRun((c) -> {
+                long writeTimeoutMS = operationContext.getTimeoutContext().getWriteTimeoutMS();
+                getChannel().write(src, writeTimeoutMS, MILLISECONDS, null,
+                        new AsyncWritableByteChannelAdapter.WriteCompletionHandler(c.asHandler()));
+            }).finish(handler.asCallback());
```
NathanQingyangXu, line 197: I might be wrong.
For the above change, to me it seems there is an implication in the sense that in vast majority cases the `handler` would be invoked twice (one for writeCompletionHandler and the other one is the the newly-added finish).

Would it have unexpected undesirable side effect?

To me the only reason to apply the `finish()` is to guard against the following statement throwing exception?
```
long writeTimeoutMS = operationContext.getTimeoutContext().getWriteTimeoutMS();
```
otherwise it seems the original one-liner is self-sufficient:
```
getChannel().write(src, operationContext.getTimeoutContext().getWriteTimeoutMS(), MILLISECONDS, null,
                    new AsyncWritableByteChannelAdapter.WriteCompletionHandler(handler));
```
if exception is thrown from this method, seems that is a bug we should fix instead.

driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
```diff
@@ -231,17 +235,20 @@ private BasicCompletionHandler(final ByteBuf dst, final OperationContext operati
...
+                    throw new MongoSocketReadException("Prematurely reached end of stream", serverAddress);
+                } else if (!localByteBuf.hasRemaining()) {
+                    localByteBuf.flip();
+                    c.complete(localByteBuf);
+                } else {
+                    long readTimeoutMS = operationContext.getTimeoutContext().getReadTimeoutMS();
+                    getChannel().read(localByteBuf.asNIO(), readTimeoutMS, MILLISECONDS, null,
+                            new BasicCompletionHandler(localByteBuf, operationContext, c.asHandler()));
+                }
+            }).finish(localHandler.asCallback());
```
NathanQingyangXu, line 251: there are some potential implications I want to point out:

- the `localHandler` would be called twice in most of times instead of once
- if `prematurely reached end of stream` happens, originally the exception would be thrown synchronously; after the change it would be thrown in async manner or AFTER the above `completed()` method has returned

To me a safer alternative to avoid the above implications is simply to add a try block to maintain sync nature in this async callback (introducing another level of async might be complicated and not worthwhile, if semantically viable):

```
           try {
                ByteBuf localByteBuf = byteBufReference.getAndSet(null);
                if (result == -1) {
                    localByteBuf.release();
                    throw new MongoSocketReadException("Prematurely reached end of stream", serverAddress);
                } else if (!localByteBuf.hasRemaining()) {
                    localByteBuf.flip();
                    c.complete(localByteBuf);
                } else {
                    long readTimeoutMS = operationContext.getTimeoutContext().getReadTimeoutMS();
                    getChannel().read(localByteBuf.asNIO(), readTimeoutMS, MILLISECONDS, null,
                            new BasicCompletionHandler(localByteBuf, operationContext, c.asHandler()));
                }
            } catch (Throwable t) {
                localHandler.failed(t);
            }
``` 
I would add special logic processing for `InterruptedException` as below, but I saw we almost never pay attention to the subtlety it, then feel free to ignore it
```
            } catch (Throwable t) {
                if (t instanceof InterruptedException) {
                }
                localHandler.failed(t);
            }
```

NathanQingyangXu (commented):
Some uncertainties that neeed author's response

driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
vbabanin: > the localHandler would be called twice in most of times instead of once

Could you elaborate why you think it could be called twice in most of the times? 
I assume it relates to the comment above, i left a comment there: https://github.com/mongodb/mongo-java-driver/pull/1761#discussion_r2191042294.


> if prematurely reached end of stream happens, originally the exception would be thrown synchronously; after the change it would be thrown in async manner or AFTER the above completed() method has returned

`beginAsync` does not start execution on a different thread or concurrently (i understand that it might look so based on the naming, it is a bit confusing). It only indicates that the following code should be handled in an asynchronous style. Any exception thrown within the lambdas passed to `beginAsync` is propagated to a callback further up to `localHandler`. This framework was intentionally designed to mimic synchronous code to improve readability and maintain alignment with the sync counterpart.

For example, see [executeBatchAsync](https://github.com/mongodb/mongo-java-driver/blob/d3e81d85c5be5cbad72f9a8115b2ae46206d0f2a/driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java#L318) and compare it with its synchronous version. The async structure closely mirrors the sync flow.

Async framework/API also guarantees single-firing:

`c.asHandler()` completes only stage c.
`finish(localHandler.asCallback())` forwards that single completion (or failure) to `localHandler.`

There’s no execution path where `localHandler` would be invoked both directly and via finish(), so double invocations aren’t possible, unless misused. 

I also shared a recording offline that walks through the Async API for more context. Happy to do a quick live walkthrough as well.

driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
vbabanin: I see how this could be confusing at first glance, as we have custom async framework. It allows chaining asynchronous operations (and  mixing in synchronous code) stage by stage. 

In the example setup:
`thenRun` takes a lambda - an `AsyncRunnable` that receives a callback `(c)`, which needs to be completed explicitly via `c.complete(c)`. This signals that the current stage is done, and the chain can proceed.

After each `thenRun`, the framework continues to the next stage, eventually reaching `finish(localHandler.asCallback())`, which ties the entire chain back to the external handler.

Example code:
```
beginAsync
    .thenRun(c -> {
        // some synchronous work
        c.complete(); // marks this stage complete
    })
    .thenRun(c -> {
        asyncFunction(c); // some async operation that will call c.complete() later
    })
    .finish(localHandler.asCallback()); // final completion callback
```

In our case:
Inside the `thenRun` block, we use `c.asHandler()`, but that `c` is the local callback for the current stage, not `localHandler`. It delegates through the chain and ultimately completes via `finish(localHandler.asCallback())`.

So there is no scenario here where` localHandler` would be invoked twice. The framework ensures each stage triggers exactly once, propagating through to the final handler.

> To me the only reason to apply the finish() is to guard against the following statement throwing exception?
`beginAsync` will catch any exceptions thrown in the lambda and propagate to a user callback, and we might have also `IO  or executor exceptions` coming from write() method.

NathanQingyangXu: thanks for the detailed explanation. My poor head is still spinning? :).

driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
NathanQingyangXu: You had explained that the stage's callback is internal one, not the external one. Thanks for the explanation.

NathanQingyangXu (approved):

driver-core/src/main/com/mongodb/internal/connection/AsynchronousChannelStream.java
NathanQingyangXu: Yeah, I misunderstood in the sense that I had assumed a separate thread is running asynchorously for `begineAsync()`. Now I see it is really confusing.

PR #1762, opened 2025-07-07
Fix Netty reference leak.

rozza:
Fixes netty byte buffer releases in edge case scenarios:

- Ensure async select server uses a callback if the cluster had been closed
- Ensure that handleReadResponse checks to see if the cluster had been closed before retaining incoming buffers
- Ensure closing the netty stream releases all references

Test fixes
- Ensure tests run using paranoid leak detection
- Fail the testsuite if a leak is detected.
- Fixed releasing buffers in the ByteBufferBsonOutputTest.

JAVA-5901

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/netty/NettyStream.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/netty/NettyStream.java
@@ -307,8 +307,7 @@ final class NettyStream implements Stream {
                             composite.addComponent(next);
                             iter.remove();
                         } else {
-                            next.retain();
-                            composite.addComponent(next.readSlice(bytesNeededFromCurrentBuffer));
+                            composite.addComponent(next.readRetainedSlice(bytesNeededFromCurrentBuffer));
                         }
                         composite.writerIndex(composite.writerIndex() + bytesNeededFromCurrentBuffer);
                         bytesNeeded -= bytesNeededFromCurrentBuffer;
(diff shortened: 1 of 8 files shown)


driver-core/src/main/com/mongodb/internal/connection/netty/NettyStream.java
```diff
@@ -378,7 +381,7 @@ public void close() {
             for (Iterator<io.netty.buffer.ByteBuf> iterator = pendingInboundBuffers.iterator(); iterator.hasNext();) {
                 io.netty.buffer.ByteBuf nextByteBuf = iterator.next();
                 iterator.remove();
-                nextByteBuf.release();
+                nextByteBuf.release(nextByteBuf.refCnt());
```
vbabanin, line 386: As i see, we retain a buffer at most three times in `NettyStream`:

-  Initial ByteBuf from the allocator
-  When adding to pendingInboundBuffers
-  When adding retainedSlice to CompositeBuffer, that is returned from read/readAsync.

The first two are internal, but the third is owned by the caller. As long as we keep the invariant we won't hit `IllegalReferenceCountException`:
```
Every ByteBuf returned by NettyStream.read*() **must** be
release()‑d before NettyStream.close() is invoked.
```
As i see, all call‑sites satisfy the invariant today, and I don’t see it changing. I suggest adding an explicit comment to document this for the future:
```suggestion
                // Drops all retains to prevent silent leaks; assumes callers have already released   
                // ByteBuffers returned by that NettyStream before calling close.
                nextByteBuf.release(nextByteBuf.refCnt());
```

vbabanin (commented):
Looks good! Just one suggestion.

PR #1764, opened 2025-07-07
Properly handle exceptions in threads created by MongoClient

abalanonline:
Hello dear mongo-java-driver maintainers,
Please allow me to contribute this bugfix. The problem with Throwable in this line is that
it catches errors (OutOfMemoryError in particular) and doesn't let JVM crash leaving it
in zombie state - unable to recover, unable to exit.

JAVA-5913

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/AsynchronousClusterEventListener.java
@@ -31,6 +31,8 @@ import com.mongodb.event.ServerListener;
 import com.mongodb.event.ServerMonitorListener;
 import com.mongodb.event.ServerOpeningEvent;
 import com.mongodb.internal.VisibleForTesting;
+import com.mongodb.internal.diagnostics.logging.Logger;
+import com.mongodb.internal.diagnostics.logging.Loggers;
 
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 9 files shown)


stIncMale:
Hi @abalanonline, thank you for the PR.

1. The place your identified is not the only place where we were incorrectly catching and swallowing `Throwable`.
2. Also, there are a number of threads created by `MongoClient` that may be silently terminated due to a throwable. I updated those places such that we log a message and rethrow, allowing applications to handle such throwables via the [default `UncaughtExceptionHandler`](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/Thread.html#setDefaultUncaughtExceptionHandler(java.lang.Thread.UncaughtExceptionHandler)).

@abalanonline, @rozza please take a look.

abalanonline (commented):
Well, catching the error, logging and rethrowing it will work. I can give verbal approval.
But since you asked for my opinion, I would kindly suggest removing all Error catching from the code and letting the JVM handle them. Alternatively, the user of the client library can handle such errors themselves if they are confident enough.

stIncMale:
> suggest removing all Error catching from the code and letting the JVM handle them. Alternatively, the user of the client library can handle such errors themselves if they are confident enough.

The only behavioral difference between my approach and the approach you propose is logging. The issue with not logging is that when an internal `MongoClient` thread terminates due to an exception, the application that does not have the default `UncaughtExceptionHandler` configured, will have no easy way of knowing that the problem occurred. And the majority of application developers probably don't know about the internal `MongoClient` threads (why would they?), thus not necessarily suspecting that it may be useful to configure the default `UncaughtExceptionHandler`. This is why I deem log messages about unexpected termination of internal threads useful.

abalanonline:
I agree. Logging the error seems to be the best option.

PR #1765, opened 2025-07-11
Fix type parameters not being passed to fallback property codec

GliczDev:
For some reason, type parameters weren't passed to fallback property codec, which caused an exception when trying to access them. Now, this issue is fixed.

### DIFF
--- a/bson/src/main/org/bson/codecs/pojo/FallbackPropertyCodecProvider.java
+++ b/bson/src/main/org/bson/codecs/pojo/FallbackPropertyCodecProvider.java
@@ -20,6 +20,8 @@ import org.bson.codecs.Codec;
 import org.bson.codecs.configuration.CodecRegistry;
 
 import java.lang.reflect.Type;
+import java.util.List;
+import java.util.stream.Collectors;
 
 final class FallbackPropertyCodecProvider implements PropertyCodecProvider {
     private final CodecRegistry codecRegistry;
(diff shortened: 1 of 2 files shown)


bson/src/main/org/bson/codecs/pojo/FallbackPropertyCodecProvider.java
```diff
@@ -19,6 +19,8 @@
 import org.bson.codecs.Codec;
 import org.bson.codecs.configuration.CodecRegistry;
 
+import java.lang.reflect.Type;
```
rozza, line 22:
```suggestion
import java.lang.reflect.Type;
import java.util.List;
import java.util.stream.Collectors;
```

bson/src/main/org/bson/codecs/pojo/FallbackPropertyCodecProvider.java
```diff
@@ -35,6 +37,6 @@ public <S> Codec<S> get(final TypeWithTypeParameters<S> type, final PropertyCode
         if (clazz == pojoCodec.getEncoderClass()) {
             return (Codec<S>) pojoCodec;
         }
-        return codecRegistry.get(type.getType());
+        return codecRegistry.get(type.getType(), type.getTypeParameters().stream().<Type>map(TypeWithTypeParameters::getType).toList());
```
rozza:
```suggestion
        List<Type> types = type.getTypeParameters().stream().<Type>map(TypeWithTypeParameters::getType).collect(Collectors.toList());
        if (!types.isEmpty()) {
            return codecRegistry.get(type.getType(), types);
        }
        return codecRegistry.get(type.getType());
```

rozza (changes requested):
I added a couple of requested changes - to support java 8+ and to ensure that types are only provided if the list is not empty.

The `LookupCountingCodecRegistry` implementation in tests will need to implement `<T> Codec<T> get(Class<T> clazz, List<Type> typeArguments) ` inorder not to throw an exception.

eg:
```
@Override
        incrementCount(clazz);
            Codec<T> codec = provider.get(clazz, typeArguments, this);
            if (codec != null) {
                return codec;
            }
        }
        return null;
}
```

It would also be good to add a regression test where this code path is used.

PR #1768, opened 2025-07-17
$dateToString aggregation with timezone outputs invalid ISO8601 string

NathanQingyangXu:
https://jira.mongodb.org/browse/JAVA-5044

The server upstream [ticket](https://jira.mongodb.org/browse/SERVER-76696) aims to fix the following bug for the `$dateToString` aggregation expresison output:

- when `format` is not specified and `timezone` is not UTC, the string should not end with `Z`, which would be interpreted as UTC time, not w.r.t. the specified non-UTC zone

The server bugfix will end up with removing the trailing `Z` from the output since v7.1.

The change is a breaking one so it was not backported to v6. From driver side, no non-testing code logic needs to be changed unless there is some tech debt as in this PR (our `MqlDate#asString(TimeZone, Format)` doesn't support optional format, which is a mandatory precondition of the bug reproduction).

For this reason, this PR will do the following:

1. added a new overloaded `MqlDate#asString(TimeZone)` method to accommodate default format
2. changed internal common implementation to only include `format` field into MQL when it was provided expecitly or through the originual `MqlDate#asString(TimeZone, Format)` method
3. added new functional testing case to cover the new method; showcased that `Z` would be absent when non-UTC timezone is used(with server range above v7.1)


driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
```diff
@@ -225,6 +226,35 @@ public void dateAsStringTest() {
...
 
+    /**
+     * Tests that with default format and non-UTC timezone, {@code $dateToString} won't output invalid ISO 8601 string ending with 'Z'.
+     * Ticket: JAVA-5044
+     */
+    @Test
+    void dateAsStringWithDefaultFormat() {
+        // the bug was reported on v6.0.3, but was fixed in v7.1 without backporting
+        // due to the concern that it is a breaking change.
+        assumeTrue(serverVersionAtLeast(7, 1));
```
vbabanin: This test doesn’t rely on server communication, do we need this check?

NathanQingyangXu: It requires for internally `assertExpression()` will call server to valuate the expression and compare with the expected result or its first parameter.

driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
```diff
@@ -225,6 +226,35 @@ public void dateAsStringTest() {
                 of(instant).asString(of("America/New_York"), of("%Y-%m-%dT%H:%M:%S.%L%z")));
     }
 
+    /**
+     * Tests that with default format and non-UTC timezone, {@code $dateToString} won't output invalid ISO 8601 string ending with 'Z'.
```
vbabanin:
```suggestion
     * Tests that with server default format and non-UTC timezone, {@code $dateToString} won't output invalid ISO 8601 string ending with 'Z'.
```

driver-core/src/main/com/mongodb/client/model/mql/MqlExpression.java
```diff
@@ -947,10 +948,24 @@ public MqlInteger millisecond(final MqlString timezone) {
...
+        return doAsString(timezone, format);
+    }
+
+    @Override
+    public MqlString asString(final MqlString timezone) {
+        Assertions.notNull("timezone", timezone);
+        return doAsString(timezone, null);
+    }
+
+    private MqlString doAsString(final MqlString timezone, @Nullable final MqlString format) {
```
vbabanin: I think, the private method adds more code than it removes. We should extract code blocks to a helper only when the lines (and clarity) it saves outweighs the boilerplate it adds.

Currently, It slightly obscures the flow for readers who now have to jump to another method to see the expression. Inlining the logic in each overload makes the code shorter and more direct, so l suggest the in-lined approach below:

```
    @Override
    public MqlString asString(final MqlString timezone, final MqlString format) {
        Assertions.notNull("timezone", timezone);
        Assertions.notNull("format", format);
        return newMqlExpression((cr) -> astDoc("$dateToString", new BsonDocument()
                .append("date", this.toBsonValue(cr))
                .append("format", toBsonValue(cr, format))
                .append("timezone", toBsonValue(cr, timezone))));
    }

    @Override
    public MqlString asString(final MqlString timezone) {
        Assertions.notNull("timezone", timezone);
        return newMqlExpression((cr) -> astDoc("$dateToString", new BsonDocument()
                .append("date", this.toBsonValue(cr))
                .append("timezone", toBsonValue(cr, timezone))));
    }
```

NathanQingyangXu: that makes sense. There is a threshold above which private common method is needed. Currently the threshold is not reached.
Adopted. Thanks.

driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
```diff
@@ -225,6 +226,35 @@ public void dateAsStringTest() {
                 of(instant).asString(of("America/New_York"), of("%Y-%m-%dT%H:%M:%S.%L%z")));
     }
 
+    /**
+     * Tests that with server default format and non-UTC timezone, {@code $dateToString} won't output invalid ISO 8601 string ending with 'Z'.
+     * Ticket: JAVA-5044
+     */
+    @Test
+    void dateAsStringWithDefaultFormat() {
```
katcharov: This is testing a new API. But, since this is a bugfix, I would expect it to test an existing API, where the test would fail before the prod change, and pass afterward. Is that not the case?

driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
```diff
@@ -225,6 +226,35 @@ public void dateAsStringTest() {
...
+                utcDateTime.format(ISO_INSTANT),
+                date.asString(of("UTC")),
+                "{'$dateToString': {'date': {'$date': '2025-07-17T18:00:36.052Z'}, "
+                        + "'timezone': 'UTC'}}");
+
+        // non-UTC timezone's default format should be "%Y-%m-%dT%H:%M:%S.%L"
+        assertExpression(
+                utcDateTime.withZoneSameInstant(ZoneId.of("America/New_York")).format(ISO_LOCAL_DATE_TIME),
+                date.asString(of("America/New_York")),
+                "{'$dateToString': {'date': {'$date': '2025-07-17T18:00:36.052Z'}, "
```
katcharov: Shouldn't this test value not have a Z at the end? From the Jira:

> It tacks on a "Z" to the end of the string by default, which implies that it is in UTC

But here, we are expecting UTC?

driver-core/src/main/com/mongodb/client/model/mql/MqlExpression.java
```diff
@@ -953,6 +953,15 @@ public MqlString asString(final MqlString timezone, final MqlString format) {
                 .append("timezone", toBsonValue(cr, timezone))));
     }
 
+    @Override
+    public MqlString asString(final MqlString timezone) {
```
katcharov, line 957: I believe that we do not need this method, and that it might actually be introducing the bug described in the Jira. That is, the API as it is now intentionally prevents the user from specifying a timezone without also forcing that user to explicitly specify a format, thus preventing the situation from arising.

For reference:
* https://github.com/mongodb/mongo-java-driver/pull/1045#discussion_r1028655799
* https://jira.mongodb.org/browse/SERVER-74002 (not as closely related)

driver-core/src/main/com/mongodb/client/model/mql/MqlExpression.java
```diff
@@ -953,6 +953,15 @@ public MqlString asString(final MqlString timezone, final MqlString format) {
...
+        Assertions.notNull("timezone", timezone);
+        return newMqlExpression((cr) -> astDoc("$dateToString", new BsonDocument()
+                .append("date", this.toBsonValue(cr))
+                .append("timezone", toBsonValue(cr, timezone))));
+    }
+
+
```
katcharov: If my other comments are correct, I think we might instead make the following changes in this PR:

```suggestion
    public MqlString asString(final MqlString timezone) {
        fail(); // intentionally not implemented, see DRIVERS-2620
    }

    public MqlString asString() {
        fail(); // intentionally not implemented, see DRIVERS-2620
    }

```

katcharov (changes requested):

driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
vbabanin: The "Z" in the expected mql belongs to the input value, not the expected output. Expected output  comes from `utcDateTime.withZoneSameInstant(ZoneId.of("America/New_York")).format(ISO_LOCAL_DATE_TIME)` which is in the format of `2025‑07‑17T14:00:36.052` - this is what server returns when timezone is not UTC and format is omitted.

driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
NathanQingyangXu: yeah, correct. The testing would only succeed for v7.1..0+; it would fail for v6, for instance.

driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
NathanQingyangXu: Thanks for @vbabanin 's explanation. Yeah, the last Mql string parameter is the expected server input, not output.

driver-core/src/main/com/mongodb/client/model/mql/MqlExpression.java
NathanQingyangXu: thanks for sharing. Yeah, also saw you raising the point in our team slack channel. I have no objection of closing this PR.

driver-core/src/main/com/mongodb/client/model/mql/MqlExpression.java
NathanQingyangXu: I think we need to decide whether to close this PR in the first place. If we want to merge this PR, I would commit your suggestion, together with necessary testing changes.

driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
katcharov: I am not referring to the difference in test outcome based on the server version. The server would write such a test.

I mean, the ticket suggests that we need to make changes in the driver. When we write a regression test (or, all tests actually), the idea is that the test should not work before the change, but should work after. If the test works both before and after the change, the test has no relation to the change and is not testing it. If it fails both times, the change did not fix the issue.

Before the change to the production API, it is not actually possible to write a test that exhibits the bug via our API. So, no error. After the change, we can now write a test that exhibits the bug. But, we block the test from running on the server version necessary for the bug to show up. So things are "green" at first for that version, but after the prod code change, they are "red". This means we introduced a bug.

driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
katcharov: I see, thanks. It is the server version assume that blocks the test from running.

driver-core/src/main/com/mongodb/client/model/mql/MqlExpression.java
katcharov: What would the testing changes be?

NathanQingyangXu: if we changed as suggested, the newly-added `dateAsStringWithDefaultFormat()` would fail invariably. Then we might need to remove it totally and replace it with some simple testing cases to assert exception is thrown?

NathanQingyangXu: nvm. as we discussed, if we adopted the above change, no testing is required.

driver-core/src/test/functional/com/mongodb/client/model/mql/TypeMqlValuesFunctionalTest.java
NathanQingyangXu: If I didn't misunderstand, you said the same testing case would fail or be red on v6, correct? That is true and I understand that is an issue.

driver-core/src/main/com/mongodb/client/model/mql/MqlExpression.java
NathanQingyangXu: > If my other comments are correct, I think we might instead make the following changes in this PR:

I committed the suggestion and found the following issue:

- `MqlExpression` has `asString()` method already, which ends up with `$toString` expression, not the `$dateToString` expression. It was covered in the existing testing case `com.mongodb.client.model.mql.TypeMqlValuesFunctionalTest#dateAsStringTest()`

Not sure whether further testing is needed; for to reproduce the bug on v6, we need to do the following steps:

1. restore the original implementation (or not `fail()`)
2. use assumption of v6 instead of v7.1+

So given our current `throw fail();` implementation, there is no way to test the bug on v6 using `assertExpression()`. Am I misunderstanding something here?

driver-core/src/main/com/mongodb/client/model/mql/MqlDate.java
```diff
@@ -138,6 +138,20 @@ public interface MqlDate extends MqlValue {
...
+     *
+     * <ul>
+     *     <li>If {@code timezone} is UTC timezone, {@code "%Y-%m-%dT%H:%M:%S.%LZ"} will be the default
+     *     <li>Otherwise the default format will be {@code "%Y-%m-%dT%H:%M:%S.%L"}
+     * </ul>
+     *
+     * @param timezone the UTC Offset or Olson Timezone Identifier.
+     * @return the resulting value.
+     */
+    MqlString asString(MqlString timezone);
```
katcharov: Why add this to the external API? This would expose a method that would always fail.

katcharov (changes requested):

driver-core/src/main/com/mongodb/client/model/mql/MqlDate.java
NathanQingyangXu: removed. I misunderstood something.

driver-core/src/main/com/mongodb/client/model/mql/MqlExpression.java
```diff
@@ -953,6 +954,10 @@ public MqlString asString(final MqlString timezone, final MqlString format) {
                 .append("timezone", toBsonValue(cr, timezone))));
     }
 
+    public MqlString asString(final MqlString timezone) {
+        throw fail(); // intentionally not implemented, see DRIVERS-2620
```
vbabanin: Could we add additional context in this comment on why we decided to not implement it? This would help us to recollect the rationale in the future. 

For example: `Given that server versions < ... return a wrong format, not implementing this method helps prevent users from encountering that bug by avoiding the method that requires a format as a parameter.`

PR #1769, opened 2025-07-18
sunset Astrolabe

NathanQingyangXu:
https://jira.mongodb.org/browse/JAVA-5798


rozza:
> @rozza , we have stopped cloning local unified testing cases but grabbed by fetching from upstream at run time, correct?
Correct - I've just merged the latest dependabot commit - so we're up to date.

NathanQingyangXu:
> > I could delete the driver-workload-executor module, but then the following two checks would fail:
> 
> Why would they fail exactly?

I found the reason. We need to change snyk scanning config to remove the module. Now the snyk issues are gone.

PR #1770, opened 2025-07-18
Adding timeoutMS for explain helpers

nhachicha:
Fixes JAVA-5579

### DIFF
--- a/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/AggregatePublisher.java
+++ b/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/AggregatePublisher.java
@@ -274,7 +274,7 @@ public interface AggregatePublisher<TResult> extends Publisher<TResult> {
     <E> Publisher<E> explain(Class<E> explainResultClass, ExplainVerbosity verbosity);
 
     /**
-     * Explain the execution plan for this operation with the server's default verbosity level
+     * Explain the execution plan for this operation.
      *
      * @param timeoutMS the timeout in milliseconds for the explain operation
      * @return the execution plan
@@ -285,7 +285,7 @@ public interface AggregatePublisher<TResult> extends Publisher<TResult> {
     Publisher<Document> explain(long timeoutMS);
 
     /**
-     * Explain the execution plan for this operation with the given verbosity level
+     * Explain the execution plan for this operation.
      *
      * @param verbosity the verbosity of the explanation
      * @param timeoutMS the timeout in milliseconds for the explain operation
@@ -297,7 +297,7 @@ public interface AggregatePublisher<TResult> extends Publisher<TResult> {
     Publisher<Document> explain(ExplainVerbosity verbosity, long timeoutMS);
 
     /**
-     * Explain the execution plan for this operation with the server's default verbosity level
+     * Explain the execution plan for this operation.
      *
      * @param <E> the type of the document class
      * @param explainResultClass the document class to decode into
@@ -310,7 +310,7 @@ public interface AggregatePublisher<TResult> extends Publisher<TResult> {
     <E> Publisher<E> explain(Class<E> explainResultClass, long timeoutMS);
 
     /**
-     * Explain the execution plan for this operation with the given verbosity level
+     * Explain the execution plan for this operation.
      *
      * @param <E> the type of the document class
      * @param explainResultClass the document class to decode into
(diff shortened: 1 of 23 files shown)


nhachicha:
@rozza  Fixed https://spruce.mongodb.com/version/687fc6361ce1270007a8f87d/test-analysis?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/AggregatePublisher.java
```diff
@@ -272,4 +272,54 @@ public interface AggregatePublisher<TResult> extends Publisher<TResult> {
      * @mongodb.server.release 3.6
      */
     <E> Publisher<E> explain(Class<E> explainResultClass, ExplainVerbosity verbosity);
+
+    /**
+     * Explain the execution plan for this operation.
+     *
+     * @param timeoutMS the timeout in milliseconds for the explain operation
+     * @return the execution plan
+     * @since 5.2
```
jyemin, line 281: All these should be `5.6`.

driver-sync/src/main/com/mongodb/client/FindIterable.java
```diff
@@ -331,4 +331,50 @@ public interface FindIterable<TResult> extends MongoIterable<TResult> {
...
+
+    /**
+     * Explain the execution plan for this operation.
+     *
+     * @param timeoutMS the timeout in milliseconds for the explain operation
+     * @return the execution plan
+     * @mongodb.driver.manual reference/command/explain/
+     * @mongodb.server.release 3.2
+     */
+    Document explain(long timeoutMS);
```
jyemin, line 343: I'm not clear why all these overloads are needed.  For the rest of CSOT, applications specify the timeout via `MongoCluster/MongoDatabase/MongoCollection#withTimeout`. Would that suffice for specifying the timeout for `explain`?

jyemin (changes requested):

vbabanin:
The changes covered by this ticket were already implemented in [JAVA-5580](https://jira.mongodb.org/browse/JAVA-5580),  where we also decided not to introduce any new knobs on explain helpers

According to the existing CSOT specification, `timeoutMS is intended to be specified per operation`. The unified spec tests also require timeoutMS to be passed explicitly for each operation.

Our API supports this model via `withTimeout()` methods on `MongoCollection`, `MongoDatabase`, and `MongoClient`, which allow users to define a timeout at the desired scope, and subsequently per operation. For example, if a user wants to apply a timeout to a specific operation, they can call:

`collection.withTimeout(100, TimeUnit.SECONDS).find(...).explain();`

In that case,`timeoutMS` is applied to the actual operation that is executed, thus explain command, not a command being explained. The `maxTimeMS` on the command being explained is ignored by the server, even if it was present.

This approach aligns with our existing API design, which already includes `withWriteConcern()` and `withReadPreference()` on the aforementioned API components. 

We've documented this design approach, along with the reasoning and tradeoffs, here:
[Design doc](https://docs.google.com/document/d/1no9gHliqcjBz0qImiqjRvoNJD9X5kDAoR1rQq_lDjA8/edit?tab=t.0#heading=h.rilq4to2erfg).

To clarify: `timeoutMS` and `maxTimeMS` are separate APIs. `timeoutMS` is the successor, and it deprecates `maxTimeMS`. As noted in the CRUD spec:

> "NOTE: This option is deprecated in favor of timeoutMS."

### Regarding `maxTimeMS` (which is going to be marked deprecated once CSOT is GA).
[The prose test in the CRUD spec](https://github.com/baileympearson/specifications/blob/7355e5efd48eead548b9cfb0ca1d4e9c58543f98/source/crud/tests/README.md#14-explain-helpers-allow-users-to-specify-maxtimems) applies only to the deprecated maxTimeMS. In [discussion comments of DRIVERS-1256](https://jira.mongodb.org/browse/DRIVERS-1256?focusedCommentId=6677162#comment-6677162), we agreed not to require drivers to implement `maxTimeMS` support on explain for drivers that already support CSOT, since it would introduce a soon-to-be-deprecated parameter. See also this related [spec PR comment](https://github.com/mongodb/specifications/pull/1640#discussion_r1750837371).

While CSOT is currently marked alpha, `timeoutMS` already works with `explain()` via `MongoCollection` and explain(), allowing users to troubleshoot operations if needed.

Given that:
- CSOT is already in place (albeit alpha).
- timeoutMS works as intended with explain() via `withTimeout` on `MongoCollection`.
- and [JAVA-5580](https://jira.mongodb.org/browse/JAVA-5580) already covered the necessary changes.

I believe, no additional changes are needed as part of this ticket JAVA-5579.

PR #1775, opened 2025-07-23
Fix parameterized test names in CrudProseTest

jyemin:
Previously, for a parameterized test like
CrudProseTest::testBulkWriteHandlesWriteErrorsAcrossBatches, the test cases in the tests results XML output (and therefore the Evergreen UI) appear as 

```
  <testcase name="[1] false" classname="com.mongodb.reactivestreams.client.CrudProseTest""/>
  <testcase name="[2] true" classname="com.mongodb.reactivestreams.client.CrudProseTest"/>
```

Making it quite confusing to figure out which test it actually is.

After this change, they will appear as

```
  <testcase name="6. MongoClient.bulkWrite handles individual WriteErrors across batches--ordered:false" classname="com.mongodb.reactivestreams.client.CrudProseTest"/>
  <testcase name="6. MongoClient.bulkWrite handles individual WriteErrors across batches--ordered:true" classname="com.mongodb.reactivestreams.client.CrudProseTest"/>
```

and in the IDEA UI like

```
 6. MongoClient.bulkWrite handles individual WriteErrors across batches
     --ordered:false 
     --ordered:true
```

and in Gradle output like

```
  6. MongoClient.bulkWrite handles individual WriteErrors across batches

    Test 6. MongoClient.bulkWrite handles individual WriteErrors across batches--ordered:false PASSED
    Test 6. MongoClient.bulkWrite handles individual WriteErrors across batches--ordered:true PASSED
```

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/CrudProseTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/CrudProseTest.java
@@ -238,7 +238,7 @@ public class CrudProseTest {
     }
 
     @DisplayName("6. MongoClient.bulkWrite handles individual WriteErrors across batches")
-    @ParameterizedTest
+    @ParameterizedTest(name = "6. MongoClient.bulkWrite handles individual WriteErrors across batches--ordered:{0}")
     @ValueSource(booleans = {false, true})
     protected void testBulkWriteHandlesWriteErrorsAcrossBatches(final boolean ordered) {
         assumeTrue(serverVersionAtLeast(8, 0));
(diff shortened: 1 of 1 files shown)


driver-sync/src/test/functional/com/mongodb/client/CrudProseTest.java
vbabanin, line 1: It looks like the only remaining use of `@ParameterizedTest` relying on `@DisplayName` is in retryTests:
https://github.com/mongodb/mongo-java-driver/blob/01aff5a0789f71b9d0b56190d4996e8ac5436827/driver-core/src/test/unit/com/mongodb/internal/async/function/RetryStateTest.java#L248-L250

Should we address this as part of this PR as well?

jyemin: I'd prefer to keep this one focused on a single class.  We can always do another PR for RetryStateTest after this one is merged.

driver-sync/src/test/functional/com/mongodb/client/CrudProseTest.java
```diff
@@ -238,7 +238,7 @@ protected void testBulkWriteCollectsWriteConcernErrorsAcrossBatches() throws Int
     }
 
     @DisplayName("6. MongoClient.bulkWrite handles individual WriteErrors across batches")
```
rozza, line 240: Shame you have to duplicate `@DisplayName` for it to work in intellij :(

jyemin: Yeah, without that you get the method name instead, which is confusing.

vbabanin (approved):

PR #1776, opened 2025-07-24
Simplify Operations by uniting Sync and Async operations

rozza:
Merged Async Read / Write Operations into their sync operations counterparts.

* AsyncWriteOperation into WriteOperation is a simple merge
* AsyncReadOperation into ReadOperation requires an additional type parameter This is due to async and sync cursors have different types.

Added extra ReadOperations interfaces to represent:
  * Read operations that return the same type for sync / async
  * Read operations that represent cursors

Removed SyncOperations and AsyncOperations builders which both use the same Operations builder class.

JAVA-5900


driver-core/src/main/com/mongodb/internal/operation/Operations.java
jyemin, line 1: Now that this class is used directly instead of being wrapped by Sync/AsyncOperations, consider changing the return values of the operation-return methods to be the interface types rather than the concrete types, as was done in AsyncOperations and SyncOperations.

Longer term, it would be nice if all the concrete operation classes were package private, and `Operations` was the only way to create them, but that's quite a bit more work (especially since the legacy driver doesn't use `Operations`)

jyemin (changes requested):

nhachicha (approved):
Nice simplification 👍

PR #1777, opened 2025-07-28
Use connectTimeoutMS for connection establishment in maintenance TimeoutContext.

vbabanin:
JAVA-5927


rozza:
@vbabanin lots of tests failing after this - are they related or transient?

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideOperationsTimeoutProseTest.java
```diff
@@ -908,6 +909,91 @@ public void shouldThrowTimeoutExceptionForSubsequentCommitTransaction() {
...
+     * is established, if timeoutMS is set at the MongoClient level, it MUST be used as the timeout for all commands sent as part of
+     * the MongoDB or authentication handshakes
+     * <p>
+     * Therefore, if timeoutMS expires before connection establishment begins, connectTimeoutMS should not be used to start a TCP connection,
+     * since the connection will be closed immediately on the first handshake command due to the expired timeout. The timeout MUST be
+     * refreshed after each command.
+     */
+    @Test
+    @DisplayName("Should throw MongoOperationTimeoutException when establishing connection in background if timeoutMs expired before starting connect")
+    public void shouldThrowMongoOperationTimeoutWhenEstablishingConnectionInBackgroundIfTimeoutMsExpiredBeforeStartingConnect() {
```
rozza: This appears to be failing in reactive streams only and is happy in sync. Does the test need tweaking?

vbabanin: I replaced this prose-style test `Should throw MongoOperationTimeoutException when establishing connection in background if timeoutMs expired before starting connect` with a unit test for the sync driver. Helper for establishing SocketStream's already had this behaviour but no test. 

The async implementation doesn’t have this and it’s not strictly required, since it’s unlikely anyone would specify such a low timeout (e.g. 1 ms) that would effectively prevent connections from being established.

driver-core/src/test/functional/com/mongodb/internal/connection/SocketStreamHelperSpecification.groovy
```diff
@@ -78,6 +81,30 @@ class SocketStreamHelperSpecification extends Specification {
...
+                                        1,
+                                        100,
+                                        100,
+                                        1,
+                                        100))),
+                socket, getSocketAddresses(getPrimary(), new DefaultInetAddressResolver()).get(0),
+                SocketSettings.builder().build(), SslSettings.builder().build())
+
+        then:
+        thrown(MongoOperationTimeoutException.class)
```
rozza: Stop codenarc complaining
```suggestion
        thrown(MongoOperationTimeoutException)
```

rozza (approved):
LGTM!

PR #1782, opened 2025-08-11
Remove internal use of MongoNamespace.COMMAND_COLLECTION_NAME

jyemin:
This removes a remnant of the pre-OP_MSG wire protocol, which required a synthetic collection name for most commands.  It's no longer needed, except for the last remaining bit where the driver uses OP_QUERY for the first message in the connection handshake.

JAVA-5939

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
@@ -408,9 +408,7 @@ public final class CommandMessage extends RequestMessage {
     }
 
     /**
-     * Gets the collection name, which may be null for some message types
-     *
-     * Gets the database name, which may be null for some message types
+     * Gets the database name
      *
      * @return the database name
      */
--- a/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
@@ -83,6 +83,7 @@ public final class CommandMessage extends RequestMessage {
     private final boolean exhaustAllowed;
     private final MessageSequences sequences;
     private final boolean responseExpected;
+    private final String database;
     /**
      * {@code null} iff either {@link #sequences} is not of the {@link DualMessageSequences} type,
      * or it is of that type, but it has not been {@linkplain #encodeMessageBodyWithMetadata(ByteBufferBsonOutput, OperationContext) encoded}.
@@ -119,7 +120,8 @@ public final class CommandMessage extends RequestMessage {
                    final boolean responseExpected, final boolean exhaustAllowed,
                    final MessageSequences sequences,
                    final ClusterConnectionMode clusterConnectionMode, @Nullable final ServerApi serverApi) {
-        super(database, getOpCode(settings, clusterConnectionMode, serverApi), settings);
+        super(getOpCode(settings, clusterConnectionMode, serverApi), settings);
+        this.database = database;
         this.command = command;
         this.commandFieldNameValidator = commandFieldNameValidator;
         this.readPreference = readPreference;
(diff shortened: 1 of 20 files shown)


driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -281,7 +275,7 @@ private int writeOpMsg(final ByteBufferBsonOutput bsonOutput, final OperationCon
 
     private int writeOpQuery(final ByteBufferBsonOutput bsonOutput) {
         bsonOutput.writeInt32(0);
-        bsonOutput.writeCString(namespace.getFullName());
+        bsonOutput.writeCString(new MongoNamespace(getDatabase(), "$cmd").getFullName());
```
jyemin, line 280: This is now the only place we need to use the synthetic collection name, so just inlined it.

Note that with DRIVERS-2944, this too will soon be removed.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/MongoOperationPublisher.java
```diff
@@ -175,7 +175,7 @@ MongoOperationPublisher<T> withDatabase(final String name) {
     }
 
     <D> MongoOperationPublisher<D> withDatabaseAndDocumentClass(final String name, final Class<D> documentClass) {
-        return withNamespaceAndDocumentClass(new MongoNamespace(notNull("name", name), "ignored"),
+        return withNamespaceAndDocumentClass(new MongoNamespace(notNull("name", name), "_ignored"),
```
jyemin, line 178: Just changed this to be consistent with other similar spots in the driver.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -411,6 +407,15 @@ private static boolean isServerVersionKnown(final MessageSettings settings) {
         return settings.getMaxWireVersion() != UNKNOWN_WIRE_VERSION;
     }
 
+    /**
+     * Gets the collection name, which may be null for some message types
+     *
+     * @return the collection name
+     */
+    public String getDatabase() {
```
NathanQingyangXu, line 415: the Javadoc seems to be copied from `getCollectionName()`, not applicable to `getDatabase()`.

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
NathanQingyangXu: if this is the only location the `$cmd` is used still, it seems we could either use the `COMMAND_COLLECTION_NAME` constant here or delete the constant which is not used; either way is better than the coexistence of both (so when we finish 2944, no dangling stuff is overlooked). Not a big deal for sure.

NathanQingyangXu (changes requested):

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: Done

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: I didn't want to keep using the constant because it should be deprecated (I'll add a ticket for that), and I can't just delete the constant because it's part of the API. But since it's not deprecated yet, I'll keep using it here for now.

NathanQingyangXu (approved):

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
jyemin: On second thought, let's just leave this as is.

PR #1785, opened 2025-08-12
Handle dropIndex behavior change

rozza:
JAVA-5941

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
@@ -80,6 +80,98 @@ public final class UnifiedTestModifications {
                 .test("client-side-operations-timeout", "WaitQueueTimeoutError does not clear the pool",
                         "WaitQueueTimeoutError does not clear the pool");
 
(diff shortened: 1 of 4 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
```diff
@@ -80,6 +80,98 @@ public static void applyCustomizations(final TestDef def) {
                 .test("client-side-operations-timeout", "WaitQueueTimeoutError does not clear the pool",
                         "WaitQueueTimeoutError does not clear the pool");
 
+        def.skipJira("https://jira.mongodb.org/browse/JAVA-5491")
+                .testContains("client-side-operations-timeout", "dropIndex")
+                .when(() -> !serverVersionLessThan(8, 3))
+                .test("client-side-operations-timeout", "operations ignore deprecated timeout options if timeoutMS is set",
```
nhachicha, line 86: Is there an upcoming PR that updates these tests? It seems the only change is that the server returns `ok:1` since 8.3. It's unfortunate to disable all these checks?

rozza: Yeah theres a drivers ticket related to this to fix / change the tests themselves.

PR #1790, opened 2025-08-15
Add support for awaiting CommandStartedEvent in Unified Test Runner.

vbabanin:
JAVA-5815

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/unified/EventMatcher.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/unified/EventMatcher.java
@@ -224,7 +224,7 @@ final class EventMatcher {
         }
     }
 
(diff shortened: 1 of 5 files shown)


driver-sync/src/test/functional/com/mongodb/client/unified/EventMatcher.java
```diff
@@ -223,6 +224,26 @@ public void waitForConnectionPoolEvents(final String client, final BsonDocument
         }
     }
 
+    public void waitForCommandEvents(final String clientId, final BsonDocument expectedCommandEvent, int count,
```
rozza:
```suggestion
    public void waitForCommandEvents(final String clientId, final BsonDocument expectedCommandEvent, final int count,
```

rozza (approved):
One checkstyle nit then LGTM

PR #1795, opened 2025-08-26
Update the DEFAULT_CODEC_REGISTRY order

rozza:
Ensure that KotlinCodecProvider is used to handle any kotlin types first, then fall back to any native java codecs.

For non kotlin users this will be a NOOP as the actual kotlin codec providers won't be on the class path

JAVA-5928


vbabanin (approved):
LGTM!

PR #1802, opened 2025-09-11
Ensure encrypted fields data exists fully

rozza:
Appears theres been a change when using mongocryptd and an empty document is returned.

JAVA-5959


.evergreen/run-csfle-tests-with-mongocryptd.sh
```diff
@@ -60,8 +60,6 @@ echo "Running tests with Java ${JAVA_VERSION}"
       ${GRADLE_EXTRA_VARS} \
       -Dorg.mongodb.test.fle.on.demand.credential.test.failure.enabled=true \
       --stacktrace --info --continue \
-      driver-legacy:test \
```
rozza, line 63: Legacy tests don't use the unified test runner (this was missed when migrating from the legacy runner)

PR #1806, opened 2025-09-22
Reverse equals check to avoid NPEs

owen-obrien:
https://jira.mongodb.org/browse/JAVA-5971


rozza:
Thanks @owen-obrien excellent PR and tests. This will be released as part of the 5.6.1 release due Oct 2nd.

PR #1807, opened 2025-09-22
ClientMetaData refactoring

rozza:
Avoid appending duplicate metadata

JAVA-5955

### DIFF
--- a/driver-core/src/main/com/mongodb/MongoDriverInformation.java
+++ b/driver-core/src/main/com/mongodb/MongoDriverInformation.java
@@ -16,13 +16,15 @@
 
 package com.mongodb;
 
+import com.mongodb.annotations.Internal;
 import com.mongodb.annotations.NotThreadSafe;
+import com.mongodb.internal.client.DriverInformation;
+import com.mongodb.internal.client.DriverInformationHelper;
 
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-import static com.mongodb.assertions.Assertions.isTrue;
 import static com.mongodb.assertions.Assertions.notNull;
 
 /**
@@ -46,9 +48,7 @@ import static com.mongodb.assertions.Assertions.notNull;
  * @mongodb.server.release 3.4
  */
 public final class MongoDriverInformation {
-    private final List<String> driverNames;
-    private final List<String> driverVersions;
-    private final List<String> driverPlatforms;
+    private final List<DriverInformation> driverInformationList;
 
     /**
      * Convenience method to create a Builder.
@@ -75,7 +75,7 @@ public final class MongoDriverInformation {
      * @return the driverNames
      */
     public List<String> getDriverNames() {
-        return driverNames;
+        return DriverInformationHelper.getNames(driverInformationList);
     }
 
     /**
@@ -84,7 +84,7 @@ public final class MongoDriverInformation {
      * @return the driverVersions
      */
     public List<String> getDriverVersions() {
-        return driverVersions;
+        return DriverInformationHelper.getVersions(driverInformationList);
     }
 
     /**
@@ -93,7 +93,15 @@ public final class MongoDriverInformation {
      * @return the driverPlatforms
      */
     public List<String> getDriverPlatforms() {
-        return driverPlatforms;
+        return DriverInformationHelper.getPlatforms(driverInformationList);
+    }
+
+    /**
+     * For internal use only
+     */
+    @Internal
+    public List<DriverInformation> getDriverInformationList() {
+        return driverInformationList;
     }
 
     /**
@@ -101,7 +109,7 @@ public final class MongoDriverInformation {
      */
     @NotThreadSafe
     public static final class Builder {
-        private final MongoDriverInformation driverInformation;
+        private final MongoDriverInformation mongoDriverInformation;
         private String driverName;
         private String driverVersion;
         private String driverPlatform;
@@ -147,38 +155,26 @@ public final class MongoDriverInformation {
          * @return the driver information
          */
         public MongoDriverInformation build() {
-            isTrue("You must also set the driver name when setting the driver version", !(driverName == null && driverVersion != null));
-
-            List<String> names = prependToList(driverInformation.getDriverNames(), driverName);
-            List<String> versions = prependToList(driverInformation.getDriverVersions(), driverVersion);
-            List<String> platforms = prependToList(driverInformation.getDriverPlatforms(), driverPlatform);
-            return new MongoDriverInformation(names, versions, platforms);
-        }
-
-        private List<String> prependToList(final List<String> stringList, final String value) {
-            if (value == null) {
-                return stringList;
-            } else {
-                ArrayList<String> newList = new ArrayList<>();
-                newList.add(value);
-                newList.addAll(stringList);
-                return Collections.unmodifiableList(newList);
+            DriverInformation driverInformation = new DriverInformation(driverName, driverVersion, driverPlatform);
+            if (mongoDriverInformation.driverInformationList.contains(driverInformation)) {
+                return mongoDriverInformation;
             }
+
+            List<DriverInformation> driverInformationList = new ArrayList<>(mongoDriverInformation.driverInformationList);
+            driverInformationList.add(driverInformation);
+            return new MongoDriverInformation(Collections.unmodifiableList(driverInformationList));
         }
 
         private Builder() {
-            List<String> immutableEmptyList = Collections.emptyList();
-            driverInformation = new MongoDriverInformation(immutableEmptyList, immutableEmptyList, immutableEmptyList);
+            mongoDriverInformation = new MongoDriverInformation(Collections.emptyList());
         }
 
         private Builder(final MongoDriverInformation driverInformation) {
-            this.driverInformation = notNull("driverInformation", driverInformation);
+            this.mongoDriverInformation = notNull("driverInformation", driverInformation);
         }
     }
 
-    private MongoDriverInformation(final List<String> driverNames, final List<String> driverVersions, final List<String> driverPlatforms) {
-        this.driverNames = driverNames;
-        this.driverVersions = driverVersions;
-        this.driverPlatforms = driverPlatforms;
+    private MongoDriverInformation(final List<DriverInformation> driverInformation) {
+        this.driverInformationList = notNull("driverInformation", driverInformation);
     }
 }
--- a/driver-core/src/main/com/mongodb/annotations/Internal.java
+++ b/driver-core/src/main/com/mongodb/annotations/Internal.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ * Copyright 2010 The Guava Authors
+ * Copyright 2011 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mongodb.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Signifies that a public API element is intended for internal use only.
+ *
+ * <p>It is inadvisable for <i>applications</i> to use Internal APIs as they are intended for <b>internal library purposes</b> only.</p>
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({
+        ElementType.ANNOTATION_TYPE,
+        ElementType.CONSTRUCTOR,
+        ElementType.FIELD,
+        ElementType.METHOD,
+        ElementType.PACKAGE,
+        ElementType.TYPE })
+@Documented
+@Alpha(Reason.CLIENT)
+public @interface Internal {
+}
--- a/driver-core/src/main/com/mongodb/annotations/Internal.java
+++ b/driver-core/src/main/com/mongodb/annotations/Internal.java
@@ -1,7 +1,5 @@
 /*
  * Copyright 2008-present MongoDB, Inc.
- * Copyright 2010 The Guava Authors
- * Copyright 2011 The Guava Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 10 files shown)


driver-core/src/main/com/mongodb/annotations/Internal.java
```diff
@@ -0,0 +1,42 @@
...
+@Target({
+        ElementType.ANNOTATION_TYPE,
+        ElementType.CONSTRUCTOR,
+        ElementType.FIELD,
+        ElementType.METHOD,
+        ElementType.PACKAGE,
+        ElementType.TYPE })
+@Documented
+@Alpha(Reason.CLIENT)
+public @interface Internal {
```
rozza, line 39: Had to add an internal method `MongoDriverInformation#getDriverInformationList` to get the list of `DriverInformation` and added this annotation as a way to highlight things for internal use only.

If this ok, then I'll add a ticket to annotation each of the `internal` packages in their `package-info.java`

driver-core/src/main/com/mongodb/internal/client/DriverInformation.java
```diff
@@ -0,0 +1,81 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mongodb.internal.client;
+
+import com.mongodb.lang.Nullable;
+
+import java.util.Objects;
+
+public final class DriverInformation {
```
rozza, line 22: The changes required  by the spec ensure that driver information as a whole can be checked for equality - so that duplicates aren't appended.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -81,57 +81,61 @@ public BsonDocument getBsonDocument() {
...
-                    mongoDriverInformationToAppend.getDriverPlatforms());
-            this.clientMetadataBsonDocument = createClientMetadataDocument(applicationName, driverInformation);
+            boolean hasAddedNewData = false;
+            for (DriverInformation driverInformation : mongoDriverInformationToAppend.getDriverInformationList()) {
+               if (!driverInformationList.contains(driverInformation)) {
+                   hasAddedNewData = true;
+                   driverInformationList.add(driverInformation);
+               }
+            }
+            if (hasAddedNewData) {
```
rozza, line 91: Only recalculate the `clientMetadataBsonDocument` if there has been a real change.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -81,57 +81,61 @@ public BsonDocument getBsonDocument() {
...
-        BsonDocument client = new BsonDocument();
-        tryWithLimit(client, d -> putAtPath(d, "application.name", applicationName));
+        BsonDocument clientMetadata = new BsonDocument();
+        tryWithLimit(clientMetadata, d -> putAtPath(d, "application.name", applicationName));
 
         // required fields:
-        tryWithLimit(client, d -> {
-            putAtPath(d, "driver.name", driverInformation.getInitialDriverName());
-            putAtPath(d, "driver.version", driverInformation.getInitialDriverVersion());
+        tryWithLimit(clientMetadata, d -> {
```
rozza, line 109: We add the initial driver information - if there is room. Then later we add the full driver information (eg the extended libraries that use the driver) if there is room.

This behavior is defined by the spec.

driver-core/src/main/com/mongodb/MongoDriverInformation.java
```diff
@@ -46,9 +48,7 @@
  * @mongodb.server.release 3.4
  */
 public final class MongoDriverInformation {
-    private final List<String> driverNames;
```
rozza, line 49: Using lists of Strings, no longer works for the new test cases - as it meant we couldn't check for duplicates effectively.

driver-core/src/test/unit/com/mongodb/MongoDriverInformationSpecification.groovy
```diff
@@ -30,7 +30,7 @@ class MongoDriverInformationSpecification extends Specification {
         options.getDriverPlatforms() ==  []
     }
 
-    def 'should not prepend data if none has been added'() {
+    def 'should not append data if none has been added'() {
```
rozza, line 33: We never prepended the data - so these tests needed some tweaking to make sense.

driver-core/src/test/unit/com/mongodb/MongoDriverInformationSpecification.groovy
```diff
@@ -40,50 +40,42 @@ class MongoDriverInformationSpecification extends Specification {
...
 
+        def options = MongoDriverInformation.builder(javaDriverInfo).driverName('mongo-scala-driver').build()
+
         expect:
         options.getDriverNames() == ['mongo-java-driver', 'mongo-scala-driver']
         options.getDriverVersions() == ['3.4.0']
         options.getDriverPlatforms() == ['Java oracle64-1.8.0.31']
     }
 
-    def 'should error if trying to set a version without setting a name'() {
```
rozza, line 79: There is a specific new test from the spec that actually tests we can do this. So removed the check and this test.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientMetadataProseTest.java
```diff
@@ -71,53 +73,38 @@ public void tearDown() {
...
-        return Stream.of(
-                Arguments.of("1.0", "Framework", "Framework Platform"),
-                Arguments.of("1.0", "Framework", null),
-                Arguments.of(null, "Framework", "Framework Platform"),
-                Arguments.of(null, "Framework", null)
-        );
-    }
-
-    @ParameterizedTest
+    @DisplayName("Test 1: Test that the driver updates metadata")
```
rozza, line 76: The new prose tests. I renamed the existing ones to make it easier for future changes / adding new tests.

driver-core/src/main/com/mongodb/internal/client/DriverInformationHelper.java
```diff
@@ -0,0 +1,84 @@
...
+                    if (platform != null) {
+                        data.add(platform);
+                    }
+                    break;
+                default:
+                    fail(format("Unknown field: %s", driverField));
+            }
+        });
+        return Collections.unmodifiableList(data);
+    }
```
vbabanin: We could simplify this by using a lambda-based extractor instead of the enum/switch.


Passing a `Function<DriverInformation, String>` removes the need for the enum and reduces branching preserving the same behaviour. 

For example:
```suggestion
    public static List<String> getNames(final List<DriverInformation> driverInformation) {
        return getDriverField(DriverInformation::getDriverName, driverInformation);
    }

    public static List<String> getVersions(final List<DriverInformation> driverInformation) {
        return getDriverInformation(DriverInformation::getDriverVersion, driverInformation);
    }

    public static List<String> getPlatforms(final List<DriverInformation> driverInformation) {
        return getDriverField(DriverInformation::getDriverPlatform, driverInformation);
    }

    private static List<String> getDriverField(final Function<DriverInformation, String> fieldSupplier,
                                                                          final List<DriverInformation> driverInformation) {
        return Collections.unmodifiableList(driverInformation.stream()
                .map(fieldSupplier)
                .filter(Objects::nonNull)
                .collect(Collectors.toList()));
    }
```

driver-core/src/main/com/mongodb/internal/client/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This package contains classes for internal client functionality.
+ */
+
+@NonNullApi
```
nhachicha, line 21: Will this impact existing methods nullability in the package? Not sure if our linter will catch this?

driver-core/src/main/com/mongodb/MongoDriverInformation.java
```diff
@@ -93,15 +93,23 @@ public List<String> getDriverVersions() {
...
      */
     public List<String> getDriverPlatforms() {
-        return driverPlatforms;
+        return DriverInformationHelper.getPlatforms(driverInformationList);
+    }
+
+    /**
+     * For internal use only
+     */
+    @Internal
```
nhachicha, line 102: Why can't this be `public` (similar to existing `getDriverPlatforms`, `getDriverVersions` etc? 
we can return a `Collections.unmodifiable` (read-only) to the user?

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
```diff
@@ -81,57 +81,61 @@ public BsonDocument getBsonDocument() {
...
-        tryWithLimit(client, d -> putAtPath(d, "env.timeout_sec", faasEnvironment.getTimeoutSec()));
-        tryWithLimit(client, d -> putAtPath(d, "env.memory_mb", faasEnvironment.getMemoryMb()));
-        tryWithLimit(client, d -> putAtPath(d, "env.region", faasEnvironment.getRegion()));
+        tryWithLimit(clientMetadata, d -> putAtPath(d, "platform", INITIAL_DRIVER_INFORMATION.getDriverPlatform()));
+        tryWithLimit(clientMetadata, d -> putAtPath(d, "platform", listToString(DriverInformationHelper.getPlatforms(driverInformationList))));
+        tryWithLimit(clientMetadata, d -> putAtPath(d, "os.name", getOperatingSystemName()));
+        tryWithLimit(clientMetadata, d -> putAtPath(d, "os.architecture", getProperty("os.arch", "unknown")));
+        tryWithLimit(clientMetadata, d -> putAtPath(d, "os.version", getProperty("os.version", "unknown")));
 
-        tryWithLimit(client, d -> putAtPath(d, "env.container.runtime", containerRuntime.getName()));
```
nhachicha, line 131: Is this naming change an issue for analytic tools (collecting the previous metrics)?

driver-sync/src/test/functional/com/mongodb/client/AbstractClientMetadataProseTest.java
```diff
@@ -169,31 +157,219 @@ void shouldAppendToDefaultClientMetadataWhenUpdatedAfterInitialization(@Nullable
...
+            // then
+            sleep(5); // wait for connection to become idle
+            mongoClient.appendMetadata(getMongoDriverInformation(new DriverInformation("framework", "1.2", "Library Platform")));
+
+            clientMetaData = executePingAndCaptureMetadataHandshake(mongoClient)
+                    .orElseThrow(AbstractClientMetadataProseTest::failOnEmptyMetadata);
+
+            // then
+            sleep(5); // wait for connection to become idle
+            mongoClient.appendMetadata(getMongoDriverInformation(new DriverInformation("library", "1.2", "Library Platform")));
```
nhachicha, line 277: The spec considers `"framework", "1.2", "Library Platform"` equal to `"library", "1.2", "Library Platform"` ?

driver-core/src/main/com/mongodb/annotations/Internal.java
nhachicha: We already have `VisibleForTesting` should we try to unify all these in a similar fashion to [ApiStatus](https://javadoc.io/doc/org.jetbrains/annotations-java5/20.1.0/org/jetbrains/annotations/ApiStatus.html)
_(already supported in IntelliJ ~)_

Image

rozza: I didnt want to misuse that as it's not for testing. It's needed for the `ClientMetaData` API as well. This all stems from `MongoDriverInformation` being lossy, I need the combined information of name, version and platform.

driver-core/src/main/com/mongodb/MongoDriverInformation.java
rozza: So it could be, but then we'd have to make `DriverInformation` public as well and that really is an internal implementation detail.  It also means we'd have to wait for 5.7.0 if we follow semver for this.

driver-core/src/main/com/mongodb/internal/connection/ClientMetadata.java
rozza: The diff makes this confusing. Most of the real changes are the name of the variable from `client` to `clientMetaData` we keep all the same metrics.

driver-core/src/main/com/mongodb/internal/client/package-info.java
rozza: As this is a totally new package for the `DriverInformation` and its helper. This is just being explicit about the nullability expectations.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientMetadataProseTest.java
rozza: No - the prose test here is checking that adding the same data as the initial data doesn't reappend it. So the test here is to ensure the data and is same as `clientMetaData` set after adding the framework definition.

nhachicha (approved):
Since it's outside the scope of this PR, we can create a ticket to consider folding the `@Internal` annotation under`APIStatus` 👍

driver-core/src/main/com/mongodb/annotations/Internal.java
vbabanin: We have other places where methods are exposed in the public API but intended for internal use. For example:
https://github.com/mongodb/mongo-java-driver/blob/5e4e96ae3301c49b08329390e4358f8a08150adf/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/ClientSession.java#L37-L55

Since we’ve introduced a new annotation here, we should make it consistent with those cases by adding the annotation there as well. What do you think?

driver-core/src/main/com/mongodb/annotations/Internal.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ * Copyright 2010 The Guava Authors
+ * Copyright 2011 The Guava Authors
+ *
```
vbabanin, line 3: Should we also mention this file in THIRD-PARTY-NOTICES, as we did for other vendored components?
https://github.com/mongodb/mongo-java-driver/blob/48fdf2a4e392f086221c6eb6834981f304354a52/THIRD-PARTY-NOTICES#L40-L43

driver-core/src/main/com/mongodb/annotations/Internal.java
vbabanin: We have other places where methods are exposed in the public API but intended for internal use. For example:
https://github.com/mongodb/mongo-java-driver/blob/5e4e96ae3301c49b08329390e4358f8a08150adf/driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/ClientSession.java#L37-L55

Since we’ve introduced a new annotation here, should we make it consistent with those cases by adding the annotation there as well? We could do this in a separate ticket.

driver-core/src/main/com/mongodb/annotations/Internal.java
rozza: Good catch thats copy pasta - will remove as this is a new unrelated annotation.

driver-core/src/main/com/mongodb/annotations/Internal.java
rozza: Added JAVA-5975

driver-core/src/main/com/mongodb/annotations/Internal.java
rozza: Done

PR #1810, opened 2025-09-24
Fix integer overflow with max limit & batchsize

rozza:
Eliminate unnecessary killCursors command when batchSize == limit can overflow if using Integer.MAX_VALUE for both.

JAVA-5667
JAVA-5970

### DIFF
--- a/driver-core/src/test/unit/com/mongodb/internal/operation/FindOperationUnitSpecification.groovy
+++ b/driver-core/src/test/unit/com/mongodb/internal/operation/FindOperationUnitSpecification.groovy
@@ -102,6 +102,27 @@ class FindOperationUnitSpecification extends OperationUnitSpecification {
         version << [[3, 2, 0]] * 10 + [[3, 4, 0]] * 10
     }
 
+    def 'should find with correct command with effective batch size'() {
+        when:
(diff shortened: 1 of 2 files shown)


driver-core/src/test/unit/com/mongodb/internal/operation/FindOperationUnitSpecification.groovy
```diff
@@ -102,6 +102,27 @@ class FindOperationUnitSpecification extends OperationUnitSpecification {
...
+        def operation = new FindOperation<BsonDocument>(namespace, new BsonDocumentCodec())
+                .batchSize(batchSize)
+                .limit(limit)
+
+        def expectedCommand = new BsonDocument('find', new BsonString(namespace.getCollectionName()))
+                .append('batchSize', new BsonInt32(commandBatchSize))
+                .append('limit', new BsonInt32(commandLimit))
+
+        then:
+        testOperation(operation, [7, 0, 0], expectedCommand, async, commandResult)
```
nhachicha, line 116: Why is this limited to 7.0? Shouldn't this run also from `[3, 2, 0]`

rozza: This just tests the operation is crafted as expected and supplies a "fake" server version.

PR #1812, opened 2025-09-30
Fix AbstractWriteSearchIndexOperation executeAsync releasing connections

rozza:
Update SyncMongoClient to assert connections are released

JAVA-5972

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncOperationHelper.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncOperationHelper.java
@@ -138,6 +138,9 @@ final class AsyncOperationHelper {
                     AsyncCallbackSupplier<R> curriedFunction = c -> function.apply(resource, c);
                     curriedFunction.whenComplete(resource::release).get(errorHandlingCallback);
                 } catch (Exception e) {
+                    if (resource.getCount() > 0) {
+                        resource.release();
+                    }
                     errorHandlingCallback.onResult(null, e);
                 }
             }


driver-core/src/main/com/mongodb/internal/operation/AbstractWriteSearchIndexOperation.java
```diff
@@ -66,9 +66,9 @@ public void executeAsync(final AsyncWriteBinding binding, final SingleResultCall
                                 writeConcernErrorTransformerAsync(binding.getOperationContext().getTimeoutContext()), (result, commandExecutionError) -> {
                                     try {
                                         swallowOrThrow(commandExecutionError);
-                                        callback.onResult(result, null);
+                                        cb.onResult(result, null);
```
rozza, line 69: This the reason for the unreleased connection - it used the original and not decorated callback that handles the releasing.

The rest of the PR is updating the test suite to ensure there are no future regressions. The core is the SyncMongoClient close method that now has a listener that counts the connections out and in. It also asserts that all connections are released.

driver-core/src/main/com/mongodb/internal/operation/AsyncOperationHelper.java
```diff
@@ -138,6 +138,9 @@ static <R, T extends ReferenceCounted> void withAsyncSuppliedResource(final Asyn
                     AsyncCallbackSupplier<R> curriedFunction = c -> function.apply(resource, c);
                     curriedFunction.whenComplete(resource::release).get(errorHandlingCallback);
                 } catch (Exception e) {
+                    if (resource.getCount() > 0) {
```
rozza, line 141: This more closely mirrors the SyncOperationHelper and will ensure any errors thrown outside of the callback also release the resource.

nhachicha (approved):
Nice refactor 👍

PR #1830, opened 2025-10-20
test: Fix TODO by adding Pattern type to BSON codec test

HJC96:
## Problem
The test had a commented-out Pattern test with TODO note:
```java
// TODO: Pattern doesn't override equals method!
```
This prevented Pattern from being tested alongside other primitive BSON types.

## Changes
- Removed TODO comment
- Added Pattern encode/decode test
- Pattern objects are decoded as `BsonRegularExpression`, so comparison is done separately


nhachicha:
Hi @HJC96 can you rebase your changes on top of `main` PR claims to be a tiny test fix, but touches 4,010 files with `+473,454 / -93,164` lines

HJC96:
@nhachicha rebased onto current main, PR diff is now scoped to the intended test change.

PR #1856, opened 2025-12-17
[JAVA-6033] ServerHeartbeatSucceededEvent is not fired for initial POLL monitoring

nhachicha:
~~https://github.com/mongodb/mongo-java-driver/pull/1900 depends on this PR.~~
Update: #1900 no longer depends on this one. Spec PR https://github.com/mongodb/specifications/pull/1880 rewrote the `backpressure-network-*-fail.yml` tests to wait on `serverDescriptionChangedEvent` instead of `serverHeartbeatSucceededEvent`.

JAVA-6033


driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -216,6 +216,18 @@ public void run() {
...
+                            && currentServerDescription.getType() != UNKNOWN
+                            && !shouldStreamResponses
+                            && connection != null && !connection.isClosed();
+
+                    if (justConnected) {
+                        // Do immediate heartbeat for newly connected server in POLL mode
+                        // to fire the ServerHeartbeatSucceededEvent
+                        continue;
+                    }
+
```
rozza:
```suggestion
            boolean isPollMode = serverSettings.getServerMonitoringMode().equals(ServerMonitoringMode.POLL);
            ServerDescription unknownConnectingServerDescription = unknownConnectingServerDescription(serverId, null);
            ServerDescription currentServerDescription = unknownConnectingServerDescription;
            try {
                while (!isClosed) {
                    ServerDescription previousServerDescription = currentServerDescription;
                    currentServerDescription = lookupServerDescription(currentServerDescription);
                    boolean shouldStreamResponses = shouldStreamResponses(currentServerDescription);
                    if (shouldStreamResponses) {
                        ensureRoundTripTimeMonitorStarted();
                    }

                    if (isClosed) {
                        continue;
                    }

                    if (currentCheckCancelled) {
                        waitForNext();
                        currentCheckCancelled = false;
                        continue;
                    }

                    // For POLL mode, if we just established initial connection, do an immediate heartbeat
                    if (isPollMode && connection != null && !connection.isClosed()
                            && previousServerDescription.equals(unknownConnectingServerDescription)
                            && currentServerDescription.equals(connection.getInitialServerDescription())) {
                        continue;
                    }

```

rozza: This is a better check to determine if its a new initial connection.

rozza (changes requested):

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
rozza: Although I think this may be simpler and enough:

```
// when not in streaming mode, if we just established a connection, do an immediate heartbeat
if (!shouldStreamResponses && connection != null && !connection.isClosed()
    && currentServerDescription.equals(connection.getInitialServerDescription())) {
    continue;
}
```

rozza (dismissed):
LGTM - if evergreen's happy, I'm happy

rozza (approved):
LGTM!

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
```diff
@@ -216,6 +217,12 @@ public void run() {
                         continue;
                     }
 
+                    // For POLL mode, if we just established initial connection, do an immediate heartbeat
+                    if (!shouldStreamResponses && connection != null && !connection.isClosed()
+                            && currentServerDescription.equals(connection.getInitialServerDescription())) {
+                        continue;
```
vbabanin, line 223: Currently, we send two hello commands back-to-back but emit only one sequence of events (`HearbeatStarted `and `"HearbeatSucceed"`) as if only one heartbeat was sent.

When establishing a new connection, the flow in this PR looks like this:
1. At line 286, `setupNewConnectionAndGetInitialDescription()` calls `logAndNotifyHeartbeatStarted()` and sets `alreadyLoggedHeartBeatStarted = true`.
2. We receive the handshake result and return it as the `ServerDescription`.
3. We detect that this is the initial connection and immediately loop again.
4. On the next loop iteration, we hit line 264, but `alreadyLoggedHeartBeatStarted` is still true, so we don’t log another “started” event.

We then call `doHeartbeat()`, which sends a second hello command, without waiting for the selection heartbeat timeout. When it succeeds at line 322, we log the “succeeded” event and reset the flag `alreadyLoggedHeartBeatStarted = false`.

As a result we get:
- 1 ServerHeartbeatStartedEvent (for the first hello / handshake)
- 2 hello commands sent back-to-back
- 1 ServerHeartbeatSucceededEvent (for the second hello)

This conflicts with the [SDAM spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.md#heartbeats), which requires a correlation between started and succeeded/failed events, and also seem to require treating an initial handshake as a heartbeat event:

> “The driver MUST guarantee that every ServerHeartbeatStartedEvent has either a correlating ServerHeartbeatSucceededEvent or ServerHeartbeatFailedEvent.”

> “The driver MUST publish a ServerHeartbeatStartedEvent before the driver opens a connection for the next heartbeat to a server…”

Judging by the above, the initial handshake appears to be treated as the first heartbeat. The unified tests confirm this expectation, for example, the [serverMonitoringMode.yml](https://github.com/vbabanin/specifications/blob/121be10445b4df54eec5c40b924d8e3c90609cd2/source/server-discovery-and-monitoring/tests/unified/serverMonitoringMode.yml#L52-L58) unified test explicitly expects this sequence in streaming mode:

```
serverHeartbeatStartedEvent (awaited: false) - handshake
serverHeartbeatSucceededEvent (awaited: false) - handshake
serverHeartbeatStartedEvent (awaited: false) - next poll after waiting
```

If we run this test with the current implementation, we get:
```
HEARTBEAT STARTED connectionId{localValue:7} awaited: false
HEARTBEAT SUCCEED connectionId{localValue:7, serverValue:242} awaited: true
```

So the correlation between` ServerHeartbeatStartedEvent` and `ServerHeartbeatSucceededEvent` (in terms of `awaited`) does not match the expectations, and the test fails. These tests are currently ignored here:
https://github.com/vbabanin/mongo-java-driver/blob/5cb23ba8e2a9da83857f3e2f7bbb7e5cfc8dcbe2/driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java#L410-L412 - the associated ticket is already closed, and I wasn’t able to find additional details on the reason for skipping these tests.

Given the above, it seems we should:
- Align the behavior with the spec (that we missed before this PR) so that the handshake acts as the first heartbeat and we don’t send an extra hello.
- Re-enable the ignored SDAM unified tests.

What do you think?

P.S: For the regular heartbeat, we send a `ServerHeartbeatSucceededEvent` right after receiving the regular heartbeat response. We could do the same for the initial connection: emit the succeeded event right after open, since open already includes receiving the handshake.
```
connection.open(operationContextFactory.create());
roundTripTimeSampler.addSample(connection.getInitialServerDescription().getRoundTripTimeNanos());

logAndNotifyHeartbeatSucceeded(.....)                                                                                                           
```

vbabanin (changes requested):

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
stIncMale, line 1: I haven't yet looked at the [comment](https://github.com/mongodb/mongo-java-driver/pull/1856#discussion_r2634105593) left by @vbabanin, so it is possible that Slava covered the same, but given that we found a discrepancy between our behavior and the behavior a spec test expects, we should investigate the relevant specification change that we are missing / implemented incorrectly, and implement / fix it, instead of just doing an ad-hoc driver change without any reference to the specification requirement (I found no reference in this PR or in JAVA-6033).

driver-core/src/main/com/mongodb/internal/connection/DefaultServerMonitor.java
stIncMale: Thank you, Slava. I now read your comment, and see that you are saying the same (but with more details) as what I warned about in https://github.com/mongodb/mongo-java-driver/pull/1856#discussion_r2636103889.

The relevant spec change is https://github.com/mongodb/specifications/commit/f1bbc245e68916ab76cb831f6372aa5d6a50ed25, and the relevant Java driver ticket is https://jira.mongodb.org/browse/JAVA-5230.

@vbabanin, the disabled tests you identified, must have been enabled when we did https://jira.mongodb.org/browse/JAVA-5230, as stated in this Jira [comment](https://jira.mongodb.org/browse/JAVA-5230?focusedCommentId=6118489&focusedId=6118489&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6118489) on that ticket, but they weren't

> it seems we should

I completely agree with what you suggested.

stIncMale (changes requested):
.

nhachicha:
@claude review

PR #1860, opened 2026-01-12
Add support for server selection's deprioritized servers to all topologies.

vbabanin:
Relevant specification changes:

- [DRIVERS-3344 - Add support for server selection's deprioritized servers to all topologies (#1865)](https://github.com/mongodb/specifications/commit/3e577e879b60d3850ed617ea8f91eeec14c106e3)
- [DRIVERS-3380 Filter deprioritized candidates by address only (#1886)](https://github.com/mongodb/specifications/commit/9e2acc6c9f77977cdbc0705cd8e236ed4d1b6551)
- [DRIVERS-3404 - Server selection deprioritization only for overload errors on replica sets (#1900)](https://github.com/mongodb/specifications/commit/9a101dec57a95ebca1593c022fcf79cfff725577) (the change has been added on April 15, 2026 to this PR)
- [DRIVERS-3406 test deprioritized selection with tag sets (#1903)](https://github.com/mongodb/specifications/commit/19ed647f6d6ca2de231e21de25bd1279f6d0bd14) - the change has not yet been addressed in this PR, because it was introduced after creation of the PR.

The change will be as part of a subsequent PR as it concerns API surface:
- [DRIVERS-3427: Finalize client backpressure implementation for phase 1 rollout (#1919)](https://github.com/mongodb/specifications/commit/1b15cd299011172f941893cbe3c9d4737cb2ee5d) - the part in `client-backpressure.md` about `enableOverloadRetargeting`, and all the prose test changes in `source/retryable-reads/tests/README.md`. 

JAVA-6021, 
JAVA-6074,
JAVA-6105
JAVA-6114

- [x]  Has AI-assistance been used?
- Prose test refactoring to abstract class was performed by Claude Opus 4.6
- Code review was done by Claude Opus 4.6 based on @rozza Claude skill: https://github.com/rozza/ai-skills/tree/main/driver-code-review
<details> 
<summary> Code review result </summary>
JAVA-6105

Correctness & Architecture

**[praise]** The gating logic in `onAttemptFailure` is clean and minimal -- one boolean, one `if`. No over-engineering.

**[praise]** Passing `ClusterType` to `updateCandidate` rather than storing it in the deprioritized set keeps the data structure simple (`Set<ServerAddress>`).

**[nit]** `OperationContext.java:220` -- `clusterType` field is uninitialized (defaults to `null`). If `onAttemptFailure` is called before `updateCandidate`, `clusterType == ClusterType.SHARDED` is false and `candidate` is null so it returns early. Safe, but an explicit initialization `private ClusterType clusterType = ClusterType.UNKNOWN` would make the intent clearer.

**[important]** `ServerSelectionSelectionTest.java` -- The comment "Thus, SystemOverloadedError used unconditionally regardless of the cluster type" is correct but the `static import` of `format` was removed and replaced with... nothing. The `format` method is now called as an instance method `format()` on the test class itself (line 115/119 in the existing code). That's from the existing `format` helper at the bottom of the class. No issue, just noting the import change is valid.

### Performance & Concurrency

**[praise]** No performance concerns. The `instanceof` + `hasErrorLabel` check in `onAttemptFailure` is on the error path only -- not a hot path.

**[praise]** No concurrency changes. `ServerDeprioritization` is explicitly documented as not needing thread safety.

### Code Quality & Tests

**[praise]** `deprioritizableClusters()` pattern in `ServerDeprioritizationTest` is elegant -- it encodes the valid (clusterType, error) pairs that trigger deprioritization, and all test method sources derive from it. Single source of truth.

**[praise]** `onAttemptFailureIgnoresIfNonShardedWithoutOverloadError` tests both `RuntimeException` and `MongoException` without `SystemOverloadedError`, covering both branches of the conditional.

[This suggestion was adopted in the PR]**[nit]** `ServerDeprioritizationTest` line 108: `of(SHARDED_CLUSTER, new RuntimeException())` -- `RuntimeException` is used as "any error" for sharded. This works but a `MongoException` might be more realistic since server errors are always `MongoException` in practice. Not wrong, just a style choice. Using `RuntimeException` actually makes the test stronger -- it proves sharded deprioritizes on *any* `Throwable`, not just `MongoException`.

**[nit]** `AbstractRetryableReadsProseTest` -- Tests 1, 2.1, 2.2 delegate to `RetryableWritesProseTest` static methods (pre-existing pattern). Tests 3.1, 3.2 are inline. This inconsistency is acceptable since the new tests don't share logic with writes, but worth noting.

**[suggestion]** `AbstractRetryableReadsProseTest` line 54 -- `COLLECTION_NAME = "test"` is a private constant used only in 3.1/3.2. The older tests (1, 2.1, 2.2) use their own collection names inside `RetryableWritesProseTest`. After `@AfterEach` drops the database anyway, the collection name doesn't matter much. Fine as-is.

### Final Verdict: **Approve**

No blocking issues. The implementation is clean, minimal, well-tested, and aligned with the spec. The parameterized test structure in `ServerDeprioritizationTest` is a good pattern.
</details>

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -68,7 +68,6 @@ import static com.mongodb.connection.ServerDescription.MIN_DRIVER_SERVER_VERSION
 import static com.mongodb.connection.ServerDescription.MIN_DRIVER_WIRE_VERSION;
 import static com.mongodb.internal.Locks.withInterruptibleLock;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
-import static com.mongodb.internal.VisibleForTesting.AccessModifier.PROTECTED;
 import static com.mongodb.internal.connection.EventHelper.wouldDescriptionsGenerateEquivalentEvents;
 import static com.mongodb.internal.event.EventListenerHelper.singleClusterListener;
 import static com.mongodb.internal.logging.LogMessage.Component.SERVER_SELECTION;
@@ -237,8 +236,7 @@ public abstract class BaseCluster implements Cluster {
         return isClosed;
     }
 
-    @VisibleForTesting(otherwise = PROTECTED)
-    public void updateDescription(final ClusterDescription newDescription) {
+    protected void updateDescription(final ClusterDescription newDescription) {
         withLock(() -> {
             if (LOGGER.isDebugEnabled()) {
                 LOGGER.debug(format("Updating cluster description to  %s", newDescription.getShortDescription()));
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -68,6 +68,7 @@ import static com.mongodb.connection.ServerDescription.MIN_DRIVER_SERVER_VERSION
 import static com.mongodb.connection.ServerDescription.MIN_DRIVER_WIRE_VERSION;
 import static com.mongodb.internal.Locks.withInterruptibleLock;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
+import static com.mongodb.internal.VisibleForTesting.AccessModifier.PROTECTED;
 import static com.mongodb.internal.connection.EventHelper.wouldDescriptionsGenerateEquivalentEvents;
 import static com.mongodb.internal.event.EventListenerHelper.singleClusterListener;
 import static com.mongodb.internal.logging.LogMessage.Component.SERVER_SELECTION;
@@ -94,7 +95,8 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.stream.Collectors.toList;
 
-abstract class BaseCluster implements Cluster {
+@VisibleForTesting(otherwise = PRIVATE)
+public abstract class BaseCluster implements Cluster {
     private static final Logger LOGGER = Loggers.getLogger("cluster");
     private static final StructuredLogger STRUCTURED_LOGGER = new StructuredLogger("cluster");
 
@@ -112,10 +114,11 @@ abstract class BaseCluster implements Cluster {
     private volatile boolean isClosed;
     private volatile ClusterDescription description;
 
-    BaseCluster(final ClusterId clusterId,
-                final ClusterSettings settings,
-                final ClusterableServerFactory serverFactory,
-                final ClientMetadata clientMetadata) {
+    @VisibleForTesting(otherwise = PRIVATE)
+    protected BaseCluster(final ClusterId clusterId,
+                          final ClusterSettings settings,
+                          final ClusterableServerFactory serverFactory,
+                          final ClientMetadata clientMetadata) {
         this.clusterId = notNull("clusterId", clusterId);
         this.settings = notNull("settings", settings);
         this.serverFactory = notNull("serverFactory", serverFactory);
@@ -234,7 +237,8 @@ abstract class BaseCluster implements Cluster {
         return isClosed;
     }
 
-    protected void updateDescription(final ClusterDescription newDescription) {
+    @VisibleForTesting(otherwise = PROTECTED)
+    public void updateDescription(final ClusterDescription newDescription) {
         withLock(() -> {
             if (LOGGER.isDebugEnabled()) {
                 LOGGER.debug(format("Updating cluster description to  %s", newDescription.getShortDescription()));
@@ -361,8 +365,7 @@ abstract class BaseCluster implements Cluster {
             final ClusterSettings settings) {
         List<ServerSelector> selectors = Stream.of(
                 getRaceConditionPreFilteringSelector(serversSnapshot),
-                serverSelector,
-                serverDeprioritization.getServerSelector(),
+                serverDeprioritization.applyDeprioritization(serverSelector),
                 settings.getServerSelector(), // may be null
                 new LatencyMinimizingServerSelector(settings.getLocalThreshold(MILLISECONDS), MILLISECONDS),
                 AtMostTwoRandomServerSelector.instance(),
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -163,7 +163,7 @@ public abstract class BaseCluster implements Cluster {
             if (serverTuple != null) {
                 ServerAddress serverAddress = serverTuple.getServerDescription().getAddress();
                 logServerSelectionSucceeded(operationContext, clusterId, serverAddress, serverSelector, currentDescription);
-                serverDeprioritization.updateCandidate(serverAddress);
+                serverDeprioritization.updateCandidate(serverAddress, currentDescription.getType());
                 return serverTuple;
             }
             computedServerSelectionTimeout.onExpired(() ->
@@ -306,7 +306,7 @@ public abstract class BaseCluster implements Cluster {
                 if (serverTuple != null) {
                     ServerAddress serverAddress = serverTuple.getServerDescription().getAddress();
                     logServerSelectionSucceeded(operationContext, clusterId, serverAddress, request.originalSelector, description);
-                    serverDeprioritization.updateCandidate(serverAddress);
+                    serverDeprioritization.updateCandidate(serverAddress, description.getType());
                     request.onResult(serverTuple, null);
                     return true;
                 }
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -67,7 +67,6 @@ import static com.mongodb.connection.ServerDescription.MAX_DRIVER_WIRE_VERSION;
 import static com.mongodb.connection.ServerDescription.MIN_DRIVER_SERVER_VERSION;
 import static com.mongodb.connection.ServerDescription.MIN_DRIVER_WIRE_VERSION;
 import static com.mongodb.internal.Locks.withInterruptibleLock;
-import static com.mongodb.internal.VisibleForTesting.AccessModifier.PACKAGE;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PROTECTED;
 import static com.mongodb.internal.connection.EventHelper.wouldDescriptionsGenerateEquivalentEvents;
@@ -115,8 +114,7 @@ public abstract class BaseCluster implements Cluster {
     private volatile boolean isClosed;
     private volatile ClusterDescription description;
 
-    @VisibleForTesting(otherwise = PACKAGE)
-    protected BaseCluster(final ClusterId clusterId,
+    BaseCluster(final ClusterId clusterId,
                           final ClusterSettings settings,
                           final ClusterableServerFactory serverFactory,
                           final ClientMetadata clientMetadata) {
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -68,7 +68,6 @@ import static com.mongodb.connection.ServerDescription.MIN_DRIVER_SERVER_VERSION
 import static com.mongodb.connection.ServerDescription.MIN_DRIVER_WIRE_VERSION;
 import static com.mongodb.internal.Locks.withInterruptibleLock;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
-import static com.mongodb.internal.VisibleForTesting.AccessModifier.PROTECTED;
 import static com.mongodb.internal.connection.EventHelper.wouldDescriptionsGenerateEquivalentEvents;
 import static com.mongodb.internal.event.EventListenerHelper.singleClusterListener;
 import static com.mongodb.internal.logging.LogMessage.Component.SERVER_SELECTION;
@@ -95,8 +94,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.stream.Collectors.toList;
 
-@VisibleForTesting(otherwise = PROTECTED)
-public abstract class BaseCluster implements Cluster {
+abstract class BaseCluster implements Cluster {
     private static final Logger LOGGER = Loggers.getLogger("cluster");
     private static final StructuredLogger STRUCTURED_LOGGER = new StructuredLogger("cluster");
 
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -69,6 +69,7 @@ import static com.mongodb.connection.ServerDescription.MIN_DRIVER_WIRE_VERSION;
 import static com.mongodb.internal.Locks.withInterruptibleLock;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PACKAGE;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
+import static com.mongodb.internal.VisibleForTesting.AccessModifier.PROTECTED;
 import static com.mongodb.internal.connection.EventHelper.wouldDescriptionsGenerateEquivalentEvents;
 import static com.mongodb.internal.event.EventListenerHelper.singleClusterListener;
 import static com.mongodb.internal.logging.LogMessage.Component.SERVER_SELECTION;
@@ -95,7 +96,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.stream.Collectors.toList;
 
-@VisibleForTesting(otherwise = PRIVATE)
+@VisibleForTesting(otherwise = PROTECTED)
 public abstract class BaseCluster implements Cluster {
     private static final Logger LOGGER = Loggers.getLogger("cluster");
     private static final StructuredLogger STRUCTURED_LOGGER = new StructuredLogger("cluster");
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -67,6 +67,7 @@ import static com.mongodb.connection.ServerDescription.MAX_DRIVER_WIRE_VERSION;
 import static com.mongodb.connection.ServerDescription.MIN_DRIVER_SERVER_VERSION;
 import static com.mongodb.connection.ServerDescription.MIN_DRIVER_WIRE_VERSION;
 import static com.mongodb.internal.Locks.withInterruptibleLock;
+import static com.mongodb.internal.VisibleForTesting.AccessModifier.PACKAGE;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
 import static com.mongodb.internal.connection.EventHelper.wouldDescriptionsGenerateEquivalentEvents;
 import static com.mongodb.internal.event.EventListenerHelper.singleClusterListener;
@@ -113,7 +114,7 @@ public abstract class BaseCluster implements Cluster {
     private volatile boolean isClosed;
     private volatile ClusterDescription description;
 
-    @VisibleForTesting(otherwise = PRIVATE)
+    @VisibleForTesting(otherwise = PACKAGE)
     protected BaseCluster(final ClusterId clusterId,
                           final ClusterSettings settings,
                           final ClusterableServerFactory serverFactory,
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -364,7 +364,7 @@ public abstract class BaseCluster implements Cluster {
             final ClusterSettings settings) {
         List<ServerSelector> selectors = Stream.of(
                 getRaceConditionPreFilteringSelector(serversSnapshot),
-                serverDeprioritization.applyDeprioritization(serverSelector),
+                serverDeprioritization.apply(serverSelector),
                 settings.getServerSelector(), // may be null
                 new LatencyMinimizingServerSelector(settings.getLocalThreshold(MILLISECONDS), MILLISECONDS),
                 AtMostTwoRandomServerSelector.instance(),
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -22,7 +22,6 @@ import com.mongodb.RequestContext;
 import com.mongodb.ServerAddress;
 import com.mongodb.ServerApi;
 import com.mongodb.connection.ClusterDescription;
-import com.mongodb.connection.ClusterType;
 import com.mongodb.connection.ServerDescription;
 import com.mongodb.internal.IgnorableRequestContext;
 import com.mongodb.internal.TimeoutContext;
@@ -228,24 +227,23 @@ public class OperationContext {
         @Nullable
         private ServerAddress candidate;
         private final Set<ServerAddress> deprioritized;
-        private final DeprioritizingSelector selector;
 
         private ServerDeprioritization() {
             candidate = null;
             deprioritized = new HashSet<>();
-            selector = new DeprioritizingSelector();
         }
 
         /**
-         * The returned {@link ServerSelector} tries to {@linkplain ServerSelector#select(ClusterDescription) select}
-         * only the {@link ServerDescription}s that do not have deprioritized {@link ServerAddress}es.
-         * If no such {@link ServerDescription} can be selected, then it selects {@link ClusterDescription#getServerDescriptions()}.
+         * The returned {@link ServerSelector} wraps the provided selector and tries to
+         * {@linkplain ServerSelector#select(ClusterDescription) select} only the {@link ServerDescription}s
+         * that do not have deprioritized {@link ServerAddress}es. If no such {@link ServerDescription} can be selected,
+         * then it retries without filtering deprioritized servers.
          */
-        ServerSelector getServerSelector() {
-            return selector;
+        ServerSelector applyDeprioritization(final ServerSelector wrappedSelector) {
+            return new DeprioritizingSelector(wrappedSelector);
         }
 
-        void updateCandidate(final ServerAddress serverAddress) {
+        public void updateCandidate(final ServerAddress serverAddress) {
             candidate = serverAddress;
         }
 
@@ -263,24 +261,35 @@ public class OperationContext {
          * which indeed may be used concurrently. {@link DeprioritizingSelector} does not need to be thread-safe.
          */
         private final class DeprioritizingSelector implements ServerSelector {
-            private DeprioritizingSelector() {
+            private final ServerSelector wrappedSelector;
+
+            private DeprioritizingSelector(final ServerSelector wrappedSelector) {
+                this.wrappedSelector = wrappedSelector;
             }
 
             @Override
             public List<ServerDescription> select(final ClusterDescription clusterDescription) {
                 List<ServerDescription> serverDescriptions = clusterDescription.getServerDescriptions();
-                if (!isEnabled(clusterDescription.getType())) {
-                    return serverDescriptions;
+
+                if (serverDescriptions.size() == 1 || deprioritized.isEmpty()) {
+                    return wrappedSelector.select(clusterDescription);
                 }
+
                 List<ServerDescription> nonDeprioritizedServerDescriptions = serverDescriptions
                         .stream()
                         .filter(serverDescription -> !deprioritized.contains(serverDescription.getAddress()))
                         .collect(toList());
-                return nonDeprioritizedServerDescriptions.isEmpty() ? serverDescriptions : nonDeprioritizedServerDescriptions;
-            }
 
-            private boolean isEnabled(final ClusterType clusterType) {
-                return clusterType == ClusterType.SHARDED;
+                if (nonDeprioritizedServerDescriptions.isEmpty()) {
+                    return wrappedSelector.select(clusterDescription);
+                }
+
+                List<ServerDescription> selected = wrappedSelector.select(
+                        new ClusterDescription(clusterDescription.getConnectionMode(), clusterDescription.getType(),
+                                nonDeprioritizedServerDescriptions,
+                                clusterDescription.getClusterSettings(),
+                                clusterDescription.getServerSettings()));
+                return selected.isEmpty() ? wrappedSelector.select(clusterDescription) : selected;
             }
         }
     }
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -112,6 +112,12 @@ public class OperationContext {
                 operationName, tracingSpan);
     }
 
+    //TODO-JAVA
+    public OperationContext withServerDeprioritization(final ServerDeprioritization serverDeprioritization) {
+        return new OperationContext(id, requestContext, sessionContext, timeoutContext, serverDeprioritization, tracingManager, serverApi,
+                operationName, tracingSpan);
+    }
+
     public long getId() {
         return id;
     }
@@ -228,7 +234,7 @@ public class OperationContext {
         private ServerAddress candidate;
         private final Set<ServerAddress> deprioritized;
 
-        private ServerDeprioritization() {
+        public ServerDeprioritization() {
             candidate = null;
             deprioritized = new HashSet<>();
         }
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -39,6 +39,8 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
+import static com.mongodb.internal.VisibleForTesting.AccessModifier.PACKAGE;
+import static com.mongodb.internal.VisibleForTesting.AccessModifier.PRIVATE;
 import static java.util.stream.Collectors.toList;
 
 /**
@@ -112,7 +114,8 @@ public class OperationContext {
                 operationName, tracingSpan);
     }
 
-    //TODO-JAVA
+    // TODO-JAVA-6058: This method enables overriding the ServerDeprioritization state.
+    //  It is a temporary solution to handle cases where deprioritization state persists across operations.
     public OperationContext withServerDeprioritization(final ServerDeprioritization serverDeprioritization) {
         return new OperationContext(id, requestContext, sessionContext, timeoutContext, serverDeprioritization, tracingManager, serverApi,
                 operationName, tracingSpan);
@@ -234,21 +237,24 @@ public class OperationContext {
         private ServerAddress candidate;
         private final Set<ServerAddress> deprioritized;
 
+        @VisibleForTesting(otherwise = PRIVATE)
         public ServerDeprioritization() {
             candidate = null;
             deprioritized = new HashSet<>();
         }
 
         /**
-         * The returned {@link ServerSelector} wraps the provided selector and tries to
-         * {@linkplain ServerSelector#select(ClusterDescription) select} only the {@link ServerDescription}s
-         * that do not have deprioritized {@link ServerAddress}es. If no such {@link ServerDescription} can be selected,
-         * then it retries without filtering deprioritized servers.
+         * The returned {@link ServerSelector} wraps the provided selector and attempts server selection in two passes:
+         * <ol>
+         *   <li>First pass: calls the wrapped selector with only non-deprioritized {@link ServerDescription}s</li>
+         *   <li>Second pass: if the first pass returns no servers, calls the wrapped selector again with all servers (including deprioritized ones)</li>
+         * </ol>
          */
         ServerSelector applyDeprioritization(final ServerSelector wrappedSelector) {
             return new DeprioritizingSelector(wrappedSelector);
         }
 
+        @VisibleForTesting(otherwise = PACKAGE)
         public void updateCandidate(final ServerAddress serverAddress) {
             candidate = serverAddress;
         }
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -116,8 +116,8 @@ public class OperationContext {
 
     // TODO-JAVA-6058: This method enables overriding the ServerDeprioritization state.
     //  It is a temporary solution to handle cases where deprioritization state persists across operations.
-    public OperationContext withServerDeprioritization(final ServerDeprioritization serverDeprioritization) {
-        return new OperationContext(id, requestContext, sessionContext, timeoutContext, serverDeprioritization, tracingManager, serverApi,
+    public OperationContext withNewServerDeprioritization() {
+        return new OperationContext(id, requestContext, sessionContext, timeoutContext, new ServerDeprioritization(), tracingManager, serverApi,
                 operationName, tracingSpan);
     }
 
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -250,6 +250,7 @@ public class OperationContext {
                 return;
             }
 
+            // As per spec: sharded clusters deprioritize on any error, other topologies only on overload
             boolean isSystemOverloadedError = failure instanceof MongoException
                     && ((MongoException) failure).hasErrorLabel(SYSTEM_OVERLOADED_ERROR_LABEL);
             if (clusterType == ClusterType.SHARDED || isSystemOverloadedError) {
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -250,7 +250,8 @@ public class OperationContext {
                 return;
             }
 
-            boolean isSystemOverloadedError = failure instanceof MongoException && ((MongoException) failure).hasErrorLabel("SystemOverloadedError");
+            boolean isSystemOverloadedError = failure instanceof MongoException 
+                    && ((MongoException) failure).hasErrorLabel(MongoException.SYSTEM_OVERLOADED_ERROR_LABEL);
             if (clusterType == ClusterType.SHARDED || isSystemOverloadedError) {
                 deprioritized.add(candidate);
             }
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -77,7 +77,7 @@ public class OperationContext {
                 null);
     }
 
-    public static OperationContext simpleOperationContext(
+    static OperationContext simpleOperationContext(
             final TimeoutSettings timeoutSettings, @Nullable final ServerApi serverApi) {
         return new OperationContext(
                 IgnorableRequestContext.INSTANCE,
@@ -160,8 +160,7 @@ public class OperationContext {
         this.tracingSpan = tracingSpan;
     }
 
-    @VisibleForTesting(otherwise = VisibleForTesting.AccessModifier.PRIVATE)
-    public OperationContext(final long id,
+    private OperationContext(final long id,
             final RequestContext requestContext,
             final SessionContext sessionContext,
             final TimeoutContext timeoutContext,
@@ -181,26 +180,6 @@ public class OperationContext {
         this.operationName = operationName;
         this.tracingSpan = tracingSpan;
     }
-
-    @VisibleForTesting(otherwise = VisibleForTesting.AccessModifier.PRIVATE)
-    public OperationContext(final long id,
-            final RequestContext requestContext,
-            final SessionContext sessionContext,
-            final TimeoutContext timeoutContext,
-            final TracingManager tracingManager,
-            @Nullable final ServerApi serverApi,
-            @Nullable final String operationName) {
-        this.id = id;
-        this.serverDeprioritization = new ServerDeprioritization();
-        this.requestContext = requestContext;
-        this.sessionContext = sessionContext;
-        this.timeoutContext = timeoutContext;
-        this.tracingManager = tracingManager;
-        this.serverApi = serverApi;
-        this.operationName = operationName;
-        this.tracingSpan = null;
-    }
-
 
     /**
      * @return The same {@link ServerDeprioritization} if called on the same {@link OperationContext}.
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -284,6 +284,7 @@ public class OperationContext {
                         new ClusterDescription(
                                 clusterDescription.getConnectionMode(),
                                 clusterDescription.getType(),
+                                clusterDescription.getSrvResolutionException(),
                                 nonDeprioritizedServerDescriptions,
                                 clusterDescription.getClusterSettings(),
                                 clusterDescription.getServerSettings()));
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -237,8 +237,7 @@ public class OperationContext {
         private ServerAddress candidate;
         private final Set<ServerAddress> deprioritized;
 
-        @VisibleForTesting(otherwise = PRIVATE)
-        public ServerDeprioritization() {
+        private ServerDeprioritization() {
             candidate = null;
             deprioritized = new HashSet<>();
         }
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -114,8 +114,10 @@ public class OperationContext {
                 operationName, tracingSpan);
     }
 
-    // TODO-JAVA-6058: This method enables overriding the ServerDeprioritization state.
-    //  It is a temporary solution to handle cases where deprioritization state persists across operations.
+    /**
+     * TODO-JAVA-6058: This method enables overriding the ServerDeprioritization state.
+     * It is a temporary solution to handle cases where deprioritization state persists across operations.
+     */
     public OperationContext withNewServerDeprioritization() {
         return new OperationContext(id, requestContext, sessionContext, timeoutContext, new ServerDeprioritization(), tracingManager, serverApi,
                 operationName, tracingSpan);
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -41,6 +41,7 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
+import static com.mongodb.MongoException.SYSTEM_OVERLOADED_ERROR_LABEL;
 import static com.mongodb.internal.VisibleForTesting.AccessModifier.PACKAGE;
 import static java.util.stream.Collectors.toList;
 
@@ -253,7 +254,7 @@ public class OperationContext {
             }
 
             boolean isSystemOverloadedError = failure instanceof MongoException 
-                    && ((MongoException) failure).hasErrorLabel(MongoException.SYSTEM_OVERLOADED_ERROR_LABEL);
+                    && ((MongoException) failure).hasErrorLabel(SYSTEM_OVERLOADED_ERROR_LABEL);
             if (clusterType == ClusterType.SHARDED || isSystemOverloadedError) {
                 deprioritized.add(candidate);
             }
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -17,11 +17,13 @@ package com.mongodb.internal.connection;
 
 import com.mongodb.Function;
 import com.mongodb.MongoConnectionPoolClearedException;
+import com.mongodb.MongoException;
 import com.mongodb.ReadConcern;
 import com.mongodb.RequestContext;
 import com.mongodb.ServerAddress;
 import com.mongodb.ServerApi;
 import com.mongodb.connection.ClusterDescription;
+import com.mongodb.connection.ClusterType;
 import com.mongodb.connection.ServerDescription;
 import com.mongodb.internal.IgnorableRequestContext;
 import com.mongodb.internal.TimeoutContext;
@@ -215,6 +217,7 @@ public class OperationContext {
     public static final class ServerDeprioritization {
         @Nullable
         private ServerAddress candidate;
+        private ClusterType clusterType;
         private final Set<ServerAddress> deprioritized;
 
         private ServerDeprioritization() {
@@ -236,8 +239,9 @@ public class OperationContext {
         }
 
         @VisibleForTesting(otherwise = PACKAGE)
-        public void updateCandidate(final ServerAddress serverAddress) {
-            candidate = serverAddress;
+        public void updateCandidate(final ServerAddress serverAddress, final ClusterType clusterType) {
+            this.candidate = serverAddress;
+            this.clusterType = clusterType;
         }
 
         public void onAttemptFailure(final Throwable failure) {
@@ -245,7 +249,11 @@ public class OperationContext {
                 candidate = null;
                 return;
             }
-            deprioritized.add(candidate);
+
+            boolean isSystemOverloadedError = failure instanceof MongoException && ((MongoException) failure).hasErrorLabel("SystemOverloadedError");
+            if (clusterType == ClusterType.SHARDED || isSystemOverloadedError) {
+                deprioritized.add(candidate);
+            }
         }
 
         /**
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -217,6 +217,7 @@ public class OperationContext {
     public static final class ServerDeprioritization {
         @Nullable
         private ServerAddress candidate;
+        @Nullable
         private ClusterType clusterType;
         private final Set<ServerAddress> deprioritized;
 
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -264,18 +264,18 @@ public class OperationContext {
             public List<ServerDescription> select(final ClusterDescription clusterDescription) {
                 List<ServerDescription> serverDescriptions = clusterDescription.getServerDescriptions();
 
-                // TODO-JAVA-5908: Evaluate whether the early-return optimization a meaningful performance impact on server selection.
+                // TODO-JAVA-5908: Evaluate whether using the early-return optimization has a meaningful performance impact on server selection.
                 if (serverDescriptions.size() == 1 || deprioritized.isEmpty()) {
                     return wrappedSelector.select(clusterDescription);
                 }
 
-                // TODO-JAVA-5908: Evaluate whether using a loop instead of Stream a meaningful performance impact on server selection.
+                // TODO-JAVA-5908: Evaluate whether using a loop instead of Stream has a meaningful performance impact on server selection.
                 List<ServerDescription> nonDeprioritizedServerDescriptions = serverDescriptions
                         .stream()
                         .filter(serverDescription -> !deprioritized.contains(serverDescription.getAddress()))
                         .collect(toList());
 
-                // TODO-JAVA-5908: Evaluate whether the early-return optimization a meaningful performance impact on server selection.
+                // TODO-JAVA-5908: Evaluate whether the early-return optimization has a meaningful performance impact on server selection.
                 if (nonDeprioritizedServerDescriptions.isEmpty()) {
                     return wrappedSelector.select(clusterDescription);
                 }
--- a/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
@@ -244,10 +244,12 @@ public class OperationContext {
         }
 
         /**
-         * The returned {@link ServerSelector} wraps the provided selector and attempts server selection in two passes:
+         * The returned {@link ServerSelector} wraps the provided selector and attempts
+         * {@linkplain ServerSelector#select(ClusterDescription) server selection} in two passes:
          * <ol>
-         *   <li>First pass: calls the wrapped selector with only non-deprioritized {@link ServerDescription}s</li>
-         *   <li>Second pass: if the first pass returns no servers, calls the wrapped selector again with all servers (including deprioritized ones)</li>
+         *   <li>First pass: selects using the wrapped selector with only non-deprioritized {@link ServerDescription}s.</li>
+         *   <li>Second pass: if the first pass selects no {@link ServerDescription}s,
+         *   selects using the wrapped selector again with all {@link ServerDescription}s, including deprioritized ones.</li>
          * </ol>
          */
         ServerSelector applyDeprioritization(final ServerSelector wrappedSelector) {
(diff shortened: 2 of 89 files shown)


driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -234,7 +237,8 @@ public boolean isClosed() {
         return isClosed;
     }
 
-    protected void updateDescription(final ClusterDescription newDescription) {
+    @VisibleForTesting(otherwise = PROTECTED)
+    public void updateDescription(final ClusterDescription newDescription) {
```
stIncMale: It seems we can avoid making this method `public` by overriding it in the subclass. We won't even need to make it `public` in the subclass, because the test code using the method will be in the same package as the subclass. However, this requires the subclass to not be anonymous (if we could use `var`, the proposed approach would have worked even for an anonymous subclass).

driver-core/src/test/unit/com/mongodb/internal/connection/ServerSelectionSelectionTest.java
```diff
@@ -87,21 +117,25 @@ public void shouldPassAllOutcomes() {
                 fail("Should not have thrown exception: " + e);
             }
             return;
+        } catch (MongoTimeoutException mongoTimeoutException) {
+            List<ServerDescription> inLatencyWindowServers = buildServerDescriptions(definition.getArray("in_latency_window"));
+            assertTrue("Expected emtpy but was " + inLatencyWindowServers.size() + " " + definition.toJson(
+                    JsonWriterSettings.builder()
+                            .indent(true).build()), inLatencyWindowServers.isEmpty());
+            return;
```
vbabanin, line 128: Since we now perform the full server-selection path (via `BaseCluster`), the behavior of "no servers selected: is observed differently than before.

Previously, these tests invoked the `Selector` directly, got an empty list, and asserted on that result. With `BaseCluster`, server selection runs the normal selection loop: it will retry until either a server becomes selectable or the selection timeout elapses.

In this setup, a server selection timeout is the expected signal that no servers are available/eligible for selection. The timeout is set to 200ms to keep the test fast, while giving enough headroom to avoid any flakiness.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -87,21 +117,25 @@ public void shouldPassAllOutcomes() {
...
         }
-        assertServers(selectedServers, suitableServers);
-
-        ServerSelector latencyBasedServerSelector = new CompositeServerSelector(asList(serverSelector,
-                new LatencyMinimizingServerSelector(15, TimeUnit.MILLISECONDS)));
         List<ServerDescription> inLatencyWindowServers = buildServerDescriptions(definition.getArray("in_latency_window"));
-        List<ServerDescription> latencyBasedSelectedServers = latencyBasedServerSelector.select(clusterDescription);
-        assertServers(latencyBasedSelectedServers, inLatencyWindowServers);
+        assertNotNull(serverTuple);
+        assertTrue(inLatencyWindowServers.stream().anyMatch(s -> s.getAddress().equals(serverTuple.getServerDescription().getAddress())));
```
vbabanin: There’s an ambiguity in the server selection spec docs about what drivers must assert for `in_latency_window`:

`tests/README.md` says to verify the set of servers in in_latency_window.
>   "Drivers implementing server selection MUST test that their implementation correctly returns the set of servers in in_latency_window."

`server-selection-tests.md`says to verify the selection returns one of the servers in in_latency_window.
>   "Drivers implementing server selection MUST test that their implementations correctly return one of the servers in in_latency_window."


This test follows `server-selection-tests.md`, so asserting that the selected server is within the expected `in_latency_window` set is consistent with the requirements in the spec. 

**P.S:**  Both files were created in this single commit with contradicting requirements:
 ```
Feb 6, 2015 - Commit 6b63123a - "Add Server Selection Spec"

  File 1: server-selection-tests.rst
  Drivers implementing server selection MUST test that their implementations
  correctly return one of the servers in ``in_latency_window``.

  File 2: tests/README.rst
  Drivers implementing server selection MUST test that their implementations
  correctly return the set of servers in ``in_latency_window``.
```

File: driver-core/src/test/functional/com/mongodb/ClusterFixture.java [hunk 4]
```diff
@@ -193,16 +193,18 @@ public static ServerVersion getServerVersion() {
...
         }
         return serverVersion;
     }
 
-    public static final OperationContext OPERATION_CONTEXT = new OperationContext(
-            IgnorableRequestContext.INSTANCE,
-            new ReadConcernAwareNoOpSessionContext(ReadConcern.DEFAULT),
-            new TimeoutContext(TIMEOUT_SETTINGS),
-            getServerApi());
+    public static OperationContext getOperationContext() {
```
stIncMale: Thank you for noticing the issue with `OperationContext` being mutable (it was mutable before this PR), and fixing it.

The methods returning `OperationContext` in this class are a mess:

- `getOperationContext()`
- `createOperationContext(TimeoutSettings timeoutSettings)`
- `createNewOperationContext(TimeoutSettings timeoutSettings)`
- `getOperationContext(ReadPreference readPreference)`

1. Let's name the methods consistently. I think, all of the aforementioned methods should use the "create" prefix.
1.1. **Let's do this automatically via IDE in a separate commit, and express in the commit message that the commit was done via automatic refactoring, so that reviewers know not to review it.**
3. Let's remove the weirdly named and trivial `createNewOperationContext` method, and inline it where it is used (fortunately, it is used only in two places in `ClusterFixture`, and nowhere else).

driver-core/src/test/functional/com/mongodb/connection/ConnectionSpecification.groovy
```diff
@@ -32,8 +32,8 @@ class ConnectionSpecification extends OperationFunctionalSpecification {
 
     def 'should have id'() {
         when:
-        def source = getBinding().getReadConnectionSource(OPERATION_CONTEXT)
-        def connection = source.getConnection(OPERATION_CONTEXT)
+        def source = getBinding().getReadConnectionSource(getOperationContext())
+        def connection = source.getConnection(getOperationContext())
```
stIncMale: I think, the `OperationContext` used here should be the same. There may be many more changed places in this PR where this is the case, and given how many changes there are, it may not be too easy to identify them all.

For brevity, I'll be marking such places with just the **"same context"** comment. For now I am not sure if it is even practically achievable to identify them all.

The problem is exacerbated by `ClusterFixture` creating a new `OperationContext` each time it needs one, which means that the `ClusterFixture.getBinding` never uses the same context as the one used by the test calling `ClusterFixture.getBinding`.

driver-core/src/test/functional/com/mongodb/connection/ConnectionSpecification.groovy
```diff
@@ -50,8 +50,8 @@ class ConnectionSpecification extends OperationFunctionalSpecification {
                                                              new BsonInt32(getDefaultMaxMessageSize())).intValue()
         def expectedMaxBatchCount = commandResult.getNumber('maxWriteBatchSize',
                                                             new BsonInt32(getDefaultMaxWriteBatchSize())).intValue()
-        def source = getBinding().getReadConnectionSource(OPERATION_CONTEXT)
-        def connection = source.getConnection(OPERATION_CONTEXT)
+        def source = getBinding().getReadConnectionSource(getOperationContext())
+        def connection = source.getConnection(getOperationContext())
```
stIncMale: [same context](https://github.com/mongodb/mongo-java-driver/pull/1860#discussion_r2786698904)

driver-core/src/test/unit/com/mongodb/internal/connection/AbstractServerDiscoveryAndMonitoringTest.java
```diff
@@ -98,7 +98,7 @@ protected void applyApplicationError(final BsonDocument applicationError) {
         switch (type) {
             case "command":
                 exception = getCommandFailureException(applicationError.getDocument("response"), serverAddress,
-                        OPERATION_CONTEXT.getTimeoutContext());
+                        getOperationContext().getTimeoutContext());
```
stIncMale: [same context](https://github.com/mongodb/mongo-java-driver/pull/1860#discussion_r2786698904)

driver-core/src/test/unit/com/mongodb/internal/connection/ServerDiscoveryAndMonitoringTest.java
```diff
@@ -154,9 +154,9 @@ private void assertServer(final String serverName, final BsonDocument expectedSe
 
         if (expectedServerDescriptionDocument.isDocument("pool")) {
             int expectedGeneration = expectedServerDescriptionDocument.getDocument("pool").getNumber("generation").intValue();
-            Timeout serverSelectionTimeout = OPERATION_CONTEXT.getTimeoutContext().computeServerSelectionTimeout();
+            Timeout serverSelectionTimeout = getOperationContext().getTimeoutContext().computeServerSelectionTimeout();
             DefaultServer server = (DefaultServer) getCluster()
-                    .getServersSnapshot(serverSelectionTimeout, OPERATION_CONTEXT.getTimeoutContext())
+                    .getServersSnapshot(serverSelectionTimeout, getOperationContext().getTimeoutContext())
```
stIncMale: [same context](https://github.com/mongodb/mongo-java-driver/pull/1860#discussion_r2786698904)

driver-core/src/test/unit/com/mongodb/internal/connection/MultiServerClusterSpecification.groovy
```diff
@@ -96,8 +96,8 @@ class MultiServerClusterSpecification extends Specification {
 
         when:
         cluster.getServersSnapshot(
-                OPERATION_CONTEXT.getTimeoutContext().computeServerSelectionTimeout(),
-                OPERATION_CONTEXT.getTimeoutContext())
+                getOperationContext().getTimeoutContext().computeServerSelectionTimeout(),
+                getOperationContext().getTimeoutContext())
```
stIncMale: [same context](https://github.com/mongodb/mongo-java-driver/pull/1860#discussion_r2786698904)

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -113,6 +114,13 @@ public OperationContext withOperationName(final String operationName) {
                 operationName, tracingSpan);
     }
 
+    // TODO-JAVA-6058: This method enables overriding the ServerDeprioritization state.
+    //  It is a temporary solution to handle cases where deprioritization state persists across operations.
+    public OperationContext withServerDeprioritization(final ServerDeprioritization serverDeprioritization) {
```
stIncMale: we should also leave a note in the ticket description that addressing comments with the `TODO-JAVA-6058` tag is in scope of the ticket. See https://jira.mongodb.org/browse/JAVA-6005, https://jira.mongodb.org/browse/JAVA-6059 as examples. Such notes are important because they draw attention of the assignee to the tagged comments. Without a note, the assignee is more likely to not even realize there are relevant tagged comments.

stIncMale: Let's turn this comment into a [documentation comment](https://docs.oracle.com/en/java/javase/25/docs/specs/javadoc/doc-comment-spec.html#documentation-comments). Such a change for a program element that is not part of the driver API seems to have benefits with no drawbacks compared to an [end-of-line comment](https://docs.oracle.com/javase/specs/jls/se25/html/jls-3.html#jls-3.7).

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: [this suggestion is not directly related to the PR]
Let's make the `simpleOperationContext(final TimeoutSettings timeoutSettings, @Nullable final ServerApi serverApi)` method package-access. It does not need to be `public`.

stIncMale: [this suggestion is not directly related to the PR]
Let's make the

```java
OperationContext(final long id,
            final RequestContext requestContext,
            final SessionContext sessionContext,
            final TimeoutContext timeoutContext,
            final ServerDeprioritization serverDeprioritization,
            final TracingManager tracingManager,
            @Nullable final ServerApi serverApi,
            @Nullable final String operationName,
            @Nullable final Span tracingSpan)
```

constructor `private`, and remove the `@VisibleForTesting` from it. It does not need to be `public`.

stIncMale: [this suggestion is not directly related to the PR]

Let's remove the
```java
OperationContext(final long id,
            final RequestContext requestContext,
            final SessionContext sessionContext,
            final TimeoutContext timeoutContext,
            final TracingManager tracingManager,
            @Nullable final ServerApi serverApi,
            @Nullable final String operationName)
```
constructor. It is dead code.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -228,24 +236,26 @@ public static final class ServerDeprioritization {
         @Nullable
         private ServerAddress candidate;
         private final Set<ServerAddress> deprioritized;
-        private final DeprioritizingSelector selector;
 
-        private ServerDeprioritization() {
+        @VisibleForTesting(otherwise = PRIVATE)
+        public ServerDeprioritization() {
```
stIncMale: This change does not seem right:

- This constructor is not called from test code, which makes `VisibleForTesting` wrong.
- There is no need for this constructor to be `public`. It should stay `private`.
  - See https://github.com/mongodb/mongo-java-driver/pull/1860#discussion_r2794482942.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
stIncMale: There is no need for callers of the `withServerDeprioritization` method to pass `ServerDeprioritization`. Let's

- remove this parameter from the method;
- let the method create new `ServerDeprioritization` on its own;
- rename the method to something akin to `withNewServerDeprioritization`.

driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
```diff
@@ -85,7 +85,11 @@ final class ChangeStreamBatchCursor<T> implements AggregateResponseBatchCursor<T
                             final int maxWireVersion) {
         this.changeStreamOperation = changeStreamOperation;
         this.binding = binding.retain();
-        this.initialOperationContext = operationContext.withOverride(TimeoutContext::withMaxTimeAsMaxAwaitTimeOverride);
+        this.initialOperationContext = operationContext
+                .withOverride(TimeoutContext::withMaxTimeAsMaxAwaitTimeOverride)
+                //TODO-JAVA-6058. Temporary workaround to reset any server deprioritization
+                // state from the previous find operation.
+                .withServerDeprioritization(new OperationContext.ServerDeprioritization());
```
stIncMale: If the `withServerDeprioritization` method is documented as a work-around, there is no need to proliferate comments by duplicating the same information everywhere we call the method. I am proposing to remove this comment.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -94,7 +94,8 @@
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.stream.Collectors.toList;
 
-abstract class BaseCluster implements Cluster {
+@VisibleForTesting(otherwise = PRIVATE)
+public abstract class BaseCluster implements Cluster {
```
stIncMale: The value of `otherwise` is incorrect here.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -112,10 +113,11 @@ abstract class BaseCluster implements Cluster {
     private volatile boolean isClosed;
     private volatile ClusterDescription description;
 
-    BaseCluster(final ClusterId clusterId,
-                final ClusterSettings settings,
-                final ClusterableServerFactory serverFactory,
-                final ClientMetadata clientMetadata) {
+    @VisibleForTesting(otherwise = PRIVATE)
+    protected BaseCluster(final ClusterId clusterId,
```
stIncMale: The value of `otherwise` is incorrect here.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -361,8 +363,7 @@ private static ServerSelector getCompleteServerSelector(
             final ClusterSettings settings) {
         List<ServerSelector> selectors = Stream.of(
                 getRaceConditionPreFilteringSelector(serversSnapshot),
-                serverSelector,
-                serverDeprioritization.getServerSelector(),
+                serverDeprioritization.applyDeprioritization(serverSelector),
```
stIncMale: Let's rename the method to `apply`. The name of the type (`ServerDeprioritization`) of the object whose instance method we call tells what is being applied. There is no need to duplicate that in the names of methods.

Let's do this automatically via IDE in a separate commit, and express in the commit message that the commit was done via automatic refactoring, so that reviewers know not to review it.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -263,24 +273,35 @@ public void onAttemptFailure(final Throwable failure) {
...
-                    return serverDescriptions;
+
+                if (serverDescriptions.size() == 1 || deprioritized.isEmpty()) {
+                    return wrappedSelector.select(clusterDescription);
                 }
+
                 List<ServerDescription> nonDeprioritizedServerDescriptions = serverDescriptions
                         .stream()
                         .filter(serverDescription -> !deprioritized.contains(serverDescription.getAddress()))
                         .collect(toList());
```
stIncMale, line 286: Let's leave a `TODO-JAVA-XXXX` comment linking this code (the `if` optimization as well as the use of `Stream`) to the existing performance ticket about server selection. We should also mention in the description of that ticket that there are `TODO-JAVA-XXXX` comments that needs to be addressed.

The open questions here are:
- whether the `if` optimization is worth itl
- whether we should use a loop instead of using `Stream`.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -263,24 +273,35 @@ public void onAttemptFailure(final Throwable failure) {
...
-                return clusterType == ClusterType.SHARDED;
+                if (nonDeprioritizedServerDescriptions.isEmpty()) {
+                    return wrappedSelector.select(clusterDescription);
+                }
+
+                List<ServerDescription> selected = wrappedSelector.select(
+                        new ClusterDescription(clusterDescription.getConnectionMode(), clusterDescription.getType(),
+                                nonDeprioritizedServerDescriptions,
+                                clusterDescription.getClusterSettings(),
+                                clusterDescription.getServerSettings()));
```
stIncMale, line 300: We should use the
```java
public ClusterDescription(final ClusterConnectionMode connectionMode, final ClusterType type,
                          @Nullable final MongoException srvResolutionException,
                          final List<ServerDescription> serverDescriptions,
                          @Nullable final ClusterSettings clusterSettings,
                          @Nullable final ServerSettings serverSettings) {
```
constructor.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -228,24 +236,26 @@ public static final class ServerDeprioritization {
...
 
         /**
-         * The returned {@link ServerSelector} tries to {@linkplain ServerSelector#select(ClusterDescription) select}
-         * only the {@link ServerDescription}s that do not have deprioritized {@link ServerAddress}es.
-         * If no such {@link ServerDescription} can be selected, then it selects {@link ClusterDescription#getServerDescriptions()}.
+         * The returned {@link ServerSelector} wraps the provided selector and attempts server selection in two passes:
+         * <ol>
+         *   <li>First pass: calls the wrapped selector with only non-deprioritized {@link ServerDescription}s</li>
+         *   <li>Second pass: if the first pass returns no servers, calls the wrapped selector again with all servers (including deprioritized ones)</li>
+         * </ol>
```
stIncMale: [optional]

```suggestion
         * The returned {@link ServerSelector} wraps the provided selector and attempts
         * {@linkplain ServerSelector#select(ClusterDescription) server selection} in two passes:
         * <ol>
         *   <li>First pass: selects using the wrapped selector with only non-deprioritized {@link ServerDescription}s.</li>
         *   <li>Second pass: if the first pass selects no {@link ServerDescription}s,
         *   selects using the wrapped selector again with all {@link ServerDescription}s, including deprioritized ones.</li>
         * </ol>
```

stIncMale (changes requested):
The last reviewed commit is f20b48277b7a390ab1fcb92b3a4ca1cff94923c7.

The files I haven't yet reviewed:
- `ServerDeprioritizationTest`
- `ServerSelectionSelectionTest`

@vbabanin I am proposing to resolve all the outstanding comments before proceeding with the changes required by DRIVERS-3404 (for details, see the description of this PR).

stIncMale:
I thought I already proposed this, but looks like I didn't: let's not implement any new spec changes in the PR (there were spec changes) until we resolve all outstanding discussions.

driver-core/src/test/unit/com/mongodb/internal/connection/ServerDeprioritizationTest.java
```diff
@@ -56,48 +67,123 @@ void beforeEach() {
...
+                of(Named.of(generateArgumentName(singletonList(SERVER_B)), singletonList(SERVER_B))),
+                of(Named.of(generateArgumentName(singletonList(SERVER_C)), singletonList(SERVER_C))),
+                of(Named.of(generateArgumentName(asList(SERVER_A, SERVER_B)), asList(SERVER_A, SERVER_B))),
+                of(Named.of(generateArgumentName(asList(SERVER_B, SERVER_A)), asList(SERVER_B, SERVER_A))),
+                of(Named.of(generateArgumentName(asList(SERVER_A, SERVER_C)), asList(SERVER_A, SERVER_C))),
+                of(Named.of(generateArgumentName(asList(SERVER_C, SERVER_A)), asList(SERVER_C, SERVER_A))),
+                of(Named.of(generateArgumentName(ALL_SERVERS), ALL_SERVERS))
+        );
+    }
+    @ParameterizedTest
```
stIncMale, line 87: An empty line is missing after the `selectNoneDeprioritized` method and the `@ParameterizedTest` annotation on the next method.

driver-core/src/test/unit/com/mongodb/internal/connection/ServerDeprioritizationTest.java
```diff
@@ -56,48 +67,123 @@ void beforeEach() {
...
+                assertEquals(ALL_SERVERS, servers);
+                return selectorResult;
+            }
+        };
+
+        assertAll(
+                () -> assertEquals(selectorResult, serverDeprioritization.applyDeprioritization(selector.get()).select(SHARDED_CLUSTER)),
+                () -> assertEquals(selectorResult, serverDeprioritization.applyDeprioritization(selector.get()).select(REPLICA_SET_CLUSTER)),
+                () -> assertEquals(selectorResult, serverDeprioritization.applyDeprioritization(selector.get()).select(UNKNOWN_CLUSTER))
+        );
```
stIncMale: 1. Let's use Mockito (more specifically, `MongoMockito`), to create the selector (the code is shorter this way, and, after all, that's why Mockito exists).
2. Let's rename `selector` -> `selectorSupplier`, because that's what it is.

```suggestion
        Supplier<ServerSelector> selectorSupplier = () -> MongoMockito.mock(ServerSelector.class, tuner ->
                Mockito.when(tuner.select(any(ClusterDescription.class)))
                        .thenAnswer(invocation -> {
                            assertEquals(asList(SERVER_A, SERVER_C), invocation.<ClusterDescription>getArgument(0).getServerDescriptions());
                            return emptyList();
                        })
                        .thenAnswer(invocation -> {
                            assertEquals(ALL_SERVERS, invocation.<ClusterDescription>getArgument(0).getServerDescriptions());
                            return selectorResult;
                        })
            );
        assertAll(
                () -> assertEquals(selectorResult, serverDeprioritization.applyDeprioritization(selectorSupplier.get()).select(SHARDED_CLUSTER)),
                () -> assertEquals(selectorResult, serverDeprioritization.applyDeprioritization(selectorSupplier.get()).select(REPLICA_SET_CLUSTER)),
                () -> assertEquals(selectorResult, serverDeprioritization.applyDeprioritization(selectorSupplier.get()).select(UNKNOWN_CLUSTER))
        );
```

driver-core/src/test/unit/com/mongodb/internal/connection/ServerDeprioritizationTest.java
```diff
@@ -56,48 +67,123 @@ void beforeEach() {
...
     }
 
     @Test
     void onAttemptFailureIgnoresIfPoolClearedException() {
         serverDeprioritization.updateCandidate(SERVER_A.getAddress());
         serverDeprioritization.onAttemptFailure(
                 new MongoConnectionPoolClearedException(new ServerId(new ClusterId(), new ServerAddress()), null));
-        assertEquals(ALL_SERVERS, serverDeprioritization.getServerSelector().select(SHARDED_CLUSTER));
+        ServerSelector selector = createAssertingSelector(ALL_SERVERS, singletonList(SERVER_A));
+        assertEquals(singletonList(SERVER_A), serverDeprioritization.applyDeprioritization(selector).select(SHARDED_CLUSTER));
```
stIncMale: Let's return `ALL_SERVERS` from `selector` in `onAttemptFailureIgnoresIfPoolClearedException`. This way, we will not only check that the input to `selector` has all servers (that is, `serverDeprioritization` does not filter anything out at that point), but also that the `serverDeprioritization` does not have any unexpected logic of filtering the servers selected by the wrapped selector. If it had such an unexpected logic, it would have been more likely to leave `SERVER_A` be when it is the only server selected by the wrapped selector, and less likely to leave `SERVER_A` be when it is one of the servers selected by the wrapped selector. Thus, implementing `selector` such that it selects more than just `SERVER_A`, makes the test stronger.

```suggestion
        ServerSelector selector = createAssertingSelector(ALL_SERVERS, ALL_SERVERS);
        assertEquals(ALL_SERVERS, serverDeprioritization.applyDeprioritization(selector).select(SHARDED_CLUSTER));
```

stIncMale (changes requested):
The last reviewed commit is f20b48277b7a390ab1fcb92b3a4ca1cff94923c7.

- I reviewed `ServerDeprioritizationTest.java`. 
- `ServerSelectionSelectionTest.java` is yet to be reviewed.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -59,6 +80,14 @@ public class ServerSelectionSelectionTest {
     private final ClusterDescription clusterDescription;
     private final long heartbeatFrequencyMS;
```
stIncMale: [this comment is not about the current PR]

Let's remove the `heartbeatFrequencyMS` field and make it a local variable, as Idea suggests?

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -67,20 +96,19 @@ public ServerSelectionSelectionTest(final String description, final BsonDocument
...
-        assumeFalse(description.endsWith("/max-staleness/tests/ReplicaSetWithPrimary/MaxStalenessWithModePrimary.json"));
-        assumeFalse(description.contains("Deprioritized")); // TODO JAVA-6021 deprioritized server selection"
-
-        ServerSelector serverSelector = null;
-        List<ServerDescription> suitableServers = buildServerDescriptions(definition.getArray("suitable_servers", new BsonArray()));
-        List<ServerDescription> selectedServers = null;
-        try {
-            serverSelector = getServerSelector();
-            selectedServers = serverSelector.select(clusterDescription);
+        assumeTrue(!description.endsWith("/max-staleness/tests/ReplicaSetWithPrimary/MaxStalenessWithModePrimary.json"));
```
stIncMale: Why was this line changed? It seemed fine before. It is still OK now, but it now has a negation that is not necessary.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -16,40 +16,61 @@
...
 
-import static java.util.Arrays.asList;
-import static org.junit.Assert.assertEquals;
+import static com.mongodb.ClusterFixture.TIMEOUT_SETTINGS;
+import static com.mongodb.connection.ServerDescription.MIN_DRIVER_WIRE_VERSION;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
-import static org.junit.Assume.assumeFalse;
+import static org.junit.Assume.assumeTrue;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
```
stIncMale: Let's either keep using JUnit 4, or rewrite this class to use JUnit 5.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -16,40 +16,61 @@
...
+import static org.mockito.Mockito.when;
 
 // See https://github.com/mongodb/specifications/tree/master/source/server-selection/tests
```
stIncMale: [this comment is not about the current PR]

- Let's convert this comment to a class documentation comment.
- Let's change the link to https://github.com/mongodb/specifications/tree/master/source/server-selection/tests/server_selection, because the current link is misleading.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -59,6 +80,14 @@ public class ServerSelectionSelectionTest {
     private final ClusterDescription clusterDescription;
     private final long heartbeatFrequencyMS;
     private final boolean error;
+    private final List<ServerAddress> deprioritizedServerAddresses;
```
stIncMale: It does not seem that we need this field. We can call `extractDeprioritizedServerAddresses` from `createOperationContext`, which is simpler than using the `deprioritizedServerAddresses` instance field to pass the result of `extractDeprioritizedServerAddresses` to `createOperationContext`.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -89,21 +117,25 @@ public void shouldPassAllOutcomes() {
                 fail("Should not have thrown exception: " + e);
             }
             return;
+        } catch (MongoTimeoutException mongoTimeoutException) {
+            List<ServerDescription> inLatencyWindowServers = buildServerDescriptions(definition.getArray("in_latency_window"));
```
stIncMale: We should not duplicate this line of code here and below. Instead, let's extract it to a place above the current `try` statement.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -67,20 +96,19 @@ public ServerSelectionSelectionTest(final String description, final BsonDocument
...
-        List<ServerDescription> selectedServers = null;
-        try {
-            serverSelector = getServerSelector();
-            selectedServers = serverSelector.select(clusterDescription);
+        assumeTrue(!description.endsWith("/max-staleness/tests/ReplicaSetWithPrimary/MaxStalenessWithModePrimary.json"));
+        ServerTuple serverTuple;
+        ServerSelector serverSelector = getServerSelector();
+        OperationContext operationContext = createOperationContext();
+        Cluster.ServersSnapshot serversSnapshot = createServersSnapshot(clusterDescription);
+        try (BaseCluster cluster = createTestCluster(clusterDescription, serversSnapshot);) {
```
stIncMale: What is the purpose of `;` here?

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -89,21 +117,25 @@ public void shouldPassAllOutcomes() {
                 fail("Should not have thrown exception: " + e);
             }
             return;
+        } catch (MongoTimeoutException mongoTimeoutException) {
+            List<ServerDescription> inLatencyWindowServers = buildServerDescriptions(definition.getArray("in_latency_window"));
+            assertTrue("Expected emtpy but was " + inLatencyWindowServers.size() + " " + definition.toJson(
+                    JsonWriterSettings.builder()
+                            .indent(true).build()), inLatencyWindowServers.isEmpty());
```
stIncMale: This is the only place where we add the `definition` to the assertion error thrown. We should either do this consistently, or not do at all. The latter makes more sense (and that's what I propose), because each parameterized test description points to the corresponding scenario file.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -89,21 +117,25 @@ public void shouldPassAllOutcomes() {
...
         }
-        assertServers(selectedServers, suitableServers);
-
-        ServerSelector latencyBasedServerSelector = new CompositeServerSelector(asList(serverSelector,
-                new LatencyMinimizingServerSelector(15, TimeUnit.MILLISECONDS)));
         List<ServerDescription> inLatencyWindowServers = buildServerDescriptions(definition.getArray("in_latency_window"));
-        List<ServerDescription> latencyBasedSelectedServers = latencyBasedServerSelector.select(clusterDescription);
-        assertServers(latencyBasedSelectedServers, inLatencyWindowServers);
+        assertNotNull(serverTuple);
+        assertTrue(inLatencyWindowServers.stream().anyMatch(s -> s.getAddress().equals(serverTuple.getServerDescription().getAddress())));
```
stIncMale: We relied on `ServerDescription.equals` previously, and there seem to be no reason to do otherwise


```suggestion
        assertTrue(inLatencyWindowServers.stream().anyMatch(s -> s.equals(serverTuple.getServerDescription())));
```

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -89,21 +117,25 @@ public void shouldPassAllOutcomes() {
...
     }
 
     @Parameterized.Parameters(name = "{0}")
     public static Collection<Object[]> data() {
         List<Object[]> data = new ArrayList<>();
+        //source/server-selection/tests/server_selection/ReplicaSetNoPrimary/read/DeprioritizedPrimary.json
```
stIncMale: What does this comment mean?

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -244,8 +281,78 @@ private ServerSelector getServerSelector() {
...
+                .map(BsonValue::asDocument)
+                .map(ServerSelectionSelectionTest::buildServerDescription)
+                .map(ServerDescription::getAddress)
+                .collect(Collectors.toList());
+    }
+
+    private OperationContext createOperationContext() {
+        OperationContext operationContext =
+                OperationContext.simpleOperationContext(
+                        new TimeoutContext(TIMEOUT_SETTINGS.withServerSelectionTimeoutMS(SERVER_SELECTION_TIMEOUT_MS)));
```
stIncMale: Let's use 0 (["immediate" timeout](https://mongodb.github.io/mongo-java-driver/5.6/apidocs/driver-core/com/mongodb/connection/ClusterSettings.Builder.html#serverSelectionTimeout(long,java.util.concurrent.TimeUnit))) and remove the `SERVER_SELECTION_TIMEOUT_MS` constant:


```suggestion
                        new TimeoutContext(TIMEOUT_SETTINGS.withServerSelectionTimeoutMS(0)));
```

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -244,8 +281,78 @@ private ServerSelector getServerSelector() {
...
+        OperationContext.ServerDeprioritization serverDeprioritization = operationContext.getServerDeprioritization();
+        for (ServerAddress address : deprioritizedServerAddresses) {
+            serverDeprioritization.updateCandidate(address);
+            serverDeprioritization.onAttemptFailure(new MongoConfigurationException("test"));
```
stIncMale: While not being important, `MongoConfigurationException` is quite a weird exception here: why would we deprioritize a server due to a client configuration exception? Of course, in production it does not matter, because such an exception does not trigger a retry. But for the sake of making this less surprising, let's use something else, like `MongoException`?

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -244,8 +281,78 @@ private ServerSelector getServerSelector() {
...
+        }
+        return operationContext;
+    }
+
+    private static Cluster.ServersSnapshot createServersSnapshot(
+            final ClusterDescription clusterDescription) {
+        Map<ServerAddress, Server> serverMap = new HashMap<>();
+        for (ServerDescription desc : clusterDescription.getServerDescriptions()) {
+            serverMap.put(desc.getAddress(), MongoMockito.mock(Server.class, server -> {
+                // Operation count selector should select any server since they all have 0 operation count.
```
stIncMale:
```suggestion
                // `MinimumOperationCountServerSelector` should select any server since they all have 0 operation count.
```

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -244,8 +281,78 @@ private ServerSelector getServerSelector() {
...
+                when(server.operationCount()).thenReturn(0);
+            }));
+        }
+        return serverMap::get;
+    }
+
+    private BaseCluster createTestCluster(final ClusterDescription clusterDescription, final Cluster.ServersSnapshot serversSnapshot) {
+        return new TestCluster(clusterDescription, serversSnapshot);
+    }
```
stIncMale: What is the purpose of this method?

P.S. If it is really needed, it should be `static`.

driver-core/src/test/unit/com/mongodb/internal/connection/ServerSelectionSelectionTest.java
```diff
@@ -244,8 +281,78 @@ private ServerSelector getServerSelector() {
...
+
+        @Override
+        public ServersSnapshot getServersSnapshot(final Timeout serverSelectionTimeout, final TimeoutContext timeoutContext) {
+            return serversSnapshot;
+        }
+
+        @Override
+        public void onChange(final ServerDescriptionChangedEvent event) {
+            // We do not expect this to be called during server selection.
+            Assertions.fail();
```
stIncMale, line 356: 1. Does this comment communicate something that is not already obvious? It seems the assertion makes it clear that we don't expect the method to be called.
2. We should call `com.mongodb.assertions.Assertions.fail` here instead of a JUnit assertion method: we do not check that the `onChange` is not called, we simply do not expect the `onChange` method to be called on this test double, so we don't care to implement it, and that's why we simply fail (it's the same behavior we have in `MongoMockito`, and we use `com.mongodb.assertions.Assertions.fail` there, see `InsufficientStubbingDetector.answer`).

stIncMale (changes requested):
The last reviewed commit is f20b48277b7a390ab1fcb92b3a4ca1cff94923c7.

Everything has been reviewed.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Ah, this wasn’t intentional. The `assumeTrue(!...)` form is the older variant (it exists on 5.5.x/5.6.x and pre-cleanup main). We later switched `main` to `assumeFalse(...)` in 41ca107ca1. This branch was created before that change, and I likely reintroduced the old form during a merge of `main` into this PR. 

I have updated it to `assumeFalse(...)` here to match main and remove the negation. Commit: 32609f64b1

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Good point. The intent of including `definition.toJson(...)` was to make CI failures self-contained (so the full scenario is visible in logs without having to open the JSON file), similar to how `UnifiedTest` assertions include the full test definition via `AssertionContext`.

I’m leaning toward including definition for all assertions so failures are easier to debug from CI logs. What do you think?

Commit: cdee4fa0b6
Subsequent commit: 9a84fc0c0921643ef1f90316a17e6cc74cad1d32

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: This was a leftover comment used as a reference example during testing. It doesn't serve any purpose now. Removed, thanks!

Commit: 12ceda72e1

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: This was a typo. Removed, thanks! 
Commit: 2f06765f5a

driver-core/src/test/functional/com/mongodb/connection/ConnectionSpecification.groovy
vbabanin: Done in: a1c052a4ad
Identified one more place: 817a5c53ab20da77b93ad36b1e938df4b6d96839

driver-core/src/test/functional/com/mongodb/ClusterFixture.java
vbabanin: Done in:
1. 893d419048
2. 19ab19086b

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
vbabanin: Done in: 98caa34ae6

driver-core/src/test/unit/com/mongodb/internal/connection/ServerDeprioritizationTest.java
vbabanin: Done in: 2bc1b2eee2

driver-core/src/test/unit/com/mongodb/internal/connection/ServerDeprioritizationTest.java
vbabanin: Done in: 8a8bf0a1cf.

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Done in: cf5c48446c

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Done in: 7d3d8625b5

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Done in: eff9e2351f

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Done in: 78d685d92d

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Done in: ff279de7fd

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Done in: b16bfb0d6b

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Done in: a7fc3fc939

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
vbabanin: Done in:  1d15a73e74

driver-core/src/test/unit/com/mongodb/internal/connection/ServerSelectionSelectionTest.java
vbabanin: Done in: 3e2c0cd0dde21f1fe9b175049574245e1da87293

driver-core/src/main/com/mongodb/internal/operation/ChangeStreamBatchCursor.java
vbabanin: Done in:  2df458066c

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
vbabanin: Done in: db936a184a

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
vbabanin: Done in: 64dd4f4869

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
vbabanin: Done in: cd67117e39

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
vbabanin: Done in: f57616d2e1

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
vbabanin: Done in: 0a4f04fe76

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
vbabanin: Done in: 08b34666

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -244,8 +281,78 @@ private ServerSelector getServerSelector() {
...
+
+        @Override
+        public ServersSnapshot getServersSnapshot(final Timeout serverSelectionTimeout, final TimeoutContext timeoutContext) {
+            return serversSnapshot;
+        }
+
+        @Override
+        public void onChange(final ServerDescriptionChangedEvent event) {
+            // We do not expect this to be called during server selection.
+            Assertions.fail();
```
vbabanin: Done in: 3e2c0cd0dd

driver-core/src/test/unit/com/mongodb/internal/connection/MultiServerClusterSpecification.groovy
vbabanin: Done in: 43108945

driver-core/src/test/unit/com/mongodb/internal/connection/ServerDiscoveryAndMonitoringTest.java
vbabanin: Done in: 43108945

driver-core/src/test/unit/com/mongodb/internal/connection/AbstractServerDiscoveryAndMonitoringTest.java
vbabanin: Done in: 43108945

driver-core/src/test/functional/com/mongodb/connection/ConnectionSpecification.groovy
vbabanin: Done in: 43108945

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
vbabanin: Done in: cd67117e3900e36111076888ef823e57178e135c

vbabanin: Done in: cd67117e3900e36111076888ef823e57178e135c

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
vbabanin: Done in: 
a6587fad99 
c97cb65c6a5046797993d09c562d5805dfc22935
012aa6b2ec793c115358394e1433ebf978b3ebc9

stIncMale:
@vbabanin, I updated the PR description with the information about DRIVERS-3427 and the corresponding JAVA-6141.

Could you please further update the description, if needed, such that it correctly reflects which of the things that are supposed to be part of the PR, have already been done in the PR, and what will be done later, once all the outstanding threads are resolved (I remember asking not to implement anything new to avoid complicating the review further).

vbabanin:
> @vbabanin, I updated the PR description with the information about [DRIVERS-3427](https://jira.mongodb.org/browse/DRIVERS-3427) and the corresponding [JAVA-6141](https://jira.mongodb.org/browse/JAVA-6141).
> 
> Could you please further update the description, if needed, such that it correctly reflects which of the things that are supposed to be part of the PR, have already been done in the PR, and what will be done later, once all the outstanding threads are resolved (I remember asking not to implement anything new to avoid complicating the review further).

@stIncMale Thank you for updating the PR description.

I've updated the PR description to reflect what is expected to be delivered in this PR and what will follow as subsequent work. 

**Note:**  You might notice a `force push` here. The PR has been rebased onto the `backpressure` feature branch (previously based on `main`). This changed all commit SHAs linked in reply comments; I've updated my comment replies with the new hashes, however the commits themselves are identical in content. The rebase was needed because this work depends on the `backpressure` branch.

driver-core/src/test/unit/com/mongodb/internal/connection/SingleServerClusterSpecification.groovy
```diff
@@ -95,8 +95,8 @@ class SingleServerClusterSpecification extends Specification {
         cluster.close()
 
         when:
-        cluster.getServersSnapshot(OPERATION_CONTEXT.getTimeoutContext().computeServerSelectionTimeout(),
-                OPERATION_CONTEXT.getTimeoutContext())
+        cluster.getServersSnapshot(createOperationContext().getTimeoutContext().computeServerSelectionTimeout(),
+                createOperationContext().getTimeoutContext())
```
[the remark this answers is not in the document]
vbabanin: Done in: https://github.com/mongodb/mongo-java-driver/pull/1860/changes/446f8fff082bd4926c4dd74470682e4982ce0635

driver-core/src/test/unit/com/mongodb/internal/connection/SingleServerClusterSpecification.groovy
```diff
@@ -78,10 +78,10 @@ class SingleServerClusterSpecification extends Specification {
         sendNotification(firstServer, STANDALONE)
 
         then:
-        cluster.getServersSnapshot(OPERATION_CONTEXT
+        cluster.getServersSnapshot(createOperationContext()
                         .getTimeoutContext()
                         .computeServerSelectionTimeout(),
-                OPERATION_CONTEXT.getTimeoutContext()).getServer(firstServer) == factory.getServer(firstServer)
+                createOperationContext().getTimeoutContext()).getServer(firstServer) == factory.getServer(firstServer)
```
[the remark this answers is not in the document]
vbabanin: Done in: https://github.com/mongodb/mongo-java-driver/pull/1860/changes/446f8fff082bd4926c4dd74470682e4982ce0635

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
vbabanin: Done in: https://github.com/mongodb/mongo-java-driver/pull/1860/changes/446f8fff082bd4926c4dd74470682e4982ce0635

driver-core/src/test/unit/com/mongodb/connection/ServerSelectionSelectionTest.java
```diff
@@ -244,8 +280,82 @@ private ServerSelector getServerSelector() {
...
+        OperationContext operationContext =
+                OperationContext.simpleOperationContext(
+                        new TimeoutContext(TIMEOUT_SETTINGS.withServerSelectionTimeoutMS(0)));
+        OperationContext.ServerDeprioritization serverDeprioritization = operationContext.getServerDeprioritization();
+        for (ServerAddress address : extractDeprioritizedServerAddresses(definition)) {
+            serverDeprioritization.updateCandidate(address, clusterDescription.getType());
+            // The spec defines deprioritized_servers as a pre-populated list to feed into the selection mechanism - not as "simulate the
+            // failure that caused deprioritization." Thus, SystemOverloadedError used unconditionally regardless of the cluster type.
+            MongoException error = new MongoException("test");
+            error.addLabel("SystemOverloadedError");
```
rozza:
```suggestion
            error.addLabel(MongoException.SYSTEM_OVERLOADED_ERROR_LABEL);
```

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -227,34 +217,43 @@ public OperationContext withOverride(final TimeoutContextOverride timeoutContext
...
         }
 
         public void onAttemptFailure(final Throwable failure) {
             if (candidate == null || failure instanceof MongoConnectionPoolClearedException) {
                 candidate = null;
                 return;
             }
-            deprioritized.add(candidate);
+
+            boolean isSystemOverloadedError = failure instanceof MongoException && ((MongoException) failure).hasErrorLabel("SystemOverloadedError");
```
rozza:
```suggestion
            boolean isSystemOverloadedError = failure instanceof MongoException 
                    && ((MongoException) failure).hasErrorLabel(MongoException.SYSTEM_OVERLOADED_ERROR_LABEL);
```

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
rozza: Resolving as has been done.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -263,24 +254,40 @@ public void onAttemptFailure(final Throwable failure) {
...
+                    return wrappedSelector.select(clusterDescription);
+                }
+
+                List<ServerDescription> selected = wrappedSelector.select(
+                        new ClusterDescription(
+                                clusterDescription.getConnectionMode(),
+                                clusterDescription.getType(),
+                                nonDeprioritizedServerDescriptions,
+                                clusterDescription.getClusterSettings(),
+                                clusterDescription.getServerSettings()));
```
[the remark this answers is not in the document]
rozza: This is now out of date.

rozza (changes requested):
Initial review a couple of nits looking good.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -227,34 +217,43 @@ public OperationContext withOverride(final TimeoutContextOverride timeoutContext
     public static final class ServerDeprioritization {
         @Nullable
         private ServerAddress candidate;
+        private ClusterType clusterType;
```
rozza, line 220: Nit: Lets be explicit about the nullability of clusterType and set it in the constructor.

```suggestion
        @Nullable
        private ClusterType clusterType;
```

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -227,34 +217,43 @@ public OperationContext withOverride(final TimeoutContextOverride timeoutContext
...
         private final Set<ServerAddress> deprioritized;
-        private final DeprioritizingSelector selector;
 
         private ServerDeprioritization() {
             candidate = null;
             deprioritized = new HashSet<>();
```
rozza, line 225:
```suggestion
            deprioritized = new HashSet<>();
            clusterType = null;
```

rozza (changes requested):
A couple more nits.

AsyncChangeStreamBatchCursor should also be updated to match its sync counter part

driver-core/src/test/unit/com/mongodb/internal/connection/DefaultServerSpecification.groovy
```diff
@@ -210,7 +210,7 @@ class DefaultServerSpecification extends Specification {
         given:
         def connectionPool = Mock(ConnectionPool)
         def serverMonitor = Mock(ServerMonitor)
-        connectionPool.get(OPERATION_CONTEXT) >> { throw exceptionToThrow }
+        connectionPool.get(createOperationContext()) >> { throw exceptionToThrow }
```
rozza:
```suggestion
```
This can be removed as per the Copilot message:

> Spock stub uses connectionPool.get(createOperationContext()), but createOperationContext() returns a new instance each call and OperationContext doesn’t override equals, so this stub will not match real invocations. Use a wildcard argument (get(_)) or bind the created operationContext to a variable and use the same instance in both the stub and the call site(s).

seems it copy pasta and wasn't needed.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
rozza: Question - is this annotation even needed now?

driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableReadsProseTest.java
```diff
@@ -0,0 +1,202 @@
...
+                        return cursor.hasNext();
+                    }
+                }, "find", false);
+    }
+
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.md#31-retryable-reads-caused-by-overload-errors-are-retried-on-a-different-replicaset-server-when-one-is-available">
+     * 3.1 Retryable Reads Caused by Overload Errors Are Retried on a Different Replicaset Server When One is Available</a>.
+     */
+    //TODO-JAVA-6141 add overloadRetargeting into tests.
```
stIncMale: We discussed at today's catchup, mentioning so that we don't forget:

This will be done as part of JAVA-6167, and the TODO tags we use in backpressure are documented in the description of https://github.com/mongodb/mongo-java-driver/pull/1918: `TODO-BACKPRESSURE`/`TODO-BACKPRESSURE <engineer-name>`.

The same applies to the comment tagged the same way below.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -112,10 +115,11 @@ abstract class BaseCluster implements Cluster {
...
 
-    BaseCluster(final ClusterId clusterId,
-                final ClusterSettings settings,
-                final ClusterableServerFactory serverFactory,
-                final ClientMetadata clientMetadata) {
+    @VisibleForTesting(otherwise = PACKAGE)
+    protected BaseCluster(final ClusterId clusterId,
+                          final ClusterSettings settings,
+                          final ClusterableServerFactory serverFactory,
+                          final ClientMetadata clientMetadata) {
```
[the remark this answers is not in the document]
vbabanin: Not a real issue, however, i moved it to the test class from `com.mongodb.connection` to `com.mongodb.internal.connection`  for simplicity.

driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableReadsProseTest.java
vbabanin: Updated, thank you!

driver-core/src/test/unit/com/mongodb/internal/connection/ServerDeprioritizationTest.java
```diff
@@ -25,93 +26,223 @@
...
+                    serverDeprioritization.updateCandidate(SERVER_B.getAddress(), clusterType);
+                    serverDeprioritization.onAttemptFailure(new RuntimeException());
+                    assertEquals(singletonList(SERVER_A), serverDeprioritization.apply(selector).select(SHARDED_CLUSTER),
+                            "Expected no deprioritization for " + clusterType + " with RuntimeException");
+                }, () -> {
+                    serverDeprioritization = createOperationContext(TIMEOUT_SETTINGS).getServerDeprioritization();
+                    serverDeprioritization.updateCandidate(SERVER_B.getAddress(), clusterType);
+                    serverDeprioritization.onAttemptFailure(new MongoException(1, "error"));
+                    assertEquals(singletonList(SERVER_A), serverDeprioritization.apply(selector).select(SHARDED_CLUSTER),
+                            "Expected no deprioritization for " + clusterType + " with no SystemOverloadedError MongoException");
```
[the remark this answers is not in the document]
vbabanin: Done in https://github.com/mongodb/mongo-java-driver/pull/1860/changes/a911780f5313c9d645492c9f77bea4848def8c03.

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
vbabanin: Good catch. Removed.

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -227,34 +218,46 @@ public OperationContext withOverride(final TimeoutContextOverride timeoutContext
...
-        ServerSelector getServerSelector() {
-            return selector;
+        ServerSelector apply(final ServerSelector wrappedSelector) {
+            return new DeprioritizingSelector(wrappedSelector);
         }
 
-        void updateCandidate(final ServerAddress serverAddress) {
-            candidate = serverAddress;
+        @VisibleForTesting(otherwise = PACKAGE)
+        public void updateCandidate(final ServerAddress serverAddress, final ClusterType clusterType) {
```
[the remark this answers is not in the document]
vbabanin: Done

driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
```diff
@@ -94,7 +95,8 @@
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.stream.Collectors.toList;
 
-abstract class BaseCluster implements Cluster {
+@VisibleForTesting(otherwise = PROTECTED)
+public abstract class BaseCluster implements Cluster {
```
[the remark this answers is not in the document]
vbabanin: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractRetryableReadsProseTest.java
```diff
@@ -0,0 +1,232 @@
...
+                event -> {
+                    ClusterDescription desc = event.getNewDescription();
+                    // We need both primary and secondary to be discovered (not UNKNOWN) before running the test.
+                    //
+                    // 1. The failpoint is set on the primary. If the primary is not yet discovered,
+                    //    primaryPreferred may route the find to a secondary, and the failpoint never fires.
+                    //
+                    // 2. When the primary is deprioritized on retry, primaryPreferred falls back to a secondary.
+                    //    If the secondaries are still UNKNOWN at that point, the fallback yields no selectable servers,
+                    //    causing the deprioritized primary to be selected again.
```
vbabanin: The specific flaky scenario encountered during debugging: [Evergreen log line](https://parsley.corp.mongodb.com/evergreen/mongo_java_driver_tests_jdk8_unsecure__version~6.0_os~linux_topology~replicaset_auth~noauth_ssl~nossl_jdk~jdk8_test_sync_task_patch_e5d8668e1a19fb4e0b5414a1867a0fa006957492_69e59688c4efd700071d8ca0_26_04_20_02_59_36/0/task?bookmarks=0,191649&shareLine=47029)

> [overloadErrorRetriedOnDifferentReplicaSetServer] [DeprioritizingSelector.select] servers=3 deprioritized=[localhost:27017] clusterType=REPLICA_SET serverDetails=localhost:27017(REPLICA_SET_PRIMARY), localhost:27018(UNKNOWN), localhost:27019(UNKNOWN)

driver-core/src/main/com/mongodb/internal/connection/OperationContext.java
```diff
@@ -227,34 +216,45 @@ public OperationContext withOverride(final TimeoutContextOverride timeoutContext
...
         public void onAttemptFailure(final Throwable failure) {
             if (candidate == null || failure instanceof MongoConnectionPoolClearedException) {
                 candidate = null;
                 return;
             }
-            deprioritized.add(candidate);
+
+            boolean isSystemOverloadedError = failure instanceof MongoException
+                    && ((MongoException) failure).hasErrorLabel(SYSTEM_OVERLOADED_ERROR_LABEL);
+            if (clusterType == ClusterType.SHARDED || isSystemOverloadedError) {
```
rozza, line 256: Nit and optional: Add a note explaining this logic

```suggestion
            // As per spec: sharded clusters deprioritize on any error, other topologies only on overload
            boolean isSystemOverloadedError = failure instanceof MongoException
                    && ((MongoException) failure).hasErrorLabel(SYSTEM_OVERLOADED_ERROR_LABEL);
            if (clusterType == ClusterType.SHARDED || isSystemOverloadedError) {
```

rozza (dismissed):
LGTM - one optional comment to help grok the logic

rozza (approved):
LGTM!

PR #1863, opened 2026-01-14
Make asynchronous ClientSession.close() idempotent.

vbabanin:
JAVA-6062


driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/MongoClientSessionSpecification.groovy
```diff
@@ -42,6 +42,11 @@ import static com.mongodb.ClusterFixture.isDiscoverableReplicaSet
 import static com.mongodb.reactivestreams.client.Fixture.getDefaultDatabase
 import static com.mongodb.reactivestreams.client.Fixture.getMongoClientBuilderFromConnectionString
 
+/**
```
rozza, line 45: Instead of doing this I just use copilot to port all the tests - it cleans up as we go along.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/MongoClientSessionFunctionalTest.java
```diff
@@ -0,0 +1,126 @@
...
+            Mono.from(collection.insertOne(session, new Document("x", 1))).block(TIMEOUT_DURATION);
+
+            // when
+            commandListener.reset();
+            session.close();
+            session.close();
+            session.close();
+
+            // wait for async abort operations to complete
+            sleep(1000);
```
rozza: This looks like a race waiting for evergreen to find

rozza (changes requested):
I think the tests might need some work.

PR #1880, opened 2026-02-03
Add RetryableWriteError to write concern error in bulk writes.

vbabanin:
This PR fixes a label-loss in bulk write error handling. When a bulk write returns `ok: 1` with a retryable `writeConcernError`, the driver may decide to attach the `RetryableWriteError` label, but the label was being added to the `writeConcernError` subdocument. `MongoBulkWriteException` parsing reads labels from the top-level `errorLabels` array, so labels attached under `writeConcernError` were not added to the thrown `MongoBulkWriteException`.

### Change
Attach `RetryableWriteError` into the top-level result`["errorLabels"]` for the relevant write concern error path (instead of result`["writeConcernError"]`).

Keep behavior consistent with server-originated labels (raw result is not exposed to users; labels are processed via existing parsing).

### Tests
Re-enabled unified test (now passing).

JAVA-6087


driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -287,8 +287,11 @@ private BulkWriteResult executeBulkWriteBatch(
                             connection.getDescription().getServerAddress(), "errMsg", timeoutContext);
                     if (writeConcernBasedError != null) {
                         if (currentBulkWriteTracker.lastAttempt()) {
-                            addRetryableWriteErrorLabel(writeConcernBasedError, maxWireVersion);
-                            addErrorLabelsToWriteConcern(result.getDocument("writeConcernError"), writeConcernBasedError.getErrorLabels());
+                            // The raw result is not exposed to the user, so labels can be added directly here.
+                            // They are further processed as if they originated from the server.
+                            if (shouldAddRetryableWriteErrorLabel(writeConcernBasedError, maxWireVersion)) {
+                                addRetryableErrorLabelToResult(result);
+                            }
```
stIncMale, line 294: Does this comment actually add something useful? I think, the driver never returns raw results, except for the `runCommand` method.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
```diff
@@ -456,9 +461,11 @@ private boolean shouldExpectResponse(final BulkWriteBatch batch, final WriteConc
         return effectiveWriteConcern.isAcknowledged() || (ordered && batch.hasAnotherBatch());
     }
 
-    private void addErrorLabelsToWriteConcern(final BsonDocument result, final Set<String> errorLabels) {
+    private void addRetryableErrorLabelToResult(final BsonDocument result) {
         if (!result.containsKey("errorLabels")) {
-            result.put("errorLabels", new BsonArray(errorLabels.stream().map(BsonString::new).collect(Collectors.toList())));
+            BsonArray labelsArray = new BsonArray();
+            labelsArray.add(new BsonString(RETRYABLE_WRITE_ERROR_LABEL));
```
stIncMale, line 467: I fail to understand why we are adding the label at all, given that the label is not part of our public contract, and its semantics is unknown to users.

It is also weird that the `RetryableWriteError label is added based on writeConcernError in pre-4.4 mongod response` test is only for MongoDB <4.4. What's different in the newer versions?

stIncMale (commented):
The last reviewed commit is 789be1315b2e470553fe7c4b639bef22f0bee455.

driver-core/src/main/com/mongodb/internal/operation/MixedBulkWriteOperation.java
stIncMale: We discussed this over Zoom, @rozza suggested that the reason the spec test verifies the label is likely to indirectly check how the retry mechanism works, and not to make sure that the application gets the `RetryableWriteError` label.

As for the MongoDB 4.4 - the difference is that starting at this version, the server adds the `RetryableWriteError` label.

stIncMale: Leaving this unresolved, as this comment may be useful for the posterity.

PR #1882, opened 2026-02-06
Fix release notes diff in github release actions

atandon2024:
Fix GitHub Actions release notes generation - Corrects an issue where draft release notes included commits from earlier releases due to an incorrect comparison range. Additional changes were necessary due to feedback from the semgrep-code bot, specifically moving variables to env vars.

Tested: https://github.com/atandon2024/mongo-java-driver/releases/tag/untagged-d68ec2c3f4e63cb2b5e0


.github/workflows/release.yml
```diff
@@ -23,11 +23,12 @@ jobs:
     steps:
       - name: "Create release output"
         run: echo '🎬 Release process for version ${{ env.RELEASE_VERSION }} started by @${{ github.triggering_actor }}' >> $GITHUB_STEP_SUMMARY
-
+      
       - uses: mongodb-labs/drivers-github-tools/secure-checkout@v2
         with:
           app_id: ${{ vars.APP_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
+          fetch-depth: 0
```
rozza, line 31: What does this do?

I think `fetch-depth: 0` fetches all history for all branches and tags. But is it just for the `mongodb-labs/drivers-github-tools/secure-checkout` or is it for the java driver repo itself?

rozza (commented):
One question about fetch depth.

.github/workflows/release.yml
atandon2024: @rozza yes fetch-depth 0 fetches all branches and tags. 

Per the GitHub Actions documentation (https://docs.github.com/en/actions/tutorials/create-an-example-workflow), drivers-github-tools is only pulled internally to run the action itself. The checkout operates on github.ref.

rozza (dismissed):
LGTM!

.github/workflows/release.yml
vbabanin, line 1: Have we tested this in the fork yet? I also noticed the note: "Needs to be tested by triggering GitHub Actions manually before merging.”

vbabanin (changes requested):

.github/workflows/release.yml
atandon2024: @vbabanin tested finally - https://github.com/atandon2024/mongo-java-driver/releases/tag/untagged-d68ec2c3f4e63cb2b5e0

.github/workflows/release.yml
```diff
@@ -42,29 +43,36 @@ jobs:
...
         #   Example: 5.2.0-beta1 => <current branch>
+        env:
+          INPUT_VERSION: ${{ inputs.version }}
         run: |
-          echo RELEASE_VERSION=${{ inputs.version }} >> $GITHUB_ENV
-          echo RELEASE_VERSION_WITHOUT_SUFFIX=$(echo ${{ inputs.version }} | awk -F- '{print $1}') >> $GITHUB_ENV
-          if [[ "${{ inputs.version }}" =~ (alpha|beta|rc)[0-9]+$ ]]; then
-            echo NEXT_VERSION=$(echo ${{ inputs.version }} | awk -F- '{print $1}') >> $GITHUB_ENV
-            echo RELEASE_BRANCH=${{ github.ref_name }} >> $GITHUB_ENV
+          echo CURRENT_TAG=$(git \
```
atandon2024, line 49: this git command is somewhat complex. what it does - a) get all tags b) teaches git that those suffixes make a version lower than just the minor/major version c) get the first one in the list (which is the most current tag)

PR #1900, opened 2026-02-25
JAVA-5949 preserve connection pool on backpressure errors when establishing connections

nhachicha:
Original PR https://github.com/mongodb/mongo-java-driver/pull/1854 accidentally closed, and had no outstanding review comments.

- ~~This current PR depends on https://github.com/mongodb/mongo-java-driver/pull/1856:~~
**Update**: This PR no longer depends on https://github.com/mongodb/mongo-java-driver/pull/1856. Spec PR https://github.com/mongodb/specifications/pull/1880 rewrote the `backpressure-network-*-fail.yml` tests to wait on `serverDescriptionChangedEvent` instead of `serverHeartbeatSucceededEvent`.

~~we should merge the latter into `main`, then merge this PR in `backpressure`, and leave the necessary `TODO` comments in https://github.com/mongodb/mongo-java-driver/pull/1918 related to the dependency on https://github.com/mongodb/mongo-java-driver/pull/1856.~~

- then, when there are no outstanding PRs for the `backpressure` branch, rebase `backpressure` on top of `main` to include the changes from the `main` (see https://github.com/mongodb/mongo-java-driver/pull/1917#issuecomment-4102007580 for more context on this process).
- then address the aforementioned `TODO` comments.

Specification changes:

- [DRIVERS-3218 Avoid clearing the connection pool when the server connection rate limiter triggers (#1855)](https://github.com/mongodb/specifications/commit/1cafc53b729df3f1ecb144dba78d51da6ffa6a55)
- [DRIVERS-3218: update SDAM error handling tests to ignore handshake failures (#1860)](https://github.com/mongodb/specifications/commit/c12c50ed0ff3df870883e2c1d51254e933f9850e)
- [DRIVERS-3218 Clarify interaction between TLS errors and backpressure (#1879)](https://github.com/mongodb/specifications/commit/28fdaf9b37851f8d479a510be9f3717338c94608) - the change has not yet been addressed in this PR, because it was introduced after creation of the PR.
- [DRIVERS-3218 clarify label handling (#1892)](https://github.com/mongodb/specifications/commit/bb9dddd8176eddbb9424f9bebedfe8c6bbf28c3a) - the change has not yet been addressed in this PR, because it was introduced after creation of the PR.
- [DRIVERS-3367: Fix racy backpressure-network-* tests (#1880)](https://github.com/mongodb/specifications/commit/7d1d59b0e5b7bbf1a09881f9dfd9e36264a1da4f) - just a unified spec tests (JSON) update; the change has not yet been addressed in this PR, because it was introduced after creation of the PR.
- [DRIVERS-1785 ](https://jira.mongodb.org/browse/DRIVERS-1785) Add test that PoolClearedEvent is emitted before ConnectionCheckedInEvent/ConnectionCheckOutFailedEvent

Note that there was a change in the description of one of the tests that belongs to this PR, but I closed the corresponding JAVA-6095 (see this [comment](https://jira.mongodb.org/browse/JAVA-6095?focusedCommentId=8258829&focusedId=8258829&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-8258829)): [DRIVERS-3394: adjust spec test description (#1894)](https://github.com/mongodb/specifications/commit/5fd020275b621110f0fb9438b590be7bb0a535f2).

JAVA-5949, JAVA-6056, JAVA-5664

### AI usage and effectiveness:
- Model Opus 4.6 then 4.7
- Used as an implementer + paired reviewer
- Spec audits
- Most architectural decisions were grounded in "what do Node and .NET do?" in addition to the reasoning from the spec text.


stIncMale:
Added the following to the description of the PR (a new piece of work to be done within the PR): [DRIVERS-3394: adjust spec test description (#1894)](https://github.com/mongodb/specifications/commit/5fd020275b621110f0fb9438b590be7bb0a535f2)(JAVA-6095).

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,79 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+            for (int i = 0; i < 100; i++) {
+                executor.submit(() -> collection.find(new Document("$where", "function() { sleep(2000); return true; }")).first());
+            }
+
+            // Wait for all operations to complete
+            executor.shutdown();
+            boolean terminated = executor.awaitTermination(20, SECONDS);
+            assertTrue("Executor did not terminate within timeout", terminated);
+
+            // Assert at least 10 ConnectionCheckOutFailedEvents occurred
```
vbabanin: Do we need this comment? The assertion message already explains the intent (e.g., `“Expected at least 10 ConnectionCheckOutFailedEvents, but got …”`).

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,79 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+            assertTrue("Expected at least 10 ConnectionCheckOutFailedEvents, but got " + connectionCheckOutFailedEventCount.get(),
+                    connectionCheckOutFailedEventCount.get() >= 10);
+
+            // Assert 0 PoolClearedEvents occurred
```
vbabanin: The same as in: https://github.com/mongodb/mongo-java-driver/pull/1900/changes#r2908976631

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,79 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+            assertEquals("Expected 0 PoolClearedEvents", 0, poolClearedEventCount.get());
+
+            // Teardown: sleep 1 second and reset rate limiter
+            Thread.sleep(1000);
+            adminDatabase.runCommand(new Document("setParameter", 1)
+                    .append("ingressConnectionEstablishmentRateLimiterEnabled", false));
```
vbabanin: This cleanup is currently conditional on the code above completing successfully. If an assertion or exception happens earlier, this teardown won’t run, which can leak state into subsequent tests.

We should move it this cleanup into `@AfterEach/afterEach` so it runs reliably regardless of how the test exits.

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,79 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+            @Override
+            public void connectionCheckOutFailed(final ConnectionCheckOutFailedEvent event) {
+                connectionCheckOutFailedEventCount.incrementAndGet();
+            }
+
+            @Override
+            public void connectionPoolCleared(final ConnectionPoolClearedEvent event) {
+                poolClearedEventCount.incrementAndGet();
+            }
+        };
```
vbabanin: Instead of introducing a new anonymous listener with counters, we can reuse the existing test listener:
```
TestConnectionPoolListener connectionPoolListener = new TestConnectionPoolListener();
```

It already provides await helpers that double as assertions and helpers to assert that zero PoolClearedEvents happened, e.g.:

`connectionPoolListener.waitForEvent(ConnectionPoolClearedEvent.class, 1, 110, SECONDS);`


This keeps the test more concise and reuses established utilities for clarity/consistency.

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
```diff
@@ -137,9 +138,27 @@ private void handleException(final SdamIssue sdamIssue, final boolean beforeHand
...
                 || (beforeHandshake && (sdamIssue.relatedToNetworkTimeout() || sdamIssue.relatedToAuth()))) {
-            updateDescription(sdamIssue.serverDescription());
-            connectionPool.invalidate(sdamIssue.exception().orElse(null));
-            serverMonitor.cancelCurrentCheck();
+            // Backpressure spec: Don't clear pool or mark server unknown for connection establishment failures
+            // (network errors or timeouts during handshake). Authentication errors after handshake should still
+            // clear the pool as they're not related to overload.
+            // TLS configuration errors (certificate validation, protocol mismatches) should also clear the pool
+            // as they indicate configuration issues, not server overload.
+            if (beforeHandshake && !sdamIssue.relatedToAuth() && !sdamIssue.relatedToTlsConfigurationError()) {
```
vbabanin: Currently we attach `SystemOverloadedError` and `RetryableError` labels in the SDAM error-handling path (effectively only for `DefaultServer`). In load-balanced mode, SDAM isn’t involved: the LB code path invalidates the pool directly (e.g., `connectionPool.invalidate(serviceId, generation)`), so the labeling logic is bypassed.

This means users running the driver in LB mode (behind an NLB) can still hit network errors, TLS handshake failures, timeouts during connection establishment or hello, but won’t get the labels.

However, these labels are a CMAP requirement, not SDAM. The CMAP spec states: `“The pool MUST add the error labels SystemOverloadedError and RetryableError to network errors or network timeouts it encounters during the connection establishment or the hello message.”`

Since this is defined as a pool behavior (topology-agnostic), it seems we should implement the labeling in the connection pool layer so it applies consistently in both default and load-balanced modes.

vbabanin: Currently we don’t distinguish DNS lookup failures `(UnknownHostException`) from other connection-establishment network errors. As a result, a DNS failure goes through the generic path (same as connection reset/timeout) and gets `SystemOverloadedError/RetryableError` labels.

The CMAP spec excludes DNS failures from backpressure labeling: `“For errors that the driver can distinguish as never occurring due to server overload, such as DNS lookup failures […] the driver MUST NOT add backpressure error labels for these error types.”.

**Proposed change:** detect DNS failure by walking the exception cause chain for `UnknownHostException` (it’s wrapped as `MongoSocketException` from `ServerAddressHelper.getSocketAddresses()`), and when present, skipping backpressure label attachment so SDAM follows the normal path (clear the pool and mark the server Unknown).

In that case, we should add coverage to assert that labeling and pool clearing behaviour. If the driver ever changes the wrapper exception type `MongoSocketException` (or stops wrapping UnknownHostException this way) and starts adding labels the test should fail.

vbabanin (changes requested):

driver-core/src/main/com/mongodb/MongoException.java
```diff
@@ -50,6 +50,22 @@ public class MongoException extends RuntimeException {
...
+    public static final String SYSTEM_OVERLOADED_ERROR_LABEL = "SystemOverloadedError";
+
+    /**
+     * An error label indicating that the operation is safely retryable.
+     *
+     * @see #hasErrorLabel(String)
+     * @since 5.7
+     */
+    public static final String RETRYABLE_ERROR_LABEL = "RetryableError";
+
```
stIncMale, line 68: I introduced these constants in https://github.com/mongodb/mongo-java-driver/pull/1918. There is more to this than the change do in the current PR. Let's revert the change in the current PR to resolve the conflict with `backpressure`.

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
```diff
@@ -137,9 +138,27 @@ private void handleException(final SdamIssue sdamIssue, final boolean beforeHand
...
         } else if (sdamIssue.relatedToNetworkNotTimeout()
                 || (beforeHandshake && (sdamIssue.relatedToNetworkTimeout() || sdamIssue.relatedToAuth()))) {
-            updateDescription(sdamIssue.serverDescription());
-            connectionPool.invalidate(sdamIssue.exception().orElse(null));
-            serverMonitor.cancelCurrentCheck();
+            // Backpressure spec: Don't clear pool or mark server unknown for connection establishment failures
+            // (network errors or timeouts during handshake). Authentication errors after handshake should still
+            // clear the pool as they're not related to overload.
+            // TLS configuration errors (certificate validation, protocol mismatches) should also clear the pool
+            // as they indicate configuration issues, not server overload.
```
stIncMale, line 145: We don't usually copy the specification prose into code as comments. I suspect, this comment was left by the AI.

Let's remove this comment, unless there is a good reason to leave it, which was not applicable to the rest of the code in this file that existed before the current PR.

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
```diff
@@ -137,9 +138,27 @@ private void handleException(final SdamIssue sdamIssue, final boolean beforeHand
...
+            if (beforeHandshake && !sdamIssue.relatedToAuth() && !sdamIssue.relatedToTlsConfigurationError()) {
+                // Don't update server description to Unknown
+                // Don't invalidate the connection pool
+                // Apply error labels for backpressure
```
stIncMale, line 149: We don't usually copy the specification prose into code as comments. I suspect, this comment was left by the AI.

Let's remove this comment, unless there is a good reason to leave it, which was not applicable to the rest of the code in this file that existed before the current PR.

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
stIncMale, line 1: [this comment is left on a file, but has nothing to do with the file; this is done so that we could reply to the comment; commenting on a PR does not allow replies - horrendous GitHub functionality]

The current PR seemingly depends on https://github.com/mongodb/mongo-java-driver/pull/1856 (see the description of the current PR). We need to decide what to do with that.

P.S. I originally left my thoughts in the description of this PR, but I don't know if that's what we are going to do.

stIncMale, line 1: [this comment is left on a file, but has nothing to do with the file; this is done so that we could reply to the comment; commenting on a PR does not allow replies - horrendous GitHub functionality]

@nhachicha

- Could you please confirm that all the specification changes listed in the "Specification changes" part of the description of the current PR have been addressed in the PR?
- If they have been addressed, let's update the description correspondingly.

stIncMale, line 1: [this comment is left on a file, but has nothing to do with the file; this is done so that we could reply to the comment; commenting on a PR does not allow replies - horrendous GitHub functionality]

JAVA-5949 has an old [comment](https://jira.mongodb.org/browse/JAVA-5949?focusedCommentId=7971036&focusedId=7971036&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-7971036), which instructs to re-enable some tests that were previously disabled when we updated the `testing/resources/specifications` submodule in `main`. Those tests were not enabled. Let's enable them and make sure they pass.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
nhachicha: This PR no longer depends on https://github.com/mongodb/mongo-java-driver/pull/1856. Spec PR https://github.com/mongodb/specifications/pull/1880 rewrote the `backpressure-network-*-fail.yml` tests to wait on `serverDescriptionChangedEvent` instead of `serverHeartbeatSucceededEvent`.

nhachicha: Description updated. All specs are implemented

nhachicha: We no longer depend on this PR see https://github.com/mongodb/mongo-java-driver/pull/1900#discussion_r3132425567

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
nhachicha: Fixed

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
nhachicha: Fixed

driver-core/src/main/com/mongodb/MongoException.java
nhachicha: Fixed

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,79 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+    public void testConnectionPoolBackpressure() throws InterruptedException {
+        assumeTrue(serverVersionAtLeast(7, 0));
+
+        AtomicInteger connectionCheckOutFailedEventCount = new AtomicInteger(0);
+        AtomicInteger poolClearedEventCount = new AtomicInteger(0);
+
+        ConnectionPoolListener connectionPoolListener = new ConnectionPoolListener() {
+            @Override
+            public void connectionCheckOutFailed(final ConnectionCheckOutFailedEvent event) {
+                connectionCheckOutFailedEventCount.incrementAndGet();
```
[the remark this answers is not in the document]
nhachicha: Rejected: not part of the spec

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,79 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+            executor.shutdown();
+            boolean terminated = executor.awaitTermination(20, SECONDS);
+            assertTrue("Executor did not terminate within timeout", terminated);
+
+            // Assert at least 10 ConnectionCheckOutFailedEvents occurred
+            assertTrue("Expected at least 10 ConnectionCheckOutFailedEvents, but got " + connectionCheckOutFailedEventCount.get(),
+                    connectionCheckOutFailedEventCount.get() >= 10);
+
+            // Assert 0 PoolClearedEvents occurred
+            assertEquals("Expected 0 PoolClearedEvents", 0, poolClearedEventCount.get());
```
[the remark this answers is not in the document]
nhachicha: Fixed

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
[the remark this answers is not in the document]
nhachicha: Fixed

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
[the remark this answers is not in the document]
nhachicha: Outdated

driver-core/src/main/com/mongodb/internal/connection/SdamServerDescriptionManager.java
```diff
@@ -162,6 +168,49 @@ boolean relatedToWriteConcern() {
...
+                    if (message != null) {
+                        String lowerMessage = message.toLowerCase();
+                        // These indicate configuration issues, not network issues
+                        if (lowerMessage.contains("certificate")
+                                || lowerMessage.contains("verify")
+                                || lowerMessage.contains("trust")
+                                || lowerMessage.contains("hostname")
+                                || lowerMessage.contains("protocol")
+                                || lowerMessage.contains("cipher")
+                                || lowerMessage.contains("handshake_failure")) {
```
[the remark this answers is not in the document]
nhachicha: Fixed

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
nhachicha: Fixed

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
nhachicha: Fixed

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
nhachicha: Fixed

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
nhachicha: It's wrapped under a finally block now

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
nhachicha: Fixed

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
nhachicha: Fixed

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
```diff
@@ -137,6 +137,9 @@ private void handleException(final SdamIssue sdamIssue, final boolean beforeHand
             serverMonitor.connect();
         } else if (sdamIssue.relatedToNetworkNotTimeout()
                 || (beforeHandshake && (sdamIssue.relatedToNetworkTimeout() || sdamIssue.relatedToAuth()))) {
+            if (sdamIssue.hasBackpressureLabel()) {
+                return;
+            }
             updateDescription(sdamIssue.serverDescription());
             connectionPool.invalidate(sdamIssue.exception().orElse(null));
             serverMonitor.cancelCurrentCheck();
```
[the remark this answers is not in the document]
nhachicha: Rejected: the spec scopes backpressure to the operational pool, not SDAM monitor.

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,72 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+
+        MongoClientSettings clientSettings = getMongoClientSettingsBuilder()
+                .applyToConnectionPoolSettings(builder -> builder
+                        .maxConnecting(100)
+                        .addConnectionPoolListener(connectionPoolListener))
+                .build();
+
+        try (MongoClient adminClient = MongoClients.create(getMongoClientSettingsBuilder().build());
+             MongoClient client = MongoClients.create(clientSettings)) {
+
```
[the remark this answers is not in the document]
nhachicha: Rejected: deviate from the spec

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,72 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+            try {
+                adminDatabase.runCommand(new Document("setParameter", 1)
+                        .append("ingressConnectionEstablishmentRateLimiterEnabled", true));
+                adminDatabase.runCommand(new Document("setParameter", 1)
+                        .append("ingressConnectionEstablishmentRatePerSec", 20));
+                adminDatabase.runCommand(new Document("setParameter", 1)
+                        .append("ingressConnectionEstablishmentBurstCapacitySecs", 1));
+                adminDatabase.runCommand(new Document("setParameter", 1)
+                        .append("ingressConnectionEstablishmentMaxQueueDepth", 1));
```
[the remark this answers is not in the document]
nhachicha: Rejected: already reported

driver-core/src/main/com/mongodb/internal/connection/BackpressureErrorLabeler.java
```diff
@@ -0,0 +1,102 @@
...
+        while (cause != null) {
+            if (cause instanceof UnknownHostException) {
+                return true;
+            }
+            cause = cause.getCause();
+        }
+        return false;
+    }
+
+    static boolean isTlsConfigurationError(final Throwable t) {
```
vbabanin: `isDnsLookupFailure` and `isTlsConfigurationError` take `Throwable`, but the `applyLabelsIfEligible` has already narrowed the exception to `MongoSocketException`. Since `isDnsLookupFailure` and `isTlsConfigurationError` are not used with non-MongoSocketException inputs on the production path, we could tighten both signatures to `MongoSocketException` to make the precondition explicit.

This would also let us simplify `sdamIssueTlsCheckDelegatesToHelper` to assert TLS classification via `applyLabelsIfEligible` (the actual contract), rather than calling the helper with inputs it never receives.

```suggestion
    private static boolean isTlsConfigurationError(final MongoSocketException t) {
```

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,72 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+                    for (int i = 0; i < 100; i++) {
+                        executor.submit(() ->
+                                collection.find(new Document("$where", "function() { sleep(2000); return true; }")).first());
+                    }
+                    executor.shutdown();
+                    assertTrue("Executor did not terminate within timeout",
+                            executor.awaitTermination(20, SECONDS));
+                } finally {
+                    if (!executor.isTerminated()) {
+                        executor.shutdownNow();
```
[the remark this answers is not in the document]
nhachicha: Rejected: overengineering

driver-sync/src/test/functional/com/mongodb/client/unified/EventMatcher.java
```diff
@@ -208,6 +209,12 @@ public void waitForConnectionPoolEvents(final String client, final BsonDocument
             case "connectionReadyEvent":
                 eventClass = ConnectionReadyEvent.class;
                 break;
+            case "connectionClosedEvent":
+                eventClass = ConnectionClosedEvent.class;
+                break;
+            case "connectionCheckedInEvent":
+                eventClass = ConnectionCheckedInEvent.class;
+                break;
```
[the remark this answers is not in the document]
nhachicha: Rejected: `assertConnectionPoolEventCount` is called after the supported events only (not the newly added ones)
```Java
            case "poolClearedEvent":
            case "poolReadyEvent":
                context.getEventMatcher().assertConnectionPoolEventCount(clientId, event, count,
                        entities.getConnectionPoolListener(clientId).getEvents());
```

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTest.java
```diff
@@ -800,6 +800,8 @@ private OperationResult executeWaitForEvent(final UnifiedTestContext context, fi
             case "poolReadyEvent":
             case "connectionCreatedEvent":
             case "connectionReadyEvent":
+            case "connectionClosedEvent":
+            case "connectionCheckedInEvent":
                 context.getEventMatcher().waitForConnectionPoolEvents(clientId, event, count, entities.getConnectionPoolListener(clientId));
```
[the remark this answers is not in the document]
nhachicha: Rejected: ditto https://github.com/mongodb/mongo-java-driver/pull/1900#discussion_r3153456579

driver-core/src/main/com/mongodb/internal/connection/BackpressureErrorLabeler.java
```diff
@@ -0,0 +1,102 @@
...
+                            || lowerMessage.contains("protocol")
+                            || lowerMessage.contains("cipher")
+                            || lowerMessage.contains("handshake_failure")) {
+                        return true;
+                    }
+                }
+            }
+            cause = cause.getCause();
+        }
+        return false;
```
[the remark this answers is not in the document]
nhachicha: Rejected: outdated (unit test added)

driver-core/src/main/com/mongodb/internal/connection/BackpressureErrorLabeler.java
```diff
@@ -0,0 +1,102 @@
...
```
vbabanin, line 131: I dug into `isTlsConfigurationError` and the TLS exception landscape across JDK versions and providers. Sharing what I found below and what we discussed over the call.

---

### Current TLS config error gaps

The keyword list (`certificate`, `verify`, `trust`, `hostname`, `protocol`, `cipher`, `handshake_failure`) catches 12 of the 25 RFC-defined TLS alert descriptions but misses these config errors.

| Alert | JDK exception message | Labeled? |
|---|---|---|
| `unknown_ca(48)` | `"Received fatal alert: unknown_ca"` | Yes -- **false postive** |
| `access_denied(49)` | `"Received fatal alert: access_denied"` | Yes -- **false positive* *|
| `insufficient_security(71)` | `"Received fatal alert: insufficient_security"` | Yes -- **false positive** |
| `unrecognized_name(112)` | `"Received fatal alert: unrecognized_name"` | Yes -- **false positive** |

BouncyCastle's JSSE provider is currently unhandled: it produces `TlsFatalAlertReceived` (extends `IOException`, not `SSLException`), however, there is no good way to catch those.

### How FIN, RST, and TLS alerts surface across providers and transports

<details>
<summary>Full exception chains (42 test cases)</summary>

**FIN (graceful close during TLS handshake):**

| Transport | Provider | Exception chain |
|---|---|---|
| SocketStream | JDK | `MongoSocketWriteException` -> `SSLHandshakeException("Remote host terminated the handshake")` -> `EOFException("SSL peer shut down incorrectly")` |
| NettyStream | JDK | `MongoSocketWriteException` -> `ClosedChannelException` |
| TlsChannelStream | JDK | `MongoSocketWriteException` -> `ClosedChannelException` |
| SocketStream | BC | `MongoSocketWriteException` -> `TlsFatalAlert("handshake_failure(40)")` |
| NettyStream | BC | `MongoSocketWriteException` -> `ClosedChannelException` |
| TlsChannelStream | BC | `MongoSocketWriteException` -> `ClosedChannelException` |

**RST (abrupt reset during TLS handshake):**

| Transport | Provider | Exception chain |
|---|---|---|
| SocketStream | JDK | `MongoSocketWriteException` -> `SocketException("Connection reset")` |
| NettyStream | JDK | `MongoSocketWriteException` -> `SslClosedEngineException("SSLEngine closed already")` |
| TlsChannelStream | JDK | `MongoSocketWriteException` -> `SocketException("Connection reset")` |
| SocketStream | BC | `MongoSocketWriteException` -> `SocketException("Connection reset")` |
| NettyStream | BC | `MongoSocketWriteException` -> `SslClosedEngineException("SSLEngine closed already")` |
| TlsChannelStream | BC | `MongoSocketWriteException` -> `SocketException("Connection reset")` |

**TLS alert example -- `unknown_ca(48)`:**

| Transport | Provider | Exception chain |
|---|---|---|
| SocketStream | JDK | `MongoSocketWriteException` -> `SSLHandshakeException("Received fatal alert: unknown_ca")` |
| NettyStream | JDK | `MongoSocketWriteException` -> `SSLHandshakeException("Received fatal alert: unknown_ca")` |
| TlsChannelStream | JDK | `MongoSocketWriteException` -> `SSLHandshakeException("Received fatal alert: unknown_ca")` |
| SocketStream | BC | `MongoSocketWriteException` -> `TlsFatalAlertReceived("unknown_ca(48)")` |
| NettyStream | BC | `MongoSocketWriteException` -> `SSLException` -> `TlsFatalAlertReceived("unknown_ca(48)")` |
| TlsChannelStream | BC | `MongoSocketWriteException` -> `SSLException` -> `TlsFatalAlertReceived("unknown_ca(48)")` |

</details>

Key observation: FIN/RST get labeled correctly today because they don't match any exclusion check . A TLS alert from the server means it actively responded, so it wasn't shedding load.
Assumption: A rate limiter drops the TCP connection (FIN/RST) before or during a TLS exchange.

### Two approaches

Both are heuristic-based -- perfect classification isn't possible since exception types and messages are provider implementation details.

**A. Improved blocklist (current approach):** Label by default, exclude what we can identify as not-overload. 
Failure mode: unrecognized config errors stay labeled (pool preserved, unnecessary retries). 

it should be better for unknown providers - if we can't classify an error, pool is preserved, which is the safer default under load.

**B. Allowlist:** Only label recognized I/O patterns (`EOFException`, `SocketException("Connection reset")`, `ClosedChannelException`). 

Failure mode: unrecognized I/O errors are not labeled (pool cleared unnecessarily). This is the pre-backpressure behavior. For example, BC represents FIN as `TlsFatalAlert("handshake_failure(40)")` - the allowlist misses it, pool gets cleared.

### TL;DR / Suggestion:

As we discussed over the call, the blocklist (Approach A) matches the CMAP spec's direction ("the pool MUST add the error labels" as default, with exclusions carved out).

However, we could check for all RFC alert description strings, so we can improve the heuristic.       

<details>
<summary>All 25 handshake-only TLS alert descriptions from OpenJDK's Alert.java (RFC 8446/5246)</summary>

These are the alert codes with `handshakeOnly=true` in OpenJDK. When received from a peer, they produce `SSLHandshakeException("Received fatal alert: <description>")`. All are definitively config/protocol errors, not I/O:

| Code | Description string | Currently caught by keywords? |
|---|---|---|
| 40 | `handshake_failure` | Yes (`handshake_failure`) |
| 41 | `no_certificate` | Yes (`certificate`) |
| 42 | `bad_certificate` | Yes (`certificate`) |
| 43 | `unsupported_certificate` | Yes (`certificate`) |
| 44 | `certificate_revoked` | Yes (`certificate`) |
| 45 | `certificate_expired` | Yes (`certificate`) |
| 46 | `certificate_unknown` | Yes (`certificate`) |
| 47 | `illegal_parameter` | **No** |
| 48 | `unknown_ca` | **No** |
| 49 | `access_denied` | **No** |
| 50 | `decode_error` | **No** |
| 51 | `decrypt_error` | **No** |
| 60 | `export_restriction` | **No** |
| 70 | `protocol_version` | Yes (`protocol`) |
| 71 | `insufficient_security` | **No** |
| 100 | `no_renegotiation` | **No** |
| 109 | `missing_extension` | **No** |
| 110 | `unsupported_extension` | **No** |
| 111 | `certificate_unobtainable` | Yes (`certificate`) |
| 112 | `unrecognized_name` | **No** |
| 113 | `bad_certificate_status_response` | Yes (`certificate`) |
| 114 | `bad_certificate_hash_value` | Yes (`certificate`) |
| 115 | `unknown_psk_identity` | **No** |
| 116 | `certificate_required` | Yes (`certificate`) |
| 120 | `no_application_protocol` | **No** |

</details>

We should keep the existing type checks (`CertificateException`, `CertPathBuilderException`, `CertPathValidatorException`, `SSLPeerUnverifiedException`, `SSLProtocolException`) - they handle local validation failures correctly on both providers (and should on others).  


**P.S** Claude was used to examine JSSE providers and create a table of RFC messages used in OpenJDK's provider.

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
```diff
@@ -439,15 +439,7 @@ public static void applyCustomizations(final TestDef def) {
...
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5949")
-                .file("server-discovery-and-monitoring", "backpressure-network-error-fail-single");
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5949")
-                .file("server-discovery-and-monitoring", "backpressure-network-timeout-error-single");
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5949")
-                .file("server-discovery-and-monitoring", "backpressure-network-error-fail-replicaset");
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5949")
-                .file("server-discovery-and-monitoring", "backpressure-network-timeout-error-replicaset");
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5949")
+        def.skipJira("https://jira.mongodb.org/browse/JAVA-6174")
```
vbabanin, line 443: I suggest marking this as `// TODO-BACKPRESSURE`, since before the release we should re-enable the test together with the fix.

driver-core/src/main/com/mongodb/internal/connection/DefaultSdamServerDescriptionManager.java
```diff
@@ -137,6 +137,9 @@ private void handleException(final SdamIssue sdamIssue, final boolean beforeHand
             serverMonitor.connect();
         } else if (sdamIssue.relatedToNetworkNotTimeout()
                 || (beforeHandshake && (sdamIssue.relatedToNetworkTimeout() || sdamIssue.relatedToAuth()))) {
+            if (sdamIssue.hasBackpressureLabel()) {
```
vbabanin: Let's name it `isSystemOverloadedError` to be precise as we do not have `Backpressure` label?

```suggestion
            if (sdamIssue.hasSystemOverloadedLabel()) {
```

driver-core/src/test/unit/com/mongodb/internal/connection/DefaultServerSpecification.groovy
```diff
@@ -258,6 +258,28 @@ class DefaultServerSpecification extends Specification {
         ]
     }
 
+    def 'DNS lookup failure should not be labeled as backpressure and should invalidate the pool'() {
+        given:
+        def exceptionToThrow = new MongoSocketException('DNS lookup failed', new ServerAddress(),
+                new UnknownHostException('no such host'))
+        BackpressureErrorLabeler.applyLabelsIfEligible(exceptionToThrow)
+        assert !exceptionToThrow.hasErrorLabel(MongoException.SYSTEM_OVERLOADED_ERROR_LABEL)
```
vbabanin: This part seems better suited for the unit tests where the `SUT` is `BackpressureLabeler` (e.g., `BackpressureErrorLabelerTest`). In this test the `SUT` is `server.getConnection()`, so it would be preferable to avoid asserting labeling internals here.

Could we instead assert the “label not applied” outcome at the behavior level in the` then:` section (e.g., on `def e = thrown(MongoException))` and keep the unit tests in `BackpressureLabeler` responsible for the labeling implementation details?

driver-core/src/test/unit/com/mongodb/internal/connection/BackpressureErrorLabelerTest.java
```diff
@@ -0,0 +1,156 @@
...
+import java.io.IOException;
+import java.net.UnknownHostException;
+import java.security.cert.CertificateException;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class BackpressureErrorLabelerTest {
+
+    private static final ServerAddress ADDRESS = new ServerAddress();
```
vbabanin, line 50: I think it’s worth extending coverage here to include broaden TLS `failure modes`/`exceptions`:

- Local TLS configuration failures (client-side), and
- Handshake failures caused by server alerts, which surface as `SSLHandshakeException` with OpenJDK’s JSSE provider.

This also looks like a good fit for a parameterized test to cover multiple exception types/cases without duplicating test logic. 

Suggested change to be more specific: https://github.com/vbabanin/mongo-java-driver/commit/d78261e12301970d45a5f3ddab0420ff4f591c36

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -268,6 +272,72 @@ public void shouldEmitHeartbeatStartedBeforeSocketIsConnected() {
...
+    public void testConnectionPoolBackpressure() throws InterruptedException {
+        assumeTrue(serverVersionAtLeast(7, 0));
+
+        TestConnectionPoolListener connectionPoolListener = new TestConnectionPoolListener();
+
+        MongoClientSettings clientSettings = getMongoClientSettingsBuilder()
+                .applyToConnectionPoolSettings(builder -> builder
+                        .maxConnecting(100)
+                        .addConnectionPoolListener(connectionPoolListener))
+                .build();
```
vbabanin, line 289: The spec says: 
> If running against a sharded cluster, this test MUST be run against only a single host

I don’t think we’re enforcing that requirement here, should we adjust the setup to ensure the test targets a single host when sharded?

nhachicha: This is enforced by Evergreen:  `getMongoClientSettingsBuilder` uses `ClusterFixture.getConnectionString()`, which is set to `MONGODB_URI` (single mongos) in https://github.com/mongodb/mongo-java-driver/blob/main/.evergreen/run-tests.sh#L85 and https://github.com/mongodb/mongo-java-driver/blob/main/.evergreen/run-tests.sh#L132

driver-core/src/test/unit/com/mongodb/internal/connection/BackpressureErrorLabelerTest.java
nhachicha: Fixed

driver-core/src/test/unit/com/mongodb/internal/connection/DefaultServerSpecification.groovy
nhachicha: Fixed

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
nhachicha: Fixed

driver-core/src/main/com/mongodb/internal/connection/BackpressureErrorLabeler.java
nhachicha: Refactored for option A

driver-core/src/main/com/mongodb/internal/connection/BackpressureErrorLabeler.java
nhachicha: Fixed

driver-core/src/main/com/mongodb/internal/connection/BackpressureErrorLabeler.java
```diff
@@ -0,0 +1,109 @@
...
+                String message = cause.getMessage();
+                if (message != null) {
+                    String lowerMessage = message.toLowerCase(Locale.ROOT);
+                    if (lowerMessage.contains("certificate")
+                            || lowerMessage.contains("verify")
+                            || lowerMessage.contains("trust")
+                            || lowerMessage.contains("hostname")
+                            || lowerMessage.contains("protocol")
+                            || lowerMessage.contains("cipher")
+                            || lowerMessage.contains("handshake_failure")
```
vbabanin: This RFC message is already covered by `"received fatal alert"`
```suggestion
```

driver-core/src/test/unit/com/mongodb/internal/connection/DefaultServerSpecification.groovy
vbabanin: I’m not sure this should be DNS-specific at the `DefaultServer` level. 

Other unlabeled connection-establishment failures (e.g., TLS configuration errors) follow the same path and also shouldn’t receive backpressure labels.

If the intent of this test is “unlabeled errors still invalidate the pool”, we have two options:

1. Expand this test to cover other unlabeled exceptions (TLS config, etc.), or
2. Make it a generic behavior test and rename accordingly (e.g., from `“DNS lookup failure …”` to `“should invalidate the pool when the exception without system overloaded label”`). It could be also parametrized on different exception types.

I’d prefer (2) to keep the test focused on `DefaultServer` behavior rather than a specific error type (as DNS in this case).

Also, I suggest adding the complementary test for the opposite branch: ` should not invalidate the pool when exception with SystemOverloadedError`. This would complete coverage of both paths.

driver-core/src/test/unit/com/mongodb/internal/connection/DefaultServerSpecification.groovy
```diff
@@ -258,6 +258,28 @@ class DefaultServerSpecification extends Specification {
         ]
     }
 
+    def 'DNS lookup failure should not be labeled as backpressure and should invalidate the pool'() {
+        given:
+        def exceptionToThrow = new MongoSocketException('DNS lookup failed', new ServerAddress(),
+                new UnknownHostException('no such host'))
+        BackpressureErrorLabeler.applyLabelsIfEligible(exceptionToThrow)
```
vbabanin: Could we remove this implementation-detail? It looks redundant in this test since a DNS exception is a no-op (doesn't apply any label), and the test would pass identically without this line. The SUT is DefaultServer.getConnection() - we're testing that an unlabeled pool DNS error stays unlabeled. If the labeler logic changes, this test would break for the different reason.

Related: #1900 (comment)
```suggestion
```

nhachicha: Fixed

driver-core/src/test/unit/com/mongodb/internal/connection/DefaultServerSpecification.groovy
nhachicha: Fixed (option 2 + additional test)

driver-core/src/test/unit/com/mongodb/internal/connection/DefaultServerSpecification.groovy
```diff
@@ -258,6 +259,56 @@ class DefaultServerSpecification extends Specification {
         ]
     }
 
+    def 'should invalidate the pool when the exception without system overloaded label'() {
+        given:
+        assert !exceptionToThrow.hasErrorLabel(MongoException.SYSTEM_OVERLOADED_ERROR_LABEL)
```
vbabanin: Do we need this assertion in given?
```suggestion
```

driver-core/src/main/com/mongodb/internal/connection/BackpressureErrorLabeler.java
```diff
@@ -0,0 +1,108 @@
...
+                return true;
+            }
+            if (cause instanceof SSLHandshakeException) {
+                String message = cause.getMessage();
+                if (message != null) {
+                    String lowerMessage = message.toLowerCase(Locale.ROOT);
+                    if (lowerMessage.contains("certificate")
+                            || lowerMessage.contains("verify")
+                            || lowerMessage.contains("trust")
+                            || lowerMessage.contains("hostname")
```
vbabanin: What scenario produces an `SSLHandshakeException` with "trust", "hostname", or "pkix" in the message which is not covered by ? I couldn’t find these strings in the [OpenJDK JSSE provider sources](https://github.com/search?q=repo%3Aopenjdk%2Fjdk+%22new+SslHandshakeException%22&type=code&p=2) - are they coming from a different provider or JDK version?

nhachicha: I saw some SO messages with these: [ex1](https://stackoverflow.com/questions/21183043/sslhandshakeexception-trust-anchor-for-certification-path-not-found-android-http) & [ex2](https://stackoverflow.com/questions/77173976/sslhandshakeexception-pkix-after-migrate-glassfish-2-to-5) and [ex3](https://stackoverflow.com/questions/37234200/sslhandshakeexception-pkix-path-building-failed) but you can argue these are coming from the cause `getMessage`, which makes these checks empirical. I'm leaning towards keeping only the `received fatal alert`

driver-core/src/main/com/mongodb/internal/connection/BackpressureErrorLabeler.java
```diff
@@ -0,0 +1,174 @@
...
+                    String lowerMessage = message.toLowerCase(Locale.ROOT);
+                    if (lowerMessage.contains("verify")
+                            || lowerMessage.contains("protocol")
+                            || lowerMessage.contains("cipher")
+                            || lowerMessage.contains("received fatal alert")) {
+                        return true;
+                    }
+                }
+            }
+            if (isBouncyCastleTlsError(cause)) {
```
vbabanin, line 125: Should we add a couple of unit tests for this path as well? It looks like this isn’t currently covered by automated tests.

Since `ByteBuddy` is already on the classpath, it could be possible to generate a class with the expected FQCN to exercise the logic, e.g.:
```
  try (DynamicType.Unloaded<Exception> unloaded = new ByteBuddy()
          .subclass(Exception.class)
          .name("org.bouncycastle.tls.TlsFatalAlert")
          .make()) { 
      unloaded.load(getClass().getClassLoader())
}
```

One caveat: we may need to check error type via FQCN string rather than Class equality.

nhachicha: Fixed: good idea using the ByteBuddy with FQCN 👏

nhachicha: Also added reference to https://jira.mongodb.org/browse/JAVA-6194 for the socks5 TODO

driver-core/src/test/unit/com/mongodb/internal/connection/BackpressureErrorLabelerTest.java
```diff
@@ -0,0 +1,260 @@
...
+                .subclass(Exception.class)
+                .name("org.bouncycastle.tls.TlsFatalAlert")
+                .make()
+                .load(BackpressureErrorLabelerTest.class.getClassLoader())
+                .getLoaded();
+        Class<?> bcSubclass = new ByteBuddy()
+                .subclass(bcParent)
+                .name("com.example.CustomBcSubclass")
+                .make()
+                .load(bcParent.getClassLoader())
```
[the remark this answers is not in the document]
nhachicha: Rejected: false positive — `WRAPPER` is the default strategy for the single-arg `load(ClassLoader)`, and it creates a *new child ClassLoader per call*, so re-defining
  the same FQCN doesn't collide. The parent (the test classloader) is never written to.                                                                                                                            
                                                                                                                                                                                                                   
  Evidence from `byte-buddy-1.14.12`:                                                                                                                                                                              
                                                                                                                                                                                                                   
  `DynamicType.java` (single-arg `load`) — falls through to `WRAPPER` for any non-`InjectionClassLoader`:                                                                                                          
                                                                                                                                                                                                                   
  ```java                                                                                                                                                                                                          
  public DynamicType.Loaded<T> load(@MaybeNull ClassLoader classLoader) {                                                                                                                                          
      if (classLoader instanceof InjectionClassLoader && !((InjectionClassLoader) classLoader).isSealed()) {                                                                                                       
          return load((InjectionClassLoader) classLoader, InjectionClassLoader.Strategy.INSTANCE);                                                                                                                 
      } else {                                                                                                                                                                                                     
          return load(classLoader, ClassLoadingStrategy.Default.WRAPPER);                                                    
      }                                                                                                                                                                                                            
  }                                                                                                                          
  ```                                                                                                                                                                                                              
                                                                                                                                                                                                                   
  `ClassLoadingStrategy.java` — `WRAPPER` Javadoc:                                                                                                                                                                 
                                                                                                                                                                                                                   
  > This strategy creates a new `ByteArrayClassLoader` with the given class loader as its **parent**. […] this strategy will lead to the unloading of these classes once this class loader, its classes or any
  instances of these classes become unreachable.                                                                                                                                                                   
                                                                                                                             
  `DynamicType.java` (interface Javadoc) — explicitly designs for repeated invocation:                                                                                                                             
                                                                          
  > **Note**: A new class is attempted to be loaded each time this method is invoked, even if a compatible class was created previously. Consider using a `TypeCache`.

driver-core/src/test/unit/com/mongodb/internal/connection/BackpressureErrorLabelerTest.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    private static Throwable newBouncyCastleStub(final String fqcn, final String message) throws Exception {
+        Class<?> cls = new ByteBuddy()
+                .subclass(Exception.class)
+                .name(fqcn)
+                .make()
+                .load(BackpressureErrorLabelerTest.class.getClassLoader())
+                .getLoaded();
+        return (Throwable) cls.getConstructor(String.class).newInstance(message);
```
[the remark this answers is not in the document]
nhachicha: Rejected: ditto https://github.com/mongodb/mongo-java-driver/pull/1900#discussion_r3202087916

driver-core/src/main/com/mongodb/internal/connection/BackpressureErrorLabeler.java
```diff
@@ -0,0 +1,165 @@
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Walks the class hierarchy comparing fully qualified names so a subclass of a known BC type
+     * still matches without requiring the BC classes to be loadable at static-init time.
+     */
```
vbabanin, line 154: Nit: "without requiring the BC classes to be loadable at static-init time" - this property only matters for testability (ByteBuddy stubs), not for production. So without the context about test constant, it might be unclear how `at static-init time` affects the implementation.

```suggestion
    /**
     * Walks the class hierarchy comparing fully qualified names so a subclass of a known BC type
     * still matches.
     */
```

vbabanin (dismissed):

PR #1911, opened 2026-03-13
Add onlyConnectOriginalUrl connection string option to skip replica set topology discovery

TheWind111:
Resolves JAVA-6128

## Motivation

When connecting to a replica set, the Java driver automatically discovers and manages cluster topology by adding and removing hosts based on information reported by primary members (via `ensureServers()`). This behavior is not always desirable.

Common scenarios where users need to disable auto-discovery:
- Certain replica set nodes are known to be unreachable from the client network
- The client should only connect to the explicitly specified seed list, without expanding to other members

## Changes

Add a new boolean connection string option `onlyConnectOriginalUrl` (default: `false`).

When set to `true`, `AbstractMultiServerCluster` will:
- Still perform necessary validity checks (server type check, replica set ghost check, set name matching)
- Skip the `ensureServers()` call in `handleReplicaSetMemberChanged`, so no new hosts are added and no existing hosts are removed based on topology discovery — similar to the behavior of `handleShardRouterChanged`

**Example usage:**
mongodb://host1:27017,host2:27018/?replicaSet=myRS&onlyConnectOriginalUrl=true

## Files Changed

- `ConnectionString.java` — parse and expose the new option
- `ClusterSettings.java` — store the flag via builder/getter and apply from connection string
- `AbstractMultiServerCluster.java` — apply guard logic in `handleReplicaSetMemberChanged`

## Testing

Added unit tests covering:
- Primary reports additional members → no new hosts added
- Primary reports reduced member list → no existing hosts removed
- Wrong server type → server still removed (safety checks take priority)
- Wrong replica set name → server still rejected
- `onlyConnectOriginalUrl=false` → original behavior unchanged

### DIFF
--- a/driver-core/src/main/com/mongodb/ConnectionString.java
+++ b/driver-core/src/main/com/mongodb/ConnectionString.java
@@ -304,6 +304,7 @@ public class ConnectionString {
     private String srvServiceName;
     private Boolean directConnection;
     private Boolean loadBalanced;
+    private Boolean onlyConnectOriginalUrl;
     private ReadPreference readPreference;
     private WriteConcern writeConcern;
     private Boolean retryWrites;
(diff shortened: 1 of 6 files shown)


jyemin:
Closing, as JAVA-6128 closed with no response.

PR #1913, opened 2026-03-17
Optimize RawBsonDocument encode and decode 

vbabanin:
- Add BsonWriter.pipe(byte[], int, int) in BsonBinaryWriter to write raw BSON bytes directly to the output, avoiding intermediate object allocation on the encode path
- Add BsonInput.pipe(BsonOutput, int) to remove the temporary byte[] copy in BsonBinaryWriter.pipeDocument() on both encode and decode paths
- Add public getBackingArray(), getByteOffset(), getByteLength() on RawBsonDocument to expose the backing byte array

Performance analyzer: [link](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=6a0e87678f137b6a464ceb77&selected_tab=data-table&percent_filter=0%7C%7C100&z_filter=0%7C%7C10&filter_type=Default)

JAVA-6133


bson/src/main/org/bson/BsonBinaryWriter.java
```diff
@@ -355,9 +366,7 @@ private void pipeDocument(final BsonReader reader, final List<BsonElement> extra
             }
             int pipedDocumentStartPosition = bsonOutput.getPosition();
             bsonOutput.writeInt32(size);
-            byte[] bytes = new byte[size - 4];
-            bsonInput.readBytes(bytes);
-            bsonOutput.writeBytes(bytes);
```
vbabanin, line 360: This avoids an extra temporary byte[] allocation by reading directly into the target buffer, reducing both allocation pressure and byte-copy/processing overhead.

bson/src/main/org/bson/BsonBinaryWriter.java
```diff
@@ -334,6 +334,17 @@ public void pipe(final BsonReader reader) {
         pipeDocument(reader, null);
     }
 
+    @Override
+    public void pipe(final byte[] bytes, final int offset, final int length) {
```
vbabanin: Instead of routing through a `BsonReader` (which wraps a BsonInput but doesn’t expose the underlying array unless copied), we write the bytes directly to the output.

bson/src/main/org/bson/io/BsonInput.java
```diff
@@ -127,6 +127,15 @@ public interface BsonInput extends Closeable {
...
     boolean hasRemaining();
 
+    /**
+     * Pipes the specified number of bytes from {@linkplain BsonInput this} input to the given {@linkplain BsonOutput output}.
+     *
+     * @param output the output to pipe to
+     * @param numBytes the number of bytes to pipe
+     * @since 5.8
+     */
+    void pipe(BsonOutput output, int numBytes);
```
rozza: Should this have a default implementation, inorder not to break binary compat?

vbabanin: Good catch! I have added the default implementation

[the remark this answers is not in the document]
vbabanin: Done.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -275,6 +275,24 @@ public boolean hasRemaining() {
...
+            int position = buffer.position();
+            int arrayOffset = buffer.arrayOffset();
+            output.writeBytes(buffer.array(), arrayOffset + position, numBytes);
+            buffer.position(position + numBytes);
+        } else {
+            // Fallback: use temporary buffer for non-array-backed buffers
+            byte[] temp = new byte[numBytes];
+            buffer.get(temp);
+            output.writeBytes(temp);
+        }
```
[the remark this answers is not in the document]
vbabanin: The old code in `pipeDocument`:

```
  byte[] bytes = new byte[size - 4];
  bsonInput.readBytes(bytes);
  bsonOutput.writeBytes(bytes);
```

Same allocation, same size. This PR doesn't make it worse - it just moves that allocation into `pipe()` as a fallback. Copilot doesn't seem to have the context of the pre-existing behavior or suggest over-optimization.

bson/src/main/org/bson/BsonBinaryWriter.java
```diff
@@ -334,6 +334,26 @@ public void pipe(final BsonReader reader) {
...
+     */
+    public void pipe(final byte[] bytes, final int offset, final int length) {
+        checkMinDocumentSize(length);
+        if (getState() == State.VALUE) {
+            bsonOutput.writeByte(BsonType.DOCUMENT.getValue());
+            writeCurrentName();
+        }
+        int pipedDocumentStartPosition = bsonOutput.getPosition();
+        bsonOutput.writeBytes(bytes, offset, length);
+        completePipeDocument(pipedDocumentStartPosition);
```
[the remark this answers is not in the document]
vbabanin: BsonInput is consumed by BsonBinaryReader, which validates BSON as it reads - size checks, type markers, structure. That's where the validation belongs, at the reading/parsing layer, not at the piping layer which is a fast-path.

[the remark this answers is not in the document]
vbabanin: `bsonOutput.writeBytes(bytes, offset, length)` delegates to `BasicOutputBuffer`, which already validates positions and bounds via `checkPosition`.  This is similar to BsonInput#readBytes.

bson/src/main/org/bson/RawBsonDocument.java
```diff
@@ -144,6 +144,36 @@ public ByteBuf getByteBuffer() {
...
 
+    /**
+     * Returns the byte array backing this document. Changes to the returned array will be reflected in this document.
+     *
+     * @return the backing byte array
+     * @since 5.8
+     */
+    public byte[] getBackingArray() {
+        return bytes;
+    }
```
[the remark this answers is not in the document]
vbabanin: Done.

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -275,6 +275,24 @@ public boolean hasRemaining() {
         return buffer.hasRemaining();
     }
 
+    @Override
+    public void pipe(final BsonOutput output, final int numBytes) {
+        ensureOpen();
```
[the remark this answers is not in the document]
vbabanin: writeBytes(byte[], int, int) on BsonOutput doesn't guard against negative length either. None of the I/O methods in this stack do - readBytes, writeBytes, ensureAvailable all rely on the underlying buffer to fail on invalid values.  Adding a guard here but not in the rest of the I/O layer would be inconsistent.

bson/src/test/unit/org/bson/io/BsonInputTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class BsonInputTest {
+
+    @Test
+    void defaultPipeShouldCopyBytesFromInputToOutput() {
+        // given
+        byte[] inputBytes = {0x4a, 0x61, 0x76, 0x61, 0x21};
```
strogiyotec: can we do 
```
"Java!".getBytes(StandardCharsets.UTF_8);
```
instead ? Same thing but for me personally it's really hard to reason about raw bytes , what do you think ?

bson/src/test/unit/org/bson/BsonBinaryWriterTest.java
```diff
@@ -802,7 +802,34 @@ public void testPipeOfDocumentWithInvalidSize() {
...
+        try (BsonBinaryWriter destWriter = new BsonBinaryWriter(destBuffer)) {
+            destWriter.pipe(documentBytes, 0, documentBytes.length);
+        }
+
+        assertArrayEquals(documentBytes, destBuffer.toByteArray());
+    }
+
+    @Test
+    public void testPipeOfRawBytesWithInvalidSize() {
+        byte[] bytes = {4, 0, 0, 0};  // minimum document size is 5
```
strogiyotec, line 827: I saw the validation for size of 5, just curious why is it 5 ?

bson/src/test/unit/org/bson/RawBsonDocumentSpecification.groovy
```diff
@@ -113,6 +113,23 @@ class RawBsonDocumentSpecification extends Specification {
         rawDocument << createRawDocumentVariants()
     }
 
+
+    def 'getBackingArray, getByteOffset and getByteLength should expose the document range'() {
```
strogiyotec: I might be mistaken but I saw a lot of PRs that also remove groovy spec class, can we move this test case to java test instead ?

bson/src/test/unit/org/bson/io/BsonInputTest.java
```diff
@@ -0,0 +1,122 @@
...
+            assertEquals(inputBytes.length, bsonInput.getPosition());
+            assertEquals(inputBytes.length, output.getPosition());
+            assertArrayEquals(inputBytes, output.toByteArray());
+        }
+    }
+
+    @Test
+    void defaultPipeShouldCopyPartialBytesFromInputToOutput() {
+        // given
+        byte[] inputBytes = {0x4a, 0x61, 0x76, 0x61, 0x21};
```
strogiyotec: https://github.com/mongodb/mongo-java-driver/pull/1913/changes#r3289947388

strogiyotec (changes requested):

bson/src/main/org/bson/codecs/RawBsonDocumentCodec.java
```diff
@@ -40,8 +40,17 @@ public RawBsonDocumentCodec() {
...
-        try (BsonBinaryReader reader = new BsonBinaryReader(new ByteBufferBsonInput(value.getByteBuffer()))) {
-            writer.pipe(reader);
+        if (writer instanceof BsonBinaryWriter) {
+            // Fast path. The pipe method should ideally exist on BsonWriter, but adding it as
+            // abstract would be a breaking change, and adding it as a default method would force
+            // BsonWriter to depend on BsonBinaryReader/ByteBufferBsonInput, violating the
+            // interface's abstraction.
+            // TODO JAVA-6211 move pipe(byte[], int, int) to BsonWriter to remove this instanceof.
+            ((BsonBinaryWriter) writer).pipe(value.getBackingArray(), value.getByteOffset(), value.getByteLength());
+        } else {
```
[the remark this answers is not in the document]
vbabanin: This is incorrect. 

The comment literally says that that `BsonWriter` does not have `pipe(byte[], int, int)` and we should consider adding it in `// TODO JAVA-6211`.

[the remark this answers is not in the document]
rozza: I think its just picked up that the description says its been added. So is a nudge to update the description.

bson/src/test/unit/org/bson/BsonBinaryWriterTest.java
vbabanin: A BSON document is minimum 5 bytes - 4 bytes for the int32 size header (which includes itself) + 1 byte for the null terminator (0x00) that marks the end of the document. 

For example: an empty document  is `\x05\x00\x00\x00\x00` - 5 bytes

bson/src/test/unit/org/bson/io/BsonInputTest.java
vbabanin: Yes, it should be more readable. I have added.

bson/src/test/unit/org/bson/RawBsonDocumentSpecification.groovy
vbabanin: Yes, that matches what we agreed on. I moved this case into the new Java test suite.

bson/src/test/unit/org/bson/io/BsonInputTest.java
vbabanin: Same as here: https://github.com/mongodb/mongo-java-driver/pull/1913#discussion_r3308793915

bson/src/main/org/bson/io/ByteBufferBsonInput.java
```diff
@@ -275,6 +275,24 @@ public boolean hasRemaining() {
...
+        if (buffer.isBackedByArray()) {
+            int position = buffer.position();
+            int arrayOffset = buffer.arrayOffset();
+            output.writeBytes(buffer.array(), arrayOffset + position, numBytes);
+            buffer.position(position + numBytes);
+        } else {
+            // Fallback: use temporary buffer for non-array-backed buffers
+            byte[] temp = new byte[numBytes];
+            buffer.get(temp);
+            output.writeBytes(temp);
```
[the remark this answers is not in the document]
vbabanin: This fallback behavior mirrors BsonBinaryWriter#pipeDocument. We don’t currently restore the underlying ByteBuffer position on exceptions in similar code paths, so keeping the same behavior here maintains consistency across the Bson APIs.

```
           byte[] bytes = new byte[size - 4];
            bsonInput.readBytes(bytes);
            bsonOutput.writeBytes(bytes);
```

bson/src/main/org/bson/BsonBinaryWriter.java
```diff
@@ -334,6 +334,26 @@ public void pipe(final BsonReader reader) {
         pipeDocument(reader, null);
     }
 
+    /**
+     * Pipes a raw BSON document from the given byte array to this writer, writing the bytes directly to the
+     * output without intermediate object allocation.
+     *
+     * @param bytes the byte array containing the BSON document
+     * @param offset the offset into the byte array
+     * @param length the length of the BSON document
```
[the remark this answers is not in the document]
vbabanin: `BSON` by definition always has the size header and null terminator.  However, i have added "encoded" prefix to be more precise.(https://github.com/mongodb/mongo-java-driver/pull/1913/commits/54292e812923bdf72821cd88524773d3dd47f883)

bson/src/main/org/bson/RawBsonDocument.java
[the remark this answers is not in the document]
vbabanin: Good point. Added.

(https://github.com/mongodb/mongo-java-driver/commit/54292e812923bdf72821cd88524773d3dd47f883)

bson/src/main/org/bson/RawBsonDocument.java
```diff
@@ -144,6 +144,40 @@ public ByteBuf getByteBuffer() {
         return new ByteBufNIO(buffer);
     }
 
+    /**
+     * Returns the byte array backing this document. The returned array may be larger than the BSON document itself;
+     * only the range from {@link #getByteOffset()} to {@code getByteOffset() + }{@link #getByteLength()} contains
+     * valid document bytes. Changes to the returned array will be reflected in this document.
```
rozza, line 150: RawBsonDocument is described as "An immutable BSON document that is represented using only the raw bytes". 

This breaks the immutability, so I think we should remove that from the class javadoc.

bson/src/main/org/bson/RawBsonDocument.java
rozza:
```suggestion
 * A BSON document that is represented using only the raw bytes.
```

rozza (dismissed):
Looking good - only one suggestion

rozza (approved):
LGTM!

rozza:
@vbabanin this seems to be have been stuck - so opened #1988 with the rebased changes. Will merge once evergreen is happy.

rozza:
Closed and merged via #1988

PR #1921, opened 2026-03-23
Added AGENTS.md instructions and SKILLS.md

rozza:
Add AI agent instructions using the AGENTS.md open standard
Adds structured AI agent documentation across the MongoDB Java Driver
using the AGENTS.md convention — an open standard for providing
codebase context to AI coding agents.

- Root AGENTS.md with build commands, style rules, testing policy, API
  design guidelines, safety-critical code boundaries, and dependency rules
- Per-module AGENTS.md files (bson, driver-core, driver-sync, etc.) with
  module-specific packages, dependencies, and patterns
- CLAUDE.md files that reference AGENTS.md (for Claude Code compatibility)
- `.agents/references/` for passive knowledge (contextually loaded):
  project-guide, style-reference, api-design, testing-guide, spec-tests
- `.agents/skills/` for procedural workflows:
  evergreen (CI validation), sync-agents-docs (documentation sync checklist)
- `scripts/symlink-claude-md.sh` for maintaining CLAUDE.md symlinks

- References over skills for static knowledge: references are loaded
  contextually based on description relevance rather than requiring
  explicit invocation. This avoids bloating every conversation with all
  documentation while still surfacing it when needed.
- Flat file layout for references (.agents/references/api-design.md)
  rather than directory-per-reference (.agents/references/api-design/REFERENCE.md):
  simpler structure; supplementary files use name-prefixing
  (e.g. testing-guide-examples.md) instead of subdirectories.
- Markdown links (See [name](path)) rather than @ syntax for referencing
  from AGENTS.md: @ auto-includes file contents on every load, which
  defeats contextual loading. Links let agents follow them on demand.
- Skills retained only for procedural workflows that have tool constraints
  (allowed-tools) or explicit trigger conditions (disable-model-invocation),
  not for passive knowledge lookup.

[JAVA-6143](https://jira.mongodb.org/browse/JAVA-6143)


.agents/references/project-guide.md
```diff
@@ -0,0 +1,63 @@
...
+name: project-guide
+description: Project structure, dependency graph, and guide for finding the right project to work in. Use when starting a task and unsure which project owns the relevant code, or when you need to understand cross-project dependencies.
+---
+# Project Guide
+
+## Project Structure
+
+| Project | Purpose |
+| --- | --- |
+| `bson` | BSON library (core serialization) |
```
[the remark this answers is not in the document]
rozza: hallcination - theres only a single pipe.

ajcvickers:
We had a bit of discussion about this in the AI/ML-sync meeting yesterday. My approach has been to use very minimal agents files (for Junie, Augie, and Claude) because I don't think it is clear what is:

* Useful
* Benign
* Harmful

I think we've all seen harmful changes--for example, telling it about tests and then having the agent get caught up attempting to run them all the time.

Things that are not harmful but also not very useful (the benign category) can end up being accidentally harmful by taking up context space and thereby forcing something important out.

We can discuss these things on a case-by-case basis, but I think that means starting small and understanding the impacts of the changes made.

Another approach is for each platform team to do something different and then we all compare notes later.

.agents/skills/api-design/SKILL.md
katcharov, line 1: Could you run Claude to review this PR, asking something like "Confirm that this PR follows best practices for creating CLAUDE.md files. Evaluate it personally, confirming that various items are generally necessary. Also evaluate it based on official and authoritative sources (official Claude documentation, online posts that are corroborated AND experimentally verified, and posts from recognized experts; but not: blog posts, social media speculations, and other uncorroborated sources. Is anything unnecessary, and easily discoverable? Is anything crucial missing?"

Independently, have it doublecheck correctness until it stops finding issues.

Please also include info about how these files were generated, and AI-reviewed, including results of the above. (This is just the typical AI usage, effectiveness comment in the description).

.agents/skills/api-design/SKILL.md
```diff
@@ -0,0 +1,45 @@
...
+  Safe to use, but not to subclass.
+- `@Deprecated` — Supported until next major release but should be migrated away from.
+
+## Design Principles
+
+- **Deep modules:** Prefer simple interfaces with powerful implementations over shallow wrappers.
+- **Information hiding:** Bury complexity behind simple interfaces.
+- **Pull complexity downward:** Make the implementer work harder so callers work less.
+- **General-purpose over special-case:** Fewer flexible methods over many specialized ones.
+- **Define errors out of existence:** Design APIs so errors cannot happen rather than detecting and handling them.
```
katcharov, line 25: I am not sure how to interpret this or check if a model has actually applied these principles. For example, the first one, "deep modules", seems to be a matter of intuition. In practice, I'm not sure how we would choose one approach vs another, or what would make a given approach have a "powerful implementation".

.agents/skills/api-design/SKILL.md
```diff
@@ -0,0 +1,45 @@
...
+
+- `@Alpha` — Early development, may be removed.
+  Not for production use.
+- `@Beta` — Subject to change or removal.
+  Libraries should not depend on these.
+- `@Evolving` — May add abstract methods in future releases.
+  Safe to use, but implementing/extending bears upgrade risk.
+- `@Sealed` — Must not be extended or implemented by consumers.
+  Safe to use, but not to subclass.
+- `@Deprecated` — Supported until next major release but should be migrated away from.
```
katcharov, line 17: I don't think a model should ever be applying some of these. I also don't know what "Libraries should not depend on these" means? Our users can depend on Beta, but we can change it.

.agents/skills/api-design/SKILL.md
```diff
@@ -0,0 +1,45 @@
...
+## Key Patterns
+
+- Static factory methods: `Filters.eq()`, `Updates.set()`, `Aggregates.match()`
+- Fluent builders: `MongoClientSettings.builder()` is the primary entry point
+- Abstract core with pluggable transports
+
+## Public API Rules
+
+- Breaking changes require a major version bump - ALWAYS warn if breaking binary compatibility
+- All `com.mongodb.internal.*` / `org.bson.internal.*` is private API — never expose in public APIs
```
katcharov: I don't understand how a model would expose these as public APIs?

.agents/skills/api-design/SKILL.md
```diff
@@ -0,0 +1,45 @@
+---
+name: api-design
```
katcharov, line 2: I ran only this skill through Claude, and I got this:

```
What's already known to models:
  - The design principles section is essentially a summary of Ousterhout's A
  Philosophy of Software Design. Any capable model already knows these
  concepts and would apply them naturally. This section adds little value.
  - "Search before implementing" is generic good practice, not
  project-specific guidance.

  What's genuinely useful (project-specific):
  - The stability annotations (@Alpha, @Beta, @Evolving, @Sealed) — these are
   driver-specific and not inferable without reading the source.
  - The com.mongodb.internal.* / org.bson.internal.* rule.
  - The concrete pattern examples (Filters.eq(),
  MongoClientSettings.builder()).
  - The package-info.java requirement.

  What's missing that would actually help:
  - How to choose which annotation for new API — when should something be
  @Alpha vs @Beta vs @Evolving? What's the promotion path?
  - Sync/async API mirroring — does every sync method need an async
  counterpart? How are they kept in sync?
  - Nullability conventions — does the driver use @Nullable/@NonNull? What's
  the stance on null parameters?
  - The module structure — what goes in driver-core vs driver-sync vs
  driver-reactive-streams? Where should new API land?
  - Default methods on interfaces — the strategy for evolving @Evolving
  interfaces without breaking implementors.
  - Concrete examples of "wrong" API decisions — models learn more from
  "don't do X because Y happened" than from abstract principles.

  Bottom line: The skill is ~30% useful as-is. The design principles section
  could be dropped entirely without loss. The project-specific rules are
  valuable but thin. The biggest improvement would be replacing the generic
  principles with concrete, driver-specific guidance on the missing items
  above — things a model can't infer from first principles.

```

I am uncertain that what is marked "useful" above will be in practice (see other comments).

bom/CLAUDE.md
```diff
@@ -0,0 +1 @@
+@AGENTS.md
```
katcharov: What is this pattern based on?

I tried to look up how it works, and info is scarce. As of 9 months ago, this convention did not imply that the file would be embedded in the other file. See comments in [this post](https://www.reddit.com/r/ClaudeAI/comments/1lr1g0d/claude_code_and_claudemd_should_you_spread_your/), someone experimented and confirmed that Claude sometimes only conditionally reads these files. I believe this means that:

1. We don't actually have a guarantee that a referenced file will be read, and the conditions under which it is read are unclear
2. Updating the referenced file while a session is in-flight will not cause an immediate context update, which makes tuning difficult
3. There will be an additional step (time, tokens) for reading this file

I don't think we should use this pattern if the above is true.

katcharov (changes requested):
Partial review. Agree with @ajcvickers points above. There is a fair bit of content here, and I am not sure how we can validate it.

.agents/skills/api-design/SKILL.md
rozza: You can ask the model to apply / create / update methods to add / remove these etc.. these. This is just a skill regarding their usage.

.agents/skills/api-design/SKILL.md
rozza: As method parameters

bom/CLAUDE.md
rozza: The internal guidelines as linked in the ticket description: https://wiki.corp.mongodb.com/spaces/MMS/pages/499158370/Making+a+Repo+Agent+Ready

The cloud team have used this pattern to great effect and it has allowed them to be model agnostic.

rozza:
@ajcvickers thanks for the feedback. Out of interest were the MMS team in the AI/ML-sync meeting? The formating of this PR and use of skills over a large Claude.md file came via Glean indiciating the Cloud teams approach : https://wiki.corp.mongodb.com/spaces/MMS/pages/499158370/Making+a+Repo+Agent+Ready

They seem to be a few months ahead with their adoption and usage of these tools. @katcharov know anyone on their team that could review this.

rozza:
So I think before going further, an ideological approach has to be agreed upon.

Do we add skills at all initially? We could .gitignore the skills directory and then manually add them on a one by one basis in other PR's?  That approach is handy and would allow users to have test their own skills without risk of accidently adding to the repo. Note the request for local skills https://github.com/anthropics/claude-code/issues/31155 has been closed so there is no local repo skills pattern other than via .gitignore.

We can start small - just the AGENTS.md files explaining the source code and how to compile and run the tests. This is useful to stop claude trying to use sbt instead of gradle everytime it tries to do anything in scala.

~Do we even agree with following the cloud teams https://agentskills.io/ approach and then @ importing the file into claude. We could just flip it and have CLAUDE.md as the source of truth and AGENTS.md being the links - as we use claude for development and copilot for reviews. Update: Claude CLI reads AGENTS.md files - so the CLAUDE.md files aren't actually needed unless there are claude specific instructions.~ Update: We need the CLAUDE.md files - checked and verified against claudes documentation.

bom/CLAUDE.md
rozza: ~Turns out both sources are out of date. AGENTS.md is all thats needed for claude now and will automatically imported. CLAUDE.md can contain claude specific logic above AGENTS.md.~

~Confirmed with Claude and read about it here: https://medium.com/data-science-collective/the-complete-guide-to-ai-agent-memory-files-claude-md-agents-md-and-beyond-49ea0df5c5a9 .~

Turns out you need the `@` pattern: https://code.claude.com/docs/en/memory#agents-md but can add overrides in your local CLAUDE.md files after the import.

.agents/skills/api-design/SKILL.md
rozza: Done. See:  https://gist.github.com/rozza/1a474291735bac0a28e399c59a977db2

.agents/skills/api-design/SKILL.md
rozza: I slightly modified the text for clarity.

rozza:
I made some changes after iterating with Claude.

Full report of the final outcomes: https://gist.github.com/rozza/1a474291735bac0a28e399c59a977db2

.agents/skills/api-design/SKILL.md
```diff
@@ -24,22 +24,8 @@ description: API stability annotations, design principles, and patterns for the
 - **General-purpose over special-case:** Fewer flexible methods over many specialized ones.
 - **Define errors out of existence:** Design APIs so errors cannot happen rather than detecting and handling them.
 
-## Search Before Implementing
```
rozza: This is covered by the AGENTS.md so was considered a duplication.

.agents/skills/api-design/SKILL.md
```diff
@@ -24,22 +24,8 @@ description: API stability annotations, design principles, and patterns for the
...
-- Check if a similar pattern is established elsewhere in the module
-- Reuse existing well-tested infrastructure over creating duplicates
-
 ## Key Patterns
 
 - Static factory methods: `Filters.eq()`, `Updates.set()`, `Aggregates.match()`
 - Fluent builders: `MongoClientSettings.builder()` is the primary entry point
 - Abstract core with pluggable transports
-
-## Public API Rules
```
rozza: Another duplication from the top level AGENTS.md

.agents/skills/style-reference/SKILL.md
```diff
@@ -31,6 +31,17 @@ description: Detailed code style rules for Java, Kotlin, Scala, and Groovy in th
 
 - **Static analysis:** CodeNarc
 
+## Javadoc / KDoc / Scaladoc
```
rozza, line 34: Claude identified these as missing

.agents/skills/sync-agents-docs/SKILL.md
```diff
@@ -0,0 +1,27 @@
+---
+name: sync-agents-docs
+description: Sync AGENTS.md files and skills after buildSrc or convention changes. Use after modifying build plugins, formatting rules, testing conventions, or task names to keep documentation consistent.
+disable-model-invocation: true
+---
+# Sync Documentation After Build Changes
```
rozza, line 6: Claude identified the buildSrc/AGENTS.md was not the correct place for this as it changes code. It also recommended this not be automatically run by claude.

.agents/skills/testing-guide/examples.md
```diff
@@ -0,0 +1,88 @@
+# Test Skeletons
```
rozza, line 1: Claude recommended having official test skeleton examples to use for new test cases.

bom/AGENTS.md
rozza, line 1: Claude recommended deleting a few AGENTS.md files as they didnt add value or were for niche projects within the codebase.

bson-kotlin/AGENTS.md
```diff
@@ -16,5 +16,3 @@ Kotlin extensions for the BSON library, providing idiomatic Kotlin codec support
 ./gradlew :bson-kotlin:test
 ./gradlew :bson-kotlin:check
 ```
-
-For global rules see [root AGENTS.md](../AGENTS.md).
```
rozza: These comments were considered unnecessary because the root agent is already in scope.

ajcvickers:
@rozza I think you should continue with this approach and on the C# driver we will start more minimal, and then let's revisit how things are working in a few weeks.

.agents/skills/api-design/SKILL.md
rozza: As these skills are tied to the codebase and the agents.md reviewing them in isolation (as generic skills) may not be useful.

However, the aim and the claude review has shown they do help provide context and reduce token usage. For example the annotations `@Alpha` are documented, but providing an overview saves tokens.

rozza:
@katcharov please re-review. See: https://gist.github.com/rozza/1a474291735bac0a28e399c59a977db2 for the claude report on the latest iteration.

dariakp:
> ~Do we even agree with following the cloud teams [agentskills.io](https://agentskills.io/) approach and then @ importing the file into claude. We could just flip it and have CLAUDE.md as the source of truth and AGENTS.md being the links - as we use claude for development and copilot for reviews.~. Update: Claude CLI reads AGENTS.md files - so the CLAUDE.md files aren't actually needed unless there are claude specific instructions.

@rozza If this is accurate, should the CLAUDE.md pointer files be removed?

dariakp (commented):
With respect to the overall content, this makes sense to me. I don't think we have to get it perfect. I didn't see anything obviously wrong here. The one note I have is that you mentioned in a previous comment that CLAUDE code does discover AGENTS.md - if so, should the rest of the CLAUDE.md pointers to AGENTS.md be removed?

rozza:
@dariakp - I updated the comment. Unfortunately the CLAUDE.md files are needed. I got a false positive from Claude when I asked if it read the AGENTS.md files - it proceeded to read them and then tell me yes - which caught me out.

Confirmation: https://code.claude.com/docs/en/memory#agents-md

dariakp (dismissed):

rozza:
@nhachicha I rebased folded all the commits into a single commit.

One main change was some of the skills I had were really references - so split them out so they will still be contextually loaded as required.

nhachicha (dismissed):
This looks great 🎉 
I did additional review via Glean (deep research) & Claude review, and summarised them below. Most of them could be dismissed or implemented in a future round of improvements...

### PR review concerns

- **High — Build / Gradle command guidance may be inaccurate or misleading**
  - Examples called out in review include `docs` vs `doc`, generated build config task names, and exact commands shown to agents.
  - **Action:** Verify every documented command against the actual Gradle build and keep only mechanically correct task names / invocation patterns.

- **High — Formatting / Spotless guidance may not match the build**
  - There is potential mismatch around `check` vs `spotlessApply` / `spotlessCheck`, and whether Java is actually formatted by Spotless.
  - **Action:** Re-audit formatting docs against the Gradle conventions and rewrite them to reflect the exact task dependencies and formatter coverage.

- **High — Some module/package references may be wrong or misleading**
  - In particular, `mongodb-crypt` and `driver-legacy` were called out as possibly referencing the wrong packages / surfaces.
  - **Action:** Verify all referenced packages/classes against the current codebase and update any incorrect guidance before merge.

- **High — `driver-core/AGENTS.md` may make an incorrect dependency claim**
  - The stated dependency on `bson-record-codec` may not match the actual dependency graph.
  - **Action:** Verify the dependency list against `driver-core/build.gradle.kts`; if `bson-record-codec` is optional, label it as optional, and if it is not a dependency, remove it.

- **Medium — `api-design.md` appears too thin / possibly truncated**
  - It seems to stop after the “Key Patterns” section and is missing some high-value driver-specific guidance.
  - **Action:** Add guidance for `@Deprecated` lifecycle, `@Alpha/@Beta` removal expectations, commonly misused patterns like `MongoNamespace` / `WriteConcern` / `ReadPreference`, and when to use `@Sealed` vs related patterns.

- **Medium — The API design reference is still too generic in places**
  - Some of the content reads like general software design advice rather than Java-driver-specific guidance.
  - **Action:** Trim generic principles and replace them with concrete driver-specific rules (annotation usage, public/internal API rules, module ownership, sync/async expectations).

- **Medium — `scripts/symlink-claude-md.sh` has unsafe / confusing behavior**
  - Concerns include name vs implementation mismatch, possible deletion of local files, and path sensitivity.
  - **Action:** Either make it a true symlink script with safe repo-root handling, or rename/document it clearly and add safeguards so it only touches generated files.

- **Medium — `symlink-claude-md.sh` is not wired into a reliable workflow**
  - As written, contributors/agents can forget to run it and commit `AGENTS.md` changes without matching `CLAUDE.md` updates.
  - **Action:** Either wire it into a pre-commit hook / install task, or explicitly document the required invocation point (`before git add` vs `before commit`) and enforce that convention in the README.

- **Medium — `sync-agents-docs` looks like a non-executable checklist disguised as a skill**
  - `disable-model-invocation: true` with no allowed tools means the agent cannot actually reason about which checklist items apply.
  - **Action:** Either remove `disable-model-invocation: true` so the agent can reason about relevance, or document clearly that this is intentionally a manual checklist only.

- **Medium — There may still be duplication across root AGENTS.md, per-module AGENTS.md, and references/skills**
  - This increases maintenance cost and risks stale or conflicting instructions.
  - **Action:** Do one final DRY pass and move repeated detail into references, keeping AGENTS.md files focused on always-needed guidance.

- **Medium — The `bom/` module is listed but has no `AGENTS.md`**
  - Even though it is low complexity, agents can still make dependency-management mistakes there.
  - **Action:** Add a minimal `bom/AGENTS.md` stating that dependency versions belong in the version catalog / central dependency management, not inline in module build files.

- **Low — `driver-lambda/` and `graalvm-native-image-app/` are listed in the project guide but have no `AGENTS.md`**
  - This may confuse agents about whether they are intended development targets.
  - **Action:** Either add minimal AGENTS files, or explicitly note in the project guide that these modules are intentionally undocumented / not normal development targets.

- **Low — The README’s reference to `agentskills.io` should be treated carefully**
  - If this is meant to be a long-term standards reference, it is worth confirming that this is the canonical external link the repo wants to depend on.
  - **Action:** Confirm the project wants to pin to `agentskills.io` long-term, or reference the underlying spec / canonical source directly if that is more stable.

- **Low — `testing-guide-examples.md` may contain a Kotlin mocking example contributors will copy incorrectly**
  - The `@Mock` example looks suspicious unless it is tied to a specific framework setup.
  - **Action:** Fix the snippet to use the intended mocking style, or add a brief note if the pattern is framework-specific.

- **Low — Markdown tables need a final rendered sanity check**
  - Some table-formatting concerns were raised during review and are easy to miss in raw diff form.
  - **Action:** Review all new tables in GitHub’s rendered view and fix any remaining formatting issues.

- **Low — Overall AGENTS content density should stay tightly controlled**
  - The root and module AGENTS files should remain minimal so they do not become high-token, low-signal context.
  - **Action:** Keep AGENTS.md minimal and move anything not needed in most sessions into references or skills.

nhachicha (approved):

PR #1928, opened 2026-03-26
[backport] Release connection between empty getMore responses 

vbabanin:
Backport of: https://github.com/mongodb/mongo-java-driver/pull/1925

> - Extract getMoreLoop to loop on empty batches while releasing the connection between each getMore, preventing pool starvation from idle tailable cursors (e.g., change streams)
> - Add test to verify connection is released and re-acquired between consecutive empty getMores
> 
> JAVA-6142

Note: 
- changes are adjusted to the current branch, as `AsyncCommandCursor` does not exists in `5.6.x`
- Additionally applied suggestion by @rozza from the main PR as well: https://github.com/mongodb/mongo-java-driver/pull/1925#pullrequestreview-4013208237.

- [x] Has AI automation been used? - Backport was done fully by Claude Code Sonnet 4.6.

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
@@ -121,7 +121,7 @@ class AsyncCommandBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T>
                 commandCursorResult = withEmptyResults(commandCursorResult);
                 funcCallback.onResult(batchResults, null);
             } else {
-                getMore(localServerCursor, funcCallback);
+                getMoreLoop(localServerCursor, funcCallback);
             }
         }, callback);
     }
@@ -183,10 +183,10 @@ class AsyncCommandBatchCursor<T> implements AsyncAggregateResponseBatchCursor<T>
 
     private void getMore(final ServerCursor cursor, final SingleResultCallback<List<T>> callback) {
         resourceManager.executeWithConnection((connection, wrappedCallback) ->
-                getMoreLoop(assertNotNull(connection), cursor, wrappedCallback), callback);
+                executeGetMoreCommand(assertNotNull(connection), cursor, wrappedCallback), callback);
     }
 
-    private void getMoreLoop(final AsyncConnection connection, final ServerCursor serverCursor,
+    private void executeGetMoreCommand(final AsyncConnection connection, final ServerCursor serverCursor,
             final SingleResultCallback<List<T>> callback) {
         connection.commandAsync(namespace.getDatabaseName(),
                 getMoreCommandDocument(serverCursor.getId(), connection.getDescription(), namespace, batchSize, comment),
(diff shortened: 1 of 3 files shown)


driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
```diff
@@ -206,19 +206,22 @@ private void getMoreLoop(final AsyncConnection connection, final ServerCursor se
...
+                    callback.onResult(commandCursorResult.getResults(), null);
+        });
+    }
 
-                    getMoreLoop(connection, nextServerCursor, callback);
+    private void getMoreLoop(final ServerCursor localServerCursor, final SingleResultCallback<List<T>> funcCallback) {
+        getMore(localServerCursor, (nextBatch, t) -> {
+            if (t != null) {
+                funcCallback.onResult(null, t);
+            } else if (resourceManager.getServerCursor() == null || (nextBatch != null && !nextBatch.isEmpty())) {
```
nhachicha, line 217: Interesting, `claude-review` flagged this as a potential _double volatile read_ and claimed that a concurrent `close()` could cause  `assertNotNull` to fail on the second read. However when I pushed back on this it realises the following 

```
You're right to push back — that claim was wrong.                                                                                                                                                                
                                                                                                                                     
  Looking at the actual state machine:                                                                                                                                                                             
                                                                                                                                                                                                                   
  1. When next() starts, tryStartOperation() transitions state from IDLE → OPERATION_IN_PROGRESS (line 148)                                                                                                        
  2. If close() is called concurrently while in OPERATION_IN_PROGRESS, it transitions to CLOSE_PENDING (line 186) — but it does not call doClose() and does not call unsetServerCursor()                           
  3. doClose() only runs later, when endOperation() sees CLOSE_PENDING (line 166-168)                                                                                                                              
  4. endOperation() is called after the next() callback completes                                                                                                                                                  
                                                                                                                                                                                                                   
  So during the entire execution of getMoreLoop's callback (lines 214-224), the state is OPERATION_IN_PROGRESS or CLOSE_PENDING. In neither case does serverCursor get unset — that only happens inside doClose(), 
  which is deferred until endOperation().                                                                                                                                                                          
                                                                                                                                                                                                                   
  This means the double volatile read on lines 217/223 is actually safe. The serverCursor cannot change between the two reads because:                                                                             
  - setServerCursor() asserts state.inProgress() and is only called from within the operation itself (line 208)
  - unsetServerCursor() is only called from doClose(), which cannot run until endOperation()                                                                                                                       
  - No other code path modifies serverCursor during an in-progress operation                                                         
```

nhachicha: Note:
This is still a warning reported by the IDE 

Image

driver-core/src/test/functional/com/mongodb/internal/operation/AsyncCommandBatchCursorTest.java
```diff
@@ -252,6 +256,32 @@ void closeShouldNotResetOriginalTimeout(final boolean disableTimeoutResetWhenClo
...
+            cb.onResult(new BsonDocument("cursor",
+                    new BsonDocument("ns", new BsonString(NAMESPACE.getFullName()))
+                            .append("id", new BsonInt64(callCount.incrementAndGet() < 3 ? 1 : 0))
+                            .append("nextBatch", new BsonArrayWrapper<>(new BsonArray()))), null);
+            return null;
+        }).when(mockConnection).commandAsync(eq(NAMESPACE.getDatabaseName()),
+                argThat(doc -> doc.containsKey("getMore")), any(), any(), any(), any(), any());
+
+        when(serverDescription.getType()).thenReturn(ServerType.STANDALONE);
+        createBatchCursor(0).next((result, t) -> {
```
nhachicha: [nit]
wrap with `try-with-resource`
```Java
try (AsyncCommandBatchCursor<Document> commandBatchCursor = createBatchCursor(0)) {
            commandBatchCursor.next((result, t) -> {
                assertNotNull(result);
                assertTrue(result.isEmpty());
                assertNull(t);
            });
        } catch (Exception e) {
        }
```

nhachicha (dismissed):

driver-core/src/main/com/mongodb/internal/operation/AsyncCommandBatchCursor.java
vbabanin: > However when I pushed back on this it realises the following

I agree with your analysis; that the code is functionally safe due to the state machine invariants. The `AsyncCommandCursor` and `CommandCursor` are not allowed to be used concurrently and protected by the state machine. The IDE warning is still legitimate from a static analysis perspective - it can't infer these runtime invariants. Therefore, it is safe to leave it as it is.

driver-core/src/test/functional/com/mongodb/internal/operation/AsyncCommandBatchCursorTest.java
vbabanin: Good catch. I just noticed that we do not do this consistently in all places. So I added `try-with-resource` everywhere in this test class - and removed catch as it is redundant, exception will fail the test anyway, wrapping was unnecessary.

driver-core/src/test/functional/com/mongodb/internal/operation/AsyncCommandBatchCursorTest.java
```diff
@@ -207,8 +202,6 @@ void closeShouldResetTimeoutContextToDefaultMaxTime() {
         try (AsyncCommandBatchCursor<Document> commandBatchCursor = createBatchCursor(maxTimeMS)) {
             // verify that the `maxTimeMS` override was applied
             timeoutContext.runMaxTimeMS(remainingMillis -> assertTrue(remainingMillis <= maxTimeMS));
-        } catch (Exception e) {
```
nhachicha, line 210: Why was this wrapped in RuntimeException in the first place?

nhachicha (approved):
Thanks for updating the try-with-resources.
Will you backport https://github.com/mongodb/mongo-java-driver/pull/1928/commits/52898a57c00cca3de3d0d2518273865005a7efb3 to `main`?

vbabanin:
> Will you backport https://github.com/mongodb/mongo-java-driver/commit/52898a57c00cca3de3d0d2518273865005a7efb3 to main?

i will create another cleanup PR for main.

PR #1918, opened 2026-03-21
Client Backpressure

stIncMale:
### How to include `main` changes into `backpressure`

Rebase `backpressure` on top of `main` when there are no outstanding PRs for `backpressure` that are being reviewed, because this leads to GitHub closing the PRs.

See https://github.com/mongodb/mongo-java-driver/pull/1917#issuecomment-4102007580 for more details.

### How to deal with TODOs

`TODO-BACKPRESSURE` is the tag we use to leave `TODO`s in PRs for the `backpressure` branch that must be addressed before the current PR is merged in `main`. When we are leaving such a comment, and we know who should be responsible for addressing it, we should add the name of that engineer in the comment (for example, @stIncMale uses `TODO-BACKPRESSURE Valentin` for the `TODO`s he should address).

JAVA-5942, JAVA-6019, JAVA-6261

PR #1929, opened 2026-03-27
JAVA-6055 Implement prose backpressure retryable writes tests

stIncMale:
The relevant spec changes:
- https://github.com/mongodb/specifications/blame/ba14b6bdc1dc695aa9cc20ccf9378592da1b2329/source/retryable-writes/tests/README.md#L265-L418
- See also https://jira.mongodb.org/browse/DRIVERS-3432 for the phrasing fixes for "Test 3 Case 3"

I recommend reviewing the first two commits one by one to simplify reviewing: the first commit does the refactoring, including the necessary refactoring (the introduction of `ConfigureFailPointCommandListener`), the second commit adds the new tests.

I used the `TODO-BACKPRESSURE Valentin` comments as described in https://github.com/mongodb/mongo-java-driver/pull/1918.

JAVA-6055

### DIFF
--- a/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableReadsProseTest.java
+++ b/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableReadsProseTest.java
@@ -22,32 +22,27 @@ import com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient;
 import org.bson.Document;
 import org.junit.jupiter.api.Test;
 
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeoutException;
-
 import static com.mongodb.client.model.Filters.eq;
 
 /**
- * See
- * <a href="https://github.com/mongodb/specifications/tree/master/source/retryable-reads/tests">Retryable Reads Tests</a>.
+ * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.md#prose-tests">
+ * Prose Tests</a>.
  */
 final class RetryableReadsProseTest {
     /**
-     * See
-     * <a href="https://github.com/mongodb/specifications/tree/master/source/retryable-reads/tests#poolclearederror-retryability-test">
-     * PoolClearedError Retryability Test</a>.
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.md#1-poolclearederror-retryability-test">
+     * 1. PoolClearedError Retryability Test</a>.
      */
     @Test
-    void poolClearedExceptionMustBeRetryable() throws InterruptedException, ExecutionException, TimeoutException {
+    void poolClearedExceptionMustBeRetryable() throws Exception {
         RetryableWritesProseTest.poolClearedExceptionMustBeRetryable(
                 SyncMongoClient::new,
                 mongoCollection -> mongoCollection.find(eq(0)).iterator().hasNext(), "find", false);
     }
 
     /**
-     * See
-     * <a href="https://github.com/mongodb/specifications/tree/master/source/retryable-reads/tests#21-retryable-reads-are-retried-on-a-different-mongos-when-one-is-available">
-     * Retryable Reads Are Retried on a Different mongos When One is Available</a>.
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.md#21-retryable-reads-are-retried-on-a-different-mongos-when-one-is-available">
+     * 2.1 Retryable Reads Are Retried on a Different mongos When One is Available</a>.
      */
     @Test
     void retriesOnDifferentMongosWhenAvailable() {
@@ -61,9 +56,8 @@ final class RetryableReadsProseTest {
     }
 
     /**
-     * See
-     * <a href="https://github.com/mongodb/specifications/tree/master/source/retryable-reads/tests#22-retryable-reads-are-retried-on-the-same-mongos-when-no-others-are-available">
-     * Retryable Reads Are Retried on the Same mongos When No Others are Available</a>.
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.md#22-retryable-reads-are-retried-on-the-same-mongos-when-no-others-are-available">
+     * 2.2 Retryable Reads Are Retried on the Same mongos When No Others are Available</a>.
      */
     @Test
     void retriesOnSameMongosWhenAnotherNotAvailable() {
--- a/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableWritesProseTest.java
+++ b/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableWritesProseTest.java
@@ -16,66 +16,53 @@
 
 package com.mongodb.reactivestreams.client;
 
-import com.mongodb.client.test.CollectionHelper;
 import com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient;
 import org.bson.Document;
-import org.bson.codecs.DocumentCodec;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeoutException;
-
 /**
- * See
- * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#prose-tests">Retryable Write Prose Tests</a>.
+ * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#prose-tests">
+ * Prose Tests</a>.
  */
-public class RetryableWritesProseTest extends DatabaseTestCase {
-    private CollectionHelper<Document> collectionHelper;
-
-    @BeforeEach
-    @Override
-    public void setUp() {
-        super.setUp();
-
-        collectionHelper = new CollectionHelper<>(new DocumentCodec(), collection.getNamespace());
-        collectionHelper.create();
-    }
-
+final class RetryableWritesProseTest {
     /**
-     * Prose test #2.
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#2-test-that-drivers-properly-retry-after-encountering-poolclearederrors">
+     * 2. Test that drivers properly retry after encountering PoolClearedErrors</a>.
      */
     @Test
-    public void poolClearedExceptionMustBeRetryable() throws InterruptedException, ExecutionException, TimeoutException {
+    void poolClearedExceptionMustBeRetryable() throws Exception {
         com.mongodb.client.RetryableWritesProseTest.poolClearedExceptionMustBeRetryable(
                 SyncMongoClient::new,
                 mongoCollection -> mongoCollection.insertOne(new Document()), "insert", true);
     }
 
     /**
-     * Prose test #3.
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#3-test-that-drivers-return-the-original-error-after-encountering-a-writeconcernerror-with-a-retryablewriteerror-label">
+     * 3. Test that drivers return the original error after encountering a WriteConcernError with a RetryableWriteError label</a>.
      */
     @Test
-    public void originalErrorMustBePropagatedIfNoWritesPerformed() throws InterruptedException {
+    void originalErrorMustBePropagatedIfNoWritesPerformed() throws Exception {
         com.mongodb.client.RetryableWritesProseTest.originalErrorMustBePropagatedIfNoWritesPerformed(
                 SyncMongoClient::new);
     }
 
     /**
-     * Prose test #4.
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#4-test-that-in-a-sharded-cluster-writes-are-retried-on-a-different-mongos-when-one-is-available">
+     * 4. Test that in a sharded cluster writes are retried on a different mongos when one is available</a>.
      */
     @Test
-    public void retriesOnDifferentMongosWhenAvailable() {
+    void retriesOnDifferentMongosWhenAvailable() {
         com.mongodb.client.RetryableWritesProseTest.retriesOnDifferentMongosWhenAvailable(
                 SyncMongoClient::new,
                 mongoCollection -> mongoCollection.insertOne(new Document()), "insert", true);
     }
 
     /**
-     * Prose test #5.
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#5-test-that-in-a-sharded-cluster-writes-are-retried-on-the-same-mongos-when-no-others-are-available">
+     * 5. Test that in a sharded cluster writes are retried on the same mongos when no others are available</a>.
      */
     @Test
-    public void retriesOnSameMongosWhenAnotherNotAvailable() {
+    void retriesOnSameMongosWhenAnotherNotAvailable() {
         com.mongodb.client.RetryableWritesProseTest.retriesOnSameMongosWhenAnotherNotAvailable(
                 SyncMongoClient::new,
                 mongoCollection -> mongoCollection.insertOne(new Document()), "insert", true);
--- a/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableWritesProseTest.java
+++ b/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableWritesProseTest.java
@@ -96,7 +96,6 @@ final class RetryableWritesProseTest {
      * Case 3: Test that drivers return the correct error when receiving some errors with NoWritesPerformed and some without NoWritesPerformed</a>.
      */
     @Test
-    @Disabled("TODO-BACKPRESSURE Valentin Enable when implementing JAVA-6055, fails on MongoDB 6.0")
     void errorPropagationAfterEncounteringMultipleErrorsCase3() throws Exception {
         com.mongodb.client.RetryableWritesProseTest.errorPropagationAfterEncounteringMultipleErrorsCase3(SyncMongoClient::new);
     }
--- a/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableWritesProseTest.java
+++ b/driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableWritesProseTest.java
@@ -18,6 +18,7 @@ package com.mongodb.reactivestreams.client;
 
 import com.mongodb.reactivestreams.client.syncadapter.SyncMongoClient;
 import org.bson.Document;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 /**
@@ -66,5 +67,37 @@ final class RetryableWritesProseTest {
         com.mongodb.client.RetryableWritesProseTest.retriesOnSameMongosWhenAnotherNotAvailable(
                 SyncMongoClient::new,
                 mongoCollection -> mongoCollection.insertOne(new Document()), "insert", true);
+    }
+
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#case-1-test-that-drivers-return-the-correct-error-when-receiving-only-errors-without-nowritesperformed">
+     * 6. Test error propagation after encountering multiple errors.
+     * Case 1: Test that drivers return the correct error when receiving only errors without NoWritesPerformed</a>.
+     */
+    @Test
+    @Disabled("TODO-BACKPRESSURE Valentin Enable when implementing JAVA-6055")
+    void errorPropagationAfterEncounteringMultipleErrorsCase1() throws Exception {
+        com.mongodb.client.RetryableWritesProseTest.errorPropagationAfterEncounteringMultipleErrorsCase1(SyncMongoClient::new);
+    }
+
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#case-2-test-that-drivers-return-the-correct-error-when-receiving-only-errors-with-nowritesperformed">
+     * 6. Test error propagation after encountering multiple errors.
+     * Case 2: Test that drivers return the correct error when receiving only errors with NoWritesPerformed</a>.
+     */
+    @Test
+    void errorPropagationAfterEncounteringMultipleErrorsCase2() throws Exception {
+        com.mongodb.client.RetryableWritesProseTest.errorPropagationAfterEncounteringMultipleErrorsCase2(SyncMongoClient::new);
+    }
+
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#case-3-test-that-drivers-return-the-correct-error-when-receiving-some-errors-with-nowritesperformed-and-some-without-nowritesperformed">
+     * 6. Test error propagation after encountering multiple errors.
+     * Case 3: Test that drivers return the correct error when receiving some errors with NoWritesPerformed and some without NoWritesPerformed</a>.
+     */
+    @Test
+    @Disabled("TODO-BACKPRESSURE Valentin Enable when implementing JAVA-6055, fails on MongoDB 6.0")
+    void errorPropagationAfterEncounteringMultipleErrorsCase3() throws Exception {
+        com.mongodb.client.RetryableWritesProseTest.errorPropagationAfterEncounteringMultipleErrorsCase3(SyncMongoClient::new);
     }
 }
--- a/driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
@@ -19,6 +19,7 @@ package com.mongodb.client;
 import com.mongodb.ConnectionString;
 import com.mongodb.Function;
 import com.mongodb.MongoClientSettings;
+import com.mongodb.MongoException;
 import com.mongodb.MongoServerException;
 import com.mongodb.MongoWriteConcernException;
 import com.mongodb.ServerAddress;
@@ -34,9 +35,11 @@ import com.mongodb.internal.connection.ServerAddressHelper;
 import com.mongodb.internal.connection.TestCommandListener;
 import com.mongodb.internal.connection.TestConnectionPoolListener;
 import com.mongodb.internal.event.ConfigureFailPointCommandListener;
+import com.mongodb.lang.Nullable;
 import org.bson.BsonDocument;
 import org.bson.BsonInt32;
 import org.bson.Document;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import java.util.HashSet;
@@ -55,6 +58,8 @@ import static com.mongodb.ClusterFixture.isDiscoverableReplicaSet;
 import static com.mongodb.ClusterFixture.isSharded;
 import static com.mongodb.ClusterFixture.isStandalone;
 import static com.mongodb.ClusterFixture.serverVersionAtLeast;
+import static com.mongodb.MongoException.RETRYABLE_ERROR_LABEL;
+import static com.mongodb.MongoException.SYSTEM_OVERLOADED_ERROR_LABEL;
 import static com.mongodb.client.Fixture.getDefaultDatabaseName;
 import static com.mongodb.client.Fixture.getMongoClientSettingsBuilder;
 import static com.mongodb.client.Fixture.getMultiMongosMongoClientSettingsBuilder;
@@ -66,6 +71,7 @@ import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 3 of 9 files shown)


driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +339,182 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+     * @param unexpectedErrorLabel {@code null} means there is no expectation.
+     */
+    private static void errorPropagationAfterEncounteringMultipleErrors(
+            final Function<MongoClientSettings, MongoClient> clientCreator,
+            final BsonDocument configureFailPoint,
+            final BsonDocument configureFailPointFromListener,
+            final Predicate<CommandEvent> configureFailPointEventMatcher,
+            final int expectedErrorCode,
+            @Nullable final String unexpectedErrorLabel) throws Exception {
+        assumeTrue(serverVersionAtLeast(6, 0));
```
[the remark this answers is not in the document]
stIncMale: Done.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableReadsProseTest.java
```diff
@@ -22,32 +22,27 @@
...
     /**
-     * See
-     * <a href="https://github.com/mongodb/specifications/tree/master/source/retryable-reads/tests#poolclearederror-retryability-test">
-     * PoolClearedError Retryability Test</a>.
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.md#1-poolclearederror-retryability-test">
+     * 1. PoolClearedError Retryability Test</a>.
      */
     @Test
-    void poolClearedExceptionMustBeRetryable() throws InterruptedException, ExecutionException, TimeoutException {
+    void poolClearedExceptionMustBeRetryable() throws Exception{
```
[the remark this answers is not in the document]
stIncMale: Done.

driver-sync/src/test/functional/com/mongodb/internal/event/ConfigureFailPointCommandListener.java
```diff
@@ -0,0 +1,101 @@
...
+        synchronized (lock) {
+            if (!failPointFuture.isDone() && eventMatcher.test(event)) {
+                FailPoint failPoint;
+                try {
+                    failPoint = FailPoint.enable(configureFailPoint, serverAddress);
+                } catch (Throwable e) {
+                    assertTrue(failPointFuture.completeExceptionally(e));
+                    return;
+                }
+                assertTrue(failPointFuture.complete(failPoint));
```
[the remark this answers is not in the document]
stIncMale: Excellent! Done.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -160,91 +150,86 @@ public static <R> void poolClearedExceptionMustBeRetryable(
...
+        BsonDocument configureFailPoint = BsonDocument.parse(
+                "{\n"
+                + "    configureFailPoint: \"failCommand\",\n"
+                + "    mode: { times: 1 },\n"
+                + "    data: {\n"
+                + "        failCommands: ['insert'],\n"
+                + "        errorLabels: ['" + RETRYABLE_WRITE_ERROR_LABEL + "'],\n"
+                + "        writeConcernError: {"
+                + "            code: 91,\n"
+                + "            errmsg: ''\n"
```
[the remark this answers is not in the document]
stIncMale: Done.

The test contradicts itself. But I did the change as Copilot suggested because this is also what we had before the refactoring.

P.S. I am too irritated with all the mess in the specs to go into another investigation regarding this old test, and to file a DRIVERS ticket to fix it.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -160,91 +150,86 @@ public static <R> void poolClearedExceptionMustBeRetryable(
...
+                + "    mode: { times: 1 },\n"
+                + "    data: {\n"
+                + "        failCommands: ['insert'],\n"
+                + "        errorCode: 10107,\n"
+                + "        errorLabels: ['" + RETRYABLE_WRITE_ERROR_LABEL + "', '" + NO_WRITES_PERFORMED_ERROR_LABEL + "']\n"
+                + "    }\n"
+                + "}\n");
+        Predicate<CommandEvent> configureFailPointEventMatcher = event -> {
+            if (event instanceof CommandSucceededEvent) {
+                CommandSucceededEvent  commandSucceededEvent = (CommandSucceededEvent) event;
```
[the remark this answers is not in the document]
stIncMale: Done.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +338,183 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+            @Nullable final String unexpectedErrorLabel) throws Exception {
+        assumeTrue(serverVersionAtLeast(6, 0));
+        assumeTrue(isDiscoverableReplicaSet());
+        try (ConfigureFailPointCommandListener commandListener = new ConfigureFailPointCommandListener(
+                configureFailPointFromListener, getPrimary(), configureFailPointEventMatcher);
+             MongoClient client = clientCreator.apply(getMongoClientSettingsBuilder()
+                     .retryWrites(true)
+                     .addCommandListener(commandListener)
+                     .build());
+             FailPoint ignored = FailPoint.enable(configureFailPoint, getPrimary())) {
```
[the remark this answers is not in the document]
stIncMale: Done.

I disagree with the explanation, but agree with the suggestion.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableReadsProseTest.java
```diff
@@ -22,32 +22,27 @@
...
-     * PoolClearedError Retryability Test</a>.
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.md#1-poolclearederror-retryability-test">
+     * 1. PoolClearedError Retryability Test</a>.
      */
     @Test
-    void poolClearedExceptionMustBeRetryable() throws InterruptedException, ExecutionException, TimeoutException {
+    void poolClearedExceptionMustBeRetryable() throws Exception {
         RetryableWritesProseTest.poolClearedExceptionMustBeRetryable(
                 SyncMongoClient::new,
                 mongoCollection -> mongoCollection.find(eq(0)).iterator().hasNext(), "find", false);
```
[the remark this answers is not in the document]
stIncMale: Rejected.

This is by design, we refer via a fully qualified name, and that's fine.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +338,184 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+                + "}\n");
+        BsonDocument configureFailPointFromListener = BsonDocument.parse(
+                "{\n"
+                + "    configureFailPoint: \"failCommand\",\n"
+                + "    mode: 'alwaysOn',\n"
+                + "    data: {\n"
+                + "        failCommands: ['insert'],\n"
+                + "        errorCode: 91,\n"
+                + "        errorLabels: ['" + RETRYABLE_ERROR_LABEL + "', '" + SYSTEM_OVERLOADED_ERROR_LABEL
+                + "', '" + NO_WRITES_PERFORMED_ERROR_LABEL + "'],\n"
```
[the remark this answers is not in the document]
stIncMale: Done.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +338,184 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+                + "}\n");
+        BsonDocument configureFailPointFromListener = BsonDocument.parse(
+                "{\n"
+                + "    configureFailPoint: \"failCommand\",\n"
+                + "    mode: 'alwaysOn',\n"
+                + "    data: {\n"
+                + "        failCommands: ['insert'],\n"
+                + "        errorCode: 10107,\n"
+                + "            errorLabels: ['" + RETRYABLE_ERROR_LABEL + "', '" + SYSTEM_OVERLOADED_ERROR_LABEL
+                + "', '" + NO_WRITES_PERFORMED_ERROR_LABEL + "'],\n"
```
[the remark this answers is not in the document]
stIncMale: Done.

driver-sync/src/test/functional/com/mongodb/internal/event/ConfigureFailPointCommandListener.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    public void close() throws InterruptedException, ExecutionException {
+        synchronized (lock) {
+            if (failPointFuture.cancel(true)) {
+                fail("The listener was closed before (in the happens-before order) it attempted to configure the fail point");
+            } else {
+                assertTrue(failPointFuture.isDone());
+                assertNotNull(failPointFuture.get()).close();
```
nhachicha, line 101: [minor]
If `failPointFuture` is completed exceptionally (via `completeExceptionally`), then `failPointFuture.get()` in `close()` will throw, so the normal `FailPoint.close()` is skipped and the fail point may remain enabled.

This is, however, low-risk since there may be nothing to clean up if `FailPoint.enable()` fails ...

driver-sync/src/test/functional/com/mongodb/internal/event/ConfigureFailPointCommandListener.java
```diff
@@ -0,0 +1,100 @@
...
+                    assertTrue(failPointFuture.completeExceptionally(e));
+                }
+            }
+        }
+    }
+
+    @Override
+    public void close() throws InterruptedException, ExecutionException {
+        synchronized (lock) {
+            if (failPointFuture.cancel(true)) {
```
nhachicha, line 97: [TIL]
`mayInterruptIfRunning` is [no-op](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/CompletableFuture.html#cancel(boolean))
> mayInterruptIfRunning - this value has no effect in this implementation because interrupts are not used to control processing.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -253,37 +238,35 @@ public static <R> void retriesOnDifferentMongosWhenAvailable(
...
         ServerAddress s1Address = ServerAddressHelper.createServerAddress(connectionString.getHosts().get(1));
-        BsonDocument failPointDocument = BsonDocument.parse(
+        BsonDocument configureFailPoint = BsonDocument.parse(
                 "{\n"
                 + "    configureFailPoint: \"failCommand\",\n"
                 + "    mode: { times: 1 },\n"
                 + "    data: {\n"
-                + "        failCommands: [\"" + operationName + "\"],\n"
+                + "        failCommands: [\"" + expectedCommandName + "\"],\n"
+                + "        errorCode: 6,\n"
```
nhachicha, line 250: [claude-review]
Trailing comma in generated JSON when `write=false`.

The old code avoided this by placing `errorCode` last

The same pattern appears in `retriesOnSameMongosWhenAnotherNotAvailable`

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +338,184 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+                + "    data: {\n"
+                + "        failCommands: ['insert'],\n"
+                + "        errorCode: 10107,\n"
+                + "        errorLabels: ['" + RETRYABLE_ERROR_LABEL + "', '" + SYSTEM_OVERLOADED_ERROR_LABEL + "']\n"
+                + "    }\n"
+                + "}\n");
+        Predicate<CommandEvent> configureFailPointEventMatcher = event -> {
+            if (event instanceof CommandFailedEvent) {
+                CommandFailedEvent commandFailedEvent = (CommandFailedEvent) event;
+                MongoException cause = assertInstanceOf(MongoException.class, commandFailedEvent.getThrowable());
```
nhachicha: [claude-review]
The predicates use `assertInstanceOf(MongoException.class, ...)` which throws `AssertionFailedError` if the type doesn't match. A `Predicate.test()` should return `false`, not throw. The exception is caught by `onEvent()` and routed to `completeExceptionally`, so it works, but the semantics are surprising. Consider using a plain instanceof check instead.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +338,184 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+                + "        failCommands: ['insert'],\n"
+                + "        errorCode: 10107,\n"
+                + "        errorLabels: ['" + RETRYABLE_ERROR_LABEL + "', '" + SYSTEM_OVERLOADED_ERROR_LABEL
+                + "', '" + NO_WRITES_PERFORMED_ERROR_LABEL + "']\n"
+                + "    }\n"
+                + "}\n");
+        Predicate<CommandEvent> configureFailPointEventMatcher = event -> {
+            if (event instanceof CommandFailedEvent) {
+                CommandFailedEvent commandFailedEvent = (CommandFailedEvent) event;
+                MongoException cause = assertInstanceOf(MongoException.class, commandFailedEvent.getThrowable());
```
nhachicha: Ditto https://github.com/mongodb/mongo-java-driver/pull/1929/changes#diff-8fe5319aef55e451d009c31e961108720a7d361d60dd5c5d2c0a99e088d0545dR378

driver-sync/src/test/functional/com/mongodb/internal/event/ConfigureFailPointCommandListener.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.function.Predicate;
+
+import static com.mongodb.assertions.Assertions.assertNotNull;
+import static com.mongodb.assertions.Assertions.assertTrue;
+import static com.mongodb.assertions.Assertions.fail;
+
+public final class ConfigureFailPointCommandListener implements CommandListener, AutoCloseable {
```
nhachicha, line 38: Worth annotating with `@ThreadSafe` ?

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -129,24 +120,23 @@ public static <R> void poolClearedExceptionMustBeRetryable(
...
-                        .append("blockTimeMS", new BsonInt32(1000)));
+        BsonDocument configureFailPoint = BsonDocument.parse(
+                "{\n"
+                + "    configureFailPoint: 'failCommand',\n"
+                + "    mode: {'times': 1},\n"
+                + "    data: {\n"
+                + "        failCommands: ['" + commandName + "'],\n"
+                + "        errorCode: 91,\n"
+                + "        blockConnection: true,\n"
+                + "        blockTimeMS: 1000,\n"
```
nhachicha, line 131: Trailing comma issue when `write == false`. I wonder if we should keep the _append_ approach of building the `BsonDocument`

similar to https://github.com/mongodb/mongo-java-driver/pull/1929/changes#r3005045523

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -295,38 +278,41 @@ public static <R> void retriesOnDifferentMongosWhenAvailable(
...
         ServerAddress s0Address = ServerAddressHelper.createServerAddress(connectionString.getHosts().get(0));
-        BsonDocument failPointDocument = BsonDocument.parse(
+        BsonDocument configureFailPoint = BsonDocument.parse(
                 "{\n"
                 + "    configureFailPoint: \"failCommand\",\n"
                 + "    mode: { times: 1 },\n"
                 + "    data: {\n"
-                + "        failCommands: [\"" + operationName + "\"],\n"
+                + "        failCommands: [\"" + expectedCommandName + "\"],\n"
+                + "        errorCode: 6,\n"
```
nhachicha, line 309: Ditto https://github.com/mongodb/mongo-java-driver/pull/1929/changes#r3005045523

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +338,184 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+                + "    configureFailPoint: \"failCommand\",\n"
+                + "    mode: 'alwaysOn',\n"
+                + "    data: {\n"
+                + "        failCommands: ['insert'],\n"
+                + "        errorCode: 91,\n"
+                + "        errorLabels: ['" + RETRYABLE_ERROR_LABEL + "', '" + SYSTEM_OVERLOADED_ERROR_LABEL
+                + "', '" + NO_WRITES_PERFORMED_ERROR_LABEL + "']\n"
+                + "    }\n"
+                + "}\n");
+        Predicate<CommandEvent> configureFailPointEventMatcher = event -> event instanceof CommandFailedEvent;
```
nhachicha: [minor]
Both the initial and the listener fail points use error code 91, for consistency and defensiveness, checking code == 91 wouldn't hurt — if something unexpected fails first (e.g., a different command), the test would configure the second fail point at the wrong time and produce a confusing failure.

nhachicha (commented):
Good idea extracting `ConfigureFailPointCommandListener` as a reusable utility 👍 
some minor comments from me & Claude review

driver-sync/src/test/functional/com/mongodb/internal/event/ConfigureFailPointCommandListener.java
stIncMale: > If `failPointFuture` is completed exceptionally (via `completeExceptionally`), then `failPointFuture.get()` in `close()` will throw, so the normal `FailPoint.close()` is skipped

No, the execution you are proposing is forbidden, and the conclusion _"so the normal `FailPoint.close()` is skipped"_ is incorrect, assuming that you imply that a reference to a `FailPoint` was returned (I have to assume this, as otherwise there is no reference which we could call `close` through):

1. Due to the `synchronized` blocks, there is a guarantee that either the whole `onEvent` is ordered before (in the happens-before order) `close`, or `close` is ordered before `onEvent`<sup>1</sup>.
2. `failPointFuture.get()` is not part of any execution, where `close` is ordered before `onEvent`.
3. In all executions where `onEvent` is ordered before `close`, `failPointFuture.get()` throws if and only if `failPointFuture.completeExceptionally(e)` was executed as part of `onEvent`. When the latter is the case, no fail point were enabled<sup>2</sup>, and, therefore, no fail point needs to be closed.

---

<sup>1</sup> Strictly speaking, that's not the right way to say what I am saying, but I don't want to make things more complex, because that will change nothing.

<sup>2</sup> Actually, `FailPoint.enable` throwing does not guarantee that the `configureFailPoint` command failed to configure the fail point. That's a bug in `FailPoint.enable`, I fixed it in `main`: https://github.com/mongodb/mongo-java-driver/pull/1931.

stIncMale: If, however, you were talking exactly about the bug in `FailPoin.enable` (I introduced it in the past when added this class):

> there may be nothing to clean up if `FailPoint.enable()` fails

then no code added/modified in the current PR is relevant, and any usage of `FailPoint` is problematic because of that bug.

If that's the case, then it should be fine when https://github.com/mongodb/mongo-java-driver/pull/1931 is merged.

driver-sync/src/test/functional/com/mongodb/internal/event/ConfigureFailPointCommandListener.java
stIncMale: Thank you, I know. But we have to pass something.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: Rejected.

The AI is reasonable, however, our parser is fine with such commas. And I wanted the order of the fields to match the specification, as that simplifies comparison for humans.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: Rejected.

The AI is wrong here, and having assertions in event matchers is desirable. Specifically in the code highlighted, the test does not expect a command to fail with anything but `MongoException`. When this expectation fails, the test should not just skip enabling the fail point, but should fail explicitly complaining about the unexpected situation. Having an assertion allows us to achieve this.

If anything, I realized that we should double down on assertions in event matchers, and replaced other comparisons with assertions where appropriate in fcd0a1c420a93c86e085bf1c74906f8233dae791.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: See https://github.com/mongodb/mongo-java-driver/pull/1929#discussion_r3014570873.

driver-sync/src/test/functional/com/mongodb/internal/event/ConfigureFailPointCommandListener.java
stIncMale: Done in f9fd19e1473caae4614bae9374913c627ceb9d32.

None of our other named (non-anonymous) implementations of `CommandListener` are annotated with `@ThreadSafe`, even though some of them are thread-safe, and some don't have to be thread-safe due to the way they are used in tests. I marked all of them with `@ThreadSafe`/`@NotThreadSafe`.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: See https://github.com/mongodb/mongo-java-driver/pull/1929#discussion_r3014570873.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: See https://github.com/mongodb/mongo-java-driver/pull/1929#discussion_r3014570873.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: Done in 78915452fa6d9634996f4fe59175d01d8fcba2da.

This made the `configureFailPointEventMatcher` in case 3 the same as the one in cases 1 and 2, which lead to a nice simplification.

stIncMale (commented):
I am ready for another review round. @nhachicha, thank you for marking AI-generated concerns with `[claude-review]`.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
stIncMale: Adding assertions proved to be immediately useful: MongoDB 6.0 seem to error when a non-existing collections is dropped. That makes no sense on its own, and also given that neither MongoDB 5.0 nor MongoDB 7.0 behaves this way. I'll have to investigate.

nhachicha: I'm resolving this one. Please create an issue to investigate the MongoDB 6.0 behaviour 👍

nhachicha (dismissed):

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -129,24 +120,23 @@ public static <R> void poolClearedExceptionMustBeRetryable(
...
+                "{\n"
+                + "    configureFailPoint: 'failCommand',\n"
+                + "    mode: {'times': 1},\n"
+                + "    data: {\n"
+                + "        failCommands: ['" + commandName + "'],\n"
+                + "        errorCode: 91,\n"
+                + "        blockConnection: true,\n"
+                + "        blockTimeMS: 1000,\n"
+                + (write
+                ? "        errorLabels: ['" + RETRYABLE_WRITE_ERROR_LABEL + "']\n" : "")
```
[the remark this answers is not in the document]
stIncMale: Rejected.

I think, the way it is currently written makes it easier to read in the source code than what Copilot suggests.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -482,6 +482,10 @@ private static void errorPropagationAfterEncounteringMultipleErrors(
         Predicate<CommandEvent> configureFailPointEventMatcher = event -> {
             if (event instanceof CommandFailedEvent) {
                 CommandFailedEvent commandFailedEvent = (CommandFailedEvent) event;
+                if (commandFailedEvent.getCommandName().equals("drop")) {
```
nhachicha, line 484: This skips any `drop` command event, regardless of the error code. The intent is to only skip `NamespaceNotFound (26)` that Mongod 6 raises when dropping a nonexistent collection...

stIncMale: > The intent is to only skip `NamespaceNotFound (26)`

That's not the intent. The `ConfigureFailPointCommandListener` created in the test is not supposed to be concerned with the `drop` commands and any events it may produce, and it would not have the code highlighted if it were not for the divergence between the server and the driver (the server error happens, but `MongoCollection.drop` succeeds). The listener ignoring all of the `drop`-related failure events is no different from it already ignoring all of the command started/succeeded events (regardless of the command) - neither leads to the test passing despite something having went wrong. For example, if dropping a collection fails (from the driver perspective), the `MongoCollection.drop` method completes abruptly and the test fails - the listener does not need to and should not be the one dealing with that.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +341,176 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+            assertEquals(expectedErrorCode, e.getCode());
+            if (unexpectedErrorLabel != null) {
+                assertFalse(e.hasErrorLabel(unexpectedErrorLabel));
+            }
+        }
+    }
+
+    private static MongoCollection<Document> dropAndGetCollection(final String name, final MongoClient client) {
+        MongoCollection<Document> result = client.getDatabase(getDefaultDatabaseName()).getCollection(name);
+        result.drop();
```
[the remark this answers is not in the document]
stIncMale: AI is wrong, `MongoCollection.drop` does not throw when the `NamespaceNotFound` error is produced by the server.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +341,176 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#case-1-test-that-drivers-return-the-correct-error-when-receiving-only-errors-without-nowritesperformed">
+     * 6. Test error propagation after encountering multiple errors.
+     * Case 1: Test that drivers return the correct error when receiving only errors without NoWritesPerformed</a>.
+     */
+    @Test
+    @Disabled("TODO-BACKPRESSURE Valentin Enable when implementing JAVA-6055")
+    void errorPropagationAfterEncounteringMultipleErrorsCase1() throws Exception {
+        errorPropagationAfterEncounteringMultipleErrorsCase1(MongoClients::create);
```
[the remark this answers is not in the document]
stIncMale: > or enable it once the corresponding backpressure behavior is implemented

That is exactly what will be done. And we'll know it has to be done because of the `TODO-BACKPRESSURE` tag - no such tags are allows to be merged into `main`.

driver-sync/src/test/functional/com/mongodb/client/RetryableWritesProseTest.java
```diff
@@ -354,4 +341,176 @@ public static <R> void retriesOnSameMongosWhenAnotherNotAvailable(
...
+
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#case-3-test-that-drivers-return-the-correct-error-when-receiving-some-errors-with-nowritesperformed-and-some-without-nowritesperformed">
+     * 6. Test error propagation after encountering multiple errors.
+     * Case 3: Test that drivers return the correct error when receiving some errors with NoWritesPerformed and some without NoWritesPerformed</a>.
+     */
+    @Test
+    @Disabled("TODO-BACKPRESSURE Valentin Enable when implementing JAVA-6055, fails on MongoDB 6.0")
+    void errorPropagationAfterEncounteringMultipleErrorsCase3() throws Exception {
+        errorPropagationAfterEncounteringMultipleErrorsCase3(MongoClients::create);
```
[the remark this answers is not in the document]
stIncMale: The test no longer fails on MongoDB 6, because of 4b9a565902cfd4f6885592b190e9125a2dd1e24c. So it can and should be enabled now, rather than later.

Done in 68fdf37638f3e31f927ab2b6bd81d6aad3722cf9.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableWritesProseTest.java
```diff
@@ -16,68 +16,88 @@
...
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#case-1-test-that-drivers-return-the-correct-error-when-receiving-only-errors-without-nowritesperformed">
+     * 6. Test error propagation after encountering multiple errors.
+     * Case 1: Test that drivers return the correct error when receiving only errors without NoWritesPerformed</a>.
+     */
+    @Test
+    @Disabled("TODO-BACKPRESSURE Valentin Enable when implementing JAVA-6055")
+    void errorPropagationAfterEncounteringMultipleErrorsCase1() throws Exception {
+        com.mongodb.client.RetryableWritesProseTest.errorPropagationAfterEncounteringMultipleErrorsCase1(SyncMongoClient::new);
+    }
```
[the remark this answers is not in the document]
stIncMale: There is nothing to do, similarly to https://github.com/mongodb/mongo-java-driver/pull/1929#discussion_r3060011014.

driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/RetryableWritesProseTest.java
```diff
@@ -16,68 +16,88 @@
...
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.md#case-3-test-that-drivers-return-the-correct-error-when-receiving-some-errors-with-nowritesperformed-and-some-without-nowritesperformed">
+     * 6. Test error propagation after encountering multiple errors.
+     * Case 3: Test that drivers return the correct error when receiving some errors with NoWritesPerformed and some without NoWritesPerformed</a>.
+     */
+    @Test
+    @Disabled("TODO-BACKPRESSURE Valentin Enable when implementing JAVA-6055, fails on MongoDB 6.0")
+    void errorPropagationAfterEncounteringMultipleErrorsCase3() throws Exception {
+        com.mongodb.client.RetryableWritesProseTest.errorPropagationAfterEncounteringMultipleErrorsCase3(SyncMongoClient::new);
+    }
```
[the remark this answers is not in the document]
stIncMale: Enabled now rather than later, similarly to https://github.com/mongodb/mongo-java-driver/pull/1929#discussion_r3060064225.

Done in 31562053d93f7cae3d4f5a12b1aec05a31e7969b.

stIncMale:
I agree with AI in https://github.com/mongodb/mongo-java-driver/pull/1929#pullrequestreview-4084826254 - the code looks suspicious. But that does not change the fact that the `TestCommandListener` contract is such that it must be thread-safe. I don't have the bandwidth to investigate that code (it wasn't modified in the PR), so I'll leave it be as is.

nhachicha (approved):

PR #1931, opened 2026-03-31
Fix `com.mongodb.client.FailPoint.enable`

stIncMale:
Given the `FailPoint` API, when `enable` completes abruptly, it must not be the case that the fail point is left enabled.

### DIFF
--- a/driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
@@ -24,6 +24,7 @@ import org.bson.conversions.Bson;
 
 import java.util.Collections;
 
+import static com.mongodb.assertions.Assertions.assertFalse;
 import static com.mongodb.client.Fixture.getMongoClientSettingsBuilder;
 
 public final class FailPoint implements AutoCloseable {
@@ -48,27 +49,10 @@ public final class FailPoint implements AutoCloseable {
                         .hosts(Collections.singletonList(serverAddress)))
                 .build();
         MongoClient client = MongoClients.create(clientSettings);
-        Throwable enableException = null;
-        try {
-            return enable(configureFailPointDoc, client);
-        } catch (Throwable e) {
-            enableException = e;
-            throw e;
-        } finally {
-            if (enableException != null) {
-                try {
-                    disableAndClose(configureFailPointDoc, client);
-                } catch (Throwable closeException) {
-                    enableException.addSuppressed(closeException);
-                }
-            }
+        try (Guard guard = new Guard(configureFailPointDoc, client)) {
+            client.getDatabase("admin").runCommand(configureFailPointDoc);
+            return guard.intoFailPoint();
         }
-    }
-
-    private static FailPoint enable(final BsonDocument configureFailPointDoc, final MongoClient client) {
-        FailPoint result = new FailPoint(configureFailPointDoc, client);
-        client.getDatabase("admin").runCommand(configureFailPointDoc);
-        return result;
     }
 
     @Override
@@ -83,6 +67,39 @@ public final class FailPoint implements AutoCloseable {
                     .append("mode", new BsonString("off")));
         } finally {
             client.close();
+        }
+    }
+
+    private static final class Guard implements AutoCloseable {
+        private final BsonDocument failPointDocument;
+        private final MongoClient client;
+        private boolean consumed;
+
+        Guard(final BsonDocument failPointDocument, final MongoClient client) {
+            this.failPointDocument = failPointDocument;
+            this.client = client;
+            consumed = false;
+        }
+
+        /**
+         * May be invoked at most once.
+         *
+         * @see #close()
+         */
+        FailPoint intoFailPoint() {
+            assertFalse(consumed);
+            consumed = true;
+            return new FailPoint(failPointDocument, client);
+        }
+
+        /**
+         * Invokes {@link #disableAndClose(BsonDocument, MongoClient)} unless {@link #intoFailPoint()} was invoked.
+         */
+        @Override
+        public void close() {
+            if (!consumed) {
+                disableAndClose(failPointDocument, client);
+            }
         }
     }
 }
--- a/driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
@@ -61,12 +61,10 @@ public final class FailPoint implements AutoCloseable {
     }
 
     private static void disableAndClose(final BsonDocument failPointDocument, final MongoClient client) {
-        try {
+        try (MongoClient ignored = client) {
             client.getDatabase("admin").runCommand(new BsonDocument()
                     .append("configureFailPoint", failPointDocument.getString("configureFailPoint"))
                     .append("mode", new BsonString("off")));
-        } finally {
-            client.close();
         }
     }
 
@@ -88,8 +86,9 @@ public final class FailPoint implements AutoCloseable {
          */
         FailPoint intoFailPoint() {
             assertFalse(consumed);
+            FailPoint result = new FailPoint(failPointDocument, client);
             consumed = true;
-            return new FailPoint(failPointDocument, client);
+            return result;
         }
 
         /**
--- a/driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
@@ -48,7 +48,22 @@ public final class FailPoint implements AutoCloseable {
                         .hosts(Collections.singletonList(serverAddress)))
                 .build();
         MongoClient client = MongoClients.create(clientSettings);
-        return enable(configureFailPointDoc, client);
+        RuntimeException enableException = null;
+        try {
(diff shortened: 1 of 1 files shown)


driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
```diff
@@ -48,7 +48,22 @@ public static FailPoint enable(final BsonDocument configureFailPointDoc, final S
...
+            throw e;
+        } finally {
+            if (enableException != null) {
+                try {
+                    disableAndClose(configureFailPointDoc, client);
+                } catch (RuntimeException closeException) {
+                    enableException.addSuppressed(closeException);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
stIncMale: Rejected.

That is one way of doing things, it's not unreasonable. Another way is to follow the documentation of `Error`: [_"An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch."_](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/lang/Error.html) I prefer the latter, which is why I created https://jira.mongodb.org/browse/JAVA-6106 in the past.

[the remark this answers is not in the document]
stIncMale: Done in be09603bf12cb3d012341516bd29dc120cee19b3.

Actually, I changed my mind. Even when we do https://jira.mongodb.org/browse/JAVA-6106, we should still make the best effort to make sure there is no dangling fail points left in the database.

driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
```diff
@@ -48,7 +48,22 @@ public static FailPoint enable(final BsonDocument configureFailPointDoc, final S
...
+        } finally {
+            if (enableException != null) {
+                try {
+                    disableAndClose(configureFailPointDoc, client);
+                } catch (Throwable closeException) {
+                    enableException.addSuppressed(closeException);
+                }
+            }
+        }
+
```
nhachicha: [claude-review]
[nit]
remove extra space

driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
```diff
@@ -48,7 +48,22 @@ public static FailPoint enable(final BsonDocument configureFailPointDoc, final S
                         .hosts(Collections.singletonList(serverAddress)))
                 .build();
         MongoClient client = MongoClients.create(clientSettings);
-        return enable(configureFailPointDoc, client);
+        Throwable enableException = null;
```
nhachicha: [claude-review]
[nit]
Suggestion: Simpler pattern using a success flag: The catch-rethrow-finally pattern works but is slightly verbose. An alternative:
```Java
  MongoClient client = MongoClients.create(clientSettings);                                                                                                                                                        
  boolean success = false;                                                                                                                                                                                         
  try {                                                                                                                                                                                                            
      FailPoint result = enable(configureFailPointDoc, client);                                                                                                                                                    
      success = true;                                                                                                                                                                                              
      return result;                                                                                                                                                                                               
  } finally {                                                                                                                                                                                                      
      if (!success) {                                                                                                                                                                                              
          disableAndClose(configureFailPointDoc, client);                                                                                                                                                          
      }                                                                                                                                                                                                            
  }
```

This loses the `addSuppressed` behavior but is more concise. The current approach is fine though — `addSuppressed` is marginally better for diagnostics.

nhachicha (dismissed):
Nice 👍

driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
stIncMale: Done in 5ff93477fcb483ef6dff24204c6a37354fe08b0d.

driver-sync/src/test/functional/com/mongodb/client/FailPoint.java
stIncMale: Thought-provoking.

The pattern I used is nothing extraordinary. It is exactly what the Java SE language specification describes for the [`try`-with-resources statement](https://docs.oracle.com/javase/specs/jls/se26/html/jls-14.html#jls-14.20.3.1). I don't want to loose the behavior, but I also like the idea of simplicity. Maybe we may have our cake and eat it too with a bit of an internal abstraction.

@nhachicha, please take a look at 20d6928af9fb2d3eac0f1cb32b8e578be7ab2543 and 86ad60d03a9e9839212254715f39036b3a7036c1.

nhachicha:
## Code Review Summary: PR #1931 — Fix `com.mongodb.client.FailPoint.enable`

**Author:** @stIncMale | **Branch:** `fixFailPoint` → `main` | **File:** `FailPoint.java` (test utility)

### What Changed

The PR fixes a resource leak in `FailPoint.enable`: if the `runCommand` call to enable a fail point threw an exception, the `MongoClient` was never closed and the fail point could be left enabled on the server.

The fix introduces a `Guard` inner class that uses try-with-resources to guarantee cleanup on all failure paths. On success, `Guard.intoFailPoint()` transfers ownership of the `MongoClient` to the returned `FailPoint`. On failure, `Guard.close()` disables the fail point and closes the client. A shared `disableAndClose` static method eliminates duplication between `FailPoint.close()` and `Guard.close()`.

### What Was Done Well

- **Guard pattern is clean and idiomatic** — the ownership-transfer via `intoFailPoint()` is a well-suited idiom for this problem
- **DRY improvement** — `disableAndClose` is extracted and reused by both `FailPoint.close()` and `Guard.close()`
- **Defensive programming** — `assertFalse(consumed)` in `intoFailPoint()` guards against accidental double-consumption

### Review Findings

| Severity | Finding |
|----------|---------|
| 🟡 Important | **`Guard.close()` attempts to disable a fail point that was never enabled.** If `runCommand` (the enable command) throws, the fail point was never activated on the server. Yet `Guard.close()` still calls `disableAndClose`, which sends a `configureFailPoint: "off"` command that will likely also fail, potentially suppressing the original exception. Consider only closing the `MongoClient` (without the disable command) when the enable was never successful. |
| 🟢 Nit | `consumed = false` in the `Guard` constructor is redundant — `boolean` fields default to `false` in Java. |
| 🟢 Nit | `try (MongoClient ignored = client)` is a valid pattern but may benefit from a brief inline comment for readability. |
| 🟢 Nit | The field name `consumed` could be more descriptive — e.g., `ownershipTransferred` — to better convey that `intoFailPoint()` transfers resource ownership. |
| 💡 Suggestion | `assertFalse(consumed)` is defensive for a `private static final` class called from exactly one place. Fine as-is for test infrastructure, but worth noting it can never trigger in practice. |

### Verdict

**Comment** — The change correctly addresses the stated goal and is well-structured. The one substantive discussion point (unnecessary disable-on-failure path) is worth considering but is not blocking, since the behavior is safe and this is best-effort test infrastructure cleanup.

stIncMale:
@nhachicha, following are the responses to the [AI review findings](https://github.com/mongodb/mongo-java-driver/pull/1931#issuecomment-4205255051)

TL;DR, there is nothing actionable in this batch:

### Review Findings

| Severity | Finding | Response |
|----------|---------|---------|
| 🟡 Important | **`Guard.close()` attempts to disable a fail point that was never enabled.** If `runCommand` (the enable command) throws, the fail point was never activated on the server. Yet `Guard.close()` still calls `disableAndClose`, which sends a `configureFailPoint: "off"` command that will likely also fail, potentially suppressing the original exception. Consider only closing the `MongoClient` (without the disable command) when the enable was never successful. | The assumption _"If `runCommand` (the enable command) throws, the fail point was never activated on the server"_ is incorrect, which makes the whole finding invalid. | 
| 🟢 Nit | `consumed = false` in the `Guard` constructor is redundant — `boolean` fields default to `false` in Java. | Yes, I know. This hasn't been a concern, see `ServerDescription.Builder.minWireVersion/maxWireVersion/cryptd`, `BsonDocumentMarkableIterator.marking`, for example. @nhachicha, if you want to propose us to avoid explicitly specifying default values where we don't have to, please do, and we can discuss that as a team. |
| 🟢 Nit | `try (MongoClient ignored = client)` is a valid pattern but may benefit from a brief inline comment for readability. | This has become a common enough pattern in the codebase, that it no longer needs any comments. See [here](https://github.com/mongodb/mongo-java-driver/blame/main/driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java#L102-L105) for example. |
| 🟢 Nit | The field name `consumed` could be more descriptive — e.g., `ownershipTransferred` — to better convey that `intoFailPoint()` transfers resource ownership. | I am not sure the proposed alternative is obviously better. If I had `ownershipTransferred`, I am sure there were an AI, maybe the same AI, that would have proposed `consumed` or something similar. @nhachicha, if you yourself think that the proposed name is better, please let me know. |
| 💡 Suggestion | `assertFalse(consumed)` is defensive for a `private static final` class called from exactly one place. Fine as-is for test infrastructure, but worth noting it can never trigger in practice. | Yes, assertions must never trigger in practice, because that would mean there is a bug in the driver code. Assertions exist to tersely communicate information to a reader and as guards against bugs. |

nhachicha (approved):

PR #1935, opened 2026-04-09
Add Windows CI testing

vbabanin:
### Changes in this PR:
- Add Windows 2022 Evergreen build variants for server 8.2 across all topologies to enable Windows CI coverage
- Add EnvironmentProvider class to centralize all System.getenv() calls, replacing the reflection-based env var override hack that fails on Windows due to JDK's ProcessEnvironment dual-map architecture
- Fix SSL keystore/truststore paths in Evergreen scripts to use native Windows paths via cygpath
- Fix JSON test file path separator handling and mongocryptd temp directory for Windows compatibility
- Skip Unix-domain-socket test on Windows

JAVA-6057

### Checklist
- [x] Jira ticket linked in the description
- [x] AI-generated code reviewed by author antagonistically
- [x] AI review completed on draft PR (Copilot or/and Claude)
- [x] GitHub comments added for complex/confusing areas.
- [x] PR is appropriately scoped (drive-by changes are allowed)
- [x] Manually triggered test variants not covered by PR automation, if applicable (e.g., Atlas, OIDC)

### AI Usage & Effectiveness
**What Claude (Opus 4.6) did (mostly boilerplate/mechanical refactoring):**
- Wrote the `EnvironmentProvider` class (after the author, without Claude, designed a detailed plan using the lambda-based approach)
- Migrated all `System.getenv()` call sites to `EnvironmentProvider.getEnv()` (`FaasEnvironment`, `OidcAuthenticator`, `BuiltInAwsCredentialSupplier`, `TracingManager`, `MicrometerTracer`)
- Removed `FaasEnvironmentAccessor` and the reflection hack from `AbstractMicrometerProseTest`
- Updated the `WithWrapper` class to use the new pattern
- Fixed the `mongocryptd /tmp/` path for Windows
- [Experiment] Used Evergreen MCP via Claude as an experiment to retrieve logs and provide a quick summary of the failure. Results: Claude was slower to retrieve those logs than manually downloading them. However, Claude was useful for analyzing the failures sometimes.

**What Claude (Opus 4.6) suggestions were rejected:**
- Rejected multiple Claude proposals for `EnvironmentProvider` (`volatile`, `ConcurrentHashMap`, `override(Map)` constructor)

**What was done manually:**
- Submitted Evergreen patches and iterated on failures during troubleshooting
- Designed the `EnvironmentProvider` approach (lambda field, no `volatile`, no `ConcurrentHashMap`)
- Updated shell scripts
- Chose server version 8.2 for Windows (based on `SERVER-116018`)
- Unit tests


driver-core/src/main/com/mongodb/internal/EnvironmentProvider.java
```diff
@@ -0,0 +1,68 @@
...
+
+        public EnvironmentOverride set(final String key, @Nullable final String value) {
+            overrides.put(key, value);
+            return this;
+        }
+
+        @Override
+        public void close() {
+            envLookup = original;
+        }
```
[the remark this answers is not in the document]
vbabanin: Rejected. This thread safety does not matter here because `envOverride()` is only called in tests, and:                                                                                                                                
  1. In production, `envLookup` is never reassigned - it stays `System::getenv` forever.                                                                                                                              
  2. In tests, execution is single-threaded, and the override is set before the code under test runs.

driver-core/src/test/unit/com/mongodb/internal/EnvironmentProviderTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    void shouldInterceptGetEnvWhenOverridden() {
+        try (EnvironmentProvider.EnvironmentOverride env = EnvironmentProvider.envOverride()) {
+            env.set("MY_TEST_VAR", "hello");
+            assertEquals("hello", EnvironmentProvider.getEnv("MY_TEST_VAR"));
+        }
+
+        assertNull(EnvironmentProvider.getEnv("MY_TEST_VAR"));
+    }
```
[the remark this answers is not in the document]
vbabanin: Rejected. The probability of `MY_TEST_VAR` existence  any CI environment is negligible.

driver-core/src/test/unit/com/mongodb/internal/EnvironmentProviderTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+            try (EnvironmentProvider.EnvironmentOverride inner = EnvironmentProvider.envOverride()) {
+                inner.set("VAR_B", "inner");
+                assertEquals("outer", EnvironmentProvider.getEnv("VAR_A"));
+                assertEquals("inner", EnvironmentProvider.getEnv("VAR_B"));
+            }
+
+            assertEquals("outer", EnvironmentProvider.getEnv("VAR_A"));
+            assertNull(EnvironmentProvider.getEnv("VAR_B"));
+        }
```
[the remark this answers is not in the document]
vbabanin: Rejected. The probability of VAR_B existence any CI environment is negligible.

driver-sync/src/test/functional/com/mongodb/client/AbstractMicrometerProseTest.java
```diff
@@ -62,23 +61,18 @@ public abstract class AbstractMicrometerProseTest {
...
     }
 
     @AfterAll
-    static void afterAll() throws Exception {
-        // restore original env var values
-        setEnv(ENV_OBSERVABILITY_ENABLED, previousEnvVarMdbTracingEnabled);
-        setEnv(ENV_OBSERVABILITY_QUERY_TEXT_MAX_LENGTH, previousEnvVarMdbQueryTextLength);
+    static void afterAll() {
+        environmentOverride.close();
     }
```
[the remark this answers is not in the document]
vbabanin: Rejected. Even if there were tests being executed in parallel - the narrower scope would not eliminate the problem. It would eliminate it only if the tests were executed concurrently without overlapping, but could be executed from different classes in undefined order. However, it does not happen in practice, they are executed in a single-thread class by class.

driver-core/src/main/com/mongodb/internal/EnvironmentProvider.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.mongodb.lang.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.UnaryOperator;
+
+/**
+ * Centralized access to environment variables. All production code should use
+ * this class instead of calling {@link System#getenv(String)} directly.
```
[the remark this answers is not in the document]
vbabanin: Rejected. LambdaTestApp is not a production code.

P.S This is an interesting inference - copilot did not understand that `LambdaTestApp` is not a production code despite having "Test" in it's name.

.evergreen/.evg.yml
```diff
@@ -2540,6 +2565,15 @@ buildvariants:
     tasks:
       - name: "csfle-tests-with-mongocryptd-task"
 
+  - matrix_name: "csfle-tests-with-mongocryptd-windows"
+    matrix_spec: { os: "windows",
+                   version: [ "8.0", "latest" ],
+                   topology: [ "replicaset" ] }
```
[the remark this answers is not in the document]
vbabanin: This particular task should be 8.0 because we do not execute tests that run crypt shared library, which has a bug fix only in 8.2.

.evergreen/setup-env.bash
```diff
@@ -1,9 +1,16 @@
...
+  export JDK21="/cygdrive/c/java/jdk21"
+else
+  export JDK8="/opt/java/jdk8"
+  export JDK11="/opt/java/jdk11"
+  export JDK17="/opt/java/jdk17"
+  export JDK21="/opt/java/jdk21"
+fi
 # note that `JDK21_GRAALVM` is used in `run-graalvm-native-image-app.sh`
 # by dynamically constructing the variable name
 export JDK21_GRAALVM="/opt/java/jdk21-graalce"
```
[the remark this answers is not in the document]
vbabanin: Rejected. Graal is not installed on Windows machines.

.evergreen/.evg.yml
```diff
@@ -1740,6 +1740,12 @@ axes:
         display_name: "8.0"
         variables:
           VERSION: "8.0"
+      # 8.2 is used solely for Windows testing. MongoDB 8.0 binaries are affected by SERVER-116018 on Windows,
+      # and the fix is only available starting from 8.2.
+      - id: "8.2"
+        display_name: "8.2"
+        variables:
+          VERSION: "8.2"
```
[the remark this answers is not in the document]
vbabanin: Good catch! I expanded all * to a list, because there is no way to hardcode without global axis.

.evergreen/.evg.yml
```diff
@@ -2403,6 +2412,22 @@ buildvariants:
...
 
+  - matrix_name: "tests-jdk-secure-windows"
+    matrix_spec: { auth: "auth", ssl: "ssl", jdk: "jdk17",
+                   version: [ "8.2" ],
+                   topology: "*", os: "windows" }
+    display_name: "${os}: ${version} ${topology} ${auth} ${ssl} ${jdk}"
+    tags: [ "tests-variant" ]
+    tasks:
+      - name: "test-sync-task"
+        exec_timeout_secs: 7200
```
vbabanin, line 2423: Tests run longer on `Windows` ~1h 30min - 1h 40 min, so it is bumped to 2 hours.

driver-core/src/main/com/mongodb/internal/EnvironmentProvider.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * <p>Tests can override values via {@link #envOverride()}.</p>
+ */
+public final class EnvironmentProvider {
```
vbabanin, line 32: ### Why this PR introduces **EnvironmentProvider** (and consolidates env handling)

We previously used a reflection-based test helper in `AbstractMucromiterProseTests` to override environment variables. That approach broke on Windows due to differences in the JDK’s  **ProcessEnvironment** implementation.

### What worked on Linux/macOS

**ProcessEnvironment** effectively exposes a single backing map. The reflection helper mutates that map, and `System.getenv() / System.getenv("KEY")` read from the same data, thus overrides are visible.

### What breaks on Windows
The JDK maintains two maps:

- theEnvironment (HashMap): backs System.getenv() (returns the full map)
- theCaseInsensitiveEnvironment (TreeMap, without case-insensitive comparator): backs System.getenv(String key) (single-key lookup)

The reflection helper only updated `theEnvironment`. Our production code uses `System.getenv("OBSERVABILITY_MONGODB_ENABLED")`, which reads from `theCaseInsensitiveEnvironment`, so the override was invisible on Windows, thus Micrometer tests failed.

### What this PR changes
Introduces **EnvironmentProvider** to avoid reflection/JDK internals entirely. Production code calls `EnvironmentProvider.getEnv()` instead of `System.getenv()`.

Tests override the provider via `envOverride(...)` (swapping the underlying supplier/lambda), which behaves consistently across platforms.

**EnviromentOverride** is `AutoCloseable`, so tests can use try-with-resources and avoid boilerplate while keeping overrides scoped and explicit.

.evergreen/.evg.yml
```diff
@@ -2332,7 +2341,7 @@ buildvariants:
     #
 
   - matrix_name: "tests-zlib-compression"
-    matrix_spec: { compressor: "zlib", auth: "noauth", ssl: "nossl", jdk: "jdk8", version: "*", topology: "standalone", os: "linux" }
+    matrix_spec: { compressor: "zlib", auth: "noauth", ssl: "nossl", jdk: "jdk8", version: [ "4.2", "4.4", "5.0", "6.0", "7.0", "8.0", "latest" ], topology: "standalone", os: "linux" }
```
strogiyotec, line 2344: note for myself
We need to list all versions here so that 8.2 is not included in linux builds

.evergreen/run-csfle-tests-with-mongocryptd.sh
```diff
@@ -39,8 +39,16 @@ provision_ssl () {
   cp ${JAVA_HOME}/lib/security/cacerts mongo-truststore
   ${JAVA_HOME}/bin/keytool -importcert -trustcacerts -file ${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem -keystore mongo-truststore -storepass changeit -storetype JKS -noprompt
 
+  # Use native paths on Windows (cygwin paths like /cygdrive/c/... are not understood by the JDK)
+  local CURRENT_DIR
+  if [ "Windows_NT" == "$OS" ]; then
```
strogiyotec, line 44: didn't know windows build uses cygwin instead of powershell 
good catch

bson/src/test/unit/util/JsonPoweredTestHelper.java
```diff
@@ -86,7 +86,8 @@ public static List<BsonDocument> getTestDocuments(final String resourcePath) {
                     public FileVisitResult visitFile(final Path filePath, final BasicFileAttributes attrs) throws IOException {
                         if (filePath.toString().endsWith(".json")) {
                             if (fileSystem == null) {
-                                files.add(getTestDocumentWithMetaData(filePath.toString().substring(filePath.toString().lastIndexOf(resourcePath))));
+                                String filePathStr = filePath.toString().replace('\\', '/');
```
strogiyotec, line 90: nit: would be great to have a comment here explaining on why this replace is needed
I didn't understand first time , the LLM can explain it fast so if you think it's not important let's keep it this way

driver-core/src/main/com/mongodb/internal/EnvironmentProvider.java
```diff
@@ -0,0 +1,68 @@
...
+    public static String getEnv(final String key) {
+        return envLookup.apply(key);
+    }
+
+    /** Exists only for testing **/
+    public static EnvironmentOverride envOverride() {
+        return new EnvironmentOverride();
+    }
+
+    public static final class EnvironmentOverride implements AutoCloseable {
```
strogiyotec, line 53: should we add @NotThreadSafe annotation here as all the tests must release the map and can't run in parallel ? 
By `can't run in parallel` I mean one test can override the same key-value used by other test

driver-core/src/main/com/mongodb/internal/EnvironmentProvider.java
```diff
@@ -0,0 +1,68 @@
...
+        private final Map<String, String> overrides = new HashMap<>();
+        private final UnaryOperator<String> original;
+
+        private EnvironmentOverride() {
+            original = envLookup;
```
strogiyotec: took me a while to understand, I didn't notice that `envLookup` is static, can we do
```
            original = EnvironmentProvider.envLookup;
```
instead ? Just my preference , feel free to ignore

driver-core/src/test/functional/com/mongodb/internal/connection/ClientMetadataTest.java
```diff
@@ -234,7 +234,7 @@ void testDockerMetadataIncluded() {
     @Test
     void testDockerAndKubernetesMetadataIncluded() {
         try (MockedStatic<Files> pathsMockedStatic = Mockito.mockStatic(Files.class)) {
-            Path path = Paths.get(File.separator + "/.dockerenv");
+            Path path = Paths.get(File.separator + ".dockerenv");
```
strogiyotec, line 237: thanks, looks like existing bug

strogiyotec (changes requested):

driver-core/src/test/functional/com/mongodb/internal/connection/SocketStreamHelperSpecification.groovy
```diff
@@ -43,6 +43,7 @@ import static java.util.concurrent.TimeUnit.SECONDS
 
 class SocketStreamHelperSpecification extends Specification {
 
+    @IgnoreIf({ System.getProperty('os.name').toLowerCase().contains('windows') })
```
strogiyotec, line 47: could we add a comment here that TCP_KEEPCOUNT is linux specific and doesn't exist in windows (I didn't know why test was excluded so had to read the logic first , feel free to ignore it )

vbabanin: Done in [8fad8cd](https://github.com/mongodb/mongo-java-driver/pull/1935/commits/8fad8cd4e36e00b3d194c96c6a21644fb1a64318).

driver-core/src/main/com/mongodb/internal/EnvironmentProvider.java
vbabanin: Done in [8fad8cd](https://github.com/mongodb/mongo-java-driver/pull/1935/commits/8fad8cd4e36e00b3d194c96c6a21644fb1a64318)

driver-core/src/main/com/mongodb/internal/EnvironmentProvider.java
vbabanin: Done in [8fad8cd](https://github.com/mongodb/mongo-java-driver/pull/1935/commits/8fad8cd4e36e00b3d194c96c6a21644fb1a64318)

bson/src/test/unit/util/JsonPoweredTestHelper.java
vbabanin: Done in [8fad8cd](https://github.com/mongodb/mongo-java-driver/pull/1935/commits/8fad8cd4e36e00b3d194c96c6a21644fb1a64318)

strogiyotec:
hi @vbabanin  let's run the formatter , looks like the checkstyle stage in evergreen has failed

strogiyotec (approved):
LGTM

PR #1936, opened 2026-04-09
JAVA-6099 Add new fields for Auto embedding

strogiyotec:
[JAVA-6099 ](https://jira.mongodb.org/browse/JAVA-6099)
The doc in JIRA list new fields that must be supported by the drivers
but because the Search index doesn't use builders, java driver makes it possible to use new fields without changing API
As part of this PR I added a new unit test to disabled test class and enabled it locally for testing
The reason it's disabled is 
1. This class was created to add support for queryText index that was only available in custom docker image ,right now it's available in community edition so we have to adjust evergreen first
2. New Search Index fields are only available in Atlas and are not in public preview yet, I had to use specific cluster in Atlas QA environment 
For other details please check the JIRA ticket [comment](https://jira.mongodb.org/browse/JAVA-6099?focusedCommentId=8294606&focusedId=8294606&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-8294606)


driver-sync/src/test/functional/com/mongodb/client/vector/AbstractAutomatedEmbeddingVectorSearchFunctionalTest.java
```diff
@@ -210,6 +214,200 @@ private void insertDocumentsForEmbedding() {
...
+        mongoClient.getDatabase(getDatabaseName()).createCollection(getCollectionName());
+        createAutoEmbeddingIndex("voyage-4-large");
+        TimeUnit.SECONDS.sleep(2L);
+        insertDocumentsForEmbedding();
+        TimeUnit.SECONDS.sleep(2L);
+
+        List<Bson> pipeline = asList(
+                vectorSearch(
+                        fieldPath(FIELD_SEARCH_PATH),
+                        textQuery("movies about love").model("voyage-4-large"),
```
vbabanin: This doesn’t actually exercise an override. The index is created with voyage-4-large, and the query also specifies `.model("voyage-4-large")` - same model. An override would mean using a different compatible model at query time.

The spec calls this out for the `voyage-4` family (interoperability within the same family). To test override, we should create the index with `voyage-4-large` and run the query with a different compatible model (e.g., `voyage-4` or `voyage-4-lite`).

driver-sync/src/test/functional/com/mongodb/client/vector/AbstractAutomatedEmbeddingVectorSearchFunctionalTest.java
vbabanin: I missed this in the previous review (#1862, https://github.com/mongodb/mongo-java-driver/pull/1862/changes
), but i think `AbstractAutomatedEmbeddingVectorSearchFunctionalTest` should be relocated to `com.mongodb.client.model.search` (where `AggregatesBinaryVectorSearchIntegrationTest` lives).

Both tests exercise the same surface area: `$vectorSearch` aggregation stage behavior - one with binary vectors, the other with auto-embedded text queries. They use the same API (`Aggregates.vectorSearch()`, `VectorSearchOptions`, `VectorSearchQuery`, `SearchPath.fieldPath()`), all of which live under `com.mongodb.client.model.search`, so co-locating the tests with that API keeps the package structure consistent.

`com.mongodb.client.vector` should remain focused on BSON vector type coverage, not aggregation-stage behavior, which is what is done in `AbstractBinaryVectorFunctionalTest` - this test is specifically for vector BSON type.

driver-sync/src/test/functional/com/mongodb/client/vector/AbstractAutomatedEmbeddingVectorSearchFunctionalTest.java
```diff
@@ -210,6 +214,200 @@ private void insertDocumentsForEmbedding() {
...
+                () -> documentCollection.createSearchIndexes(Collections.singletonList(indexModel)),
+                "Expected index creation to fail because autoEmbed field cannot be used as a filter field"
+        );
+    }
+
+    @Test
+    @DisplayName("should create auto embedding index and run query with model override")
+    void shouldCreateAutoEmbeddingIndexAndRunQueryWithModelOverride() throws InterruptedException {
+        mongoClient.getDatabase(getDatabaseName()).createCollection(getCollectionName());
+        createAutoEmbeddingIndex("voyage-4-large");
```
vbabanin: This comment is related to the suggestion in: https://github.com/mongodb/mongo-java-driver/pull/1936#discussion_r3076771291

Currently, these tests mix two concerns: index setup and running a query. When a test covers both, a failure doesn’t tell us - without digging into logs/code - whether the setup broke (index creation) or the operation broke (query execution).

This applies to:
`shouldCreateAutoEmbeddingIndexAndRunQueryWithModelOverride`
`shouldCreateAutoEmbeddingIndexAndRunExactVectorSearchQuery`
`shouldCreateAutoEmbeddingIndexAndRunVectorSearchQuery`

The `createAutoEmbeddingIndex("voyage-4-large")` is the exact same setup duplicated across all three - it's not the `SUT`, it's a prerequisite: 

```
        createAutoEmbeddingIndex("voyage-4-large");
        TimeUnit.SECONDS.sleep(2L);
        insertDocumentsForEmbedding();
        TimeUnit.SECONDS.sleep(2L);
```

I think we should follow the pattern in `AggregatesBinaryVectorSearchIntegrationTest`: create the index and shared test data once, and have each test method exercise a single query concern (approximate, exact, model override), e.g. via `@BeforeAll` (and later it will be a pre-created index on the QA cluster).

The same pattern we did before: 
- AggregatesBinaryVectorSearchIntegrationTest (38fc561c) - index setup in the `BeforeAll`.
- Migration to QA cluster.

driver-sync/src/test/functional/com/mongodb/client/vector/AbstractAutomatedEmbeddingVectorSearchFunctionalTest.java
```diff
@@ -210,6 +214,200 @@ private void insertDocumentsForEmbedding() {
...
+                        )),
+                SearchIndexType.vectorSearch()
+        );
+        List<String> result = documentCollection.createSearchIndexes(Collections.singletonList(indexModel));
+        Assertions.assertFalse(result.isEmpty());
+    }
+
+    @Test
+    @DisplayName("should create auto embedding index with filter field")
+    void shouldCreateAutoEmbeddingIndexWithFilterField() {
```
vbabanin: Index creation and vector search queries are separate concerns, and the codebase currently reflects this split:

Index management tests (index creation, deletion etc) live in `AbstractAtlasSearchIndexManagementProseTest`
Vector search query tests live in `AggregatesBinaryVectorSearchIntegrationTest` / `AbstractAutomatedEmbeddingVectorSearchFunctionalTest`

I suggest we move the index creation tests in this PR to `AbstractAtlasSearchIndexManagementProseTest`:

- `shouldCreateAutoEmbeddingIndexWithQuantization`
- `shouldCreateAutoEmbeddingIndexWithCustomNumDimensions`
- `shouldCreateAutoEmbeddingIndexWithFilterField`
- `shouldFailWhenMixingVectorAndAutoEmbedTypes`
- `shouldFailWhenDuplicatePathsAreUsed`
- `shouldFailWhenAutoEmbedFieldUsedAsFilterField`

They aren’t spec-defined prose tests today, but colocating this additional (non-prose) coverage alongside the prose tests is still a good fit: it extends index-management coverage in the same area, and if/when the spec adds auto-embedding prose tests, they’d naturally land there as well. The C# driver took the same approach by adding auto-embed index creation tests under their Atlas Search index management tests ([mongo-csharp-driver PR #1900](https://github.com/mongodb/mongo-csharp-driver/pull/1900/files)).

The class would stay then quite simple and would have only:
  1. shouldExecuteVectorSearchQuery (existing, from previous PR)                                                                 
  2. shoulduExecuteVectorSearchWithModelOverride (new, added in this PR)                                                     
  3. shouldExecuteExactVectorSearchQuery (new, added in this PR)

driver-sync/src/test/functional/com/mongodb/client/vector/AbstractAutomatedEmbeddingVectorSearchFunctionalTest.java
strogiyotec: was addressed [here ](https://github.com/mongodb/mongo-java-driver/pull/1936/changes#diff-18ddb2481926113f83d6490e4bd557ba8aabca7483ba8574b0bd41f7b219f849R134)

driver-sync/src/test/functional/com/mongodb/client/vector/AbstractAutomatedEmbeddingVectorSearchFunctionalTest.java
strogiyotec: class was moved to a different package [here](https://github.com/mongodb/mongo-java-driver/pull/1936/changes#diff-ab924f3c9f5633f5e6632139d023e1cdc4551cc01a9c85f0d383b7c9c59ad5fbR17)

driver-sync/src/test/functional/com/mongodb/client/vector/AbstractAutomatedEmbeddingVectorSearchFunctionalTest.java
strogiyotec: Addressed it [here ](https://github.com/mongodb/mongo-java-driver/pull/1936/changes#diff-18ddb2481926113f83d6490e4bd557ba8aabca7483ba8574b0bd41f7b219f849R72)
Class is annotated as TEST PER CLASS so that I can access non static `mongoClient` within the test

driver-sync/src/test/functional/com/mongodb/client/vector/AbstractAutomatedEmbeddingVectorSearchFunctionalTest.java
strogiyotec: Addressed it [here](https://github.com/mongodb/mongo-java-driver/pull/1936/changes#diff-cdd0d614e5d6ca573bd3830974cc8bbcaca9a78af116482d1455cec1226f8ba6R291)
Right now these methods are skipped as atlas search is not in public preview

rozza (dismissed):
LGTM

vbabanin (dismissed):
LGTM!

driver-sync/src/test/functional/com/mongodb/client/model/search/AbstractAutoEmbeddingVectorSearchFunctionalTest.java
```diff
@@ -44,36 +44,33 @@
...
 
 /**
- * The test cases were borrowed from
- * <a href="https://github.com/mongodb-labs/ai-ml-pipeline-testing/blob/main/.evergreen/mongodb-community-search/self_test.py"> this repository</a>.
+ * Tests for auto-embedding vector search queries.
+ * Index creation and validation tests are in {@link com.mongodb.client.AbstractAtlasSearchIndexManagementProseTest}.
  */
-public abstract class AbstractAutomatedEmbeddingVectorSearchFunctionalTest extends OperationTest {
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public abstract class AbstractAutoEmbeddingVectorSearchFunctionalTest extends OperationTest {
```
vbabanin: Do we need to extend `OperationTest`? 

OperationTest has a `@BeforeEach` that calls `CollectionHelper.drop(getNamespace())`. With `@TestInstance(PER_CLASS)`, JUnit 5 still invokes inherited `@BeforeEach` before each test method - so when `Assumptions.assumeTrue(false)` is removed `(TODO-JAVA-6059)`, the collection will be dropped before each test.

strogiyotec: you are right , addressed it [here ](https://github.com/mongodb/mongo-java-driver/pull/1936/changes/95cb82d06bae6020bd4ed883ad4fe6d6eb32c1e7)

vbabanin (approved):

PR #1938, opened 2026-04-11
Change CSOT exception hierarchy

vbabanin:
**MongoTimeoutException** is stable public API documented as "waiting for a server or connection to become available," implying the operation was not executed. **MongoOperationTimeoutException** does not provide that guarantee: the operation may have been sent to the server and may have executed. Keeping it as a subclass of **MongoTimeoutException** is therefore semantically incorrect and may mislead consumers that catch **MongoTimeoutException** into assuming no operation was executed.

## Changes:
- Make MongoOperationTimeoutException extend MongoClientException instead of MongoTimeoutException to avoid breaking the semantic guarantee that MongoTimeoutException implies the operation was not executed
- Handle MongoOperationTimeoutException explicitly in instanceof MongoTimeoutException checks to preserve existing behavior for transaction labeling, connection pool events, and logging

## Options considered:

- **extends MongoClientException** (what is proposed in this PR) - sibling to MongoTimeoutException. Both are client-side failures, no shared timeout semantics.                                                                   
- **extends MongoException** - too broad. **MongoException** is the root for everything including server errors. **MongoOperationTimeoutException** is a client-side concern.                                              
-  New intermediate class like **MongoOperationException** - adds a class just for one subclass. Unnecessary abstraction.                                                                                           
- **extends RuntimeException** directly - breaks out of the **MongoException** hierarchy entirely. Users catching **MongoException** would miss it.       

JAVA-5438

### DIFF
--- a/driver-core/src/main/com/mongodb/MongoOperationTimeoutException.java
+++ b/driver-core/src/main/com/mongodb/MongoOperationTimeoutException.java
@@ -40,7 +40,7 @@ import java.util.concurrent.TimeUnit;
  * @since 5.2
  */
 @Alpha(Reason.CLIENT)
-public final class MongoOperationTimeoutException extends MongoTimeoutException {
+public final class MongoOperationTimeoutException extends MongoClientException {
     private static final long serialVersionUID = 1L;
 
     /**
--- a/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/BaseCluster.java
@@ -427,7 +427,7 @@ abstract class BaseCluster implements Cluster {
                 "Timed out while waiting for a server that matches %s. Client view of cluster state is %s",
                 serverSelector, clusterDescription.getShortDescription());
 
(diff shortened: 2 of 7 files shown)


driver-core/src/main/com/mongodb/MongoOperationTimeoutException.java
```diff
@@ -40,7 +40,7 @@
  * @since 5.2
  */
 @Alpha(Reason.CLIENT)
-public final class MongoOperationTimeoutException extends MongoTimeoutException {
+public final class MongoOperationTimeoutException extends MongoClientException {
```
stIncMale, line 43: We also need to rollback the change done to `MongoClientException.java` in https://github.com/mongodb/mongo-java-driver/commit/2b00a83268ef4e13dac205247bff6b0814eaad65 (and only to that file, the rest of the commit changes should stay in place).

stIncMale (changes requested):

rozza (changes requested):
A couple of tests need updating to use `MongoOperationTimeoutException`:

* `BaseClusterSpecification`
* `ClientSideOperationTimeoutProseTest`

PR #1953, opened 2026-04-29
Retry KMS requests on transient errors

rozza:
Add libmongocrypt CAPI bindings for KMS retry support and wire retry logic through the sync and reactive driver stacks. Transient KMS HTTP and network errors are retried with backoff delays managed by libmongocrypt; retry is enabled unconditionally.

- Add native bindings: mongocrypt_setopt_retry_kms, mongocrypt_kms_ctx_usleep, mongocrypt_kms_ctx_feed_with_retry, mongocrypt_kms_ctx_fail
- Add sleepMicroseconds(), feedAndRetry(), fail() to MongoKeyDecryptor
- Enable KMS retry unconditionally in MongoCryptImpl
- Rewrite sync Crypt.decryptKey() with retry loop, timeout-aware
- Add retry logic to reactive KeyManagementService.decryptKey()
- Fix TlsChannelImpl.read() to preserve bytes delivered alongside close_notify (already fixed upstream in marianobarrios/tls-channel)
- Add spec Section 24 KMS retry integration tests (sync + reactive)
- Add Evergreen CI task for KMS retry tests

JAVA-5391


.evergreen/run-kms-retry-tests.sh
```diff
@@ -0,0 +1,44 @@
...
+
+./gradlew --stacktrace --info ${GRADLE_EXTRA_VARS} -Dorg.mongodb.test.uri=${MONGODB_URI} \
+    -Dorg.mongodb.test.kms.retry.ca.path="${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem" \
+    driver-sync:cleanTest driver-sync:test --tests ClientSideEncryptionKmsRetryProseTest
+first=$?
+echo $first
+
+./gradlew --stacktrace --info ${GRADLE_EXTRA_VARS} -Dorg.mongodb.test.uri=${MONGODB_URI} \
+    -Dorg.mongodb.test.kms.retry.ca.path="${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem" \
+    driver-reactive-streams:cleanTest driver-reactive-streams:test --tests ClientSideEncryptionKmsRetryProseTest
```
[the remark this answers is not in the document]
rozza: Rejected its scoped to: `driver-reactive-streams:test`

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -486,6 +487,21 @@ public interface mongocrypt_random_fn extends Callback {
...
+     * Opt-into handling the MONGOCRYPT_CTX_NEED_KMS state with retry logic.
+     *
+     * <p>If opted in, KMS requests will include retry information accessible via
+     * {@link #mongocrypt_kms_ctx_usleep}, {@link #mongocrypt_kms_ctx_feed_with_retry},
+     * and {@link #mongocrypt_kms_ctx_fail}.
+     *
+     * @param crypt  The @ref mongocrypt_t object to update
+     * @param enable Whether to enable KMS retry
+     * @return A boolean indicating success. If false, an error status is set.
+     * @since 5.8
```
stIncMale: Why do we specify `@since` on an internal API? (the same applies to `MongoKeyDecryptor`) I fail to see an answer myself.

I see that some existing program elements in `CAPI` have `@since` specified, while some other do not (like `mongocrypt_setopt_enable_multiple_collinfo`, which is relatively new).

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -486,6 +487,21 @@ public interface mongocrypt_random_fn extends Callback {
...
+    /**
+     * Opt-into handling the MONGOCRYPT_CTX_NEED_KMS state with retry logic.
+     *
+     * <p>If opted in, KMS requests will include retry information accessible via
+     * {@link #mongocrypt_kms_ctx_usleep}, {@link #mongocrypt_kms_ctx_feed_with_retry},
+     * and {@link #mongocrypt_kms_ctx_fail}.
+     *
+     * @param crypt  The @ref mongocrypt_t object to update
+     * @param enable Whether to enable KMS retry
+     * @return A boolean indicating success. If false, an error status is set.
```
stIncMale: Let's add the

> Retrieve it with @ref mongocrypt_ctx_status

sentence from the documentation in `mongocrypt.h` to the `@return` text. It is useful, and we seem to have it in the documentation of other methods. Though it should be formatted properly according to the Java documentation comment rules.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -486,6 +487,21 @@ public interface mongocrypt_random_fn extends Callback {
     public static native void
     mongocrypt_setopt_bypass_query_analysis (mongocrypt_t crypt);
 
+    /**
+     * Opt-into handling the MONGOCRYPT_CTX_NEED_KMS state with retry logic.
+     *
+     * <p>If opted in, KMS requests will include retry information accessible via
+     * {@link #mongocrypt_kms_ctx_usleep}, {@link #mongocrypt_kms_ctx_feed_with_retry},
```
stIncMale: - `mongocrypt_kms_ctx_feed_with_retry` was addded later in mongodb/libmongocrypt#1011 (https://jira.mongodb.org/browse/MONGOCRYPT-763).
- The original prose, which JAVA-5391 and this PR are supposed to implement, instructed us to use `mongocrypt_kms_ctx_feed`. The current prose, added in the above libmongocrypt PR, instructs us to use either `mongocrypt_kms_ctx_feed` or `mongocrypt_kms_ctx_feed_with_retry`.

Thus, we do not have to use `mongocrypt_kms_ctx_feed_with_retry`. If we decide to introduce this method and use it, then we should
also
1. fully implement https://jira.mongodb.org/browse/MONGOCRYPT-763 in this PR (it seems, it mostly was implemented?);
   - example: the documentation of `mongocrypt_kms_ctx_fail` was updated in mongodb/libmongocrypt#1011, and that change is not reflected in the current PR.
2. specify JAVA-5772 in addition to JAVA-5391 in the description of the current PR
3. resolve https://jira.mongodb.org/browse/JAVA-5772 (split from https://jira.mongodb.org/browse/MONGOCRYPT-763) once this PR is merged.

I wonder how the PR ended up using a method from a different change, why nothing was explicitly said about that to help orienting a reviewer, and what was the plan for JAVA-5772 given that this PR implements most of what should be done in JAVA-5772.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -1164,6 +1180,47 @@ public interface mongocrypt_random_fn extends Callback {
...
+     * @param kms          The @ref mongocrypt_kms_ctx_t.
+     * @param bytes        The bytes to feed.
+     * @param should_retry Receives whether the driver should retry the KMS request.
+     * @return A boolean indicating success.
+     * @since 5.8
+     */
+    public static native boolean
+    mongocrypt_kms_ctx_feed_with_retry(mongocrypt_kms_ctx_t kms,
+                                       mongocrypt_binary_t bytes,
+                                       ByteByReference should_retry);
```
stIncMale, line 1215: For some reason the documentation of this method in the PR omits parts of the documentation from `mongocrypt.h`. Let's bring them all here, unless there is a good reason to omit any specific part.

driver-core/src/main/com/mongodb/internal/connection/tlschannel/impl/TlsChannelImpl.java
```diff
@@ -236,10 +236,13 @@ public long read(ByteBufferSet dest) throws IOException {
           case NOT_HANDSHAKING:
           case FINISHED:
             UnwrapResult res = readAndUnwrap(Optional.of(dest));
+            bytesToReturn = res.bytesProduced;
             if (res.wasClosed) {
-              return -1;
+              // JAVA-5391: return any bytes produced alongside close_notify; the next read
+              // sees shutdownReceived and returns -1. Fixed in upstream marianobarrios/tls-channel;
+              // this is the minimal patch until the vendored snapshot is refreshed.
+              return bytesToReturn > 0 ? bytesToReturn : -1;
```
stIncMale, line 244: Could you please
- Either explain the problem in this GitHub thread, or link to an existing explanation / bug report.
   - Also explain why the fix is required for the current PR.
- Share a link to the upstream PR/commit with the fix.

I think, given that this change is in a vendored code, we should do it in a separate PR, to avoid complicating the future work of updating the vendored code.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptorImpl.java
```diff
@@ -96,6 +105,29 @@ public void feed(final ByteBuffer bytes) {
...
+    public long sleepMicroseconds() {
+        return mongocrypt_kms_ctx_usleep(wrapped);
+    }
+
+    @Override
+    public boolean feedAndRetry(final ByteBuffer bytes) {
+        try (BinaryHolder binaryHolder = toBinary(bytes)) {
+            // Default 0 means "do not retry"; libmongocrypt writes 1 only when the driver should retry.
+            ByteByReference shouldRetry = new ByteByReference();
+            boolean success = mongocrypt_kms_ctx_feed_with_retry(wrapped, binaryHolder.getBinary(), shouldRetry);
```
stIncMale, line 118: - Let's make this information part of the `should_retry` parameter documentation on the `mongocrypt_kms_ctx_feed_with_retry` method, and remove the comment. Also, "default" seems to be inapplicable here.
- Was this comment generated by AI?

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptorImpl.java
```diff
@@ -96,6 +105,29 @@ public void feed(final ByteBuffer bytes) {
         }
     }
 
+    @Override
+    public long sleepMicroseconds() {
+        return mongocrypt_kms_ctx_usleep(wrapped);
+    }
+
+    @Override
+    public boolean feedAndRetry(final ByteBuffer bytes) {
```
stIncMale, line 114: The current name `feedAndRetry` suggests that the method also retries, which is not true. Let's find a name that is less misleading. Maybe `feedWithRetry` (as is in `CAPI`), or `feedWithRetrySupport`?

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptor.java
```diff
@@ -24,6 +24,15 @@
...
 public interface MongoKeyDecryptor {
 
+    /**
+     * Initial read size after re-sending a KMS request. Matches libmongocrypt's DEFAULT_MAX_KMS_BYTE_REQUEST
+     * and is used when {@link #bytesNeeded()} still returns 0 because libmongocrypt's should_retry flag has
+     * not yet been cleared by {@link #feedAndRetry}.
+     *
+     * @since 5.8
+     */
+    int DEFAULT_KMS_READ_SIZE = 1024;
```
stIncMale, line 34: Where does this constant and the rules accompanying it come from? It seem to  be in conflict with both the [current `integration.md` instructions](https://github.com/mongodb/libmongocrypt/blob/0ea5b7449ac2c7e774827a04495f96af14586117/integrating.md?plain=1#L245-L247) and the API documentation of the methods [`mongocrypt_kms_ctx_bytes_needed`](https://github.com/mongodb/libmongocrypt/blob/0ea5b7449ac2c7e774827a04495f96af14586117/src/mongocrypt.h#L1175-L1188), [`mongocrypt_kms_ctx_feed_with_retry`](https://github.com/mongodb/libmongocrypt/blob/0ea5b7449ac2c7e774827a04495f96af14586117/src/mongocrypt.h#L1190-L1205).

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -486,6 +487,21 @@ public interface mongocrypt_random_fn extends Callback {
     public static native void
     mongocrypt_setopt_bypass_query_analysis (mongocrypt_t crypt);
 
+    /**
+     * Opt-into handling the MONGOCRYPT_CTX_NEED_KMS state with retry logic.
```
stIncMale: Let's make `MONGOCRYPT_CTX_NEED_KMS` a `@link` to the corresponding program element.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
stIncMale, line 1: - Let's update the documentation of the `mongocrypt_kms_ctx_bytes_needed` method that existed before this PR, such that it mentions not only to the `mongocrypt_kms_ctx_feed` method, but also to the `mongocrypt_kms_ctx_feed_with_retry` method.
- Let's express those notions using `@link`.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -1164,6 +1180,47 @@ public interface mongocrypt_random_fn extends Callback {
...
+     * Signal to libmongocrypt that a network error occurred on this KMS request.
+     *
+     * <p>Requires {@link #mongocrypt_setopt_retry_kms} to be enabled.
+     *
+     * @param kms The @ref mongocrypt_kms_ctx_t.
+     * @return True if the request should be retried, false if retries are exhausted.
+     * @since 5.8
+     */
+    public static native boolean
+    mongocrypt_kms_ctx_fail(mongocrypt_kms_ctx_t kms);
```
stIncMale, line 1227: > True if the request should be retried

The documentation of the native method says "may be retried", not "should", which is different. Let's make our documentation consistent with that of the native API, and update the documentation of `MongoKeyDecryptor.fail` accordingly.

stIncMale (changes requested):
Some questions and suggestions after a partial review (the smaller part of it).

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: Thanks - Let me review JAVA-5772 and ensure all is covered there. I followed integrating.md but that most likely was the latest version and effectively have done both tickets.

driver-core/src/main/com/mongodb/internal/connection/tlschannel/impl/TlsChannelImpl.java
rozza: Yes it was causing the tests to fail (I can't recall which one) but it was not returning bytes that needed to be read. Most likely the mock kms server.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptorImpl.java
```diff
@@ -42,6 +46,11 @@
 import static com.mongodb.internal.crypt.capi.CAPIHelper.toByteBuffer;
 import static org.bson.assertions.Assertions.notNull;
 
+/**
+ * Note: Not thread-safe: methods mutate the underlying native {@code mongocrypt_kms_ctx_t} and must be invoked serially.
+ * Callers perform retries sequentially — the sync driver in a {@code while} loop and the reactive driver via
+ * {@code Mono.flatMap} — so no external synchronization is required.
+ */
 class MongoKeyDecryptorImpl implements MongoKeyDecryptor {
```
stIncMale, line 54: 1. We have `@NotThreadSafe` for this, let's use it.
   - Also, given that for some non-obvious reason (do you know what it is?) we use `MongoKeyDecryptorImpl` via the `MongoKeyDecryptor` interface, we should document the interface as non-thread-safe.
3. If you want to document explicitly that the wrapped `mongocrypt_kms_ctx_t` is not thread-safe, then let's do that in the documentation of that type, not here.
4. The "methods mutate the underlying..." text is not needed - it does not add anything useful to the "not thread-safe" information, nor does it on its own imply non-thread-safety.
5. "Callers perform retries sequentially ..." - all that information may be a comment on the calling code, if it's really needed there. I think, the documentation of `MongoKeyDecryptorImpl` is a wrong place for the comments for the code that uses the type.

.evergreen/.evg.yml
```diff
@@ -1632,6 +1642,17 @@ tasks:
           AUTH: "noauth"
           SSL: "nossl"
 
+  - name: "test-kms-retry-task"
```
stIncMale, line 1645: This new task was not run on Evergreen in the patch created from the PR: https://spruce.corp.mongodb.com/version/6a16d0923d7cda00071535ab/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&taskName=test-kms-retry-task.

Even if it does not need to be run for all PRs, we should run it for this PR. I triggered it in the same build.

stIncMale (changes requested):
I am pausing the review until we address the somewhat structurally important concerns:

- https://github.com/mongodb/mongo-java-driver/pull/1953#discussion_r3327641085
- https://github.com/mongodb/mongo-java-driver/pull/1953#discussion_r3335720783
- https://github.com/mongodb/mongo-java-driver/pull/1953#discussion_r3337300603
- https://github.com/mongodb/mongo-java-driver/pull/1953#discussion_r3344210916 - the tests should pass. Update: they [pass](https://spruce.corp.mongodb.com/task/mongo_java_driver_kms_retry_test__version~5.0_os~linux_topology~standalone_test_kms_retry_task_patch_40ee8d511e4541727de2edc4527d209d842fd607_6a16d0923d7cda00071535ab_26_05_27_11_08_07/tests?execution=0&sorts=STATUS%3AASC).

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: Just convention I suppose and I suppose its slightly informative - we don't publish the javadocs either yet write them.

rozza: Resolving as its a Nit / question and not worth changing

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptor.java
rozza: As per the comment above - it matches `DEFAULT_MAX_KMS_BYTE_REQUEST`.

The `fail()` function resets the parser and sets `should_retry = true`, which deliberately makes `bytes_needed` return `0`. This creates an awkward contract: the driver must somehow know to do a read even though bytes_needed says "I don't need anything." 

The integrating guide doesn't document this gap — it just says "loop while bytes_needed > 0" and "call fail() on network error, retry if it returns true," without acknowledging that those two instructions are incompatible without a fallback.

.evergreen/.evg.yml
rozza: I also ran it multiple times in preparing this PR

driver-core/src/main/com/mongodb/internal/connection/tlschannel/impl/TlsChannelImpl.java
rozza: Ah yes it was:

> ClientSideEncryptionKmsRetryProseTest > testCreateDataKeyAndEncryptWithHttpRetry(String) > Case 2: HTTP retry with aws FAILED

If causes: MongoClientException: Exception in encryption library: Allocation size must be greater than zero. 

It doesn't happen on sync because it uses SSLSocket/InputStream, which correctly delivers data before signalling EOF on close_notify.

It isn't triggered by gcp or azure because the failpoint fires on the OAuth token request. So the actual key operation uses another connection and succeeds. As AWS doesn't use a token, and fails with data unconsumed on the wire.

According to Claude:

> The TLS spec allows close_notify to arrive in the same flight as application data. > Any real-world TLS server can legitimately:
>  - Send Connection: close and the response in one write (common for HTTP/1.0 or error responses)
>  - Send a response and immediately initiate TLS shutdown
>  - Have the OS coalesce the response and close_notify into a single TCP segment due to Nagle's algorithm or buffering

TLSChannelImpl fixed this behaviour in 0.5.0

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: Confirmed this includes JAVA-5772. I didn't see a drivers ticket for it and missed it completely. Not that its very clear either.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptorImpl.java
rozza: Done

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptor.java
rozza: I asked Claude to compare my implementation to that of the python binding. Seems I over complicated it (or shortcutted it) and I can just revisit the State machine loop. Will update.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptorImpl.java
rozza: Opted to remove it. Unfortunately, `@NotThreadSafe` isn't available to use here.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: I've got Claude to update them all - as most are very out of date.

I've also added JAVA-6225 to remove any unused statics / methods that either we never or no longer implement

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: Updated all javadocs

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: Updated all javadocs

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptorImpl.java
rozza: Cleaned up - and made sure we explicitly check for 1 rather than != 0.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: Updated all javadocs.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: Updated all Javadocs to bring inline with mongocrypt.h as of 1.18.1

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: Ended up removing all `@since` tags here - as you mentioned the usage was mixed. Also its really tied to the libmongocrypt version.

PR #1956, opened 2026-04-30
Fix NamespaceExists test failure

nhachicha:


rozza (approved):
LGTM - I reckon it would be good for all unified tests to clean up like this. (as long as they aren't running on a pre existing db)

PR #1959, opened 2026-05-05
Change stream events are not emitted for timeseries

rozza:
Removed UnifiedTestModification skipping for timeseries change stream test 
Removed Timeseries prose test

JAVA-6181


strogiyotec (dismissed):
approved, just in case, shouldn't we have a separate PR to merge latest specification changes ?

strogiyotec (approved):

PR #1962, opened 2026-05-06
Add vectorSearch operator for $search pipeline stage

rozza:
JAVA-6130


driver-core/src/main/com/mongodb/client/model/search/VectorSearchOperator.java
```diff
@@ -0,0 +1,56 @@
...
+@Beta(Reason.CLIENT)
+public interface VectorSearchOperator extends SearchOperator {
+
+    /**
+     * Creates a new {@link VectorSearchOperator} with the exact search option.
+     *
+     * @param exact Whether to use exact (ENN) search. If {@code true}, runs exact nearest neighbor search.
+     * @return A new {@link VectorSearchOperator}.
+     */
+    VectorSearchOperator exact(boolean exact);
```
strogiyotec: I didn't really understand why do we need this `exact` if SearchOperator already has `vectorSearch` with `exact` false (client doesn't set it , but according to the [doc ](https://www.mongodb.com/docs/atlas/atlas-search/operators-collectors/vectorSearch/?interface=driver&language=python&vector-search-type=enn&prefilter-type=compound#std-label-fts-vectorSearch-ref/) the default value is false if omitted) and `vectorSearchExact`

driver-core/src/test/unit/com/mongodb/client/model/search/SearchOperatorTest.java
```diff
@@ -1002,6 +1003,125 @@ void regex() {
         );
     }
 
+    @Test
+    void vectorSearch() {
```
strogiyotec, line 1007: one of the acceptance criteria from the [doc ](https://docs.google.com/document/d/1WE6DxUQJqi3xHD5kJAAgTsK0uW2NGOGZOh4Df0EUARQ/edit?tab=t.0) is to
> Implement all necessary testing (unit, integration, e2e) and metrics.
we don't have e2e tests , is it because feature is not available in atlas ?

driver-core/src/main/com/mongodb/client/model/search/VectorSearchOperator.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * A {@link SearchOperator} that performs vector search within the {@code $search} pipeline stage.
+ *
+ * @mongodb.atlas.manual atlas-search/operators-and-collectors/#operators Search operators
+ * @since 5.8
+ */
+@Sealed
+@Beta(Reason.CLIENT)
+public interface VectorSearchOperator extends SearchOperator {
```
strogiyotec, line 30: what about `knnVector` ? From the [doc ](https://docs.google.com/document/d/1WE6DxUQJqi3xHD5kJAAgTsK0uW2NGOGZOh4Df0EUARQ/edit?tab=t.0) I understood that index could be created with `knnVector` field
```
{
  "mappings": {
    "dynamic": false,
    "fields": {
      "embeddings": {
        "type": "knnVector",
```
Should we add a fluent index builder that can set `type` to knnVector ?

strogiyotec (changes requested):

driver-core/src/test/unit/com/mongodb/client/model/search/SearchOperatorTest.java
rozza: We don't have e2e testing yet at a drivers level, I'm expecting this work flow should be convered by a Drivers rather than the doc. 

I'll make sure to add a DRIVERS ticket to ensure its documented in the Specs repo and / or unified tests are added.

driver-core/src/main/com/mongodb/client/model/search/VectorSearchOperator.java
rozza: True - it does allow users to programmatically build a vectorSearch and set the exact value. `vectorSearchExact` is a convience helper that helps with the discoverability of ENN searches.

Lets remove it - will update.

driver-core/src/main/com/mongodb/client/model/search/VectorSearchOperator.java
rozza: The index work was covered in #1960 - this is for creating the query eg:

```
{
  $search: {
    "index": "<index name>", // optional, defaults to "default"
    "vectorSearch": {
      "exact": true | false,
      "path": "<field-to-search>",
      "queryVector": [<array-of-numbers>],
      "filter": {<filter-specification>},
      "limit": <number-of-results>,
      "numCandidates": <number-of-candidates>,
      "score": {<options>}
    }
  }
}
```

Which can be done via:

```
  SearchOperator.vectorSearch(
          fieldPath("embedding"),
          asList(1.0, 2.0),
          10,
          50
  ).filter(SearchOperator.text(fieldPath("title"), "hello"))
          .score(boost(2f))
```

I added `VectorSearchOperator` to allow for overrides that vector search can do but aren't general for all `SearchOperator` instances.  `VectorSearchOperatorConstructibleBsonElement` is the actual implementation for creating the bson.

This follows the existing `SearchOperator` conventions and builds upon them for vector search.

driver-core/src/main/com/mongodb/client/model/search/VectorSearchOperator.java
rozza: Done - removed. Good catch.

strogiyotec (approved):

PR #1965, opened 2026-05-07
JAVA-6197 Fix javadoc warnings in micrometer

rozza:
JAVA-6197


nhachicha (approved):
Thanks for the quick fix.

PR #1975, opened 2026-05-19
Close connection on responseTo mismatch to prevent pool reuse

rozza:
When ReplyMessage detects that the responseTo field in a server response does not match the expected requestId, close the underlying connection.

Previously, MongoInternalException propagated as a regular error without closing the stream. Since PooledConnection.close() only prunes connections that are closed or match shouldPrune() criteria (generation, idle time, lifetime), the desynchronized connection was returned to the pool and could be handed to subsequent operations — risking further data corruption.

A responseTo mismatch means the TCP stream is definitively desynchronized. The connection is unusable and must not be recycled.

JAVA-6210


rozza:
Working on more defensive hardening and will repurpose JAVA-6210 to cover all these cases.

PR #1977, opened 2026-05-20
Add storedSource support for vector search indexes and returnStoredSource for $vectorSearch queries 

rozza:
JAVA-6129

### DIFF
--- a/driver-core/src/main/com/mongodb/client/model/search/VectorSearchConstructibleBson.java
+++ b/driver-core/src/main/com/mongodb/client/model/search/VectorSearchConstructibleBson.java
@@ -17,6 +17,7 @@ package com.mongodb.client.model.search;
 
 import com.mongodb.annotations.Immutable;
 import com.mongodb.internal.client.model.AbstractConstructibleBson;
+import org.bson.BsonBoolean;
 import org.bson.BsonDocument;
 import org.bson.Document;
 import org.bson.conversions.Bson;
@@ -46,6 +47,11 @@ final class VectorSearchConstructibleBson extends AbstractConstructibleBson<Vect
     @Override
     public VectorSearchOptions filter(final Bson filter) {
         return newAppended("filter", notNull("name", filter));
+    }
+
+    @Override
+    public VectorSearchOptions returnStoredSource(final boolean returnStoredSource) {
+        return newAppended("returnStoredSource", new BsonBoolean(returnStoredSource));
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


driver-core/src/main/com/mongodb/client/model/search/VectorSearchConstructibleBson.java
```diff
@@ -48,6 +49,11 @@ public VectorSearchOptions filter(final Bson filter) {
         return newAppended("filter", notNull("name", filter));
     }
 
+    @Override
+    public VectorSearchOptions returnStoredSource(final boolean returnStoredSource) {
```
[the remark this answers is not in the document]
rozza: Fixed :) thanks copilot

driver-scala/src/test/scala/org/mongodb/scala/model/SearchIndexDefinitionSpec.scala
```diff
@@ -77,6 +77,20 @@ class SearchIndexDefinitionSpec extends BaseSpec {
     )
   }
 
+  it should "create a vectorSearch definition with storedSource" in {
+    toBson(
+      vectorSearch(
```
strogiyotec, line 82: I am just curios , why do we also add scala tests when it uses existing java classes that we already unit tested ?

strogiyotec (approved):

driver-core/src/main/com/mongodb/client/model/search/VectorSearchOptions.java
```diff
@@ -41,6 +41,16 @@ public interface VectorSearchOptions extends Bson {
...
 
+    /**
+     * Creates a new {@link VectorSearchOptions} that instructs to return only stored source fields.
+     *
+     * @param returnStoredSource The option to return only stored source fields.
+     * @return A new {@link VectorSearchOptions}.
+     * @mongodb.atlas.manual atlas-vector-search/vector-search-stage/ $vectorSearch
+     * @since 5.8
+     */
+    VectorSearchOptions returnStoredSource(boolean returnStoredSource);
```
strogiyotec, line 52: do you think it's better to make it 
`returnStoredSourceOnly()` not sure what is better 
by default it's false which according to [documentation ](https://www.mongodb.com/docs/atlas/atlas-search/stored-source-definition/)is 
>  implicit full document lookup on the backend database
so I believe users will only use this builder method to only set it to `true`

strogiyotec (approved):

driver-scala/src/test/scala/org/mongodb/scala/model/SearchIndexDefinitionSpec.scala
rozza: To ensure that the added aliases types don't need companion objects to allow idomatic access

driver-core/src/main/com/mongodb/client/model/search/VectorSearchOptions.java
rozza: Agreed, but it does follow the conventions of allowing a value to be configured explicitly either way.

PR #1980, opened 2026-05-21
Widen getSocketAddresses catch to Exceptions for custom resolver failures

rozza:
Custom InetAddressResolver implementations may throw RuntimeException rather than UnknownHostException. Widening the catch ensures all resolver failures are consistently wrapped in MongoSocketException.

JAVA-5855


rozza:
Closing - seems this will change expected behavior with no benefit.

PR #1985, opened 2026-05-27
Mark micrometer OSGi imports as optional and add bundle resolution test

rozza:
The micrometer-observation dependency is optional but its packages were imported as required in the OSGi metadata, causing bundle resolution failure in OSGi containers.

Adds io.micrometer.*;resolution:=optional to driver-core Import-Package.

Also adds a new testing/osgi-test module that uses Apache Felix to verify all driver bundles resolve without optional dependencies present. This runs in CI via static-checks.sh to prevent future regressions.

JAVA-6215


rozza:
This is just a cherry pick and has already been reviewed. Prepping a 5.7.1 release.

PR #1986, opened 2026-05-27
Add BinaryVector and VectorSearchQuery vectorSearch overloads to Scala driver

rozza:
The Scala Aggregates wrapper only exposed the `Iterable[Double]` overload of vectorSearch. This adds the `BinaryVector` (since 5.3) and `VectorSearchQuery` auto-embedding (since 5.7) overloads. Changes the existing overload's parameter from `Iterable[java.lang.Double]` to `Iterable[Double]` to resolve Scala type inference ambiguity when multiple overloads are present.

Also fixes scaladoc warnings for unresolvable cross-module Java type links.

JAVA-5699

### DIFF
--- a/driver-scala/src/main/scala/org/mongodb/scala/model/Aggregates.scala
+++ b/driver-scala/src/main/scala/org/mongodb/scala/model/Aggregates.scala
@@ -18,7 +18,8 @@ package org.mongodb.scala.model
 
 import com.mongodb.annotations.{ Beta, Reason }
 import com.mongodb.client.model.fill.FillOutputField
-import com.mongodb.client.model.search.FieldSearchPath
+import com.mongodb.client.model.search.{ FieldSearchPath, VectorSearchQuery }
+import org.bson.BinaryVector
 
 import scala.collection.JavaConverters._
 import com.mongodb.client.model.{ Aggregates => JAggregates }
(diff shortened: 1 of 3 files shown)


driver-scala/src/main/scala/org/mongodb/scala/model/Aggregates.scala
```diff
@@ -740,12 +741,71 @@ object Aggregates {
...
+   *
+   * @param path The field to be searched.
+   * @param queryVector The `BinaryVector` query vector. The number of dimensions must match that of the `index`.
+   * @param index The name of the index to use.
+   * @param limit The limit on the number of documents produced by the pipeline stage.
+   * @param options Optional `\$vectorSearch` pipeline stage fields.
+   * @return The `\$vectorSearch` pipeline stage.
+   * @see [[https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/ \$vectorSearch]]
+   * @note Requires MongoDB 6.0.10 or greater
+   * @since 5.3
```
vbabanin: Should `@since` be 5.8 (when this method was introduced) rather than 5.3?
```suggestion
   * @since 5.8
```

driver-scala/src/main/scala/org/mongodb/scala/model/Aggregates.scala
```diff
@@ -740,12 +741,71 @@ object Aggregates {
...
+   *
+   * @param path The field to be searched.
+   * @param query The query specification, typically created via `VectorSearchQuery.textQuery`.
+   * @param index The name of the index to use.
+   * @param limit The limit on the number of documents produced by the pipeline stage.
+   * @param options Optional `\$vectorSearch` pipeline stage fields.
+   * @return The `\$vectorSearch` pipeline stage.
+   * @see [[https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/ \$vectorSearch]]
+   * @note Requires MongoDB 6.0.10 or greater
+   * @since 5.7
```
vbabanin: Should `@since` be 5.8 (when this method was introduced) rather than 5.7?
```suggestion
   * @since 5.8
```

vbabanin (dismissed):
LGTM! Just two comment regarding `@since`.

PR #1988, opened 2026-05-28
Optimize RawBsonDocument encode and decode by eliminating intermediate allocations

rozza:
- Add BsonInput.pipe(BsonOutput, int) to remove the temporary byte[] copy in BsonBinaryWriter.pipeDocument() on both encode and decode paths
- Add public getByteBacking(), getByteOffset(), getByteLength() on RawBsonDocument to expose the backing byte array

JAVA-6133


rozza:
@vbabanin #1913 seems to be stuck (likely due to yesterdays github outage). Just rebased it here and doing a final evergreen pass before merging.

PR #1998, opened 2026-06-10
Improvements for connection desynchronization and hardening of error handling with connections

rozza:
JAVA-6210


driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -956,10 +977,15 @@ private ResponseBuffers receiveResponseBuffers(final OperationContext operationC
                     Compressor compressor = getCompressor(compressedHeader);
 
                     ByteBuf buffer = getBuffer(compressedHeader.getUncompressedSize());
-                    compressor.uncompress(messageBuffer, buffer);
+                    try {
```
nhachicha: Do we need an explicit try/catch here? can't we reuse the one online [999](https://github.com/mongodb/mongo-java-driver/pull/1998/changes?diff=split#diff-b4c0702b1d188ac6a1328c1a0807e4ea4235781735aff239169e6d37f167d0bdR999) to also release the buffer?

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -1042,10 +1079,15 @@ public void onResult(@Nullable final ByteBuf result, @Nullable final Throwable t
...
-                            compressor.uncompress(result, buffer);
-
-                            buffer.flip();
-                            replyHeader = new ReplyHeader(buffer, compressedHeader);
+                            try {
+                                compressor.uncompress(result, buffer);
+
+                                buffer.flip();
+                                replyHeader = new ReplyHeader(buffer, compressedHeader);
+                            } catch (Throwable localThrowable) {
```
nhachicha: ditto can't we reuse the catch on line [1102](https://github.com/mongodb/mongo-java-driver/pull/1998/changes?diff=split#diff-b4c0702b1d188ac6a1328c1a0807e4ea4235781735aff239169e6d37f167d0bdR1102) to release the buffer if it's not null?

driver-core/src/test/unit/com/mongodb/internal/connection/InternalStreamConnectionTest.java
```diff
@@ -0,0 +1,1273 @@
...
+                "Underlying stream should be closed to release socket resources");
+    }
+
+    /**
+     * Verifies that when {@code stream.readAsync()} throws a {@link RuntimeException},
+     * the connection is closed, just as for an {@link Error}.
+     */
+    @Test
+    @DisplayName("Async: connection closed when readAsync throws Exception during body read")
+    void asyncClosesConnectionWhenReadAsyncThrowsException() {
```
nhachicha: This test is almost similar to the one above. can't we parametrise using the exception type (ex `new OutOfMemoryError("Java heap space")`, `new RuntimeException("Simulated allocation failure")` ?

nhachicha (commented):
Minor comments, I didn't complete reviewing the entire test file.
**Q1:**
Isn't there a risk of false positive/negative with this approach? we assume all `MongoInternalException` needs to close the connection...what about a future desync condition that might throw some other type?

**Q2:**
Can we flip the logic to be more safe? i.e (Flip the default to closed) and only allow a certain command error to reuse the connection? This approach favour integrity over connection churn 

<details>
<summary>AI review Opus 4.8 1M</summary>

- 🟢 [nit] L586: Sync catch (Exception e) vs async catch (Throwable) asymmetry. If getCommandResult throws an Error (e.g. OutOfMemoryError) during decode in the sync path, no CommandFailedEvent is emitted,
  whereas the async path catches Throwable and emits one. Note this is purely about failed-event symmetry — neither path closes the connection on a non-MongoInternalException Error (framing is intact, so it
  isn't a desync), so connection safety is unaffected. Consider widening for consistency, or leave as-is since an Error during decode is effectively unrecoverable.
   
  - 🟢 [nit] L1031, L1065 (the t != null branches) & callback close() calls: These close() calls are defensively redundant — the callback they forward to (the lambda in receiveMessageAsync, L857-864) also closes
  on a non-null throwable, and readAsync already closes on its own failures. close() is idempotent so this is harmless, but a one-line comment noting "close() is idempotent; this is belt-and-suspenders" would
  stop a future reader from assuming any single one is the sole guardian.
   
  - 🟢 [nit] L1098-1105 (MessageCallback.onResult failure path): Robustness only — the path relies on new ResponseBuffers(...) being the last statement in the try and never throwing, so responseBuffers is never
  leaked. Safe today; a brief comment would harden it against future edits that add throwing code after the assignment.

  driver-core/src/test/unit/com/mongodb/internal/connection/InternalStreamConnectionTest.java


  - 🟢 [nit] L363, L402: asyncClosesConnectionWhenReadAsyncThrowsError/...ThrowsException assert the exception type but not that the original OutOfMemoryError/RuntimeException is preserved as the cause. The sync
  counterpart (L626) does assert assertInstanceOf(OutOfMemoryError.class, thrown.getCause()). Mirror it for symmetry.

  - 🟢 [suggestion] No explicit sync test for buffer-release-before-close ordering on a responseTo mismatch (async has asyncReleasesResponseBufferBeforeCloseOnResponseToMismatch). try-with-resources guarantees
  the ordering and the compressed sync path's release is covered by syncReleasesUncompressedBufferWhenReplyHeaderParsingFails, so this is optional hardening.

  - ✅ No concern. All production changes are within com.mongodb.internal.connection (non-public), plus a new test and a doc. No public/protected signatures, visibility, or types changed.

</details>

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -571,8 +572,8 @@ private <T> T receiveCommandMessageResponse(final Decoder<T> decoder, final Comm
                         new BsonDocumentCodec()), description.getServerAddress(), operationContext.getTimeoutContext());
             }
 
-            commandSuccessful = true;
             commandEventSender.sendSucceededEvent(responseBuffers);
+            commandSuccessful = true;
```
rozza, line 611: ### Why `sendSucceededEvent(...)` comes *before* `commandSuccessful = true`

`commandSuccessful` is the guard that prevents emitting **both** a succeeded and a failed monitoring event for a single command:

```java
commandEventSender.sendSucceededEvent(responseBuffers);
commandSuccessful = true;
...
} catch (Throwable t) {
    if (!connectionIsReusable(t)) {
        close();
    }
    if (!commandSuccessful) {
        commandEventSender.sendFailedEvent(t);   // only if we never signalled success
    }
    ...
    throw t;
}
```

The subtlety is what happens when `sendSucceededEvent` itself throws. It isn't a no-op — it does real work **before** the listener is notified:

```java
public void sendSucceededEvent(final ResponseBuffers responseBuffers) {
    sendSucceededEvent(responseBuffers.getResponseDocument(message.getId(), new RawBsonDocumentCodec()));
}
```

`getResponseDocument(message.getId(), ...)` builds a `ReplyMessage`, which validates `responseTo` against the request id and decodes the body. On a **desync (responseTo mismatch)** it throws `MongoInternalException`; on a **corrupt body** it throws `BsonSerializationException` — both while *constructing* the event, before `commandListener.commandSucceeded(...)` is ever called.

So with the current order:

- `sendSucceededEvent` throws → `commandSuccessful` stays `false` → the catch sends a **failed** event. Correct: a desync/corrupt response should surface as a failure, and the listener was never told it succeeded.

If the two lines were swapped (the old order):

- `sendSucceededEvent` throws → `commandSuccessful` is already `true` → the catch **suppresses** the failed event → monitoring gets a "started" with no terminal event. That's the bug this ordering avoids.

There's no double-event risk on the normal path either: if the listener itself throws, `ProtocolHelper.sendCommandSucceededEvent` wraps `commandListener.commandSucceeded(...)` in `try/catch (Exception)` and swallows it, so `sendSucceededEvent` returns normally and the catch is never entered.

This is covered by `syncSendsFailedEventOnResponseToMismatchInOkResponse` (and its async equivalent) and `syncSendsFailedEventWhenResponseDocumentParsingFails`, which assert exactly one started + one failed event.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
rozza: Done

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
rozza: Done

rozza:
I opted for explicit known errors eg:
> Can we flip the logic to be more safe? i.e (Flip the default to closed) and only allow a certain command error to reuse the connection? This approach favour integrity over connection churn

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -605,6 +606,45 @@ private <T> T receiveCommandMessageResponse(final Decoder<T> decoder, final Comm
...
+     * </ul>
+     * Any other failure (e.g. a responseTo mismatch or an unexpected error) may have left the stream
+     * desynchronized, so the connection must be closed to prevent pool reuse.
+     */
+    private static boolean connectionIsReusable(final Throwable failure) {
+        return failure instanceof MongoCommandException
+                || failure instanceof MongoExecutionTimeoutException
+                || failure instanceof MongoWriteConcernWithResponseException
+                || failure instanceof MongoOperationTimeoutException
+                || failure instanceof BsonSerializationException;
```
nhachicha, line 666: Can we expand `BsonSerializationException` to `BSONException`?

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -605,6 +606,45 @@ private <T> T receiveCommandMessageResponse(final Decoder<T> decoder, final Comm
...
+
+    /**
+     * Handles a failure raised while sending or processing a command, identically on the sync and async paths:
+     * closes the connection unless the failure leaves it {@linkplain #connectionIsReusable reusable}, and emits a
+     * command-failed event unless a succeeded event has already been sent.
+     */
+    private void onCommandFailure(final Throwable failure, final boolean commandSuccessful,
+            final CommandEventSender commandEventSender) {
+        if (!connectionIsReusable(failure)) {
+            close();
```
nhachicha: If `close` throws we're not sending the failed event ... maybe add a suppressed exception to make it resilient ?
```Java
try {
          if (!connectionIsReusable(failure)) {
              close();
          }
      } catch (Throwable closeFailure) {
          failure.addSuppressed(closeFailure)
      }
      if (!commandSuccessful) {
          sender.sendFailedEvent(failure);
      }
```

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -730,34 +770,56 @@ private <T> void sendCommandMessageAsync(final int messageId, final Decoder<T> d
...
+     * as soon as the framed response has been read, before the body is processed. In particular it runs before
+     * {@link #updateSessionContext} could advance session state (operation/cluster time, snapshot timestamp, recovery
+     * token) from a reply that does not belong to this operation. The failure is raised on the same path that
+     * {@linkplain #onCommandFailure handles command failures}, which closes the connection to keep the desynchronized
+     * stream out of the pool.
+     */
+    private void assertResponseToMatches(final ResponseBuffers responseBuffers, final int requestId) {
+        int actualResponseTo = responseBuffers.getReplyHeader().getResponseTo();
+        if (requestId != actualResponseTo) {
+            throw new MongoInternalException(ReplyMessage.responseToMismatchMessage(actualResponseTo, requestId));
```
nhachicha, line 859: I wonder if this could happen because of a misbehaving L4 proxy/ load balancer (multiplexing...) ?

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -583,20 +586,18 @@ private <T> T receiveCommandMessageResponse(final Decoder<T> decoder, final Comm
...
                     if (ctx != null) {
-                        ctx.setResponseStatusCode(String.valueOf(((MongoCommandException) e).getErrorCode()));
+                        ctx.setResponseStatusCode(String.valueOf(((MongoCommandException) t).getErrorCode()));
                     }
                 }
-                tracingSpan.error(e);
+                tracingSpan.error(t);
             }
-            throw e;
+            throw t;
```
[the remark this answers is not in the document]
rozza: False: This compiles cleanly due to Java 7's precise rethrow — the compiler verifies no checked exceptions are thrown in the try block, so t is statically known to be unchecked. Verified with ./gradlew :driver-core:compileJava

nhachicha (commented):
Minor things from me (test pending review) and
<details>
<summary>more comments from AI as well</summary>

### `connectionIsReusable` + `MongoOperationTimeoutException`: latent desync risk

  Not a blocker and I don't think it's a live bug today — but I want to flag a fragility here, because it touches the exact failure mode this PR exists to prevent.

  `MongoOperationTimeoutException` is on the reusable allow-list, and the Javadoc says these types "are only raised after the full framed response has been read off the wire." That isn't actually true for this
  type. `translateReadException` converts a CSOT read-timeout — i.e. an **incomplete** read, a genuine desync — into a `MongoOperationTimeoutException`:

  ```java
  if (operationContext.getTimeoutContext().hasTimeoutMS()) {
      if (e instanceof SocketTimeoutException) {
          return createMongoTimeoutException(createReadTimeoutException((SocketTimeoutException) e));
      } else if (e instanceof MongoSocketReadTimeoutException) {
          return createMongoTimeoutException(e);
      }
  }
  ```

  So `connectionIsReusable`, taken on its own, classifies a timed-out-mid-read connection as **reusable**.

  The reason it's safe right now is purely that the read layer closes the connection *before* that exception is ever produced:

  ```java
  // receiveResponseBuffers (sync)
  } catch (Throwable t) {
      ...
      close();                                            // closed FIRST
      rethrowIfError(t);
      throw translateReadException(t, operationContext);  // only now becomes MongoOperationTimeoutException
  }
  ```

  and likewise `readAsync` calls `close()` in both its `failed` handler and `catch` before `translateReadFailure` runs. By the time `onCommandFailure` consults the predicate, `isClosed()` is already true, so the
  misclassification is inert and the pool discards the connection.

  The concern: this correctness depends entirely on a **non-local invariant** (the read paths always pre-close) that lives in a different method from the predicate that's now meant to be "the" reuse decision. If
  someone later refactors the read path to defer the close to `connectionIsReusable` — which looks reasonable given that's the centralized decision point now — a timed-out read would start handing stale-byte
  connections back to the pool, and the next checkout reads the previous response → `responseTo` mismatch. That's precisely the canonical desync this PR is hardening against, reintroduced.

  **Suggestion (non-blocking):** rather than just correcting the Javadoc, consider routing only *post-read* failures through `onCommandFailure` on the sync path — mirroring what async already does, where read
  failures take the `t1 != null` branch and never reach the predicate. Then `connectionIsReusable` is only ever applied to failures for which "the full framed response was read" is literally true, the read-layer
  pre-close stops being load-bearing for correctness, and `MongoOperationTimeoutException` can't be misjudged regardless of how the read path evolves.

  If it's easier to keep the current structure, at minimum I'd update the Javadoc to say that read-path failures (including read-timeout `MongoOperationTimeoutException`) are already closed by
  `receiveResponseBuffers`/`readAsync`, and that the predicate governs the reuse decision for *post-read* failures only.




### InternalStreamConnection.java

  - 🟡 [important] L609-630 — connectionIsReusable's contract is inaccurate for MongoOperationTimeoutException, and its safety leans on the very non-local invariant this refactor was meant to remove.
  The Javadoc says these types "are only raised after the full framed response has been read off the wire." That is false for MongoOperationTimeoutException: translateReadException (L996-1001) converts a CSOT
  read-timeout — i.e. an incomplete read, a genuine desync — into createMongoTimeoutException(...), which returns MongoOperationTimeoutException. In the sync path that exception can reach onCommandFailure
  through the try-with-resources resource initializer (L569-570), where connectionIsReusable returns true.
  This is not a live bug — receiveResponseBuffers (L1066) already calls close() before throwing, so the connection is closed regardless of the predicate's verdict. But the correctness now depends on remembering
  that the read paths pre-close, which is exactly the implicit, action-at-a-distance coupling the allow-list was introduced to eliminate. A future change that stops pre-closing in receiveResponseBuffers would
  silently start pooling desynced connections.
  Async doesn't have this problem: read failures take the t1 != null branch (L767) and never reach connectionIsReusable. The clean fix is to make sync mirror async — route only post-read failures through
  onCommandFailure, so the predicate is only ever applied to failures for which its contract is literally true. Minimum fix: correct the Javadoc to state that read-path failures (incl. read-timeout
  MongoOperationTimeoutException) are already closed by receiveResponseBuffers/readAsync, and the predicate governs the close decision only for post-read failures.

  - 🟢 [nit] L570 + ReplyMessage:47 — responseTo is now validated twice. assertResponseToMatches validates it upfront, and ReplyMessage's constructor validates it again on the command path (reached via
  getResponseDocument/getCommandResult). The shared message string prevents drift and ReplyMessage is used from other call sites, so this is reasonable defense-in-depth — worth a one-line note that the second
  check is redundant on the command path so nobody "optimizes" it away assuming it's the only guard.

###  InternalStreamConnectionTest.java
  - 🟡 [important] Two allow-list members lack a "NOT closed" test. MongoCommandException, MongoExecutionTimeoutException, and BsonSerializationException are each covered by a not-closed test;
  MongoWriteConcernWithResponseException and MongoOperationTimeoutException (the reusable write-concern-timeout / CSOT-derived case) are not. Given the allow-list is the safety-critical surface — and the
  testing-guide's own "test both sides of the condition" rule — these two should each get a connection-stays-open test driven from a fully-read body.
</details>

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
rozza: Perhaps - in the reported case its most likely due to misconfigured publisher / and or terminating the publisher mid lifecycle.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
rozza: We could do but the aim here is to be explicit about known errors and favour integrity over connection churn.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
rozza: Done

nhachicha (approved):
LGTM 👍 
Some nits from Fable
```
- onCommandFailure doesn't rethrow Error from close() — it only catches RuntimeException, so an Error from close() would replace the original failure. Given close() is best-effort cleanup, catching Throwable (minus re-throwing nothing) and suppressing would be marginally more robust; very low priority.

- receiveCommandResponseBuffers catches Throwable and calls sendFailedEvent(t) before rethrowing — if the event sender itself throws (it shouldn't, given ProtocolHelper's guards, but LoggingCommandEventSender.sendFailedEvent does some work outside them), the original failure is masked. Consider a try/suppress there for symmetry with onCommandFailure.
```

PR #1999, opened 2026-06-11
KMS Retry support

rozza:
Added KMS Retry support.

4 Commits: 

- Update CAPI javadoc and added rules to: ‎mongodb-crypt/AGENTS.md‎
- Implement KMS Retry API in mongo-crypt
- Implemented KMS Retries in sync / reactive streams
- CAPI cstring NPE hardening - found by copilot

JAVA-5391 / JAVA-5772


mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -52,25 +53,36 @@ public String toString() {
...
+     * corresponding (handle)_status function. E.g. {@link #mongocrypt_ctx_status}.
      */
     public static class mongocrypt_status_t extends PointerType {
     }
 
     /**
-     * Contains all options passed on initialization of a @ref mongocrypt_ctx_t.
+     * Contains all options passed on initialization of a {@link mongocrypt_ctx_t}.
      */
     public static class mongocrypt_opts_t extends PointerType {
```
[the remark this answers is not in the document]
rozza: Fixed - now links to `mongocrypt_t`. Note `mongocrypt_opts_t` has no counterpart in mongocrypt.h 1.18.1 at all (it is a legacy binding class, declared but unused), so its Javadoc is Java-authored rather than header-copied; the doc now also notes it is retained for backwards compatibility.

mongodb-crypt/AGENTS.md
```diff
@@ -15,6 +15,24 @@ Client-side field-level encryption (CSFLE) support via JNA bindings to libmongoc
...
 
+## CAPI.java — JNA Binding Declarations
+
+`CAPI.java` declares the JNA native method signatures for libmongocrypt. Javadoc on each method
+must match the documentation in [mongocrypt.h](https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h).
+
+When adding or updating bindings:
+- Copy the doc comment from `mongocrypt.h` verbatim — do not add, remove, or reword sentences,
+  even when extra context (e.g. cross-references to related setopt functions) seems helpful
+- Convert Doxygen markup to Javadoc semantics:
```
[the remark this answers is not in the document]
rozza: Verified the rewritten Javadoc against mongocrypt.h at the 1.18.1 tag (the version this module binds to, `downloadRevision` in build.gradle.kts): the updated blocks - `mongocrypt_status_t`, `mongocrypt_binary_t`, `mongocrypt_t`, `mongocrypt_binary_data`/`_len`, and the four new retry bindings - match the header verbatim modulo the documented Doxygen-to-Javadoc conversion and omitted `@code{.c}` examples. The apparent additions come from comparing against an older header revision.

The underlying point about clarity stands though: the guideline did not cover declarations with no header counterpart (`cstring`, `mongocrypt_opts_t`), omitted code examples, or Javadoc validity. AGENTS.md now pins the comparison target to the bound libmongocrypt version, scopes the verbatim rule to declarations with a header counterpart, allows omitting `@code` blocks, and requires the result to generate cleanly with `./gradlew :mongodb-crypt:javadoc`.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/KeyManagementService.java
```diff
@@ -109,60 +129,90 @@ private void streamWrite(final Stream stream, final MongoKeyDecryptor keyDecrypt
...
         }
     }
 
-    private static void handleError(final Throwable t, final OperationContext operationContext, final MonoSink<Void> sink) {
+    private static void failOrHandleError(final Throwable t, final MongoKeyDecryptor keyDecryptor,
+            final OperationContext operationContext, final MonoSink<Void> sink) {
         if (isTimeoutException(t) && operationContext.getTimeoutContext().hasTimeoutMS()) {
-            sink.error(TimeoutContext.createMongoTimeoutException(TIMEOUT_ERROR_MESSAGE, t));
+            sink.error(TimeoutContext.createMongoTimeoutException(KMS_TIMEOUT_ERROR_MESSAGE, t));
+        } else if (keyDecryptor.fail()) {
```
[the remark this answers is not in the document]
rozza: Fixed.  now passes the original  through / to the error handler, where  checks actual deadline expiry (matching the sync implementation). A socket timeout with budget remaining now falls through to `keyDecryptor.fail()` rather than being classified as a CSOT timeout. Added `KeyManagementServiceKmsRetryTest` with the same two scenarios as the sync `CryptKmsRetryTest` to cover this.

driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/KeyManagementService.java
```diff
@@ -74,6 +82,20 @@ public void close() {
...
 
     Mono<Void> decryptKey(final MongoKeyDecryptor keyDecryptor, @Nullable final Timeout operationTimeout) {
+        return Mono.defer(() -> {
+            long sleepMicros = keyDecryptor.sleepMicroseconds();
+            if (sleepMicros > 0) {
+                checkKmsRetryBackoff(operationTimeout, sleepMicros);
+                // Reactor's scheduler resolves delays at ~1 ms granularity on most JVMs, so sub-millisecond backoff
+                // values are effectively rounded up to 1 ms.
+                return Mono.delay(Duration.of(sleepMicros, ChronoUnit.MICROS))
+                        .then(attemptDecryptKey(keyDecryptor, operationTimeout));
```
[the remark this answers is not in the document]
rozza: `Duration.of(long, TemporalUnit)` is present in the Java 8 `Duration` Javadoc, and the Java 8 API docs explicitly list `Duration.of(long amount, TemporalUnit unit)`. 
`ChronoUnit.MICROS` also exists in the Java 8 docs

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -917,65 +964,67 @@ public interface mongocrypt_random_fn extends Callback {
...
      * FLE2RangeFindDriverSpec is a BSON document with one of these forms:
      *
+     * <pre>
+     * <code>
      * 1. A Match Expression of this form:
-     *    {$and: [{&#60;field&#62;: {&#60;op&#62;: &#60;value1&#62;, {&#60;field&#62;: {&#60;op&#62;: &#60;value2&#62; }}]}
+     *    {$and: [{<field>: {<op>: <value1>, {<field>: {<op>: <value2> }}]}
      * 2. An Aggregate Expression of this form:
-     *    {$and: [{&#60;op&#62;: [&#60;fieldpath&#62;, &#60;value1&#62;]}, {&#60;op&#62;: [&#60;fieldpath&#62;, &#60;value2&#62;]}]
+     *    {$and: [{<op>: [<fieldpath>, <value1>]}, {<op>: [<fieldpath>, <value2>]}]
```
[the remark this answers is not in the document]
rozza: the javadoc convention excludes **/com/mongodb/**/internal/**, and CAPI so it's never processed by the doclet

rozza:
Errors in evergreen are related to: DRIVERS-3547

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -248,11 +265,12 @@ public static class mongocrypt_kms_ctx_t extends PointerType {
...
      *
      * @param status The status object.
-     * @param len an optional length of the returned string. May be NULL.
-     * @return An error message or an empty string.
+     * @param len    An optional length of the returned string (excluding the
+     *               trailing NULL byte). May be NULL.
+     * @return A NULL terminated error message or NULL.
      */
     public static native cstring
     mongocrypt_status_message(mongocrypt_status_t status, Pointer len);
```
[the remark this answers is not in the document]
rozza: Fixed - in an separate commit.

Although this has been here since pre merge to this repo. Its a simple hardening fix to ensure NPE's dont propgate to the user. I can move this out of this review if required.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoCryptImpl.java
```diff
@@ -198,6 +199,9 @@ class MongoCryptImpl implements MongoCrypt {
             mongocrypt_setopt_use_need_kms_credentials_state(wrapped);
         }
 
+        // always enabled; backoff and budget management are the driver's responsibility
+        configure(() -> mongocrypt_setopt_retry_kms(wrapped, true));
```
stIncMale, line 202: Is this comment useful? For me it brings only confusion.

- _"always enabled"_ - that is pretty obvious, as `mongocrypt_setopt_retry_kms` is called unconditionally. What useful information does this part add?
- _"backoff and budget management are the driver's responsibility"_ - as far as I understand, whether to retry and how long to sleep before retrying is managed by `lybmongocrypt`, not the driver, while the comment seem to imply the opposite.

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
```diff
@@ -43,34 +45,61 @@ public cstring(String string) {
             setPointer(m);
         }
 
+        // By default JNA returns a null cstring reference when a native function returns a NULL
+        // pointer (e.g. mongocrypt_status_message for a status with no message). Override fromNative
+        // to return a cstring with a null pointer instead, so callers never receive a null reference
+        // and toString below can normalize it to an empty string.
+        @Override
+        public Object fromNative(final Object nativeValue, final FromNativeContext context) {
```
stIncMale, line 53: I did not find the JNA behavior mentioned in the comment anywhere in the JNA documentation. Let's add the https://github.com/java-native-access/jna/commit/3fc6bc7e539378145813d307d7ffde273515bd34 commit link to the comment as the supporting evidence.

mongodb-crypt/src/test/java/com/mongodb/internal/crypt/capi/CstringTest.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class CstringTest {
+
+    // A native function declared to return a cstring (e.g. mongocrypt_status_message) may return a
+    // NULL pointer. By default JNA would marshal that to a null cstring reference, causing callers
+    // that immediately call toString() to throw a NullPointerException. The fromNative override must
+    // return a non-null cstring instead.
+    @Test
+    void fromNativeReturnsEmptyStringBackedInstanceForNullPointer() {
```
stIncMale: Let's use the documentation comment here, as the comment documents the test.

stIncMale: This test method name is misleading. `fromNative(null, null)` does not return an "empty string–backed instance", as can be observed by `result.getPointer().getString(0)` throwing `NullPointerException`. That is, the returned `cstring` is backed by `Pointer.NULL`, but its `toString` method returns the empty `String`.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+    private static final String FAILPOINT_URL_BASE = "https://" + FAILPOINT_SERVER_ADDRESS;
+
+    @Nullable
+    private static volatile SSLContext failpointSslContext;
+
+    @NonNull
+    protected abstract ClientEncryption getClientEncryption(ClientEncryptionSettings settings);
+
+    @BeforeEach
+    public void setUp() {
```
stIncMale: `@BeforeEach`, `@AfterEach`, `@ParameterizedTest`, `@Test` methods do not have to be `public`. Let's make them all package-access.

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

driver-sync/src/test/functional/com/mongodb/client/ClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.mongodb.client;
+
+import com.mongodb.ClientEncryptionSettings;
+import com.mongodb.client.vault.ClientEncryption;
+import com.mongodb.client.vault.ClientEncryptions;
+
+public class ClientSideEncryptionKmsRetryProseTest extends AbstractClientSideEncryptionKmsRetryProseTest {
+    @Override
+    public ClientEncryption getClientEncryption(final ClientEncryptionSettings settings) {
```
stIncMale: Let's make the class package-access and the `getClientEncryption` method `protected` (as it is in the superclass).

The same applies to the `ClientSideEncryptionKmsRetryProseTest` class in `driver-reactive-streams`.

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+        assumeTrue(System.getProperty("org.mongodb.test.kms.retry.ca.path") != null,
```
stIncMale: The `"org.mongodb.test.kms.retry.ca.path"` string literal is encountered 3 times in code, 1 time in the documentation comment. Furthermore, its content is part another string literal. Let's introduce a `private static final` constant for this value, and use it in all of the aforementioned places (in the documentation comment we can use `{@value #CONSTANT}`).

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+ * <p>Requires the {@code org.mongodb.test.kms.retry.ca.path} system property pointing to the CA cert for the
+ * failpoint server.
+ */
+public abstract class AbstractClientSideEncryptionKmsRetryProseTest {
+
+    private static final String FAILPOINT_SERVER_ADDRESS = "127.0.0.1:9003";
+    private static final String FAILPOINT_URL_BASE = "https://" + FAILPOINT_SERVER_ADDRESS;
+
+    @Nullable
+    private static volatile SSLContext failpointSslContext;
```
stIncMale: - `failpointSslContext` does not need to be `volatile`. `volatile` may be needed to avoid data races if accesses to this field are not already sequential. However, in that case, there is a race condition in the test code, which is a bug, and `volatile` can't fix it.
- Instead of initializing `failpointSslContext` in a `synchronized createFailpointSslContext` method, we should initialize `failpointSslContext` in a `@BeforeAll void beforeAll() {...}`  method. No `synchronized` will be needed, and the test code will be able to read the `failpointSslContext` variable directly.

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+
+        Map<String, Object> gcpCredentials = new HashMap<>();
+        gcpCredentials.put("email", getEnv("GCP_EMAIL"));
+        gcpCredentials.put("privateKey", getEnv("GCP_PRIVATE_KEY"));
+        gcpCredentials.put("endpoint", FAILPOINT_SERVER_ADDRESS);
+
+        Map<String, Map<String, Object>> kmsProviders = new HashMap<>();
+        kmsProviders.put("aws", awsCredentials);
+        kmsProviders.put("azure", azureCredentials);
+        kmsProviders.put("gcp", gcpCredentials);
```
stIncMale, line 188: `kmsProviders` is what the driver API documentation calls the ["map of KMS provider properties"](https://mongodb.github.io/mongo-java-driver/5.9/apidocs/driver-core/com/mongodb/ClientEncryptionSettings.html#getKmsProviders()). Accordingly, let's rename `awsCredentials`, `azureCredentials`, `gcpCredentials` to `...Properties`.

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

.evergreen/run-kms-retry-tests.sh
```diff
@@ -0,0 +1,43 @@
...
+
+############################################
+#            Main Program                  #
+############################################
+RELATIVE_DIR_PATH="$(dirname "${BASH_SOURCE:-$0}")"
+. "${RELATIVE_DIR_PATH}/setup-env.bash"
+echo "Running KMS Retry tests"
+
+cp ${JAVA_HOME}/lib/security/cacerts mongo-truststore
+${JAVA_HOME}/bin/keytool -importcert -trustcacerts -file ${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem -keystore mongo-truststore -storepass changeit -storetype JKS -noprompt
```
stIncMale, line 16: `${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem` currently occurs in 3 places. It's better to extract it into a variable:
```shell
readonly KMS_TLS_CERT_PATH="${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem"
```

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

.evergreen/.evg.yml
```diff
@@ -1632,6 +1642,17 @@ tasks:
...
 
+  - name: "test-kms-retry-task"
+    tags: [ "kms-retry" ]
+    commands:
+      - func: "start-mongo-orchestration"
+        vars:
+          TOPOLOGY: "server"
+          AUTH: "noauth"
+          SSL: "nossl"
+      - func: "start-csfle-servers"
```
stIncMale, line 1653: [informational comment]

This function runs `${DRIVERS_TOOLS}/.evergreen/csfle/setup.sh`, which starts the KMS failpoint server required by the [prose test](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.md#24-kms-retry-tests). One can see that `${DRIVERS_TOOLS}/.evergreen/csfle/setup.sh` indeed starts it at https://github.com/mongodb-labs/drivers-evergreen-tools/blob/18aed4f176dab1ed505c9a2a53466ddf3f4796ec/.evergreen/csfle/start-servers.sh#L70-L74:

```shell
echo "Starting Failpoint Server..."
$COMMAND kms_failpoint_server.py --port 9003 --cert-file "$CSFLE_TLS_FAILPOINT_CERT_FILE" --ca-file "$CSFLE_TLS_FAILPOINT_CA_FILE" >failpoint.log 2>&1 &
echo "$!" >>kmip_pids.pid
echo "Starting Failpoint Server...done."
sleep 1
```

.evergreen/run-kms-retry-tests.sh
stIncMale, line 16: [informational comment]

The file `${DRIVERS_TOOLS}/.evergreen/x509gen/ca.pem` comes from https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/x509gen/ca.pem, which is copied by the `prepare-resources` function in the `.evg.yml`.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+    @Nullable
+    private static volatile SSLContext failpointSslContext;
+
+    @NonNull
+    protected abstract ClientEncryption getClientEncryption(ClientEncryptionSettings settings);
+
+    @BeforeEach
+    public void setUp() {
+        assumeTrue(System.getProperty("org.mongodb.test.kms.retry.ca.path") != null,
+                "org.mongodb.test.kms.retry.ca.path system property is not set");
```
stIncMale: Do we really need a message here? Other `assumeTrue` methods have no message.

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+            TrustManagerFactory tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            tmf.init(trustStore);
+
+            SSLContext sslContext = SSLContext.getInstance("TLS");
+            sslContext.init(null, tmf.getTrustManagers(), null);
+            failpointSslContext = sslContext;
+            return sslContext;
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to create SSL context for failpoint server", e);
+        }
```
stIncMale: This only practical aspect in which the created `SSLContext` is different from the default one, is that it trusts the certificate at `org.mongodb.test.kms.retry.ca.path`. However, since `.evergreen/run-kms-retry-tests.sh` modifies the default trust store with `mongo-truststore` that contains the same certificate as a trusted one, this difference does not even exist.

That is, we need `createFailpointSslContext` neither when creating `ClientEncryption` (we don't need to call `ClientEncryptionSettings.Builder.kmsProviderSslContextMap` there at all), nor when creating `HttpsURLConnection`.

Let's remove all that we don't need.

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+
+    @AfterEach
+    public void tearDown() {
+        // runs even when setUp's assumptions aborted the test, so re-check the environment
+        if (System.getProperty("org.mongodb.test.kms.retry.ca.path") == null || !hasEncryptionTestsEnabled()) {
+            return;
+        }
+        // leave the shared failpoint server clean for whoever runs next; a test that aborts
+        // mid-retry (e.g. on operation timeout) leaves unconsumed failure counts armed
+        resetFailpoints();
```
stIncMale, line 82: Is this a valuable code comment? I am certain that any driver maintainer who looks at this code will immediately understand why an `@AfterEach` method calls `resetFailpoints`.

Similarly, we do not comment in every place where `com.mongodb.client.FailPoint` is used that we use it in a `try`-with-resources block to make sure it is disabled.

Let's remove this comment.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+    private static void setFailpoint(final String failpointType, final int count) {
+        postToFailpointServer("/set_failpoint/" + failpointType, "{\"count\": " + count + "}");
+    }
+
+    /**
+     * Clears any failpoint counts left armed by an earlier test or suite - the failpoint server's
+     * state is global and outlives the test JVMs. In particular, the operation-timeout test arms
+     * more network failures than it consumes before its deadline expires.
+     */
+    private static void resetFailpoints() {
```
stIncMale: Is this documentation comment useful? It communicates tree pieces of information, which either add nothing new or are irrelevant:

1. [adds nothing new] The method _"Clears any failpoint counts"_, which we already know from the method name (though seemingly for no reason instead of reusing the word "reset", another word "clear" is used).
2. [irrelevant] _"the failpoint server's state is global and outlives the test JVMs"_ - this information is irrelevant. The test class neither starts nor shuts down the KMS failpoint server, and this fact alone is enough to know that it should leave the server in the "before" state. Similarly to how tests setting MongoDB failpoints, always disable them.
3. [irrelevant] The "operation-timeout test" leaves active failpoints when it completes successfully - this information is irrelevant because the code should make sure no failpoints are left active regardless of the tests succeeding or failing, which requires resetting them unconditionally. That is, even the "operation-timeout test" did not leave active failpoints when completing successfully, the failpoints would still have to be reset in an `@AfterEach` method.

Let's remove the comment.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+        try {
+            SSLContext sslContext = createFailpointSslContext();
+            URL url = new URL(FAILPOINT_URL_BASE + path);
+            HttpsURLConnection connection = (HttpsURLConnection) url.openConnection();
+            try {
+                connection.setConnectTimeout(10_000);
+                connection.setReadTimeout(10_000);
+                connection.setSSLSocketFactory(sslContext.getSocketFactory());
+                // test-only: self-signed cert, hostname verification intentionally disabled
+                connection.setHostnameVerifier((hostname, session) -> true);
```
stIncMale: If you actually consider this comment useful, let's remove the `test-only:` prefix, as it adds no information, given that the code is in the `srs/test` directory.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+                // the JDK keep-alive cache holds the socket open for several seconds even after
+                // disconnect(). Asking the server to close the connection avoids stalling the
+                // KMS requests that follow.
+                connection.setRequestProperty("Connection", "close");
+
+                byte[] bodyBytes = body.getBytes(StandardCharsets.UTF_8);
+                connection.setRequestProperty("Content-Length", String.valueOf(bodyBytes.length));
+
+                try (OutputStream os = connection.getOutputStream()) {
+                    os.write(bodyBytes);
```
stIncMale, line 251: Let's call `os.flush()` here, because we must even if it works right now without doing that.

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+                }
+
+                int responseCode = connection.getResponseCode();
+                assertEquals(200, responseCode,
+                        "Failpoint server request to " + path + " failed, HTTP status: " + responseCode);
```
stIncMale: - `org.junit.jupiter.api.Assertions.assertEquals` is inappropriate here. Let's either throw a `RuntimeException` or, use a method from `com.mongodb.assertions.Assertions`.
- Let's add `body` to the error message.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+                try (InputStream is = connection.getInputStream()) {
+                    while (is.read() != -1) {
+                        // drain the response so the exchange completes cleanly
+                    }
+                }
+            } finally {
+                connection.disconnect();
+            }
+        } catch (Exception e) {
+            throw new RuntimeException("Failpoint server request to " + path + " failed", e);
```
stIncMale: Let's add `body` to the error message.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
+
+    /**
+     * Prose test: createDataKey fails when the operation timeout expires mid-retry. Configures a 100ms
+     * operation timeout and a failpoint that triggers repeated network errors, so the cumulative retry
+     * backoff normally pushes the operation past its deadline.
+     */
+    @Test
+    public void testCreateDataKeyTimesOutDuringRetry() {
```
stIncMale: Given https://jira.mongodb.org/browse/DRIVERS-3580, there seem to be no reason to add a CSOT-specific test that is not even in the spec. Let's remove it, together with the `createClientEncryptionForRetryTest(Long timeoutMS)` overload.

If we, for some reason, decide to add it nevertheless, I will need to review it.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
```diff
@@ -0,0 +1,327 @@
...
+public abstract class AbstractClientSideEncryptionKmsRetryProseTest {
+
+    private static final String FAILPOINT_SERVER_ADDRESS = "127.0.0.1:9003";
+    private static final String FAILPOINT_URL_BASE = "https://" + FAILPOINT_SERVER_ADDRESS;
+
+    @Nullable
+    private static volatile SSLContext failpointSslContext;
+
+    @NonNull
+    protected abstract ClientEncryption getClientEncryption(ClientEncryptionSettings settings);
```
stIncMale, line 65: IntelliJ IDEA wants us to repeat the `@NonNull` annotation on the overriding methods. Let's do that, or, alternatively, remove the annotation, as it's usefulness here seems very questionable to me.

I implemented the proposed change in https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78, tested [here](https://spruce.corp.mongodb.com/version/6a61584c52f68400071c727c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
```diff
@@ -313,6 +315,87 @@ public void testExplicitEncryptionDecryptionKeyAltName() throws IOException, URI
...
+        assertEquals(getResourceAsDocument("command-reply.json"), decryptedDocument);
+
+        decryptor.close();
+        mongoCrypt.close();
+    }
+
+    @Test
+    public void testKmsRetryExhaustsBudget() {
+        MongoCrypt mongoCrypt = createMongoCrypt();
+        assertNotNull(mongoCrypt);
```
stIncMale: If we really need this `assertNotNull`, which I doubt, let's move it into `createMongoCrypt`, and remove it from other places where it is also present.

The same applies to `testKmsRetryOnHttpErrorAndNetworkError`.

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
```diff
@@ -313,6 +315,87 @@ public void testExplicitEncryptionDecryptionKeyAltName() throws IOException, URI
...
+
+        int retriesAllowed = 0;
+        while (retriesAllowed < 100 && keyDecryptor.fail()) {
+            retriesAllowed++;
+        }
+        assertTrue(retriesAllowed > 0, "expected at least one retry to be allowed");
+        assertTrue(retriesAllowed < 100, "expected the retry budget to be exhausted");
+
+        decryptor.close();
+        mongoCrypt.close();
```
stIncMale: Let's use `try`-with-resources for `decryptor` and `mongoCrypt`.

The same applies to `testKmsRetryOnHttpErrorAndNetworkError`.

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
```diff
@@ -313,6 +315,87 @@ public void testExplicitEncryptionDecryptionKeyAltName() throws IOException, URI
...
+    }
+
+    private MongoKeyDecryptor reenterNeedKms(final MongoCryptContext context) {
+        assertEquals(State.NEED_KMS, context.getState());
+        MongoKeyDecryptor keyDecryptor = context.nextKeyDecryptor();
+        assertNotNull(keyDecryptor);
+        return keyDecryptor;
+    }
+
+    private MongoKeyDecryptor feedKeysAndGetKeyDecryptor(final MongoCryptContext context) {
```
stIncMale: Let's make the method `static`.

The same applies to `reenterNeedKms`.

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
```diff
@@ -313,6 +315,87 @@ public void testExplicitEncryptionDecryptionKeyAltName() throws IOException, URI
...
+
+        // The context stays in NEED_KMS and the key decryptor is re-presented for the retry
+        keyDecryptor = reenterNeedKms(decryptor);
+        assertTrue(keyDecryptor.bytesNeeded() > 0);
+        assertTrue(keyDecryptor.getMessage().remaining() > 0);
+
+        // A network error is retryable while budget remains
+        assertTrue(keyDecryptor.fail());
+        keyDecryptor = reenterNeedKms(decryptor);
+
```
stIncMale: Let's move the following two `assertTrue` invocations

```java
assertTrue(keyDecryptor.bytesNeeded() > 0);
assertTrue(keyDecryptor.getMessage().remaining() > 0);
```

into `reenterNeedKms`? This way, we will neither miss them on the second quoted invocation of `reenterNeedKms`, nor will have to duplicate them there.

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
```diff
@@ -52,6 +54,32 @@
  */
 public final class MongoCryptHelper {
 
+    public static final String KMS_TIMEOUT_ERROR_MESSAGE = "KMS key decryption exceeded the timeout limit.";
+
+    /**
+     * Throws a {@code MongoOperationTimeoutException} if the operation timeout has expired or the
```
stIncMale: Let's `@link` to `MongoOperationTimeoutException`

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
```diff
@@ -313,6 +315,87 @@ public void testExplicitEncryptionDecryptionKeyAltName() throws IOException, URI
...
+        // A complete HTTP 429 response marks the request as retryable. The backoff is jittered,
+        // so it may legitimately be zero; only that the call succeeds can be asserted.
+        assertTrue(keyDecryptor.feedWithRetry(getHttpResourceAsByteBuffer("kms-reply-429.txt")));
+        keyDecryptor.sleepMicroseconds();
+
+        // The context stays in NEED_KMS and the key decryptor is re-presented for the retry
+        keyDecryptor = reenterNeedKms(decryptor);
+        assertTrue(keyDecryptor.bytesNeeded() > 0);
+        assertTrue(keyDecryptor.getMessage().remaining() > 0);
```
stIncMale: [informational comment]

The PR description says that it implements [JAVA-5772: Support retrying a KMS in-place](https://jira.mongodb.org/browse/JAVA-5772) (and indeed, the PR uses the `MongoKeyDecryptor.feedWithRetry` (`mongocrypt_kms_ctx_feed_with_retry(mongocrypt_kms_ctx_t, ...)`) method from that ticket, yet `reenterNeedKms` calls `MongoCryptContext.nextKeyDecryptor` (`mongocrypt_ctx_next_kms_ctx(mongocrypt_ctx_t)`) after feeding a KMS error response. This surprised me, as in an in-place retry [_"If a KMS context indicates retry, retry the KMS request and feed the new response to the same KMS context"_](https://github.com/mongodb/libmongocrypt/blame/54d1959457d7bc5fb86c91b1e9d9ae18444ff9a9/integrating.md#L262-L264).

`com.mongodb.client.internal.Crypt.decryptKeys` also does not do an in-place retry (at the moment, I am unsure about `com.mongodb.reactivestreams.client.internal.crypt.Crypt.decryptKeys`) - it calls `MongoCryptContext.nextKeyDecryptor` each time it retries.

Interestingly, `testKmsRetryExhaustsBudget` does reuse the `MongoKeyDecryptor` (`mongocrypt_kms_ctx_t`), thus testing an in-place retry of a KMS network error.

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
```diff
@@ -52,6 +54,32 @@
...
+                // infinite timeout: no CSOT budget to enforce; libmongocrypt's retry count is the only limit
+                () -> { },
+                remainingMicros -> {
+                    if (remainingMicros < backoffMicros) {
+                        throw TimeoutContext.createMongoTimeoutException(KMS_TIMEOUT_ERROR_MESSAGE);
+                    }
+                },
+                () -> {
+                    throw TimeoutContext.createMongoTimeoutException(KMS_TIMEOUT_ERROR_MESSAGE);
+                });
```
stIncMale:
```suggestion
        operationTimeout.shortenBy(backoffMicros, TimeUnit.MICROSECONDS).onExpired(() -> {
            throw TimeoutContext.createMongoTimeoutException(KMS_TIMEOUT_ERROR_MESSAGE);
        });
```

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
stIncMale, line 1: The changes in this file are all CSOT-related, and @dariakp confirmed that we should not do any more CSOT-related work (except for removing CSOT, or making sure the code still compiles and works). Thus, I think we should remove all the CSOT-related changes from the PR, including CSOT-related tests in `MongoCryptTest`, `KeyManagementServiceKmsRetryTest`.

stIncMale (changes requested):
This is a partial review. I will come back to reviewing this PR once we are out of the woods with the backpressure project.

**The [previous PR](https://github.com/mongodb/mongo-java-driver/pull/1953) was closed and replaced with this one. Let's not do the same to this PR (also, let's not force push to the PR branch)**, as that generally complicates the review process.

Review status of individual files as of `38a268fd7abd774a246a68023c5467ade21ce86c` (✓ means the file was reviewed, ✗ means it wasn't):
- ✓ `.evergreen/.evg.yml`
- ✓`.evergreen/run-kms-retry-tests.sh`
- ✓ `driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java`
- ✗ `driver-reactive-streams/src/main/com/mongodb/reactivestreams/client/internal/crypt/KeyManagementService.java`
- ✓ `driver-reactive-streams/src/test/functional/com/mongodb/reactivestreams/client/ClientSideEncryptionKmsRetryProseTest.java`
- ✗ `driver-reactive-streams/src/test/unit/com/mongodb/reactivestreams/client/internal/crypt/KeyManagementServiceKmsRetryTest.java`
- ✗ `driver-sync/src/main/com/mongodb/client/internal/Crypt.java`
- ✓ `driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java`
- ✓ `driver-sync/src/test/functional/com/mongodb/client/ClientSideEncryptionKmsRetryProseTest.java`
- ✗ `driver-sync/src/test/unit/com/mongodb/client/internal/CryptKmsRetryTest.java`
- ✓ `mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java` - **I did not review the unrelated API documentation changes, and do not plan to review them**.
- ✓ `mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoCryptImpl.java`
- ✓ `mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptor.java`
- ✓ `mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoKeyDecryptorImpl.java`
- ✓ `mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java`
- ✓ `mongodb-crypt/src/test/java/com/mongodb/internal/crypt/capi/CstringTest.java`
- ✓ `mongodb-crypt/src/test/resources/kms-reply-429.txt`
- ✓ `mongodb-crypt/AGENTS.md`

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/CAPI.java
rozza: I don't think we need to add a commit link the comment is clear enough.
Its all because of the getPointer method returns a null - https://java-native-access.github.io/jna/4.2.1/com/sun/jna/Pointer.html#getPointer-long-

mongodb-crypt/src/main/com/mongodb/internal/crypt/capi/MongoCryptImpl.java
rozza: I can see your confusion - its the responsibility of the driver to sleep. Removed.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: This is such a minor nit for test code - I wonder if its even worth doing and reorganising the commits for.

driver-sync/src/test/functional/com/mongodb/client/ClientSideEncryptionKmsRetryProseTest.java
rozza: I question the worth of this for test code - let alone the time it takes to bake back into the organized commits done previously.

~Won't do.~ but claude will.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: These are the common naming conventions already in place across the code. So won't change as we'll just introduce a new convention and therefore add cognative load for this one file.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Just helps debugging - again little value in removing and could be useful

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: I'll keep it in for now for the following reasons:

 * DRIVERS-3580 didnt exist when this PR was created 
 * DRIVERS-3580 - may be rejected
 * We do currently have alpha support for it

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
rozza: Neat

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
rozza: I'm going to keep them in for now - again we haven't removed CSOT yet and the process of removing it is not set. Eg. deprecation for a major release before another major release or straight removal as its alpha

.evergreen/run-kms-retry-tests.sh
rozza: Good shout

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
rozza: Yeah, its copy-pasta, will update all instances.

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
rozza: Again copy pasta - will update all instances.

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
rozza: Confirmed correct — the retry is in-place at the libmongocrypt level.

The contract is about the native `mongocrypt_kms_ctx_t`, not the Java wrapper:

- `feedWithRetry` setting `should_retry` (HTTP 429/503) or `fail()` returning `true` (network error) marks **that same** `kms_ctx` for retry. libmongocrypt keeps it pending rather than done.
- `mongocrypt_ctx_next_kms_ctx` then hands back **the same** `kms_ctx`.
- `nextKeyDecryptor()` wraps it in a new `MongoKeyDecryptorImpl`, but `wrapped` is the same native context — so the retried request is fed to the same context, as `integrating.md` requires.

This is just libmongocrypt's documented sync loop:

```
next_kms_ctx → send → feed_with_retry / fail → next_kms_ctx → … → kms_done
```

The binding docs back this up: `feed_with_retry` says *"retry in-place without calling `mongocrypt_kms_ctx_fail`"*, and `fail` says the ctx *"may be reused"*.

Tests confirm it:

- `testKmsRetryOnHttpErrorAndNetworkError` — `reenterNeedKms(decryptor)` asserts a non-null decryptor with `bytesNeeded() > 0` after both a 429 feed and a `fail()`. If `next_kms_ctx` didn't re-present retryable contexts, this would fail.
- `testKmsRetryExhaustsBudget` — drives repeated in-place `fail()` retries on the same `MongoKeyDecryptor`.

Reactive `Crypt.decryptKeys` uses the same model (recursing via `doOnSuccess`), so it behaves identically.

The only non-in-place bit is the per-iteration wrapper allocation — a thin JNA wrapper over the same native pointer, no behavioral effect.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Done - and added a new rule to `.agents/references/testing-guide.md` seems the driver is consistently inconsistent here! 

It would be good to have these nits automated away via spotless or some other tool in the future.

driver-sync/src/test/functional/com/mongodb/client/ClientSideEncryptionKmsRetryProseTest.java
rozza: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Thanks - I used your approach from https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Thanks - I used your approach from https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Rejected due to the existing naming conventions. Its also the credentials for each KMS provider that could be used. I suppose the logic is the key and value combined becomes a `awsProvider` but the aws provider provides the credentials 🤷

.evergreen/run-kms-retry-tests.sh
rozza: I used your approach from https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: I used your approach from https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78 so this is no longer an issue.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Thanks - I used your approach from https://github.com/stIncMale/mongo-java-driver/commit/189a6d677940e4b56ba1b5cddb44ece00f663d78

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Done

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsRetryProseTest.java
rozza: Done

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
rozza: Done

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
rozza: Done

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
rozza: Done

mongodb-crypt/src/test/java/com/mongodb/crypt/capi/MongoCryptTest.java
rozza: Done

driver-core/src/main/com/mongodb/internal/capi/MongoCryptHelper.java
rozza: Done

rozza:
Thanks @stIncMale - I've actioned most of the comments.

Main open point at the moment is the CSOT implementation as its piped through. I'm the tech lead on [DRIVERS-3580](https://jira.mongodb.org/browse/DRIVERS-3580) and its yet to go through triage. Also the deprecation / removal process hasn't been determined. So its a cheap add at the moment and we may release this improvement before DRIVERS-3580 is implemented.

rozza:
@stIncMale note recent changes in server behaviour are causing some of the CSOT tests to fail. These fail against main as well and are not related to the changes in this PR. See: SERVER-132246

PR #2001, opened 2026-06-16
Update Specs to latest

rozza:
Replaces #2000 - has the necessary skips for pending work.

PR #2002, opened 2026-06-16
JAVA check for DRIVERS-3547: Explicit Encryption Case 2 removes contention-0 find step

rozza:
Servers implementing SERVER-91887 require find payloads to exactly match the collection's configured contention factor, not just avoid exceeding it. Drop the find with contentionFactor=0 (which the server now rejects) and recreate the collection with contention=10 before inserting.


rozza:
Failing tests are addressed in #2001

PR #2006, opened 2026-06-23
JAVA-6236 add builder API for facet without operator and clean up search api from beta Reason.SERVER

strogiyotec:
[JAVA-6236](https://jira.mongodb.org/browse/JAVA-6236) add builder API for facet without operator
Existing API marks `operator` as mandatory but latest `facet` API makes it optional 
https://www.mongodb.com/docs/search/query/operators-collectors/facet/
image

Also removed `@Beta Reason.SERVER` from all classes in `model/search`  except for VectorSearchQuery, TextVectorSearchQuery


driver-core/src/main/com/mongodb/client/model/search/SearchCollector.java
```diff
@@ -54,6 +54,23 @@ static FacetSearchCollector facet(final SearchOperator operator, final Iterable<
...
+     * for each of those groups, faceting over the entire collection.
+     * <p>
+     * Unlike {@link #facet(SearchOperator, Iterable)}, this method omits the search operator, so the facets are computed
+     * across all documents in the collection.</p>
+     *
+     * @param facets The non-empty facet definitions.
+     * @return The requested {@link SearchCollector}.
+     * @mongodb.atlas.manual atlas-search/facet/ facet collector
+     */
+    @Beta({Reason.CLIENT, Reason.SERVER})
```
stIncMale: Looks like the server had GA of the search functionality in 2020 (see https://www.mongodb.com/docs/atlas/search-changelog/#9-june-2020-release), and there is no information on when specific parts of MQL came out of preview. The DOCSP team said they [do not document availability of search operators by DB version](https://jira.mongodb.org/browse/DOCSP-61702?focusedCommentId=8620682&focusedId=8620682&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-8620682).

Since the only information we have, is that currently the collector is not documented as being in preview, we are to interpret it as it not being in preview, despite us not knowing when GA happened. I suspect, the same applies to all the operators, but this has to be checked. We should remove `Reason.SERVER` here and everywhere in `driver-core/src/main/com/mongodb/client/model/search`, except for `VectorSearchQuery`, `TextVectorSearchQuery`, which are [still in preview](https://www.mongodb.com/docs/vector-search/query/aggregation-stages/vector-search-stage/?interface=driver&language=java-sync&deployment-type=atlas&embedding=auto).

We do not have to do that in this PR, but then, we should do that in the same driver release this PR goes into. Because otherwise it will be very weird to see one of the following:

-  The old `SearchCollector.facet` method marked as beta because of `Reason.SERVER`, while the new one not marked the same way.
- The new `SearchCollector.facet` method marked as beta because of `Reason.SERVER` despite the facet collector is not beta anymore.

driver-core/src/main/com/mongodb/client/model/search/SearchCollector.java
```diff
@@ -54,6 +54,23 @@ static FacetSearchCollector facet(final SearchOperator operator, final Iterable<
                 .append("facets", combineToBson(facets)));
     }
 
+    /**
+     * Returns a {@link SearchCollector} that groups results by values or ranges in the specified faceted fields and returns the count
+     * for each of those groups, faceting over the entire collection.
+     * <p>
+     * Unlike {@link #facet(SearchOperator, Iterable)}, this method omits the search operator, so the facets are computed
+     * across all documents in the collection.</p>
```
stIncMale: I see two issues with the current wording:

1. The second sentence (the one after `<p>`) duplicates part of the first sentence.
2. The _"over the entire collection"_ / _"across all documents in the collection"_ phrases are not future-proof.
   - The pipeline stages that search collectors are used in (`$search`/`$searchMeta`), just like all other pipeline stages, act on _the input documents_ (see https://www.mongodb.com/docs/manual/core/aggregation-pipeline/, https://mongodb.github.io/mongo-java-driver/5.8/apidocs/driver-core/com/mongodb/client/model/Aggregates.html). Generally speaking, those documents don't have to come from or be present in any collection (https://www.mongodb.com/docs/manual/reference/operator/aggregation/documents/ is an example). Even if they come from a collection, they may not the all the documents from that collection. Granted, _currently_ `$search`/`$searchMeta` _"must be the first stage of any pipeline"_ and _"fields must be covered by a MongoDB Search index"_, meaning that _currently_, the aforementioned wordings are correct. However, if any of those requirements ever changes, they will stop being correct.
   - I think, the wording should not implicitly rely on any of the two aforementioned current requirements, and should talk about all the input documents instead of talking about all documents in the collection.

I propose us to change the wording for both the existing and the new `facet` methods as per https://github.com/stIncMale/mongo-java-driver/commit/d086d434dca694c7e3a1faf8dfbe5cd54c46e5fd.

stIncMale (changes requested):

stIncMale (changes requested):
added by mistake

driver-core/src/main/com/mongodb/client/model/search/SearchCollector.java
stIncMale, line 1: I forgot to say that we should also update the Scala API, more specifically, `org.mongodb.scala.model.search.SearchCollector`. Kotlin APIs do not have their own `SearchCollector`.

stIncMale (changes requested):

driver-core/src/main/com/mongodb/client/model/search/SearchCollector.java
strogiyotec: removed `Reason.SERVER` from all `model/search` classes except for VectorSearchQuery, TextVectorSearchQuery

let's keep the change in 1 PR (I updated PR title and description)

driver-core/src/main/com/mongodb/client/model/search/SearchCollector.java
strogiyotec: thanks, updated the comment in java and scala methods

driver-core/src/main/com/mongodb/client/model/search/SearchCollector.java
strogiyotec: addressed it [here](https://github.com/mongodb/mongo-java-driver/pull/2006/changes#diff-ed3ba81ad2def4662da5f2f0ac6c22aadff107e810afe5895ad6daa53ce80312R58)

PR #2019, opened 2026-07-20
JAVA-6224 Ensure Kotlin can encode ByteArrays correctly

rozza:
* Fixes regression in bson-kotlin
* Adds explicit ByteArray support to bson-kotlinx

JAVA-6224

### DIFF
--- a/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/ArrayCodecProvider.kt
+++ b/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/ArrayCodecProvider.kt
@@ -25,7 +25,9 @@ public class ArrayCodecProvider : CodecProvider {
     override fun <T : Any> get(clazz: Class<T>, registry: CodecRegistry): Codec<T>? = get(clazz, emptyList(), registry)
 
     override fun <T : Any> get(clazz: Class<T>, typeArguments: List<Type>, registry: CodecRegistry): Codec<T>? =
-        if (clazz.isArray) {
+        // ByteArrays must be encoded as compact BSON Binary by ByteArrayCodec.
+        // Returning null lets the registry fall through to ByteArrayCodec.
+        if (clazz.isArray && clazz != ByteArray::class.java) {
             ArrayCodec.create(clazz.kotlin, typeArguments, registry)
         } else null
 }
--- a/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/ArrayCodecProvider.kt
+++ b/bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/ArrayCodecProvider.kt
@@ -24,10 +24,12 @@ import org.bson.codecs.configuration.CodecRegistry
 public class ArrayCodecProvider : CodecProvider {
     override fun <T : Any> get(clazz: Class<T>, registry: CodecRegistry): Codec<T>? = get(clazz, emptyList(), registry)
 
+    @Suppress("UNCHECKED_CAST")
     override fun <T : Any> get(clazz: Class<T>, typeArguments: List<Type>, registry: CodecRegistry): Codec<T>? =
-        // ByteArrays must be encoded as compact BSON Binary by ByteArrayCodec.
-        // Returning null lets the registry fall through to ByteArrayCodec.
-        if (clazz.isArray && clazz != ByteArray::class.java) {
+        if (clazz == ByteArray::class.java) {
+            // ByteArrays are encoded as compact BSON Binary, not as a BSON Array of Int32.
+            LenientByteArrayCodec as Codec<T>
+        } else if (clazz.isArray) {
             ArrayCodec.create(clazz.kotlin, typeArguments, registry)
         } else null
 }
--- a/bson-kotlin/src/test/kotlin/org/bson/codecs/kotlin/DataClassCodecTest.kt
+++ b/bson-kotlin/src/test/kotlin/org/bson/codecs/kotlin/DataClassCodecTest.kt
@@ -45,6 +45,7 @@ import org.bson.codecs.kotlin.samples.DataClassWithBsonExtraElements
 import org.bson.codecs.kotlin.samples.DataClassWithBsonId
 import org.bson.codecs.kotlin.samples.DataClassWithBsonIgnore
 import org.bson.codecs.kotlin.samples.DataClassWithBsonProperty
+import org.bson.codecs.kotlin.samples.DataClassWithByteArray
 import org.bson.codecs.kotlin.samples.DataClassWithCollections
 import org.bson.codecs.kotlin.samples.DataClassWithDataClassMapKey
 import org.bson.codecs.kotlin.samples.DataClassWithDefaults
@@ -122,6 +123,11 @@ class DataClassCodecTest {
             | "arraySimple": ["a", "b", "c", "d"],
             | "nestedArrays":  [["e", "f"], [], ["g", "h"]],
             | "arrayOfMaps":  [{"A": ["aa"], "B": ["bb"]}, {}, {"C": ["cc", "ccc"]}],
+            | "byteArray": {"${'$'}binary": {"base64": "AQIDBA==", "subType": "00"}},
+            | "nestedByteArrays": [
+            |     {"${'$'}binary": {"base64": "AQI=", "subType": "00"}},
+            |     {"${'$'}binary": {"base64": "AwQF", "subType": "00"}}
+            | ],
             |}"""
                 .trimMargin()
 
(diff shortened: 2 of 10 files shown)


bson-kotlin/src/test/kotlin/org/bson/codecs/kotlin/DataClassCodecTest.kt
```diff
@@ -122,6 +123,11 @@ class DataClassCodecTest {
             | "arraySimple": ["a", "b", "c", "d"],
             | "nestedArrays":  [["e", "f"], [], ["g", "h"]],
             | "arrayOfMaps":  [{"A": ["aa"], "B": ["bb"]}, {}, {"C": ["cc", "ccc"]}],
+            | "byteArray": {"${'$'}binary": {"base64": "AQIDBA==", "subType": "00"}},
+            | "nestedByteArrays": [
+            |     {"${'$'}binary": {"base64": "AQI=", "subType": "00"}},
+            |     {"${'$'}binary": {"base64": "AwQF", "subType": "00"}}
+            | ],
             |}"""
```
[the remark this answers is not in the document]
rozza: It is parsed and passes the test - so won't do for now.

nhachicha (commented):
LGTM, Claude review revealed some minor things, the important one to consider is the migration/breaking existing users with this fix

## Review — blocking & important items

### `bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt`

- 🟡 **[important]** Decoding data written by the regressed versions is now a hard failure with no migration story. Affected releases persisted `ByteArray` fields as BSON arrays of int32; after this fix those documents decode via `ByteArrayCodec`, which throws `BsonInvalidOperationException` ("readBinaryData can only be called when CurrentBSONType is BINARY"). Either add a lenient fallback that reads a BSON array of int32 when `currentBsonType == ARRAY`, or explicitly decide against leniency and document the break in JAVA-6224 / release notes (naming which released versions produced the array form). A test pinning whichever behavior is chosen would document it.

### `bson-kotlin/src/test/kotlin/org/bson/codecs/kotlin/DataClassCodecTest.kt`

- 🔴 **[blocking]** `testDataClassWithObjectArrayEncodesAsBsonArray` is a byte-for-byte duplicate of the updated `testDataClassWithArrays` — identical `DataClassWithArrays` instance and semantically identical expected JSON (differences are whitespace only). It adds no coverage and doubles maintenance. Suggest keeping one: either revert the byte-array additions to `testDataClassWithArrays` and keep the new focused test, or delete the new one.
- 🟡 **[important]** In `testDataClassWithByteArrayDoesNotExpandDocumentSize`, `document.toBsonDocument()` is redundant — `document` is already a `BsonDocument` (the call returns `this`). Use `document.getBinary("byteArray").data.size` directly.


### `bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/ArrayCodecProvider.kt`

- 🟡 **[important]** Returning `null` for `ByteArray` assumes the registry contains a `ByteArrayCodec`. Users composing `ArrayCodecProvider` in a custom registry without `ValueCodecProvider` previously got a (verbose) working codec and will now get `CodecConfigurationException`. This is the intended pre-regression behavior, but worth a line in the release notes.

### `bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/ByteArrayAsBsonBinary.kt`

- 🟡 **[important]** The public `serializersModule` property has no KDoc. In an `explicitApi()` module — and referenced from the class KDoc as one of the two opt-in mechanisms — it should document what registering it does (`@Contextual ByteArray` fields use this serializer), with its own `@since 5.10`.

### `bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecTest.kt`

- 🟡 **[important]** The `@Contextual` + `ByteArrayAsBsonBinary.serializersModule` path advertised in the KDoc has no test. Add one (data class with `@Contextual val byteArray: ByteArray`, codec built with the module) — contextual resolution for primitive-array types has kotlinx-serialization edge cases worth pinning. Also missing: empty/nullable `ByteArray` and a negative test that `deserialize` fails on a non-binary BSON value.
- 🟡 **[important]** Comment wrapping in the two new tests is broken mid-clause ("(subType / 00),", "per-type-argument / annotation") far short of the 120-col limit, and much of the second comment narrates the change rather than stating a constraint. Suggest trimming to the one non-obvious point — the annotation inside `Array<@Serializable(...) ByteArray>` applies per element — and reflowing the rest.

rozza:
@nhachicha updated based on feedback. 

Only issue not addressed because this was a regression is for `bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/DataClassCodec.kt`

rozza:
@nhachicha the latest change adds support for decoding `Array<int32>` into `ByteArray` with validation. So it wont break existing users decoding. Encoding now uses BsonBinary. We will have to document this change and remediation steps.

bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/ArrayCodecProvider.kt
```diff
@@ -24,10 +24,12 @@ import org.bson.codecs.configuration.CodecRegistry
 public class ArrayCodecProvider : CodecProvider {
     override fun <T : Any> get(clazz: Class<T>, registry: CodecRegistry): Codec<T>? = get(clazz, emptyList(), registry)
 
+    @Suppress("UNCHECKED_CAST")
     override fun <T : Any> get(clazz: Class<T>, typeArguments: List<Type>, registry: CodecRegistry): Codec<T>? =
-        // ByteArrays must be encoded as compact BSON Binary by ByteArrayCodec.
-        // Returning null lets the registry fall through to ByteArrayCodec.
-        if (clazz.isArray && clazz != ByteArray::class.java) {
+        if (clazz == ByteArray::class.java) {
+            // ByteArrays are encoded as compact BSON Binary, not as a BSON Array of Int32.
```
nhachicha, line 30: Fable picked this.

**Nested `Array<ByteArray>` legacy decoding still fails with the default registry**

`ArrayCodec.create` resolves element codecs via `codecRegistry.get(byte[].class)`. In `MongoClientSettings.DEFAULT_CODEC_REGISTRY`, `ValueCodecProvider` precedes `KotlinCodecProvider`, so nested elements get the plain `ByteArrayCodec` instead of `LenientByteArrayCodec`. The new test passes only because the test registry lists `ArrayCodecProvider` first.

Repro (production ordering):

```kotlin
data class WithNested(val nested: Array<ByteArray>)

val registry = fromProviders(ValueCodecProvider(), ArrayCodecProvider(), DataClassCodecProvider())
DataClassCodec.create(WithNested::class, registry)!!
    .decode(BsonDocumentReader(BsonDocument.parse("""{"nested": [[1, 2], [3, 4]]}""")), DecoderContext.builder().build())
// BsonInvalidOperationException: readBinaryData can only be called when CurrentBSONType is BINARY...
```

Suggested fix in `ArrayCodec.create` — resolve `ByteArray` elements directly, like the existing top-level special cases:

```kotlin
val codec: Codec<Any?> =
    if (valueClass == ByteArray::class.java) {
        LenientByteArrayCodec as Codec<Any?>
    } else if (nestedTypes.isEmpty()) {
        codecRegistry.get(valueClass)
    } else {
        codecRegistry.get(valueClass, nestedTypes)
    }
```

nhachicha (changes requested):

bson-kotlin/src/main/kotlin/org/bson/codecs/kotlin/ArrayCodecProvider.kt
rozza: Done!

bson-kotlinx/src/test/kotlin/org/bson/codecs/kotlinx/KotlinSerializerCodecTest.kt
```diff
@@ -299,6 +305,132 @@ class KotlinSerializerCodecTest {
...
+        assertRoundTrips(
+            expected,
+            DataClassWithContextualByteArray(byteArrayOf(1, 2, 3, 4)),
+            serializersModule = ByteArrayAsBsonBinary.serializersModule)
+    }
+
+    @Test
+    fun testByteArrayAsBsonBinaryDecodesLegacyBsonArray() {
+        // The built-in ByteArray serializer, and bson-kotlin 5.2.0 - 5.9.x,
+        // regressed abd encoded a ByteArray as a BSON Array of Int32. Those documents must still
```
nhachicha: Typo?
```suggestion
        // regressed and encoded a ByteArray as a BSON Array of Int32. Those documents must still
```

nhachicha (dismissed):
Nice work! 🎉

nhachicha (approved):

PR #2022, opened 2026-07-23
JAVA-6222 Support trace context propagation to the server

nhachicha:
- JAVA-6222

- This is the reference implementation for the Spec https://github.com/mongodb/specifications/pull/1966/ 

- This PR also bump the wire protocol to 29 per https://jira.mongodb.org/browse/JAVA-6164


driver-core/src/test/unit/com/mongodb/internal/observability/micrometer/MicrometerTraceParentTest.java
```diff
@@ -0,0 +1,176 @@
...
+            Class<?> traceContextInterface = Class.forName(
+                    "com.mongodb.internal.observability.micrometer.TraceContext", true, isolatedLoader);
+            Object traceParent = assertDoesNotThrow(
+                    () -> traceContextInterface.getMethod("traceParent").invoke(traceContext),
+                    "traceParent() must not throw NoClassDefFoundError when micrometer-tracing is absent");
+
+            assertNull(traceParent);
+
+            spanInterface.getMethod("closeScope").invoke(span);
+            spanInterface.getMethod("end").invoke(span);
```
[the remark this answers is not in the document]
nhachicha: Rejected: The observation registry there is test-local, so the leak contaminates a dead registry's thread-local rather than other tests' observations

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -277,11 +333,32 @@ private int writeOpMsg(final ByteBufferBsonOutput bsonOutput, final OperationCon
...
+        String traceParent = commandSpan.context().traceParent();
+        if (traceParent == null) {
+            return;
+        }
+        bsonOutput.writeByte(PAYLOAD_TYPE_3_TELEMETRY);
+        BsonDocument telemetry = new BsonDocument("otel",
+                new BsonDocument("traceparent", new BsonString(traceParent)));
+        new BsonDocumentCodec().encode(new BsonBinaryWriter(bsonOutput), telemetry,
+                EncoderContext.builder().build());
+    }
```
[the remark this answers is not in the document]
nhachicha: Rejected: section is now hand-written (refactored)

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -194,6 +217,24 @@ BsonDocument getCommandDocument(final ByteBufferBsonOutput bsonOutput) {
         }
     }
 
+    /**
+     * The command name (first key of the command document. Available before {@code encode()} runs, unlike
+     * {@link #getCommandDocument(ByteBufferBsonOutput)}; identical to the encoded document's first
+     * key, since encoding only appends fields.
+     */
```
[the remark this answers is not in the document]
nhachicha: Applied

gradle/libs.versions.toml
```diff
@@ -186,6 +186,7 @@ project-reactor-test = { module = "io.projectreactor:reactor-test" }
 reactive-streams-tck = { module = " org.reactivestreams:reactive-streams-tck", version.ref = "reactive-streams" }
 reflections = { module = "org.reflections:reflections", version.ref = "reflections" }
 
+micrometer-tracing = { module = "io.micrometer:micrometer-tracing", version.ref = "micrometer-tracing" }
 micrometer-tracing-integration-test-bom = { module = " io.micrometer:micrometer-tracing-bom", version.ref = "micrometer-tracing" }
 micrometer-tracing-integration-test = { module = " io.micrometer:micrometer-tracing-integration-test" }
```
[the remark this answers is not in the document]
nhachicha: Applied

driver-core/src/main/com/mongodb/internal/operation/ServerVersionHelper.java
```diff
@@ -31,6 +31,9 @@ public final class ServerVersionHelper {
     public static final int SIX_DOT_ZERO_WIRE_VERSION = 17;
     public static final int SEVEN_DOT_ZERO_WIRE_VERSION = 21;
     public static final int EIGHT_DOT_ZERO_WIRE_VERSION = 25;
+    // Server 9.0 (WIRE_VERSION_90 = 29). Minimum wire version for the OP_MSG telemetry
+    // section (OTel trace-context propagation, DRIVERS-3454).
+    public static final int NINE_DOT_ZERO_WIRE_VERSION = 29;
     public static final int LATEST_WIRE_VERSION = EIGHT_DOT_ZERO_WIRE_VERSION;
```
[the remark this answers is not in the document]
nhachicha: Bumped to 29

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -156,15 +174,20 @@ BsonDocument getCommandDocument(final ByteBufferBsonOutput bsonOutput) {
...
 
-                // If true, it means there is at least one `PAYLOAD_TYPE_1_DOCUMENT_SEQUENCE` section in the OP_MSG
+                // If true, there are more sections after the `PAYLOAD_TYPE_0_DOCUMENT` section: either one or more
+                // `PAYLOAD_TYPE_1_DOCUMENT_SEQUENCE` sections, and/or a trailing `PAYLOAD_TYPE_3_TELEMETRY` section.
                 if (byteBuf.hasRemaining()) {
                     BsonDocument commandBsonDocument = byteBufBsonDocument.toBaseBsonDocument();
 
                     // Each loop iteration processes one Document Sequence
                     // When there are no more bytes remaining, there are no more Document Sequences
                     while (byteBuf.hasRemaining()) {
```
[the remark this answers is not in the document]
nhachicha: Refactored

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -194,6 +216,24 @@ BsonDocument getCommandDocument(final ByteBufferBsonOutput bsonOutput) {
         }
     }
 
+    /**
+     * The command name (the first key of the command document). Available before {@code encode()} runs, unlike
+     * {@link #getCommandDocument(ByteBufferBsonOutput)}; identical to the encoded document's first
+     * key, since encoding only appends fields.
+     */
+    public String getCommandName() {
+        return command.getFirstKey();
```
rozza, line 225: Is it worth caching this in the constructor? I don't think theres a path where we don't call it?

driver-core/src/main/com/mongodb/internal/observability/micrometer/MicrometerTracer.java
```diff
@@ -47,11 +47,28 @@
  * @since 5.7
  */
 public class MicrometerTracer implements Tracer {
+    /**
+     * {@code micrometer-tracing} is an optional dependency: consumers may have only {@code micrometer-observation}
+     * on the classpath (e.g. metrics/logging-only observability setups). Guard any use of {@code io.micrometer.tracing.*}
+     * types with this flag so that such setups never trigger a {@link NoClassDefFoundError} or {@link LinkageError}.
+     */
+    static final boolean MICROMETER_TRACING_ON_CLASSPATH = isMicrometerTracingOnClasspath();
```
rozza, line 55: Nit: Lets annotation this as visible for testing otherwise private:

```
@VisibleForTesting(otherwise = AccessModifier.PRIVATE)
```

I would have suggested but imports are annoying with suggestions.

driver-core/src/main/com/mongodb/internal/observability/micrometer/MicrometerTracer.java
```diff
@@ -116,6 +133,68 @@ private static class MicrometerTraceContext implements TraceContext {
...
+    }
+
+    /**
+     * Isolates all hard references to {@code io.micrometer.tracing.*} types. This class must only be loaded/invoked
+     * after {@link #MICROMETER_TRACING_ON_CLASSPATH} has been confirmed {@code true}, so that JVMs which resolve
+     * classes referenced by bytecode eagerly (e.g. during verification) do not trigger a {@link NoClassDefFoundError}
+     * when {@code micrometer-tracing} is absent from the classpath.
+     */
+    private static final class MicrometerTracingSupport {
+        @Nullable
```
rozza, line 179: Suggest adding a simple javadoc to help grok this.

```suggestion
          /**
           * Computes the W3C {@code traceparent} for {@code observation}'s active span.
           *
           * @param observation the observation whose span supplies the trace context
           * @return the {@code traceparent} as specified by {@link TraceContext#traceParent()}, or {@code null} when
           *         {@code observation} has no span or its trace/span ids are missing, zero, or malformed
           */
        @Nullable
```

driver-core/src/main/com/mongodb/internal/observability/micrometer/TraceContext.java
```diff
@@ -16,8 +16,18 @@
...
 public interface TraceContext {
-    TraceContext EMPTY = new TraceContext() {
-    };
+    TraceContext EMPTY = () -> null;
+
+    /**
+     * The 55-char W3C {@code traceparent} string for this context
+     * ({@code 00-<32 hex traceId>-<16 hex spanId>-<flags>}; flags {@code 01} sampled / {@code 00} unsampled),
+     * or {@code null} when there is no valid context (no-op span, missing/zero/malformed ids).
+     * Never includes tracestate.
```
rozza, line 29: 👍

driver-core/src/main/com/mongodb/internal/observability/micrometer/TracingManager.java
```diff
@@ -224,9 +221,9 @@ public Span createTracingSpan(final CommandMessage message,
             ConnectionId connectionId = connectionIdSupplier.get();
             mongodbContext.setConnectionId(connectionId);
 
-            if (command.containsKey("getMore")) {
-                long cursorId = command.getInt64("getMore").longValue();
-                mongodbContext.setCursorId(cursorId);
+            BsonInt64 getMoreCursorId = message.getGetMoreCursorId();
+            if (getMoreCursorId != null) {
+                mongodbContext.setCursorId(getMoreCursorId.longValue());
```
rozza, line 226: See previous suggestion for using a `Long`


```suggestion
            Long getMoreCursorId = message.getGetMoreCursorId();
            if (getMoreCursorId != null) {
                mongodbContext.setCursorId(getMoreCursorId);
```

driver-core/src/test/unit/com/mongodb/internal/connection/CommandMessageOtelTraceContextTest.java
```diff
@@ -0,0 +1,293 @@
...
+        return mock(OperationContext.class, mock -> {
+            when(mock.getSessionContext()).thenReturn(sessionContext);
+            when(mock.getTimeoutContext()).thenReturn(timeoutContext);
+        });
+    }
+
+    /**
+     * Walks the OP_MSG sections skip the type-0 body document, then for each subsequent section read the kind byte;
+     * for kind 1 (document sequence) skip past the {@code int32} section size, and for kind 3 (telemetry) decode and
+     * return the BSON document payload. Returns {@code null} if no kind-3 section is found.
```
rozza, line 256: Nit: Slightly easier to read

```suggestion
      /**
       * Walks the OP_MSG sections: skips the type-0 body document, then reads the kind byte of each subsequent
       * section. For kind 1 (document sequence) it skips past the {@code int32} section size; for kind 3 (telemetry)
       * it decodes and returns the BSON document payload. Returns {@code null} if no kind-3 section is found.
       */
```

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -235,6 +275,21 @@ protected void encodeMessageBody(final ByteBufferBsonOutput bsonOutput, final Op
...
     }
 
+    /**
+     * Encodes the message, attaching the OP_MSG telemetry section with {@code commandSpan}'s trace context when
+     * the gating conditions hold (see {@link #writeTelemetryContextSection}). The 2-arg {@link
+     * #encode(ByteBufferBsonOutput, OperationContext)} never attaches the section.
+     */
+    public void encode(final ByteBufferBsonOutput bsonOutput, final OperationContext operationContext,
+            @Nullable final Span commandSpan) {
+        this.commandSpanForEncoding = commandSpan;
```
rozza, line 285: Question: This feels like an antipattern - global state that is safe with the try finally but its a bit hidden. 

Would it be clearer if this was threaded into encode as a parameter instead?

driver-core/src/main/com/mongodb/internal/connection/CommandMessage.java
```diff
@@ -194,6 +216,24 @@ BsonDocument getCommandDocument(final ByteBufferBsonOutput bsonOutput) {
...
+    }
+
+    /**
+     * The cursor id if this is a {@code getMore} command, or {@code null} otherwise.
+     */
+    @Nullable
+    public BsonInt64 getGetMoreCursorId() {
+        BsonValue value = command.get("getMore");
+        return value instanceof BsonInt64 ? (BsonInt64) value : null;
+    }
```
rozza, line 235: I'd recommend changing to a `Long` as that mirrors the `getCursorId` methods in the code base. See: `CommandCursorResult` setting of `cursorId`

```suggestion
    public Long getGetMoreCursorId() {
        BsonValue value = command.get("getMore");
        return value.isNumber() ? value.asNumber().longValue() : null;
    }
```

rozza: I also made a suggestion below to change `getGetMoreCursorId` use.

driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java
```diff
@@ -441,20 +439,19 @@ public boolean reauthenticationIsTriggered(@Nullable final Throwable t) {
     @Nullable
     private <T> T sendAndReceiveInternal(final CommandMessage message, final Decoder<T> decoder,
```
rozza, line 440: I got this feedback from Claude:

```
🟡 `[important]` — `InternalStreamConnection.java`: test gap

  The connection-level rewiring introduced here has no direct coverage. The new CommandMessage/traceparent tests exercise encode(...) in
  isolation and never construct an InternalStreamConnection, and the existing InternalStreamConnectionSpecification predates these changes.

  Not covered:

    1. That the sync and async paths actually call encode(bsonOutput, operationContext, tracingSpan) (the 3-arg overload) with the span from
       createTracingSpan. A regression to the 2-arg overload or null would pass all current tests yet silently stop emitting the telemetry
       section.
    2. The reorder — span created before encode.
    3. Exactly-once semantics under a send failure: the widened catch (Throwable) firing sendFailedEvent + span.error()/closeScope()/end()
       once each, without the receive path double-firing.
    4. The intentional sync/async asymmetry in the catch (async does no closeScope() / direct sendFailedEvent).

  Notes:

    • End-to-end server acceptance of the kind-3 section is 9.0-gated and reasonably deferred.
    • The wiring / ordering / failure paths are unit-testable with the existing TestInternalConnection + Spock harness and a mocked
      TracingManager/Span.
    • Per AGENTS.md ("every code change must include tests"), this should be covered on a safety-critical send path.
```

It may be worth auto migrating `InternalStreamConnectionSpecification.groovy` to Junit 5 and adding extra coverage tests.

rozza (changes requested):
Looks good - have a few comments / suggestions / questions and nits.

A wider question is should we bump micrometer in versions?  Currently we have:

```
micrometer-tracing = "1.6.0-M3" # This version has a fix for https://github.com/micrometer-metrics/tracing/issues/1092
micrometer-observation-bom = "1.15.4"
```

Should we just bump to 1.16.6 (latest 1.16.6) or 1.17.0 (latest)?

PR #2026, opened 2026-07-28
 Support $vectorSearch against nested embeddings and arrays of embeddings

rozza:
JAVA-5987

Also some `@Beta` annotations clean ups


driver-core/src/main/com/mongodb/client/model/Aggregates.java
```diff
@@ -28,8 +28,10 @@
 import com.mongodb.client.model.search.SearchOperator;
 import com.mongodb.client.model.search.SearchOptions;
 import com.mongodb.client.model.search.TextVectorSearchQuery;
+import com.mongodb.client.model.search.VectorSearchNestedOptions;
 import com.mongodb.client.model.search.VectorSearchOptions;
 import com.mongodb.client.model.search.VectorSearchQuery;
+import com.mongodb.client.model.search.VectorSearchScoreMode;
```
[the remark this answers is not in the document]
rozza: Not an issue here: Checkstyle's `UnusedImports` has `processJavadoc=true` by default, so imports referenced from `{@link}` are not flagged. `./gradlew :driver-core:checkstyleMain` passes with these imports in place, and keeping them shorter/readable in the Javadoc is the existing convention in this file. Leaving as is.

driver-scala/src/test/scala-2/org/mongodb/scala/ApiAliasAndCompanionSpec.scala
```diff
@@ -308,13 +308,15 @@ class ApiAliasAndCompanionSpec extends BaseSpec {
...
-      .asScala
-      .filter(_.getPackage.getName == packageName)
-      .filter(classFilter)
-      .map(_.getSimpleName)
-      .toSet
+    val wrapped =
+      (new Reflections(packageName, new SubTypesScanner(false))
+        .getSubTypesOf(classOf[Object])
+        .asScala ++
+        new Reflections(packageName, new SubTypesScanner(false)).getSubTypesOf(classOf[Enum[_]]).asScala)
```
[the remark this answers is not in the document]
rozza: Fixed - the two scans are now a single reused `Reflections` instance. (The separate `getSubTypesOf(classOf[Enum[_]])` query is still needed, since `Enum` is outside the scanned package so enums are not reachable transitively from `Object`.)

rozza:
Added JAVA-6263 - for tracking integration tests.

driver-core/src/main/com/mongodb/client/model/search/SearchOptions.java
```diff
@@ -54,7 +54,6 @@ public interface SearchOptions extends Bson {
      * @param option The counting option.
      * @return A new {@link SearchOptions}.
      */
-    @Beta({Reason.CLIENT, Reason.SERVER})
```
rozza, line 57: Note the whole SearchOptions file is annotated `@Beta(Reason.CLIENT)`

driver-core/src/test/unit/com/mongodb/client/model/search/VectorSearchNestedOptionsTest.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Test
+    void optionNullValue() {
+        assertThrows(IllegalArgumentException.class, () ->
+                VectorSearchNestedOptions.vectorSearchNestedOptions().option("scoreMode", null));
+    }
+
+    @Test
+    void vectorSearchNestedOptionsIsUnmodifiable() {
```
rozza, line 96: I'm not 100% sold on the value of this test and the one below - but both `SearchOptionsTest` and `FuzzySearchOptionsTest` have the same tests.

driver-scala/src/main/scala/org/mongodb/scala/model/search/package.scala
```diff
@@ -310,7 +310,6 @@ package object search {
    * @since 4.11
    */
   @Sealed
-  @Beta(Array(Reason.SERVER))
```
rozza, line 313: Keeping Scala types inline with the Java package. I also added a rule to AGENTS.md

driver-core/src/test/unit/com/mongodb/client/model/search/VectorSearchOptionsTest.java
```diff
@@ -23,8 +23,9 @@
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
-final class BinaryVectorSearchOptionsTest {
+final class VectorSearchOptionsTest {
```
rozza, line 28: Opted to rename so to follow the existing convention of `<Name>Test`

driver-core/src/test/unit/com/mongodb/client/model/search/VectorSearchNestedOptionsTest.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Test
+    void optionEquivalentToScoreMode() {
+        assertEquals(
+                VectorSearchNestedOptions.vectorSearchNestedOptions()
+                        .scoreMode(VectorSearchScoreMode.AVG)
+                        .toBsonDocument(),
+                VectorSearchNestedOptions.vectorSearchNestedOptions()
+                        .option("scoreMode", "avg")
```
strogiyotec, line 72: do we really need this type unsafe `option` method when we have a fluent type safe builders like `scoreMode` I am not sure why user would chose `option` over `scoreMode` or any other methods we expose

driver-core/src/main/com/mongodb/client/model/search/VectorSearchNestedOptions.java
```diff
@@ -0,0 +1,69 @@
...
+     *      .scoreMode(VectorSearchScoreMode.AVG);
+     *  VectorSearchNestedOptions options2 = VectorSearchNestedOptions.vectorSearchNestedOptions()
+     *      .option("scoreMode", "avg");
+     * }</pre>
+     *
+     * @param name The option name.
+     * @param value The option value.
+     * @return A new {@link VectorSearchNestedOptions}.
+     */
+    VectorSearchNestedOptions option(String name, Object value);
```
strogiyotec, line 59: looks like from the docs https://www.mongodb.com/docs/vector-search/query/aggregation-stages/vector-search-stage/?deployment-type=atlas&embedding=byo&interface=driver&language=nodejs
nested options only supports `scoreMode ` 
Do we need to include this `option` method ? I think it's better to add new fluent builder method once nestedOptions supports more fields
What do you think ?

driver-core/src/main/com/mongodb/client/model/search/VectorSearchOptions.java
```diff
@@ -41,6 +44,41 @@ public interface VectorSearchOptions extends Bson {
...
+     * for example it controls how the scores of the individual matching embeddings within a document are
+     * aggregated (see {@link VectorSearchNestedOptions#scoreMode(VectorSearchScoreMode)}).</p>
+     *
+     * @param nestedOptions The options for a nested (embedded) {@code $vectorSearch}.
+     * @return A new {@link VectorSearchOptions}.
+     * @mongodb.atlas.manual atlas-vector-search/vector-search-stage/ $vectorSearch
+     * @mongodb.server.release 8.3
+     * @since 5.10
+     */
+    VectorSearchOptions nestedOptions(VectorSearchNestedOptions nestedOptions);
```
strogiyotec, line 80: Should we add fluent builders for the rest of the fields ? I highlighted with yellow what this class already supports  (except for numCandidates and exact)

Image

strogiyotec (approved):

driver-core/src/main/com/mongodb/client/model/search/VectorSearchOptions.java
strogiyotec: Not relevant, I double checked and found
queryVector and numCandidates

driver-core/src/test/unit/com/mongodb/client/model/search/VectorSearchNestedOptionsTest.java
rozza: This is a common as all of `SearchOptions`, `FuzzySearchOptions` and `VectorSearchOptions` have this as it allows for expandability for any future server options in this driver.

driver-core/src/main/com/mongodb/client/model/search/VectorSearchNestedOptions.java
rozza: As this may change over time I opted to add to keep in line with the other Search option classes.

PR #2028, opened 2026-07-31
Add code design rules for AI

stIncMale:
Part of [JAVA-6240](https://jira.mongodb.org/browse/JAVA-6240), as discussed in https://github.com/stIncMale/mongo-java-driver/pull/3#discussion_r3626303384.

This change was reviewed using Claude. Claude was also used to explain what the change should look like.

### DIFF
--- a/.agents/references/code-design.md
+++ b/.agents/references/code-design.md
@@ -1,0 +1,11 @@
+---
+name: code-design
+description: Design rules for implementation code in the MongoDB Java Driver. Use when adding or modifying an internal program element, or changing the implementation of a public API program element.
+---
+# Code Design
+
+## Executors
+
+- When instantiating an executor, prefer the `MongoThreadPoolExecutor` and `MongoScheduledThreadPoolExecutor` implementations.
+- Any task executed, submitted, or scheduled via an executor must not allow an `Exception` to be propagated;
+  `Error`s should generally not be caught, but if they are, they must still be propagated.
--- a/.agents/references/code-design.md
+++ b/.agents/references/code-design.md
@@ -4,8 +4,23 @@ description: Design rules for implementation code in the MongoDB Java Driver. Us
 ---
 # Code Design
 
+## Access Modifiers
+
+- Use the most restrictive access modifier that is sufficient for now.
+- If there is an internal program element suitable for the task but not accessible, relax its access modifier
+  to the least permissive one that makes it accessible,
+  unless it contradicts the intent expressed in the documentation of the program element in question.
+  Be careful not to make the program element part of the public API accidentally.
+- When access is relaxed only for tests, annotate the program element with `VisibleForTesting`.
+    
+
 ## Executors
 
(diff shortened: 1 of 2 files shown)


stIncMale:
@strogiyotec, please ignore the review request. The bot is too stupid and insists on its choice even after having been manually overridden.

.agents/references/code-design.md
```diff
@@ -0,0 +1,26 @@
+---
+name: code-design
+description: Design rules for implementation code in the MongoDB Java Driver. Use when adding or modifying an internal program element, or changing the implementation of a public API program element.
+---
+# Code Design
+
```
vbabanin: Let's create a section ##Hard constraints so the agent know what "Use", "Annotate" means - there is a possibility that agents in large context might treat it as soft guidelines.

.agents/references/code-design.md
```diff
@@ -0,0 +1,26 @@
...
+## Access Modifiers
+
+- Use the most restrictive access modifier that is sufficient for now.
+- If there is an internal program element suitable for the task but not accessible, relax its access modifier
+  to the least permissive one that makes it accessible,
+  unless it contradicts the intent expressed in the documentation of the program element in question.
+  Be careful not to make the program element part of the public API accidentally.
+- When access is relaxed only for tests, annotate the program element with `VisibleForTesting`.
+
```
vbabanin: It emerges from prompt / agent instruction design because agents need to distinguish:

goals: what to achieve
guidance: preferred style / approach
constraints: what must not be violated
examples: concrete applications of the rule

The current text covers relaxing access modifiers, but it does not clearly cover the public-API -> internal-code boundary case.

```
When code in a public API package needs to call internal implementation code, that internal code must remain in an internal package.

If access rules require the internal program element to be `public` so it can be called from the public API package, make it `public` in the internal package.

Do not move internal classes or interfaces into public API packages only to keep them package-private.
```

.agents/references/code-design.md
```diff
@@ -0,0 +1,26 @@
...
+    
+
+## Executors
+
+- Avoid instantiating new executors/threads. Consider using the existing `CommonExecutor` or `AsyncClientExecutor`.
+- If a new executor/thread is unavoidable, prefer instantiating a new executor with a single thread over a bare thread.
+  It should be created and managed either by `CommonExecutor`, `AsyncClientExecutor`,
+  or a class whose instance is accessible via them. This may require changing their design, implementation, documentation. 
+- When instantiating an executor, prefer the `MongoThreadPoolExecutor` and `MongoScheduledThreadPoolExecutor` implementations.
+- Use daemon threads, see `DaemonThreadFactory`.
```
vbabanin, line 24: Should we make this a MUST? We only spawn daemon threads, and I don’t foresee us using non-daemon threads.

```suggestion
- Must use daemon threads, see `DaemonThreadFactory`.
```

vbabanin (approved):
LGTM!

PR #2032, opened 2026-08-07
Stop BufferPoolPruner thread when last MongoClient closes

### LINKED ISSUE #2029: Resource Leak: BufferPoolPruner thread is not stopped on MongoClient.close() when using Tailable Cursors on Capped Collections
When using the MongoDB Java Driver within an Apache Tomcat environment, undeploying or reloading the web application triggers a severe memory leak warning regarding the BufferPoolPruner thread. 

This issue specifically surfaces when the application utilizes a continuous listening mechanism (e.g., a Tailable Cursor with Await) on a Capped Collection. Calling MongoClient.close() during the application shutdown lifecycle (e.g., inside a ServletContextListener) fails to terminate this background thread. Because the thread is tied to the static PowerOfTwoBufferPool.DEFAULT instance, it remains actively running and holds references, preventing Tomcat from garbage collecting the webapp.

Steps to Reproduce:
1. Deploy a web application using the MongoDB Java driver in Apache Tomcat.
2. Initialize a MongoClient during application startup.
3. Open a continuous query/listener using a Tailable Cursor (e.g., CursorType.TailableAwait) on a Capped Collection.
4. Trigger the application shutdown phase and call MongoClient.close().
5. Undeploy or reload the application via Tomcat Manager.
6. Check the Tomcat logs.

Actual Results:
Tomcat reports the following severe warning preventing proper garbage collection of the WebappClassLoader:
"The web application [ROOT] appears to have started a thread named [BufferPoolPruner-1-thread-1] but has failed to stop it. This is very likely to create a memory leak."

Expected Behavior:
The driver should properly interrupt and clean up the BufferPoolPruner thread when MongoClient.close() is called, even if tailable cursors or background listeners were active on capped collections. Alternatively, a public API method should be provided to gracefully shut down static pool threads upon container undeployment.

Current Workaround:
To prevent the memory leak on undeploy, we currently have to rely on reflection to access driver internals and explicitly call the disablePruning() method on the static pool to stop the thread:

    /**
     * The driver owns a static PowerOfTwoBufferPool.DEFAULT which starts a "BufferPoolPruner" thread.
     * When using tailable cursors on capped collections, MongoClient.close() does not stop it, 
     * causing Tomcat to report a leak on undeploy.
     * Both the field holder and disablePruning() are driver internals, requiring reflection.
     */
    private static void stopBufferPoolPruner() {
        try {
            Class<?> poolClass = Class.forName("com.mongodb.internal.connection.PowerOfTwoBufferPool");
            Object pool = poolClass.getField("DEFAULT").get(null);
            java.lang.reflect.Method disablePruning = poolClass.getDeclaredMethod("disablePruning");
            disablePruning.setAccessible(true);
            disablePruning.invoke(pool);
        } catch (Exception e) {
            Fx.log("disablePruning---->");
            e.printStackTrace();
        }
    }

arimu1:
## Summary

Fixes #2029.

`PowerOfTwoBufferPool.DEFAULT` starts a `BufferPoolPruner` daemon thread that was never stopped when `MongoClient.close()` was called. In container environments (e.g. Tomcat), that leaves a live thread holding the webapp classloader after undeploy, producing:

> The web application [...] appears to have started a thread named [BufferPoolPruner-1-thread-1] but has failed to stop it.

### Changes

- Make `enablePruning()` / `disablePruning()` **idempotent and restartable** (recreate the scheduled executor after shutdown; await termination so the thread is gone before `close()` returns).
- Add **retain/release** reference counting on the pool so the shared DEFAULT pruner:
  - stays running while any `MongoClient` is open
  - is stopped when the **last** client is closed (safe for multi-client apps)
- Hook retain/release into sync and reactive `MongoClientImpl` for real clients (those with a non-null `externalResourceCloser` / `StreamFactoryFactory`).
- Unit tests for stop, restart, retain/release, and pruner thread termination.

### Testing

```bash
./gradlew :driver-core:test --tests "com.mongodb.internal.connection.PowerOfTwoBufferPoolTest" -PskipCryptVerify=true
# 8/8 PASSED
./gradlew :driver-sync:compileJava :driver-reactive-streams:compileJava -PskipCryptVerify=true
# BUILD SUCCESSFUL
```

Temurin 21 / macOS.

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
+++ b/driver-core/src/main/com/mongodb/internal/connection/PowerOfTwoBufferPool.java
@@ -19,6 +19,7 @@ package com.mongodb.internal.connection;
 import com.mongodb.internal.diagnostics.logging.Logger;
 import com.mongodb.internal.diagnostics.logging.Loggers;
 import com.mongodb.internal.thread.DaemonThreadFactory;
+import com.mongodb.lang.Nullable;
 import org.bson.ByteBuf;
 import org.bson.ByteBufNIO;
 
@@ -31,6 +32,7 @@ import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * <p>This class is not part of the public API and may be removed or changed at any time</p>
@@ -63,7 +65,10 @@ public class PowerOfTwoBufferPool implements BufferProvider {
 
     private final Map<Integer, BufferPool> powerOfTwoToPoolMap = new HashMap<>();
     private final long maxIdleTimeNanos;
-    private final ScheduledExecutorService pruner;
+    private final Object prunerLock = new Object();
+    private final AtomicInteger pruneRetainCount = new AtomicInteger();
+    @Nullable
+    private ScheduledExecutorService pruner;
 
     /**
      * Construct an instance with a highest power of two of 24.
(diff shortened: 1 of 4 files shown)


strogiyotec:
hi @arimu1 , thanks for raising this PR , we will talk with the team next week and will add it to our board for review

strogiyotec:
@arimu1 hi , after looking at this ticket with a team we decided to postpone the review as we are already working on a feature that touches the same code flow , we will post the update as soon as it's released

PR #2034, opened 2026-08-11
Fix regression in JSON primitive handling

rozza:
JSON floating-point literals with an integral value were encoded as integer BSON types.

Change to type the literal from its text instead. A literal containing a fraction or an exponent is encoded as a BSON double, matching how the driver's own JSON parser types numbers.

Outside the range of the matching BSON type a literal still widens to a Decimal128 rather than losing data.

Because kotlinx.serialization has no BigDecimal support, a BigDecimal in a JsonObject is stored as its toString() and cannot be distinguished from a hand-written literal, so BigDecimal("1E+19") encodes as a double. Callers needing an exact BSON type should declare a typed property such as Decimal128 or BsonValue.

JAVA-6280

### DIFF
--- a/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/JsonBsonEncoder.kt
+++ b/bson-kotlinx/src/main/kotlin/org/bson/codecs/kotlinx/JsonBsonEncoder.kt
@@ -17,6 +17,7 @@ package org.bson.codecs.kotlinx
 
 import java.math.BigDecimal
 import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.SerializationException
 import kotlinx.serialization.SerializationStrategy
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonArray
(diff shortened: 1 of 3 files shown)


rozza:
Here is an overview of the changes:

| literal | 5.6.0 | 5.7 - 5.9 | Now |
|---|---|---|---|
| `3` | `INT32` | `INT32` | `INT32` |
| `3.0` | `DOUBLE` | **`INT32`** | `DOUBLE` |
| `-3.0` | **`DECIMAL128`** | **`INT32`** | `DOUBLE` |
| `0.0` | **`DECIMAL128`** | **`INT32`** | `DOUBLE` |
| `-0.0` | **`DECIMAL128`** | **`INT32`** | `DOUBLE` (sign kept) |
| `1.1` | `DOUBLE` | `DOUBLE` | `DOUBLE` |
| `-1.1` | **`DECIMAL128`** | `DOUBLE` | `DOUBLE` |
| `11.0` | `DOUBLE` | **`INT32`** | `DOUBLE` |
| `1.1E1` | **crash** | **`INT32`** | `DOUBLE` |
| `3.0E1` | **crash** | **`INT32`** | `DOUBLE` |
| `1e20` | `DOUBLE` | **`DECIMAL128`** | `DOUBLE` |
| `-1e20` | **`DECIMAL128`** | **`DECIMAL128`** | `DOUBLE` |
| `1E5` | `DOUBLE` | **`INT32`** | `DOUBLE` |
| `1E+19` | `DOUBLE` | **`DECIMAL128`** | `DOUBLE` |
| `1E-1` | `DOUBLE` | `DOUBLE` | `DOUBLE` |
| `1.7976931348623157E308` (`Double.MAX_VALUE`) | `DOUBLE` | **`DECIMAL128`** | `DOUBLE` |
| `1e-320` (subnormal) | `DOUBLE` | `DOUBLE` | `DOUBLE` |
| `1e400` | `DECIMAL128` | `DECIMAL128` | `DECIMAL128` |
| `1e-330` | `DECIMAL128` | `DECIMAL128` | `DECIMAL128` |
| `9223372036854775808` | `DECIMAL128` | `DECIMAL128` | `DECIMAL128` |

IamPritamAcharya:
Hi @rozza , I noticed this overlaps with #2033, which I opened earlier to address the same floating-point JSON literal regression.

This PR looks like it expands the fix with additional validation and edge-case handling. Is #2034 intended to supersede #2033? Happy to coordinate or update #2033 if there are changes you'd prefer to see there.

PR #2041, opened 2026-08-18
Implement the overload retry policy for commands that are read/write retryable

stIncMale:
This PR was originally drafted as https://github.com/stIncMale/mongo-java-driver/pull/4.

- [JAVA-6248](https://jira.mongodb.org/browse/JAVA-6248) - of the [relevant spec changes](https://github.com/mongodb/specifications/commit/1125200e4a6161d87cb5090860597eb8e8e90bf1), [`client-backpressure.md`](https://github.com/mongodb/specifications/blob/8a8a7c56429c80b51ec62268dcafc5e5e3c477ef/source/client-backpressure/client-backpressure.md) and [`transactions.md`](https://github.com/mongodb/specifications/blob/1125200e4a6161d87cb5090860597eb8e8e90bf1/source/transactions/transactions.md) was implemented for commands that are read/write retryable.
- [JAVA-6055](https://jira.mongodb.org/browse/JAVA-6055) - conducted the audit as per the `TODO-BACKPRESSURE` in the ticket. The audited logic is in `SpecRetryPolicy.decideWriteProspectiveFailedResult`. We are good, no code changes are needed.


driver-core/src/main/com/mongodb/internal/time/ExponentialBackoff.java
```diff
@@ -43,17 +43,38 @@ private ExponentialBackoff() {
...
+        return Duration.ofMillis(calculateBackoffMs(100, 10000, 2, attemptNumber));
+    }
+
+    /**
+     * Calculate the backoff in milliseconds for transaction retries.
+     *
+     * @param attemptNumber attempt number > 0
+     * @return The calculated backoff in milliseconds.
+     */
+    public static long calculateBackoffMs(final double baseMs, final double maxMs, final double growth, final int attemptNumber) {
```
stIncMale: `VAKOTODO` Make `private`.

driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
```diff
@@ -333,6 +384,8 @@ private int getMaxAttempts() {
                 maxRetries = Descriptor.WRITE.maxRetries;
             } else if (policies.containsKey(Descriptor.READ)) {
                 maxRetries = Descriptor.READ.maxRetries;
+            } else if (policies.containsKey(Descriptor.OVERLOAD)) {
+                maxRetries = overload().map(State.Overload::getMaxAttempts).orElse(Descriptor.OVERLOAD.maxRetries);
```
stIncMale: `VAKOTODO` Assert `overload()` returns non-empty `Optional`, because `policies` are never empty (see `assertValid`).

driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
```diff
@@ -462,6 +523,107 @@ public String toString() {
...
+                    requirementsMet = effectiveRetrySetting;
+                    this.maxAdaptiveRetriesSetting = maxAdaptiveRetriesSetting;
+                    sessionScopedState = null;
+                }
+
+                int getMaxAttempts() {
+                    return maxAttempts(maxAdaptiveRetriesSetting == null
+                            ? IndividualPolicies.Descriptor.OVERLOAD.maxRetries
+                            : maxAdaptiveRetriesSetting);
+                }
```
stIncMale: `VAKOTODO` The `getMaxAttempts` method is called to compute `maxRetries` in `IndividualPolicies.getMaxAttempts`, which is then converted to max attempts again via `maxAttempts(maxRetries)`. This is wrong.

driver-core/src/main/com/mongodb/internal/time/ExponentialBackoff.java
stIncMale: Done in 9e447d5d60a2dadde08eec0234db6bf1db874f15.

driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
stIncMale: Done in 85f1df599d0f72ad3b5ebe6bb760a79490a4688b.

driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
stIncMale: Done in f677474271a162739581b9cc5d1a32fba23a0524.

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
stIncMale, line 1: @vbabanin Create a ticket on refactoring `BaseClientSessionImpl` and its subclasses (`ClientSessionImpl`, `ClientSessionPublisherImpl`):

- We need to reduce the state and code duplication between subclasses.
- We need to split this god object into multiple objects each implemented by a different class responsible for its own area. `BaseClientSessionImpl.OverloadRetryPolicyState` (`BaseClientSessionImpl.overloadRetryPolicyState`) is a precedent of the new state and logic being put in separate classes, and `BaseClientSessionImpl` being composed of their instances.

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
```diff
@@ -249,4 +263,301 @@ protected enum TransactionState {
...
+        @Nullable
+        CommitScoped getCommitScoped();
+
+        /**
+         * @see #openCommitScope()
+         */
+        void closeCommitScope();
+
+        /**
+         * A part of {@link DefaultOverloadRetryPolicyState} restricted to the execution of a command
```
stIncMale: `VAKOTODO` `DefaultOverloadRetryPolicyState` -> `OverloadRetryPolicyState`

driver-core/src/main/com/mongodb/internal/async/function/RetryControl.java
```diff
@@ -80,15 +80,22 @@ public P getPolicy() {
...
         try {
             if (disabled) {
                 throw attemptFailedResult;
             }
-            // this `RetryControl` must not be mutated before calling `onAttemptFailure`
-            Decision decision = onAttemptFailure(policy, this, prospectiveFailedResult(), attemptFailedResult);
+            Throwable prospectiveFailedResult = prospectiveFailedResult();
+            if (attemptFailedResult instanceof Error) {
+                onAttemptFatalFailure(policy, prospectiveFailedResult, (Error) attemptFailedResult);
+                policy.onAttemptFatalFailure();
```
stIncMale: `VAKOTODO` Remove `policy.onAttemptFatalFailure();`

driver-core/src/main/com/mongodb/internal/async/function/RetryControl.java
```diff
@@ -80,15 +80,22 @@ public P getPolicy() {
...
-    RetryAttemptInfo advanceOrThrow(final Throwable attemptFailedResult) throws RuntimeException {
+    RetryAttemptInfo advanceOrThrow(final Throwable attemptFailedResult) throws RuntimeException, Error {
         assertNotNull(attemptFailedResult);
         try {
             if (disabled) {
                 throw attemptFailedResult;
             }
-            // this `RetryControl` must not be mutated before calling `onAttemptFailure`
-            Decision decision = onAttemptFailure(policy, this, prospectiveFailedResult(), attemptFailedResult);
+            Throwable prospectiveFailedResult = prospectiveFailedResult();
```
stIncMale: `VAKOTODO` Use `getProspectiveFailedResult().orElse(null)`.

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
stIncMale: Done in 0a3a76cf413ab1c6933b2988541e9768fb135a08.

driver-core/src/main/com/mongodb/internal/async/function/RetryControl.java
stIncMale: Done in 68359e0d61025cff3612c7e95fd43165de8da9e5.

driver-core/src/main/com/mongodb/internal/async/function/RetryControl.java
stIncMale: Done in 6b8d5767af8669a020c765666ded8a0c626de92a.

driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
```diff
@@ -462,6 +521,107 @@ public String toString() {
...
+                            ? IndividualPolicies.Descriptor.OVERLOAD.maxRetries
+                            : maxAdaptiveRetriesSetting;
+                }
+
+                void onAttemptStart(final RetryContext retryContext, final BaseClientSessionImpl.OverloadRetryPolicyState sessionScopedState) {
+                    if (retryContext.isFirstAttempt()) {
+                        sessionScopedState.openCommandExecutionScope();
+                    }
+                    this.sessionScopedState = sessionScopedState;
+                }
```
stIncMale, line 554: `VAKOTODO` Move `this.sessionScopedState = sessionScopedState;` at the start of the method body.

stIncMale: Done in ac231b3e8ad43601f92ac6cd24fa083fbc8d866c.

driver-core/src/main/com/mongodb/session/ClientSession.java
```diff
@@ -34,17 +36,17 @@
  * @see ClientSessionOptions
  */
 @NotThreadSafe
+@Sealed
```
stIncMale: `VAKOTODO` Remove `@Sealed` here and in other `ClientSession` interfaces.

This change must only be done in a major release. Furthermore, we may need to change the design of our session API in a way that does not necessitate exposing internal methods in the public API.

stIncMale: Done in d684f1fee22d77b4041c8bac7bd366ccc57a34ec.

driver-core/src/main/com/mongodb/internal/session/BaseClientSessionImpl.java
```diff
@@ -249,4 +263,301 @@ protected enum TransactionState {
...
+         * @see #getCommandExecutionScoped()
+         */
+        void openCommandExecutionScope();
+
+        /**
+         * @see #openCommandExecutionScope()
+         * @see #closeCommandExecutionScope()
+         */
+        @Nullable
+        CommandExecutionScoped getCommandExecutionScoped();
```
stIncMale, line 288: `VAKOTODO` Document when it may be `null`. The same applies to `getCommitScoped`.

stIncMale: Done in 7d06cf695fca21a49fc6eea1d63b38a155c6970d.

driver-core/src/main/com/mongodb/internal/operation/ClientBulkWriteOperation.java
```diff
@@ -412,8 +417,10 @@ private ExhaustiveClientBulkWriteCommandOkResponse executeBulkWriteCommandAndExh
             return null;
         }
         ClientBulkWriteCommandOkResponse response = new ClientBulkWriteCommandOkResponse(okResponseDocument);
-        List<List<BsonDocument>> cursorExhaustBatches = retryControl.doWhileDisabled(() ->
-                exhaustBulkWriteCommandOkResponseCursor(connectionSource, operationContext, connection, response));
+        List<List<BsonDocument>> cursorExhaustBatches = retryControl.doWhileDisabled(() -> {
+            onNotTryingToStartTransaction(operationContext.getSessionContext().getOverloadRetryPolicyState());
```
vbabanin, line 421: As we discussed offline:

One suggestion regarding the `onNotTryingToStartTransaction(...)` hook currently in `ClientBulkWriteOperation`: I think we can eliminate it by introducing a wire-level success notification on `SessionContext` and calling it from the single command-execution funnel.

### Proposed shape

Add to `SessionContext`:
```java
void notifyMessageProcessedSuccessfully(); (this name is used as an example)
```

Then wire it into `CommandProtocolImpl` on the success path only (sync + async). `ok:0` responses throw from `ProtocolHelper.getCommandFailureException(...)` before the notification is reached, so the failure/retry path remains owned by the retry policy. The wire layer would not need to inspect error labels.

### Why this helps

Since every command flows through `DefaultServerConnection.command(...)` -> `CommandProtocolImpl` -> `CommandMessage`, putting the notification at that funnel means the outer bulk write command’s success is observed at the wire layer. That would naturally clear the flag before any inner `getMore` is encoded.

In this case no per-operation hook would be required, and any future operation that fans out sub-commands via `doWhileDisabled(...)` benefits automatically.

Draft commit: https://github.com/mongodb/mongo-java-driver/commit/a2a9d5dfcaba10b713457dc816900f6005c0e0f7

**It is non-blocking we can address it later.**

driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
```diff
@@ -462,6 +523,107 @@ public String toString() {
...
+                    return labelInfo;
+                }
+
+                private void onAnyAttemptFailure(final boolean retryableOverloadError) {
+                    BaseClientSessionImpl.OverloadRetryPolicyState localSessionScopedState = assertNotNull(sessionScopedState);
+                    assertNotNull(localSessionScopedState.getCommandExecutionScoped()).onAnyAttemptFailure(retryableOverloadError);
+                    BaseClientSessionImpl.OverloadRetryPolicyState.CommitScoped commitScopedState = localSessionScopedState.getCommitScoped();
+                    if (commitScopedState != null) {
+                        commitScopedState.onAnyAttemptFailure(retryableOverloadError);
+                    }
```
vbabanin, line 572: As we discussed offline: 

Instead of exposing `getOverloadRetryPolicyState()` from `SessionContext` we could consider exposing flat lifecycle-notification methods on `SessionContext`.

For example:

```java
void notifyCommandExecutionStarted();                     
void notifyCommandExecutionEnded(); 
void notifyAttemptFailed(boolean retryableOverloadError); 
```

Then `SessionContext` would own the routing to `CommandExecutionScoped` internally.

This would improve encapsulation in a few ways:

- `SpecRetryPolicy` would no longer need to import or know about `OverloadRetryPolicyState` / `CommandExecutionScoped`; it would depend only on `SessionContext`.
- Moving the routing into the `SessionContext` keeps the invariants session-internal.

**It is non-blocking we can address it later.**

vbabanin:
LGTM!

vbabanin (approved):

driver-sync/src/test/functional/com/mongodb/client/unified/UnifiedTestModifications.java
```diff
@@ -565,14 +565,6 @@ public static void applyCustomizations(final TestDef def) {
...
                 .test("transactions", "client bulkWrite transactions",
                         "client bulkWrite with writeConcern in a transaction causes a transaction error");
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5956 TODO-JAVA-5956")
-                .file("transactions", "backpressure-retryable-writes");
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5956 TODO-JAVA-5956")
-                .file("transactions", "backpressure-retryable-reads");
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5956 TODO-JAVA-5956")
-                .file("transactions", "backpressure-retryable-commit");
-        def.skipJira("https://jira.mongodb.org/browse/JAVA-5956 TODO-JAVA-5956")
-                .file("transactions", "backpressure-retryable-abort");
```
vbabanin, line 575: Given that we enabled these tests and the specifications submodule includes unified tests for [JAVA-6117](https://jira.mongodb.org/browse/JAVA-6117), the consider [JAVA-6117](https://jira.mongodb.org/browse/JAVA-6117) addressed in this scope.

PR #2042, opened 2026-08-20
Java 6173 CSFLE/QE Support for HTTP Proxies

strogiyotec:
[Java 6173](https://jira.mongodb.org/browse/JAVA-6173) CSFLE/QE Support for HTTP Proxies
[Spec](https://github.com/mongodb/specifications/blob/92b3c0b9287bfba1b0ec4084300858d05c654f8c/source/client-side-encryption/client-side-encryption.md#kmsconnectcallback) 

This implementation does not support reactive driver because the ProxySetting do not support reactive driver either see [jira.mongodb.org/browse/JAVA-5205](https://jira.mongodb.org/browse/JAVA-5205) 

### Why this does not implement `kmsConnectCallback`
The spec makes the callback optional:

> Drivers are required to support an HTTP proxy but MAY omit `kmsConnectCallback` if they provide an alternative mechanism for proxy support.

This PR takes that route and configures the proxy declaratively instead. Here is why:
The callback returns a `java.net.Socket`, which cannot be turned into a `SocketChannel` (`Socket.getChannel()` is null, and an `SSLSocket` never is one), so `TlsChannel` cannot accept it. 

### API

```java
ClientEncryptionSettings.builder()
    .keyVaultNamespace("keyvault.datakeys")
    .kmsProviders(kmsProviders)
    .proxySettings(ProxySettings.builder()
            .protocol(ProxyProtocol.HTTP)      // or HTTPS to reach the proxy over TLS
            .host("proxy.example.com")
            .port(8080)
            .build())
    .build();
```
I decided to reuse `ProxySetting` even though it was designed for socks5 mongo server communication only, the name itself suggests we should treat it as a regular proxy settings which include http proxy 

## TODO
1. ~I added support for proxy auth https://www.rfc-editor.org/info/rfc9110/#section-11.7.2 though spec doesn't say we need to support it , I dm Kevin~ (looks like the auth was rejected, I left a link to the discussion doc in the jira ticket)
2. Need to follow up with docsp to show code samples

## AI Policy
Claude helped me to implement `KmsSocketConnector` and `HttpProxyTunnel`

### DIFF
--- a/driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
+++ b/driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
@@ -16,25 +16,21 @@
 
 package com.mongodb.client.internal;
 
+import com.mongodb.KmsConnectCallback;
 import com.mongodb.ServerAddress;
 import com.mongodb.internal.TimeoutContext;
-import com.mongodb.internal.connection.SslHelper;
+import com.mongodb.internal.capi.KmsSocketConnector;
 import com.mongodb.internal.diagnostics.logging.Logger;
 import com.mongodb.internal.diagnostics.logging.Loggers;
 import com.mongodb.internal.time.Timeout;
 import com.mongodb.lang.Nullable;
 import com.mongodb.lang.NonNull;
 
-import javax.net.SocketFactory;
 import javax.net.ssl.SSLContext;
-import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocket;
-import javax.net.ssl.SSLSocketFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.net.SocketException;
 import java.nio.ByteBuffer;
@@ -48,10 +44,14 @@ import static java.util.concurrent.TimeUnit.NANOSECONDS;
 class KeyManagementService {
     private static final Logger LOGGER = Loggers.getLogger("client");
     private final Map<String, SSLContext> kmsProviderSslContextMap;
+    @Nullable
+    private final KmsConnectCallback kmsConnectCallback;
     private final int timeoutMillis;
 
-    KeyManagementService(final Map<String, SSLContext> kmsProviderSslContextMap, final int timeoutMillis) {
+    KeyManagementService(final Map<String, SSLContext> kmsProviderSslContextMap,
+            @Nullable final KmsConnectCallback kmsConnectCallback, final int timeoutMillis) {
         this.kmsProviderSslContextMap = notNull("kmsProviderSslContextMap", kmsProviderSslContextMap);
+        this.kmsConnectCallback = kmsConnectCallback;
         this.timeoutMillis = timeoutMillis;
     }
 
@@ -61,18 +61,15 @@ class KeyManagementService {
         LOGGER.info("Connecting to KMS server at " + serverAddress);
         SSLContext sslContext = kmsProviderSslContextMap.get(kmsProvider);
 
-        SocketFactory sslSocketFactory = sslContext == null
-                    ? SSLSocketFactory.getDefault() : sslContext.getSocketFactory();
-        SSLSocket socket = (SSLSocket) sslSocketFactory.createSocket();
-        enableHostNameVerification(socket);
+        SSLSocket socket = KmsSocketConnector.connect(sslContext, kmsConnectCallback, kmsProvider, serverAddress,
+                timeoutMillis, remainingMillis(operationTimeout));
 
-        try {
-            socket.setSoTimeout(timeoutMillis);
-            socket.connect(new InetSocketAddress(InetAddress.getByName(serverAddress.getHost()), serverAddress.getPort()), timeoutMillis);
-        } catch (IOException e) {
+        // A KmsConnectCallback performs blocking I/O and may overrun the deadline it was given, which cannot be
+        // preempted. Re-check before issuing a request whose time budget is already spent.
+        Timeout.nullAsInfinite(operationTimeout).onExpired(() -> {
             closeSocket(socket);
-            throw e;
-        }
+            TimeoutContext.throwMongoTimeoutException("Connecting to KMS server exceeded the timeout limit.");
+        });
 
         try {
             OutputStream outputStream = socket.getOutputStream();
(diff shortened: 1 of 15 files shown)


driver-core/src/main/com/mongodb/connection/ProxySettings.java
```diff
@@ -31,16 +32,36 @@
 import static com.mongodb.assertions.Assertions.notNull;
 
 /**
- * This setting is only applicable when communicating with a MongoDB server using the synchronous variant of {@code MongoClient}.
- * <p>
- * This setting is furthermore ignored if:
+ * The settings for reaching a destination through a proxy server.
+ *
+ * <p>These settings are applied in two independent places, which support different
+ * {@linkplain ProxyProtocol protocols}:</p>
```
strogiyotec: the name of this class is `ProxySetting` though it was only used as a socks5 proxy to mongo server

Instead of duplicating similar logic I introduced an inner protocol which defaults to `socks5` so that existing code base won't have to be changed, but kms proxy can change protocol to https

driver-core/src/main/com/mongodb/connection/ProxyProtocol.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * <p>This protocol is currently supported only for Key Management Service (KMS) requests made by in-use
+     * encryption, configured via {@link ClientEncryptionSettings.Builder#proxySettings(ProxySettings)} or
+     * {@link AutoEncryptionSettings.Builder#proxySettings(ProxySettings)}. Configuring it for connections to a
+     * MongoDB server is rejected when the client is created.</p>
+     *
+     * <p>A port must be specified explicitly with {@link ProxySettings.Builder#port(int)}, as there is no
+     * standard port for an HTTP proxy.</p>
+     */
+    HTTP,
```
strogiyotec: I was thinking about dropping http support and only keep https but based on this [comment ](https://jira.mongodb.org/browse/DRIVERS-2920?focusedCommentId=8471526&focusedId=8471526&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-8471526) customer is using a plain http

driver-core/src/main/com/mongodb/connection/ProxySettings.java
```diff
@@ -59,6 +80,9 @@ public final class ProxySettings {
     private final String username;
     @Nullable
     private final String password;
+    private final ProxyProtocol protocol;
```
strogiyotec: @vbabanin I need your opinion, instead of going with a callback as recommended by the [spec ](https://github.com/mongodb/specifications/blob/92b3c0b9287bfba1b0ec4084300858d05c654f8c/source/client-side-encryption/client-side-encryption.md#kmsconnectcallback)
 though not mandatory `The kmsConnectCallback property MAY be specified`
I went with modifying the ProxySettings class

This class was designed to support SOCKS5 proxy with mongo server , but the naming suggests it should work as a regular proxy so I decided to modify this class to carry the SSLContext and the protocol which default to Socks5 

Do you think it's a good choice ?

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
```diff
@@ -48,10 +44,13 @@
...
     private static final Logger LOGGER = Loggers.getLogger("client");
     private final Map<String, SSLContext> kmsProviderSslContextMap;
+    private final ProxySettings proxySettings;
     private final int timeoutMillis;
 
-    KeyManagementService(final Map<String, SSLContext> kmsProviderSslContextMap, final int timeoutMillis) {
+    KeyManagementService(final Map<String, SSLContext> kmsProviderSslContextMap,
+            final ProxySettings proxySettings, final int timeoutMillis) {
         this.kmsProviderSslContextMap = notNull("kmsProviderSslContextMap", kmsProviderSslContextMap);
+        this.proxySettings = notNull("proxySettings", proxySettings);
```
strogiyotec: the ProxySettings must be present but to use non proxy the `isProxyEnabled` will return false

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
```diff
@@ -61,18 +60,15 @@ public InputStream stream(final String kmsProvider, final String host, final Byt
         LOGGER.info("Connecting to KMS server at " + serverAddress);
         SSLContext sslContext = kmsProviderSslContextMap.get(kmsProvider);
 
-        SocketFactory sslSocketFactory = sslContext == null
-                    ? SSLSocketFactory.getDefault() : sslContext.getSocketFactory();
-        SSLSocket socket = (SSLSocket) sslSocketFactory.createSocket();
```
strogiyotec, line 66: this removed logic was moved into https://github.com/mongodb/mongo-java-driver/pull/2042/changes#diff-b17d0a25554cd6a81a2942131b25338c22e88e9e9bd5f5eb30b6794d969d42f2R75

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
```diff
@@ -94,13 +90,20 @@ public InputStream stream(final String kmsProvider, final String host, final Byt
         }
     }
 
-    private void enableHostNameVerification(final SSLSocket socket) {
```
strogiyotec, line 97: this method was moved to [here](https://github.com/strogiyotec/mongo-java-driver/blob/e7d726a0974be7cf0f01cb2802b8e0f0f901f128/driver-core/src/main/com/mongodb/internal/capi/KmsSocketConnector.java#L75-L80)

However I removed 
```
 if (sslParameters == null) {
            sslParameters = new SSLParameters();
        }
```
because according to the [javadoc](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLSocket.html#getSSLParameters--)
> Returns the SSLParameters in effect for this SSLSocket. The ciphersuites and protocols of the returned SSLParameters are always non-null.

so it's never null

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
```diff
@@ -94,13 +90,20 @@ public InputStream stream(final String kmsProvider, final String host, final Byt
...
+    /**
+     * Determines the time available for reaching the KMS server, which the specification requires to be the time
+     * remaining in the operation when CSOT is in use.
+     *
+     * <p>Visible for testing.</p>
+     *
+     * @return the remaining time available for connecting to the KMS server, in milliseconds, never larger than the
+     * configured connect timeout.
+     */
+    long remainingMillis(@Nullable final Timeout operationTimeout) {
```
strogiyotec, line 103: from the [spec](https://github.com/mongodb/specifications/blob/92b3c0b9287bfba1b0ec4084300858d05c654f8c/source/client-side-encryption/client-side-encryption.md#kmsconnectcallback)
> Drivers supporting CSOT MUST pass remaining timeoutMS

driver-core/src/main/com/mongodb/internal/connection/HttpProxyTunnel.java
```diff
@@ -0,0 +1,147 @@
...
+     * @param target the host the tunnel should reach
+     * @param proxySettings the proxy settings, used for optional {@code Basic} authentication
+     * @throws IOException if the request cannot be written, the response cannot be read, or the proxy declines
+     */
+    public static void establishTunnel(final Socket socket, final ServerAddress target,
+            final ProxySettings proxySettings) throws IOException {
+        writeConnectRequest(socket.getOutputStream(), target, proxySettings);
+        String statusLine = readResponse(socket.getInputStream(), target);
+        int statusCode = parseStatusCode(statusLine, target);
+        if (statusCode / 100 != 2) {
```
strogiyotec: this looks weird , but according to the connect spec https://www.rfc-editor.org/info/rfc9110/#name-connect
> Any [2xx (Successful)](https://www.rfc-editor.org/info/rfc9110/#status.2xx) response indicates that the sender (and all inbound proxies) will switch to tunnel mode immediately after the response header section

driver-core/src/main/com/mongodb/internal/connection/HttpProxyTunnel.java
```diff
@@ -0,0 +1,147 @@
...
+ * end-to-end with the target host over the same socket.</p>
+ *
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public final class HttpProxyTunnel {
+
+    /**
+     * Bounds the response so that a proxy which never sends the end of the header block cannot cause an unbounded read.
+     */
+    private static final int MAX_RESPONSE_BYTES = 8192;
```
strogiyotec: https://www.rfc-editor.org/info/rfc9110/#section-4.1-3
> It is RECOMMENDED that all senders and recipients support, at a minimum, URIs with lengths of 8000 octets in protocol elements

PR #2043, opened 2026-08-25
Java 6173 CSFLE/QE Support for HTTP Proxies 

strogiyotec:
[Java 6173](https://jira.mongodb.org/browse/JAVA-6173) CSFLE/QE Support for HTTP Proxies

The first iteration - https://github.com/mongodb/mongo-java-driver/pull/2042 I completely missed the scope doc that suggests that implementing http proxy + http proxy connect in the driver side in discouraged 

This iteration implements the callback , specifically it adds `kmsConnectCallback`: the application supplies the connection, and the driver negotiates TLS with the KMS host over it.

### API

```java
KmsConnectCallback viaProxy = context -> {
    Socket socket = new Socket();
    int timeout = (int) context.getTimeoutMillis();
    socket.connect(new InetSocketAddress("proxy.example.com", 8080), timeout);
    socket.setSoTimeout(timeout);

    String target = context.getServerAddress().getHost() + ":" + context.getServerAddress().getPort();
    socket.getOutputStream().write(
            ("CONNECT " + target + " HTTP/1.1\r\nHost: " + target + "\r\n\r\n").getBytes(US_ASCII));
    readAndCheckProxyResponse(socket.getInputStream());   // application's code

    return socket;   // the driver adds TLS
};

ClientEncryptionSettings.builder()
        .keyVaultNamespace("keyvault.datakeys")
        .kmsProviders(kmsProviders)
        .kmsConnectCallback(viaProxy)
        .build();
```

This implementation does not support reactive driver because the ProxySetting does not support reactive driver either see [jira.mongodb.org/browse/JAVA-5205](https://jira.mongodb.org/browse/JAVA-5205)

AI Policy

Claude helped me to implement `KmsSocketConnector.java` and write test cases


driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
```diff
@@ -61,18 +61,15 @@ public InputStream stream(final String kmsProvider, final String host, final Byt
         LOGGER.info("Connecting to KMS server at " + serverAddress);
         SSLContext sslContext = kmsProviderSslContextMap.get(kmsProvider);
 
-        SocketFactory sslSocketFactory = sslContext == null
-                    ? SSLSocketFactory.getDefault() : sslContext.getSocketFactory();
-        SSLSocket socket = (SSLSocket) sslSocketFactory.createSocket();
-        enableHostNameVerification(socket);
```
strogiyotec, line 67: this method was moved to [here](https://github.com/mongodb/mongo-java-driver/pull/2043/changes#diff-b17d0a25554cd6a81a2942131b25338c22e88e9e9bd5f5eb30b6794d969d42f2R77)

However I removed
```
 if (sslParameters == null) {
            sslParameters = new SSLParameters();
        }
```
because according to the [javadoc](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLSocket.html#getSSLParameters--)

> Returns the SSLParameters in effect for this SSLSocket. The ciphersuites and protocols of the returned SSLParameters are always non-null.

so it's never null

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
```diff
@@ -94,13 +91,20 @@ public InputStream stream(final String kmsProvider, final String host, final Byt
...
+    /**
+     * Determines the time available for reaching the KMS server, which the specification requires to be the time
+     * remaining in the operation when CSOT is in use.
+     *
+     * <p>Visible for testing.</p>
+     *
+     * @return the remaining time available for connecting to the KMS server, in milliseconds, never larger than the
+     * configured connect timeout.
+     */
+    long remainingMillis(@Nullable final Timeout operationTimeout) {
```
strogiyotec: from the [spec](https://github.com/mongodb/specifications/blob/92b3c0b9287bfba1b0ec4084300858d05c654f8c/source/client-side-encryption/client-side-encryption.md#kmsconnectcallback)

    Drivers supporting CSOT MUST pass remaining timeoutMS

driver-core/src/main/com/mongodb/KmsConnectCallback.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * <p>When a callback is configured, the driver invokes it instead of connecting to the KMS host itself, and uses the
+ * socket it returns for the KMS request. This enables routing KMS requests through an intermediary, most commonly an
+ * HTTP proxy via the {@code HTTP CONNECT} method.</p>
+ *
+ * <p>The driver always negotiates TLS with the KMS host itself over the returned socket, using the
+ * {@link javax.net.ssl.SSLContext} configured for the KMS provider. Server Name Indication and certificate hostname
+ * verification target {@link KmsConnectContext#getServerAddress()}, not the address the callback actually connected to.
+ * Implementations therefore MUST NOT negotiate TLS with the KMS host themselves; they must return a socket over which
+ * a TLS handshake with the KMS host can be performed. An implementation may use TLS for its own connection to an
```
jyemin, line 36: I didn't know this was possible.

driver-core/src/main/com/mongodb/KmsConnectContext.java
```diff
@@ -0,0 +1,98 @@
...
+
+import com.mongodb.annotations.Immutable;
+
+import static com.mongodb.assertions.Assertions.isTrueArgument;
+import static com.mongodb.assertions.Assertions.notNull;
+
+/**
+ * The details of a Key Management Service (KMS) connection that a {@link KmsConnectCallback} is asked to establish.
+ *
+ * <p>Instances are created by the driver. Further properties may be added in future releases, so implementations of
```
jyemin: In practice, what does this mean?  What do we expect a callback to do in the face of a driver upgrade that adds a new field to the context?

driver-core/src/main/com/mongodb/KmsConnectContext.java
```diff
@@ -0,0 +1,98 @@
...
+     * @return the address of the KMS host, never null
+     */
+    public ServerAddress getServerAddress() {
+        return serverAddress;
+    }
+
+    /**
+     * Gets the time available to establish the connection, in milliseconds.
+     *
+     * <p>When a timeout is configured this is the time remaining in the operation, so it shrinks as the operation
```
jyemin: This is referring to CSOT?  If so, this is dangerous because CSOT remains alpha and is going to be removed rather than be made GA.

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
jyemin: I checked history and JAVA-2876 is the reason for the null check.  Safer to keep it I would say.

driver-sync/src/main/com/mongodb/client/internal/KmsSocketConnector.java
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * <p>When a {@link KmsConnectCallback} is configured, the callback establishes the underlying connection, which may be
+ * to an intermediary such as an HTTP proxy, and TLS for the KMS host is layered on top of the socket it returns. TLS is
+ * always negotiated end-to-end with the KMS host: Server Name Indication and certificate hostname verification are
+ * configured from the KMS address rather than from wherever the socket is actually connected, so an intermediary
+ * relays the session without being able to read it.</p>
+ *
+ * <p>This class is not part of the public API and may be removed or changed at any time</p>
+ */
+public final class KmsSocketConnector {
```
jyemin, line 47: Can this be moved to driver-sync?  I don't see anything in here that could be re-used for the reactive driver.

driver-core/src/test/unit/com/mongodb/internal/capi/KmsSocketConnectorTunnelTest.java
```diff
@@ -0,0 +1,476 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Verifies that the driver layers TLS for the KMS host on the socket a {@link KmsConnectCallback} returns, including
```
jyemin: Does this test anything not already tested by the functional test?

driver-core/src/main/com/mongodb/KmsConnectCallback.java
```diff
@@ -0,0 +1,85 @@
...
+ * };
+ * }</pre>
+ *
+ * @see ClientEncryptionSettings.Builder#kmsConnectCallback(KmsConnectCallback)
+ * @see AutoEncryptionSettings.Builder#kmsConnectCallback(KmsConnectCallback)
+ * @since 5.11
+ */
+@ThreadSafe
+@FunctionalInterface
+public interface KmsConnectCallback {
```
jyemin, line 81: I'm concerned that this callback interface won't be re-usable for the reactive driver, since in reactive I don't think we would want the application to return a `Socket`. Instead, it would have to return a `SocketChannel` in non-blocking mode, I imagine.  But that call in itself would have to be async and would have to return  `Publisher<SocketChannel>` instead.

So there are a few options:

1. Commit to never supporting the reactive driver
2. Add a second interface if we ever support the reactive driver
3. Make this a non-functional interface, and add a second method for reactive, either now or in the future.

driver-core/src/main/com/mongodb/KmsConnectContext.java
```diff
@@ -0,0 +1,98 @@
...
+ * @since 5.11
+ */
+@Immutable
+public final class KmsConnectContext {
+    private final String kmsProvider;
+    private final ServerAddress serverAddress;
+    private final long timeoutMillis;
+
+    /**
+     * This constructor is not part of the public API and may be removed or changed at any time.
```
jyemin: Is this a pattern we use in other similar places?  It's a bit awkward.

driver-core/src/main/com/mongodb/KmsConnectCallback.java
jyemin, line 81: There's a test failure relating to this new API: https://parsley.corp.mongodb.com/test/mongo_java_driver_scala_tests__version~7.0_os~ubuntu_topology~replicaset_auth~noauth_ssl~nossl_jdk~jdk17_scala~2.13_scala_test_task_patch_9af9be1fa428effe0ea10e05c0e1256d2286ec3c_6a8d13bb56debd00076a11bb_26_08_25_04_02_07/0/8da065851070e91f08c4e66ae00ff2bc?bookmarks=0,46&shareLine=0

jyemin (changes requested):

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
jyemin: Also, there's now a stale SpotBugs exclude for `KeyManagementService#enableHostNameVerification` that can be removed.

driver-core/src/main/com/mongodb/KmsConnectCallback.java
strogiyotec: this is according to the spec
https://github.com/mongodb/specifications/blob/92b3c0b9287bfba1b0ec4084300858d05c654f8c/source/client-side-encryption/client-side-encryption.md#kmsconnectcallback
> The driver then wraps the returned socket with TLS (using the KMS provider's configured [TLS options](https://github.com/mongodb/specifications/blob/92b3c0b9287bfba1b0ec4084300858d05c654f8c/source/client-side-encryption/client-side-encryption.md#kms-provider-tls-options))

I also didn't know we can do it

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
strogiyotec: I updated the spotbug [here](https://github.com/mongodb/mongo-java-driver/pull/2043/changes#diff-015f0d0522426e70c6b5fdaacdaa07550e15d229072299a61830a9119f7f6a0cL77) 
and kept the null check with added [comment ](https://github.com/mongodb/mongo-java-driver/pull/2043/changes#diff-70a4b02c1cb1ae95a178d0a04da3ac0343e39a7a8a1bbc063ba0c41e95d37781R129)to explain why we have it

driver-core/src/main/com/mongodb/KmsConnectContext.java
strogiyotec: so far spec only suggests to pass the kms address and our implementation passes the same context as c# does 
https://github.com/adelinowona/mongo-csharp-driver/blob/916a6b57cbaf2783040b6fd251121225b7566a0b/src/MongoDB.Driver/Encryption/KmsConnectionContext.cs 
I thought maybe add this sentence in case spec will be changed , for now the doc was removed

driver-sync/src/main/com/mongodb/client/internal/KmsSocketConnector.java
strogiyotec: [Moved ](https://github.com/mongodb/mongo-java-driver/pull/2043/changes#diff-70a4b02c1cb1ae95a178d0a04da3ac0343e39a7a8a1bbc063ba0c41e95d37781) the class to the sync driver

driver-core/src/test/unit/com/mongodb/internal/capi/KmsSocketConnectorTunnelTest.java
strogiyotec: removed the test case, everything is covered by the functional test already ,originally  I asked claude to have test cases for every new class

driver-core/src/main/com/mongodb/KmsConnectCallback.java
strogiyotec: you are right this won't work with a reactive driver, to justify it I have two arguments
1. `ProxySetting` never worked with a reactive driver
2. The client who wants this feature uses sync driver 
This interface can be accessed from the reactive driver so I added a [guard ](https://github.com/mongodb/mongo-java-driver/pull/2043/changes#diff-66fc2324936ae09d3734167754e38b6f0d8ed18e8098406b7000ed131549ea93R89) that will throw an error if users try to pass the callback

driver-core/src/main/com/mongodb/KmsConnectContext.java
strogiyotec: that's the only place with this comment, I replaced it with a regular `Construct an instance`

driver-core/src/main/com/mongodb/KmsConnectCallback.java
strogiyotec: Thanks, I didn't notice the failure in scala driver 
Addressed it [here ](https://github.com/mongodb/mongo-java-driver/pull/2043/changes#diff-675a731dbf71157d1a4c026894b87d721d0d7a84878530650f75ec5aa5075623)

driver-core/src/main/com/mongodb/KmsConnectCallback.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.IOException;
+import java.net.Socket;
+
+/**
+ * A callback that establishes the connection used for a Key Management Service (KMS) request made by in-use encryption
+ * (client-side field level encryption or queryable encryption).
+ *
+ * <p>When a callback is configured, the driver invokes it instead of connecting to the KMS host itself, and uses the
+ * socket it returns for the KMS request. This enables routing KMS requests through an intermediary, most commonly an
+ * HTTP proxy via the {@code HTTP CONNECT} method.</p>
```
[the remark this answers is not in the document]
strogiyotec: it's not finalized yet

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsConnectCallbackProseTest.java
```diff
@@ -0,0 +1,432 @@
...
+            }
+
+            // read with the unencrypted client to confirm the value is stored encrypted
+            Document stored = client.getDatabase("db").getCollection("coll").find(Filters.eq("_id", 1)).first();
+            assertNotNull(stored);
+            assertInstanceOf(org.bson.types.Binary.class, stored.get("encrypted_string"));
+        }
+
+        // only one KMS request is expected, since the decrypted key is cached
+        assertTrue(getConnectCount(false) >= 1, "expected KMS requests to be routed through the proxy");
```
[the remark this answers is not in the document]
strogiyotec: fixed

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsConnectCallbackProseTest.java
```diff
@@ -0,0 +1,449 @@
...
+ * <p>All cases require real AWS KMS credentials and are skipped when they are not available. The KMS HTTP proxy is
+ * started by {@code drivers-evergreen-tools}: {@code .evergreen/csfle/start-servers.sh} runs {@code kms_http_proxy.py}
+ * on port 9004 in plain HTTP mode and on port 9005 in HTTPS mode. Cases are skipped when the proxy is unreachable, so
+ * that this test does not fail when run outside that environment.</p>
+ *
+ * <p>Case 6, "Retry", is not implemented: it is to be skipped by drivers that do not implement DRIVERS-1541, and this
+ * driver does not retry KMS requests.</p>
+ *
+ * @see <a href="https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.md#28-kms-connect-callback">
+ * Prose test 28</a>
```
strogiyotec, line 79: https://github.com/mongodb/specifications/blob/92b3c0b9287bfba1b0ec4084300858d05c654f8c/source/client-side-encryption/tests/README.md#28-kms-connect-callback

driver-core/src/main/com/mongodb/KmsConnectCallback.java
jyemin: Cool.  Is the behavior tested?

driver-core/src/main/com/mongodb/KmsConnectCallback.java
jyemin: So you're choosing option 2: add a second interface if we ever support the reactive driver ?

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
jyemin: Right, but we're _not_ supporting CSOT.  All CSOT-related API is `@Alpha` and will be removed soon.

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsTlsTest.java
```diff
@@ -133,6 +135,51 @@ public void testInvalidKmsCertificate() {
...
     }
 
+    /**
+     * The driver negotiates TLS with the KMS host over the socket a {@link KmsConnectCallback} returns, so certificate
+     * hostname verification must be applied on that path too. A callback that connects straight to the KMS host is
+     * enough to exercise it: the certificate served on port 9001 is issued for a different host, so the handshake the
+     * driver performs must fail.
+     */
+    @Test
+    public void testInvalidKmsCertificateWhenConnectingViaCallback() {
```
jyemin: Is this a test that is not in the spec?  If so, why is it needed? And why is it in this file and not the new file with all the other callback-related tests?

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsConnectCallbackProseTest.java
```diff
@@ -0,0 +1,449 @@
...
+        } finally {
+            connection.disconnect();
+        }
+    }
+
+    // --- the proxy's CA ----------------------------------------------------------------------------------------
+
+    private static volatile SSLContext proxySslContext;
+
+    private static SSLContext proxySslContext() {
```
jyemin, line 370: I don't think the double-checked locking idiom is required here.  We don't run any of our tests in parallel.

driver-core/src/main/com/mongodb/KmsConnectCallback.java
jyemin: It is, I checked.

driver-sync/src/main/com/mongodb/client/internal/KmsSocketConnector.java
```diff
@@ -0,0 +1,154 @@
...
+            throw e;
+        }
+
+        try {
+            // Even though the callback's connection is already established, the TLS handshake has not been performed
+            // yet, so SSL parameters can still be set. They target the KMS host, not any intermediary.
+            enableHostNameVerification(socket, kmsAddress.getHost());
+            socket.setSoTimeout(soTimeoutMillis);
+            // Handshake explicitly so that a TLS failure is reported here rather than on the first write.
+            socket.startHandshake();
```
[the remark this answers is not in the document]
strogiyotec: CSOT was removed so not relevant anymore

driver-core/src/main/com/mongodb/KmsConnectCallback.java
```diff
@@ -0,0 +1,85 @@
...
+ *     socket.getOutputStream().write(
+ *             ("CONNECT " + target + " HTTP/1.1\r\nHost: " + target + "\r\n\r\n").getBytes(StandardCharsets.US_ASCII));
+ *
+ *     // Read the status line and confirm a 2xx status, throwing an IOException otherwise. Match the status code
+ *     // rather than the whole status line, as proxies differ in the HTTP version they reply with. Read the response
+ *     // one byte at a time, up to the end of the header block, so that no byte of the TLS handshake that the driver
+ *     // performs over this socket is consumed.
+ *     readAndCheckProxyResponse(socket.getInputStream());
+ *
+ *     return socket;
```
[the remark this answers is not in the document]
strogiyotec: I think it's fine for a public example , we don't have to change it here but rather in the docsp

driver-core/src/main/com/mongodb/KmsConnectCallback.java
strogiyotec: I went with option 2 , the second interface will be introduced (if ever) to support reactive proxy

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
strogiyotec: ~I removed the `remainingMillis` ~

driver-core/src/main/com/mongodb/KmsConnectContext.java
strogiyotec: I removed the CSOT from the ~callback~  from the context

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsConnectCallbackProseTest.java
strogiyotec: removed the double checked locking

driver-sync/src/test/functional/com/mongodb/client/AbstractClientSideEncryptionKmsTlsTest.java
strogiyotec: the test was originally created to test the case  when the driver still verifies the KMS certificate when it layers TLS over a socket the callback created
For now I removed it and will follow up with Kevin if it's needed, if it's there will be a follow up PR to add this test , until then I don't see a reason to keep it

driver-sync/src/main/com/mongodb/client/internal/KeyManagementService.java
strogiyotec: restored remainingMillis when CSOT is in use while the Context doesn't have a timeout anymore

jyemin (approved):
LGTM!

PR #2044, opened 2026-08-25
Add overload retry policy for runCommand operations

vbabanin:
- Extend SpecRetryPolicy so the overload sub-policy can carry write-shape error propagation independently of includeWrite(), enabling runCommand's client-backpressure rules without pulling in unwanted write-retry decisions
- Split CommandReadOperation into an abstract base plus a concrete runCommand subclass that composes overload-only retries gated on retryReads && retryWrites; explain operations continue to use the base with no retry

JAVA-6294

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/AggregateOperation.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/AggregateOperation.java
@@ -157,7 +157,7 @@ public class AggregateOperation<T> implements ReadOperationExplainable<T> {
         return createExplainableOperation(verbosity, resultDecoder);
     }
 
-    <R> CommandReadOperation<R> createExplainableOperation(@Nullable final ExplainVerbosity verbosity, final Decoder<R> resultDecoder) {
+    <R> ExplainCommandOperation<R> createExplainableOperation(@Nullable final ExplainVerbosity verbosity, final Decoder<R> resultDecoder) {
         return new ExplainCommandOperation<>(getNamespace().getDatabaseName(), getCommandName(),
                 (operationContext, serverDescription, connectionDescription) -> {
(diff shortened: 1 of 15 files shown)


driver-core/src/main/com/mongodb/internal/operation/AggregateOperation.java
```diff
@@ -157,7 +157,7 @@ public <R> ReadOperationSimple<R> asExplainableOperation(@Nullable final Explain
         return createExplainableOperation(verbosity, resultDecoder);
     }
 
-    <R> CommandReadOperation<R> createExplainableOperation(@Nullable final ExplainVerbosity verbosity, final Decoder<R> resultDecoder) {
+    <R> ExplainCommandOperation<R> createExplainableOperation(@Nullable final ExplainVerbosity verbosity, final Decoder<R> resultDecoder) {
```
vbabanin, line 160: Just a small cleanup.

driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
```diff
@@ -648,4 +687,13 @@ private int maxAttempts(final IndividualPolicies policies) {
             }
         }
     }
+
+    /**
+     * Selects the error propagation shape for overload-only policy
+     * compositions (see {@link IndividualPolicies#includeOverload(Integer, ErrorPropagation)}).
+     */
+    enum ErrorPropagation {
+        AS_READ_POLICY,
```
vbabanin, line 696: `ErrorPropagation.AS_READ_POLICY` is needed for the next ticket / follow-up PR: JAVA-6296.

jyemin (approved):
LGTM!

PR #2052, opened 2026-09-10
Add baseBackoffMS support

vbabanin:
## Summary

This PR brings the driver's Client Backpressure implementation up to date with the latest
specification changes: the server-supplied `baseBackoffMS` override, the corrected
exponential backoff formula exponent, and the versioned `backpressure: "2"` handshake
flag. Also adds MongoDB 9.0 to the Evergreen version matrix so the new 9.0-gated
prose test has a CI variant to run against.

## Tickets

| JAVA | DRIVERS | Description |
|------|---------|-------------|
| [JAVA-6238](https://jira.mongodb.org/browse/JAVA-6238) | [DRIVERS-3535](https://jira.mongodb.org/browse/DRIVERS-3535) | Client Backpressure with baseBackoffMS — backoff formula exponent, `baseBackoffMS` server override, handshake `backpressure: "2"` |
| [JAVA-6255](https://jira.mongodb.org/browse/JAVA-6255) | [DRIVERS-3578](https://jira.mongodb.org/browse/DRIVERS-3578) | Correct sum of backoffs for the convenient transactions API backoff prose test |
| [JAVA-6265](https://jira.mongodb.org/browse/JAVA-6265) | [DRIVERS-3584](https://jira.mongodb.org/browse/DRIVERS-3584) | Provide updated retryable overload error code examples for 9.0 — **out of scope for this PR** (documentation deliverable, tracked separately as TODO) |

Upstream spec commits:
- [`d6fc118`](https://github.com/mongodb/specifications/commit/d6fc118bd1dd6854fa72f9be0ff5d6a57e8f68e1) (DRIVERS-3535)
- [`615e0f9`](https://github.com/mongodb/specifications/commit/615e0f9ca5f554614636c098225fbbf1be55565d) (DRIVERS-3578)

## Commits

| Commit | Ticket | Summary |
|--------|--------|---------|
| `a8c755e829` | JAVA-6238 | use retry number as backoff exponent |
| `407b3f0430` | JAVA-6255 | correct withTransaction backoff test sum to 2.3s |
| `ac31095dc9` | JAVA-6238 | update overload backoff prose test to 0.6s sum per new formula |
| `9b6d1a4995` | JAVA-6238 | send backpressure: "2" in handshake |
| `64d1a438b1` | JAVA-6238 | support server-supplied baseBackoffMS override for overload backoff |
| `d5377190cd` | JAVA-6238 | add baseBackoffMS override prose test (Test 5) |
| `7a42e3ac5b` | JAVA-6238 | add MongoDB 9.0 to Evergreen version matrix |
| `5472e7b055` | JAVA-6238 | extract baseBackoffMS field name to a constant |

### Backoff formula exponent (JAVA-6238 / DRIVERS-3535)

The spec changed the exponential backoff to use the retry number directly as the
exponent, instead of one less than the retry number:

- Overload retry: `jitter * min(MAX_BACKOFF, BASE_BACKOFF * 2^attempt)` (was `2^(attempt-1)`)
  → first retry 200ms, second 400ms (was 100ms / 200ms).
- Convenient transactions API: `jitter * min(BACKOFF_INITIAL * 1.5^attempt, BACKOFF_MAX)`
  (was `1.5^(attempt-1)`).

Both share `ExponentialBackoff.calculateBackoffMs`, so a single-line exponent fix
covers both. `BASE_BACKOFF` remains 100ms; `MAX_BACKOFF` remains 10000ms.

### Server-supplied `baseBackoffMS` override (JAVA-6238 / DRIVERS-3535)

When an overload error document contains a positive top-level `baseBackoffMS` field,
the driver now uses it in place of the 100ms `BASE_BACKOFF` constant:

- New `ExponentialBackoff.calculateOverloadBackoff(int, @Nullable Long baseBackoffMs)`
  overload. A `null` or non-positive value falls back to the 100ms default; a positive
  value replaces it. The 10000ms cap is preserved.
- `SpecRetryPolicy` extracts `baseBackoffMS` from `MongoCommandException.getResponse()`
  and passes it through. Extraction is lenient: absent, non-numeric, or non-positive
  values are ignored.

### Handshake `backpressure: "2"` (JAVA-6238 / DRIVERS-3535)

The handshake now sends `backpressure: "2"` — the BSON **string** `"2"`, not a boolean
and not the number 2 — to version the supported backpressure specification. Previously
the driver sent `backpressure: true`. The value is sent unconditionally on every hello
(no server-version gating), matching the Python and Node reference implementations;
the flag is currently unused by the server and older servers ignore unexpected
handshake field values.

### Test updates

- **Prose Test 1** (`BackpressureProseTest.operationRetryUsesExponentialBackoff`):
  expected overload backoff sum 0.3s → 0.6s.
- **New prose Test 5** (`BackpressureProseTest.overloadErrorsWithBaseBackoffMsOverrideBaseBackoff`):
  verifies `baseBackoffMS` overrides the default backoff. Gated on
  `serverVersionAtLeast(9, 0)`; pins jitter to 1, toggles
  `externalClientBaseBackoffMS` via `setParameter`, and asserts timing bounds
  (default run ≥ 600ms; override run ≥ 300ms and < 600ms).
- **Handshake prose Test 9** (`AbstractClientMetadataProseTest`): asserts the
  `backpressure` field is the string `"2"`.
- **withTransaction backoff test** (`WithTransactionProseTest.testRetryBackoffIsEnforced`):
  expected sum of 13 backoffs 1800ms → 2300ms.
- **Unit tests** (`ExponentialBackoffTest`): updated the transaction backoff expected
  array for the new exponent; added 4 tests covering the `baseBackoffMS` override
  (default-when-absent, positive-override, non-positive-ignored, max-cap-preserved).
- **Groovy unit spec** (`InternalStreamConnectionInitializerSpecification`): updated
  the 4 handshake expectations to `backpressure: "2"`.

### Evergreen (JAVA-6238)

Added a `9.0` entry to the `version` axis in `.evergreen/.evg.yml` and included `9.0`
in all test matrix variants that currently test against `8.0`, mirroring the Python
and Node drivers. This gives the 9.0-gated prose Test 5 a pinned CI variant. The
Atlas-specific `MONGODB_VERSION: "8.0"` deploy target is left untouched (Atlas
manages its own server version).

### DIFF
--- a/driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
@@ -70,6 +70,7 @@ import static java.util.Arrays.asList;
  */
 final class SpecRetryPolicy implements RetryPolicy {
     private static final int INFINITE_ATTEMPTS = Integer.MAX_VALUE;
+    private static final String BASE_BACKOFF_MS_FIELD = "baseBackoffMS";
 
     private final IndividualPolicies policies;
     private int maxAttempts;
@@ -300,10 +301,10 @@ final class SpecRetryPolicy implements RetryPolicy {
             return null;
         }
         BsonDocument response = ((MongoCommandException) attemptFailedResult).getResponse();
-        if (!response.containsKey("baseBackoffMS")) {
+        if (!response.containsKey(BASE_BACKOFF_MS_FIELD)) {
             return null;
         }
-        BsonValue value = response.get("baseBackoffMS");
+        BsonValue value = response.get(BASE_BACKOFF_MS_FIELD);
         if (!value.isNumber()) {
             return null;
         }
--- a/driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
+++ b/driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
@@ -16,6 +16,7 @@
 package com.mongodb.internal.operation;
 
 import com.mongodb.MongoClientSettings;
+import com.mongodb.MongoCommandException;
 import com.mongodb.MongoConnectionPoolClearedException;
 import com.mongodb.MongoException;
 import com.mongodb.MongoOperationTimeoutException;
@@ -32,6 +33,9 @@ import com.mongodb.internal.connection.OperationContext.ServerDeprioritization;
 import com.mongodb.internal.session.BaseClientSessionImpl;
 import com.mongodb.internal.time.ExponentialBackoff;
 import com.mongodb.lang.Nullable;
+
+import org.bson.BsonDocument;
+import org.bson.BsonValue;
 
 import java.time.Duration;
 import java.util.EnumMap;
@@ -285,9 +289,26 @@ final class SpecRetryPolicy implements RetryPolicy {
         assertFalse(attemptFailedResult instanceof OperationHelper.ResourceSupplierInternalException);
         if (attemptFailedResult instanceof MongoException
                 && ((MongoException) attemptFailedResult).hasErrorLabel(SYSTEM_OVERLOADED_ERROR_LABEL)) {
-            return ExponentialBackoff.calculateOverloadBackoff(immediateNextAttempt);
+            return ExponentialBackoff.calculateOverloadBackoff(immediateNextAttempt, extractBaseBackoffMs(attemptFailedResult));
         }
         return Duration.ZERO;
+    }
+
+    @Nullable
+    private static Long extractBaseBackoffMs(final Throwable attemptFailedResult) {
+        if (!(attemptFailedResult instanceof MongoCommandException)) {
+            return null;
+        }
+        BsonDocument response = ((MongoCommandException) attemptFailedResult).getResponse();
+        if (!response.containsKey("baseBackoffMS")) {
+            return null;
+        }
+        BsonValue value = response.get("baseBackoffMS");
+        if (!value.isNumber()) {
+            return null;
+        }
+        long parsed = value.asNumber().longValue();
+        return parsed > 0 ? parsed : null;
     }
 
     private static int maxAttempts(final int maxRetries) {
--- a/driver-sync/src/test/functional/com/mongodb/client/BackpressureProseTest.java
+++ b/driver-sync/src/test/functional/com/mongodb/client/BackpressureProseTest.java
@@ -137,6 +137,57 @@ public class BackpressureProseTest {
         }
     }
 
+    private static Duration measureFailedInsertDuration(final MongoCollection<Document> collection) {
+        StartTime startTime = StartTime.now();
+        assertThrows(MongoServerException.class, () -> collection.insertOne(Document.parse("{a: 1}")));
+        return startTime.elapsed();
+    }
+
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/client-backpressure/tests/README.md#test-5-overload-errors-with-basebackoffms-override-base-backoff">
+     * Test 5: Overload Errors with baseBackoffMS override base backoff</a>.
+     */
+    @Test
+    void overloadErrorsWithBaseBackoffMsOverrideBaseBackoff() throws InterruptedException {
+        assumeTrue(serverVersionAtLeast(9, 0));
+        BsonDocument configureFailPoint = BsonDocument.parse(
+                "{\n"
+                + "    configureFailPoint: 'failCommand',\n"
+                + "    mode: 'alwaysOn',\n"
+                + "    data: {\n"
+                + "        failCommands: ['insert'],\n"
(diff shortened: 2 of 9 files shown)


driver-sync/src/test/functional/com/mongodb/client/BackpressureProseTest.java
```diff
@@ -137,6 +137,57 @@ private static Duration measureFailedInsertDuration(final MongoCollection<Docume
...
 
+    private static Duration measureFailedInsertDuration(final MongoCollection<Document> collection) {
+        StartTime startTime = StartTime.now();
+        assertThrows(MongoServerException.class, () -> collection.insertOne(Document.parse("{a: 1}")));
+        return startTime.elapsed();
+    }
+
+    /**
+     * <a href="https://github.com/mongodb/specifications/blob/master/source/client-backpressure/tests/README.md#test-5-overload-errors-with-basebackoffms-override-base-backoff">
+     * Test 5: Overload Errors with baseBackoffMS override base backoff</a>.
```
jyemin, line 166: Let's put this after test 4 so they remain in order.

jyemin (changes requested):
All looking good to me.  Just one nit in the prose test ordering.

driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
```diff
@@ -570,6 +613,14 @@ static final class Overload extends State {
...
                 @Nullable
                 private BaseClientSessionImpl.OverloadRetryPolicyState sessionScopedState;
+                /**
+                 * Whether all failures observed so far within the current command execution are
+                 * retryable overload errors. {@code true} is also the case if no failures have been observed.
+                 *
+                 * @see MongoException#RETRYABLE_ERROR_LABEL
+                 * @see MongoException#SYSTEM_OVERLOADED_ERROR_LABEL
+                 */
+                private boolean observedNoneOrOnlyRetryableOverloadErrors;
```
stIncMale, line 623: I apologize for leaving questions without really reviewing the code. If they are stupid, please ignore.

1. Is it true that we can't reuse  `DefaultCommandExecutionScoped.observedNoneOrOnlyRetryableOverloadErrors` instead of introducing this new field?
2. If the answer to the above questions is "yes", maybe we then could use this new field instead of `DefaultCommandExecutionScoped.observedNoneOrOnlyRetryableOverloadErrors`? The duplication (at least, that is how it looks like at first glance) is concerning.

driver-sync/src/test/functional/com/mongodb/client/BackpressureProseTest.java
vbabanin: Good catch. I changed the order.

jyemin (dismissed):
LGTM

jyemin (approved):

driver-core/src/main/com/mongodb/internal/operation/SpecRetryPolicy.java
vbabanin: Update after offline discussion:

1. Technically, yes, we could reuse `observedNoneOrOnlyRetryableOverloadErrors`.
2. I don’t think we should do that in this PR. 

`observedNoneOrOnlyRetryableOverloadErrors` currently lives on the command-execution scope. Reusing it here would require exposing it through another getter, which goes in the opposite direction of the cleanup we discussed in https://github.com/mongodb/mongo-java-driver/pull/2041#discussion_r3833989334 / JAVA-6315: encapsulating `SessionContext` internals behind notify-style methods.

There are also two separate concerns mixed into this state:

- Whether the session should still attach startTransaction.
- Whether SpecRetryPolicy should check if the previously selected server is still eligible.

Ideally, `SpecRetryPolicy` should track retry-policy state and notify the session that it observed overload retryable error, while the session should mark that `startTransaction` is not needed anymore.

If we later decide that `observedNoneOrOnlyRetryableOverloadErrors` should span commands at the session level - for example, command A fails all attempts with retryable overload errors, then command B starts and we still need to decide whether to attach `startTransaction` - that should be handled inside the session scope not command scope. In that case, we should not have to modify `SpecRetryPolicy` because the session-side concern changed.

So for now, the lesser of two evils appears to be benign duplication. I added a note to JAVA-6315 that we should decide what to do with this field as part of the broader SessionContext notify-method refactoring.

stIncMale: @vbabanin Added the info to https://jira.mongodb.org/browse/JAVA-6315 and https://jira.mongodb.org/browse/JAVA-6316, so that we don't forget to refactor this if we ever take on such work.

P.S. All the follow-up tickets caused by the backpressure epic https://jira.mongodb.org/browse/JAVA-6019 are / will be linked as "caused by", so that we can easily find them via the epic ticket in its "causes" links.

PR #2054, opened 2026-09-15
JAVA-6235: Configurable DNS domain validation for SRV records

apmasell:
Add support for `srvAllowedHostsSuffix` using the public suffix list as a deny list as a safeguard and add docs explaining the use of this parameter (and its potential security implications when used).


driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -1355,6 +1371,22 @@ public String getSrvServiceName() {
...
 
+    /**
+     * Gets the SRV allowed hosts suffix.
+     *
+     * <p>If present, its value is used as the domain for SRV host name validation, replacing the domain inferred from
+     * the SRV host name. The value is normalized: a leading {@code "."} is prepended if absent, so the returned value
+     * always begins with {@code "."}.</p>
+     *
+     * @return the normalized SRV allowed hosts suffix, always beginning with {@code "."}. Defaults to null.
+     * @since 5.9
```
nhachicha, line 1382: The feature will probably be released in `5.13` (next minor)
```suggestion
     * @since 5.13
```

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
```diff
@@ -15,16 +15,103 @@
...
+            throw new IllegalArgumentException("srvAllowedHostsSuffix must not contain whitespace");
+        }
+        // A single leading '.' is allowed (it is the documented suffix form); everything after it must be one or more
+        // non-empty domain labels.
+        boolean hasLeadingDot = srvAllowedHostsSuffix.startsWith(".");
+        String labels = hasLeadingDot ? srvAllowedHostsSuffix.substring(1) : srvAllowedHostsSuffix;
+        if (labels.isEmpty()) {
+            throw new IllegalArgumentException("srvAllowedHostsSuffix must contain at least one domain label");
+        }
+        String[] parts = labels.split("\\.", -1);
```
nhachicha, line 58: I believe this is redundant since the check below `isDomainName` rejects `"build..10gen.cc"`  you can  keep it if you think this gives more clarity (specific error message)

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
```diff
@@ -15,16 +15,103 @@
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static boolean isPublicSuffix(final String suffix) {
+        try (Scanner scanner = new Scanner(
+                Objects.requireNonNull(
+                        DomainNameUtils.class.getResourceAsStream("public_suffix_list.dat"), "Missing DNS suffix list"),
```
nhachicha, line 86:
```suggestion
                        DomainNameUtils.class.getResourceAsStream("public_suffix_list.dat"), "Missing DNS public suffix list"),
```

nhachicha, line 86: Caught by Claude:

GraalVM: `public_suffix_list.dat` isn't registered as a resource

Native image drops unregistered classpath resources, so `getResourceAsStream("public_suffix_list.dat")` will NPE for anyone setting `srvAllowedHostsSuffix`. We hand-maintain this metadata. Add to `driver-core/src/main/resources/META-INF/native-image/resource-config.json`:
```
{ "pattern":"\\Qcom/mongodb/internal/connection/public_suffix_list.dat\\E" }
```

Note `graalvm-native-image-app-task` passes either way today — missing resources fail at runtime on access, not at build, and `NativeImageApp` never hits this path. Worth adding a small component beside `DnsSpi` that builds a `ConnectionString` with the option set (no server needed) so CI can actually catch this.

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
```diff
@@ -15,16 +15,103 @@
...
+                "UTF-8")) {
+            int firstDot = suffix.indexOf('.');
+            String rootDomain = firstDot >= 0 ? suffix.substring(firstDot + 1) : suffix;
+            boolean invalidMatchWildcard = false;
+            while (scanner.hasNextLine()) {
+                String line = scanner.nextLine();
+                if (line.startsWith("//") || line.isEmpty()) {
```
nhachicha, line 93: You shouldn't need this logic if we're using the [public_suffix_list.data](https://github.com/mongodb/specifications/blob/master/source/public-suffix-list/public_suffix_list.dat) [provisioned](https://github.com/mongodb/specifications/blob/master/source/public-suffix-list/public-suffix-list.md#obtaining-the-list) from spec

> Drivers SHOULD NOT fetch the list from the network at runtime, and SHOULD instead resolve it from a copy shipped with the driver.

driver-core/src/main/com/mongodb/connection/ClusterSettings.java
```diff
@@ -185,6 +189,26 @@ public Builder srvServiceName(final String srvServiceName) {
...
+         *
+         * <p>If set, its value is used as the domain for SRV host name validation, replacing the domain inferred from
+         * the SRV host name. The value is normalized: a leading {@code "."} is prepended if absent, so
+         * {@link #getSrvAllowedHostsSuffix()} always returns a value beginning with {@code "."}. This setting is only
+         * used with SRV. Specifying an overly broad suffix (for example a bare TLD) weakens SRV host name validation and
+         * is the responsibility of the caller.</p>
+         *
+         * @param srvAllowedHostsSuffix the SRV allowed hosts suffix; may not be null or empty
+         * @return this
+         * @since 5.9
```
nhachicha, line 203:
```suggestion
         * @since 5.13
```

driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
```diff
@@ -39,12 +58,50 @@ public void resetDefaultProviderUrl() {
...
+        DefaultDnsResolver resolver = resolverReturning("localhost.build.10gen.cc.");
+        MongoConfigurationException e = assertThrows(MongoConfigurationException.class,
+                () -> resolver.resolveHostFromSrvRecords(TEST_HOST, "mongodb", ".test.build.10gen.cc"));
+        assertTrue(e.getMessage().contains("srvAllowedHostsSuffix"));
+    }
+
+    @Test
+    public void shouldThrowWhenAnyResolvedHostDoesNotEndWithSrvAllowedHostsSuffix() {
+        DefaultDnsResolver resolver = resolverReturning("ok.build.10gen.cc.", "bad.evil.example.com.");
+        assertThrows(MongoConfigurationException.class,
```
nhachicha, line 96: should we assert the message include `bad.evil.example.com.`? Making sure the exception include it for investigation/Root Cause Analysis...

nhachicha (changes requested):
Thanks for taking this over! The implementation in general is ok, a couple of things left:

- 1. The official spec tests never run. The submodule isn't bumped.

- 2. `public_suffix_list.dat` The PSL spec says drivers MUST obtain it from the specifications repo, where it's pre-stripped of comments (partially my fault since I pointed you to https://github.com/mongodb/mongo-python-driver/pull/2903/ instead of https://github.com/mongodb/mongo-python-driver/pull/2868/ which contains the correct file)

- 3. Claude `/driver-code-review` caught some additional points, please assess and pushback / ignore if they're irrelevant
<details> 
<summary>Claude Opus (1M Context) review</summary>

### 🔴 Blocking

- **`DOMAIN_PATTERN` widening is a regression.** Loosening the TLD class from `[a-zA-Z]{2,63}` to `[a-zA-Z0-9\-]{2,63}` to admit `xn--` labels changes a **shared** method. `xn--frosch-6ya.w23` is an existing *negative* case in `DomainNameUtilsTest`, so that test now fails; and `SocksSocket:135` uses `isDomainName` for SOCKS5 proxy hosts, which now accepts `foo.123`, `foo.--`, `evil.-a-`. Suggest validating with the original pattern *before* IDN conversion, or special-casing the `xn--` prefix.
- **MPL-2.0 attribution missing.** The vendored list ships in the Apache-2.0 `driver-core` jar; neither `THIRD-PARTY-NOTICES` nor `sbom.json` is updated. Needs a notice entry and legal/human sign-off.

### 🟡 Important

- **Trailing/leading dot not stripped (spec step 1).** `.build.10gen.cc.` currently throws; the spec requires stripping, and `srvAllowedHostsSuffix-trailing-dot` expects success. The current test enshrines the wrong behaviour.
- **`isPublicSuffix` isn't the publicsuffix.org algorithm.** The `invalidMatchWildcard` latch converts to `return true` on the *next* ordinary line and is never reset, so results depend on upstream line ordering (it works only because `!www.ck` sits immediately after `*.ck`). Suggest a rule index + the documented algorithm: exceptions win, then longest match.
- **Per-call cost got worse.** `IDN.toASCII` now runs on every rule line — ~16.4k conversions per validation, doubled because `ConnectionString` and `ClusterSettings.Builder` both normalize. An ASCII fast-path plus a lazily-built index fixes both.
- **Misleading error messages.** `.com` fails with "is not a valid domain" (it trips `isDomainName` as a single label) rather than the public-suffix message; for bare TLDs the PSL check is effectively unreachable.
- **`ClusterSettings.Builder` javadoc still contradicts the code** (Copilot's comment, unaddressed) — says a bare TLD is the caller's responsibility, but public suffixes are rejected. The `ConnectionString` option docs also carry no security warning, which is the path most users take.
- **New Spock tests** in `ClusterSettingsSpecification.groovy` — `.agents/references/testing-guide.md`: "Do not add new Spock tests." Mechanical signature updates are fine; the new blocks should be JUnit 5.
- **No direct test for the reworked matcher.** `DomainNameUtilsTest` already exists — add uppercase, multi-label public suffixes, wildcard/exception ordering, U-label vs A-label, and missing-resource cases.
- **Out-of-scope behaviour changes.** `applySrvConnectionStringOptions` now applies `srvServiceName` on the LOAD_BALANCED path for the first time (genuine fix, but unrelated), and the unwrapped `MongoConfigurationException` rethrow changes messages on the pre-existing inferred-domain path. Please split or call out in the description.
- **`ClusterSettings.build()` doesn't reject the option without SRV**, unlike the URI path.

### 🟢 Nit

- `assertNotEquals(first.hashCode(), second.hashCode())` asserts more than the `hashCode` contract (Copilot's comment, unaddressed).
- `DOMAIN_PATTERN` allows `localhost`, so `srvAllowedHostsSuffix=localhost` slips past the ≥2-label rule.
- Duplicated comment on the two wildcard tests — inaccurate on the first.
- Stray double blank line in `DefaultDnsResolverTest`; over-length lines in `LoadBalancedClusterTest`.
- `Builder.srvAllowedHostsSuffix` can't be reset to null; field lacks `@Nullable` though the getter has it.
</details>

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
```diff
@@ -15,16 +15,103 @@
...
+     * consists only of a leading {@code "."}), contains an empty domain label (consecutive {@code "."} characters
+     * or a trailing {@code "."}), or, is a public domain suffix (e.g., top-level domain, AWS data center)
+     */
+    public static String normalizeSrvAllowedHostsSuffix(final String srvAllowedHostsSuffix) {
+        if (containsWhitespace(srvAllowedHostsSuffix)) {
+            throw new IllegalArgumentException("srvAllowedHostsSuffix must not contain whitespace");
+        }
+        // A single leading '.' is allowed (it is the documented suffix form); everything after it must be one or more
+        // non-empty domain labels.
+        boolean hasLeadingDot = srvAllowedHostsSuffix.startsWith(".");
```
nhachicha, line 65: [srvAllowedHostsSuffix-trailing-dot.json](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/tests/replica-set/srvAllowedHostsSuffix-trailing-dot.json) is failing on all 16 variants in the [latest CI run](https://parsley.corp.mongodb.com/test/mongo_java_driver_tests_jdk8_unsecure__version~4.2_os~linux_topology~replicaset_auth~noauth_ssl~nossl_jdk~jdk8_test_sync_task_patch_be53a3e117717154f8d54ff3301efae37dce22d6_6aad68e6a3aa6800076b751d_26_09_18_16_38_00/0/3d1678f7024df03550e4d465c1b21d82?bookmarks=0,47&shareLine=0). Per [spec step 1](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md?plain=1#L77), a trailing `.` is legal and must be stripped, not rejected.

Note: this will break `ConnectionStringUnitTest.shouldThrowWhenSrvAllowedHostsSuffixHasTrailingDot`  which currently asserts the opposite

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
```diff
@@ -15,16 +15,103 @@
...
+import java.util.Scanner;
 import java.util.regex.Pattern;
 
 /**
  * <p>This class is not part of the public API and may be removed or changed at any time</p>
  */
 public class DomainNameUtils {
-    private static final Pattern DOMAIN_PATTERN =
-            Pattern.compile("^(?=.{1,255}$)((([a-zA-Z0-9]([a-zA-Z0-9\\-]{0,61}[a-zA-Z0-9])?\\.)+[a-zA-Z]{2,63}|localhost))$");
+    private static final Pattern DOMAIN_PATTERN = Pattern.compile(
```
nhachicha, line 32: - The extra escape puts a literal backslash in the final label's class, so `a\b` and `evil.com\x` pass `isDomainName` (leading labels use the correct class, so `a\b.com` is still rejected). Fix: final group \\\\- → \\-.

- And `*` (correctly) allows single labels for the proxy path, but `normalizeSrvAllowedHostsSuffix` still routes through `isDomainName`, so `srvAllowedHostsSuffix=internal` is now accepted, this needs an extra check in `normalizeSrvAllowedHostsSuffix`

nhachicha (changes requested):
There are still outstanding [comments](https://github.com/mongodb/mongo-java-driver/pull/2054#pullrequestreview-5234739568) to be fixed.

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
apmasell: Fixed point 1, but I am unclear what you're asking in the second point.

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
apmasell: Fixed.

driver-core/src/test/unit/com/mongodb/internal/dns/DefaultDnsResolverTest.java
apmasell: Sure.

driver-core/src/main/com/mongodb/connection/ClusterSettings.java
apmasell: Updated.

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
apmasell: Added with a build rule to copy it from the specifications repository.

driver-core/src/main/com/mongodb/ConnectionString.java
apmasell: Updated.

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
apmasell: Agreed. I was trying to leave the original checks intact.

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
apmasell: Updated.

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
nhachicha: Per spec [step 4](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md#srvallowedhostssuffix) a single label is only valid if it's one of nine allowlisted names, and step 5 skips the public-suffix check for those,  we implement neither, and since the pattern became `(label\.)*label` (`*`, correctly, so the proxy path accepts `mongo`/`localhost`), single labels now pass ex (`mycorp`, `intranet` are currently accepted).

possible fix 
```Java
private static final Set<String> VALID_SINGLE_LABEL_SUFFIXES = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
        // RFC 6761 special use names
        "test", "localhost", "invalid", "example",
        // RFC 6762 multicast DNS
        "local",
        // reserved by ICANN for private use
        "internal",
        // not reserved by ICANN, but commonly used privately
        "corp", "home", "mail")));
...
// step 4
boolean allowedSingleLabel = VALID_SINGLE_LABEL_SUFFIXES.contains(labels);
if (labels.indexOf('.') < 0 && !allowedSingleLabel) {
    throw new IllegalArgumentException("srvAllowedHostsSuffix must contain at least two domain labels");
}
// step 5 
if (!allowedSingleLabel && isPublicSuffix(labels)) {
    throw new IllegalArgumentException("srvAllowedHostsSuffix must not be a public domain suffix");
}

```

driver-core/build.gradle.kts
```diff
@@ -66,6 +66,13 @@ dependencies {
     }
 }
 
+tasks.processResources {
+    from(
```
nhachicha: Caught by claude-review:

Gradle's `from()` treats a non-existent source as an empty copy spec, so if the `testing/resources/specifications` submodule isn't initialised the build succeeds and silently ships a jar with no `public_suffix_list.dat` => every `srvAllowedHostsSuffix` will NPE

Possible fix:
```Kotlin
val publicSuffixList = rootProject.layout.projectDirectory.file(
    "testing/resources/specifications/source/public-suffix-list/public_suffix_list.dat")

tasks.processResources {
    inputs.file(publicSuffixList).withPropertyName("publicSuffixList")
    from(publicSuffixList) {
        into("com/mongodb/internal/connection")
    }
}
```

driver-core/src/main/com/mongodb/internal/connection/SocksSocket.java
```diff
@@ -164,12 +164,14 @@ private void sendConnect(final Timeout timeout) throws IOException {
         final int hostLength = bytesOfHost.length;
 
         AddressType addressType;
-        byte[] ipAddress = null;
-        if (isDomainName(host)) {
+        byte[] ipAddress = InetAddressUtils.ipStringToBytes(host);
```
nhachicha, line 167: Caught via Claude review:

The IP-first reorder is needed with the relaxed DOMAIN_PATTERN, but malformed IP literals like 999.999.999.999 now match as domain names and get forwarded to the proxy, instead of failing fast locally with SocketException as before.

Per [RFC 3696 §2](https://www.rfc-editor.org/rfc/rfc3696#section-2) a TLD is never all-numeric — so a simple guard restores the old behavior without losing single-label hosts or punycode TLDs:

```java
} else if (isDomainName(host) && !lastLabelAllNumeric(host)) {
    addressType = DOMAIN_NAME;
} else {
    throw new SocketException(host + IP_PARSING_ERROR_SUFFIX);
}

private static boolean lastLabelAllNumeric(final String host) {
    String lastLabel = host.substring(host.lastIndexOf('.') + 1);
    return !lastLabel.isEmpty() && lastLabel.chars().allMatch(Character::isDigit);
}
```
Rejects `999.999.999.999` / `foo.123`; keeps `myhost`, `mongo.xn--3pxu8k`, `xn--frosch-6ya.w23`. A test pinning these cases would also make the IP-first ordering contract explicit.

nhachicha (changes requested):

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
```diff
@@ -27,4 +29,81 @@ public class DomainNameUtils {
...
+        if (labels.isEmpty()) {
+            throw new IllegalArgumentException("srvAllowedHostsSuffix must contain at least one domain label");
+        }
+        String[] parts = labels.split("\\.", -1);
+        for (String label : parts) {
+            if (label.isEmpty()) {
+                throw new IllegalArgumentException("srvAllowedHostsSuffix must not contain empty domain labels");
+            }
+        }
+        if (isTopLevelDomain(hasLeadingDot ? srvAllowedHostsSuffix.substring(1) : srvAllowedHostsSuffix)) {
```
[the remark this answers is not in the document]
apmasell: I think this is outdated.

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
```diff
@@ -27,4 +29,81 @@ public class DomainNameUtils {
...
+            throw new IllegalArgumentException("srvAllowedHostsSuffix must not be a top-level domain");
```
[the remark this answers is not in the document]
apmasell: Outdated.

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
```diff
@@ -27,4 +29,81 @@ public class DomainNameUtils {
     static boolean isDomainName(final String domainName) {
         return DOMAIN_PATTERN.matcher(domainName).matches();
     }
+
+    /**
+     * Validates and normalizes a {@code srvAllowedHostsSuffix} value for use as the domain in SRV host validation.
+     * A leading {@code "."} is prepended if absent, so the returned value always begins with {@code "."}; this is what
+     * is stored and returned to callers. The suffix must contain at least one non-empty domain label and no
+     * whitespace; an overly broad suffix is permitted and is the caller's responsibility.
```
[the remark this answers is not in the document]
apmasell: Outdated.

nhachicha:
`MPL-2.0` attribution missing. `public_suffix_list.dat` is now copied from the specifications submodule at build time, but it still ships inside the Apache-2.0 driver-core jar. `THIRD-PARTY-NOTICES` doesn't have an entry.

driver-core/src/test/unit/com/mongodb/ConnectionStringUnitTest.java
```diff
@@ -84,6 +84,14 @@ void equalAndHashCode(final String connectionString) {
         );
     }
 
+    @Test
+    void shouldNotBeEqualWhenSrvAllowedHostsSuffixDiffers() {
+        ConnectionString first = new ConnectionString("mongodb+srv://test12.test.build.10gen.cc/?srvAllowedHostsSuffix=.build.10gen.cc");
+        ConnectionString second = new ConnectionString("mongodb+srv://test12.test.build.10gen.cc/?srvAllowedHostsSuffix=.other.10gen.cc");
+        assertNotEquals(first, second);
+        assertNotEquals(first.hashCode(), second.hashCode());
```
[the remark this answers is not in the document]
apmasell: I assume Ross had a reason for this, but I think it should be removed.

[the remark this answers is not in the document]
apmasell: Removed.

apmasell:
We're removing the PSL, so a lot of the comments are no longer applicable. I am documenting in the ticket.

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -266,6 +267,8 @@
  * <ul>
  * <li>{@code srvServiceName=string}: The SRV service name. See {@link ClusterSettings#getSrvServiceName()} for details.</li>
  * <li>{@code srvMaxHosts=n}: The maximum number of hosts from the SRV record to connect to.</li>
+ * <li>{@code srvAllowedHostsSuffix=string}: The hostname suffix used to validate hosts returned via SRV lookup, replacing the domain
```
nhachicha, line 270: The [spec](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md#srvallowedhostssuffix) requires a clearly visible warning: 

> Notably, srvAllowedHostsSuffix relaxes existing security measures and must be used with caution. Thus, drivers MUST document that this parameter is a dangerous option. For example, something like "WARNING: Modifying the default SRV domain name validation can create vulnerabilities." should be clearly visible in the documentation.

Suggest adding the above plus a sentence that multi-label public suffixes are not rejected see [C#](https://github.com/mongodb/mongo-csharp-driver/pull/2109/changes#diff-bb6e5ee1c6a4e0689c10b24183b0aff018244cfa92924278a39c0b66265905d9R71) and [Python](https://github.com/mongodb/mongo-python-driver/pull/2868/changes#diff-a3d1d8e2a95339043c3bbeb634677606fffa1548d45c77cca40f31d804d64c51R457).

driver-core/src/main/com/mongodb/connection/ClusterSettings.java
```diff
@@ -185,6 +189,26 @@ public Builder srvServiceName(final String srvServiceName) {
             return this;
         }
 
+        /**
+         * Sets the SRV allowed hosts suffix used to validate hosts returned via SRV lookup.
```
nhachicha, line 193: [ditto](https://github.com/mongodb/mongo-java-driver/pull/2054/changes#r4091931360)

driver-core/src/main/com/mongodb/internal/connection/DomainNameUtils.java
```diff
@@ -15,16 +15,81 @@
...
+     * Validates and normalizes a {@code srvAllowedHostsSuffix} value for use as the domain in SRV host validation.
+     * A leading {@code "."} is prepended if absent, so the returned value always begins with {@code "."}; this is what
+     * is stored and returned to callers. The suffix must contain at least one non-empty domain label and no
+     * whitespace. An overly broad suffix is restricted based on a public suffix list.
+     *
+     * @param srvAllowedHostsSuffix the non-null suffix to validate
+     * @return the normalized suffix, always beginning with {@code "."}
+     * @throws IllegalArgumentException if the suffix contains whitespace, contains no domain label (it is empty or
+     * consists only of a leading {@code "."}), contains an empty domain label (consecutive {@code "."} characters
+     * or a trailing {@code "."}), or, is a public domain suffix (e.g., top-level domain, AWS data center)
```
nhachicha, line 57: Javadoc is stale after the PSL removal: "An overly broad suffix is restricted based on a public suffix list" and the @throws clause "…is a public domain suffix (e.g., top-level domain, AWS data center)" describe a check that no longer exists. Suggest rewording to the label-count rule actually enforced

driver-core/src/main/com/mongodb/connection/ClusterSettings.java
```diff
@@ -421,6 +456,21 @@ public String getSrvServiceName() {
...
     }
 
+    /**
+     * Gets the SRV allowed hosts suffix used to validate hosts returned via SRV lookup.
+     *
+     * <p>If present, its value is used as the domain for SRV host name validation, replacing the domain inferred from
+     * the SRV host name. The value is normalized to always begin with {@code "."}.</p>
+     *
+     * @return the normalized SRV allowed hosts suffix, always beginning with {@code "."}. Defaults to null.
+     * @since 5.9
```
nhachicha, line 466:
```suggestion
     * @since 5.13
```

nhachicha (commented):
Minor things left 👍

nhachicha (approved):
LGTM 🎉

driver-core/src/main/com/mongodb/connection/ClusterSettings.java
```diff
@@ -185,6 +189,26 @@ public Builder srvServiceName(final String srvServiceName) {
...
+         * {@link #getSrvAllowedHostsSuffix()} always returns a value beginning with {@code "."}. This setting is only
+         * used with SRV. Specifying an overly broad suffix (for example a bare TLD) weakens SRV host name validation and
+         * is the responsibility of the caller.</p>
+         *
+         * @param srvAllowedHostsSuffix the SRV allowed hosts suffix; may not be null or empty
+         * @return this
+         * @since 5.13
+         * @see #getSrvAllowedHostsSuffix()
+         */
+        public Builder srvAllowedHostsSuffix(final String srvAllowedHostsSuffix) {
```
stIncMale: We also need to add this new client setting to `MongoClientOptions`.

driver-core/src/main/com/mongodb/ConnectionString.java
```diff
@@ -266,6 +266,13 @@
...
```
stIncMale: We also need to add this new connection string option to `MongoClientURI`.

driver-sync/src/test/functional/com/mongodb/client/InitialDnsSeedlistDiscoveryTest.java
```diff
@@ -97,21 +96,6 @@ public InitialDnsSeedlistDiscoveryTest(final String filename, final String paren
...
         this.executePingCommand = executePingCommand;
-
-        // TODO-JAVA-6235 https://jira.mongodb.org/browse/JAVA-6235
-        assumeFalse(Arrays.asList(
-                        "srvAllowedHostsSuffix-case-insensitive.json",
-                        "srvAllowedHostsSuffix-mismatch.json",
-                        "srvAllowedHostsSuffix-period-only.json",
-                        "srvAllowedHostsSuffix-psl-not-public-suffix.json",
-                        "srvAllowedHostsSuffix-psl-public-suffix-capitalized.json",
-                        "srvAllowedHostsSuffix-psl-public-suffix.json",
```
stIncMale: [just a note] I was surprised `srvAllowedHostsSuffix-psl-public-suffix.json` does not cause build failures. But it indeed does run and pass, because it uses `srvAllowedHostsSuffix=cc`, which is invalid regardless of whether a driver implements PSL validation or not. I find this specification test weird, as it does not check anything PSL-related. The same goes about `srvAllowedHostsSuffix-psl-public-suffix-capitalized.json`.

stIncMale (changes requested):
I glanced over the PR (that is not a review in any way), and I think there are changes needed before we can release.

PR #2062, opened 2026-09-18
JAVA-6312 Increase testConnectionPoolBackpressure completion time on CI

nhachicha:
https://jira.mongodb.org/browse/JAVA-6312

`testConnectionPoolBackpressure` storms 100 `find({$where: sleep(2000)})` operations and asserts they complete within a deadline. The previous 20s deadline fails on MongoDB 9.0, where concurrent JS `sleep()` is slow under contention — a known server issue, [SERVER-134156](https://jira.mongodb.org/browse/SERVER-134156) (fix on master; backport to 9.0 undecided). Investigation: [HELP-100331](https://jira.mongodb.org/browse/HELP-100331).

## Measured completion times on CI ([patch build](https://spruce.corp.mongodb.com/version/6aad1938db012900070c77b1))

| Server version | Variants | Test duration |
|---|---|---|
| 7.0 / 8.0 / latest (noauth/nossl) | all topologies | 15.1–15.5s |
| 7.0 / 8.0 (auth/ssl) | all topologies | 17.2–17.5s |
| 9.0 (noauth/nossl) | all topologies | 30.5–31.0s |
| 9.0 (auth/ssl) | standalone, replica set | 35.1–35.5s |
| 9.0 (auth/ssl) | sharded | **44.2s** |

## Change

Bump the deadline to **90s**: the measured worst case is 44.2s, so 90s gives ~2× headroom over the slowest legitimate run while still catching genuine hangs (the old 20s had only ~2.5s headroom even against healthy servers). The test's spec assertions (≥10 `ConnectionCheckOutFailedEvent`, 0 `PoolClearedEvent`) are unchanged and pass on all versions; a completion-time log line is kept so future drift is visible in CI logs.


nhachicha:
## Measured completion times ([patch build](https://spruce.corp.mongodb.com/version/6aad1938db012900070c77b1))

| Server version | Variants | Test duration |
|---|---|---|
| 7.0 (auth/ssl) | standalone, replica set, sharded | 17.2–17.3s |
| 8.0 (auth/ssl) | standalone, replica set, sharded | 17.3–17.5s |
| 7.0 / 8.0 (noauth/nossl) | all three topologies | 15.1s |
| latest (noauth/nossl) | all three topologies | 15.4–15.5s |

All 15 variants that run the test passed; durations are remarkably consistent across topologies (±0.3s), with auth/ssl adding ~2s.

**Why a new value:** even against healthy servers the storm takes 15–17.5s on CI hosts, so the previous 20s deadline had only ~2.5–5s of headroom — one slow host away from flaking. Against MongoDB 9.0 it extrapolates to ~25–35s because concurrent JS `sleep()` is slow under contention ([SERVER-134156](https://jira.mongodb.org/browse/SERVER-134156), investigation in [HELP-100331](https://jira.mongodb.org/browse/HELP-100331)) — which is what made the 9.0 variants fail outright. (This build has no 9.0 variants; the 9.0 estimate comes from DEBUG-instrumented runs on the backpressure branch.)

**Change:** deadline set to **60s** — ~2–3× headroom over the slowest legitimate run (9.0), while still failing fast on genuine hangs. The completion-time log line is kept so future regressions are visible in CI logs.

nhachicha:
**Correction to the comment above:** this build *did* include 9.0 variants (my earlier scan dropped them). The test **passed on all six 9.0 variants** under the measurement ceiling, with these durations:

| 9.0 variant | Test duration |
|---|---|
| noauth/nossl — standalone / replica set / sharded | 30.5–31.0s |
| auth/ssl — standalone / replica set | 35.1–35.5s |
| auth/ssl — sharded | **44.2s** |

So the measured worst case is **44.2s** (9.0, auth/ssl, sharded) vs 15–17.5s on 7.0/8.0/latest — confirming the ~2× JS `sleep()` contention penalty ([SERVER-134156](https://jira.mongodb.org/browse/SERVER-134156)) plus auth/ssl/sharded overhead.

⚠️ That means **60s gives only ~1.36× headroom** over the worst legitimate run, not the 2–3× stated above. Suggest **90s** (~2× headroom) unless SERVER-134156 is expected to be backported to 9.0.x soon.

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -314,8 +314,16 @@ public void testConnectionPoolBackpressure() throws InterruptedException {
...
                     executor.shutdown();
-                    assertTrue("Executor did not terminate within timeout",
-                            executor.awaitTermination(20, SECONDS));
+                    // The operation storm takes 15-17.5s on CI hosts even against healthy servers, and
+                    // ~25-35s against MongoDB 9.0, whose concurrent JS sleep() is slow under contention
+                    // (SERVER-134156, see HELP-100331) — so the previous 20s deadline had almost no
+                    // headroom. The measured worst case is 44.2s (9.0, auth/ssl, sharded), so 90s gives ~2x headroom
+                    // while still catching real hangs.
+                    long startMillis = System.currentTimeMillis();
+                    boolean terminated = executor.awaitTermination(90, SECONDS);
```
stIncMale, line 323: Judging from https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.md#connection-pool-backpressure, the original `awaitTermination(20, SECONDS)` was not only arbitrary, but not required by the prose test. It's just us trying to make sure the test does not hand forever.

If my understanding above is correct, then all we need to do is to change 20 -> 90 without any comment - neither the prose of the test, no any reader has an expectation for the wait or its timeout. We have the tickets filed, they are even mentioned in the PR description. But none of that information seems to be needed as a comment in the code we store in the repository.

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -314,8 +314,16 @@ public void testConnectionPoolBackpressure() throws InterruptedException {
...
                     }
                     executor.shutdown();
-                    assertTrue("Executor did not terminate within timeout",
-                            executor.awaitTermination(20, SECONDS));
+                    // The operation storm takes 15-17.5s on CI hosts even against healthy servers, and
+                    // ~25-35s against MongoDB 9.0, whose concurrent JS sleep() is slow under contention
+                    // (SERVER-134156, see HELP-100331) — so the previous 20s deadline had almost no
+                    // headroom. The measured worst case is 44.2s (9.0, auth/ssl, sharded), so 90s gives ~2x headroom
+                    // while still catching real hangs.
+                    long startMillis = System.currentTimeMillis();
```
stIncMale, line 322: `System.currentTimeMillis` is not monotonic and it is incorrect to use it for measuring time intervals. Let's use `StartTime startTime = StartTime.now()` and `startTime.elapsed()`, which exist specifically for this purpose.

P.S. `StartTime` does use the monotonic `System.nanoTime()` via `SystemNanoTime.get()`, which furthermore allows static mocking that is sometimes useful in tests.

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
```diff
@@ -314,8 +314,16 @@ public void testConnectionPoolBackpressure() throws InterruptedException {
...
+                    boolean terminated = executor.awaitTermination(90, SECONDS);
+                    LOGGER.info("testConnectionPoolBackpressure: operations completed="
```
stIncMale, line 324: I am surprised to see logging from tests, especially the performance-optimized logging code like the one below:
```java
if (LOGGER.isInfoEnabled()) {
    LOGGER.info("Polled " + element);
}
```

We do not usually log from tests, and here in particular, is exception with the relevant information not enough? What are the situations when we need to see an info-level log message about the executor shutting down successfully?

stIncMale (changes requested):

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
stIncMale: Actually, I realized that we don't need to measure the duration here at all. The timeout we give to `awaitTermination` is the duration we can and should report in the exception.

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
stIncMale: This is hysterical: I was the author of that performance-optimized logging test code. No words I can find, only emotions 💀

nhachicha: This was used to get the previous completion measurements. I can remove it.

driver-sync/src/test/functional/com/mongodb/client/ServerDiscoveryAndMonitoringProseTests.java
nhachicha: We can remove the extra logging it was used to measure completion on various platform

stIncMale (approved):

